PR #1772, opened 2019-01-03
MSC1772: Matrix spaces

ara4n:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/msc1772/proposals/1772-groups-as-rooms.md)

N.B. group access control has been split out into #2962

Obsoletes #1215

PR #3784, opened 2022-04-25
MSC3784: Using room type of `m.policy` for policy rooms

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-doc/blob/FSG-Cat-Policy-List-Room-Type/proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md)

This simple MSC aims to help stakeholders know if its likely to be a room containing a policy list or not. 

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/3784#issuecomment-4114179097)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3784#issuecomment-4114179807)


proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,63 @@
+# MSC3784 Using room type of m.policy for policy rooms
```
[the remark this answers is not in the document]
FSG-Cat: Well the way i personally see them used is as dedicated rooms mostly. This proposal aims to help that usecase but is not mutually exclusive with the usecases you mentioned in my head. Yes its probably a bit of a misuse of the type attribute in the sense that policy lists can contain other information unlike the other examples.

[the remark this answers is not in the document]
Gnuxie: >because there's zero practical reason to do so

@turt2live You're being a bit too certain about this :disappointed:. 

>, it was originally intended that Matrix HQ itself would be the "ban list" plus regular chat, making it easy for others to subscribe to the list and see what was going on. 

Original intent is fair to raise for context, but this isn't how policies are used, and it would be bad to use Matrix HQ this way. Not only is HQ a difficult room to query room state from because of the number of state events it has, but it's just not an appropriate place for them. 

>and see what was going on

Which would also be difficult because policy changes would be mixed with the chat timeline right?

> Policies are just arbitrary state events though and can live anywhere. It's very much a feature that rooms can contain multiple kinds of information when possible, which means a space which is also a policy room (for example).

This isn't how they are used though, even matrix.org has dedicated rooms with chat disabled.  

It makes a lot of sense to show these dedicated rooms differently in clients so that people can browse the rules and admins can edit them.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,63 @@
...
+
+This proposal is well quite simple. Allow for Policy rooms to be marked as such using the 
+`"type": "m.policy"`. The precedent for this comes from Spaces where they introduced this system in a
+limited capacity. This proposal expands this system to help all stakeholders quickly identify 
+policy rooms in a machine compatible way that is computationally cheap. It has additional benefits like
+allowing clients that are capable of editing policy to display editing tools for policy rooms when they
+detect that a room is a policy room using this mechanic. For machine interaction with policy rooms this
+proposal supplies a very fast way to tell if a room is definitively supposed to be a policy room or if 
+the user might have supplied a legacy room or typed in the wrong room ID / alias depending on how things
+are configured.
```
deepbluev7, line 25: To expand on this a bit, I really need something like this. Without this you run into a bootstrapping issue. Is this room supposed to be a policy room before the first policy is sent?

Without a room type you need to inspect the full room state to look for policy rules before you switch to the policy editor design. Slash commands can't warn the user properly that they are sending the rule into the wrong room.

So yes, I very much need this to provide a nice experience in my client. I can somewhat paper over it using account data, but it is not great. Everywhere we are told rooms are cheap. So far no rooms send policy rules into rooms you can send messages in. Most moderation tools even make it impossible to message in policy rooms, since that could possibly break the policy room when you actually need it to fight spam.

Thank you for writing this MSC!

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
+# MSC3784 Using room type of m.policy for policy rooms
```
FSG-Cat, line 1: Making this thread to track Implementation status outside of the PR descript. 

Implementation Tracking

Mjolnir as of [da08432](https://github.com/matrix-org/mjolnir/commit/da084328a9dbb12049dce8d8744b82711f8aa881) and that commit is included in [Release 1.6.0](https://github.com/matrix-org/mjolnir/releases/tag/v1.6.0)
Nheko in mtxclient has added initial support to recognise it in [4bd39bce](https://nheko.im/nheko-reborn/mtxclient/-/commit/4bd39bcedd6943ebd87faeac851bbf29b1167c07). Further progres unknown at time of writing.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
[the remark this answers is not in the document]
clokep: Sounds like this makes a good argument and this info just needs to make its way back into the MSC of why the current status quo isn't enough (or just isn't used). It is usually good to reference historical context too.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
clokep: Does Nheko use it besides just declaring the constant?

FSG-Cat: Policy Editing that utilises MSC3784 to aid in detecting if a room is a policy list is implemented in RMU. Version that implemented first is unknown but current Version as of writing relies on this as one of its 2 detection methods. It also supports legacy detection.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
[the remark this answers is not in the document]
turt2live: considering resolved: a few years later, it feels best to specify that policy lists *should* have the room type, but don't have to.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+
+This simple MSC aims to make it easier for machines and people alike to be made instantly aware if a 
+room is a policy list. To facilitate this this MSC recommends using the already established `type` 
+attribute to flag rooms as policy rooms. This builds upon the precedent set by Spaces and their 
+`"type": "m.space"`. Adopting this for Policy rooms allows all stakeholders to instantly know if its 
+a Policy room or not for rooms covered by this proposal. 
+
+For the purposes of this MSC policy lists can at times be called a policy room. This MSC does not intend 
+to change the name of the feature its just used to be very clear that we are talking about rooms with a 
+specific usecase.
```
turt2live: Introductions for MSCs ideally contain:
* Links to existing spec, where possible.
* A brief description of what needs fixing/improving.
* An even briefer description of what changes are being made.

We should also aim to avoid claims like "this MSC is simple" because it may not be by some metrics. 

```suggestion
Matrix already has [Moderation Policy Lists](https://spec.matrix.org/v1.17/client-server-api/#moderation-policy-lists)
which describe policy recommendations through state events. In typical usage,
these policy recommendations tend to appear in dedicated rooms, though are not 
required to. 

Having the policy recommendations in dedicated rooms helps clients better understand
the purpose of a room, and potentially hide it if the client knows it can't do
much with that dedicated purpose. For example, a messaging client might want to
hide policy list rooms by default to avoid showing "non-conversational" rooms to 
the user.

This proposal introduces a [room type](https://spec.matrix.org/v1.17/client-server-api/#types)
to denote such dedicated rooms.
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+the user might have supplied a legacy room or typed in the wrong room ID / alias depending on how things
+are configured. 
+
+Legacy rooms in this proposal are defined as all rooms that predate this proposal and they are free to upgrade
+if they desire to but they are also free to stay on their current room version and not upgrade their
+room to include a type event in the creation event. All stakeholders are expected to gracefully support 
+interacting with legacy rooms until a future proposal changes this recommendation. 
+
+The Author of this MSC believes that even with the problem of legacy rooms not being covered this MSC will
+be useful in the future with more use of policy rooms.
```
turt2live: With a strong introduction, we can also shorten the actual proposal body too. Modern MSCs ideally contain the minimum set of changes to the spec, and in some cases can pre-empt the Potential Issues section a bit too.


```suggestion
Rooms dedicated to containing policy list recommendations SHOULD use a newly
defined `m.policy` room type. How clients choose to (not) handle the new type is
left as an implementation detail. 

Consumers of policy lists SHOULD note that recommendations can still appear 
outside of `m.policy` rooms. Existing rooms probably won't have the new room 
type, and some communities might mix conversation and recommendations in the 
same room (therefore not dedicating the room to recommendations).
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+interacting with legacy rooms until a future proposal changes this recommendation. 
+
+The Author of this MSC believes that even with the problem of legacy rooms not being covered this MSC will
+be useful in the future with more use of policy rooms. 
+
+
+## Potential issues
+
+As is covered in the proposal section this MSC has the potential issue relating to Legacy rooms.
+The Author of this MSC thinks its that its an acceptable trade-off.
```
turt2live: We can also expand the potential issues to touch on some implementation considerations, especially because we alluded to it in the introduction and because the MSC has received comments from client authors wanting to have dedicated UX.


```suggestion
Clients which have dedicated UX for policy rooms might not be able to apply that
UX to rooms missing the `m.policy` room type. They can try to look for recommendation
state events contained in the room, though if their dedicated UX is invasive then
it may hide conversation being had in the room. This proposal does not attempt to
make rooms without the `m.policy` room type, but still contain policy recommendations,
invoke any dedicated UX a client might have - clients can choose how/if they handle
this case.
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+As is covered in the proposal section this MSC has the potential issue relating to Legacy rooms.
+The Author of this MSC thinks its that its an acceptable trade-off. 
+
+
+## Alternatives
+
+The alternative of using Peeking over fed to try to figure out if a room is a policy list has been
+considered but considering that its currently considered at the time of writing to be a very in development
+technology its deemed wholly unsuited for this application. The system this proposal recommends is the same
+one that was used for spaces. Spaces have proven them self's in the real world and so has this system.
```
turt2live: This just uses the root word "consider" too many times.


```suggestion
Peeking over federation, like in [MSC2444](https://github.com/matrix-org/matrix-spec-proposals/pull/2444),
can help clients identify rooms which contain policy list recommendation state
events. However, as identified above, this may be insufficient to properly
identify rooms *dedicated* to containing those state events.
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Security considerations
+
+None that the author is aware of at the time of writing due to the mostly informative nature of the data.
+Due to that no client should trust that just because it says m.policy it is indeed a policy room the security
+implications should be minimal.
```
turt2live: I think this is trying to say something like:


```suggestion
This change is largely informative and carries no direct security impact. Clients 
which interpret the `m.policy` room type will need to consider security in their
implementations. For example, if hiding the rooms then notifications in those 
rooms will need some consideration.
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Unstable prefix
+
+If you want to implement this MSC before its merged your free to use the unstable type of 
+`support.feline.policy.lists.msc.v1`.
```
turt2live:
```suggestion
If you want to implement this MSC before its merged you're free to use the unstable type of 
`support.feline.policy.lists.msc.v1`. 
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+Due to that no client should trust that just because it says m.policy it is indeed a policy room the security
+implications should be minimal.
+
+## Unstable prefix
+
+If you want to implement this MSC before its merged your free to use the unstable type of 
+`support.feline.policy.lists.msc.v1`. 
+
+After this MSC gets merged if a stakeholder has elected to remove its support for the unstable prefix if any
+support existed or support never existed revert to Legacy room behavior for rooms that used the unstable prefix.
```
turt2live:
```suggestion
Because the room type is immutable, rooms which use the unstable room type might
find themselves "unsupported" when/if implementations drop support for the unstable
type when this proposal becomes stable.
```

turt2live (approved):
My general feeling is this is something we should have in the spec to identify rooms dedicated to containing policy list recommendations. There might still be some rooms which want to mix recommendations with conversation - those rooms wouldn't use this type and might not get the benefits of dedicated UX from a client, but that's okay.

I've given this MSC a sizable editorial review. It's not my intention to completely change how this proposal works, so if that's happened then please correct me 😅. The editorial comments here reflect that ~650 MSCs have been opened in the time since this one was, and that modern MSCs are typically held to a different standard than older ones. I've also tried to make some areas a bit more clear & concise, as would happen in normal review.

My approval denotes that the MSC is ready for proposed-FCP with the edits (or something reasonably close). The listed implementations appear to satisfy the requirements we'd have for this proposal.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
TheArcaneBrony: https://mru.rory.gay/PolicyLists is the particular page in RMU that uses this MSC to find policy rooms. I'm not sure whether it has legacy detection enabled or not, but legacy detection here was implemented by checking for the Mjolnir shortcode state event. It does not show rooms that *happen* to have policy events, as querying /sync or /state for rooms is very expensive, especially compared to cherry picking 2 events.

(Oops, I still have to get rid of the "Create policy list" popup being there by default, which also uses the MSC's unstable type on room creation, additionally I've just uncovered a crash related to fetching room names... That's for the tomorrow pile).

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,68 @@
...
+to denote such dedicated rooms.
+
+## Proposal
+
+Rooms dedicated to containing policy list recommendations SHOULD use a newly
+defined `m.policy` room type. How clients choose to (not) handle the new type is
+left as an implementation detail. 
+
+Consumers of policy lists SHOULD note that recommendations can still appear 
+outside of `m.policy` rooms. Existing rooms probably won't have the new room
```
Johennes: SHOULD feels slightly odd to me here because consumers have complete freedom in how they handle (or don't handle) policy recommendations. There is nothing normative here and I think this should rather be a remark.


```suggestion
Consumers of policy lists should note that recommendations can still appear 
outside of `m.policy` rooms. Existing rooms probably won't have the new room 
```

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,68 @@
...
+can help clients identify rooms which contain policy list recommendation state
+events. However, as identified above, this may be insufficient to properly
+identify rooms *dedicated* to containing those state events.
+
+## Security considerations
+
+This change is largely informative and carries no direct security impact. Clients 
+which interpret the `m.policy` room type will need to consider security in their
+implementations. For example, if hiding the rooms then notifications in those 
+rooms will need some consideration.
```
Johennes: Missing notifications feels like a potential issue rather than a security consideration to me. Should there be a note in the proposal section that `m.policy` rooms can still include normal conversation? We already have the complementary remark there that regular rooms can still contain policy lists.

turt2live: Hidden/unobtainable notifications are occasionally considered a security concern in an implementation because of the disruption they cause. That's the only reason to mention it here.

`m.policy` rooms should *not* contain conversation, per the MSC.

Johennes: > `m.policy` rooms should _not_ contain conversation, per the MSC.

The proposal doesn't seem to explicitly say that. It speaks of "rooms dedicated to containing policy list recommendations" which may intend to say just that but this isn't very clear in my opinion.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
FSG-Cat: This part may be a bit wierdly worded due to being translated from the old legacy rooms handling that the MSC defined. And yes a strong but optional world like SHOULD is correct to the intent.

The intended way this is meant to be handled is that policy manipulation tooling should allow you to override the automatic detection the room type gives you and instead use manual detection. Be that temporary as in not persisted or persisted personal overrides stored via implementation specific means.

Secondary detection measures like checking for shortcode is also allowed but as @TheArcaneBrony can testify to its cludgy compared to pure room type detection.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
FSG-Cat: There is no enforcement what so ever so your client should NOT trust that nobody in the long history of doing things like this for the lulz on matrix, decides to make a policy list with room type and short code and all just to have it as their offtopic lounge style room.

Like i have a long history of seeing people intentionally toy with edge cases like this on matrix and so does several SCT members. 

Enforcement would require a room version and that is unneeded complexity. Policy lists should always already be created via specialised tools that will prevent these types of situations via strategic configuration of the powerlevels for non creators. This is because normal clients cant set room types for non space rooms. Only specialised clients or power user / developer clients will have the ability to fall into this trap. And when your using a client that intentionally lets you use the full power of matrix and has no guardrails against footguns you are on your own.

Does [MSC1772](https://github.com/matrix-org/matrix-spec-proposals/pull/1772) explicitly define that conversation can not exist in a space as far as auth rules are concerned? I refer to the Spaces MSC for one simple reason. It should already have debated to death this concern as this MSC is essentially the same type of MSC just for policy lists instead of introducing the concept of spaces and room types.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,68 @@
...
+## Proposal
+
+Rooms dedicated to containing policy list recommendations SHOULD use a newly
+defined `m.policy` room type. How clients choose to (not) handle the new type is
+left as an implementation detail. 
+
+Consumers of policy lists SHOULD note that recommendations can still appear 
+outside of `m.policy` rooms. Existing rooms probably won't have the new room 
+type, and some communities might mix conversation and recommendations in the 
+same room (therefore not dedicating the room to recommendations).
```
clokep: This is very confusing and I don't understand what it is recommending. I understand at a protocol level you can send whatever you want into  room, but what's the expected/recommended behavior? Is this like a space that it isn't expected to have conversation or is it expected?

@mscbot concern Confusing behavior around whether or not conversation occurs in policy rooms.

FSG-Cat: Policies live in a bit of a grey area there its essentially communicating. 99% of policy lists work just as spaces usually do. Pure Policies no conversations just like how Spaces usually are exclusively spaces nothing else.

Its just someone might decide to mix policies into conversation for some reason. Be that the original idea that Travis talked about in https://github.com/matrix-org/matrix-spec-proposals/pull/3784/changes#r858892009 or some other reason.

We are also trying to address what was refered to as the legacy problem originally. Aka what to do with policy lists like CME that predate the MSC or Matrix.org COC. Both these lists predate the MSC and predate the mjolnir support for the MSC that was later inherited by Draupnir and implemented by Meowlnir.

Some of this comment may potentially have to make it into the file in a more refined form ofc. Just not sure what if anything from this comment needs to land inside of the MSC it self.

Gnuxie: Conversation is not mixed in policy rooms. The highlighted text is attempting to clarify that moderation policies can exist in rooms without the `m.policy` type.

Gnuxie:
```suggestion
Where the `m.policy` room type is used, conversation is not expected. Clients SHOULD note that moderation policies can appear in rooms without the `m.policy` type, where conversation could be present.
```

FSG-Cat: This change would not reflect the second more important reason why the type is lacking. Lists that predate the type. Other than that i think its a good suggestion.

clokep: The current text is confusing. I would make it clear that `m.policy` rooms are not expected to have conversation.

Maybe use a paragraph break to talk about legacy policy rooms and that nothing in this MSC applies to them?

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: Spaces left whether a conversation can happen in the room as an implementation detail because it was theorized as possibly useful.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: I feel the suggestion implies the lack of type sufficiently. This will need line wrapping, but is otherwise great - thank you! 

@clokep @Johennes - would this resolve your concerns around the MSC not spelling it out?

clokep: Yeah, I think that is clearer.

Johennes: Yes, thanks.

turt2live: Would be resolved by https://github.com/matrix-org/matrix-spec-proposals/pull/3784#discussion_r3050235722

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: I believe this would be resolved by https://github.com/matrix-org/matrix-spec-proposals/pull/3784#discussion_r3050235722

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: I believe this may be mostly/all resolved by https://github.com/matrix-org/matrix-spec-proposals/pull/3784#discussion_r3050235722

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: @FSG-Cat this looks like it'd unblock the MSC.

FSG-Cat: Suggestion with line wrapping has been applied.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -22,10 +22,9 @@ Rooms dedicated to containing policy list recommendations SHOULD use a newly
 defined `m.policy` room type. How clients choose to (not) handle the new type is
 left as an implementation detail.
 
-Consumers of policy lists SHOULD note that recommendations can still appear
-outside of `m.policy` rooms. Existing rooms probably won't have the new room
-type, and some communities might mix conversation and recommendations in the
-same room (therefore not dedicating the room to recommendations).
+Where the `m.policy` room type is used, conversation is not expected.
```
clokep, line 25: Small quibble that I wouldn't block this MSC on, but should we use `m.moderation_policy` to be a bit more descriptive? I could think of multiple types of policies.

TheArcaneBrony: Shouldn't the policy server MSC then also be updated to use `m.moderation_policy` as an event type rather than `m.policy` or whatever the final variant was?

FSG-Cat: Im not married to any particular final name for the room type. I was just copying the precedent set by spaces at the time of writing originally.

So i will take this to the saftey guild and ask them for input on what the name should be.

clokep: Maybe we're already using this term, and that's ok. Thanks for providing the context on why this was chosen. It just stuck out to me as really vague.

Would be curious for other SCT input here.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
turt2live: I'm assuming the concern was also supposed to be resolved.

@mscbot resolve Confusing behavior around whether or not conversation occurs in policy rooms.

clokep: Yes, thank you!

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Unstable prefix
+
+If you want to implement this MSC before its merged you're free to use the unstable type of
+`support.feline.policy.lists.msc.v1`.
+
+After this MSC gets merged if a stakeholder has elected to remove its support for the unstable prefix if any
+Because the room type is immutable, rooms which use the unstable room type might
+find themselves "unsupported" when/if implementations drop support for the unstable
+type when this proposal becomes stable.
```
richvdh: I am utterly unable to parse this sentence.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
richvdh: I feel like @turt2live would have more context on whether this is a sensible name or not. I agree that `m.policy` feels a bit vague.

turt2live: As the person who advocated for `m.room.policy` to enter the spec as-is, I'm not sure my opinion is super reliable 😅. I'm personally fine with `m.policy` as a room type because it's "a room containing a policy". The 'policy' happens to be expressed with primarily ban rules, but in future could be more than that (anonymous reputation data, shades of gray policy definitions to determine on-topic-ness, etc). 

I'm also generally of the opinion that calling it a "moderation policy list" in the spec is wrong, but I'm not concerned enough to actually raise a bug fix PR or issue for it. The use case is to communicate general policy data, not specific to moderation. This could be a security policy, entrance policy, community volunteer policy, etc.

FSG-Cat: A practical use case that exists today where policies are not "moderation policies" would actually be the Access controls for Mjolnir and Draupnir Appservice mode. While they are broken in Draupnir currently they work in Mjolnir as far as i know because it used to work in Draupnir and we did not touch that part of the codebase to my knowledge in that way. (I should note we use the moderation policy recommendations in this context but thats also more of a easiest path forward type of situation i would argue. But thats also only for the disallow policy the allow policy is a custom one in the org.matrix namespace.)

In this case the policies simply control if your bot is allowed to start or not and if your doing Appservice hosting as a service you can easily hook this into a subscription management system. Be that paid or free subscriptions.

Point im trying to make is that we have a concrete already in the wild use case where policies are not strict moderation but something a bit different and we have lots of exploration wanting to extend this system further.

Personally i would say that room sub typing also comes into play here as if we had sub typing then definitively could argue if moderation should be its own subtype and the flagship subtype for the first MSC to try to define a policy subtype due to its clear and proven track record.

Now the fact that the appservice admin room is where Mjolnir and Draupnir stores the policy list is a design choice that as far as i understand exists out of convenience not any actual necessity. Appservice mode was a small beta project that had beta issues that make complete sense in context. The admin room having the policies is one of them and is a choice that i personally would argue for moving away from if one wants to modernize the mode (and is something i may personally look at in the future inside of the Draupnir project) as the room type benefits are quite large for specialized tools like RMU and any other matrix client that wants to implement policy manipulation.

proposals/3784-using-room-type-of-m.policy-for-policy-rooms.md
FSG-Cat: This sentence is meant to read in a simplified form as. If a client supports the unstable room type and later removes it then rooms that used to be "supported" as in recognized are no longer recognized.

The practical example is that if you try to look up policy lists using a client that does not support the unstable version you wont find a single policy list in all of matrix via room type today as no policy lists with the stable room type exist.

richvdh: Can you please fix the MSC so that it can be understood?

Note that my comment applies to all four lines "After this ... becomes stable". It seems to be one huge sentence.

FSG-Cat: A updated version of the text in question has been pushed that hopefully resolves this concern.

richvdh: thanks. much clearer now.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2411

turt2live:
Merged 🎉

PR #3440, opened 2021-10-13
MSC3440: Threading via `m.thread` relation

gsouquet:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/gsouquet/threading-via-relations/proposals/3440-threading-via-relations.md)

[FCP proposal](https://github.com/matrix-org/matrix-doc/pull/3440#issuecomment-1034692453)

Implementation:
* Element Web: Threads P0: Web vector-im/element-meta#3
* Element Android: Threads P0: Android vector-im/element-meta#7
* Element iOS: Threads P0: iOS  vector-im/element-meta#9

Preview: https://pr3440--matrix-org-previews.netlify.app

PR #3785, opened 2022-04-27
Clarify serialization properties of the latest thread event from MSC3440.

clokep:
This clarifies some bits of the `latest_thread` property of the bundled aggregations form of the `m.thread` relation [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) from a few things that came up during implementation review:

* Correct the format of the `latest_event` (matrix-org/synapse#12085); this was somewhat clarified during the MSC review in [`89c4b5e` (#3440)](https://github.com/matrix-org/matrix-spec-proposals/pull/3440/commits/89c4b5e4b65eef20e8d8a9e971bceba17a5e60fa), but I've tried to make it clearer.
* Clarify that the `latest_event` includes bundled aggregations (which was only implied from the above) (matrix-org/synapse#12270).

### DIFF
--- a/proposals/3440-threading-via-relations.md
+++ b/proposals/3440-threading-via-relations.md
@@ -53,6 +53,9 @@ would include additional information in the `unsigned` field:
 
 * `latest_event`: The most recent event which relates to this event, with 
   `rel_type` of `m.thread`.
+
+  The latest event should be serialised in the same form as the event itself;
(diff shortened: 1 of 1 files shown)


t3chguy (approved):

turt2live (approved):
thanks!

PR #2174, opened 2019-07-14
MSC2174: Move the `redacts` key to a sane place

richvdh:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/rav/proposals/move_redacts_key/proposals/2174-move-redacts-key.md).

This would fix #502.

PR #3788, opened 2022-05-04
Add an unstable room version to MSC2174

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/2174

### DIFF
--- a/proposals/2174-move-redacts-key.md
+++ b/proposals/2174-move-redacts-key.md
@@ -15,3 +15,8 @@ when serving such events over the Client-Server API.
 For improved compatibility with *newer* clients, servers should add a `redacts`
 property to the `content` of `m.room.redaction` events in *older* room versions
 when serving such events over the Client-Server API.
+
+## Unstable prefix
+
+Implementations should use `org.matrix.msc2174` for the room version containing
+this change until assigned to a specified room version.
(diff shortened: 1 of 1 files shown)


proposals/2174-move-redacts-key.md
```diff
@@ -15,3 +15,9 @@ when serving such events over the Client-Server API.
 For improved compatibility with *newer* clients, servers should add a `redacts`
 property to the `content` of `m.room.redaction` events in *older* room versions
 when serving such events over the Client-Server API.
+
+## Unstable prefix
+
+Implementations should use `org.matrix.msc2174` for the room version containing
```
uhoreg, line 21: It probably doesn't matter much, but does anything actually use this room version?  The synapse impl. seems to bundle msc2174 and msc2176 together into `org.matrix.msc2176`

turt2live: That implementation is actually incomplete :D 

At a glance, it looks like the implementation was missing the bit where the key moved into `content`, but I guess if that's been done then there's no real need for a specific room version.

turt2live: It ended up needing a room version: https://github.com/matrix-org/synapse/pull/12634

PR #3790, opened 2022-05-09
MSC3790: Register Clients

xarvic:
[Rendered](https://github.com/xarvic/matrix-spec-proposals/blob/main/proposals/3790-register-clients.md)

Signed-off-by: Christoph König <jan.christoph.koenig@posteo.net>


proposals/3788-register-clients.md
```diff
@@ -0,0 +1,121 @@
+# MSC3788: Register Clients
+It should be possible to delegate specific tasks to other clients which are able to handle them. To make this possible the spec should define a special private room for the user, which is used by clients to register their abilities, to which device they belong and by which method they can be startet.
```
turt2live, line 2: Please wrap your lines to roughly 120 characters.

turt2live (commented):
Please also sign off on your changes when you get a chance.

xarvic:
Wow that was quick, thanks for the reply :)

> Please also sign off on your changes when you get a chance.

Just by editing the comment?

proposals/3788-register-clients.md
```diff
@@ -0,0 +1,121 @@
+# MSC3788: Register Clients
```
turt2live, line 1: Your MSC number is actually 3790 ftr - please update the file name and such :)

turt2live:
Sign off can also be a comment in the PR description using the template:

```
Signed-off-by: Your Name <email@example.com>
```

proposals/3788-register-clients.md
xarvic: ok

proposals/3790-register-clients.md
```diff
@@ -0,0 +1,129 @@
...
+        },
+    ]
+    
+  }
+}
+```
+
+`client_name` is the app name.
+`client_device_id` is the id the device this client is located on.
+`launch_command` is the command to run when the user wants to inspect one of the associated_event_types.
```
uhoreg, line 82: This will only work if the command exists on the computer that your current client is running on.  Similarly, if you try to use `launch_url` on a different computer/browser, then it will create a new session instead of using the existing session, which is presumably what you would want it to do.

proposals/3790-register-clients.md
```diff
@@ -0,0 +1,129 @@
...
+
+When running the client:
+- When launch_command is excuted the enviorment varaible MATRIX_CLIENT_STARTUP_EVENT is set to `<room_id>/<message_id>`.
+- When launch_url is used the url is extended by `?matrix_client_startup_event=<room_id>/<message_id>`.
+- When an action is started the enviorment varaible MATRIX_CLIENT_ACTION_LOCATION is set to `<room_id>`.
+- When an action_url is used the url is extended by `?matrix_client_action_location=<room_id>`.
+
+Using enviorment varaibles has the advantage that this functionality can be implemented as a plugin for an existing app.
+
+To highlight certain events a client can use "m.event_description" event.
```
uhoreg, line 98: It's not clear why a client would want to "highlight" an event, what it's supposed to do, and how it's related to the rest of this proposal.

xarvic:
After reading the MSC again i noticed one problem, that it is not that easy to detect whether two clients are one the same device, since the `device_id` is actually a `client_id`. Maybe querying the device name could help?

proposals/3790-register-clients.md
```diff
@@ -0,0 +1,129 @@
...
+
+If a matrix chat-client finds an event which is not related to messaging, it searches the registry for clients
+that can process the event and offers to launch them.
+
+For an example a remote-control app could use matrix to establish a p2p connection. Therefore the app would not
+need its own backend and account system. Another example is collaborative editing directly in matrix.
+
+## Proposal
+
+The spec should reserve the special room "client_registrations".
```
uhoreg, line 14: That's not a room ID or alias.  How does the client find that room?

proposals/3790-register-clients.md
xarvic: You are right. Is there a way to detect if two clients are on the same device? If so the chat client should probably filter out the rest.

proposals/3790-register-clients.md
xarvic: The idea was that by default events just get a small infotext combined with a link to click on like this:
image

Where as events which have a highlight get a bigger area with icon, header, description and a button. Something like this:
image

The aim is to distinguish between important and unimportant events so that the user doesn't get spammed.

Do thinks this could be useful?

proposals/3790-register-clients.md
xarvic: Having a fixed room id is probably best, should i change it to `#client_registration:yourhomeserver.com`?

proposals/3790-register-clients.md
```diff
@@ -0,0 +1,129 @@
+# MSC3790: Register Clients
```
Johennes, line 1: The "Rendered" link in the pull request description appears to be a 404.

xarvic: That is strange, for me it works.

Johennes: Hm, looks like it works for me, too, now. 🤷‍♂️

proposals/3790-register-clients.md
KitsuneRal: Strictly speaking, no. Browser sandboxes are actually supposed to prevent web applications from doing that, and application confinement mechanisms  like Flatpak normally deny most of insight into the local environment to the application as well.

proposals/3790-register-clients.md
uhoreg: That's not an ID, that's an alias.  Also, that would mean that any homeservers that already have a room with such an alias would get confused clients.

proposals/3790-register-clients.md
uhoreg: I don't see why something like that needs to be sent as a separate event.

KitsuneRal (changes requested):
So far the MSC is very sketchy and has quite a few gaps. The note in "Potential issues" is not completely accurate (the room is private to the user after all) but still provides ample opportunity for lateral propagation of an attack once one device has been compromised, making the suggestion a non-starter without proper security mitigations.

What you're trying to devise is a mechahism for experience hand-over really. Whether the client is running on the same device or not doesn't really matter - I might want to open some event on a desktop while running through the feed on a mobile phone, e.g. You really shouldn't deal with application invocation - it's a big security can of worms (have you heard of Windows OLE? You're half-way reinventing it for Matrix) and there are plenty of things in operating systems - think XDG specifications on Linux, e.g. - that are aimed at doing exactly that, with proper safeguards and local context.

That said, I think that advertising support of certain event types across devices of a user has some promise. I'd be interested to see an MSC focusing specifically on this - e.g., with a client at a certain device id claiming support of certain event types and other clients of the user unhiding these events even if they are unknown locally and hinting at the user's device(s) that can show them. This is a more compact and less controversial function.

xarvic:
> So far the MSC is very sketchy and has quite a few gaps. The note in "Potential issues" is not completely accurate (the room is private to the user after all) but still provides ample opportunity for lateral propagation of an attack once one device has been compromised, making the suggestion a non-starter without proper security mitigations.
> 
> What you're trying to devise is a mechahism for experience hand-over really. Whether the client is running on the same device or not doesn't really matter - I might want to open some event on a desktop while running through the feed on a mobile phone, e.g. You really shouldn't deal with application invocation - it's a big security can of worms (have you heard of Windows OLE? You're half-way reinventing it for Matrix) and there are plenty of things in operating systems - think XDG specifications on Linux, e.g. - that are aimed at doing exactly that, with proper safeguards and local context.

Thanks, for your input. I took a look at OLE and XDG. You are right. This is what I wanted to have. But yes, i see the problem.

> That said, I think that advertising support of certain event types across devices of a user has some promise. I'd be interested to see an MSC focusing specifically on this - e.g., with a client at a certain device id claiming support of certain event types and other clients of the user unhiding these events even if they are unknown locally and hinting at the user's device(s) that can show them. This is a more compact and less controversial function.

Maybe i will try that.

PR #3414, opened 2021-09-27
MSC3414: Encrypted state events

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/encrypted-state/proposals/3414-encrypted-state.md)

See also:
* https://github.com/matrix-org/matrix-doc/issues/794

PR #3813, opened 2022-05-10
Obfuscated events

tusooa:


proposals/3813-obfuscated-events.md
```diff
@@ -0,0 +1,186 @@
...
+a template. There's no major expected impact in this proposal, so it doesn't list one. If your proposal
+was more invasive (such as proposing a change to how servers discover each other) then that would be
+a good thing to list here.*
+
+*If you're having troubles coming up with a description, a good question to ask is "how
+does this proposal improve Matrix?" - the answer could reveal a small impact, and that is okay.*
+
+There can never be enough templates in the world, and MSCs shouldn't be any different. The level
+of detail expected of proposals can be unclear - this is what this example proposal (which doubles
+as a template itself) aims to resolve.-->
```
turt2live, line 19: don't forget to remove the template comments from the proposal :)

turt2live (commented):
Please also [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes so we can incorporate the MSC later on in the process.

tusooa:
> Please also [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes so we can incorporate the MSC later on in the process.

i'll squash the commits later

uranuspucksaxophone:
I wonder if this PR could be implemented along with MSC3414 ( #3414 ), both would mean a very high metadata resistance, which would benefit the privacy of users. The only thing I don't like about matrix are the metadata leaks along with the centralized identity system. Anyways, It would be interesting to see how these two proposals scale to Matrix P2P.

PR #3814, opened 2022-05-12
[WIP] MSC3814: Dehydrated devices with SSSS

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/shrivelled_sessions/proposals/3814-dehydrated-devices-with-ssss.md)


proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+Once a client calls `POST /dehydrated_device/{device_id}/events`, the server
+can delete the device (though not necessarily its to-device messages).  Once a
+client calls `POST /dehydrated_device/{device_id}/events` with a `next_batch`
+token, the server can delete any to-device messages delivered in previous
+batches.  It is recommended that, for the last batch of messages, the server
+still send a `next_batch` token, and return an empty `events` array when called
+with that token, so that it knows that the client has successfully received all
+the messages.
```
dkasak, line 129: Can we please not leave this as just a recommendation? Either we should commit to it fully and require it or delete it.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+If the client is able to decrypt the data and wants to use the dehydrated
+device, the client retrieves the to-device messages sent to the dehydrated
+device by calling `POST /dehydrated_device/{device_id}/events`, where
+`{device_id}` is the ID of the dehydrated device.  Since there may be many
+messages, the response can be sent in batches: the response can include a
+`next_batch` parameter, which can be used in a subsequent call to `POST
+/dehydrated_device/{device_id}/events` to obtain the next batch.
+
+```
+POST /dehydrated_device/{device_id}/events
```
nico-famedly, line 104: Why is this a POST and not a GET like /sync and /messages?

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+  "events": [
+    // array of to-device messages, in the same format as in
+    // https://spec.matrix.org/unstable/client-server-api/#extensions-to-sync
+  ],
+  "next_batch": "token to obtain next events" // optional
+}
+```
+
+Once a client calls `POST /dehydrated_device/{device_id}/events`, the server
+can delete the device (though not necessarily its to-device messages).  Once a
```
nico-famedly, line 123: Why should the server delete the device? Shouldn't this rather be done by the client explicitly in a delete call?

Imo it is not quite obvious that fetching the events should "break" the device. A client might fail to properly restore and now you lost all the intermediate sessions. instead the client should replace the device once it is somewhat sure it restored successfully and has uploaded the megolm keys to online backup.

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: IIRC, the rationale was because the call has side-effects (deleting the device).

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: The idea is that when the client starts getting events, it means that the client is signalling its intention to use the dehydrated device, and it has been "claimed", so it shouldn't be used by anyone else.  At this point, there isn't much that can be done if the client, e.g. fails to decrypt some messages.  If it fails to decrypt messages with the dehydrated device, it's unlikely that leaving the device around will fix anything in the future -- any future attempts would likely fail as well.  So the best thing to do is to replace the dehydrated device with a new one anyways.

I'm not insistent on this endpoint deleting the dehydrated device, but I think that once you start using a dehydrated device, you'll want to create a new device no matter what.

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: It's a bit weird that it doesn't follow the pattern of /messages, /events or /sync imo. I'll try implementing it as a GET without the device deletion first and see how that works out, I think.

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: It's more that if the device fetches the first few events but then the user closes the browser and it never gets to upload the devices, then you have effectively thrown the dried fish out the window without properly getting to use it. So in that case it should either only delete the device, when it deletes the first few messages (i.e. by the client paginating with a next token), or just wait for the user to send a new device. Since you CAN still use the same dried device from another device, I think. All of the messages will be PRE KEY messages, so you can decrypt them as long as you haven't deleted the one time keys from the pickled device. So even if a client downloads the first batch of messages and then starts with the next batch and the first batch gets deleted, a different client should still be able to pick up from there.

I agree that you want to create a new device no matter what, but that can just be done by uploading a new one instead of implicitly doing it when receiving messages.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+If no dehydrated device is available, the server responds with an error code of
+`M_NOT_FOUND`, http code 404.
+
+If the client is able to decrypt the data and wants to use the dehydrated
+device, the client retrieves the to-device messages sent to the dehydrated
+device by calling `POST /dehydrated_device/{device_id}/events`, where
+`{device_id}` is the ID of the dehydrated device.  Since there may be many
```
nico-famedly, line 98: Can this endpoint be called on devices that are not dehydrated devices?

nico-famedly: And if yes, should it error out on unknown devices or not?

uhoreg: If `device_id` is not the ID of a dehydrated device, it should return some sort of error.  I think maybe a 404 if the device ID is unknown, and some sort of 400 if the device ID is known, but doesn't belong to a dehydrated device?

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: Yeah, the issue of a client starting to load events and then dying somehow is possible, but seems like it would be extremely rare.

I think another consideration is that a client could forget to replace the dehydrated device.  If the device gets deleted automatically, then it makes it obvious that the client didn't do that.

In any event, I think it's fine to try it out with `GET` and without automatically deleting the device, and see how it goes.

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: I currently have it implemented that it returns 403, if you try to sync with a device that isn't the one from your access token and not a dehydrated device. I actually wanted an endpoint that just syncs to_device messages for a while, but I can remove the support for syncing your own to_device messages.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+After the dehydrated device is uploaded, the client will upload the encryption
+keys using `POST /keys/upload/{device_id}`, where the `device_id` parameter is
+the device ID given in the response to `PUT /dehydrated_device`.  The request
+and response formats for `POST /keys/upload/{device_id}` are the same as those
+for `POST /keys/upload` with the exception of the addition of the `device_id`
+path parameter.
+
```
nico-famedly, line 66: There is a bit of a race condition here, when creating the device and uploading the account stuff to SSSS. Not sure how to best mitigate that.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+## Unstable prefix
+
+While this MSC is in development, the `/dehydrated_device` endpoints will be
+reached at `/unstable/org.matrix.msc3814.v1/dehydrated_device`, and the
+`/dehydrated_device/{device_id}/events` endpoint will be reached at
+`/unstable/org.matrix.msc3814.v1/dehydrated_device/{device_id}/events`.  The
+dehydration algorithm `m.dehydration.v1.olm` will be called
+`org.matrix.msc3814.v1.olm`.  The SSSS name for the dehydration key will be
+`org.matrix.msc3814` instead of `m.dehydrated_device`.
+
```
nico-famedly, line 221: Client implementation: https://gitlab.com/famedly/company/frontend/famedlysdk/-/merge_requests/1111

Server implementation: https://github.com/matrix-org/synapse/pull/13581

Both not merged yet and notably missing is the dehydrated device format.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,224 @@
...
+the private one-time keys.  This will come at the expense of implementations
+such as libolm needing to implement extra functions to support dehydration, but
+will have the advantage that we don't need to figure out a format that will fit
+into every possible implementation's idiosyncrasies.  The format will be
+encrypted, which leads to ...
+
+#### Encryption key
+
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`.
```
nico-famedly, line 146: Hm, I just stored the device encrypted in SSSS instead of a key to encrypt the device with. Is that what this intended or do we need some account data to store the actual device in while the encryption key for the device is stored via SSSS?

uhoreg: I guess it isn't clear enough, but the way it's supposed to work in this MSC is that SSSS stores the key to de/rehydrate the device, and the actual (encrypted) device data gets transferred via the `PUT/GET /dehydrated_device` endpoint.  The reason that SSSS stores the key rather than the actual dehydrated device is that some clients may not want to store the main SSSS key, but only store the secrets that it is actually using.  If the dehydrated device data was stored in SSSS, then clients would have to remember the main SSSS key to periodically upload a new dehydrated device.

nico-famedly: Ah, then I implemented this wrong. So this would be stored as the device_data, I guess? The whole device_data is the dehydrated device or is it some subkey?

proposals/3814-dehydrated-devices-with-ssss.md
reivilibre: There's also an issue if there's a connection problem during the `POST` request; the client presumably won't be able to re-try the request because the device will have been deleted.
We generally try to design our APIs so that they work / can be retried even if the request fails half-way through for some reason.

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: Yes, the whole `device_data` is the (encrypted) dehydrated device.

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: Actually, the idea is that the dehydrated device is "deleted" in the sense that no other client can claim it, and if a client queries for the dehydrated device, it won't be returned. But the events associated with it are still there and can be retrieved (until the events get deleted as described elsewhere in the MSC), so if the client re-tries the `POST` request, it will still get the events.

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: Actually, now that I implemented it correctly, there is no race condition :)

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: I implemented it that way now. It is a bit more difficult, but it gets rid of the race that otherwise exists and as you say, it is needed to properly rotate the device in the background (which I haven't implemented yet, since I need to decide where to store the device age :3)

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: Yes, the problem is just that if the client fails to replace the device, then a failed login attempt will break the device dehydration until the next successfull login, since there is no way to receive messages in the meantime (while that would work fine if the device is just kept).

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: Added a note about error code.  We can bikeshed about the specific error code if needed, but I'm going to resolve this conversation.

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: I've made this a requirement instead of a recommendation.

proposals/3814-dehydrated-devices-with-ssss.md
nico-famedly: I actually ran into another race condition here in production. We currently have it implemented that the PUT of a new device removes the old device. However, since uploading a new device takes several requests (claim new device, upload keys, sign it, upload encrypted device), we run into a race condition, where the user closes the browser window during one of the steps and maybe only signs back in later. That means we have an unhydrateable device and we again lose messages over the gap. Ideally there would be some way to make this atomic to prevent this race condition.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,228 @@
...
+This may be addressed by using fallback keys as described in
+[MSC2732](https://github.com/matrix-org/matrix-doc/pull/2732).
+
+To reduce the chances of one-time key exhaustion, if the user has an active
+client, it can periodically replace the dehydrated device with a new dehydrated
+device with new one-time keys.  If a client does this, then it runs the risk of
+losing any megolm keys that were sent to the dehydrated device, but the client
+would likely have received those megolm keys itself.
+
+Alternatively, the client could perform a `/sync` for the dehydrated device,
```
BillCarsonFr, line 176: Does this sill works with v2? can we still sync on the dehydrated device?

nico-famedly: You can't sync as a different device in this proposal. You can fetch the events for that device, but this proposal implicitly deletes the device in that case, which means you can't keep the device alive after that. So imo your only option is to replace it (which is somewhat easy to do, but you might need to authenticate the new signature upload/device?).

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,228 @@
...
+  }
+}
+```
+
+If no dehydrated device is available, the server responds with an error code of
+`M_NOT_FOUND`, HTTP code 404.
+
+If the client is able to decrypt the data and wants to use the dehydrated
+device, the client retrieves the to-device messages sent to the dehydrated
+device by calling `POST /dehydrated_device/{device_id}/events`, where
```
clokep, line 97: Why include a `device_id` if you can only have a single dehydrated device? It has implications that you can provide more than one (and causes additional error checking of whether the provided device ID matches the dehydrated device ID).

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,228 @@
...
+The dehydrated device may run out of one-time keys, since it is not backed by
+an active client that can replenish them.  Once a device has run out of
+one-time keys, no new olm sessions can be established with it, which means that
+devices that have not already shared megolm keys with the dehydrated device
+will not be able to share megolm keys.  This issue is not unique to dehydrated
+devices; this also occurs when devices are offline for an extended period of
+time.
+
+This may be addressed by using fallback keys as described in
+[MSC2732](https://github.com/matrix-org/matrix-doc/pull/2732).
```
clokep, line 168: Should we update this with a link to the spec?

proposals/3814-dehydrated-devices-with-ssss.md
poljar: I agree that the device ID is redundant. Though since this MSC has been written a new use-case for this endpoint has been found.

In the sliding sync world we have split out the fetching of to-device events into a separate sync loop. Namely one of the biggest problems of the existing `/sync` mechanism is that you get too much data all at once and the downloading and processing of that data prevents the UI from being updated.

To-device events are one of those things that are not directly related to the things that a client will want to display in a room or room list, so putting it into a separate sync loop allows the main loop to quickly send updates while to-device moves along in the background. More info here: https://github.com/matrix-org/matrix-rust-sdk/issues/1928

I think that old sync could handle such a split as well, so I would suggest here to rename the endpoint to become `/sync/to_device/{device_id}` where `device_id` might be optional and used only in the case of a dehydrated device.

proposals/3814-dehydrated-devices-with-ssss.md
poljar: I agree that having this be a `PUT` is the wrong pattern here. If the whole idea here is for a Client to fetch to-device events and room keys then we should design the flows in a way that minimizes the risk of loss of such room keys.

As it stands the flows can't be resumed once you start fetching the to-device events, and the fetching of the to-device events will be by far the longest operation here.

What would allow perfect resumption is:

1. The dehydrated device only gets deleted if the client requests so.
2. The to-device events only get deleted¹ when the dehydrated device gets deleted, this is opposed to the current mechanism, where to-device events get deleted once the server sees a `next_batch` from a previous request. 

No 2. would ensure that, even if a device that attempts a rehydration gets stopped and deleted mid-rehydration, another, new device can restart the rehydration process.

> I actually ran into another race condition here in production. We currently have it implemented that the PUT of a new device removes the old device. However, since uploading a new device takes several requests (claim new device, upload keys, sign it, upload encrypted device),

Agree here as well, `PUT` of a new device should happen in a single request which should upload the dehydrated device, its device keys and any one-time keys. I implemented a draft version of this behavior in this patch: https://github.com/matrix-org/synapse/commit/777b3056531ff37c91014abbd029ac079d8d7c25

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: The reason for the `device_id` parameter is that, while one client is fetching events, another client could create a new dehydrated device.  Without the `device_id` parameter, the server could think that the client wants to fetch the events for the new device which, if there are any, it won't be able to decrypt since it's for a device that it doesn't know about.  With the `device_id` parameter, the server will at least be able to say that there are no more events (since the device has been replaced by a new one).

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+Once a client calls `POST /dehydrated_device/{device_id}/events`, the server
+can delete the device (though not necessarily its to-device messages).  Once a
+client calls `POST /dehydrated_device/{device_id}/events` with a `next_batch`
+token, the server will delete any to-device messages delivered in previous
+batches.  For the last batch of messages, the server will still send a
+`next_batch` token, and return an empty `events` array when called with that
+token, so that it knows that the client has successfully received all the
+messages and can clean up all the to-device messages for that device.
```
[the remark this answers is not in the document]
uhoreg: I think that in your example, the "last batch" of messages is 2. where the client requests with `next_batch=A`.  So the server still returns a `next_batch` of `B` even though it knows that the next call will be empty.  Then at 3. when the client requests with `next_batch=B`, the server discards all messages, and returns an empty array of messages and no `next_batch`.  I agree that the text could be clearer.

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: Yes.  For reference, until I manage to update this, it's https://spec.matrix.org/v1.7/client-server-api/#one-time-and-fallback-keys

proposals/3814-dehydrated-devices-with-ssss.md
clokep: This sounds quite racy to me -- how does the server know that one dehydrated device is claimed? How would the client know to make a new one instead of claim the old one?

uhoreg: > how does the server know that one dehydrated device is claimed?

It's OK for multiple clients to rehydrate the same device (unlike in the previous proposal), because it never becomes a real device.  So the server can just wait until some client fetches all the events before dropping the device.

> How would the client know to make a new one instead of claim the old one?

Making a new device and rehydrating an old one are two different use cases.  Rehydration happens after you log in, and you're setting up encryption and trying to get keys.  It only happens once in the device's lifetime.  Creating a new dehydrated device would happen after you've already set up your encryption and already attempted to rehydrate a device.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+   │                        Pickle                             │
+   ├───────────────────────────────────────────────────────────┤
+   │Name                    │ Type          │ Size (bytes)     │
+   ├────────────────────────┼───────────────┼──────────────────┤
+   │Version                 │ u32           │ 4                │
+   │Ed25519 key pair        │ KeyPair       │ 64               │
+   │Curve25519 key pair     │ KeyPair       │ 64               │
+   │Number of one-time keys │ u32           │ 4                │
+   │One-time keys           │ [OneTimeKey]  │ N * 69           │
+   │Fallback keys           │ FallbackKeys  │ 2 * 69           │
```
uhoreg, line 206: I think we only need one fallback key.  Normal clients store two fallback keys because they upload a new fallback key after the current fallback key is used.  But with dehydrated devices, we will never upload a new fallback.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+   │Next key ID             │ u32           │ 4                │
```
uhoreg, line 207: Likewise, I don't think we need this because we won't upload new OTKs after we upload the dehydrated device.  It also makes assumptions about the structure of the OTK IDs, which may not be true for all olm implementations.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+   │Private key             │ [u8; 32]      │ 32               │
+   └────────────────────────┴───────────────┴──────────────────┘
+
+   ┌───────────────────────────────────────────────────────────┐
+   │                      OneTimeKey                           │
+   ├────────────────────────┬───────────────┬──────────────────┤
+   │Name                    │ Type          │ Size (bytes)     │
+   ├────────────────────────┼───────────────┼──────────────────┤
+   │Key ID                  │ u32           │ 4                │
+   │Is published            │ u8            │ 1                │
```
uhoreg, line 224: I don't think we need these, because the key ID doesn't get used when decrypting events, and any OTKs that are included in the dehydrated device will surely have been published -- any unpublished OTKs should just be omitted.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+   │Curve 25519 key pair    │ KeyPair       │ 69               │
```
uhoreg, line 225:
```suggestion
   │Curve 25519 key pair    │ KeyPair       │ 64               │
```

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+    "algorithm": "m.dehydration.v1.olm",
+    "other_fields": "other_values"
+  }
+}
+```
+
+If no dehydrated device is available, the server responds with an error code of
+`M_NOT_FOUND`, HTTP code 404.
+
+If the client is able to decrypt the data and wants to use the dehydrated
```
uhoreg, line 123: I wonder if we can say something like: the server is allowed to discard any non-`m.room.encrypted` to-device message that it receives for the dehydrated device.  There's no point in keeping key requests sent to the dehydrated device because it won't send anything back.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+the dehydrated device, the public device keys, and one-time keys are all
+uploaded in the same request. This change should prevent the creation of
+dehydrated devices which do not support end-to-end encryption.
+
+To upload a new dehydrated device, a client will use `PUT /dehydrated_device`.
+Each user has at most one dehydrated device; uploading a new dehydrated device
+will remove any previously-set dehydrated device.
+
+The client *must* use the public [Curve25519] [identity key] of the device,
+encoded as unpadded Base64, as the device ID.
```
uhoreg, line 43: Since the device ID ends up in a path parameter, we should probably make this URL-safe Base64 to avoid issues with clients failing to URL-encode the ID.

```suggestion
encoded as unpadded URL-safe Base64, as the device ID.
```

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+    "device_id": "<device_id>",
+    "valid_until_ts": <millisecond_timestamp>,
+    "algorithms": [
+        "m.olm.curve25519-aes-sha2",
+    ]
+    "keys": {
+        "<algorithm>:<device_id>": "<key_base64>",
+    },
+    "signatures:" {
+        "<user_id>" {
```
uhoreg, line 69:
```suggestion
    "signatures": {
        "<user_id>": {
```

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+  "device_id": "dehydrated_device_id",
+  "device_data": {
+    "algorithm": "m.dehydration.v1.olm"
+    "other_fields": "other_values"
+  },
+  "initial_device_display_name": "foo bar", // optional
+  "device_keys": {
+    "user_id": "<user_id>",
+    "device_id": "<device_id>",
+    "valid_until_ts": <millisecond_timestamp>,
```
uhoreg, line 61: This property isn't defined anywhere

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+
+For the last batch of messages, the server will still send a
+`next_batch` token, and return an empty `events` array when called with that
+token, this signals to the client that it has received all to-device events and
+it can delete the dehydrated device and create a new one.
+
+If the given `device_id` is not the dehydrated device ID, the server responds
+with an error code of `M_FORBIDDEN`, HTTP code 403.
+
+### Deleting a dehydrated device
```
uhoreg, line 165: We should probably specify what happens when we use [`POST /delete_devices`](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3delete_devices) and [`DELETE /devices/{deviceId}`](https://spec.matrix.org/unstable/client-server-api/#delete_matrixclientv3devicesdeviceid) on the dehydrated device.  Also [`POST /logout/all`](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3logoutall).  Presumably those would delete the dehydrated device.  So maybe we don't need our own `DELETE` endpoint here?  Though this endpoint allows you to delete the dehydrated device without knowing the device ID, so might still be useful.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,392 @@
...
+
+When decoding, clients *must* ignore the public keys and instead derive the
+public key from the private one.
+
+#### Encryption key
+
+TODO: Explain why the double derivation is necessary.
+
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`.
```
uhoreg, line 250: if I'm reading the iOS implementation correctly, the key is encoded with unpadded base64 (as is done with the other keys in secret storage)

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg:
```suggestion
This may be addressed by using [fallback keys](https://spec.matrix.org/v1.9/client-server-api/#one-time-and-fallback-keys).
```

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,391 @@
...
+
+```jsonc
+{
+  "device_id": "dehydrated_device_id",
+  "device_data": {
+    "algorithm": "m.dehydration.v1.olm"
+    "other_fields": "other_values"
+  },
+  "initial_device_display_name": "foo bar", // optional
+  "device_keys": {
```
uhoreg, line 58: We're planning on adding a new property to the `device_keys` for dehydrated devices: a property called `dehydrated`, which will be set to `true`.  Since this will be covered by the signatures, it can't be spoofed by the server.  This will allow clients to treat the dehydrated device specially.  For example:

- give it a localized name, rather than using the device display name, to reduce confusion
- display it separately from other devices, possibly even not displaying it as a device, but as an encryption feature
- not sending room keys to that device unless it is a user's only device (or possibly if the user only has one other device?), with the assumption that if the user has another device, then that device can help new devices get room keys via backup/key sharing

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: This was previously considered, but was rejected as we would end up with two different representations of the same data (as device ID and as the key), which would be confusing.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+
+Once a client calls `POST /dehydrated_device/{device_id}/events` with a
+`next_batch` token, unlike the `/sync` endpoint, the server should *not* delete
+any to-device messages delivered in previous batches. This should prevent the
+loss of messages in case the device performing the rehydration gets deleted. In
+the case the rehydration process gets aborted, another device will be able to
+restart the process.
+
+For the last batch of messages, the server will still send a
+`next_batch` token, and return an empty `events` array when called with that
```
richvdh, line 174: This is not what https://spec.matrix.org/v1.9/appendices/#pagination says should happen. I'd like to see this changed before this stabilises.

uhoreg: The original reason why this endpoint used an empty `events` array to signal the end was that the client using the final `next_batch` token would signal to the server that it could delete the to-device messages.  Since we aren't doing that any more, we can make it work like the appendix says it should.

proposals/3814-dehydrated-devices-with-ssss.md
richvdh: A GET endpoint with side-effects seems like a big no-no to me. Everyone expects a GET request to have approximately zero side-effects.

richvdh: oh, but we're also proposing removing the side-effects? SGTM in that case

uhoreg: Yup, the current implementation no longer automatically deletes the device on the server side, but relies on the client to delete/create a new device.  So we're going to try to make this a GET.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+devices; this also occurs when devices are offline for an extended period of
+time.
+
+This may be addressed by using [fallback keys](https://spec.matrix.org/v1.9/client-server-api/#one-time-and-fallback-keys).
+
+To reduce the chances of one-time key exhaustion, if the user has an active
+client, it can periodically replace the dehydrated device with a new dehydrated
+device with new one-time keys.  If a client does this, then it runs the risk of
+losing any megolm keys that were sent to the dehydrated device, but the client
+would likely have received those megolm keys itself.
```
richvdh, line 340: Are we doing this [replacing the dehydrated device periodically] or not?

It seems like both have serious downsides. If we *do* replace it, we have a very racy operation that is certain to cause UTDs in practice. If we *don't* replace it, then we'll end up with no remaining OTKs at all, and an incredibly long list of to-device messages all of which have to be downloaded and decrypted by any new clients.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+The plain-text is encrypted with [AES-256] in [CBC] mode with [PKCS#7] padding,
+using the key $`AES\_KEY`$ and the IV $`AES\_IV`$ to give the cipher-text.
+
+Then the cipher-text are passed through [HMAC-SHA-256]. The first 8 bytes of the
+MAC are appended to the cipher-text.
+
+The cipher-text, including the appended MAC tag, are encoded using unpadded
+Base64 to give the device pickle.
```
cvwright, line 296: Why is this key derivation so different from the existing symmetric scheme for encrypting secrets in secret storage?  I'm looking at https://spec.matrix.org/v1.10/client-server-api/#msecret_storagev1aes-hmac-sha2-1

If the existing `m.secret_storage.v1.aes-hmac-sha2` is not good enough for this key, then why are we still using it for the other secrets?

If it is good enough, then why do something randomly different here?

IMO if there's going to be a new scheme for encrypting secrets in secret storage, then that should be a new MSC all on its own.  Otherwise it feels like meaningless duplication of effort, and more chance for implementations to make a stupid mistake.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+The `device_key` is then further expanded into a AES256 key, HMAC key and
+initialization vector.
+
+
+```math
+\begin{aligned}
+    AES\_KEY \parallel HMAC\_KEY \parallel AES\_IV
+    &= \text{HKDF}\left(0,DEVICE\_KEY,\text{``Pickle"},80\right)
```
cvwright, line 285: Why is the IV not random here?

Is there a reason why this encryption needs to be deterministic?  I don't think there is.  But even if so, why not use a well known deterministic encryption mode like SIV?

Otherwise this looks pretty sketchy.  In practice it's probably not going to lead to an exploit, but it should probably get flagged in a security audit.

cvwright: Ok I think I see what's going on.  Based on this comment https://github.com/matrix-org/matrix-rust-sdk/blob/794b11a0cecd2fab8e43931639eba5212fb89921/crates/matrix-sdk-crypto/src/dehydrated_devices.rs#L347 you're using the libolm pickle encryption scheme?

But given the warning in that Rust comment, it sounds like the potential for IV reuse here is a real problem.  If an application fails to generate a new random key when it creates a new dehydrated device, then you're going to reuse the IV.

So I guess this just brings me back to my question below: Why not just use the existing `m.secret_storage.v1.aes-hmac-sha2` construction here?

proposals/3814-dehydrated-devices-with-ssss.md
poljar: As you have noticed, this is a scheme already used in the stack, it's used by [Olm/Megolm](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/docs/olm.md#version-1). The `libolm` pickle format is using it as well.

Not that I'm disagreeing with you on your other points, but the scheme isn't randomly different.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`.
+
+The randomly generated encryption key *must* be expanded using the HMAC-based
+Key Derivation function defined in [RFC5869].
+
+```math
+\begin{aligned}
+    DEVICE\_KEY
+    &= \text{HKDF} \left(\text{``Device ID``}, RANDOM\_KEY, \text{``dehydrated-device-pickle-key"}, 32\right)
```
dkasak, line 274: `RANDOM_KEY` is mentioned in the example but not defined in the text. Should have a sentence to define what we mean by it unambiguously. I left a suggestion to resolve the concern above.

proposals/3814-dehydrated-devices-with-ssss.md
dkasak, line 285: Incidentally, I don't think we define the HKDF parameter order anywhere. Not in this MSC and not in the spec. The only place that defines it is the [Olm specification](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/docs/olm.md?ref_type=heads).

We should define it in the MSC and then reflect in the spec as well.

proposals/3814-dehydrated-devices-with-ssss.md
dkasak, line 296: Can we name the RFC we are basing our Base64 encoding on?

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+
+#### Encryption key
+
+TODO: Explain why the double derivation is necessary.
+
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`.
+
+The randomly generated encryption key *must* be expanded using the HMAC-based
+Key Derivation function defined in [RFC5869].
```
dkasak, line 269:
```suggestion
The randomly generated encryption key (`RANDOM_KEY` in the example below)
*must* be expanded using the HMAC-based Key Derivation function defined in
[RFC5869].
```

proposals/3814-dehydrated-devices-with-ssss.md
dkasak: Also the quotes around `"Device ID"` make me think this is a constant. Will leave a suggestion for that as well.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+
+```math
+\begin{aligned}
+    DEVICE\_KEY
+    &= \text{HKDF} \left(\text{``Device ID``}, RANDOM\_KEY, \text{``dehydrated-device-pickle-key"}, 32\right)
+\end{aligned}
+```
```
dkasak, line 276:
````suggestion
```math
\begin{aligned}
    \text{DEVICE\_KEY}
    &= \text{HKDF} \left(\text{Device ID}, \text{RANDOM\_KEY}, \text{``dehydrated-device-pickle-key"}, 32\right)
\end{aligned}
```
````

File: proposals/3814-dehydrated-devices-with-ssss.md [hunk 30]
```diff
@@ -0,0 +1,407 @@
...
+
+The `device_key` is then further expanded into a AES256 key, HMAC key and
+initialization vector.
+
+
+```math
+\begin{aligned}
+    AES\_KEY \parallel HMAC\_KEY \parallel AES\_IV
+    &= \text{HKDF}\left(0,DEVICE\_KEY,\text{``Pickle"},80\right)
+\end{aligned}
```
dkasak, line 286:
```suggestion
\begin{aligned}
    \text{AES\_KEY} \parallel \text{HMAC\_KEY} \parallel text{AES\_IV}
    &= \text{HKDF}\left(0,\text{DEVICE\_KEY},\text{``Pickle"},80\right)
\end{aligned}
```

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+The plain-text is encrypted with [AES-256] in [CBC] mode with [PKCS#7] padding,
+using the key $`AES\_KEY`$ and the IV $`AES\_IV`$ to give the cipher-text.
```
dkasak, line 290:
```suggestion
using the key $`\text{AES\_KEY}`$ and the IV $`\text{AES\_IV}`$ to give the cipher-text.
```

dkasak (changes requested):

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,407 @@
...
+something like "shrivelled sessions", so that the full expansion of this MSC
+title would be "Shrivelled Sessions with Secure Secret Storage and Sharing", or
+SSSSSS.  However, despite the alliterative property, the term "shrivelled
+sessions" is less pleasant, and "dehydrated device" is already commonly used to
+refer to this feature.
+
+The alternatives discussed in MSC2697 are also alternatives here.
+
+
+## Security considerations
```
dkasak, line 375: This section should contain a discussion of malicious dehydrated devices injected by the server.

The MSC in its current form is almost purely focused on the mechanics of creation and rehydration of dehydrated devices. There is very little discussion of how these devices should be treated by message senders.

And yet, these are not ordinary devices, as evidenced by the fact that we are proposing special UI/UX for them in the section concerning the `dehydrated` flag. The obvious concern is that dehydrated devices could end up being hidden or obscured from the user's view—even more so than ordinary devices—and therefore bear an even greater risk of malicious injection.

The long-term plan is to move to a model where a user's devices must be signed by the user's cryptographic identity in order to be considered valid (see [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)) . Given that context, and the fact that dehydrated devices are a completely new feature, I strongly recommend that this MSC should require that a dehydrated device MUST be signed by a pinned (TOFU-trusted) user identity in order to be considered valid. If the dehydrated device is not signed, or is signed by a user identity which is not the one that is currently pinned by the client, the dehydrated device MUST be ignored by senders as if it it does no exist. That is, clients MUST NOT send any to-device messages to such a device nor accept any to-device messages from it.

uhoreg: > That is, clients MUST NOT send any to-device messages to such a device nor accept any to-device messages from it.

Dehydrated devices shouldn't be sending to-device messages, so it's probably safe to say that we should not accept any to-device messages from any devices marked as dehydrated.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,390 @@
...
+  will run out of one-time keys, and reduce the number of events that the
+  dehydrated device will need to decrypt.
+
+`PUT /dehydrated_device`
+
+```jsonc
+{
+  "device_id": "dehydrated_device_id",
+  "device_data": {
+    "algorithm": "m.dehydration.v2.olm"
```
zecakeh, line 75: It seems like the `v2` algorithm name doesn't end with `.olm`?

uhoreg: Good catch.  Fixed

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,390 @@
...
+
+The `device_keys`, `one_time_keys`, and `fallback_keys` fields use the same
+structure as for the [`/keys/upload`] request.
+
+We add a new optional property to the device keys: `dehydrated`, which is set to
+`true` for dehydrated devices.  Defaults to `false`.  Clients that support
+dehydrated devices *must not* encrypt to devices marked as being a dehydrated
+device if they are not cross-signed.  Clients should also drop any to-device
+messages from a device marked as being a dehydrated device, since dehydrated
+device should not be sending messages.  Clients can use also this flag to for
```
andybalaam, line 55:
```suggestion
device should not be sending messages.  Clients can use also this flag for
```

andybalaam, line 55: In the case of encrypted messages, we can identify the sending device using the sender key, or MSC4147 device info. 2 questions:

1. What about unencrypted messages? Even if the event content contains a device id, we can't trust that a malicious homeserver didn't just fake it. Should we just allow all unencrypted to-device messages?
2. What if the sender key is not found, maybe because the device is logged out? Should we assume the device is not dehydrated and process it as normal? This gives a way to work around the messages being dropped, but it's hard to see what else we can do in this case.

andybalaam: cc @dkasak

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,406 @@
...
+This may be addressed by using [fallback
+keys](https://spec.matrix.org/v1.9/client-server-api/#one-time-and-fallback-keys),
+and clients are recommended to create a fallback key for the dehydrated device.
+
+To reduce the chances of one-time key exhaustion, if the user has an active
+client, it should periodically (e.g. once a week) replace the dehydrated device
+with a new dehydrated device with new one-time keys.  If a client does this,
+then it runs the risk of losing any megolm keys that were sent to the dehydrated
+device, but the client should have received those megolm keys itself, so this
+should not be a problem.
```
richvdh, line 319: I believe this is the approach taken in the sample implementation? We should probably hoist this sentence ("clients should periodically replace the dehydrated device") into the main proposal: I wouldn't necessarily read through all the "Potential issues" when I was writing an impl

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,406 @@
...
+## Proposal
+
+### Dehydrating a device
+
+The dehydration process is similar as in [MSC2697]. One important change is that
+the dehydrated device, the public device keys, and one-time keys are all
+uploaded in the same request. This change should prevent the creation of
+dehydrated devices which do not support end-to-end encryption.
+
+To upload a new dehydrated device, a client will use `PUT /dehydrated_device`.
```
richvdh, line 38: we should maybe write down how we know if the server supports the feature. AIUI the sample impl calls `GET /dehydrated_device` and checks for an `M_UNRECOGNISED` error?

proposals/3814-dehydrated-devices-with-ssss.md
andybalaam: Differences:

* Omit `next_batch` instead of returning empty `events` array
* Rename the request parameter `from` instead of `next_batch`
* (Check that we never delete to-device messages in the latest version of this MSC)

proposals/3814-dehydrated-devices-with-ssss.md
Johennes: Is this intended to be an optional feature for servers?

benkuly (changes requested):

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+#### Encryption key
+
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`, encoded using
+unpadded base64.
+
+A 96-bit (12-byte) nonce is randomly generated; each time a device is
+dehydrated, a new nonce must be generated.
+
+The plain-text is encrypted with ChaCha20-Poly1305 as defined in
```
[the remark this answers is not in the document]
Johennes: There is a brief summary of why the ChaCha algorithms were introduced over AES in the [RFC](https://datatracker.ietf.org/doc/html/rfc8439#section-1). I think ideally the MSC should motivate its choice explicitly though.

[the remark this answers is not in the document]
benkuly: As I understand it, the advantages mentioned in the RFC (performance, cache-collision timing attacks) are not relevant for this use case. The disadvantages mentioned above, on the other hand, strongly outweigh them in my opinion.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+will not be able to share megolm keys.  This issue is not unique to dehydrated
+devices; this also occurs when devices are offline for an extended period of
+time.
+
+This may be addressed by using [fallback
+keys](https://spec.matrix.org/v1.9/client-server-api/#one-time-and-fallback-keys),
+and clients are recommended to create a fallback key for the dehydrated device.
+
+To reduce the chances of one-time key exhaustion, if the user has an active
+client, it should periodically (e.g. once a week) replace the dehydrated device
```
benkuly, line 330: This information should be stored somewhere to synchronize the creation between active devices and prevent that dehydrated devices are rotated too often. Maybe next to the `"dehydrated": true,` as `"createdAt": "2025-05-06T18:25:43.511Z"`. That way, the information is signed and cannot be manipulated by a malicious homeserver. Another advantage would be that this information is synced via `device_lists` and no polling or an additional local cache for this metadata is needed.

benkuly (changes requested):
I am currently in the process of implementing the MSC, but in my opinion there are still fundamental concerns here that should be looked at again. There are far too many stumbling blocks in the way:

1. A completely new crypto algorithm instead of utilising existing algorithms from the Matrix Spec.
2. A lack of consideration of the rotation.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+### Dehydrating a device
+
+The dehydration process is similar as in [MSC2697]. One important change is that
+the dehydrated device, the public device keys, and one-time keys are all
+uploaded in the same request. This change should prevent the creation of
+dehydrated devices which do not support end-to-end encryption.
+
+To upload a new dehydrated device, a client will use `PUT /dehydrated_device`.
+Each user has at most one dehydrated device; uploading a new dehydrated device
+will remove any previously-set dehydrated device.
```
MadLittleMods, line 40: Why is there a limitation of one dehydrated device per user?

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+uploaded using the `PUT /dehydrated_device`, so calling `DELETE
+/dehydrated_device` is not necessary when replacing the dehydrated device.  It
+is only necessary to call when the dehydrated device is to be removed and may
+not be replaced.  For example, if a client resets the user's cross-signing keys,
+but does not create a new dehydrated device, it should call this endpoint, as
+the dehydrated device would be signed with an outdated old cross-signing key.
+
+`DELETE /dehydrated_device`
+
+Response:
```
jevolk, line 220: Should DELETE for a non-existent device return `200 {}` or `404`

jevolk (approved):
Very well written MSC. My pre-read prior to implementing gave useful preparation. API surface is succinct; elegantly reuses existing device infrastructure. Objects are partitioned well against cryptographic misuse.

Perhaps top~10 on quality in the corpus. This should be shepherded into the next phases.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+  `m.dehydration.v2` algorithm.
+- `initial_device_display_name` (optional): the display name for the device
+- `device_keys` (required), `one_time_keys` (optional), `fallback_keys`
+  (optional): the dehydrated device's device keys, one-time keys, and fallback
+  keys, respectively, using the same structure as for the [`/keys/upload`]
+  request.
+
+We add a new optional property to the device keys: `dehydrated`, which is set to
+`true` for dehydrated devices.  Defaults to `false`.  Clients that support
+dehydrated devices *must not* encrypt to devices marked as being a dehydrated
```
dkasak, line 64:
```suggestion
dehydrated devices *MUST NOT* encrypt to devices marked as being a dehydrated
```

proposals/3814-dehydrated-devices-with-ssss.md
uhoreg: No, it's not intended to be optional.  But it's not uncommon for clients to check for support in ways other than checking `/versions`, e.g. if a server implements the feature after FCP but before it gets released in a spec release, or to use the unstable endpoints.

proposals/3814-dehydrated-devices-with-ssss.md
```diff
@@ -0,0 +1,421 @@
...
+   └────────────────────────┴───────────────┴──────────────────┘
+```
+
+The data is then encrypted and encoded as follows.
+
+#### Encryption key
+
+The encryption key used for the dehydrated device will be randomly generated
+and stored/shared via SSSS using the name `m.dehydrated_device`, encoded using
+unpadded base64.
```
richvdh, line 287: Note: assuming this lands before MSC4108, we'll need to extend QR code login to include it.

martinezvelasco25-art (approved):

proposals/3814-dehydrated-devices-with-ssss.md
richvdh: Fixed in 72c0656.

proposals/3814-dehydrated-devices-with-ssss.md
richvdh: The server no longer deletes the dehydrated device, as of 3827bc0

proposals/3814-dehydrated-devices-with-ssss.md
richvdh: Done: 9ea19e2f

PR #3815, opened 2022-05-13
[WIP] 3D Worlds

robertlong:


proposals/3815-worlds.md
```diff
@@ -0,0 +1,67 @@
...
+}
+```
+
+### `m.world` state event
+
+Describes the 3D scene to use for a Matrix room as well as a thumbnail to use while the scene is loading or for other use in the client UI.
+
+```json
+{
+  "type": "m.world",
```
ara4n, line 29: what's the state key? presumably we should be able to support multiple GLBs per world in order to represent the compose hierarchy?

robertlong: glTF composition will likely happen within the glTF itself. This allows us to reference and parent glTF scenes to nodes, override properties, and more. That'll be dependent on a glTF spec proposal. I think that's more desirable than just trying to merge multiple glTFs into one scene and allows us to hopefully eventually specify this within the correct standards body.

ara4n: kk. if you compose like that, will the GLTFs (or GLB) binary diff nicely against each other?

PR #3816, opened 2022-05-17
MSC3816: Clarify Thread Participation

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/threads-participated/proposals/3816-clarify-thread-participation.md)

----

Implementation: matrix-org/synapse#12766
FCP proposal: https://github.com/matrix-org/matrix-spec-proposals/pull/3816#issuecomment-1129139715

### DIFF
--- a/proposals/3440-threading-via-relations.md
+++ b/proposals/3440-threading-via-relations.md
@@ -57,8 +57,10 @@ would include additional information in the `unsigned` field:
   The latest event should be serialised in the same form as the event itself;
   this includes adding any bundled aggregations for the event (and applying edits).[^1]
 * `count`: An integer counting the number of `m.thread` events
-* `current_user_participated`: A flag set to `true` if the current logged in user
-  has participated in the thread
+* `current_user_participated`: A boolean flag, which is set to `true` if the
+  current logged in user has participated in the thread. The user has participated if:
+  * They created the current event.
+  * They created an event with a `m.thread` relation targeting the current event.
 
 #### Rich replies in a thread
 
--- a/proposals/3816-clarify-thread-participation.md
+++ b/proposals/3816-clarify-thread-participation.md
@@ -31,8 +31,8 @@ participated in the thread (e.g. both tables would be `true` in the example abov
 
 ## Proposal
 
-The definition of the `current_user_participated` flag from
-[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) is updated to be:
+The [definition of the `current_user_participated` flag](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#event-format)
+from the bundled aggregations for `m.thread` relations is updated:
 
 > A boolean flag, which is set to `true` if the current logged in user has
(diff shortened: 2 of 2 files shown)


proposals/3440-threading-via-relations.md
```diff
@@ -57,8 +57,10 @@ would include additional information in the `unsigned` field:
   The latest event should be serialised in the same form as the event itself;
   this includes adding any bundled aggregations for the event (and applying edits).[^1]
 * `count`: An integer counting the number of `m.thread` events
-* `current_user_participated`: A flag set to `true` if the current logged in user
-  has participated in the thread
+* `current_user_participated`: A boolean flag, which is set to `true` if the
+  current logged in user has participated in the thread. The user has participated if:
+  * They created the current event.
+  * They created an event with a `m.thread` relation targeting the current event.
```
gsouquet, line 63: We will eventually allow polls within a thread. Would we consider a poll response as a participation in to a thread?

If we say yes to this, we'll need to refine that bullet point a bit

clokep: That heavily complicates things -- would a reaction to a message in a thread count as a participant then? I don't believe so. Looking at related events of all events in a thread will be quite expensive / complicated to implemented.

Polls in particular didn't seem to have been considered in the original design parameters though: vector-im/element-web#19846, while reactions were specifically considered and discarded.

clokep: I think I'm going to punt this and say that polls needs to figure this out since threads is already merged. 🗳️

turt2live (approved):

turt2live:
@mscbot fcp merge

turt2live:
@mscbot concern Implementation looks good enough to continue, but would like to see it reviewed before being comfortable

clokep:
> @mscbot concern Implementation looks good enough to continue, but would like to see it reviewed before being comfortable

@erikjohnston has approved the Synapse changes. ✅

turt2live:
@mscbot resolve Implementation looks good enough to continue, but would like to see it reviewed before being comfortable

richvdh:
I should say: all of those comments are suggestions for clarifications. I don't think they change the meaning in any way.

proposals/3816-clarify-thread-participation.md
```diff
@@ -0,0 +1,70 @@
...
+
+[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) defines relations
+for threads and includes a `current_user_participated` flag, which is not fully
+defined:
+
+> A flag set to `true` if the current logged in user has participated in the thread
+
+Known implementations implement this as whether the currently logged in user has
+sent an event with an `m.thread` relation targeting the event, but this does not
+make sense if the requesting user was the original event in the thread.
```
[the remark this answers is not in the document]
clokep: Ah, yeah it was meant to say "if the requesting user had sent the original event in the thread". Your version is much clearer though, thank you!

turt2live:
Technically this is merged because it's already had its intended effect - counting as such to clear the blocker.

PR #3588, opened 2021-12-23
WIP: MSC3588: Encrypted Stories As Rooms

krille-chan:
[Rendered](https://github.com/krillefear/matrix-doc/blob/main/proposals/3588-stories-as-rooms.md)

Signed-off-by: Christian Pauly christian-pauly@posteo.de

PR #3818, opened 2022-05-18
MSC3818: Copy room type on upgrade

Mikaela:
This or equivalent MSC is blocking vector-im/element-web#19208 <s>and matrix-org/synapse#11896.</s>

[Rendered](https://github.com/Mikaela/matrix-spec-proposals/blob/copy-room-type-on-upgrade/proposals/3818-copy-room-type-on-upgrade.md)

Resolves: matrix-org/matrix-spec#911

Signed-off-by: Aminda Suomalainen 

Implementations: https://github.com/matrix-org/synapse/pull/12786

FCP proposal: https://github.com/matrix-org/matrix-spec-proposals/pull/3818#issuecomment-1189171267

### DIFF
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -23,7 +23,9 @@ It becomes:
 
 ## Potential issues
 
-Non-applicable.
+Some room types such as Spaces also require copying over state events as a part of the update progress,
+in case of Spaces, `m.space.child` events. However as that can be changed later and done by the client,
+it's out of scope for this MSC.
 
 ## Alternatives
 
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -29,7 +29,11 @@ it's out of scope for this MSC.
 
 ## Alternatives
 
-Non-applicable.
+A suggested alternative is having every room type specify their own update process if they use other
+room types. However this would complicate the MSC process with even simple client-side proposals
+requiring also a server-side implementation. This could also result in room types dependent on a
+particular server software or discourage using Matrix for a smaller project where an MSC wasn't
+otherwise consider necessary.
 
 ## Security considerations
 
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Copy room type on upgrade.
+# MSC3818: Copy room type on upgrade
 
 There can never be enough templates in the world, and MSCs shouldn't be any different. The level
 of detail expected of proposals can be unclear - this is what this example proposal (which doubles
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -1,9 +1,5 @@
 # MSC3818: Copy room type on upgrade
 
-There can never be enough templates in the world, and MSCs shouldn't be any different. The level
-of detail expected of proposals can be unclear - this is what this example proposal (which doubles
-as a template itself) aims to resolve.
-
 Unless the room upgrade API specifies that room type must be copied over, clients cannot rely on
 rooms staying the same type leading to trouble.
 
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -17,7 +17,7 @@ The Spec currently specfies this in [section 11.32.3. server behaviour](https://
 It becomes:
 
 > 2. Creates a replacement room with a `m.room.create` event containing a `predecessor` field, a
-> `type` field if it was set in the previous room, and the applicable `room_version`.
+> `type` field set to what it was in the previous room (if it was set), and the applicable `room_version`.
 
 
 ## Potential issues
--- a/proposals/3818-copy-room-type-on-upgrade.md
+++ b/proposals/3818-copy-room-type-on-upgrade.md
@@ -8,7 +8,7 @@ rooms staying the same type leading to trouble.
 
 This MSC proposes that the room upgade API MUST copy the [room type](https://spec.matrix.org/v1.2/client-server-api/#types)
 over to the new room. Otherwise clients cannot trust that to happen and [Spaces](https://spec.matrix.org/v1.2/client-server-api/#spaces)
-or [MSC3588](https://github.com/matrix-org/matrix-doc/pull/3588) Story rooms may incorrectly become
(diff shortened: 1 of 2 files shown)


proposals/3818-copy-room-type-on-upgrade.md
```diff
@@ -0,0 +1,41 @@
...
+rooms breaking user-experience.
+
+The Spec currently specfies this in [section 11.32.3. server behaviour](https://spec.matrix.org/v1.2/client-server-api/#server-behaviour-16):
+
+> 2. Creates a replacement room with a `m.room.create` event containing a `predecessor` field and the applicable `room_version`.
+
+It becomes:
+
+> 2. Creates a replacement room with a `m.room.create` event containing a `predecessor` field, a
+> `type` field if it was set in the previous room, and the applicable `room_version`.
```
andybalaam, line 20:
```suggestion
> `type` field set to what it was in the previous room (if it was set), and the applicable `room_version`.
```

andybalaam (approved):
Looks good to me!

proposals/3818-copy-room-type-on-upgrade.md
```diff
@@ -0,0 +1,42 @@
...
+
+Non-applicable.
+
+## Unstable prefix
+
+Non-applicable.
+
+## Dependencies
+
+Non-applicable.
```
FSG-Cat, line 42:
```suggestion
```

These parts are redundant and therefore should be safe to remove. This is in the class of simple MSCs that dont need to follow  the template as religiously.

Mikaela: `#matrix-spec:matrix.org` discussion:

me:

>> https://github.com/matrix-org/matrix-spec-proposals/pull/3818 is the initial version, can I just erase the non-applicable sections?

@clokep :

> Usually it is best to leave most of them to show you've thought about it. 🙂 

* https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$9pH8bBCsoKyCVgvuJ2n4hvTuCjoGz2hX2yfkgL-ayC4?via=matrix.org&via=libera.chat&via=element.io

FSG-Cat: This matter was discussed out of band before the comment by @uhoreg and i agree with them.

proposals/3818-copy-room-type-on-upgrade.md
```diff
@@ -0,0 +1,42 @@
...
+
+It becomes:
+
+> 2. Creates a replacement room with a `m.room.create` event containing a `predecessor` field, a
+> `type` field set to what it was in the previous room (if it was set), and the applicable `room_version`.
+
+
+## Potential issues
+
+Non-applicable.
```
FSG-Cat, line 26:
```suggestion
The MSC can be safely implemented without going thru a unstable phase due to that this behavior is covered as an implementation detail before this MSC is made spec. 

## Potential issues

This MSC is dropin compatible with the current spec and therefore no issues are forseen.
```

This bit of text should clarify that the MSC is safe to essentially implement as fast as possible without it being a concern. Ofc all implementations should go thru the vetting processes of the individual projects.

clokep: I think there *is* a potential issue though, for spaces there's special behavior, the state events for rooms in a space must be moved over (the `m.space.child` events). If other room types have special behavior which is not implemented then the room type might get moved over without moving over the other related information...which could break things slightly.

FSG-Cat: @clokep good catch. This should probably be addressed.

uhoreg: With additional state events that need to be moved over, such as the `m.space.child` events, that can always be done after the upgrade is done, and could be done client-side.  So that is not necessarily a blocker for this MSC.  However, the `type` field has to be copied by the server since the `create` event can't be changed.

So it's a question of: do we want this MSC to fix every problem with upgrading spaces, or just focus on the one specific issue?

(Also, with upgrading spaces, you'd want to update the state events in the child rooms to point to the new parent.  Which is probably more problematic.)

turt2live: We also have several other cases where changing the creation content would be useful - maybe `/upgrade` can take a similar object shape to `/createRoom`, allowing clients/operators to change things like `m.federate`?

Mikaela: Does https://github.com/matrix-org/matrix-spec-proposals/pull/3818/commits/b2b1889e8f774ea24af1ae26b7c2641db81d563e appropiatively resolve this discussion?

proposals/3818-copy-room-type-on-upgrade.md
```diff
@@ -0,0 +1,48 @@
...
+
+Some room types such as Spaces also require copying over state events as a part of the update progress,
+in case of Spaces, `m.space.child` events. However as that can be changed later and done by the client,
+it's out of scope for this MSC.
+
+## Alternatives
+
+A suggested alternative is having every room type specify their own update process if they use other
+room types. However this would complicate the MSC process with even simple client-side proposals
+requiring also a server-side implementation. This could also result in room types dependent on a
```
Mikaela, line 34: For example https://github.com/matrix-org/matrix-spec-proposals/pull/3588 is entirely clientside implemented by FluffyChat and MinesTrix with the exception of having a custom room type (and blacklist, which in my opinion is trivially recreated). This alternative would also require a server-side implementation.

Additionally can servers store infinite amount of different types of rooms and their unstable prefixes as something to be copied over?

KitsuneRal (approved):

proposals/3818-copy-room-type-on-upgrade.md
FSG-Cat: I personally think that it might be counted as resolving my personal concerns.

Mikaela: @clokep @uhoreg @turt2live May I mark this conversation as resolved as per https://github.com/matrix-org/matrix-spec-proposals/commit/b2b1889e8f774ea24af1ae26b7c2641db81d563e?

clokep: Seems fine to resolve it.

turt2live:
The author believes this is ready for FCP.

anoadragon453 (approved):
LGTM for the most part. Note that whether this field is copied over is an implementation detail today, and Synapse will already copy it over. That being said, as the type can be quite crucial to the functionality of the room, mandating it to be copied over makes sense in a lot of ways.

I think this is ready for a review from a wider audience, so I'm going to push it forward.

@mscbot fcp merge

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1198

turt2live:
Merged 🎉

PR #3819, opened 2022-05-19
Allowing widgets to send/receive to-device messages

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/widgets-send-recv-toDevice/proposals/3819-to-device-messages-for-widgets.md)

### DIFF
--- a/proposals/3819-to-device-messages-for-widgets.md
+++ b/proposals/3819-to-device-messages-for-widgets.md
@@ -1,0 +1,247 @@
+# MSC3819: Allowing widgets to send/receive to-device messages
+
+Widgets (embedded HTML applications in Matrix) currently have a relatively large surface area
+they can use for interacting with their attached client, primarily in the context of a room. They
+can [send/receive events with MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762),
+[navigate to rooms with MSC2931](https://github.com/matrix-org/matrix-spec-proposals/pull/2931),
+and even [open dialogs with MSC2790](https://github.com/matrix-org/matrix-spec-proposals/pull/2790),
+but they can't act as a whole other Matrix client just yet.
+
+This MSC forms part of a larger, ongoing, question about how to embed other Matrix clients into another
+client or room for access. An increasingly more popular client development option is to build out an
+entirely new Matrix client and want to embed that within another client (as a widget) to avoid the
+user needing to switch apps. To support this, we need to consider both long term and short term impact
+of the changes we propose. This MSC aims closer to the short term.
+
+A longer term solution to the problem of clients wanting to be embedded in other clients might still
+be widgets, though with a system like [MSC3008](https://github.com/matrix-org/matrix-spec-proposals/pull/3008)
+to restrict access to the client-server API more effectively. For this MSC's purpose though, we're
+aiming to cover a specific subset of the client-server API: to-device messages.
+
+While we could expose the entire client-server API over `postMessage` (or similar) for embedded
+clients to access, the permissions model gets hairy and difficult to secure on the client side. Instead,
+we're exploring what it would look like to special case what is needed for specific applications, as
+needed, starting with to-device messages.
+
+To-device messaging is described [here](https://spec.matrix.org/v1.2/client-server-api/#send-to-device-messaging)
+with practical applications for widget-ized clients being implementations of
+[MSC3401 - Native group VoIP](https://github.com/matrix-org/matrix-spec-proposals/pull/3401) for now.
+
+## Prerequisite background
+
+*Author's note: This is copied from [MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762).*
+
+Widgets are relatively new to Matrix and so the terminology and behaviour might not be known to all
+readers. This section should clarify the components of widgets that are applicable to this MSC without
+going on a deep dive into widgets in general.
+
+Widgets are embedded HTML/JS/CSS applications in a client which use the `postMessage` API to talk
+to the client. This communication allows widgets to provide enhanced functionality such as sticker
+pickers (when applied to a user) or performance dashboards (in rooms).
+
+One of the first things that happens over this communication channel is a "capabilities negotiation"
+where the client asks the widget what permissions it wants, and the widget replies with its ideal
+set. The client then either decides or asks the user if the permissions requested are okay.
+
+All communication over the channel is done in a simple request/response flow, using actions to
+describe the request. For the capabilities negotiation, this would be the client sending the widget
+a request with an `action` of `capabilities`, and the widget would respond to that request with a
+response object.
+
+The channel in which communication occurs is called a "session", where the session is "established"
+after the capabilities negotiation. Sessions can only be terminated by the client.
+
+The Widget API is split into two parts: `toWidget` (client->widget) and `fromWidget` (widget->client).
+They are differentiated by where the request originates.
+
+## Proposal
+
+Inspired heavily from [MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762), we
+introduce new capabilities for to-device messages:
+
+* `m.send.to_device:<event type>` (eg: `m.send.to_device:m.call.invite`) - Used for sending to-device
+  messages of a given type.
+* `m.receive.to_device:<event type>` (eg: `m.receive.to_device:m.call.invite`) - Used for receiving
+  to-device messages of a given type.
+
+These capabilities open up access to the following respective actions, when approved:
+
+**`fromWidget` action of `send_to_device`**
+
+```json5
+{
+  // This is a standardized widget API request.
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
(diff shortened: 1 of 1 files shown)


proposals/3819-to-device-messages-for-widgets.md
```diff
@@ -0,0 +1,247 @@
...
+  "requestid": "generated-id-1234",
+  "action": "send_to_device", // value defined by this proposal
+  "data": {
+    // Same structure as the `/sendToDevice` HTTP API request body
+    "@target:example.org": {
+      "DEVICEID": { // can also be a `*` to denote "all of the user's devices"
+        "example_content": "put your real message here"
+      }
+    }
+  }
```
robintown, line 85: We need to include the event type somewhere in here. I'd suggest an interface of

```typescript
{
  eventType: string;
  messages: { [userId: string]: { [deviceId: string]: unknown } };
}
```

robintown (changes requested):

proposals/3819-to-device-messages-for-widgets.md
robintown: We also need to control whether the event is encrypted or not

```typescript
{
  type: string;
  encrypt: boolean;
  messages: { [userId: string]: { [deviceId: string]: unknown } };
}
```

proposals/3819-to-device-messages-for-widgets.md
```diff
@@ -0,0 +1,247 @@
...
+  "api": "toWidget", // note that we're sending *to* the widget here
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "send_to_device", // value defined by this proposal
+  "data": {
+    "type": "m.call.invite",
+    "sender": "@source:example.org",
+    "content": {
+      // ... as required for the event schema
+    }
```
robintown, line 161: This needs a field to indicate whether the content was encrypted, so widgets can determine whether they trust the message's integrity.

```suggestion
  "data": {
    "type": "m.call.invite",
    "sender": "@source:example.org",
    "encrypted": true,
    "content": {
      // ... as required for the event schema
    }
```

proposals/3819-to-device-messages-for-widgets.md
```diff
@@ -0,0 +1,247 @@
...
+  },
+  "response": {}
+}
+```
+
+## Potential issues
```
Fox32, line 189: While working with the current support of MSC3819 in Element we noticed a problem:

The send endpoint of the toDevice message API allows to target specific device
ids, but there is currently no way to know a specific device id for a widget.
[Neither is it possible to embed a device id into the URL](https://github.com/matrix-org/matrix-widget-api/blob/master/src/templating/url-template.ts#L31-L42),
nor does a received toDevice message indicate the sender device to reply to it.
Sure, it is possible to target all devices via `*`, but using a specific device
id makes sending much more efficient.

Element Call, which this MSC initially was created for, is cheating a little bit.
As Element Call is a special widget anyway, it is created with
[custom code instead of the normal widget API setup code](https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/models/Call.ts#L654).
This allows to embed additional parameters into the widget URL, including the device id.
From the perspective of bringing the product to the market quickly, I think it
is fine, but we should focus on getting this MSC into a complete state.

Therefore I suggest to amend small section here:

```
**Passing device id to widgets**

A new template variable is added to the available options for a widget URL:

* `device_id` - The device id of the client instance which is rendering the widget.

```

And extend the unstable prefix section with:

```
* `org.matrix.msc3819.device_id` in place of the proposed variable.
```

If this extension is desired, I'm open to add the implementation to the `matrix-react-sdk` and `matrix-widget-api`.

Fox32: I started working on implementing this addition over at https://github.com/matrix-org/matrix-widget-api/pull/78

dhenneke: In matrix-org/matrix-widget-api#78, we decided to slightly rename the parameter. Thus the amendment should be as follows:

```
**Passing device id to widgets**

A new template variable is added to the available options for a widget URL:

* `matrix_device_id` - The device id of the client instance which is rendering the widget.
```

And extend the unstable prefix section with:

```
* `org.matrix.msc3819.matrix_device_id` in place of the proposed variable.
```

proposals/3819-to-device-messages-for-widgets.md
BillCarsonFr: I am not sure about this API. Currently if you want to send a message to several devices, the content will be repeated n times as part of the map<user_id, map<device_id, content>

maybe something like
```
{
  type: string;
  encrypted: boolean;
  recipient_devices: { [userId: string]: [deviceId: string] };
  content: any
}

File: proposals/3819-to-device-messages-for-widgets.md [hunk 4]
```diff
@@ -0,0 +1,248 @@
...
+{
+  // This is a standardized widget API request.
+  "api": "toWidget", // note that we're sending *to* the widget here
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "send_to_device", // value defined by this proposal
+  "data": {
+    "type": "m.call.invite",
+    "sender": "@source:example.org",
+    "encrypted": true,
```
BillCarsonFr, line 159: This sounds like a bad idea? If a clear text event is sent with an `encrypted` field in the event then it could be miss-interpreted has beeing a decrypted plain text message

toger5: I am just noticing, that this is still outside of the `content` is there a way for the sender to modify this part of the json?

The homeserver however could right?


The spec could either enforce a widget driver to overwrite this. (but custom events that want to use a field named like this cannot exist anymore then. We would basically need to enforce in the event spec that encrypted is a reserved field.)
So naming it sth like `m.encrypted` would make more sense if we keep the structure as is

Another solution one can imagine is changing the format to:
```json
"data":{
  "content" : {...what_was_previous_stored_in_data}
  "room_id": string
  "encrypted": boolean
}
```
So that the content from the HS is never altered and that the fields the widget driver adds can never be injected by the sender.
This would very much a breaking change.

Last we could put this info next to data. But that would be odd since it would then be on the same level of:
`action`,`response`,`api`... which are clearly specified widget api fields.
Maybe introducing another toplevel property called `request_info` or `data_meta` or just `widget_info` that is exclusive to the widget and since it is on the top level only the widget driver can define it. It would then include all the data that gets added by the widget driver and is always the trusted source.

BillCarsonFr: At this point I thing I am agains letting the widget decide to encrypt when sending, and also letting the widget if it should accept a clear to-device message.

My proposal would be to remove this `encrypted` field.

And when sending a to-device:
- If it is a room widget and the room is encrypted then encrypt the to-device, if the room is not encrypted then do not encrypt. 
- If it is a "global" widget, then always encrypt.

On receiving a to-device:
- If it is a room widget: If the room is encrypted only forward encrypted to-device (sucessfully decrypted). This should respect the room encryption settings; That is, if the room only allow verified devices, then to-device sent by un-verified devices should not be forwarded. It should also respect the exclusion of insecure device as per the user setting
- For global widget, use the default user preferences for insecure devices

toger5: The only feature we would loose is unencrypted to-device in encrypted rooms.
This has quiet a little use-case. The only reason i could find for this is if to-device is used as an ephemeral message to a lot of participants. Then encryption is expensive bandwidth and compute wise.
But this is somewhat artificial since it very much sounds like the exact scenario room events were designed for.
But in all other scenarios encrypted to-device just make much more sense (compared to unencrypted).
I will draft a change and see how this feels.

proposals/3819-to-device-messages-for-widgets.md
```diff
@@ -0,0 +1,248 @@
...
+      "DEVICEID": { // can also be a `*` to denote "all of the user's devices"
+        "example_content": "put your real message here"
+      }
+    }
+  }
+}
+```
+
+The client upon receipt of this will validate that the widget has an appropriate capability to send
+the to-device message. If the widget is approved for such a capability, the client **MUST** encrypt
```
BillCarsonFr, line 90: I think we need to be a lot more restricted.
Widgets should only have access to a sub-set of the to-device traffic, internal/crypto to-device traffic should be excluded by default (`m.room.key`, `m.secret.send`, `m.verification` ...)
And even more there should be a way to scope the to-device traffic to the room.

My proposal would be to add a new `scope` property in the event content:
```
{
        "sender": "@alice:example.com",
        "type": "m.custom",
        "content": {
            "scope": {
                "room_id": "!726s6s6q:example.com",
                "id": "org.matrix.msc3401.call"
            }
           "example_content": "put your real message here"
        }
}
```

This would allow for the widget to only get traffic scoped to the room it sits in. (Global widget would get all traffic -excluding reserved crypto types-?). To device content with no `scope` will not be passed to widgets.

On the sending side the scope would be added automatically by the widget driver.

PR #3823, opened 2022-05-23
MSC3823: Account Suspension

Yoric:
[Rendered](https://github.com/Yoric/matrix-doc/blob/yoric/errcode-suspended/proposals/3823-code-for-account-suspension.md)

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3823#issuecomment-2433390226)

### DIFF
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -1,0 +1,81 @@
+# MSC3823: A code for account suspension
+
+This MSC introduces a new error code that servers may send to clarify that an account has been
+suspended *temporarily* but may still be reactivated.
+
+## Proposal
+
+### Introduction
+
+Matrix has a code `M_USER_DEACTIVATED` that a server may return to describe an account that has been
+deactivated. So far, this code has been used to represent accounts that have been *permanently*
+deactivated. In particular, clients that interpret this error code display it imply that the account
+has been *permanently* deactivated.
+
+However, some countries (e.g. UK) have laws that require the ability to appeal account
+deactivations. This requires the ability to specify that an account is *reversibly*
+suspended and let users know about the appeals procedure.
+
+This MSC simply introduces a new error code `M_USER_ACCOUNT_SUSPENDED` that servers may send to
+clarify that an account has been suspended but that the solution may still be resolved either by
+an appeal or by e.g. clearing up some abusive messages.
+
+This MSC does *not* specify a mechanism to suspend or unsuspend the account or to handle appeals.
+
+### Proposal
+
+Introduce a new error code `M_USER_ACCOUNT_SUSPENDED`. This error code MAY be sent by the server
+whenever a client attempts to use an API on behalf of a user whose account has been suspended.
+
+| Name | Type | Value |
+|------|------|-------|
+| `href` | string | (optional) If specified, a URL containing more information for the user, such as action needed. |
+
+The client is in charge of displaying an error message understandable by the user in case of `M_USER_ACCOUNT_SUSPENDED`,
+as well as a link to `href`.
+
+The web server serving `href` is in charge of localizing the message, using existing HTTP mechanisms,
+to adapt the page to the end user's locale.
+
+#### Examples
+
+Returning a static page:
+
+```json
+{
+  "errcode": "M_USER_ACCOUNT_SUSPENDED",
+  "error": "The user account has been suspended, see link for details",
+  "href": "https://example.org/help/my-account-is-suspended-what-can-i-do
+}
+```
+
+Returning a dynamic page customized for this specific user:
+
+```json
+{
+  "errcode": "M_USER_ACCOUNT_SUSPENDED",
+  "error": "The user account has been suspended, see link for details",
+  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org
+}
+```
+
+
+### Potential issues
+
+See security considerations.
+
+### Alternatives
+
+We could reuse `M_USER_DEACTIVATED` and introduce an additional field:
+
+| Name | Type | Value |
+|------|------|-------|
+| `permanent` | boolean | (optional) If `false`, the account may still be reactivated. |
+
+in addition to the fields mentioned previously.
+
+### Security considerations
+
+This has the potential to expose private data.
+
+To avoid this, any `M_USER_ACCOUNT_SUSPENDED` MUST NOT be sent without authentication.
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -82,4 +82,4 @@ To avoid this, any `M_USER_ACCOUNT_SUSPENDED` MUST NOT be sent without authentic
 
 ### Unstable prefixes
 
-During testing, `M_USER_ACCOUNT_SUSPENDED` will be prefixed as `M_ORG_MATRIX_MSC3823_USER_ACCOUNT_SUSPENDED`.
+During testing, `M_USER_ACCOUNT_SUSPENDED` will be prefixed as `ORG.MATRIX.MSC3823.USER_ACCOUNT_SUSPENDED`.
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -79,3 +79,7 @@ in addition to the fields mentioned previously.
 This has the potential to expose private data.
 
 To avoid this, any `M_USER_ACCOUNT_SUSPENDED` MUST NOT be sent without authentication.
+
+### Unstable prefixes
+
+During testing, `M_USER_ACCOUNT_SUSPENDED` will be prefixed as `M_ORG_MATRIX_MSC3823_USER_ACCOUNT_SUSPENDED`.
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -1,15 +1,20 @@
 # MSC3823: Account Suspension
 
-Unlike [account locking](https://github.com/matrix-org/matrix-spec-proposals/pull/3939), suspension
+Unlike [account locking](https://spec.matrix.org/v1.12/client-server-api/#account-locking), suspension
 allows the user to have a (largely) readonly view of their account. Homeserver administrators and
 moderators may use this functionality to temporarily deactivate an account, or place conditions on
 the account's experience. Critically, like locking, account suspension is reversible, unlike the
 deactivation mechanism currently available in Matrix - a destructive, irreversible, action.
 
-This proposal introduces an error code for communicating suspension to a user, alongside some
-guidelines for how suspension could be implemented by a server. APIs to invoke or clear suspension
-are not introduced, and left as an implementation detail. These will typically be done through an
-administrator-only API.
+This proposal introduces a concept of suspension to complement locking for server admins to use. Locking
+is typically more used in narrow scenarios, where the server admin wants to prevent the account from
+engaging any further. An example of this may be too many failed password attempts. Suspension is more
+general purpose to create a barrier to further action - a "something weird is going on -> suspend"
+kind of button.
+
+The error code introduced by this proposal is accompanied by guidelines on how servers can implement
+suspension. APIs to invoke or clear suspension are not introduced, and left as an implementation detail.
+These will typically be done through an administrator-only API.
 
 ## Proposal
 
@@ -30,12 +35,17 @@ detail, however a server SHOULD permit the user to:
   and write associated [cross-signing data](https://spec.matrix.org/v1.10/client-server-api/#cross-signing).
 * [Populate their key backup](https://spec.matrix.org/v1.10/client-server-api/#server-side-key-backups).
 * Leave rooms & reject invites.
-* Redact events.
+* Redacting their own events.
 * Log out/delete any device of theirs, including the current session.
 * Deactivate their account, potentially with a deliberate time delay to discourage making a new
   account right away.
 * Change or add [admin contacts](https://spec.matrix.org/v1.10/client-server-api/#adding-account-administrative-contact-information),
-  but not remove.
+  but not remove. Servers are recommended to only permit this if they keep a changelog on contact information
+  to prevent misuse.
+
+The recommendation for users to continue receiving/reading messages is largely so the administrator
+can maintain contact with the user, where applicable. Future MSCs may improve upon the admin<>user
+communication, and account locking may also be used to prevent access to messages.
 
 The suggested set of explicitly forbidden actions is:
 
@@ -43,6 +53,7 @@ The suggested set of explicitly forbidden actions is:
 * Sending messages.
 * Sending invites.
 * Changing profile data (display name and avatar).
+* Redacting other users' events (when a moderator/admin of a room, for example).
 
 ## Potential issues
 
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -69,7 +69,10 @@ flag, though ideally each error code should provide meaning on its own.
 
 The related concept of locking, as discussed in places like [MSC3939](https://github.com/matrix-org/matrix-spec-proposals/pull/3939)
 and [matrix-org/glossary](https://github.com/matrix-org/glossary), is semantically different from
-suspension.
+suspension. Suspension has the key difference that the user is *not* logged out of their client,
+typically used by safety teams, while locking *does* log the user out and would primarily be used by
+security teams. A future MSC may combine or split the two related error codes into a descriptive set
+of capabilities the user can perform.
 
 ## Unstable prefixes
 
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -45,7 +45,7 @@ Returning a static page:
 {
   "errcode": "M_USER_ACCOUNT_SUSPENDED",
   "error": "The user account has been suspended, see link for details",
-  "href": "https://example.org/help/my-account-is-suspended-what-can-i-do
+  "href": "https://example.org/help/my-account-is-suspended-what-can-i-do"
 }
 ```
 
@@ -55,7 +55,7 @@ Returning a dynamic page customized for this specific user:
 {
   "errcode": "M_USER_ACCOUNT_SUSPENDED",
   "error": "The user account has been suspended, see link for details",
-  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org
+  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org"
 }
 ```
 
--- a/proposals/3823-code-for-account-suspension.md
+++ b/proposals/3823-code-for-account-suspension.md
@@ -1,85 +1,66 @@
-# MSC3823: A code for account suspension
+# MSC3823: Account Suspension
 
-This MSC introduces a new error code that servers may send to clarify that an account has been
-suspended *temporarily* but may still be reactivated.
+Unlike [account locking](https://github.com/matrix-org/matrix-spec-proposals/pull/3939), suspension
+allows the user to have a (largely) readonly view of their account. Homeserver administrators and
+moderators may use this functionality to temporarily deactivate an account, or place conditions on
+the account's experience. Critically, like locking, account suspension is reversible, unlike the
+deactivation mechanism currently available in Matrix - a destructive, irreversible, action.
+
+This proposal introduces an error code for communicating suspension to a user, alongside some
+guidelines for how suspension could be implemented by a server. APIs to invoke or clear suspension
+are not introduced, and left as an implementation detail. These will typically be done through an
+administrator-only API.
 
 ## Proposal
 
-### Introduction
+When an account is suspended, any [Client-Server API](https://spec.matrix.org/v1.10/client-server-api/)
+endpoint MAY return a 403 HTTP status code with `errcode` of `M_USER_SUSPENDED`. This indicates to
+the user that the associated action is unavailable.
 
-Matrix has a code `M_USER_DEACTIVATED` that a server may return to describe an account that has been
-deactivated. So far, this code has been used to represent accounts that have been *permanently*
-deactivated. In particular, clients that interpret this error code display it imply that the account
-has been *permanently* deactivated.
+Clients should note that for more general endpoints, like `/send/:eventType`, suspension MAY only be
+applied to a subset of request parameters. For example, a user may be allowed to *redact* events but
+not send messages.
 
-However, some countries (e.g. UK) have laws that require the ability to appeal account
-deactivations. This requires the ability to specify that an account is *reversibly*
-suspended and let users know about the appeals procedure.
+The specific list of permitted actions during suspension is left as a deliberate implementation
+detail, however a server SHOULD permit the user to:
 
-This MSC simply introduces a new error code `M_USER_ACCOUNT_SUSPENDED` that servers may send to
-clarify that an account has been suspended but that the solution may still be resolved either by
-an appeal or by e.g. clearing up some abusive messages.
+* Log in/create additional sessions (which should also behave as suspended).
+* See and receive messages, particularly via `/sync` and `/messages`.
+* [Verify their other devices](https://spec.matrix.org/v1.10/client-server-api/#device-verification)
+  and write associated [cross-signing data](https://spec.matrix.org/v1.10/client-server-api/#cross-signing).
+* [Populate their key backup](https://spec.matrix.org/v1.10/client-server-api/#server-side-key-backups).
+* Leave rooms & reject invites.
+* Redact events.
+* Log out/delete any device of theirs, including the current session.
+* Deactivate their account, potentially with a deliberate time delay to discourage making a new
+  account right away.
+* Change or add [admin contacts](https://spec.matrix.org/v1.10/client-server-api/#adding-account-administrative-contact-information),
+  but not remove.
 
-This MSC does *not* specify a mechanism to suspend or unsuspend the account or to handle appeals.
+The suggested set of explicitly forbidden actions is:
 
-### Proposal
+* Joining or knocking on rooms, including accepting invites.
+* Sending messages.
+* Sending invites.
+* Changing profile data (display name and avatar).
 
-Introduce a new error code `M_USER_ACCOUNT_SUSPENDED`. This error code MAY be sent by the server
-whenever a client attempts to use an API on behalf of a user whose account has been suspended.
+## Potential issues
 
-| Name | Type | Value |
-|------|------|-------|
-| `href` | string | (optional) If specified, a URL containing more information for the user, such as action needed. |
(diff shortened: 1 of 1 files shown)


proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,81 @@
+# MSC3823: A code for account suspension
```
FSG-Cat, line 1: The Rendered link is wrong btw. Just wanted to address this.

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Security considerations
+
+This has the potential to expose private data.
+
+To avoid this, any `M_USER_ACCOUNT_SUSPENDED` MUST NOT be sent without authentication.
+
+### Unstable prefixes
+
+During testing, `M_USER_ACCOUNT_SUSPENDED` will be prefixed as `M_ORG_MATRIX_MSC3823_USER_ACCOUNT_SUSPENDED`.
```
turt2live, line 85: `M_` is a reserved namespace:
```suggestion
During testing, `M_USER_ACCOUNT_SUSPENDED` will be prefixed as `ORG.MATRIX,MSC3823.USER_ACCOUNT_SUSPENDED`.
```

turt2live (dismissed):

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### Examples
+
+Returning a static page:
+
+```json
+{
+  "errcode": "M_USER_ACCOUNT_SUSPENDED",
+  "error": "The user account has been suspended, see link for details",
+  "href": "https://example.org/help/my-account-is-suspended-what-can-i-do
```
agraven, line 48: The closing quote is missing here

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+Returning a dynamic page customized for this specific user:
+
+```json
+{
+  "errcode": "M_USER_ACCOUNT_SUSPENDED",
+  "error": "The user account has been suspended, see link for details",
+  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org
```
agraven, line 58: Closing quote is also missing here

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,85 @@
+# MSC3823: A code for account suspension
```
babolivier, line 1: What should a client do with the data it has stored when seeing a `M_USER_ACCOUNT_SUSPENDED` error? Should it consider it synonymous to a logout and delete all its data? Similarly, is the expectation that all access tokens from the user have been revoked following the suspension?

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,85 @@
...
+Matrix has a code `M_USER_DEACTIVATED` that a server may return to describe an account that has been
+deactivated. So far, this code has been used to represent accounts that have been *permanently*
+deactivated. In particular, clients that interpret this error code display it imply that the account
+has been *permanently* deactivated.
+
+However, some countries (e.g. UK) have laws that require the ability to appeal account
+deactivations. This requires the ability to specify that an account is *reversibly*
+suspended and let users know about the appeals procedure.
+
+This MSC simply introduces a new error code `M_USER_ACCOUNT_SUSPENDED` that servers may send to
```
babolivier, line 19: Should the error code be `M_USER_SUSPENDED` to be consistent with `M_USER_DEACTIVATED`?

proposals/3823-code-for-account-suspension.md
MTRNord: I feel like the user should have the ability to keep their data (a logout would delete it). So maybe handle it as a softlock in the client?

As in similar how it would be when you have no Internet connection.

That way a user still can export their stuff if needed.

FSG-Cat: The concern @MTRNord has i think can actually be resolved by applying the flow for Mutes to said user under MSC3907 and its recomendations for how homeservers apply mute recomendations to their users think. The recomended way to apply a Mute as a homeserver is actually to render the user into Read only mode. The intension from my prespective is actually something arlong these lines. Let the user keep data access but Read only.

MTRNord: Imho freezing the state makes sense here. As a suspension means they were due to their actions probably removed from participating in matrix with the account. They can keep their data but should not be able to further read imho.

proposals/3823-code-for-account-suspension.md
dklimpel, line 48:
```suggestion
  "href": "https://example.org/help/my-account-is-suspended-what-can-i-do"
```

proposals/3823-code-for-account-suspension.md
dklimpel, line 58:
```suggestion
  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org"
```

dklimpel (commented):
Only typos

turt2live:
I've updated this to match a modern template, and removed `href` from the proposal text. In a future MSC, the T&S team will be proposing a way to inform the user of actions taken against their account, and an appeals system - we'll pick up `href` then, as needed.

I've also added a set of permitted/forbidden actions a user can take while suspended. This is based upon the T&S team's current research in the area as it applies to regulatory/legislation compliance.

proposals/3823-code-for-account-suspension.md
```diff
@@ -1,85 +1,66 @@
...
 
-```json
-{
-  "errcode": "M_USER_ACCOUNT_SUSPENDED",
-  "error": "The user account has been suspended, see link for details",
-  "href": "https://example.org/action-needed/please-redact-events?event-id=$event_1:example.org&event-id=$event_2:example.org"
-}
-```
+When an account is suspended, any [Client-Server API](https://spec.matrix.org/v1.10/client-server-api/)
+endpoint MAY return a 403 HTTP status code with `errcode` of `M_USER_SUSPENDED`. This indicates to
```
MTRNord, line 17: I think its crucial to communicate the reason optional to the user to allow them to either act on it or to give them a reasoning in terms of "You violated our COC". A blind errorcode seems like a bitter taste imho.

proposals/3823-code-for-account-suspension.md
```diff
@@ -1,85 +1,66 @@
...
+applied to a subset of request parameters. For example, a user may be allowed to *redact* events but
+not send messages.
 
-### Potential issues
+The specific list of permitted actions during suspension is left as a deliberate implementation
+detail, however a server SHOULD permit the user to:
 
-See security considerations.
+* Log in/create additional sessions (which should also behave as suspended).
+* See and receive messages, particularly via `/sync` and `/messages`.
```
MTRNord, line 28: Doesnt this go against the original proposal? Why should a suspended account still be allowed to read messages (Apart from possibly admin DMs)? If the user can still read all messages it feels like the suspension isnt really useful. 

Imho the suspension would be applied in 2 major cases: 

1. Spamming from an account
2. People using accounts as not allowed by COC crawlers.

However case 2 is not a thing you can do with this. Though maybe there is a reasoning behind this?

proposals/3823-code-for-account-suspension.md
```diff
@@ -1,85 +1,66 @@
...
+* [Verify their other devices](https://spec.matrix.org/v1.10/client-server-api/#device-verification)
+  and write associated [cross-signing data](https://spec.matrix.org/v1.10/client-server-api/#cross-signing).
+* [Populate their key backup](https://spec.matrix.org/v1.10/client-server-api/#server-side-key-backups).
+* Leave rooms & reject invites.
+* Redact events.
+* Log out/delete any device of theirs, including the current session.
+* Deactivate their account, potentially with a deliberate time delay to discourage making a new
+  account right away.
+* Change or add [admin contacts](https://spec.matrix.org/v1.10/client-server-api/#adding-account-administrative-contact-information),
```
MTRNord, line 37: I am not sure if its a good idea to allow a suspended user to change their contact information in a suspended state. It may be used to hide the problematic information. We currently have no auditing system in matrix to see these changes so imho I am in favor of freezing the account more rather than less.

proposals/3823-code-for-account-suspension.md
jimmackenzie: There are a few reasons why a suspended user would need access to messages (not exhaustive):
- To access admin messages about the suspension (this bit needs more work/the future MSC referenced in the updates)
- To appeal their suspension, if erroneous
- To exercise their GDPR (or equivalent) rights (including export)
- To take actions to restore their account to good standing (e.g. redacting messages that broke the code of conduct or terms of service of their homeserver)

For situations where preventing the user from reading messages makes sense, account locking via MSC3939 could be one way of approaching it.

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,66 @@
...
+The specific list of permitted actions during suspension is left as a deliberate implementation
+detail, however a server SHOULD permit the user to:
+
+* Log in/create additional sessions (which should also behave as suspended).
+* See and receive messages, particularly via `/sync` and `/messages`.
+* [Verify their other devices](https://spec.matrix.org/v1.10/client-server-api/#device-verification)
+  and write associated [cross-signing data](https://spec.matrix.org/v1.10/client-server-api/#cross-signing).
+* [Populate their key backup](https://spec.matrix.org/v1.10/client-server-api/#server-side-key-backups).
+* Leave rooms & reject invites.
+* Redact events.
```
HarHarLinks, line 33: Should this be limited to redacting only their own events and expressively NOT continuing to act on their moderator/admin role, potentially acting disruptive by redacting other's events?

proposals/3823-code-for-account-suspension.md
turt2live: sorry, looks like I forgot to clean up old threads yesterday: I think this is resolved with clearer direction in the MSC. Please open a new thread if something is still unclear.

proposals/3823-code-for-account-suspension.md
turt2live: (yes)

proposals/3823-code-for-account-suspension.md
turt2live: Errors have an optional `error` for human-readable description of what went wrong. A reason for suspension can be included here, though clients may not show it. 

The potential issues section covers this thread in a bit more detail - if there's still a concern about the approach, please open a thread there.

proposals/3823-code-for-account-suspension.md
turt2live: (what Jim said 😇)

This may need bringing into the MSC itself, but for now leaving this open for reader context.

proposals/3823-code-for-account-suspension.md
turt2live: Oh good call, yes, definitely.

proposals/3823-code-for-account-suspension.md
turt2live: I'm not sure it'd be appropriate to have such an auditing system built into the protocol, but we can rely on implementation details here. Possibly we should clarify that only servers which support contact change history should allow this action.

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,66 @@
+# MSC3823: Account Suspension
+
+Unlike [account locking](https://github.com/matrix-org/matrix-spec-proposals/pull/3939), suspension
+allows the user to have a (largely) readonly view of their account. Homeserver administrators and
+moderators may use this functionality to temporarily deactivate an account, or place conditions on
+the account's experience. Critically, like locking, account suspension is reversible, unlike the
+deactivation mechanism currently available in Matrix - a destructive, irreversible, action.
```
clokep, line 7: I discussed this with @turt2live in #matrix-spec, but this paragraph made me think this is an alternative to MSC3939, but is meant to complement it so that you end up with a kind of flow between them:

Active -> Suspended -> Locked --> Deactive

It is a bit unclear to me if "Suspended" is meant to be closed to deactivated than locked or not though?

jimmackenzie: I see them as parallels, used in different situations. Locking has narrower applications (one example would be for account compromise scenarios, where the homeserver security team locks access to the user's account, to prevent further intrusion and to enable out-of-band recovery). Suspension is the more broadly useful state, where the user account is restricted, but with access preserved.

For most servers, I'd expect a flow of `active -> suspended -> deactivated` to be the standard flow for dealing with accounts that violate the terms of service. Locking would rarely feature, but would be available as a more restrictive option where appropriate.

We should clarify that this complements MSC3939, rather than competing with it!

jimmackenzie: Yep. There's definitely cases where it would be useful to go from Locked -> Deactivated or from Suspended -> Locked, but the flowchart captures most of the uses.

turt2live: The example I was largely thinking of is Active -> Suspended due to suspicious activity (volumetric spam, etc), then Suspended -> Locked if the user/client keeps trying anyways. This would largely only apply if the server had material load on it caused by the suspended account, and the operations team didn't want to fully deactivate the account in response.

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,66 @@
...
+* Log out/delete any device of theirs, including the current session.
+* Deactivate their account, potentially with a deliberate time delay to discourage making a new
+  account right away.
+* Change or add [admin contacts](https://spec.matrix.org/v1.10/client-server-api/#adding-account-administrative-contact-information),
+  but not remove.
+
+The suggested set of explicitly forbidden actions is:
+
+* Joining or knocking on rooms, including accepting invites.
+* Sending messages.
```
Kladki, line 43: There could be value in suggesting that this could be restricted to a set of explicitly allowed rooms, e.g. if there is an appeals room where suspended users can apologize, explain themselves, etc.

turt2live: This feels like something the appeals MSC would cover, if it decides to use a room.

Kladki: Ah my bad, I missed line 51-52

proposals/3823-code-for-account-suspension.md
turt2live, line 1: Implementation requirements:

* [x] Server - https://github.com/element-hq/synapse/pull/17051
* [ ] Client

turt2live: The client implementation requirement has been questioned a bit - I think for this MSC, evidence of clients not exploding would be fine. The list won't be exhaustive or fully representative of the ecosystem, but would indicate the typical user experience.

Specifically, I think the following clients need to be tested with a suspended account:
* [ ] Element Web
* [ ] Element X (iOS or Android, or both)
* [ ] Cinny
* [ ] FluffyChat
* [ ] Nheko

"Exploding" would be defined as the user being logged out, or otherwise unable to interact with their account. Failing to show the error message ("your account is suspended") is fine, but bonus points to the client if it makes it clear to the user. If the user can (minimally) interact with their rooms without being kicked to an error screen or login page, the client "works", even if clunky.

turt2live: This testing has been completed by the T&S team at the Foundation. Results are compiled here: https://docs.google.com/document/d/1oTVKRLHCc6x4iYti8bkhRN3YbBO29oAjqmZE-mj-xXY/edit?usp=sharing

TLDR: The above-listed clients all passed, but a few had some weird-ish errors/rough edges. With my SCT hat, I don't think these rough edges are blockers for merging.

proposals/3823-code-for-account-suspension.md
turt2live: I've attempted to add the context here to the MSC. Please open a new comment if things need further refinement.

proposals/3823-code-for-account-suspension.md
turt2live: (added - please open a new thread if further clarification is needed)

turt2live:
I think this is ready to go.

@mscbot fcp merge

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,77 @@
...
+their account and allow the user to appeal that action.
+
+## Alternatives
+
+No significant alternatives are plausible. `M_USER_DEACTIVATED` could be expanded with a `permanent`
+flag, though ideally each error code should provide meaning on its own.
+
+The related concept of locking, as discussed in places like [MSC3939](https://github.com/matrix-org/matrix-spec-proposals/pull/3939)
+and [matrix-org/glossary](https://github.com/matrix-org/glossary), is semantically different from
+suspension.
```
clokep, line 72: I think this is still very confusing -- as a server operator I would constantly be unsure whether to lock or suspend an account. It is unclear when they would be used. From the glossary (which I don't know if it is anything official or not?) it says:

> Account locking: User is unable to log in to their account, and it is preserved in-place. This action is reversible.
> Account suspension: Similar to locking, though the user can still log in and read their account. Writes are not possible. This action is reversible.

Which sounds to me like the only difference is users being able to login or not. I don't understand the differentiation.

jimmackenzie: An example for each might help: 
- As a server admin, you detect unusual login activity for a user. While you investigate, you lock the account.
- As a server admin, you receive a number of reports about a user posting content that is against your Terms of Service. You suspend the user - this could be until they remove the offending content, until they successfully appeal your decision, or until a set period of time elapses, whatever you decide/your Terms of Service state.

Locking is for scenarios where preventing any activity is important. It's a narrower use case, but there is overlap where you might reach for either locking or suspension.

clokep: I've read this a bunch of times and can't really figure out the difference from either an operator or user point of view.

@mscbot concern Confusion/overlap with account locking is unclear.

turt2live: (did Jim's comment help, or is it still confusing? I'm struggling to figure out how to add clarity here, honestly)

clokep: It didn't help clarify. It mostly felt like synonyms were being used. E.g. Why not just lock the account for a period of time or during investigation?

jimmackenzie: You could choose to use locking. It would be a valid option! 

The semantic difference is whether the user should have any access to their account while locked (no) or suspended (yes, but limited). 

With only locking, you couldn't offer the user the option to remove offending content, as they wouldn't be able to perform the redaction. With suspension, they could. 

Suspension also opens up opportunities for improved flows for handling abuse reporting, follow-up conversations and appeals within Matrix, rather than out-of-band. 

Would a different presentation of what a suspended user can do compared to a locked user help to clarify? Or perhaps some examples of where you might prefer to use locking or suspension?

KitsuneRal: I agree with @clokep that having two separate things for what looks to me like two different flavours of locked down user is somewhat confusing. Rather than folding suspension into deactivation, I would consider an alternative of adding flexibility to the locking function - I don't see a security problem with relaxing the user locking concept from the current "can't login" semantics but I might miss something?

Separately, I feel there's a bit of contradiction in saying "the account becomes read-only, can't send messages in particular" while at the same time trying to maintain that conversation with admins is possible within Matrix. I understand there's an intention to have a separate MSC for appeals or other admin-user communication but I feel like without that additional functionality the whole functionality is a bit hollow.

As a client author, I would be more interested in having a single `M_USER_LOCKED` status with, maybe but not necessarily, some information on the level of lockdown (e.g.: can't do anything; can't login until the password is changed; can login and redact but nothing else; the above and DCs with a subset of users). Having additional error codes looks to me more like we're encroaching into the homeserver administration API realm, which we historically tried to stay away from.

jimmackenzie: I disagree - a single `M_USER_LOCKED` status that I need to parse to determine what the user is permitted to do is more complex and confusing than distinct status codes that communicate the permissions directly.

- `M_USER_LOCKED` - the account is locked. They can't access it. They need to resolve that out of band. Primarily useful for security teams.
- `M_USER_SUSPENDED` - the account is suspended. Some permissions have been suspended. Primarily useful for Safety teams.

Separate codes also makes it easier for clients to tailor the messaging to the user about what has happened to their account, and why, and how to go about resolving it. Some hypothetical examples of how those messages could look:

> Your account is locked. Contact your system administrator by email at support@domain.tld

> Your account is suspended. Review the message from @safety:server.tld in your account for next steps. You cannot send messages while your account is suspended.

The Matrix.org safety team would like to use suspension now. How do we get there?

If the suggestion is that this is best suited to homeserver administration APIs, and shouldn't be part of the spec, we can work with that. My preference would be for the spec to include the fundamentals for effective safety tooling, so I'd prefer to find a way to resolve these concerns and get this merged into the spec.

clokep: > `M_USER_SUSPENDED` - the account is suspended. Some permissions have been suspended. Primarily useful for Safety teams.

I wonder if calling this "limited" instead of "suspended" would be a potential resolution?

clokep: > The semantic difference is whether the user should have any access to their account while locked (no) or suspended (yes, but limited).

This didn't quite jive with the definitions I pulled out above from the linked glossary:

> Account locking: User is unable to log in to their account, and it is preserved in-place. This action is reversible.
> Account suspension: Similar to locking, though the user can still log in and read their account. Writes are not possible. This action is reversible.

So I double checked the spec to make sure I wasn't misunderstanding, which states that locking:

> When an account is locked, servers MUST return a 401 Unauthorized error response with an M_USER_LOCKED error code and [soft_logout](https://spec.matrix.org/v1.12/client-server-api/#soft-logout) set to true on all but the following Client-Server APIs (`/logout`).
>
> Servers SHOULD NOT invalidate access tokens on locked accounts unless the client requests a logout (using the above endpoints). This ensures that users can retain their sessions without having to log back in if the account becomes unlocked.

I think the glossary is overly simplistic -- it isn't just that users cannot *login* it is also that their current sessions no longer work, but aren't destroyed.

jimmackenzie: I'm not a fan of calling a user limited, which is what `M_USER_LIMITED` does. It's an unfortunate word association. If we had a pattern of `M_ACCOUNT_<state>` it would likely help to avoid that unfortunate interplay, but we don't.

That would also disqualify `M_USER_DISABLED` but not `M_USER_BANNED`.

In the realm of safety tooling, the general consensus is to use suspended, although there are some other examples:
Twitter and Discord use suspended, Whatsapp uses banned, Facebook uses suspended, Instagram uses disabled. Gmail/Google uses suspended. Github uses suspended.

Using `M_USER_BANNED` risks confusion with room level bans, while this is an account state. I think that leaves `M_USER_SUSPENDED` as the best of the choices within the pattern we have already.

How else could we provide guidance to folks about when this state is useful, compared to locking or deactivation?

clokep: > I'm not a fan of calling a user limited, which is what `M_USER_LIMITED` does.

I don't see any hits for `M_USER_LIMITED` in the spec?

jimmackenzie: Ah! I was taking your suggestion (and tulir's +1) to use 'limited' instead of 'suspended' to the conclusion of that thought, and pointing out the pitfalls there. Sorry for the confusion!

proposals/3823-code-for-account-suspension.md
```diff
@@ -0,0 +1,77 @@
...
+
+The error code introduced by this proposal is accompanied by guidelines on how servers can implement
+suspension. APIs to invoke or clear suspension are not introduced, and left as an implementation detail.
+These will typically be done through an administrator-only API.
+
+## Proposal
+
+When an account is suspended, any [Client-Server API](https://spec.matrix.org/v1.10/client-server-api/)
+endpoint MAY return a 403 HTTP status code with `errcode` of `M_USER_SUSPENDED`. This indicates to
+the user that the associated action is unavailable.
```
richvdh, line 23: someone remind me why this should be a 403 rather than a 401 (and why does locking specify a 401)?

proposals/3823-code-for-account-suspension.md
richvdh: Personally, I'm pretty relaxed about having both concepts. If this is a useful tool for T&S, I'm prepared to believe that (and I am minded to agree with Jim that a clear distinction from locking is nicer to work with than defining locking to have shades of grey.

I wouldn't mind some more discussion within the proposal doc about how the two methods differ in functionality (and what the user is expected to see in their client), and how the anticipated usecases differ, but I don't consider that a blocker.

clokep: > Ah! I was taking your suggestion (and tulir's +1) to use 'limited' instead of 'suspended' to the conclusion of that thought, and pointing out the pitfalls there. Sorry for the confusion!

That was my suggestion. I'm not following your concerns about why `M_USER_LIMITED` would be an issue.

richvdh: > I'm not following your concerns about why `M_USER_LIMITED` would be an issue.

I wouldn't mind a clarification from Jim on this point, but I suspect "user limited" could be read as an ableist term?

jimmackenzie: Exactly that. It's an unfortunate combination we should avoid.

clokep: Seems like using `M_USER_ACCESS_LIMITED` or something like that would be ok?

turt2live: Given the widespread use of "suspended", I'm heavily inclined to follow that precedent. 

I'll add some text to the proposal to clarify locking vs suspension in any case.

proposals/3823-code-for-account-suspension.md
turt2live: 401 is "token doesn't work anymore" while 403 is "your token works, but server says no".

proposals/3823-code-for-account-suspension.md
KitsuneRal: To clarify: I'd really prefer this to be future-proof, i.e. if we need to introduce additional levels of access (either for safety or security purposes) or if different homeservers end up providing different levels of access to suspended users, the solution should be able to incorporate it.

turt2live: Improving communication to the user on what they can/can't do is planned for a future MSC as part of a regulatory compliance series relating to T&S. I'd encourage us to think about sub-codes or similar at that stage rather than delaying a needed feature.

KitsuneRal (approved):

clokep:
@mscbot resolve Confusion/overlap with account locking is unclear.

Agreed with @tulir -- I don't feel strongly enough to block this but I think it is confusing. If suspension is the industry term then I guess it makes sense.

turt2live:
Spec PR, for when FCP finishes: https://github.com/matrix-org/matrix-spec/pull/2014

turt2live:
Merged 🎉

PR #3826, opened 2022-05-27
Minor clarifications to MSC2675

turt2live:
To improve understanding for purposes of spec writing.


erikjohnston (approved):

clokep (approved):
This seems to have the same meaning as the original intention. 👍

PR #3827, opened 2022-05-27
MSC3827: Filtering of `/publicRooms` by room type

SimonBrandner:
Fixes https://github.com/vector-im/element-meta/issues/338

<hr>

#### `synapse`

+ https://github.com/matrix-org/synapse/pull/13031

#### `js-sdk`

+ https://github.com/matrix-org/matrix-js-sdk/pull/2469

#### `react-sdk`

+ https://github.com/matrix-org/matrix-react-sdk/pull/8866
+ https://github.com/matrix-org/matrix-react-sdk/pull/8936

<hr>

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/SimonBrandner/msc/space-explore/proposals/3827-space-explore.md)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3827#issuecomment-1170150880)

### DIFF
--- a/proposals/3827-space-explore.md
+++ b/proposals/3827-space-explore.md
@@ -25,7 +25,7 @@ POST /_matrix/client/v3/publicRooms
 }
 ```
 
-The response is also modified to include a `room_type` field, so that clients can show
+The response for both `POST /publicRooms` and `GET /publicRooms` is also modified to include a `room_type` field, so that clients can show
 the room type metadata.
 
 ```json
--- a/proposals/3827-space-explore.md
+++ b/proposals/3827-space-explore.md
@@ -11,7 +11,7 @@ issue, this MSC proposes a mechanism for filtering the response of
 
 This MSC proposes adding a `room_types` filter to the `POST /publicRooms` endpoint.
 The value of `room_types` is an array of room types - possible values of the
-`type` field in the `m.room.create` state event - which should be returned.
+`type` field in the `m.room.create` state event's `content` dictionary - which should be returned.
 
 An example request could look like this:
 
--- a/proposals/3827-space-explore.md
+++ b/proposals/3827-space-explore.md
@@ -9,8 +9,8 @@ issue, this MSC proposes a mechanism for filtering the response of
 
 ## Proposal
 
-This MSC proposes adding a `room_type` filter to the `/publicRooms` endpoint.
-The value of `room_type` is an array of room types - possible values of the
+This MSC proposes adding a `room_types` filter to the `/publicRooms` endpoint.
+The value of `room_types` is an array of room types - possible values of the
 `type` field in the `m.room.create` state event - which should be returned.
 
 An example request could look like this:
@@ -19,7 +19,7 @@ An example request could look like this:
 POST /_matrix/client/v3/publicRooms
 {
     "filter": {
-        "room_type": ["m.space"]
+        "room_types": ["m.space"]
     },
     "limit": 10
 }
@@ -45,13 +45,12 @@ the room type metadata.
     "prev_batch": "p1902",
     "total_room_count_estimate": 115
 }
+```
 
 If the client wants to get rooms of the default type, it should include `null` in
-the `room_type` array. If the `room_type` filter is not specified, the server should
-return only the rooms of the default type. If the `room_type` array is empty, the
+the `room_types` array. If the `room_types` filter is not specified, the server should
+return only the rooms of the default type. If the `room_types` array is empty, the
 server should return _all_ rooms no matter the type.
-
-```
 
 ## Alternatives
 
@@ -73,6 +72,7 @@ client and the server side.
 While this MSC is not considered stable, implementations should use
 `org.matrix.msc3827` as a namespace.
 
-|Release    |Development                   |
-|-----------|------------------------------|
-|`room_type`|`org.matrix.msc3827.room_type`|
+|Release     |Development                    |
+|------------|-------------------------------|
+|`room_type` |`org.matrix.msc3827.room_type` |
+|`room_types`|`org.matrix.msc3827.room_types`|
--- a/proposals/3827-space-explore.md
+++ b/proposals/3827-space-explore.md
@@ -46,6 +46,11 @@ the room type metadata.
     "total_room_count_estimate": 115
 }
 
+If the client wants to get rooms of the default type, it should include `null` in
+the `room_type` array. If the `room_type` filter is not specified, the server should
+return only the rooms of the default type. If the `room_type` array is empty, the
+server should return _all_ rooms no matter the type.
+
 ```
 
 ## Alternatives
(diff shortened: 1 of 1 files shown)


proposals/3827-space-explore.md
```diff
@@ -0,0 +1,73 @@
...
+to have [a way to to discover
+them](https://github.com/vector-im/element-web/issues/17264). To solve this
+issue, this MSC proposes a mechanism for filtering the response of
+`/publicRooms` by room type.
+
+## Proposal
+
+This MSC proposes adding a `room_type` filter to the `/publicRooms` endpoint.
+The value of `room_type` is an array of room types - possible values of the
+`type` field in the `m.room.create` state event - which should be returned.
```
clokep, line 14: Does this make it possible to filter for non-spaces, e.g. rooms without a `room_type`? I guess the following could do this:

```json
{
    "filter": {
        "room_type": []
    }
}
```

But this precludes searching for spaces and "normal" rooms while ignoring other types of rooms?

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,73 @@
...
+{
+    "filter": {
+        "room_type": ["m.space"]
+    },
+    "limit": 10
+}
+```
+
+the response then also includes a `room_type` field, so that clients can show
+the room type metadata.
```
clokep, line 29: 🎉  This brings it closer in-line to `/hierarchy`!

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,73 @@
...
+POST /_matrix/client/v3/publicRooms
+{
+    "filter": {
+        "room_type": ["m.space"]
+    },
+    "limit": 10
+}
+```
+
+the response then also includes a `room_type` field, so that clients can show
```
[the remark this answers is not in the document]
clokep: I assumed the intention was to always show it, but it doesn't quite say that.

```suggestion
The response is also modified to include a `room_type` field, so that clients can show
```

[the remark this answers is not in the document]
SimonBrandner: Would `servers should now also include a `room_type` field in the response, so that clients can show the room type metadata.` sound more clear?

proposals/3827-space-explore.md
SimonBrandner: Hmm, the ideal would be adding a default room type (`m.room`) - is that planned, has it ever been discussed, would it ok using `m.room` just in this MSC?

proposals/3827-space-explore.md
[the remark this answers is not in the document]
SimonBrandner: > I assumed the intention was to always show it, but it doesn't quite say that.

This is probably the clearest :+1:

proposals/3827-space-explore.md
clokep: Another option could be to use a special `null` to mean "missing"? Not pretty but eh.

SimonBrandner: > Another option could be to use a special `null` to mean "missing"? Not pretty but eh.

Yeah, that's not exactly the nicest but would work...

turt2live:
I'm also comfortable with the level of implementation here:

@mscbot fcp merge

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,78 @@
...
+[MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772) defines Spaces as
+rooms with `type` of `m.space`. One of the most requested features for Spaces is
+to have [a way to to discover
+them](https://github.com/vector-im/element-web/issues/17264). To solve this
+issue, this MSC proposes a mechanism for filtering the response of
+`/publicRooms` by room type.
+
+## Proposal
+
+This MSC proposes adding a `room_types` filter to the `/publicRooms` endpoint.
```
uhoreg, line 12: Nitpick: since we have both `POST` and `GET` versions of the endpoint, it should specify which one it applies to

```suggestion
This MSC proposes adding a `room_types` filter to the `POST /publicRooms` endpoint.
```

(presumably, this is only being added to the `POST` version, since the `GET` version doesn't have a filter).

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,78 @@
...
+POST /_matrix/client/v3/publicRooms
+{
+    "filter": {
+        "room_types": ["m.space"]
+    },
+    "limit": 10
+}
+```
+
+The response is also modified to include a `room_type` field, so that clients can show
```
uhoreg, line 28: Again, we should specify whether the change applies to both `POST` and `GET`, or to only one.  Probably makes sense to change both (and I think that's what the synapse impl does), so

```suggestion
The response for both `POST /publicRooms` and `GET /publicRooms` is also modified to include a `room_type` field, so that clients can show
```

clokep (approved):

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,78 @@
...
+to have [a way to to discover
+them](https://github.com/vector-im/element-web/issues/17264). To solve this
+issue, this MSC proposes a mechanism for filtering the response of
+`/publicRooms` by room type.
+
+## Proposal
+
+This MSC proposes adding a `room_types` filter to the `POST /publicRooms` endpoint.
+The value of `room_types` is an array of room types - possible values of the
+`type` field in the `m.room.create` state event - which should be returned.
```
anoadragon453, line 14:
```suggestion
`type` field in the `m.room.create` state event's `content` dictionary - which should be returned.
```

anoadragon453 (approved):
I see this as a stop-gap solution to the "space discovery" problem until searching the contents of spaces becomes more efficient in implementation.

Outside of that use case, being able to filter `/publicRooms` by room type could be handy in a number of scenarios, even those outside of Spaces. So I think this is still a worthwhile change.

proposals/3827-space-explore.md
```diff
@@ -0,0 +1,78 @@
...
+[MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772) defines Spaces as
+rooms with `type` of `m.space`. One of the most requested features for Spaces is
+to have [a way to to discover
+them](https://github.com/vector-im/element-web/issues/17264). To solve this
+issue, this MSC proposes a mechanism for filtering the response of
+`/publicRooms` by room type.
+
+## Proposal
+
+This MSC proposes adding a `room_types` filter to the `POST /publicRooms` endpoint.
```
richvdh, line 12: please link to the original spec, to help readers of the MSC.

```suggestion
This MSC proposes adding a `room_types` filter to the [`POST /publicRooms`](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3publicrooms) endpoint.
```

richvdh (approved):

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1199

turt2live:
Merged 🎉

Karam19925 (commented):
ch

hillbilly3281 (approved):

PR #3831, opened 2022-06-07
Let call member events expire

robintown:
This is to combat stale device data so clients can show an accurate view of who's actively participating in the call at all times. See https://github.com/matrix-org/matrix-react-sdk/pull/8776 for a similar PoC implementation of this for Element Web's video rooms.

### DIFF
--- a/proposals/3401-group-voip.md
+++ b/proposals/3401-group-voip.md
@@ -101,9 +101,9 @@ We mandate at most one call per room at any given point to avoid UX nightmares -
 
 Users who want to participate in the call declare this by publishing a `m.call.member` state event using their matrix ID as the state key (thus ensuring other users cannot edit it).  The event contains a timestamp named `m.expires_ts` describing when this data should be considered stale, and an array `m.calls` of objects describing which calls the user is participating in within that room.  This array must contain one item (for now).
 
-When sending an `m.call.member` event, clients should set the `m.expires_ts` value to one hour in the future. If the user stays connected for longer than an hour, the client must actively update the state event with a new expiration timestamp (for example, every 45 minutes to allow for some amount of clock desync between participants).
+When sending an `m.call.member` event, clients should choose a reasonable value for `m.expires_ts` in case they go offline unexpectedly. If the user stays connected for longer than this time, the client must actively update the state event with a new expiration timestamp.
 
-`m.call.member` state events must be ignored if the `m.expires_ts` field indicates they have expired, or if their user's `m.room.member` event's membership field is not `join`. They must also be ignored if the `m.expires_ts` value is more than 75 minutes in the future, to prevent uncooperative clients from getting themselves permanently shown as 'connected' by setting the expiration date to way in the future.
+`m.call.member` state events must be ignored if the `m.expires_ts` field indicates they have expired, or if their user's `m.room.member` event's membership field is not `join`.
 
 The fields within the item in the `m.calls` contents are:
 
--- a/proposals/3401-group-voip.md
+++ b/proposals/3401-group-voip.md
@@ -99,9 +99,11 @@ We mandate at most one call per room at any given point to avoid UX nightmares -
 
 ### Call participation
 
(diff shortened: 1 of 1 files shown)


proposals/3401-group-voip.md
```diff
@@ -99,9 +99,11 @@ We mandate at most one call per room at any given point to avoid UX nightmares -
...
+Users who want to participate in the call declare this by publishing a `m.call.member` state event using their matrix ID as the state key (thus ensuring other users cannot edit it).  The event contains a timestamp named `m.expires_ts` describing when this data should be considered stale, and an array `m.calls` of objects describing which calls the user is participating in within that room.  This array must contain one item (for now).
 
-`m.call.member` state events must be ignored if their user's `m.room.member` event's membership field is not `join`.
+When sending an `m.call.member` event, clients should set the `m.expires_ts` value to one hour in the future. If the user stays connected for longer than an hour, the client must actively update the state event with a new expiration timestamp (for example, every 45 minutes to allow for some amount of clock desync between participants).
```
dbkr, line 104: I probably wouldn't specify that it has to be an hour: leaving it up to clients should be fine and allows for more flexibility, although we can suggest it as a sensible default (If it's always an hour we could just use origin_server_ts + 1 hour anyway). I would probably also use this value as the actual time after which the event is invalid. We can still leave it up to clients when to refresh. eg. SIP registrations time out after the expiry time but are generally refreshed once half that time has elapsed.

proposals/3401-group-voip.md
```diff
@@ -99,9 +99,11 @@ We mandate at most one call per room at any given point to avoid UX nightmares -
...
+
+`m.call.member` state events must be ignored if the `m.expires_ts` field indicates they have expired, or if their user's `m.room.member` event's membership field is not `join`. They must also be ignored if the `m.expires_ts` value is more than 75 minutes in the future, to prevent uncooperative clients from getting themselves permanently shown as 'connected' by setting the expiration date to way in the future.
```
dbkr, line 106: Oh, also: worth noting that call invites have a 'lifetime' field which is a duration for which it's valid after the "time of the event", the idea being that you can work that out by using the `age` field you get from the homeserver and then it doesn't rely on your clients' clocks being synced. This approach is much simpler and avoids using the (slightly contentious and difficult to use) `age` field but obviously relies on time being correct (ish) on clients. I think this is probably the right approach because 1) almost all user devices automatically NTP sync nowadays and 2) we're dealing with a much longer span of time than the amount of time a call rings for, so if a clock is off by a couple of minutes, it won't make much difference here (whereas for a call invite, it could mean the call would never ring at all).

In other words, I think this is correct, but could be worth noting in case we come back in several years and wonder why event expiry is handled differently between invites & group call member events.

proposals/3401-group-voip.md
robintown: > (If it's always an hour we could just use origin_server_ts + 1 hour anyway)

The issue with that is, if you try to set the same state event content twice to say "I'm still participating in the call", Synapse will see that it's the same thing and won't actually generate a new state event. So semantically, the timestamp has to be part of the state event's content.

My motivation for specifying exactly an hour is to make this clause work:

> They must also be ignored if the `m.expires_ts` value is more than 75 minutes in the future, to prevent uncooperative clients from getting themselves permanently shown as 'connected' by setting the expiration date to way in the future.

So if we let clients pick their own timeouts, I think we should still specify some kind of upper limit on them. Perhaps 24 hours?

dbkr: We chatted about this in standup but just so it's here too: I'm not sure there's much use in trying to prevent clients from setting stupidly long timeouts - it adds complexity to other clients to ignore them if they're too far in the future, so I'd probably say just let them. I don't think there's any particular advantage to gaming the system by setting an expiration for the end of time.

dbkr:
I'd say this is OK to go ahead & merge into Matthew's branch

PR #3832, opened 2022-06-10
Update MSC2175 "removed auth rule" note

turt2live:
We use the markdown version for spec writing, so important to get this detail in there.


richvdh (approved):
thanks.

PR #3834, opened 2022-06-14
MSC: Opportunistic user key pinning (TOFU)

duxovni:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/fayed/tofu/proposals/XXXX-tofu.md)

### DIFF
--- a/proposals/XXXX-tofu.md
+++ b/proposals/XXXX-tofu.md
@@ -1,0 +1,134 @@
+# MSCXXXX: Opportunistic user key pinning (TOFU)
+
+In the Matrix spec at present, there are no real provisions for
+properly authenticating devices belonging to other users, aside from
+the fairly rare cases where users actually take the time to verify
+each other out-of-band. Most clients encourage users to cross-sign
+their own devices, but this has little practical value for most users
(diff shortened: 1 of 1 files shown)


maralorn:
This is great! Thank you for pushing TOFU. I am though not sure that it’s acceptable UI to have this enabled by default. I suggest that clients targeting a large user base should only activate the warnings if the user has somehow demonstrated that they understand key signing, by either manually verifying a key or enabling an option in the settings. One could even consider a setting where users manually need to activate TOFU per user.

I think showing users who don‘t understand E2E warnings they don‘t understand (like Signal does) desensitizes them for security warnings and is bad UX.

I have already voiced my opinion in https://github.com/vector-im/element-web/issues/12719 (which I know see you self assigned, so I will stop here.)

Thanks for taking this on!!!

richvdh:
Since I have to look it up each time I come across it: the difference between this and what we're doing today with the TOFU trusting of the MSK is that it provides a way to synchronise TOFU trust between devices. I'm unsure that adding another cross-signing key is the best way to achieve that, however.

PR #3837, opened 2022-06-23
Cascading profile tags for push rules

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/cascading-profile-tags/proposals/XXXX-cascading-profile-tags-for-push-rules.md)


proposals/3837-cascading-profile-tags-for-push-rules.md
```diff
@@ -0,0 +1,104 @@
+# MSC3837: Cascading profile tags for push rules
+
+The push rule system includes support for per-device rule sets via
```
turt2live, line 3: we also want to remove the per-device push rules stuff, fwiw:
* https://github.com/matrix-org/matrix-spec/issues/637
* https://github.com/matrix-org/matrix-spec/issues/651

PR #2787, opened 2020-09-23
MSC2787: Portable Identities

neilalexander:

PR #1768, opened 2019-01-03
Proposal to authenticate with public keys

friedger:
[rendered](https://github.com/friedger/matrix-doc/blob/auth_public_key/proposals/1768-auth-proof.md)

PR #3839, opened 2022-06-26
MSMSC3839: primary-identity-as-key

zander:
[Rendered](https://github.com/zander/matrix-spec-proposals/blob/pubkey/proposals/3839-primary-identity-as-key.md)

### DIFF
--- a/proposals/3839-primary-identity-as-key.md
+++ b/proposals/3839-primary-identity-as-key.md
@@ -1,0 +1,120 @@
+# MSC3839: primary identity as public / private key-pair
+
+In Matrix in general and most homeservers in particular, users are
+registered and authenticated using a username and password.
+Additionally, tokens are used and sessions are registered as separate entities.  
+Encryption keys are quite often yet another datafile stored on a homeserver
+encrypted using yet another password.
+
+Users that need to backup or migrate their data are finding this
+challenging at best.
+
+*Note: The initial author is absolutely not a Matrix protocol expert, the
+proposal here is likely to lack the historical or even technical
+understanding of Matrix.  
+The author asks those better known with Matrix to help out improve this
+idea into a fully usable upgrade for the Matrix ecosystem.*
+
+Password based authentication is also getting more frowned upon in the
+wider tech industry, and for good reason. The amount of password leaks has
+increased dramatically in the last years with billions of username/password
+pairs being available on the darknet.
+The suggestion that people come up with better passwords or stop reusing
+them is simply trying to fix the wrong thing. Passwords are on its way out
+because many in the industry are realizing that you can't change human
+behavior.
+
+Passwords are also a home-server operator nightmare. The current Matrix
+design makes the password plain text available to the server and thus the
+server operator. Next to that, the one solution to allow a password reset
+is to add an email address. But that instantly gets you into the territory
+of privacy laws and those issues makes the homeserver a target for hacking.
+Paired with this is the issue that plenty of accounts on public homeservers
+simply are abandoned because their owner never wrote down the password or
+username and lost access.
+
+An established cryptographers concept called pubic/private key technology
+can be used to provide a much better user experience and at the same time
+increase the security all round.
+
+Wikipedia; https://en.wikipedia.org/wiki/Public-key\_cryptography
+
+## Proposal
+
+A user identity can be coupled to a cryptographic public key where the
+users device / app holds the private key. Doing a backup simply saves the
+private key in a file or prints it in one QR code.
+
+This allows a chat-client to authenticate the user to a homeserver without
+username or password for more security and a better user experience.
+Naturally a homeserver may choose to support web-app users and keep the
+password design for those that don't want to have an actual app, this
+proposal doesn't demand a removal of the old login but does aim to replace
+all non-webbrowser based chat-identities with public keys.
+
+The basic idea is that a user-device generates a random private key on
+first account creation and derives a public key from that which will be the
+main identification of the user from then on. Homeservers will be able to
+store the public key instead of a username/password pair for identity.
+Login can be implemented by the homeserver sending a random token (with a
+timestamp to avoid replay attacks). The client simply signs it using the
+private key and returns it to the homeserver which can validate the
+signature using the public key that is the users-ID.
+
+A chat application that creates or receives messages in an end-to-end
+encrypted chat will also create local keys (possibly tied to that session).
+Such a client can use the private-key-as-identity in order to encrypt the
+keys-used-for-end2end encryption. Uploading those to the users homeserver
+is then safe and hassle free. Same with cross-signing.
+
+Matrix has a lot of security features and for the average user they can be
+overwhelming. The goal of making Matrix clients at least as easy as the
+competition will be much easier to reach when username and passwords are
+removed. Some of the really big chat apps do not have them either.
+At the same time the option stays open to be as strict on security as the
+user deems needed. The reader may have worked with ssh and knows they can
+put a password on their private key, the same can be added in clients.
+Again, should the user feel this is needed.
+
+A login feature that competitors have, where the main client scans a QR
+from a webbrowser to login the web-app can also be easily implemented
+for users using a public key as identity. Even with minimal homeserver
+support. A client that gets a challenge can simply show this in a QR
+for the user to scan and sign in her main chat-app, to send that answer to
+the server allowing the webapp to authenticate their session without
+password.
+Some way to resolve the public key from the related account name is needed
+on the server side.
+
+## Potential issues
+
+The relation between a private key and a user-session is expected to stay
+unchanged. A client requests a token and has a session.  
+But having a password change is no longer something that happens on the
+server, it happens completely local on the client. Many users will simply
+have an empty password since the private key is protected from leakage by
+their OS.
+So, when a user resets a password this has no effect on sessions unless the
+user explicitly goes and invalidates those sessions.
+
+As such the risk increases where a user that somehow shared their private
+key which had no password will not be able to protect their account from
+others taking over.
+
+In my opinion this is an acceptable risk. The attack model doesn't really
+change and phone-local (or laptop local) protections are the same as any
+other sensitive data stored on that device.
+
+## Alternatives
+
+None yet.
+
+## Security considerations
+
+Public key crypto is well established and there are many different options
+to choose from. The main consideration is likely going to be what
+already-in-use libraries provide.
+
+## Dependencies
+
+None


uhoreg:
I'm marking this as a draft since it needs more details before it can be considered.  For example, what encryption algorithms are used, the format for the QR code, the shape of the API endpoints, etc.

izN8nu6RyeneG5XnBoBgyRMVGH6H43WF:
@uhoreg  Is this a duplicate of #1768?

zander:
@izN8nu6RyeneG5XnBoBgyRMVGH6H43WF 

the referred proposal is quite different and in my biased opinion not very useful.

It does not propose to have the identity to be a key, for starters. Which is the main point of this proposal.

Indeed, the secondary effect of such an ID is that you can login with public-key-cryptography. An obvious enough idea as this has been done for many years in the industry as you found there is another proposal that bolts this powerful construct onto matrix.

But to do it elegant it doesn't make sense to just bolt on login-with-crypto-primitives. Indeed, the power in a key-as-identity is that it makes such a login possible, even trivial, but it is much wider than that as basic encryption of session keys and similar now becomes possible too. It indeed paves the way to allow clients to move away from username / password credentials alltogether.

zander:
As an aside, the older comment stating this is a "draft" because it does not specify which cryptographic signing algorithms are used, which endpoints exactly are needed, is a bit depressing in that it states now this proposal will not be considered.

The fact is that this kind of proposals don't get written by one person. They don't even get comments when asked on the drafts matrix rooms. Most people don't know these topics.

So when this topic is proposed and the higher level overview is stated (so without those endpoints and details), the next step is that the people which are active in Matrix can look how this can actually fit into the existing designs. Something that I and other crypto-geeks don't have the experience for.

Saying this will not be considered means only matrix devs can make contributions. Which is sad.

izN8nu6RyeneG5XnBoBgyRMVGH6H43WF:
@zander If this MSC concerns MXIDs specifically, I would oppose baking "public key as an MXID" into the Matrix spec, as it impairs user account mobility.

Consider if a user's private key is compromised under your proposal. This is a very real possibility for a user who had their key on a stolen phone, or who gets infected with malware. They then have to update all their contacts and other systems of a replacement identity, and there is no way to reclaim the compromised original, as the corresponding private key cannot change. This is like a user being unable to update their password after somebody looked over their shoulder, instead having to create a new account. This is why decoupling identity and authentication is so important.

Instead, I would recommend specifying a DID resolution _method_ in the Matrix spec, per [Decentralized Identifiers (DID)](https://www.w3.org/TR/2022/REC-did-core-20220719), a recently approved W3C recommendation. The _method_ would specify how to _resolve_ authentication information (such as a public key) for a user identity, but we could write the Matrix spec to permit the authentication method to be changed, while keeping the identity intact. We would then update the _method_ part of the Matrix spec to list all authentication methods a server must support.

I think this would permit the public key authentication scheme you describe, while improving user account portability, and allowing the Matrix spec to permit other authentication methods that meet a variety of needs. For more information about account portability, see #2787.

zander:
> Consider if a user's private key is compromised under your proposal. 

This is a common misunderstanding of the space. If you have an ssh-key you'll see where you made the mistake. The same crypto concepts are used in ssh.

The simple solution is to have a local password on your private key. Similar to 'ssh-add' and naturally gpg and others that do the same.

This is why the spec writes;

> having a password change is no longer something that happens on the server, it happens completely local on the client. 

This means that even if your private key is leaked, it is an encrypted file that can't be used without the password. You even increase your security because you can have a different password on different devices while the server knows only one identity.

I believe that solves the issue you broght up, right?

Please realize that just because its published by w3c doesn't mean its great. Its still design by committee, its still not actually used in practice and issues are likely to come up. A good standard takes time and while w3c has some good ones it would be mistaking correlation for causation by assuming its good because its w3c. Please judge on merit, not on source. Its also called experimental.

Already now the idea of DiD is generally seen as missing the point and throwing the baby out with the bathwater. A fancy word like blockchain, wielded with the hope it will be the next best thing.

Back in the real world experts know, **you can't bolt security on top after the fact**. For instance, you can't improve password security by adding crypto on top. Security is always about the weakest link, with that simple fact you can realize the idea of tying various authentication methods to one identity is the worst possible solution. A weakness in one comprimises all.

Do a search on google with "Paypal login bypass" to get the picture.

izN8nu6RyeneG5XnBoBgyRMVGH6H43WF:
>@zander said:
>The simple solution is to have a local password on your private key...if your private key is leaked, it is an encrypted file that can't be used without the password.

This is a password on your private key _file_. That is not a solution. Private keys are not private key files. The private key must be loaded from the decrypted file, unprotected, into local memory at some point. Whether the key remains in memory is entirely dependent on the implementation of the crypto libraries and the host operating system.

In your ideal world, everyone's crypto implementation is perfect, the key is promptly and permanently erased from memory, and is never unprotected in the first place due to homomorphic encryption. A file password might guarantee secrecy up to the password in that case. But in the real world, if a key exists, it can be compromised. Matrix is a real-world protocol, so it must be designed accordingly.

>Its...design by committee
>A good standard takes time

The W3C DID Core standard was developed with feedback from industry experts and theorists the world around. It is intentionally designed to be an amalgamation of their various use cases, so its semantics provide a framework generic enough to describe nearly any conceivable identity use case, including centralized or federated (read: Matrix) identity systems.

Also, the standard is over 15 years in the making. Draft-writing for this standard began over 3 years ago, and the need for decentralized identifiers was already being discussed by 2007. Weitzner, in ["Whose Name is it Anyway? Decentralized Identity Systems on the Web"](http://dig.csail.mit.edu/2007/06/ieee-ic-decentralized-identity-weitzner.html), explicitly describes "Being Known on the Web as your URI", and describes how OpenID partially solves that issue.

The notion that a "committee" suddenly or single-handedly designed this, at least, any more so than a "committee" designed Matrix, is nonsense.

>Its also called experimental

The [spec](https://www.w3.org/TR/2022/REC-did-core-20220719/#sotd) explicitly states:

>At the time of publication, no additional substantive issues, changes, or modifications are expected.

It is not called experimental. In fact, the only mention of "experimental" is with respect to certain DID _methods_ submitted to the [DID Specification Registries](https://www.w3.org/TR/did-spec-registries/), and some implementations. That those methods are experimental is irrelevant, as the only _method_ Matrix would concern itself with is a yet-to-be-defined "matrix" _method_. DID Core is intended to be used as an identity framework for globally unique identifiers, regardless of which _methods_ are available.

>judge on merit
>blockchain

I read the entire spec, and I'm not recommending W3C DIDs on a whim, nor on their relevance to blockchain technologies. DIDs are not designed only for blockchains. Rather, I recommend it because it was designed with decentralization and interoperability in mind, the very same principles Matrix is based on. In addition, the spec provides a framework for describing identity-handling information about other protocols (such as Matrix, a decentralized web service, or your favorite blockchain), including, but not limited to:

- differences in how your protocol treats human-readability and other aspects of identifiers
- how your protocol handles identification and authentication, and whether they handle them separately
- how your protocol handles identity stewardship or _control_
- how your protocol validates changes to an identity, including authentication methods or _control_
- your protocol's intended _dereferencing_ of identifiers into resources or URLs

DID semantics obviously support Matrix's use case, but more importantly, it probably supports any conceivable identity system and security model Matrix uses in the future. **No other identification scheme in the world can make this claim.** By using DIDs, we allow the Matrix protocol to evolve in response to newly discovered technical or legal needs, without having to worry about making changes to our identifiers, so users can rest assured that their identifiers won't be deprecated.

>you can't bolt security on top after the fact
>a weakness in one [authentication method] comprimises[sic] all

There's several problems with this:

1. Whether you can bolt "security" on after the fact depends on your "security" model, and the change you are making.
2. "Security" in Matrix must be defined relative to the project goals of making a highly interoperable federated communications system.
3. "Security" is not merely confidentiality. Undergrad-level security courses almost always note the CIA triangle, A as in "availability".
4. The method you suggest risks making your Matrix identity unavailable in case of private key loss or compromise. This is unacceptable to the average user, and the Matrix Foundation has made it quite clear that Matrix must support average users' security models.
5. Likewise, the risk of compromising confidentiality is often an acceptable risk for average users, provided that having multiple authentication methods improves availability.
6. None of this is even relevant to the sufficiency of DIDs, because the entire purpose of W3C DIDs is to allow Matrix to choose its security model independently of its identifier scheme.

>password security

The Matrix spec describes ways for Matrix identities to authenticate. This part of the spec can be handled independently of the identification scheme, and DIDs are designed with this in mind.

Also, under the current protocol, homeservers generally decide what authentication methods to permit. Many homeservers provide multiple authentication methods, including 3rd-party SSO, or even an SSH-like authentication method like you suggest (perhaps via a protocol extension). If we were to follow your suggestion, users would be _required_ to use this authentication scheme. So, not only is this proposal not backwards compatible with existing authentication methods, but it also violates the Matrix principle of federation, since you're constraining how homeservers may authenticate their users.

---

Overall, I think saying that a general-purpose protocol like Matrix shouldn't use W3C DIDs is like saying a web browser shouldn't use URLs. Just as failing to support URLs results in fragmentation in what should be a scheme for a "globally unique" resource identifier, failing to support DIDs also results in fragmentation. Indeed, the DID spec requires that DIDs are "globally unique", in the same sense that URLs are unique in practice.

This is the gist of my case for DIDs being preferable to public key identifiers, having read the spec.

I think making a case against DIDs would also require actually reading the spec.

MTRNord:
Could you please use a thread? It really starts to make this pr unreadable :/

izN8nu6RyeneG5XnBoBgyRMVGH6H43WF:
>@zander said
>please keep this on-topic

Perhaps I misunderstood the topic. When I discuss a PR, I ultimately do not care about the solution, but rather the problem which the solution aims to solve. You described a twofold problem:

- There is a lack of unified authentication for the various keys used in Matrix. Although they are associated with a single identity, the secrets used to protect these keys are often distinct, and not necessarily the same as the login authentication secret (usually a password). Among other issues, this makes backups more difficult.
- Matrix generally uses password-based authentication. Passwords are less secure than public key authentication.

And the rest of your MSC describes why, among other things, it's a good idea to permanently use the unified public key as a user identifier. The point of my response:

- It's not a good idea, as it violates Matrix design principles, not good to force this on average users, etc.
- There exists an alternative, with justification.

Part of the justification for the alternative is that it addresses other identity-related problems plaguing Matrix. This isn't just about your MSC, but about the Matrix ecosystem as a whole. I tried to keep the scope of the comment as closely related to the problem at hand, and I still stand by it as being wholly on-topic, albeit a "wall of text". I could have elaborated on how a Matrix DID _method_ would also address the portability issue, or issues in the non-inclusive 5-point list, but I didn't, except for identification and authentication methods, which are on-topic.

>"In your ideal world" is [a personal attack]

I described "ideal" conditions needed to avoid having to change user identifiers occasionally. You described a system where identifiers shouldn't need to be changed, so I naturally assumed we had the same "ideal" conditions in mind. If my analysis was mistaken, please correct the analysis.

proposals/3839-primary-identity-as-key.md
```diff
@@ -0,0 +1,120 @@
+# MSC3839: primary identity as public / private key-pair
```
richvdh, line 1: @zander [wrote](https://github.com/matrix-org/matrix-spec-proposals/pull/3839#issuecomment-1351865576)

> Its really off-topic to write lots of pages of text about another proposal on this PR

I don't think that's fair. I take no view as to which solution is preferable, but if there are two proposals to solve a problem, it is *absolutely* on-topic to describe why you see another solution as preferable.

Everyone: we aim to achieve consensus on what constitutes a good change to the spec, and we do that via *discussion*. People will have different opinions, and that is fine, but please be nice about it, and assume good intent rather than malice. Expressions such as "Back in the real world" and "Do a search for..." are, frankly, condescending and are more likely to aggravate than lead to quality discussion.

proposals/3839-primary-identity-as-key.md
```diff
@@ -0,0 +1,120 @@
...
+key which had no password will not be able to protect their account from
+others taking over.
+
+In my opinion this is an acceptable risk. The attack model doesn't really
+change and phone-local (or laptop local) protections are the same as any
+other sensitive data stored on that device.
+
+## Alternatives
+
+None yet.
```
richvdh, line 110: @izN8nu6RyeneG5XnBoBgyRMVGH6H43WF [wrote](https://github.com/matrix-org/matrix-spec-proposals/pull/3839#issuecomment-1349910897)

> Is this a duplicate of https://github.com/matrix-org/matrix-spec-proposals/pull/1768?

@zander replied

> the referred proposal is quite different and in my biased opinion not very useful.

Nevertheless the two set out to solve very similar problems. This is the place to mention that alternative proposals do exist, and to summarise why your proposal is superior.

richvdh:
@zander @izN8nu6RyeneG5XnBoBgyRMVGH6H43WF Please use comments on [the file changes view](https://github.com/matrix-org/matrix-spec-proposals/pull/3839/files) to start discussions, so that Github tracks them as separate threads and they can be marked as "Resolved" once addressed.

I've hidden the comments in the main PR for that reason.

proposals/3839-primary-identity-as-key.md
zander: > if there are two proposals to solve a problem, it is absolutely on-topic to describe why you see another solution as preferable.

Everyone should be able to say what is wrong with this one, which is better solved in another one. But maybe the details of how its better solved should be on "the other one". Especially when its mostly personal opinion as what we observed this week is mostly propoaganda of another solution and coming up with not very well researched reasons why this one is bad.

Thank you for hiding the messages, I hope we can see more pragmatic work here going forward.

realdeal1984:
My Brothers Best friend showed me stolen CIA software that reads minds by radiation remote neural monitoring 

Untreated?
He's using it to spy on kids! Help me take this lunatic, CJ Bess

PR #3842, opened 2022-07-02
[WIP] MSC3842: Placeholder for power levels & extensible events

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/extev/power-levels/proposals/3842-power-levels-message-events.md)

### DIFF
--- a/proposals/3842-power-levels-message-events.md
+++ b/proposals/3842-power-levels-message-events.md
@@ -1,0 +1,40 @@
+# MSC3842: Power levels on message (extensible) events
+
+Currently in Matrix anyone can send an event of any type in most rooms, though some rooms (like
+announcement-only) rooms limit what some people can send. It's additionally been desirable to
+be able to ban images or other rich media in a room which might be bridged to a less capable
+platform, though this use-case is less common.
+
+This proposal aims to try and figure out power levels for events from the perspective of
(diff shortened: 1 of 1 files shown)


proposals/3842-power-levels-message-events.md
```diff
@@ -0,0 +1,40 @@
+# MSC3842: Power levels on message (extensible) events
```
turt2live, line 1: Thought: when a client decrypts a message, it will need to determine if that event still meets the power levels, but with extensible events it's trivial to bypass "you can't send images" (for example) because you can just create a custom event type with `m.image` fallback. 

If we solve the problem by making power levels apply to content blocks, we're potentially causing problems because an `m.poll.start` might be legal for senders, but `m.image` is ignored. Some clients would render the event while others wouldn't. 

A potential solution might be dual consideration? If the event type is specifically listed in the power levels, assign that required PL to the event. If the event type isn't in the power levels, instead try to look for the implied events for fallback representations. If those implied event types are further not specifically listed, then the event is presumed allowed.

This would allow a room to support `m.poll.start`, for example, but prevent a malicious user from inventing an event type to spam images. There's still potential risk that a client doesn't support polls and therefore ends up picking the `m.image` representation, but presumably we can further have clients verify that the implied fallback representation's event type is also allowed, with `m.text` always being allowed for fallback.

PR #3215, opened 2021-05-24
Draft: MSC3215: Aristotle - Moderation in all things

Yoric:
A step towards decentralized moderation/reputation.

[Rendered](https://github.com/Yoric/matrix-doc/blob/aristotle/proposals/3215-towards-decentralized-moderation.md).

PR #2938, opened 2021-01-05
MSC2938: Report content to moderators (via server notice rooms)

Yoric:
As a followup to https://github.com/vector-im/element-meta/issues/1848, this proposal introduces a mechanism to report content to room moderators instead of reporting it to homeserver administrators.

Rendered: https://github.com/Yoric/matrix-doc/blob/master/proposals/2938-report-content-to-moderators.md

PR #3843, opened 2022-07-11
MSC3843: Reporting content over federation (API approach)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/federated-reports/proposals/3843-federated-reports.md)

### DIFF
--- a/proposals/3843-federated-reports.md
+++ b/proposals/3843-federated-reports.md
@@ -1,0 +1,82 @@
+# MSC3843: Reporting content over federation
+
+Clients can currently [report content](https://spec.matrix.org/v1.3/client-server-api/#reporting-content)
+to their homeserver, however these reports don't currently translate over federation to affected
+servers.
+
+Some content might want to be reported directly to the server which generated it, or tooling on a
+given server could find it desirable to generate a report to be sent to a specific server. For example,
+if a homeserver plugin detected spam then it might wish to report that directly to the server which
+generated the content, bypassing any local report handling which might be attached to the local server's
+client-server API endpoint.
+
+## Proposal
+
+Mirroring the client-server API a bit, we introduce a new endpoint for reporting events in rooms over
+federation: `POST /_matrix/federation/v1/rooms/{roomId}/report/{eventId}`. To report a whole room the
+caller would reference the `m.room.create` event. To report a user within the context of that room the
+caller would use the `m.room.member` event for that user. Otherwise, the potentially objectionable
+content's event ID would be used to report that content.
+
+The new endpoint takes very similar body parameters to the client-server API, though with the `score`
+notably missing and `reason` being explicitly required. For example:
+
+```
+POST /_matrix/federation/v1/rooms/!bad:example.org/report/$spammyspam
+{"reason":"This message is spam"}
+```
+
+`reason` cannot be blank, unlike the client-server API.
+
+It is intentional that the original reporter is not sent over federation, however given the endpoint
+is authenticated using the server's name it may still be possible to identify the original reporter.
+If the homeserver holds exactly one known user then there's a pretty good chance they reported the
+content.
+
+How the report is handled is deliberately left as an implementation detail. Homeservers might wish to
+add the record to their existing reports table/system, or create a new handling pipeline for it.
+Similarly, when to call this endpoint is also left as an implementation detail. It may not be desirable
+for all client-server `/report` calls to be proxied over federation, for example.
+
+The content being reported *should* originate from the server it is being reported to, but that is
+not a requirement. It may be valid to notify other homeservers that there is spam in a room to help
+all affected users, for example. It is up to the implementation if it wants to handle content generated
+from other servers - if it doesn't, it can respond with an `M_UNACTIONABLE` error code (new with this
+proposal) and an HTTP `400` status. Further, if the server just doesn't want to deal with federated
+reports at all it can return a `400` `M_UNACTIONABLE` error for everything.
+
+Homeservers are strongly encouraged to rate limit the number of reports which come in. A possible
+approach might be to limit servers to 10 reports per minute, or similar.
+
+## Potential issues
+
+This could open up servers to receiving spammy or abusive reports, however given most servers have
+a federation-capable reporting mechanism already (email, community room, etc) this is not expected
+to have much of an impact. The added recommendation that things be rate limited also helps a lot.
+
+## Alternatives
+
+Briefly while writing this proposal it was considered to expose a shared policy room between the servers,
+where somehow Server A and Server B negotiate a room where they can publish
+[policy rules](https://spec.matrix.org/v1.3/client-server-api/#moderation-policy-lists) that the other
+side picks up and actions, if needed. This feels overly complex for the time being, but if proven useful
+could be explored in a different MSC.
+
+## Security considerations
+
+Servers should protect themselves from abusive or spammy reports. Rate limiting and rejecting content
+that didn't originate from itself will help a lot on that.
+
+Like with the client-server API, servers should be cautious about hooking up automated responses to
+reported content. Without the added context of *who* made the report, it's very possible that the content
+is being reported 40 times by the same person rather than 40 different people - this could easily trip
+an automated response of spam handling or deactivation if not careful.
+
+## Unstable prefix
+
+While this MSC is not considered stable, implementations should use `/_matrix/federation/unstable/org.matrix.msc3843/rooms/{roomId}/report/{eventId}`
+as the endpoint. The body stays as-proposed.
+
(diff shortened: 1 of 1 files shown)


Yoric:
At first glance, I fear that this may have the usual problem that it reports content to the homeserver administrator, who typically cannot do much about it.

proposals/3843-federated-reports.md
```diff
@@ -0,0 +1,82 @@
+# MSC3843: Reporting content over federation
```
turt2live, line 1: @Yoric says:

> At first glance, I fear that this may have the usual problem that it reports content to the homeserver administrator, who typically cannot do much about it.

Yoric: Let me elaborate.

I agree that we definitely should have mechanisms for reporting content over federation. However, the current mechanism for reporting content is essentially broken in many ways. The most important one in my eyes being that content reporting is a mechanism to get in touch with the (current) homeserver's administrator who:

- often cannot actually look at that content;
- often cannot actually redact that content;
- in the P2P future, is the same person as the offender.

While this proposal does address an existing problem, I believe that it does so by extending a broken mechanism.

I believe that an approach comparable to https://github.com/matrix-org/matrix-doc/pull/2938 (disclaimer: I'm the author, so I'm biased), which directs reports towards room moderators, is the way forward.

aaronraimist: It would be good to have ways to do both things. Room moderators can take more effective action in that specific room but only server admins can deactivate the account to prevent it from spamming/abusing other rooms.

turt2live: it's worth noting that the current spec for `/report` doesn't actually say how you're supposed to handle the report. Synapse just dumps it into a database table, but it has provision and capability to notify room moderators, server admins, etc if it wanted to - the federation side is simply to introduce this same generic approach to ensure things route to a sensible place.

It is not expected that 100% of calls to the client-server `/report` endpoint result in a federation call. Instead, it is expected that the reporting server be intelligent to know when another *server* can deal with the report rather than the room moderators/local server. 

Such examples would be reporting spam originating from a server where the server admin might want to shut down the account, or reporting illegal content on a server where the server admins need to get involved to handle removal. 

If a server automatically reported everything to remote servers, it would be in bad taste and likely blocked by the servers who receive effectively reporting spam.

proposals/3843-federated-reports.md
```diff
@@ -0,0 +1,82 @@
...
+## Proposal
+
+Mirroring the client-server API a bit, we introduce a new endpoint for reporting events in rooms over
+federation: `POST /_matrix/federation/v1/rooms/{roomId}/report/{eventId}`. To report a whole room the
+caller would reference the `m.room.create` event. To report a user within the context of that room the
+caller would use the `m.room.member` event for that user. Otherwise, the potentially objectionable
+content's event ID would be used to report that content.
+
+The new endpoint takes very similar body parameters to the client-server API, though with the `score`
+notably missing and `reason` being explicitly required. For example:
```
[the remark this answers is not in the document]
turt2live: The `score` isn't used by anything, and wouldn't really inform any useful data anyways. Making the `reason` required is to prevent abuse of the system: reporting over federation without reason is just spam.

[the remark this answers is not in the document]
timokoesters: But now servers can't automatically forward reports to the sender's server and instead have to make up a reason or require manual action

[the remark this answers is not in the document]
turt2live: [They shouldn't be doing that](https://github.com/matrix-org/matrix-spec-proposals/pull/3843#discussion_r920399646)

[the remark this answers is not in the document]
turt2live: Sorry, it's an entire use case I am not willing to support in this MSC at this time. Automated reports are less than helpful, as highlighted by other threads already.

proposals/3843-federated-reports.md
Yoric: I would suggest that `/report` is not a good example to be followed. At the very least, its current implementation in Synapse feels designed to block it from being useful.

turt2live: err, if the Synapse implementation is bad that doesn't mean the spec is bad. I quite like the generic handling of the endpoint, but personally don't particularly enjoy Synapse's handling of it.

Kladki: > It is expected that the reporting server be intelligent to know when another server can deal with the report rather than the room moderators/local server.

Would something like [MSC2938](https://github.com/matrix-org/matrix-spec-proposals/pull/2938) or [MSC3215](https://github.com/matrix-org/matrix-spec-proposals/pull/3215) not allow for the server knowing whether it can be handled by a remote server admin?

Also as per [reports without reasons being spam](https://github.com/matrix-org/matrix-spec-proposals/pull/3843#discussion_r920400955) and automatic report forwarding, surely the admin would be viewing the event anyways to see if the concern is actually correct, so they would be able to tell if the content is against their terms of service, and hence the `reason` shouldn't be required and automatic report forwarding should be able to happen.

turt2live: Those MSCs are part of the wider system for reporting, and could be used to help make decisions. 

When we're working over federation, it's difficult to reliably assume that the remote server is handling reports safely. For example, they may be automatically banning users they receive too many reports for, or may be using reports to send abuse back to the reporter. Sometimes this is malicious, but often it's unsafe code being run on the receiving side. We can prevent a lot of these cases by discouraging automatic forwarding, and, more on the Foundation side, building tools which exemplify safe reporting practices.

Kladki: > automatically banning users they receive too many reports

The problem with this is that even if all compliant homeservers do not automatically forward reports, any bad actor can spam reports anyways, basically making it a lose-lose (as not only do you lose the benefits of automatic forwarding, but you also don't gain anything in return as bad actors can abuse your assumptions). IMO automatically banning is the type of behavior we should be discouraging instead.

> using reports to send abuse back to the reporter

When doing manual reports by trying to find the server admin's contacts and then reporting content that way, this issue still can occur.

Automatic report forwarding could have the potential to significantly reduce the friction of reporting bad actors over federation, as once the report is sent you only have to wait for the remote admin to take action. This of course can cause the bad actor to be stopped before they can do any more damage, compared to if in addition you have to wait for your admin to then forward the request. I really hope you reconsider your stance on this. 🙏

turt2live: A sending server can choose to utilize automatic sending, but this MSC will not be encouraging it due to the safety concerns.

proposals/3843-federated-reports.md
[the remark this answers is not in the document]
Kladki: > reporting over federation without reason is just spam.

As per my other comment:
> surely the admin would be viewing the event anyways to see if the concern is actually correct, so they would be able to tell if the content is against their terms of service. So how would it be concidered spam if there is no reason?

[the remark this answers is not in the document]
turt2live: The server admin has nothing to go on for how to start investigating the report. They're being told about something, but they're not being told *why* they should look into it. This is akin to DMing the server admin with permalinks - they will have follow-up questions because it's nonsensical and looks like spam.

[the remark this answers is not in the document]
Kladki: I mean it's a content report, so it's closer to telling the admin that you want to report a user because of the content in the permalinks, but like you said, not specifying the reason alongside the link. Considering that the admins should be able to access the reported content, they can assess whether a punishment is required and to what extent.

For example, if I get a content report with no reason specified, I can view the content and see if there is anything against my TOS. This holds true even if you get given a fake reason, and notably doesn't change even when a true reason is given.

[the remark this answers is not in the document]
turt2live: Not all the content is safe to view, nor is the report always going to be a piece of content. Something this MSC doesn't (yet) consider is that a new reporting system is on the horizon where different types of things can be reported: events, rooms, servers, policies, decisions, appeals, etc. 

I do strongly consider a reason as required in lieu of a more detailed reporting API which covers these cases. Particularly and API which may be better suited to classifying the report, giving the receiving admins/moderators information about the report without having to view the content.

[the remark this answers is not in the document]
turt2live: Most likely, yes. It's also less important for local reports, but over federation the amount of anonymity increases risk.

proposals/3843-federated-reports.md
turt2live:
```suggestion
# MSC3843: Reporting content over federation (API approach)
```

PR #3846, opened 2022-07-13
Add proposal for allowing widgets to access TURN servers

robintown:
[Rendered](https://github.com/robintown/matrix-doc/blob/widget-turn-servers/proposals/3846-widget-turn-servers.md)

### DIFF
--- a/proposals/3846-widget-turn-servers.md
+++ b/proposals/3846-widget-turn-servers.md
@@ -1,0 +1,140 @@
+# MSC3846: Allowing widgets to access TURN servers
+
+*The following is adapted from [MSC3819](https://github.com/matrix-org/matrix-spec-proposals/pull/3819).*
+
+Widgets (embedded HTML applications in Matrix) currently have a relatively large surface area they can use for interacting with their attached client, primarily in the context of a room. They can send/receive [events with MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762) and [to-device events with MSC3819](https://github.com/matrix-org/matrix-spec-proposals/pull/3819), allowing them to simulate an entire Matrix client in application code for specific, limited use cases.
+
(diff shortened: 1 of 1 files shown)


proposals/3846-widget-turn-servers.md
```diff
@@ -0,0 +1,140 @@
...
+
+While we could expose the entire client-server API over `postMessage` (or similar) for embedded clients to access, the permissions model gets hairy and difficult to secure on the client side. Instead, we're exploring what it would look like to special case what is needed for specific applications, as needed.
+
+The specific goal of exposing TURN servers is to make it possible for widget-ized Matrix client to implement [MSC3401 - Native group VoIP](https://github.com/matrix-org/matrix-spec-proposals/pull/3401), when combined with the abilities of [MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/3819) and [MSC3819](https://github.com/matrix-org/matrix-spec-proposals/pull/3819).
+
+## Proposal
+
+We introduce a new capability, `m.turn_servers`, used for accessing TURN server credentials. When approved, it opens up access to the following actions:
+
+**`fromWidget` action of `watch_turn_servers`**
```
turt2live, line 19: it's not clear that widgets would need ongoing updates of TURN server data - wouldn't it immediately request/receive the servers and use them, effectively discarding any future updates?

(the explanation for either should be in the MSC)

proposals/3846-widget-turn-servers.md
```diff
@@ -0,0 +1,182 @@
...
+  "widgetId": "20220712_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "unwatch_turn_servers", // value defined by this proposal
+  "data": {} // nothing
+}
+```
+
+This action tells the client to stop sending the widget TURN server updates. As with `watch_turn_servers`, the `data`
+and `response` fields for this action are empty. If the widget was not already watching for TURN servers, the client
+sends back an error response.
```
dbkr, line 90: It might be a bit more matrix-y to return success if it wasn't listening before, that way it's idempotent and probably a bit easier for the client.

dbkr (approved):
Looks good from my PoV - subscription model is a nice way of doing it and avoids the widget & the client having different ideas on when they want fresh credentials.

PR #3659, opened 2022-01-19
MSC3653: Invite Rules

joshqou:
[Rendered](https://github.com/joshqou/matrix-doc/blob/85b49895e2530ef5982bfc623be3ab37fe2b3ce4/proposals/3653-invite-rules.md)

PR #3847, opened 2022-07-13
Yoric/individual policy rooms

Yoric:


proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,137 @@
...
+2. We build upon the mechanism of recommendations, also defined in MSC2313,
+and create a new recommendation for ignoring invites from a specific user, from
+a specific server or towards a specific room.
+
+
+## Proposal
+
+### Background
+
+MSC2313 defines policy rooms. A policy room is a room in which rules such
```
turt2live, line 25: The MSC number is significantly less important than the spec which has existed for years :( 

please link to the spec rather than the MSC.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,137 @@
...
+- any client or trusted agent acting on behalf of Alice will monitor room R for
+    new policies and apply the recommendations on behalf of Alice.
+
+### A recommendation for ignoring invites
+
+We expand the **`enum`** `recommendation` with the following value
+
+| Value | Description |
+|---------|-------------|
+| `m.invites.ignore`  | The user's client should not display any invite from/to the entity specified in `entity`. |
```
turt2live, line 72: If we're introducing a new recommendation, let's shorten it to `m.ignore` to retain the generic nature of recommendations.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,137 @@
...
+
+In particular, if Alice has a policy with `recommendation` `m.invites.ignore`:
```
turt2live, line 74: this should be a suggestion for implementation, not a requirement.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,137 @@
...
+clarifying how to associate an issuer with a policy room.
+
+### Associating a policy room with a user
+
+For individual users, we introduce a new event `m.policies`, designed
+to be used as part of account data. This event has content:
+
+| Content | Type | Description |
+|---------|------|-------------|
+| `room`  | Room ID or Alias | The main room in which a user may publish policies to be applied on their behalf. |
```
turt2live, line 51: I think it'd be better to make this an array, for decentralized reputation reasons.

turt2live (commented):
initial feedback copied from MSC3840

proposals/3847-ignoring-invites-with-policy-rooms.md
turt2live: What do we gain by being overly restrictive? :P 

The original design of policy rooms and the `recommendation` is that we would only ever need ~5 to cover *everything*, because we deliberately and intentionally do not assign behaviour to the recommendation - we assign ideas and suggestions. `m.ignore` *suggests* that the implementation ignore the described entity, but doesn't explain how or why that might be done. Context clues, like "here are the policy lists I want you to ignore invites from", help inform a better understanding of what to do with that policy rule/room. Making it specific just means we need more MSCs to introduce more recommendations.

Yoric: I don't really understand the rationale.

My objective is to let users click on "ignore invites" in the UI of the client and have invites ignored. If the user clicks on "ignore invite", why would they want the client to do anything other than ignoring invites?

proposals/3847-ignoring-invites-with-policy-rooms.md
Yoric: Well, the idea was to add the array as a followup MSC, for decentralized reputation reasons, and keep this MSC simpler by focusing on a single source.

Regardless of whether it's a single source or several, we need one of them to be primary. Otherwise, the client that is used to issue policy events won't know where to send these events. I suppose that I could arbitrarily pick the first from the array, but that feels really fragile.

proposals/3847-ignoring-invites-with-policy-rooms.md
Yoric: So, if I understand correctly what you write, we still need to specify somewhere the meaning of `m.ban`, but we need to store the information somewhere else. If this is meant to work across clients, I suppose that this means either in a state event within the policy room ("hey, in this room, `m.ban` means ignore!") or as an annotation in `m.policies` in the account state ("hey, in *that* room, `m.ban` means ignore!")

Do I understand correctly?

turt2live: The point is we don't specify the meaning of `m.ban` - we don't today, and wouldn't with this MSC either. It'd most likely be a flag on the account data. Maybe something like this?

```json5
{
  "type": "m.policies",
  "content": {
    "m.ignore.invites": [ "!room:example.org", "!room2:example.org" ],
    "m.ignore.users": ["!room:example.org" ] // we should be able to reuse the rooms for different purposes
  }
}
```

gives us a bit of flexibility too to define custom/future policy usages (hypothetical decentralized reputation, for example).

turt2live: for clarity: the context is *implied* rather than *explicit*, which is how we get away with using `m.ban` (or a generic `m.ignore`).

turt2live: it has been useful in the past, yea. It's how the original Mjolnir bot, synapse module, and Element Web implementation all did different things with exactly the same information.

Yoric: So, the latest version of my understanding, based on our conversation, is that "policy lists" are not about policies at all, and could just have well be named "entity lists". Also, that `recommendation` is actually kind of useless?

turt2live: They're still policy lists containing rules. Eventually we want to put non-entity rules in there as well, which is where the `recommendation` also comes in.

The original design for policy lists included future proofing for decentralized reputation, access control, and more.

Yoric: I'll see what I can do.
Note that we still need the ability to have a "main" room in which to issue these rules.

proposals/3847-ignoring-invites-with-policy-rooms.md
turt2live: The client can make decisions about which one to post to based on how the room is structured. If someone subscribes to the code of conduct list on matrix.org then they won't have write access, but if the client sees a room that the user created then it can publish there. 

Arrays are ordered in JSON, so would heavily rely on that as well.

Yoric: > The client can make decisions about which one to post to based on how the room is structured. If someone subscribes to the code of conduct list on matrix.org then they won't have write access, but if the client sees a room that the user created then it can publish there.
>
> Arrays are ordered in JSON, so would heavily rely on that as well.

Sure, we could make the information implicit (or even use a heuristic such as "first room in the array in which we can write state events"), but I feel that this is much more brittle. For instance, if a user suddenly becomes homeserver moderator and gains rights to the homeserver's coc room, we don't want this to accidentally change into which room they store their individual policies.

So, if we introduce an array, I strongly believe that we need to make an explicit difference between the target room and the source rooms.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,146 @@
...
+| `m.ignore.invites` | `Policies` | A list of rooms in which policies should be understood as "ignore invite" lists. |
+
+We expect that future MSCs will expand upon this event `m.policies` and
+add further fields with other lists of rooms in which policies should be
+understood differently. This is, however, beyond the scope of the current
+MSC.
+
+We also expect that future MSCs will expand upon this event `m.policies`
+to add semantics in other positions than account data, e.g. community rooms.
+This is also beyond the scope of the current MSC.
```
deepbluev7, line 60: I do have some concerns regarding that, which maybe should affect the design even now:

I moderate 3 very different communities with partially shared and partially separate banlists (and one of them is a bigger community moderation effort across very different communities). In this case we currently have usually 2 different policy rooms per community, although this varies a bit. So in my case that would be 5 separate policy rooms.

Currently we only issue `m.ban` and having a specific list for ignores or invite ignores probably has different rules, but in general I think there needs to be a method to pick out of several policy lists as a target. This can of course have different methods, that might be useful. For example since usually community rooms are inside a space, it might make sense to pick the policy rooms for that space. This would probably be set in the room account data. Alternatively targets could be a list, where the client automatically picks the only list, if there is only one, but if there are multiple elements, it could offer a choice.

Now yes, this is out of scope for this MSC, but I think if it adds something incompatible with such goals, that might interfere with some use cases in the future?

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,146 @@
...
+
+Where `Policies` is defined as:
+
+| Content     | Type | Description |
+|-------------|------|-------------|
+| `target`    | Room ID or Alias   | A room in which to issue new policies. |
+| `sources`   | Room ID or Alias []| A list of rooms from which to apply new policies. Should generally include `target`. |
```
deepbluev7, line 67: Is an alias "safe" in this case? While this allows one to easily upgrade the room, generally one should already be a member of this room anyway (unless peeking over federation works reliably), but having the alias on a different server possibly would allow a foreign entity to rewrite what rooms you trust without being a member of that room itself. I'm not sure if that is worse than having a malicious actor in the room, but it does sound a bit scary? And if that is ever intended to be written to a global event, for roomids you will possibly need a list of servers to join via, so possibly this should be an object?

proposals/3847-ignoring-invites-with-policy-rooms.md
Yoric: So you'd like all the users who join any room in your space to share a list of ignored invites? Or do I misunderstand?

deepbluev7: No, but if this is ever reused for something like m.ban, I don't want all my bans to automatically go to one list, I want it to go to different lists depending on what list I have open. I can't have a single "target" list in account data.

Yoric: So, if I understand correctly your concern, you're afraid that some Matrix client could start issuing policies into a community policy room (e.g. a space's CoC room) instead of your own individual policy room (e.g. the list of invites that you, as a user, don't want to see). Is that correct?

Normally, this shouldn't happen.

The idea is that all your individual bans (i.e. the bans that are expected to affect only events you receive) will go into your individual policy rooms, as specified by the `target`. However, your community bans (i.e. the bans that you're enacting as a moderator/admin/... for a room/a space/a server) go to their policy rooms, just as they do right now.

Unless you specifically configure your account to use a community room as your individual policy room (and I don't see any reason why someone would do that), these lists are entirely distinct. Chances are that they are going to be managed by different tools, too (your usual Matrix client for the individual policies, Mjölnir or another bot for the community policies).

Does this make sense?

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,146 @@
...
+with on the Matrix network. To alleviate this, Matrix needs a mechanism to
+let users specify that they are not interested in invites from specific
+users or towards specific rooms.
+
+In this proposal, we introduce the necessary extensions to let users do
+this and to let clients perform filtering of unwanted invites.
+
+We proceed as follows:
+
+1. We build upon the mechanism of policy rooms, defined in MSC2313, and define
```
uhoreg, line 13:
```suggestion
1. We build upon the mechanism of policy rooms, defined in [MSC2313](https://github.com/matrix-org/matrix-doc/pull/2313), and define
```

proposals/3847-ignoring-invites-with-policy-rooms.md
uhoreg: It's probably at least worth mentioning in the "Security Considerations" section.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,146 @@
...
+
+Just as `m.ignored_users_list` is handled mostly on the server, we could handle `m.invites.ignore`
+largely on the server. However, this would make it much harder to undo erroneous ignores (i.e.
+implementing some sort of recovery from trashcan) on the client.
+
+Consequently, we prefer handling things on the client, even if this may require more traffic.
+
+## Security considerations
+
+Can't think of any right now.
```
uhoreg, line 138: Maybe mention that you (obviously) need to trust the policy room admins to not be malicious and cause you to ignore invites from people that you actually want to talk to?

proposals/3847-ignoring-invites-with-policy-rooms.md
deepbluev7: No, my concern is that this duplicates/conflicts with how I planned to have Nheko apply policy rules in the next major release, where you can define a list of target policy rooms per space to which the ban recommendations get posted. The plan was to control that by adding a policy target account data event to each space, which has a list of targets and sources from which policy rules are taken or to which the policy rules are posted.

Yoric: This looks complementary, no?

You're defining source/target per space, I'm defining source/target for individual users.

proposals/3847-ignoring-invites-with-policy-rooms.md
Yoric: I've given it a few days trying to work things out on paper and I believe that, while context + rule is one way to express semantics, this is the wrong direction to head. 

# Option 1: One `recommendation` per semantics

In this case, this means adding a new `recommendation` `m.invite.ignore` or `m.ignore.invite`, rather than one new context of policy room.

Adding one `recommendation` per semantics has essentially a linear cost in terms of both specification size and cognitive load: if we want one new behavior, we add one `recommendation` and one paragraph in the spec.

It *might* make it harder to reuse lists in contexts for which they were not intended. For the time being, I don't see any real benefit to this.

# Option 2: Reuse `recommendation`s, introduce new meta-policies

By meta-policies, I mean ToS, CoC, specifying `m.ignore.invites` as a key in `m.policies`, etc.

Essentially, adding a new behavior to solve a problem means:

- either introduce a new meta-policy; or
- when a meta-policy is not sufficient to represent the expected behavior (e.g. numeric opinions, allow lists), introduce a new `recommendation`.

This has an explosive cost, both in terms of specification size and in terms of cognitive load:

- harder to determine whether we need to introduce a new meta-policy or a new `recommendation`;
- whenever we introduce either a meta-policy or a new `recommendation`, we need to specify all the semantics for each (meta-policy,  `recommendation`) pair and one paragraph in the spec for each pair — or worse, not specify it at all, leaving it undocumented territory;
- similarly, when implementing a tool that works with policy lists, this means more unexpected combinations to handle, which makes it harder to write, test and trust.

Costs for both options are in addition to whatever specifications we're going to need for composition if/when policies start becoming non-trivial (e.g. allow lists and block lists, numeric opinions, etc.)

proposals/3847-ignoring-invites-with-policy-rooms.md
deepbluev7: Well, usually you have an inheritance change of a global default (for the account) and then finer grained settings for each space/room.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,167 @@
...
+to be used as part of account data. This event has content:
+
+| Content | Type | Description |
+|---------|------|-------------|
+| `m.ignore.invites` | `Policies` | A list of rooms in which policies should be understood as "ignore invite" lists. |
+
+We expect that future MSCs will expand upon this event `m.policies` and
+add further fields with other lists of rooms in which policies should be
+understood differently. This is, however, beyond the scope of the current
+MSC.
```
Mikaela, line 56: Could this MSC include requiring invite reasons to be specified for an invite to be acceptable or is that something for a future MSC that these lines speak of?

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,167 @@
...
+
+A client could of course reject invites instead of ignoring them. However, experience shows that
+spam/bullying tools monitor invite rejections to send invite spam. This entire MSC is meant to
+offer an alternative mechanism that would close this gap and let users ignore invites without
+notifying the sender.
+
+### Server-side filtering
+
+Just as `m.ignored_users_list` is handled mostly on the server, we could handle `m.invites.ignore`
+largely on the server. However, this would make it much harder to undo erroneous ignores (i.e.
```
Johennes, line 142: It might be nice to mention https://github.com/matrix-org/matrix-spec-proposals/pull/3659 here which does the filtering on the server.

proposals/3847-ignoring-invites-with-policy-rooms.md
```diff
@@ -0,0 +1,167 @@
...
+
+There is a risk that the list of ignored invites of some users may grow a lot, which might have
+performance impact, both during initial sync and during filtering. We believe that this risk is
+limited. If necessary, clients may decide to cleanup ignored invites after some time.
+
+### Sync size
+
+With this proposal, any invite ignored with `m.ignore.invites` will still show up at each `/sync`.
+In time, this can grow expensive. This is, however, necessary to be able to un-ignore invites
+from the client.
```
joshqou, line 125: Literal rule matches can be filtered out [using a sync filter](https://spec.matrix.org/v1.10/client-server-api/#_matrixclientv3useruseridfilterfilterid_roomfilter)

proposals/3847-ignoring-invites-with-policy-rooms.md
joshqou: Both specs aim to fix the same problem but #3659 drops invites as opposed to ignoring them. imo dropping invites is better than a client-side ignore since it wouldn't require all a user's clients to support the spec, even if such behaviour is more "destructive".

Johennes: Sorry, I merely meant to say that mentioning #3659 here would be nice for context on prior art, not necessarily that I'd prefer one over the other. On that matter, filtering invites on the server sounds more efficient but it won't allow you to review ignored invites (and possibly even unignore them). So both seem to have pros and cons.

proposals/3847-ignoring-invites-with-policy-rooms.md
turt2live, line 1: Overall: Based on https://github.com/matrix-org/matrix-spec-proposals/pull/4192, I feel as though this MSC would be best positioned as an extension to https://github.com/matrix-org/matrix-spec-proposals/pull/4155 rather than being a dedicated mechanism (for the parts that aren't already covered).

Half-Shot: I think in practice this means:

 - `m.policies` is no longer a dedicated event, but a key under `m.invite_permission_config`. 
 - Policies should be evaluated *after* the other rules. Both so users can set their own overrides, and for efficiency. 
 - This should be evaluated on the server, too.

PR #3848, opened 2022-07-19
MSC0000: Introduce errcodes for specific membership change failures

Half-Shot:


proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Proposal
+
+New `errcode` would be introduced into the error body of a response
+(https://spec.matrix.org/latest/client-server-api/#standard-error-response). 
+
+`M_ALREADY_JOINED` would be fired when a membership action fails when the user
+is already joined to the room.
+This would cover endpoints:
+  - [POST /_matrix/client/v3/rooms/{roomId}/invite](https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3roomsroomidinvite)
```
turt2live, line 18: please don't use latest, as latest changes - v1.3 or current version is preferred.

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,60 @@
...
+`M_ALREADY_JOINED` would be fired when a membership action fails when the user
+is already joined to the room.
+This would cover endpoints:
+  - [POST /_matrix/client/v3/rooms/{roomId}/invite](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidinvite)
+  - [POST /_matrix/client/v3/rooms/knock/{roomIdOrAlias}](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3knockroomidoralias)
+  - [PUT  /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+Note that it would not cover endpoints where trying to join a room when the
+user is already joined would no-op, like `POST /_matrix/client/v3/join/{roomIdOrAlias}`.
+
+`M_INSUFFICIENT_POWER` would be when your user does not have the specific required power level to
```
Half-Shot, line 24: Note, this came up when thinking about other kinds of explicit errors we'd like when dealing with membership (as it would be nice to know if your user was denied inviting another user due to PLs).

Powerlevels extend beyond membership though, as they also control who can send events / redact and other fun things so it might be worth slightly extending the scope of the MSC to cover any event sending endpoints which might return a power level failure.

tulir:
mautrix-python implementation (includes `M_NOT_JOINED`): https://github.com/mautrix/python/commit/d71b1497ad348083480ee332cbc3d489f050b254

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,70 @@
...
+perform an action in the room.
+  - [POST /_matrix/client/v3/rooms/{roomId}/invite](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidinvite)
+  - [POST /_matrix/client/v3/rooms/knock/{roomIdOrAlias}](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3knockroomidoralias)
+  - [POST /_matrix/client/v3/rooms/{roomId}/unban](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidban)
+  - [POST /_matrix/client/v3/rooms/{roomId}/ban](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidban)
+  - [POST /_matrix/client/v3/rooms/{roomId}/kick](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidkick)
+  - [PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+  - [PUT /_matrix/client/v3/rooms/{roomId}/send/{eventType}/{txnId}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
+
```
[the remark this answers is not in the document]
Half-Shot: We settled on `M_NOT_JOINED` to be aligned with `M_ALREADY_JOINED`

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,73 @@
...
+  - [POST /_matrix/client/v3/rooms/knock/{roomIdOrAlias}](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3knockroomidoralias)
+  - [PUT  /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+Note that it would not cover endpoints where trying to join a room when the
+user is already joined would no-op, like `POST /_matrix/client/v3/join/{roomIdOrAlias}`.
+
+`M_INSUFFICIENT_POWER` would be when the authenticated user does not have the specific required power level to
+perform an action in the room.
+`M_NOT_JOINED` would be when the authenticated user is not joined to a room, but attempts to perform
+an action in it.
+Both errcodes would cover endpoints:
```
tulir, line 34: Synapse also returns not in room when `/leave`ing a room you're not in (I'm not sure if that should actually just be a no-op without erroring, but if it returns an error, it should be included in endpoints that return `M_NOT_JOINED`)

Half-Shot: I tested this on Synapse and found:
- Leaving a room that doesn't exist on the host returns "M_UNKNOWN" -> Not a known room
- Leaving a room that I was previously in will always no-op with a 200.
- Leaving a room on the host that I was never in I am unsure of, as I need to go test that.

I don't know whether a room not existing on a host should be a not-joined or not tbh.

Half-Shot: https://github.com/matrix-org/synapse/issues/13360

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,73 @@
...
+error handling in clients much more complicated. There is precedence already in the spec for specific
+error codes for specific failures, so there is little need to subclass.
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+While this MSC is not considered stable for implementation, implementations should use `org.matrix.unstable.errcode`
```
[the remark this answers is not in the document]
tulir: A generic unstable errcode field would be nice so each MSC improving errors wouldn't need a different one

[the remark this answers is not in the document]
Half-Shot: The counterpoint made was that other MSCs can re-use this MSC's key https://github.com/matrix-org/synapse/pull/13343#discussion_r928670044

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
...
+
+This would cover endpoints:
+  - [POST /_matrix/client/v3/rooms/{roomId}/invite](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidinvite)
+  - [POST /_matrix/client/v3/rooms/knock/{roomIdOrAlias}](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3knockroomidoralias)
+  - [PUT  /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+
+Note that it would not cover endpoints where trying to join a room when the
+user is already joined would no-op, like `POST /_matrix/client/v3/join/{roomIdOrAlias}`.
+
+`M_INSUFFICIENT_POWER` would be when the authenticated user does not have the specific required power level to
```
jaller94, line 32:
```suggestion
`M_INSUFFICIENT_POWER` would be when the authenticated user does not have the required power level to
```

"specific" seems like a filler word and does not add any meaning next to "required" in this context.

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
...
+  - [PUT /_matrix/client/v3/rooms/{roomId}/send/{eventType}/{txnId}](https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
+
+
+## Potential issues
+
+Changing long-established error codes in Matrix will be fraught with risk, as many
+clients will need updating to support the new error types. Failure to do so might lead
+to unexpected behaviours or confusing error messages. For this reason, the unstable implementation
+will continue to provide the old errcode in the body of the error while providing the
+new proposed errcode under it's own field. This gives clients a chance to adapt to the
```
jaller94, line 54:
```suggestion
new proposed errcode under its own field. This gives clients a chance to adapt to the
```

proposals/3848-errcode-specific-failure.md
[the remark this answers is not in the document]
tulir: I guess that makes sense, hopefully we remember to use that key

richvdh:
@Half-Shot says this has an implementation and is ready for review

richvdh:
(I haven't checked the implementation)

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
+# MSC3848: Introduce errcodes for specific event sending failures.
```
turt2live, line 1: I'm concerned that this MSC is trying to move faster than safety allows for. While implementations can (and should) move faster than the spec, and the spec should not arbitrarily delay process, this is the sort of MSC which needs time to settle and think about the problem at hand.

Specifically, I'm concerned that this MSC hasn't taken into consideration backwards compatibility of the ecosystem: we have a very long tail of old clients which will not have had a chance to upgrade, and may never will. To help move the spec along faster than the eternal tide of old clients, we'd effectively have to consider this a breaking change. 

Breaking changes to endpoints are relatively easy, but also a bit messy in implementation: we'd bump the endpoint version and declare that the error codes are only valid on the new endpoint version. Thus, clients opt-in to using the error codes. We would also deprecate the now-old endpoints to force clients to consider using the new endpoints. Eventually, we'd remove the old ones.

The complexity for implementations is that clients would need to version check the server (or possibly rely on 404-like semantics), and servers would need to specifically handle their error codes per endpoint version (it would not be legal to return the new error codes on the old endpoint version).

turt2live (changes requested):
per comments, I don't believe this is ready for FCP given how quickly it is trying to move at the moment.

proposals/3848-errcode-specific-failure.md
Half-Shot: The intention isn't to move particularly fast (and doesn't have an urgent project backing it), but while there is spare bandwidth going it felt wise to keep moving on an MSC before it gets forgotten about. That said, happy to have this sit for a bit and have implementations play with it so we can get a wider understanding of whether the MSC works.

I can happily include a passage about requiring v4 (or vNext) endpoints for this errcode to be returned, otherwise servers will continue to respond with `M_FORBIDDEN`.

Wrt implementations, would it be better for us to continue to have an experimental errcode key returned (like the proposal currently suggests) or use a unstable version prefix like `/_matrix/client/org.matrix.msc3848/room/{rId}/invite` which would probably match the reality of what clients have to do?

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
...
+## Potential issues
+
+Changing long-established error codes in Matrix will be fraught with risk, as many
+clients will need updating to support the new error types. Failure to do so might lead
+to unexpected behaviours or confusing error messages. For this reason, the unstable implementation
+will continue to provide the old errcode in the body of the error while providing the
+new proposed errcode under its own field. This gives clients a chance to adapt to the
+new errcode / ensure their behaviours with unexpected errcodes are acceptable.
+
+## Alternatives
```
Half-Shot, line 57: Another alternative is making `errcode` an array of errors, which the client can pick from based on what errcodes it's aware of.

The drawbacks are that this is still a breaking change today, and potentially over-engineering the problem. If we expect errcodes to change infrequently, it might be better to continue to version bump the spec. The other problem is maintaining a potentially long tail of errcodes for older and older generations of the spec.

proposals/3848-errcode-specific-failure.md
turt2live: The MSC would likely claim v4, and if for some reason something else beats it to it then we might consider bundling the two MSCs into one version, or ask that this MSC use v5 or whatever. For now though, it's much simpler to assume v4 will be the thing.

I'd encourage both implementations, I think. Not at the same time mind you, but it can be worth seeing which of the approaches works "better" for clients and what their concerns are. We might also explore alternative error code formats, as suggested in the SCT office room (array of error codes?).

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
...
+Some examples of failures which require more standardized error information
+include already-joined errors when a user tries to invite another user to a room,
+or being unable to send an event into a room due to lacking the required power level.
+
+For this reason, this proposal suggests including new `errcode` definitions
+which provide more specific information about the failure.
+
+## Proposal
+
+New `errcode` would be introduced into the error body of a response
```
richvdh, line 18: this sounds like we're adding a whole new property, which I don't think is the case?

proposals/3848-errcode-specific-failure.md
richvdh: +1 - I don't think we can, in good conscience, change the error codes on existing endpoints. It's frustrating, because this is very much unspecced behaviour that servers *should* be within their rights to change, but our continual failure to spec useful error codes means that the existing error codes become part of the de-facto spec.

I'd be more relaxed about replacing `M_UNKNOWN`. Anyone attaching semantic meaning to that deserves to keep both pieces when it breaks.

So yeah, new endpoint versions here, please, unless you can justify any particular change as being ok.

(OTOH: I *very* much applaud efforts to make the error codes more useful.)

proposals/3848-errcode-specific-failure.md
```diff
@@ -0,0 +1,77 @@
...
+(https://spec.matrix.org/v1.3/client-server-api/#standard-error-response). 
+
+`M_ALREADY_JOINED` would be fired when a membership action fails when the authenticated user
```
richvdh, line 21: could you maybe structure this so that it's clearer which text applies to which error code? Maybe subsection headers?

richvdh (commented):
generally :+1:.

turt2live:
@Half-Shot I believe this has received the sanity review this MSC was looking for and therefore am removing it from the SCT board. If this is false, please raise it in the [SCT office](https://matrix.to/#/#sct-office:matrix.org) for reconsideration.

PR #3531, opened 2021-11-26
Draft: MSCXXXX: Letting moderators hide messages

Yoric:

PR #3849, opened 2022-07-19
MSC3849: Observations and Reinforcement

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/triple/proposals/3849-observations.md)

### DIFF
--- a/proposals/3849-observations.md
+++ b/proposals/3849-observations.md
@@ -46,9 +46,29 @@ Instead we propose sharing observations.
   "content": {
     "entity": "@alice:example.org",
     "observation": "spam",
-    "author": "@alice:example.org",
-    "author_server": "example.org"
-  }
+    "observed_server": "example.org"
+  },
+  "type": "m.policy.observation.user"
+```
+
+We're undecided on whether to propose an observation for an event
+but in cases where the observation is about an action or
+something an author is associated with we propose linking
+to them with an `observed` and `observed_server` field.
+This is so that you can get a general idea of the content
+being created by a user/server without depending on
+direct observations. It also makes auditing and trailing
+a decision a little easier.
+
+
+```
+  "content": {
+    "observation": "spam",
+	"entity": "$w23r0932ir:foo.net"
+    "observed": "@alice:example.org",
+    "observed_server": "example.org"
+  },
+  "type": "m.policy.observation.event"
 ``` 
 
 Observations can be made about any entity in the same way as policy rooms.


proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+
+Instead we propose sharing observations.
+
+```
+  "content": {
+    "entity": "@alice:example.org",
+    "observation": "spam",
+    "observed_server": "example.org"
+  },
+  "type": "m.policy.observation.user"
```
uhoreg, line 51: Just to clarify, these are non-state events, correct?  So far, the only event types defined for policy rooms (`m.policy.rule.*`) are state events, so it may be worth calling this out.

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+agree or disagree with observations. 
+
+Reinforcement can also actually just be done with the same observation
+system, but meh.
+
+
+```
+{
+  "content": {
+    "observation": "$event_id",
```
uhoreg, line 119: Can we use the message relationships/aggregations system here, instead of creating a new key?  Also, does this need to be a state event?

proposals/3849-observations.md
Gnuxie: Sorry, they are state events, It's incomplete because I'm pussyfooting over which entities i want to include in the proposal right now. Users, rooms, servers are a given but i think events/media are certainly blocked on the "directory of abuse" issue. I'll edit them all in with a discussion for each for now.

proposals/3849-observations.md
Gnuxie: Ok, thanks for pointing this out, i totally ignored this detail about relations. The following isn't addressed directly to your question but is my thoughts about the proposal in general.

Yes another problem is that for this to work well there needs to be relationships to events in other rooms.

This is because you will want to make references to labels from other policy lists (if we are keeping with the extensible+hierarchical format) but also so that you can make observations about events. The other thing is that if you do make an observation about another event, you will need a relationship to both the event and the label, which isn't something that is possible or can feasibly will become possible in the future of matrix. 

So then I am forced to choose between the following:

- water down labels into a free form key (which to me sucks because then they can't be hierarchical e.g. i want to trust nico's judgement on all labels under the moderation hierarchy). 
- Ignore matrix's support for relations completely and propose something ad-hoc (just embedding event/room id into each observation to the label definition).
- Specify a set of labels (No way)

- Wait a minute, you could leave the forming of a hierarchy between labels to the client? Yes that actually might work.

uhoreg: Ah, OK, if you're referencing things from other rooms, that might be a reason not to use relations.  Relations are also not generally used for state events.  Anyways, if you decide not to use relations for whatever reason, it's probably saying something about why relations aren't used.

bkil:
I think I have shared this in some PR, but I'd prefer if we had a structured way to declare the given instances (observations) of abuse that reinforce the given moderation action. I.e., in our ban list, I always add the room(s) where the given spammer has acted up recently.

Strictly speaking, an event ID (or timestamp) would also be needed, but for all intents and purposes, the room was sufficient for us with the present rate of ham vs. spam and was much easier to add by hand.

Right now, this mostly works if the abuse happened in at least one abandoned room where it is not redacted, but the future possibility of hiding abuse for a few days before redacting it would provide a better solution to this.

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+This is so that you can get a general idea of the content
+being created by a user/server without depending on
+direct observations. It also makes auditing and trailing
+a decision a little easier.
+
+
+```
+  "content": {
+    "observation": "spam",
+	"entity": "$w23r0932ir:foo.net"
```
Kladki, line 67: Problem is that you often would want to redact these events. I wonder if [MSC3531](https://github.com/matrix-org/matrix-spec-proposals/pull/3531) would allow for this to be fully utilized.

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+    "observation": "spam",
+    "observed_server": "example.org"
+  },
+  "type": "m.policy.observation.user"
+```
+
+We're undecided on whether to propose an observation for an event
+but in cases where the observation is about an action or
+something an author is associated with we propose linking
+to them with an `observed` and `observed_server` field.
```
Kladki, line 57: `observed_server` should just be able to be inferred from the server name of the `observed` user id.

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+It's generic to all kinds of entities and gray. You can make positive
+or negative observations.
+
+The `author` and `author_server` fields are added to make lookup of observations
+about entities created by one user or server easier.
+
+
+### Labels to provide Context (This one is a maybe)
+
+Labels are used to provide context for an observation.
```
Kladki, line 86: Not too sure as to what the purpose of this is, is the event id not context in of itself?

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+Reinforcement can also actually just be done with the same observation
+system, but meh.
+
+
+```
+{
+  "content": {
+    "observation": "$event_id",
+    "policy_room": "$room_id",
+    "reinforcement": "agree",
```
Kladki, line 121: Might be useful to namespace, not too sure.

Kladki (changes requested):

proposals/3849-observations.md
```diff
@@ -0,0 +1,216 @@
...
+In order to warn the wider community about `@evil:example.org`'s behaviour
+we have to recommend they also have to be banned, and existing tools will
+take the recommendation directly as soon as they are aware of a rule.
+
+Instead we propose sharing observations.
+
+```
+  "content": {
+    "entity": "@alice:example.org",
+    "observation": "spam",
```
Kladki, line 48: When handling multiple observations (e.g. racism & spam) from a single user (and event, as below), is it recommended to send multiple of these observation events, or should `observation` be replaced with an array of `observations`?

PR #3851, opened 2022-07-21
MSC3851: Allow custom room presets when creating a room

anoadragon453:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/anoa/custom_room_presets/proposals/3851-custom-room-presets.md)

### DIFF
--- a/proposals/3851-custom-room-presets.md
+++ b/proposals/3851-custom-room-presets.md
@@ -1,0 +1,87 @@
+# MSC3851: Allow custom room presets when creating a room
+
+When a client requests for a room to be created using the [`POST
+/_matrix/client/v3/createRoom`](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3createroom)
+endpoint, they are able to specify a "room preset", which can be one of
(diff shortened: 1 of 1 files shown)


proposals/3851-custom-room-presets.md
```diff
@@ -0,0 +1,87 @@
...
+client implementations.
+
+
+## Proposal
+
+This proposal advocates for the ability of a client to set a custom, namespaced
+string for the `room_preset` field when creating a room. This string identifies
+a desired configuration, such as `com.example.room_preset.dm` which the
+homeserver would map to a set of initial room state. Such mapping could be
+updated immediately without needing to modify client code.
```
HarHarLinks, line 31: how would presets be created? maybe it would be up to the implementation. for use cases I see having an admin define presets would be enough.

however i think there should be an endpoint that lets clients list presets with appropriate metadata so they can present all available presets in their "create room" UI.

for example, working in an organisation that uses "trusted" invite-only rooms and "untrusted" invite-only rooms, it would be nice to have both presets. trusted rooms would by default allow everyone to room ping or manage pinned messages, for example.

PR #3852, opened 2022-07-22
MSC3852: Expose user agent information on `Device`

kerryarchibald:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kerry/user-agent-on-device/proposals/3852-user-agent-on-device.md)

Implementation: https://github.com/matrix-org/synapse/pull/13549
Client implementation: https://github.com/matrix-org/matrix-js-sdk/pull/2728

[FCP Boxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3852#issuecomment-1297157396)

### DIFF
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -10,8 +10,8 @@ sessions. This gives users  more confidence in removing stale or suspicious sess
 ## Proposal
 Homeservers already record the user agent per session today to expose it in the admin API [GET
 /_matrix/client/v3/admin/whois/{userId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid).
-This MSC proposes optionally exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
-Client-Server API endpoints [GET
+This MSC proposes optionally exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model
+returned by Client-Server API endpoints [GET
 /_matrix/client/v3/devices](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices) and [GET
 /_matrix/client/v3/devices/{deviceId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices). If
 the user agent is not available, or the server chooses not to expose it, the value should be omitted from the response. 
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -10,10 +10,15 @@ sessions. This gives users  more confidence in removing stale or suspicious sess
 ## Proposal
 Homeservers already record the user agent per session today to expose it in the admin API [GET
 /_matrix/client/v3/admin/whois/{userId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid).
-This MSC proposes exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
+This MSC proposes optionally exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
 Client-Server API endpoints [GET
 /_matrix/client/v3/devices](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices) and [GET
-/_matrix/client/v3/devices/{deviceId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices).
+/_matrix/client/v3/devices/{deviceId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices). If
+the user agent is not available, or the server chooses not to expose it, the value should be omitted from the response. 
+
+| Name | Type | Description |
+|------|------|-------------|
+| `last_seen_user_agent` | string | **Optional** The latest recorded user agent for the session. 
 
 ```jsonp
 {
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -31,7 +31,35 @@ the user agent is not available, or the server chooses not to expose it, the val
 ```
 
 ## Alternatives
-N/A
+### HTTP client hints
+User agent strings are on their way to being [deprecated.](https://www.chromium.org/updates/ua-reduction/). Instead of
+relying on UA string, the server could use [user agent hints](https://wicg.github.io/ua-client-hints/#http-ua-hints) to
+record equivalent information about sessions.
+
+A server should set an `Accept-CH: Sec-CH-UA, Sec-CH-UA-Mobile, Sec-CH-UA-Platform` header. When `Sec-CH-UA,
+Sec-CH-UA-Mobile, Sec-CH-UA-Platform` headers are present in server requests the values should be saved against the
+session. The latest recorded values should be exposed on the device model:
+
+```jsonp
+{
+  "device_id": "QBUAZIFURK",
+  "display_name": "android",
+  "last_seen_ip": "1.2.3.4",
+  "last_seen_ts": 1474491775024,
+  "platform": "macOS",
+  "isMobile": "false",
+  "clientBrand": "Firefox",
+  "clientVersion": "123"
+}
+```
+
+[Not yet supported](https://caniuse.com/?search=Sec-CH-UA) on Firefox or Safari.
+
+### Explicitly save client information from matrix clients on device model
+Add optional client information fields to the device model, and allow Matrix clients to set these values using existing
+device update APIs. It is up to the client to use user agent, client hints, mobile platform's standard library, etc, or
+to opt out of recording client information. The new fields should be returned as part of the device model (as above). As
+it relies on the client to detect changes in values and update them manually, it is easy for data to get stale.
 
 ## Security considerations
 The user agent is currently only exposed in the admin API and following this MSC would be accessible to normal users.
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -8,7 +8,7 @@ By exposing more detailed and up to date session information, users will be able
 sessions. This gives users  more confidence in removing stale or suspicious sessions.
 
 ## Proposal
-Home servers already record the user agent per session today to expose it in the admin API [GET
+Homeservers already record the user agent per session today to expose it in the admin API [GET
 /_matrix/client/v3/admin/whois/{userId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid).
 This MSC proposes exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
 Client-Server API endpoints [GET
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -8,10 +8,12 @@ By exposing more detailed and up to date session information, users will be able
 sessions. This gives users  more confidence in removing stale or suspicious sessions.
 
 ## Proposal
-Home servers already record the user agent per session today to expose it in the admin API
-[/whois](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid). This MSC proposes exposing
-the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by Client-Server API
-[/devices](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices) endpoints.
+Home servers already record the user agent per session today to expose it in the admin API [GET
+/_matrix/client/v3/admin/whois/{userId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid).
+This MSC proposes exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
+Client-Server API endpoints [GET
+/_matrix/client/v3/devices](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices) and [GET
+/_matrix/client/v3/devices/{deviceId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices).
 
 ```jsonp
 {
--- a/proposals/3852-user-agent-on-device.md
+++ b/proposals/3852-user-agent-on-device.md
@@ -1,4 +1,4 @@
-# MSCXXX: Expose user agent information on `Device`
+# MSC3852: Expose user agent information on `Device`
 
 Currently, sessions are only easily recognisable by their `device_name`. Depending on client implementation, this may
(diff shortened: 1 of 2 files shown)


proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,36 @@
...
+  "last_seen_ip": "1.2.3.4",
+  "last_seen_ts": 1474491775024,
+  "last_seen_user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36"
+}
+```
+
+## Alternatives
+N/A
+
+## Security considerations
```
MTRNord, line 29: Please add that this data is currently public to all users which are in a room with you. So this essentially leaks private data to participants. (which is also why I down voted this as this seems like a) a security concern and b) is difficult with gdpr. Especially the ip which is considered identifyable private information as per gdpr. )

MTRNord: Ah well I guess this is already here. Just differently worded. 🤔

kerryarchibald: This MSC only concerns the [device management endpoints](https://spec.matrix.org/v1.3/client-server-api/#device-management) that only return device information about the currently authenticated user. The amount of room member device information available will not change.

proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,36 @@
...
+The user agent is currently only exposed in the admin API and following this MSC would be accessible to normal users.
```
Half-Shot, line 30: I'd imagine some of these concerns could be alleviated by allowing this to be opt-in, i.e. a query parameter on device creation that states whether the user_agent should be tracked.

I would also make the field optional for homeservers that just don't want to track that information, because I can see contexts where storing that information could be difficult legally.

anoadragon453: Note that as stated in the MSC, homeservers already expose user agent information via [`GET /_matrix/client/v3/admin/whois/{userId}`](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid). There's also nothing stopping homeserver implementations from simply setting these fields to an empty string or `null`.

On the client-side, if the client wishes to not share their user agent with the homeserver, then don't send a user agent with your request.

proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,36 @@
...
+%(osName)s)'`). This information can become stale, and if edited by the user any device detail is lost.
+
+By exposing more detailed and up to date session information, users will be able to more easily recognise their
+sessions. This gives users  more confidence in removing stale or suspicious sessions.
+
+## Proposal
+Home servers already record the user agent per session today to expose it in the admin API
+[/whois](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid). This MSC proposes exposing
+the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by Client-Server API
+[/devices](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3devices) endpoints.
```
anoadragon453, line 14: I think it would help improve clarity for the exact HTTP method + endpoints to be listed out here, instead of pointing to the `/devices` endpoints. As it stands, this MSC refers to `PUT  /_matrix/client/v3/devices/{deviceId}` and `DELETE /_matrix/client/v3/devices/{deviceId}`, which I don't believe is the intention.

proposals/3852-user-agent-on-device.md
anoadragon453: @MTRNord This MSC only exposes further device information to the user that owns the devices. It does not propose exposing user agents to other users.

turt2live: Considering resolved - if this is false, please ping or open a new thread with more information.

proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,38 @@
...
+
+Currently, sessions are only easily recognisable by their `device_name`. Depending on client implementation, this may
+include some stringified information about the session. (For example, Element web uses `'%(appName)s (%(browserName)s,
+%(osName)s)'`). This information can become stale, and if edited by the user any device detail is lost.
+
+By exposing more detailed and up to date session information, users will be able to more easily recognise their
+sessions. This gives users  more confidence in removing stale or suspicious sessions.
+
+## Proposal
+Home servers already record the user agent per session today to expose it in the admin API [GET
```
turt2live, line 11: Nit: it's "homeserver"
```suggestion
Homeservers already record the user agent per session today to expose it in the admin API [GET
```

turt2live, line 11: for the historical record (not that it affects this MSC at all): it's believed to be a mistake that *any* admin endpoints made it into the spec, though this is pending [further discussion](https://github.com/matrix-org/matrix-spec/issues/328).

However, a server should still have realtively easy access to user agent information if it needed to start storing it.

proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,38 @@
...
+/_matrix/client/v3/admin/whois/{userId}](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3adminwhoisuserid).
+This MSC proposes exposing the latest recorded user agent as `last_seen_user_agent` on the `Device` model returned by
```
turt2live, line 13: is it required? (probably shouldn't be, but need to be explicit)

turt2live: and what would the value be if the server didn't have user agent information?

turt2live (commented):
overall the feature looks good to me - once an implementation (client & server) is linked in the PR description I think this would be ready for FCP.

@kerryarchibald apologies for the delay on review here. I have this MSC listed as needing "sanity review", which it has now received. If it needs something else to happen, please raise it in the [SCT office](https://matrix.to/#/#sct-office:matrix.org) for further consideration.

turt2live (approved):
This seems pretty reasonable to move forward with, and as suggested by the team it's ready for FCP - let's do that.

turt2live:
@mscbot fcp merge

proposals/3852-user-agent-on-device.md
```diff
@@ -0,0 +1,43 @@
+# MSC3852: Expose user agent information on `Device`
+
+Currently, sessions are only easily recognisable by their `device_name`. Depending on client implementation, this may
+include some stringified information about the session. (For example, Element web uses `'%(appName)s (%(browserName)s,
+%(osName)s)'`). This information can become stale, and if edited by the user any device detail is lost.
+
+By exposing more detailed and up to date session information, users will be able to more easily recognise their
+sessions. This gives users  more confidence in removing stale or suspicious sessions.
```
dbkr, line 8: I'm probably a little late with this kind of feedback, but I do wonder if user agent strings are a thing we really want to be showing to end users. They've started their journey towards being deprecated anyway in favour of user agent hints (although obviously it will be a long road). Would adding a device name or similar that's just not visible to other users achieve a similar thing?

uhoreg: (I have an MSC in the pipeline to make device names completely hidden from other users.)

Johennes: > Would adding a device name or similar that's just not visible to other users achieve a similar thing?

I believe the problem is that on web the only way to get information about the platform (operating system, browser, browser version) is the user agent. Without using the user agent we couldn't make this information available to clients.

dbkr: Surely any information in the user agent we could equally get with javascript and put into a field in the request?

Johennes: My JS-fu is weak. Would that not still have to use the user agent? I'm not sure I understand how that is better.

anoadragon453: For immediate concerns, https://www.chromium.org/updates/ua-reduction/ details Chrome's current plan to reduce the information shared in a user agent. As of Chrome 107 (which released 12 days ago), the changes in the document have been switched on by default for most platforms. The changes are fairly minor though, and would still be useful: https://www.chromium.org/updates/ua-reduction/#sample-ua-strings-final-reduced-state

[User agent hints](https://wicg.github.io/ua-client-hints/#http-ua-hints) on the other hand, expose structured client information to remote servers through a HTTP header-based system, and [User-Agent Client Hints API](https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API#concepts_and_usage) is a JavaScript API intended to return the same information. Note that what information is returned is broken up into two buckets: "low entropy" and "high entropy". Would we be able to get away with only publishing low entropy data between devices?

So far, [Firefox doesn't support the User Agent Client Hints JavaScript API](https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API#browser_compatibility). Their [position on the subject is "neutral"](https://mozilla.github.io/standards-positions/#ua-client-hints), which may indicate that they'll add support in the future.

To be more future-proof for all of this, perhaps instead of mandating a user agent be published, we define the fields (such as "Brand", "significant version" and "device model") that the majority of use cases could use? Then it'd be up to the client to figure out whether it should get this information from a user agent, or client hints, or a mobile platform's standard library etc.

Something like `GET /_matrix/client/v3/devices` returning:

```json5
{
  "device_id": "QBUAZIFURK",
  "display_name": "android",
  "last_seen_ip": "1.2.3.4",
  "last_seen_ts": 1474491775024,
  "device_information": {
    "brand": "Firefox",
    "version": "106.0",
    "platform": "Linux"
}
```

@mscbot concern Should we rely on User Agent strings to identify devices?

Johennes: Part of the reason we decided to use the user agent was that we wanted to enable a richer device management UI _before_ OIDC arrives. Exposing the user agent (which is already recorded and already exposed in the admin API) seemed like a minimally invasive way to enable that. If there is now concern about adding the user agent into this specific endpoint, I think rather than further complicating it, we could just add the information into the account data events defined in https://github.com/vector-im/element-meta/blob/develop/spec/matrix_client_information.md and leave the spec be for the time being.

dbkr: @anoadragon453 's suggestion is basically what I was thinking: perhaps even as just a text string rather than split out into fields since the client may be able to come up with a better string than (eg. web clients could include the URL they were running at), although of course at the expense of i18n. "Element Web 1.11.1 on Firefox 106.0, Linux" looks a lot better to a user than, "Mozilla/5.0 (X11; Linux i686; rv:107.0) Gecko/20100101 Firefox/106.0". The disadvantage is the the client would probably have to set this is a separate call occasionally, whereas the user agent string is sent with every request already.

Potential compromise option: call it `device_description` instead, leave scope for it being set explicitly by the client later and for now, leave it up to the HS what it wants to put there, which could either be the raw user agent or a human-readable English (probably just English) string made by very basic parsing of the user agent string?

Johennes: > "Element Web 1.11.1 on Firefox 106.0, Linux" looks a lot better to a user than, "Mozilla/5.0 (X11; Linux i686; rv:107.0) Gecko/20100101 Firefox/106.0".

This probably wasn't clear from the proposal but the primary intention was not showing the user agent itself to users. Rather the idea was for clients to parse bits out of the user agent and display them in a structured way. So if we make this just a single `device_description` string, would the constant assembly / disassembly potentially be a little inconvenient?

anoadragon453: I'm also not a fan of a free-form text field. We already have device names for that.

dbkr: Isn't the point here that device names are basically what we want, but we just want an equivalent that's only visible to the user rather than everyone? Either way, if we'd rather have this in structured data then that's great too.

I hadn't realised the intention was to parse the UA on the receiving clients. In which case, isn't that even more of an argument for having a field we can start by populating based on UA parsing and then move to data explicitly provided by the client later? Otherwise we're stuck making every client parse user agents forever.

Johennes: > Isn't the point here that device names are basically what we want, but we just want an equivalent that's only visible to the user rather than everyone?

I think in principle yes but lumping all the information into a single string isn't very user friendly. Element's new session manager that currently builds on this MSC splits the values out into a table with descriptive labels.

dbkr: This makes @anoadragon453's suggestion the obvious choice then, I think?

Johennes: What about the [concern](https://github.com/matrix-org/matrix-spec-proposals/pull/3852#discussion_r1017057889) of OIDC rendering all of this redundant? My understanding is that in OIDC the authentication server will offer device management?

dbkr: It seems the plan is indeed to move all the session/device management to the IdP web UI and out of the clients. In which case, isn't this all redundant and if you need a temporary hack, you could just add a vendor namespaced field?

Johennes: I guess we did not consider exposing the user agent a hack. Hence, we took the effort to write this MSC separately from the unspec'ed account data storage mentioned above. If the SCT does consider this a hack, then I agree that keeping this vendor specific is better than making larger scale changes to the existing API – which I already expressed [here](https://github.com/matrix-org/matrix-spec-proposals/pull/3852#discussion_r1017057889).

dbkr: To be clear, I can't speak on behalf of the entire SCT. I don't think it's that bad, just worth considering other options (and the alternatives section is blank, after all).

kerryarchibald: Have added two options to the alternatives section, please let me know if I've transcribed them accurately.

anoadragon453: Thanks for adding those @kerryarchibald. It spells out the problem and potential solutions concisely.

Note that if folks are particularly taken with either alternative approach, I'd recommend they create a new MSC and mention that it is an alternative approach to this one.

---

As for this MSC, I remain concerned with having a field explicitly for user agent. Perhaps a third alternative could be a mix of the two presented already: we have separate fields for `platform`, `clientBrand` and `clientVersion`, but they are set by the homeserver using HTTP client hints, or the client's latest user agent as a fallback.

That would eliminate the concern of clients not keeping the information up to date, while additionally not enshrining user agents into the spec.

turt2live: @kerryarchibald are you able to take a look at this thread? It appears to be the blocker for this MSC entering FCP at the moment.

kerryarchibald:
Given that the device management UI that this MSC enables will be moving to OIDC providers in the short-to-medium term, I don't think expanding the device model with specific fields would be a constructive addition to the spec. This MSC should be closed.

turt2live:
@mscbot fcp cancel

PR #3854, opened 2022-07-25
Specify how servers detect support for MSC3827

SimonBrandner:
For https://github.com/matrix-org/matrix-spec-proposals/pull/3827


turt2live (approved):
This feels a bit complicated for the feature at hand, but should be fine.

Realistically it can just send both without version detection.

PR #2836, opened 2020-10-28
MSC2836: Threading

kegsay:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/kegan/msc/threading/proposals/2836-threading.md)

Implementation: https://cerulean.matrix.org/
Server: https://github.com/matrix-org/dendrite/tree/master/setup/mscs/msc2836
Client: https://github.com/matrix-org/cerulean/

PR #3856, opened 2022-07-26
MSC3856: Threads List API

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/threads-list-api/proposals/3856-threads-list-api.md)

----

Implementations:

* Synapse: matrix-org/synapse#13394
* matrix-js-sdk: matrix-org/matrix-js-sdk#2602

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3856#issuecomment-1238553130)

### DIFF
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -111,10 +111,18 @@ Query Parameters:
   A token corresponding to the start of `chunk`. This will be the same as the
   value given in `from`.
 
-XXX Define how ignored users are handled, which has two cases:
+If the sender of an event is ignored by the current user the results are modified
+slightly. This has two situations:
 
-1. If the ignored user sent the root thread event.
-2. If the ignored user sent the latest thread event.
+1. If the ignored user sent the root thread event: the server should return the
+   redacted form of the root event, but otherwise act as normal. This matches the
+   information that a client would have if the threads list was aggregated locally
+   (and generally matches the behavior if a thread root is unavailable, e.g. due
+   to room history visibility).
+2. If the ignored user send the latest thread event: the server should treat the
+   latest event as not existing and replace it with the latest event from a
+   non-ignored user; with the caveat that the ordering of the threads is not
+   re-arranged due to this replacement.
 
 #### Example request:
 
@@ -163,7 +171,7 @@ server-side maximum.
 It does not seem useful to be able to paginate in reverse order, i.e. starting with
 the thread which was least recently updated. If there becomes a future need of this
 a `dir` parameter could be added which takes an enum value of `[f b]` defaulting to
-`b` to maintain backwards compatibility with this proposal. 
+`b` to maintain backwards compatibility with this proposal.
 
 ## Unstable prefix
 
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -50,15 +50,13 @@ GET /_matrix/client/v1/rooms/{roomId}/threads
 
 Query Parameters:
 
-* **`filter`**: `enum`
+* **`include`**: `enum`
 
   Whether to include all threads in the room or only threads which the user has
   participated in, meaning that the user has created the root event of the thread
   or have created an event with a `m.thread` relation targeting the root.
 
   One of `[all participated]`. Defaults to `all`.
-
-  XXX Rename this to something besides `filter`.
 * **`from`**: `string`
 
   The token to start returning events from. This token can be obtained from a
@@ -68,7 +66,7 @@ Query Parameters:
   If it is not provided, the homeserver shall return a list of threads from the
   last visible event in the room history for the requesting user.
 * **`limit`**: Optional: a client-defined limit to the maximum
-  number of rooms to return per page. Must an integer greater than zero.
+  number of rooms to return per page. Must be an integer greater than zero.
 
   Server implementations should impose a maximum value to avoid resource
   exhaustion.
@@ -107,7 +105,7 @@ XXX Define how ignored users are handled, which has two cases:
 ```
 GET /_matrix/client/v1/rooms/%21ol19s%3Ableecker.street/threads?
     limit=25&
-    filter=participated
+    include=participated
 ```
 
 #### Example response:
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -22,7 +22,7 @@ paginate through all of the threads in the room, inspect the latest event from
 the bundled aggregations and attempt to sort them. This can require many round
 trips to the server and is wasteful for both the client and server.
 
-Unfortunately even when a client has all the threads in a room is not able to accurately
+Unfortunately even when a client has all the threads in a room it is not able to accurately
 sort the threads since the client lacks the proper topological ordering of events. (The
 closest available information is the `age` or `origin_server_ts` of the events, but this
 is not always correct.)
@@ -160,7 +160,7 @@ a generic `/event_relationships` endpoint, but it is overly complex for
 MSC2836 attempts to solve a larger problem, including allowing for arbitrary
 branching of threads (and many levels of event relations). MSC3440 forbids creating
 threads off a threaded message, allowing for a simpler design. Additionally, the
-MSC2836 design is more computensively intensive for both clients and servers to
+MSC2836 design is more computationally intensive for both clients and servers to
 implement due to the tree-like nature of the query.
 
 A benefit to the MSC2836 design is that it supports querying over federation for
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -7,10 +7,12 @@ issues:
 2. Clients wish to display a list of threads the user is interested in.
 
 It is currently difficult for clients to sort threads by the most recently
-responded to. Clients can use the [`/messages`](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmessages) API with a filter of
-`"related_by_rel_types": ["m.thread"]` to fetch the list of threads in a room. This
-returns the root thread events in topological order of those events (either
-forwards or backwards depending on the `dir` parameter).
+responded to. Clients can use the [`/messages`](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmessages)
+API with a filter of `"related_by_rel_types": ["m.thread"]` (as defined in
+[MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#fetch-all-threads-in-a-room))
+to fetch the list of threads in a room. This returns the root thread events in
+topological order of those events (either forwards or backwards depending on the
+`dir` parameter).
 
 Each event also includes bundled aggregation, which will include the latest
 event in each thread.
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -72,7 +72,7 @@ Query Parameters:
   the thread or replied to the thread (they have created an event with a `m.thread`
   relation targeting the root event).
 
-  One of `[all participated]`. Defaults to `all`.
+  One of `[all, participated]`. Defaults to `all`.
 * **`from`**: `string`
 
   The token to start returning events from. This token can be obtained from a
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -1,10 +1,10 @@
 # MSC3856: Threads List API
 
 An endpoint specific to listing the threads in a room is proposed to solve two
-issues:
+client problems:
 
-1. Clients wish to display threads by the most recently active.
-2. Clients wish to display a list of threads the user is interested in.
+1. Clients wish to display threads ordered by the most recently active.
+2. Clients wish to display a list of threads the user has participated in.
 
 It is currently difficult for clients to sort threads by the most recently
 responded to. Clients can use the [`/messages`](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmessages)
@@ -22,16 +22,25 @@ paginate through all of the threads in the room, inspect the latest event from
 the bundled aggregations and attempt to sort them. This can require many round
 trips to the server and is wasteful for both the client and server.
 
-Additionally, even with all of the threads a client is not able to accurately
-sort the threads since they lack the proper topological ordering of events. (The
-closest they can get is by using `age` or `origin_server_ts`, but this is not
-correct.)
+Unfortunately even when a client has all the threads in a room is not able to accurately
+sort the threads since the client lacks the proper topological ordering of events. (The
+closest available information is the `age` or `origin_server_ts` of the events, but this
+is not always correct.)
 
-For the second problem, it is currently not possible for a client to query for
-threads that the user has participated in (as defined in
-[MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#event-format)).
-A client could add the user's MXID to the filter, e.g. `"related_by_senders":["@alice:example.com"]`,
-but this misses threads where the user sent the root message and has not yet replied.
+Additionally, it is currently not possible for a client to query for threads that
+the user has participated in, as defined in
+[MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#event-format):
+
+> The user has participated if:
+>
+> * They created the current event.
+> * They created an event with a m.thread relation targeting the current event.
+
+Currently, clients add the requesting user's MXID to the `related_by_senders` filter
+(as defined in
+[MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#fetch-all-threads-in-a-room)),
+e.g. `"related_by_senders":["@alice:example.com"]`, but this results in missing
+threads where the user sent the root message and has not yet replied.
 
 ## Proposal
 
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -158,6 +158,13 @@ filtering in the future.
 As with other endpoints that accept a `limit`, homeservers should apply a hard
 server-side maximum.
 
+## Future extensions
+
+It does not seem useful to be able to paginate in reverse order, i.e. starting with
+the thread which was least recently updated. If there becomes a future need of this
+a `dir` parameter could be added which takes an enum value of `[f b]` defaulting to
+`b` to maintain backwards compatibility with this proposal. 
+
 ## Unstable prefix
 
 The client-server API will be: `/_matrix/client/unstable/org.matrix.msc3856/rooms/{roomId}/threads`
--- a/proposals/3856-threads-list-api.md
+++ b/proposals/3856-threads-list-api.md
@@ -80,7 +80,7 @@ Query Parameters:
   an `end` token returned by a previous request to this endpoint.
 
   If it is not provided, the homeserver shall return a list of thread roots starting
-  from the last visible event in the room history for the requesting user.
+  from the most recent visible event in the room history for the requesting user.
 * **`limit`**: Optional: a client-defined limit to the maximum
   number of threads to return per page. Must be an integer greater than zero.
 
@@ -119,7 +119,7 @@ slightly. This has two situations:
    information that a client would have if the threads list was aggregated locally
    (and generally matches the behavior if a thread root is unavailable, e.g. due
    to room history visibility).
-2. If the ignored user send the latest thread event: the server should treat the
+2. If the ignored user sent the latest thread event: the server should treat the
    latest event as not existing and replace it with the latest event from a
    non-ignored user; with the caveat that the ordering of the threads is not
    re-arranged due to this replacement.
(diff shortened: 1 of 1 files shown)


proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+  events), this property is omitted from the response.
+* **`start`**: `string` **Required**
+
+  A token corresponding to the start of `chunk`. This will be the same as the
+  value given in `from`.
+
+XXX Define how ignored users are handled, which has two cases:
+
+1. If the ignored user sent the root thread event.
+2. If the ignored user sent the latest thread event.
```
clokep, line 101: Since this is replacing existing functionality, I believe the behavior of the `/messages` API as exists now would be:

1. If the ignored user sent the root thread event then the thread would be completely filtered out.
2. If the ignored user sent the latest thread event, a replacement is attempted to be found.

----

1) feels pretty weird, but I don't know what other behavior to give?

2) seems OK, but there's weird edge-cases -- what happens if all events in the thread are from ignored users; is this no longer a thread?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+paginate through all of the threads in the room, inspect the latest event from
+the bundled aggregations and attempt to sort them. This can require many round
+trips to the server and is wasteful for both the client and server.
+
+Additionally, even with all of the threads a client is not able to accurately
+sort the threads since they lack the proper topological ordering of events. (The
+closest they can get is by using `age` or `origin_server_ts`, but this is not
+correct.)
+
+For the second problem, it is currently not possible for a client to query for
```
richvdh, line 28: I'm not quite sure what "second problem" we are referring to here?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
+# MSC3856: Threads List API
+
+An endpoint specific to listing the threads in a room is proposed to solve two
+issues:
+
+1. Clients wish to display threads by the most recently active.
+2. Clients wish to display a list of threads the user is interested in.
+
+It is currently difficult for clients to sort threads by the most recently
+responded to. Clients use the `/messages` API with a filter of
```
richvdh, line 10:
```suggestion
responded to. Clients can use the [`/messages`](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmessages) API with a filter of
```

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+`"related_by_rel_types": ["m.thread"]` to fetch the list of threads in a room. This
```
richvdh, line 11: `related_by_rel_types` doesn't seem to be in the spec yet. Can you link to the relevant MSC?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+* **`from`**: `string`
+
+  The token to start returning events from. This token can be obtained from a
+  `prev_batch` or `next_batch` token returned by the `/sync` endpoint, or from
+  an `end` token returned by a previous request to this endpoint.
+
+  If it is not provided, the homeserver shall return a list of threads from the
+  last visible event in the room history for the requesting user.
+* **`limit`**: Optional: a client-defined limit to the maximum
+  number of rooms to return per page. Must be an integer greater than zero.
```
richvdh, line 69:
```suggestion
  number of threads to return per page. Must be an integer greater than zero.
```

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+
+  One of `[all participated]`. Defaults to `all`.
+* **`from`**: `string`
+
+  The token to start returning events from. This token can be obtained from a
+  `prev_batch` or `next_batch` token returned by the `/sync` endpoint, or from
+  an `end` token returned by a previous request to this endpoint.
+
+  If it is not provided, the homeserver shall return a list of threads from the
+  last visible event in the room history for the requesting user.
```
richvdh, line 67: this doesn't quite read like what you mean, I think... it sounds like it will only return threads whose root is that one event?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,151 @@
...
+Query Parameters:
+
+* **`include`**: `enum`
+
+  Whether to include all threads in the room or only threads which the user has
+  participated in, meaning that the user has created the root event of the thread
+  or have created an event with a `m.thread` relation targeting the root.
+
+  One of `[all participated]`. Defaults to `all`.
+* **`from`**: `string`
```
richvdh, line 60: loath as I am to broach the pagination subject again, do we need a `dir` param?

proposals/3856-threads-list-api.md
richvdh: on 1: I *think* that if I ignore a user, but the rest of the room starts a thread based on an event the bad user sent, I still want to see the thread (ideally without the root event). After all, the thread might go on a long time without the involvement of the bad user.

We might need to think of some way of flagging to the client that a given event came from an ignored user, so should be ignored?

richvdh (commented):
A few comments, but generally looks sensible to me. If you could clarify a couple of things we can probably put this up for FCP.

proposals/3856-threads-list-api.md
clokep: > on 1: I _think_ that if I ignore a user, but the rest of the room starts a thread based on an event the bad user sent, I still want to see the thread (ideally without the root event). After all, the thread might go on a long time without the involvement of the bad user.

I think this makes sense -- it matches what the client would see if they aggregated the thread locally (or for some other reason couldn't see the initial event).

proposals/3856-threads-list-api.md
clokep: The two problem statements at the top of the MSC, the second of which is:

> Clients wish to display a list of threads the user is interested in.

The bits above this all refer to the first problem. Do you have suggestions of how to make this clearer?

proposals/3856-threads-list-api.md
clokep: Will do, it is in MSC3440.

clokep: (Note this is linked below where we go into more detail about it.)

proposals/3856-threads-list-api.md
clokep: > it sounds like it will only return threads whose root is that one event?

That doesn't make sense -- a event can only be the root of a single thread.

Would it be clearer if this entire section talked about "events", which is what is actually returned?

proposals/3856-threads-list-api.md
clokep: A `dir=f` would be equivalent to "get me the oldest threads in the room". I don't think there's a use-case for that at the moment?

@gsouquet can you confirm this is accurate?

clokep: (Note that "oldest threads" means "threads which have had replies the longest time ago".)

clokep: There's currently no plan to need a `dir` parameter on here (there doesn't seem to be any valid use-cases).

proposals/3856-threads-list-api.md
gsouquet: Totally agree with @clokep , this sounds like a scenario where we should act similarly to what clients do when the initial event is unreachable (for example the initial event was posted before I was joined a room and the history visibility setting does not allow me to grab that event). That behaviour seems sane

> We might need to think of some way of flagging to the client that a given event came from an ignored user, so should be ignored?

That is an interesting thought for sure, but as of today clients receive absolutely 0 infomations from ignored users. All events are erased from sync and scrollback. That would mean potentially creating a new event type with redacted content... which could be useful to extend to all events from ignored users so that client can decide to show a trail of events from an ignored individual?

proposals/3856-threads-list-api.md
clokep: Hopefully 8ef8232fc4c07ab03e38b9e98c2ab07a0b368159 clarifies this.

proposals/3856-threads-list-api.md
clokep: I tried to clarify that the returned values are the thread roots -- 75afe347d23c2358187a1966103516d29d5afa20

proposals/3856-threads-list-api.md
clokep: I added some info at dd11005295a82acf80ca69081e8e61142c5062ff that I think is the conclusion from above!

I did not include anything about communicating and ignored event to the client -- this sounds like an MSC by itself!

proposals/3856-threads-list-api.md
clokep: I added this as a "future extension".

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
+# MSC3856: Threads List API
```
kegsay, line 1: How does this work over federation as the local server may not have all the thread roots?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+A new endpoint is proposed to query for threads in a room. This endpoint requires
+authentication and is subject to rate-limiting.
+
+The endpoint returns events, which represent thread roots and includes
+[bundled aggregations](https://spec.matrix.org/v1.3/client-server-api/#aggregations)
+in the response (which includes the "latest event" of each thread, see
+[MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#event-format)
+for the format of bundled aggregations of threads).
+
+The returned events are ordered by the latest event of each thread.
```
kegsay, line 58: Latest by what metric? `origin_server_ts`? Arrival time on the server? Topological ordering?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+
+  One of `[all participated]`. Defaults to `all`.
+* **`from`**: `string`
+
+  The token to start returning events from. This token can be obtained from a
+  `prev_batch` or `next_batch` token returned by the `/sync` endpoint, or from
+  an `end` token returned by a previous request to this endpoint.
+
+  If it is not provided, the homeserver shall return a list of thread roots starting
+  from the last visible event in the room history for the requesting user.
```
kegsay, line 83: I don't think I understand why `from` is a thing. If I want _all_ thread roots for a room, and I don't specify `from` then it defaults to "a list of thread roots starting from the last visible event in the room history for the requesting user" which feels like the opposite of what I want. Assuming I have say the last 50 events in the room, I can slap on `from=prev_batch_from_sync` but that also will presumably just return the thread roots _from_ that position _to_ the latest current position?

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+* **`limit`**: Optional: a client-defined limit to the maximum
+  number of threads to return per page. Must be an integer greater than zero.
+
+  Server implementations should impose a maximum value to avoid resource
+  exhaustion.
+* **`to`**: `string`
+
+  The token to stop returning events at. This token can be obtained from a
+  `prev_batch` or `next_batch` token returned by the `/sync` endpoint, or from
+  an `end` token returned by a previous request to this endpoint.
```
kegsay, line 93: Why would you ever want this?

proposals/3856-threads-list-api.md
clokep: It does not change any behavior there -- a server could backfill to fetch additional thread roots if it wanted to, but that seems rather implementation specific.

proposals/3856-threads-list-api.md
clokep: This is meant to be rather vague and it is further defined below under `chunk`, although that doesn't fully answer the question either:

> The order is chronological by the latest event in that thread.

Note that this is mostly copied from the `chunk` definition of `/messages`, which includes the following text:

> (The exact definition of `chronological` is dependent on the server implementation.)

proposals/3856-threads-list-api.md
clokep: This is here mostly for consistency with `/messages` (and it generally seems to be how pagination of all the APIs in Matrix are done).

gsouquet:
Proof of client side implementation for this proposal, https://github.com/matrix-org/matrix-js-sdk/pull/2602

turt2live:
The people involved here believe this is ready for FCP. Implementations have not been checked yet though: if suitable, the label will be removed and review continue. If not suitable, comments will be left.

proposals/3856-threads-list-api.md
turt2live, line 83: because "last" is confusing with pagination:
```suggestion
  from the most recent visible event in the room history for the requesting user.
```

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+
+If the sender of an event is ignored by the current user the results are modified
+slightly. This has two situations:
+
+1. If the ignored user sent the root thread event: the server should return the
+   redacted form of the root event, but otherwise act as normal. This matches the
+   information that a client would have if the threads list was aggregated locally
+   (and generally matches the behavior if a thread root is unavailable, e.g. due
+   to room history visibility).
+2. If the ignored user send the latest thread event: the server should treat the
```
turt2live, line 122:
```suggestion
2. If the ignored user sent the latest thread event: the server should treat the
```

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+  "end": "...",
+  "start": "..."
+}
+```
+
+### MSC3440 Filtering
+
+This MSC deprecates the [event filters added in MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#fetch-all-threads-in-a-room)
+(`related_by_rel_types` and `related_by_senders`) as the only known use-case is
+more efficiently solved by this MSC.
```
turt2live, line 149: We can also outright remove it, if preferred. 
```suggestion
This MSC replaces the [event filters added in MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#fetch-all-threads-in-a-room)
(`related_by_rel_types` and `related_by_senders`) as the only known use-case is
more efficiently solved by this MSC.
```

as a "stable but not written" thing, it's entirely valid to rip out the functionality without having to go through a whole process of MSCs.

turt2live (approved):
Generally seems fine to me, though I'm not looking forward to specifying another pagination API 🙃 

That's a future problem, though.

turt2live:
@mscbot fcp merge

proposals/3856-threads-list-api.md
clokep: That would be desired, I think. 👍

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,182 @@
...
+
+This MSC replaces the [event filters added in MSC3440](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#fetch-all-threads-in-a-room)
+(`related_by_rel_types` and `related_by_senders`) as the only known use-case is
+more efficiently solved by this MSC.
+
+## Potential issues
+
+None forseen.
+
+## Alternatives
```
ara4n, line 155: It might be useful to list https://github.com/matrix-org/matrix-spec-proposals/pull/2836 as an alternative to this, in terms of having all the thread navigation/listing activity on a single /relations endpoint outside of /messages or /threads.

ara4n:
lgtm modulo wanting to list #2836 as an alt.

proposals/3856-threads-list-api.md
clokep: I added something about 6002d81ce86922f8f219ff8f7bba19d8cd14de6d, hopefully I hit the highlights, but the tl;dr was "that API is much more complicated than threads needs".

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,199 @@
...
+Query Parameters:
+
+* **`include`**: `enum`
+
+  Whether to include all thread roots in the room or only thread roots which the
+  user has participated in, meaning that the user has created the root event of
+  the thread or replied to the thread (they have created an event with a `m.thread`
+  relation targeting the root event).
+
+  One of `[all participated]`. Defaults to `all`.
```
erikjohnston, line 75:
```suggestion
  One of `[all, participated]`. Defaults to `all`.
```

clokep: I don't have a strong opinion here, but we seem to usually not include commas in the spec:

https://spec.matrix.org/v1.2/client-server-api/#mroomhistory_visibility

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,199 @@
...
+* **`from`**: `string`
+
+  The token to start returning events from. This token can be obtained from a
+  `prev_batch` or `next_batch` token returned by the `/sync` endpoint, or from
+  an `end` token returned by a previous request to this endpoint.
```
erikjohnston, line 80: Is this actually something useful? We do it in `/messages` as we need to stitch the timelines together, but I don't think we ever really want to do that? Or are we expecting clients to keep the list of threads updated as things come down sync?

proposals/3856-threads-list-api.md
erikjohnston: well that just makes me sad, but I'll live

proposals/3856-threads-list-api.md
justjanne: We need to at least have the ability to paginate the threads list if it is more than one page long. That means we need to be able to specify a token to only return threads that are older than that.

This token doesn't necessarily have to be the same type of token as the ones used by other APIs, and for our current client implementation so far we only need to paginate in one direction, though.

I haven't looked into updating the threads list based on what we get through /sync yet, so I can't tell if that's planned or not. I believe we've done that before already, though, and it would be unlikely that we'd want to remove that functionality.

erikjohnston: Yup, agreed we need pagination tokens. Just wandering if they need to be specced to be compatible with the sync tokens or not

proposals/3856-threads-list-api.md
turt2live: I'd count that a spec bug tbh

proposals/3856-threads-list-api.md
gsouquet: I can't think of a use case where we would need this to have compatibility with the sync tokens at this time

proposals/3856-threads-list-api.md
clokep: I can file something (or fix it) -- which way do we prefer?

clokep: matrix-org/matrix-spec#1240 attempts to fix this.

proposals/3856-threads-list-api.md
clokep: I simplified this quite a bit in 78ee0e0d297d1ed6d30568aa5ae90a036dcc8628

@gsouquet  / @justjanne I think this would still be reasonably usable by a client, but wanted to double check.

proposals/3856-threads-list-api.md
clokep: See https://github.com/matrix-org/matrix-spec-proposals/pull/3856#discussion_r969612505

proposals/3856-threads-list-api.md
clokep: See https://github.com/matrix-org/matrix-spec-proposals/pull/3856#discussion_r969612505

KitsuneRal (changes requested):
Mostly LGTM but I have a concern on the way messages from ignored users are handled.

KitsuneRal:
@mscbot concern should we care about further updates to the threads list on the client side? Notably wrt messages from ignored users

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Security considerations
+
+As with other endpoints that accept a `limit`, homeservers should apply a hard
+server-side maximum.
+
+## Future extensions
+
+It does not seem useful to be able to paginate in reverse order, i.e. starting with
+the thread which was least recently updated. If there becomes a future need of this
```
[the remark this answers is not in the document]
clokep: That's a good point, perhaps on mobile this would be quite useful.

proposals/3856-threads-list-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+1. If the ignored user sent the root thread event: the server should return the
+   redacted form of the root event, but otherwise act as normal. This matches the
+   information that a client would have if the threads list was aggregated locally
+   (and generally matches the behavior if a thread root is unavailable, e.g. due
+   to room history visibility).
+2. If the ignored user sent the latest thread event: the server should treat the
+   latest event as not existing and replace it with the latest event from a
+   non-ignored user; with the caveat that the ordering of the threads is not
+   re-arranged due to this replacement.
```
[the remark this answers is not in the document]
clokep: I don't think it would be easy for clients to match.

Unfortunately having the threads appear in the proper spot in pagination was quite difficult (unless you're constantly re-evaluating ignored users). It *might* be more doable now that we're not using `/sync` tokens though. I'll take another look and see if it is doable.

KitsuneRal:
@mscbot resolve should we care about further updates to the threads list on the client side? Notably wrt messages from ignored users

KitsuneRal:
The concern is formally resolved in order to let the FCP start and the threads get into the spec within this cycle, with a caveat that we will either fix the logic for ignored events or document the discrepancy I described in the comment as a known issue (that will be addressed in a separate MSC).

proposals/3856-threads-list-api.md
gsouquet: @clokep that looks reasonable to me yes! I don't foresee problems with the changes you've made 👍

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1254

turt2live:
Merged 🎉

PR #3857, opened 2022-08-02
Welcome messages

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/welcome-message/proposals/3857-welcome-message.md)

### DIFF
--- a/proposals/3857-welcome-message.md
+++ b/proposals/3857-welcome-message.md
@@ -1,0 +1,108 @@
+# MSC3857: Welcome messages/screening
+
+Many communities are interested in ways to show a README-style message to new users, giving further
+context for how the community works, what sort of rules there may be, etc. Discord takes this a step
+further by [requiring users to accept](https://support.discord.com/hc/en-us/articles/1538570466882-Rules-Screening-FAQ)
+the dialog - while it is *possible* for Matrix to enforce this, this MSC deliberately leaves this as
+an implementation detail or for a future MSC to consider. Some ideas are covered in the Future
+Considerations section of this proposal, however.
+
+## Proposal
+
+We introduce a new state event, `m.room.welcome`, which uses [MSC1767 - Extensible Events](https://github.com/matrix-org/matrix-spec-proposals/pull/1767)
+to create a renderable `m.message`-like event, similar to [MSC3765 - Rich text topics](https://github.com/matrix-org/matrix-spec-proposals/pull/3765).
+
+Such an event would look like:
+
+```json5
+{
+  "type": "m.room.welcome",
+  "content": {
+    "m.room.welcome": {
+      // we use an object to support future extensions like a theoretical "must_accept: true" flag
+
+      "m.message": [ // exactly what it looks like: an m.message schema goes here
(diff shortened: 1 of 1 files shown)


proposals/3857-welcome-message.md
```diff
@@ -0,0 +1,110 @@
...
+In a later MSC, it would be desirable to support a way to prevent the user from interacting with the room
+until they've accepted/agreed to the welcome message, akin to Discord's Rules Screening feature mentioned
+in the introduction.
+
+This could be achieved by adding the user's acceptance to their membership event (or as a new event), linking
+acceptance into the event authorization rules. This would enforce at a federation level that the acceptance
+must be given in order to send messages, which avoid polluting power levels (promoting a user from -1 to 0
+would mean a potentially lengthy list of users in a power levels event).
+
+A consideration for that future MSC is whether restricting users in this way is effective: unlike Discord,
```
FSG-Cat, line 83: I would like to point out that this section is not correct. If we assume that we are doing this for anti spam we have to throw out any idea of that the user is not willing to violate the ToS out of the window. A client mod can automatically accept this check and so can a spam bot. Screen scraping if you want to be extremely primitive in your attack can do this or plain old AHK macros. 

Yes for matrix due to having a fully open API and a complete expectation of third party clients we are more vulnerable to this as an attack method but well Discord is very vulnerable too. Its actually more effective as an anti spam method to lock your guild behind a react check done correctly than using the built in check if your attackers are just going in blind. If your being targeted specifically then well the checks have to be more complicated. 

Point with all this is that No. This feature does not provide a proper barier to serious spammers on discord other than maby slowing them down a bit but that would be it. The section about matrix is something i do agree about since it makes it clear we cant trust this to protect us.

proposals/3857-welcome-message.md
```diff
@@ -0,0 +1,110 @@
...
+Having lack of overly-strict rules for when to show a welcome message could lead to inconsistencies
+between clients interacting with the same room, as could clients not implementing this feature at all.
+This is considered low risk for the time being until a future MSC handles considerations for preventing
+interaction before acceptance of the dialog.
+
+## Alternatives
+
+There's a thousand different ways to represent a README/welcome/rules/topic message - Extensible Events
+seems the most reasonable given it solves this sort of problem and re-uses the message rendering functions
+clients would already have.
```
deepbluev7, line 65: Another method that reuses normal message rendering is pinned messages, which many clients already use, but Element can't display (by default).

turt2live: Pinned messages have a different use-case than welcome messages.

proposals/3857-welcome-message.md
turt2live: Discord is less vulnerable than Matrix in this respect, as Discord does consider this an anti-spam barrier. This section is not implying it solves spam, just that it might not be valued high enough to be worth considering an anti-spam measure in Matrix.

FSG-Cat: My point is that Discord is simply wrong in considering this an anti spam barrier unless they are doing Server side magic to use this as a signal to determine if their anti spam system should kick in and sanction the user.

PR #3859, opened 2022-08-04
MSC3859: Add well known media domain proposal

Fizzadar:
[Rendered](https://github.com/Fizzadar/matrix-spec-proposals/blob/media-domain/proposals/3859-well-known-media-domain.md)

Signed off by Nick @ Beeper (@fizzadar) <nick@beeper.com>.

UPDATE (2022-08-18): we are closing this MSC in favour of [MSC3860](https://github.com/matrix-org/matrix-spec-proposals/pull/3860) & [MSC3870](https://github.com/matrix-org/matrix-spec-proposals/pull/3870).


proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,38 @@
+# MSCXXX: .well-known Media Domain
+
+Currently Matrix clients and servers discover homeservers and identify servers via the `.well-known/matrix/client` and `.well-known/matrix/server` endpoints. This is currently limited to the homeserver and for clients the identity server. This MSC proposes adding an additional field that can be used to specify a media server that can optionally be used for `/_matrix/media/*` endpoints.
```
turt2live, line 3: Please wrap to ~120 characters

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,38 @@
+# MSCXXX: .well-known Media Domain
```
turt2live, line 1: Your MSC number is 3859.

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,38 @@
...
+
+Currently Matrix clients and servers discover homeservers and identify servers via the `.well-known/matrix/client` and `.well-known/matrix/server` endpoints. This is currently limited to the homeserver and for clients the identity server. This MSC proposes adding an additional field that can be used to specify a media server that can optionally be used for `/_matrix/media/*` endpoints.
+
+The reasoning behind this change is to support hosting media behind a CDN designed and optimised specifically for serving media. Since media transfers and endpoints are very different to the other JSON based endpoints a separate domain offers maximum flexibility for server owners.
```
turt2live, line 5: Can we get a worked example of this in the proposal? It's not clear how a CDN impacts this (I, for instance, already have my media repo backed by a CDN but just haven't turned on the flags to route all media through that CDN)

turt2live (changes requested):

turt2live:
For sign off: We will need an email address, or the form completed with dco@matrix.org

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,38 @@
...
+This is the same as above but for the server endpoint:
+
+```
+{
+  "m.server": "matrix.example.com:443",
+  "m.media_server": "matrix-media.example.com:443"
+}
+```
+
+As above, the non-media endpoint must also serve media requests.
```
uhoreg, line 38: Please follow the proposal template.  You're missing some sections that I would expect to see for this type of proposal, in particular "Alternatives", "Security considerations", and "Unstable prefix".

Some possible alternatives:

- allow the homeserver to send an HTTP (3xx status code) on `/media` queries
- include the media server in the `/login` response, rather than in `.well-known`

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,38 @@
...
+Currently Matrix clients and servers discover homeservers and identify servers via the `.well-known/matrix/client` and `.well-known/matrix/server` endpoints. This is currently limited to the homeserver and for clients the identity server. This MSC proposes adding an additional field that can be used to specify a media server that can optionally be used for `/_matrix/media/*` endpoints.
+
+The reasoning behind this change is to support hosting media behind a CDN designed and optimised specifically for serving media. Since media transfers and endpoints are very different to the other JSON based endpoints a separate domain offers maximum flexibility for server owners.
+
+
+## Proposals
+
+### Extend the client .well-known
+
+Add a new optional field, `m.media_server` that can specify a separate URL to be used for media requests. Clients can then optionally use this field for media endpoints:
```
uhoreg, line 12: Can you clarify whether this is only for fetching media, or for uploading media as well?

proposals/3859-well-known-media-domain.md
Fizzadar: https://github.com/matrix-org/matrix-spec-proposals/pull/3859/commits/f1ca6064193c26d4d173e8cf241be8a8267e76e9

proposals/3859-well-known-media-domain.md
Fizzadar: Whoops, thanks for spotting that - https://github.com/matrix-org/matrix-spec-proposals/pull/3859/commits/f1ca6064193c26d4d173e8cf241be8a8267e76e9

proposals/3859-well-known-media-domain.md
Fizzadar: Added in https://github.com/matrix-org/matrix-spec-proposals/pull/3859/commits/f1ca6064193c26d4d173e8cf241be8a8267e76e9 (both up & down).

proposals/3859-well-known-media-domain.md
Fizzadar: Missng sections added in https://github.com/matrix-org/matrix-spec-proposals/pull/3859/commits/86123b0b6a5a331b35a984b1b9524c2009b2b591.

Re: login response I don't think this works because media is accessible without authentication with a known MXID which login would prevent.

proposals/3859-well-known-media-domain.md
Fizzadar: Have expanded the reasoning section in https://github.com/matrix-org/matrix-spec-proposals/pull/3859/commits/7ba95c8174c62e81c5af7754b198ff01478016a6. In short: path optimisation for client upload/download requests to the media repo.

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,86 @@
...
+
+Currently Matrix clients and servers discover homeservers and identify servers via the `.well-known/matrix/client` and 
+`.well-known/matrix/server` endpoints. This is currently limited to the homeserver and for clients the identity server.
+This MSC proposes adding an additional field that can be used to specify a media server that can optionally be used 
+for `/_matrix/media/*` endpoints.
+
+The reasoning behind this change is to support hosting media behind a CDN designed and optimised specifically for 
+serving media. Since media transfers and endpoints are very different to the other JSON based endpoints a separate 
+domain offers maximum flexibility for server owners. This is particularly important for path optimisation between
+client and server over long dinstances where home internet providers generally offer inferior routing between countries
```
turt2live, line 11:
```suggestion
client and server over long distances where home internet providers generally offer inferior routing between countries
```

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,86 @@
...
+than CDN providers do. By allowing clients to upload/download from the closest CDN POP they will benefit from improved
```
turt2live, line 12:
```suggestion
than CDN providers do. By allowing clients to upload/download from the closest CDN Point of Presence they will benefit from improved
```

(most people should know what a CDN is in this context, but POP is less common as an acronym)

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,86 @@
...
+upload and download performance.
+
+This cannot be achieved currently unless the entire homeserver is served behind a CDN, which may have other cost or
+management implications (or simply be impossible, depending on CDN). A separate domain allows server admins to
+decide the best setup for their users.
```
turt2live, line 17: This needs expanding, or at least covering in the alternatives. As mentioned, for my own server I'm one flag away from having CDN-driven media without spec changes. Is there a particular CDN provider in mind for your use case which prevents working within the existing spec?

For reference, the setup I've been looking at is effectively deploying my own CDN on top of a cloud provider: there's a reverse proxy in several geographical areas using DNS to direct traffic there, with the media repo running in that datacenter to serve media from its CDN-backed backend (so all media is available everywhere). The traffic routing for client-server and server-server traffic is effectively unchanged in this setup, though might use a backbone back to the real server hosted in a single geographical area. The latency would be nearly the same as if a user was talking to the server over an ocean directly. 

Granted, this is not a super cost effective solution, but it does work within how Matrix expects a server to operate. In practice, we'd also want to support proper geographically-segmented homeservers for traffic balancing in that respect, however that's much harder to do (and likely an implementation problem rather than spec). 

A benefit of this proposal is it allows the media domain to be different from the homeserver domain, which is also something we want to do, though I do wonder if making the media repo be a proper first class citizen in the spec is a good idea.

proposals/3859-well-known-media-domain.md
```diff
@@ -0,0 +1,86 @@
...
+{
+  "com.beeper.msc3859.m.media_server": {
+    "base_url": "https://matrix-media.example.com"
+  }
+}
+# Server
+{
+  "com.beeper.msc3859.m.media_server": "matrix-media.example.com:443"
+}
+```
```
turt2live, line 86: so github doesn't complain about json syntax:
```suggestion
```json5
// Client
{
  "com.beeper.msc3859.m.media_server": {
    "base_url": "https://matrix-media.example.com"
  }
}
// Server
{
  "com.beeper.msc3859.m.media_server": "matrix-media.example.com:443"
}
```
```

proposals/3859-well-known-media-domain.md
Fizzadar: Will add this to alternatives! The reverse proxy / CDN over everything solution certainly works but forces all traffic to be routed via that solution. Our use case specifically is we'd like to run Synapse inside AWS (as we currently do) but have the media routes served elsewhere due to AWS' prohibitive bandwidth costs. Without a separate domain the only option for that is a solution similar to yours where a matrix-aware reverse proxy needs to sit on the CDN side routing requests accordingly.

While this is certainly possible (cloudflare workers, fly.io, custom build CDN) it's represents a significant amount of work vs. a separate domain which to me feels like a very simple and easy solution. I can't see any negatives to doing this (esp. as an optional field).

Does having a separate domain really make the media repo a first class citizen? Perhaps there should be some kind of reliance on a homeserver to avoid that, or perhaps it's actually a good thing? (Multiple homeserver owners could share a media repo install or something along those lines, way beyond this proposal).

proposals/3859-well-known-media-domain.md
Fizzadar: Nice! Had not heard of json5 :)

proposals/3859-well-known-media-domain.md
uhoreg: > Re: login response I don't think this works because media is accessible without authentication with a known MXID which login would prevent.

That seems reasonable.  I'd suggest still including that in the "Alternatives" section -- part of the purpose of that section is to indicate other possible ways to achieve the goal, and say why the proposed solution is better.

Fizzadar:
Based upon internal discussion we no longer think this is the right approach and thus we are closing this MSC in favour of [MSC3860](https://github.com/matrix-org/matrix-spec-proposals/pull/3860) & [MSC3870](https://github.com/matrix-org/matrix-spec-proposals/pull/3870).

PR #3758, opened 2022-03-27
MSC3758: Add `event_property_is` push rule condition kind

Fizzadar:
[Rendered](https://github.com/Fizzadar/matrix-spec-proposals/blob/main/proposals/3758-expand-push-rule-conditions.md)

Implementation:
* matrix-org/synapse#14964.
* matrix-org/matrix-js-sdk#3179

Signed-off-by: Nick Mills-Barrett <nick@beeper.com>

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3758#issuecomment-1428764454)

PR #2223, opened 2019-08-08
Check for a state_key on the tombstone push rule

turt2live:
This is an oversight from the proposal: https://github.com/matrix-org/matrix-doc/pull/1930

PR #3862, opened 2022-08-05
MSC3862: event_match (almost) anything

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/event-match-almost-anything/proposals/3862-event-match-almost-anything.md)

Implementations:

https://github.com/matrix-org/synapse/pull/13466

### DIFF
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -62,7 +62,8 @@ These values are not possible to differentiate with the generalised
 practice it is rare for two events to differ only in the data type of a
 particular field. If that happens, it’s most probably due to the
 evolution of an unstable event type in which case it’s likely desirable
-to treat both data types identically anyway.
+to treat both data types identically anyway. The same applies to
+situations where clients accidentally sent the wrong type.
 
 ## Alternatives
 
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -12,7 +12,7 @@ primitive value.
 
 ## Proposal
 
-In order to apply an `event_match` condition, home servers transform all
+In order to apply an `event_match` condition, homeservers transform all
 primitive values in the event body into a lowercase string
 representation. In particular this means
 
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -1,6 +1,6 @@
 # MSC3862: event_match (almost) anything
 
-The `event_match` condition kind on push rules currently only matches
+The [`event_match`] condition kind on push rules currently only matches
 string values. This is limiting because some event types have important
 information in boolean or integer fields that currently cannot be
 matched against.
@@ -27,7 +27,7 @@ representation. In particular this means
 After the transformation the condition is evaluated on the string
 representations.
 
-In order to reflect these changes, the wording in the spec is updated.
+In order to reflect these changes, the wording in the [spec] is updated.
 Specifically
 
 > **event_match** This is a glob pattern match on a field of the event.
@@ -74,5 +74,7 @@ e.g. [matrix-org/synapse#13466]) and use.
 
 None.
 
+  [`event_match`]: https://spec.matrix.org/v1.3/client-server-api/#conditions-1
+  [spec]: https://spec.matrix.org/v1.3/client-server-api/#conditions-1
   [MSC3758]: https://github.com/matrix-org/matrix-spec-proposals/pull/3758
   [matrix-org/synapse#13466]: https://github.com/matrix-org/synapse/pull/13466
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -1,4 +1,4 @@
-# MSC0000: event_match (almost) anything
+# MSC3862: event_match (almost) anything
 
 The `event_match` condition kind on push rules currently only matches
 string values. This is limiting because some event types have important
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -1,9 +1,9 @@
 # MSC3862: event_match (almost) anything
 
-The [`event_match`] condition kind on push rules currently only matches
-string values. This is limiting because some event types have important
-information in boolean or integer fields that currently cannot be
-matched against.
+The [`event_match`] condition kind on push rules currently only
+matches string values. This is limiting because some event types have
+important information in boolean or integer fields that currently cannot
+be matched against.
 
 [MSC3758] tries to address this by introducing a new condition kind
 `exact_event_match`. The current proposal takes a different route by
@@ -20,15 +20,41 @@ representation. In particular this means
 - `true` / `false` become `"true"` / `"false"`
 - Integers such as `123` become `"123"`
 - `null` becomes `"null"`
-- Arrays remain unchanged
-- Nested objects are handled recursively, their values being transformed
-  according to the rules above
+
+This transformation is applied regardless of how deeply the value is
+nested but not if the value is part of an array. Arrays – which are not
+primitive values – and their elements are not converted. This means
+that, as before, it's not possible to match anything inside an array.
+
+For the avoidance of doubt, the exemplary event body
+
+    {
+      "a": "string",
+      "b": true,
+      "c": {
+        "d": 1,
+        "e": null
+      },
+      "f": [1]
+    }
+
+is transformed into
+
+    {
+      "a": "string",
+      "b": "true",
+      "c": {
+        "d": "1",
+        "e": "null"
+      },
+      "f": [1]
+    }
 
 After the transformation the condition is evaluated on the string
 representations.
 
-In order to reflect these changes, the wording in the [spec] is updated.
-Specifically
+In order to reflect these changes, the wording in the
+[spec] is updated. Specifically
 
 > **event_match** This is a glob pattern match on a field of the event.
 
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -53,28 +53,6 @@ is transformed into
 After the transformation the condition is evaluated on the string
 representations.
 
-In order to reflect these changes, the wording in the
-[spec] is updated. Specifically
-
-> **event_match** This is a glob pattern match on a field of the event.
-
-is changed to
-
-> **event_match** This is a glob pattern match on any primitive field of
-> the event. Non-string values are converted to their lowercased string
-> representation before attempting the match.
-
-and
-
-> If the property specified by `key` is completely absent from the
-> event, or does not have a string value, then the condition will not
-> match, even if `pattern` is `*`.
-
-is changed to
-
-> If the property specified by `key` is completely absent from the
-> event, then the condition will not match, even if `pattern` is `*`.
-
 ## Potential issues
 
 It is possible for JSON values of different data types to map to the
--- a/proposals/3862-event-match-almost-anything.md
+++ b/proposals/3862-event-match-almost-anything.md
@@ -16,7 +16,7 @@ In order to apply an `event_match` condition, homeservers transform all
 primitive values in the event body into a lowercase string
 representation. In particular this means
 
-- Strings are converted into lowercase (this already happens today)
+- Strings are converted into case-insensitive strings (this already happens today)
 - `true` / `false` become `"true"` / `"false"`
 - Integers such as `123` become `"123"`
 - `null` becomes `"null"`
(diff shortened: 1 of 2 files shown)


proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,78 @@
...
+[MSC3758] tries to address this by introducing a new condition kind
+`exact_event_match`. The current proposal takes a different route by
+generalising the existing `event_match` condition kind to work on any
+primitive value.
+
+## Proposal
+
+In order to apply an `event_match` condition, home servers transform all
+primitive values in the event body into a lowercase string
+representation. In particular this means
```
richvdh, line 17: @Johennes in order to help readers understand your proposal, please could you link to appropriate parts of the current spec?

In particular, https://spec.matrix.org/v1.3/client-server-api/#conditions-1 would be helpful here.

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Strings are converted into lowercase (this already happens today)
+- `true` / `false` become `"true"` / `"false"`
+- Integers such as `123` become `"123"`
+- `null` becomes `"null"`
+- Arrays remain unchanged
+- Nested objects are handled recursively, their values being transformed
+  according to the rules above
```
richvdh, line 25: I don't really understand what this means. you can convert the values of a nested object, but that still leaves you with an object rather than a string?

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,78 @@
...
+
+In order to apply an `event_match` condition, home servers transform all
+primitive values in the event body into a lowercase string
+representation. In particular this means
+
+- Strings are converted into lowercase (this already happens today)
+- `true` / `false` become `"true"` / `"false"`
+- Integers such as `123` become `"123"`
+- `null` becomes `"null"`
+- Arrays remain unchanged
```
richvdh, line 23: ... which is to say, they will never be matched?

proposals/3862-event-match-almost-anything.md
Johennes: Let's say you have something like

```
{
  "foo": true,
  "bar": {
    "foobar": false
  }
}
```

Applying the rules

> - Strings are converted into lowercase (this already happens today)
> - `true` / `false` become `"true"` / `"false"`
> - Integers such as `123` become `"123"`
> - `null` becomes `"null"`
> - Arrays remain unchanged

to the top level values gives

```
{
  "foo": "true",
  "bar": {
    "foobar": false
  }
}
```

and then you recurse into the object stored under "bar" and apply the rules again to produce

```
{
  "foo": "true",
  "bar": {
    "foobar": "false"
  }
}
```

In reconsidering maybe describing this as recursion is overcomplicated and we could just say that any string, integer, boolean or `null` _value_ - regardless of how deeply it is nested - is transformed according to the listed rules?

proposals/3862-event-match-almost-anything.md
Johennes: Yeah, exactly. We could possibly introduce the same "match the pattern on any element of the array" semantics that [MSC3758](https://github.com/matrix-org/matrix-spec-proposals/pull/3758) includes. I wasn't sure if array matching wouldn't rather deserve a condition kind of its own because there could be other use cases such as matching at a specific index.

richvdh: I think it makes sense to limit the number of things we're changing at once. But it would be good to clarify that they will still never be matched in the proposal.

proposals/3862-event-match-almost-anything.md
richvdh: > In reconsidering maybe describing this as recursion is overcomplicated and we could just say that any string, integer, boolean or null value - regardless of how deeply it is nested - is transformed according to the listed rules?

Right, yes, I think that would be clearer.

proposals/3862-event-match-almost-anything.md
Johennes: Agreed. I've added a phrase to call out the exclusion of arrays more clearly.

proposals/3862-event-match-almost-anything.md
Johennes: 👍 Reformulated accordingly.

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
...
+`event_match` condition kind as described in this proposal. However, in
+practice it is rare for two events to differ only in the data type of a
+particular field. If that happens, it’s most probably due to the
+evolution of an unstable event type in which case it’s likely desirable
+to treat both data types identically anyway.
+
+## Alternatives
+
+As mentioned earlier, [MSC3758] provides an alternative by introducing
+a new condition kind `exact_event_match`. While being more powerful,
```
Johennes, line 92: Adding an excerpt of a related discussion in the matrix-spec room:

Screenshot 2022-08-29 at 12 51 35

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Proposal
+
+In order to apply an `event_match` condition, home servers transform all
+primitive values in the event body into a lowercase string
+representation. In particular this means
+
+- Strings are converted into lowercase (this already happens today)
+- `true` / `false` become `"true"` / `"false"`
+- Integers such as `123` become `"123"`
```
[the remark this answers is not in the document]
turt2live: An added concern is I don't think any amount of documentation will save us from folks using language-specific features to encode numbers, which can easily lead to improper notation. The implementation likely wouldn't notice until it's too late, and generally cause the same sorts of problems we saw with Canonical JSON: implementations differing in how strict they are to the spec, leading to mismatched notifications.

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
+# MSC3862: event_match (almost) anything
```
turt2live, line 1: From a client perspective, I don't think either this MSC or https://github.com/matrix-org/matrix-spec-proposals/pull/3758 is particularly troublesome, though personally I do lean more towards this MSC as it shows what push rules "should" be: matches on events irrespective of data type (there's no reason to punish someone for `"42"` when they meant `42`, at least not in this subsystem). However, the appeal of a clean slate to "fix" push rules with 3758 is equally as high.

Both are equally complex to implement and would be considered a breaking change: here, clients won't agree with servers on what the notification counts are (not that they do already...<sup>[1]</sup>), and with 3758 it's a whole new condition type. In essence, both MSCs would be writing a new condition type into the client code-wise.

<sup>[1]</sup>Element Android for instance doesn't handle keyword notifications, leading to nasty surprises when the user moves from mobile to desktop.

proposals/3862-event-match-almost-anything.md
turt2live: I'm not sure this is a valid concern as-written. Potentially the context has further details, but the screenshot itself sounds very much like how the MSC process works: if your feature sits there long enough, it needs to un-bitrot every so often.

Johennes: There wasn't more context to it in the [conversation](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$tXB3rItH6Oclm_V61NpiDu6E1V0NyDBlimRuD-roSO4?via=matrix.org&via=libera.chat&via=element.io), sadly. 😔

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
...
+generalising the existing `event_match` condition kind to work on any
+primitive value.
+
+## Proposal
+
+In order to apply an `event_match` condition, home servers transform all
+primitive values in the event body into a lowercase string
+representation. In particular this means
+
+- Strings are converted into lowercase (this already happens today)
```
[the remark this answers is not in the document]
Johennes: Changed the wording accordingly

proposals/3862-event-match-almost-anything.md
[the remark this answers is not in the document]
Johennes: > We should clarify that it's a requirement to encode "as written" or whatever terminology is required to say that encoding to 123e^7 is invalid (as some int-to-string functions like to do on large numbers).

This made me think, what if somebody used the exponential form in the JSON? E.g. should `"pattern": 123e7` be converted into `"123e7"`?

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
...
+        "e": "null"
+      },
+      "f": [1]
+    }
+
+After the transformation the condition is evaluated on the string
+representations.
+
+In order to reflect these changes, the wording in the
+[spec] is updated. Specifically
```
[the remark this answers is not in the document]
Johennes: Sorry, I thought this was helpful. I have removed it now.

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,106 @@
...
+## Potential issues
+
+It is possible for JSON values of different data types to map to the
+same string representation (e.g. `0` and `"0"` or `true` and `"true"`).
+These values are not possible to differentiate with the generalised
+`event_match` condition kind as described in this proposal. However, in
+practice it is rare for two events to differ only in the data type of a
+particular field. If that happens, it’s most probably due to the
+evolution of an unstable event type in which case it’s likely desirable
+to treat both data types identically anyway.
```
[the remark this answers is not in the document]
Johennes: Good point. Added a corresponding phrase.

proposals/3862-event-match-almost-anything.md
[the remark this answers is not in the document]
turt2live: I don't believe it's allowable by Canonical JSON, but if it is then ugh/yay.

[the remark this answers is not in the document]
Johennes: Ah, you're right. It [isn't](https://spec.matrix.org/v1.3/appendices/#grammar). I've added a corresponding sub-bullet to explain the conversion.

proposals/3862-event-match-almost-anything.md
richvdh: > matches on events irrespective of data type (there's no reason to punish someone for "42" when they meant 42, at least not in this subsystem)

Do you really think so? It seems to me that this sort of "sloppy" matching could lead to things like https://github.com/matrix-org/matrix-spec-proposals/pull/2223, where the push rules end up matching something different to the rest of the system. (That was the cause of security vulnerabilities in synapse and matrix-js-sdk.)

richvdh: I think my primary concern with #3758 is that a new push rule condition type might be more likely to break clients than extending the existing one. If @turt2live's opinion, as a client developer, is that there's not much to choose, my inclination is to go with the more complete, and more elegant, #3758.

ara4n: I also prefer #3758, as it avoids casting everything to strings, and the dotted notation is more powerful.

turt2live: > > matches on events irrespective of data type (there's no reason to punish someone for "42" when they meant 42, at least not in this subsystem)
> 
> Do you really think so? It seems to me that this sort of "sloppy" matching could lead to things like #2223, where the push rules end up matching something different to the rest of the system. (That was the cause of security vulnerabilities in synapse and matrix-js-sdk.)

It's not a strong opinion, tbh. For things like notifications it'd be a nice-to-have to support typeless matches, leaving the type validation for elsewhere (whether an event can be rendered, etc). It has been the cause of some security vulnerabilities, but push rules in general are an open avenue for multiple different kinds of implementation-specific attack to begin with. It would be a challenge to keep the "sloppiness" to the push rules though: maintaining a higher standard for other subsystems to counteract the typelessness is not traditionally something that's gone well.

proposals/3862-event-match-almost-anything.md
```diff
@@ -0,0 +1,91 @@
+# MSC3862: event_match (almost) anything
```
richvdh, line 1: @ara4n:

> the dotted notation is more powerful.

What do you mean? Isn't the dotted notation the same?

ara4n: oh, i failed to remember that push rules have dotted notation in general.

Eitherway, it feels kludgey to cast everything to a string; #3758 feels nicer in terms of actually matching on specific types.

richvdh: thanks. I edited your comment to clarify.

proposals/3862-event-match-almost-anything.md
turt2live: I'm counting this as resolved - the author can raise it here if they have more detail, per the linked conversation.

Johennes:
Closing this since #3758 is moving ahead.

PR #3015, opened 2021-02-19
MSC3015: Room state personal overrides

MurzNN:
[Rendered](https://github.com/MurzNN/matrix-doc/blob/personal-room-names/proposals/3015-room-personal-overrides.md)

Related issues:
- [Personal ('pet name') room name for chats](https://github.com/vector-im/element-meta/issues/725)
- [Ability to override the displayname for users & rooms that get served to you (SPEC-427)](https://github.com/matrix-org/matrix-doc/issues/690)

Signed-off-by: Alexey Korepov <MurzNN@gmail.com>

PR #3864, opened 2022-08-07
MSCXXXX: User-given nicknames & descriptions for rooms & users

Zocker1999NET:


turt2live:
@Zocker1999NET when you get the chance, please [Sign Off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes.

Zocker1999NET:
I've found another proposal #3015 which may overlap with this one. I will check how both overlap and if, looking at the feature set and way of implementation, one is superior to the other. If so, they may should be merged.

Mart-Bogdan:
@Zocker1999NET I think that other proposal is abandoned. So if you would like to continue it would be great.

I agree with comments from discussion on other MSC, abut naming. Perhaps move `user_given` to the end. Or we can just leave it as is: `m.user_given.user.displayname` .

What do you think? I guess we need to resume discussion in room you've created: [#mscs-3015-3864-3865:matrix.org](https://matrix.to/#/#mscs-3015-3864-3865:matrix.org)

PR #1769, opened 2019-01-03
MSC1769: Extensible profiles as rooms

ara4n:

PR #3865, opened 2022-08-09
[WIP] MSCXXXX: User-given attributes for users

Zocker1999NET:
split from [MSC3864](https://github.com/matrix-org/matrix-spec-proposals/pull/3864)

### DIFF
--- a/proposals/user-nicknames.md
+++ b/proposals/user-nicknames.md
@@ -1,0 +1,237 @@
+# MSCXXXX: User-given attributes for users
+
+> This MSC is similar to [MSC3864](https://github.com/matrix-org/matrix-spec-proposals/pull/3864),
+> which specifies user-given attributes for rooms.
+> They were split because it makes more sense for now to store user attributes in another way
+> and because it shall be possible to set a user-given name for any MXID restricted to a single room.
+> However, they may share the same ideas and reasonings.
+
+User-given displaynames loaded from an addressbook are great,
+they allow you to personalize your experience in using an IM
+while others do not need to change anything.
+Like when you write with privacy-concerned users,
+they might use a nickname for their account on their own,
+but you may want to have them listed by their real name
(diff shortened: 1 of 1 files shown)


dklimpel:
Related to:
-  #1769

Mart-Bogdan:
As I understend size of single item in User Config is limited at 64 KiB

https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$ArNvLz_XK2CoXVb64g99ns6o-8DYIbDz0Ze_6NXMBaI?via=matrix.org&via=envs.net&via=element.io

So it can be an issue if somebody store too much data in `m.user_given.user.description` 

Perhaps we need to store data for each user individually? like `m.user_given.@alice:example.com.description`.

Sadly it would explode amount of user data stored, IDK if this can cause problems.

Other option is to create  `m.user_given.user.description.2` when first one run out of storage.

tusooa:
Just so you know, I am implementing this in kazv: https://iron.lily-is.land/T77

tusooa:
As this is done in kazv ( https://iron.lily-is.land/T77 ) I think we should remove the needs-implementation label.

proposals/3865-user-nicknames.md
```diff
@@ -0,0 +1,239 @@
...
+if they shall only apply for this certain room.
+Both kinds of client configs are only visible to the user who set them
+and are synced across all devices of the user,
+which makes them a perfect fit for storing user-given attributes.
+
+
+### Implementation
+
+For each attribute of a user profile, a new event type shall be defined.
+To get the new event type, the prefix `m.user_given.user.` shall be appended to the attribute's name.
```
uhoreg, line 76:
```suggestion
To get the new event type, the prefix `m.user_given.user.` shall be prepended to the attribute's name.
```

is more appropriate, since it goes at the beginning.

PR #3866, opened 2022-08-12
MSC3866: `M_USER_AWAITING_APPROVAL` error code

babolivier:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/babolivier/m_not_approved/proposals/3866-user-not-approved-error.md)

Synapse implementation: https://github.com/matrix-org/synapse/pull/13556

### DIFF
--- a/proposals/3866-user-not-approved-error.md
+++ b/proposals/3866-user-not-approved-error.md
@@ -1,4 +1,4 @@
-# MSC3866: `M_USER_NOT_APPROVED` error code
+# MSC3866: `M_USER_AWAITING_APPROVAL` error code
 
 Over the past few years, there has been some demand for the ability to let
 administrators of homeservers approve any new user created on their homeserver
@@ -9,9 +9,9 @@ additional details such as an email address.
 
 ## Proposal
 
-This document proposes the addition of a new `M_USER_NOT_APPROVED` error code to
-the Matrix specification. This error code can be returned in two scenarios:
-registration and login.
+This document proposes the addition of a new `M_USER_AWAITING_APPROVAL` error
+code to the Matrix specification. This error code can be returned in two
+scenarios: registration and login.
 
 This proposal does not describe a way for the homeserver to alert an
 administrator about new accounts that are waiting to be reviewed, or a way for
@@ -25,11 +25,11 @@ admin room).
 When a user successfully registers on a homeserver that is configured so that
 new accounts must be approved by an administrator, the final `POST
 /_matrix/client/v3/register` request is responded to with a `403 Forbidden`
-response that includes the `M_USER_NOT_APPROVED` error code. For example:
+response that includes the `M_USER_AWAITING_APPROVAL` error code. For example:
 
 ```json
 {
-    "errcode": "M_USER_NOT_APPROVED",
+    "errcode": "M_USER_AWAITING_APPROVAL",
     "error": "This account needs to be approved by an administrator before it can be used."
 }
 ```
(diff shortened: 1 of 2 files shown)


proposals/3866-user-not-approved-error.md
```diff
@@ -0,0 +1,90 @@
...
+Over the past few years, there has been some demand for the ability to let
+administrators of homeservers approve any new user created on their homeserver
+before their account can be used. This allows for better control upon
+registration and a potential option for mitigating abuse as an alternative to
+disabling registration on a homeserver or forcing new users to fill in
+additional details such as an email address.
+
+## Proposal
+
+This document proposes the addition of a new `M_USER_NOT_APPROVED` error code to
```
uhoreg, line 12: As a bike-shed-y comment, "not approved" could be interpreted as the user having been rejected.  Maybe "...AWAITING_APPROVAL" or "...APPROVAL_PENDING" would be better.

babolivier: Good point. I've updated the error code to `M_USER_AWAITING_APPROVAL`.

anoadragon453:
The author of this MSC has indicated that it is ready for review by the SCT.

proposals/3866-user-not-approved-error.md
```diff
@@ -0,0 +1,91 @@
...
+{
+    "errcode": "M_USER_AWAITING_APPROVAL",
+    "error": "This account needs to be approved by an administrator before it can be used."
+}
+```
+
+### Login
+
+When a user whose account is still pending approval by a server administrator
+attempts to log in, `POST /_matrix/client/v3/login` requests are responded to
```
deepbluev7, line 40: Is this also supposed to handle, which never uses the register step?

proposals/3866-user-not-approved-error.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+This error can be returned in any login request, as long as the homeserver is
+confident that the user trying to log in is pending approval - as opposed to
+registration requests where only the last one can return such an error, in order
+to ensure the registration completes.
+
+Once an account is approved, homeserver must allow the user to log in (unless
+the account becomes unavailable for unrelated reasons, e.g. by getting
+deactivated).
```
deepbluev7, line 58: I think I would like to see, how that looks like in practice. If someone registers on my server, they might not know, how they will get contacted about their account request being approved or denied. Which means they basically just end up retrying to login over and over. Some way to respond with a message about how the user can expect to be notified of the approval or some long polling endpoint might be useful?

turt2live (changes requested):
Per SCT room, and in line with https://github.com/matrix-org/matrix-spec-proposals/pull/3848, I think this needs a client implementation to prove the following facts:

* Existing, unmodified, clients don't explode when they see this error code (this can be done with a manual test for the major platforms)
* At least 2 clients on different major platforms are capable of adopting the change

Once these are proven, it may very well be ready for FCP. I'm always a bit nervous about going from open MSC to FCP this quickly though, especially for error codes which have bitten us badly in the past (hence the ask for more involved client development).

proposals/3866-user-not-approved-error.md
babolivier: I'm assuming the missing word here is "SSO", in which case the answer is yes, will update.

proposals/3866-user-not-approved-error.md
babolivier: I'm not sure long-polling makes much sense given approval is manual, which means it could take hours or days/weeks to come through depending on how responsive the admin is.
A field in the response that indicates how the user should be informed of approval is a good idea. Though the issue here is that at this point we might not have a good way to reach out to the user, so if e.g. they didn't register with an email address (either because the server doesn't support it, or made it optional, etc), then we don't have much choice but to fall back to telling the user that their admin might get in touch with them oob or just to try logging in again later.
Will update the MSC with the additional response field and mention this issue in the "Potential issues" section.

proposals/3866-user-not-approved-error.md
babolivier: Done: https://github.com/matrix-org/matrix-spec-proposals/pull/3866/commits/3df48d3349538081cc015948a0be12ccb783707a

babolivier:
Heads up I am formally abandoning this proposal since I don't currently have time for Matrix spec contributions (I'm not sure whether I should close it or wait for someone to adopt it).

PR #3468, opened 2021-11-03
MXCXXXX: MXC to Hashes

ShadowJonathan:
[Rendered (but not really)](https://www.youtube.com/watch?v=dQw4w9WgXcQ)

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`

PR #3868, opened 2022-08-15
[WIP] MSCXXX: Room Contribution

jae1911:
[Rendered](https://github.com/jae1911/matrix-spec-proposals/blob/0d213afe09c046848e59a5f93a507d5fed663948/proposals/xxxx-room-contribution.md)

Signed-off-by: Aminda Suomalainen <suomalainen+git@mikaela.info>
Signed-off-by: Jae Lo Presti <me@jae.fi>

### DIFF
--- a/proposals/xxxx-room-contribution.md
+++ b/proposals/xxxx-room-contribution.md
@@ -1,0 +1,48 @@
+# MSCXXXX: Room Contribution
+
+Many projects and communities have a room on Matrix nowadays and many may wish to show their friends how to support them.  
+The goal, would be to have something only visible to room & space members and to provide something easier to manage & display than putting links into the room topic or announcing them periodically.
+
+## Proposal
+
+This proposal implements a new `m.room.contribute` state event. This state event contains a "uri" parameter that may link to a git repo or any kind of URI and a "text" parameter that may contain the text of the button directing to the support link. There may be multiple "uri" and "text" parameters, each adding a different button containing a different link. Alternatively to the "uri" parameter, a "copy" one can be specified and will tell clients to copy the value in the clipboard.
+An example of this event would look like this:
+
+```json
+{
+  "links": [
+    {
+      "uri": "https://git.me/contribute",
+      "text": "Contribute code!"
+    },
+    {
+      "uri": "https://translate.me/contribute",
+      "text": "Translate the project!"
+    },
+    {
+      "uri": "mailto:team@verygoodproje.ct",
+      "text": "Get in touch!"
+    },
+    {
+      "text": "Donation IBAN",
+      "copy": "CH5604835012345678009"
+    }
+  ]
+}
+```
+
+## Potential issues
+
+// To be determined
+
+## Alternatives
+
+Historically, that sort of advertisement is either taken care of by adding the said URIs to the room topic or periodically sending an announcement in the channel.
+
+## Security considerations
+
+State events are currently not encrypted which could leak some data.
+
+## Dependencies
+
+None.


proposals/3868-room-contribution.md
```diff
@@ -0,0 +1,48 @@
+# MSC3868: Room Contribution
+
+Many projects and communities have a room on Matrix nowadays and many may wish to show their friends how to support them.  
+The goal, would be to have something only visible to room & space members and to provide something easier to manage & display than putting links into the room topic or announcing them periodically.
+
+## Proposal
+
+This proposal implements a new `m.room.contribute` state event. This state event contains a "uri" parameter that may link to a git repo or any kind of URI and a "text" parameter that may contain the text of the button directing to the support link. There may be multiple "uri" and "text" parameters, each adding a different button containing a different link. Alternatively to the "uri" parameter, a "copy" one can be specified and will tell clients to copy the value in the clipboard.
```
turt2live, line 8: Wrap to ~120 characters please

turt2live:
(converting this to a draft given TODO-style comments within the text)

proposals/3868-room-contribution.md
jae1911: Fixed in https://github.com/matrix-org/matrix-spec-proposals/pull/3868/commits/aafc392975bb84dc8e5f7853442fb69a0d5c340b.

proposals/3868-room-contribution.md
```diff
@@ -0,0 +1,55 @@
+# MSC3868: Room Contribution
```
MTRNord, line 1: Not actually relevant for this MSC (and imho not a blocker), but I want to have it noted at an area we can link to later:

This seems also a pretty good thing to have when we have profiles as rooms eventually :) Basically having "personal links" via this in a nice way.

MayeulC:
A lot of Github projects display badges. It might be interesting to allow images as well as urls, possibly with an extra key: `"badge-url" : "mxc://something"`

jae1911:
> A lot of GitHub projects display badges. It might be interesting to allow images as well as urls, possibly with an extra key: `"badge-url" : "mxc://something"`

I find that to be a good idea, however, after a discussion with @Mikaela we are a bit concerned over some points:
 - Media repos, as of now (and at least the default one for Synapse), tends to conserve medias forever, even when events are deleted (https://github.com/matrix-org/synapse/issues/1263)
 - If a regular URI to a badge/image is used, the image would have to be proxied by the media repo to avoid the user loading the image directly and maybe leaking IP addresses and that would add complexity to the implementation of this MSC

MayeulC:
To be clearer:
* I think a regular URI should not be identified, only mxc content
* I did not initially think of "dynamic" badges, mostly something like these instead:

| image | image | image |
|---|---|---|
| image | image | etc |

You are right though, it would be tempting to upload a lot of images for dynamic badges.
It would be simpler if mxc links were content-based #3468, as that would de-duplicate them.

For cleanup, there is still the admin API: https://matrix-org.github.io/synapse/latest/admin_api/media_admin_api.html#delete-local-media-by-date-or-size and purge remote media. Those files should be relatively small, but I don't really see a solution short of allowing a background-image and formatted_body, which sounds relatively complex to implement.

jae1911:
Thanks for the input @MayeulC and what you said makes sense.
Added it to the MSC!

MayeulC:
Looks good!
Maybe advise against uploading new content too frequently?
I am not entirely sure about images:
* maybe base64 encoding would be better?
* is it necessary to specify width/height limits? Or is it implied that clients will only likely draw them with small dimensions, therefore it is advised to use images that work well at small size? Usually with such images, aspect ratio is the issue, as images have a max height of ~1em and no real max width (though in practice they do).

I assume people will just pick something that looks good on their client.

jae1911:
I assume as well that clients will format it the best it looks for them and will probably fix height and width limits on their own.

I'm gonna look into base64 but I think the MXC URIs are solid and the most practical choice for now.

PR #3715, opened 2022-02-07
MSC3715: Add a pagination direction parameter to `/relations`

clokep:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/clokep/relations-dir/proposals/3715-relations-parity-messages.md)

Implementation: matrix-org/synapse#11941.

[FCP Call](https://github.com/matrix-org/matrix-spec-proposals/pull/3715#issuecomment-1061935416)

PR #3869, opened 2022-08-16
MSCXXXX: Add an MSC for a new Widget API action to read related events from a matrix room

dhenneke:
[Rendered](https://github.com/nordeck/matrix-spec-proposals/blob/nic/feat/widgetapi-read-relations/proposals/0000-widgetapi-read-event-relations.md)

Signed-off-by: Dominik Henneke <dominik.henneke@nordeck.net>


proposals/3869-widgetapi-read-event-relations.md
```diff
@@ -0,0 +1,171 @@
...
+[MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762).
+
+The following rules apply:
+
+1. If the requested event-id belongs to a event, and the widget didn't receive the respective
+   `m.receive.event:<event type>` or `m.receive.state_event:<event type>` capability, the widget
+   rejects the request.
+2. The list of related events only include events that the widget has the respective
+   `m.receive.event:<event type>` or `m.receive.state_event:<event type>` capability for. Other
+   events are silently ignored.
```
robintown, line 34: I think this would make it more clear that the client is the one that 'ignores' the events, rather than the widget:

```suggestion
   events are silently dropped.
```

proposals/3869-widgetapi-read-event-relations.md
```diff
@@ -0,0 +1,171 @@
...
+defined, all types will be returned.
+
+The `event_type` parameter specifies the type of child events to search for. If not defined, all
+types will be returned.
+
+The `limit`, `from`, and `to` parameters work like described in
+([MSC2675](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2675-aggregations-server.md)).
+
+The `direction` parameter is used to specify the direction to search for relations. It has the same
+semantic as defined by ([MSC2675](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2675-aggregations-server.md)).
```
robintown, line 79: I'm failing to find the `direction` parameter in either MSC2675 or the specification.

proposals/3869-widgetapi-read-event-relations.md
```diff
@@ -0,0 +1,171 @@
...
+  "action": "read_relations",
+  "data": {
+    "event_id": "$event-id"
+  }
+}
+```
+
+The client SHOULD NOT modify the data of the request. The widget is responsible for producing valid
+events - the client MUST pass through any errors, such as permission errors, to the widget using the
+standard error response in the Widget API.
```
robintown, line 97: I think we should be careful to not imply that the client shouldn't do any validation of the request at all. I'd suggest a slightly different wording:

```suggestion
The client SHOULD NOT modify the data of the request. Given that the widget produces a valid
request, the client MUST pass through any errors, such as permission errors, to the widget using the
standard error response in the Widget API.
```

robintown (commented):
Generally looks like what I was expecting :+1:

proposals/3869-widgetapi-read-event-relations.md
dhenneke: You're right. I mixed up the MSC numbers. I meant [MSC3715](https://github.com/matrix-org/matrix-spec-proposals/pull/3715).

It is available as an experimental option in synapse:

https://github.com/matrix-org/synapse/blob/3dd175b628bab5638165f20de9eade36a4e88147/synapse/rest/client/relations.py#L59-L67

And also part of the `fetchRelations` function in the `matrix-js-sdk` where I come across it. Though the name of the query parameter doesn't match with the synapse implementation 🤔:

https://github.com/matrix-org/matrix-js-sdk/blob/8502759e3ee3a2b244a442b8dac2b67809d9270e/src/client.ts#L7298

We don't have a real use case for this feature, yet. So we could as well skip it for now. It also seems like the MSC in question [isn't final yet](https://github.com/matrix-org/matrix-spec-proposals/pull/3715#discussion_r931343116).

robintown: Hm, well I guess it's ultimately up to you whether to include the parameter, since MSC3715 is fairly close to FCP

proposals/3869-widgetapi-read-event-relations.md
dhenneke: I copied it from I think MSC2762. I also hesitated to add it here because it looks more like a kind of a general rule on how the widget interaction works in general. Maybe we can just leave it out.

proposals/3869-widgetapi-read-event-relations.md
dhenneke: Then let's keep it in and hope that it becomes stable in synapse and get's fixed in the `matrix-js-sdk` so we can actually use it. It is also useful because then the `relations(...)` endpoint in the `RoomWidgetClient` will also be fully functioning.

proposals/3869-widgetapi-read-event-relations.md
robintown: Makes sense, I think I'd still be favor of changing the wording even if it is copied from MSC2762, but if you want to keep it or leave it out, that's also fine.

proposals/3869-widgetapi-read-event-relations.md
clokep: > And also part of the `fetchRelations` function in the `matrix-js-sdk` where I come across it. Though the name of the query parameter doesn't match with the synapse implementation 🤔:
> 
> [matrix-org/matrix-js-sdk@8502759/src/client.ts#L7298](https://github.com/matrix-org/matrix-js-sdk/blob/8502759e3ee3a2b244a442b8dac2b67809d9270e/src/client.ts#L7298)

This is vector-im/element-web#22501, I believe.

dbkr (commented):
This feels broadly along the lines of our continuing extensions of the CS API into the widget API which feels like a lot of duplication, but if it gets us to a better place in the short term it may be viable.

proposals/3869-widgetapi-read-event-relations.md
```diff
@@ -0,0 +1,161 @@
...
+  "data": {
+    "event_id": "$event-id"
+  }
+}
+```
+
+If the widget doesn't have appropriate permission, or an error occurs anywhere along the send path,
+a standardized widget error response is returned.
+
+If the request was successful, the client sends the following response:
```
toger5, line 95: something along the lines of:
```suggestion
Widgets can rely on the complete relation lists.
The client is responsible to fetch all missing events form the matrix server.
If the request was successful, the client sends the following response:
```

toger5 (approved):
Looks good to me.
I am very happy about the doors this will open. Widgets finally can read "unlimited" event lists!
I commented a couple of things that I think would be cool to have mentioned here (or to be more explicit) for future reference.

proposals/3869-widgetapi-read-event-relations.md
```diff
@@ -0,0 +1,161 @@
...
+
+We could also add features to let the client "scroll up the room", i.e. trigger the pagination for a
+room timeline and stick with the original read interfaces. However, this would potentially load a
+lot of unrelated events which slows the read process down. In addition, the client must potentially
+decrypt all the messages in the room before being able to filter them accordingly. Using the relations
+feature, the decryption problem is still present, but the set of events that must be decrypted and
+searched is minimized.
+
+The same limitations would apply if we would consider to provide direct access to the
+`GET /_matrix/client/v3/rooms/{roomId}/messages` endpoint.
```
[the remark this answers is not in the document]
dhenneke: Not sure what you mean here. Filters won't work for e2e encryption because one would need to read and decrypt every message in the room to decide it which is not practical. But we could add a comment to clarify it.

proposals/3869-widgetapi-read-event-relations.md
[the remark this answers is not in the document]
dhenneke: You mean that the client could also respond to this with a local cache of the relations? I implied that it should always get it from CS-API because it's also how it is (and was already) implemented in the `matrix-js-sdk`/Element. But it might be a good idea to clarify it here.

PR #2246, opened 2019-08-24
MSC2246: Asynchronous media uploads

tulir:
[Rendered](https://github.com/tulir/matrix-doc/blob/asynchronous_uploads/proposals/2246-asynchronous-uploads.md)

Implementations:
* Server: https://github.com/turt2live/matrix-media-repo/pull/364, https://github.com/matrix-org/synapse/pull/12484
* Client upload (actually a bridge): https://github.com/mautrix/go/commit/3b9911029134eb2d5de96f7898a70064199061a5 + https://github.com/mautrix/whatsapp/commit/f79ca422e8c3df7801f3e5e6a896ccf725eb1f14
* Client download should work with all existing clients to some extent with the 20s default stalling

Signed-off-by: Tulir Asokan <tulir@maunium.net>

--- 

[FCP Proposal](https://github.com/matrix-org/matrix-spec-proposals/pull/2246#issuecomment-1075371292)

PR #3870, opened 2022-08-17
MSCXXXX: Media Async Upload URLs

Fizzadar:
Draft to expand upon grabbing an MSC number.

### DIFF
--- a/proposals/xxx-media-async-upload-url.md
+++ b/proposals/xxx-media-async-upload-url.md
@@ -1,0 +1,3 @@
+# MSCXXX: Media Async Upload URL
+
+Draft: extension on async uploads to return a URL to upload direct to in the create media response. Ex. an S3 presigned URL.


proposals/xxx-media-async-upload-url.md
```diff
@@ -0,0 +1,3 @@
+# MSCXXX: Media Async Upload URL
```
turt2live, line 1: From the title, this sounds like it might conflict with #2246

Fizzadar: Fixed/updated!

proposals/3870-media-async-upload-url.md
```diff
@@ -0,0 +1,63 @@
...
+
+`POST /_matrix/media/v3/create`
+
+**Example response**
+
+```json
+{
+  "content_uri": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+  "unused_expires_at": 1647257217083,
+  "upload_url": "https://cdn.example.com/media-repo/upload/XAPw4CtrzArk?signed=h4tGOHvCu"
```
uhoreg, line 34: What HTTP method does the client use to upload the file?  `POST`?  `PUT`?  Or do you need another parameter to specify the method?

proposals/3870-media-async-upload-url.md
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+If the client chooses to upload media via the `upload_url` field, the media server must implement
+it's own methods of detecting when an upload is complete if desired (eg. thumbnailing or spam
+detection).
+
+**Alternative**: a new endpoint that clients `POST` to after uploading media to the `upload_url`
+used to notify the media server. However in the case of spam detection this makes it possible for
```
uhoreg, line 43: It may be worthwhile having this endpoint so that clients can get an error message from the Matrix server in case the server rejects the upload for whatever reason.  Regarding bypassing spam detection by bad clients, wouldn't it be possible to make the file inaccessible until the endpoint is called?  In fact, maybe you could change it so that you get the `mxc:` URL after you call this endpoint, rather than when you call the `create` endpoint.

Fizzadar: Totally agree; I've expanded this in https://github.com/matrix-org/matrix-spec-proposals/pull/3870/commits/ba15fa9fc5f2d2e70258567e563714585006f057. The content ID has to come down in creation stage otherwise no underlying event can be sent. This does open the door for some strange behaviour if uploads get rejected, however, so I've noted that.

proposals/3870-media-async-upload-url.md
turt2live: We can probably encode it into the field here, making it generic against providers.

```suggestion
  "upload_location": "PUT https://cdn.example.com/media-repo/upload/XAPw4CtrzArk?signed=h4tGOHvCu"
```

proposals/3870-media-async-upload-url.md
```diff
@@ -0,0 +1,87 @@
...
+download redirect URLs) this makes it possible to run a "lean" media server that orchestrates where
+media is uploaded to and downloaded from but does not directly handle much media itself. Note this 
+doesn't include thubmnails or URL previews which would still go through the media server.
+
+
+## Proposal
+
+The proposal modifies the MSC2246 create endpoint to optionally include a URL that a compatible
+client may use to upload the content for the created MXID. This is considered optional and the
+client may continue to upload via the new upload endpoint defined in MSC2246.
```
turt2live, line 24: Clients should be required to use the URL when the server generates one, I think. The server is almost certainly trying to shed bandwidth load when it generates one of these URLs, and it's already gone through the effort to allocate a storage location for the client. That's a lot of wasted effort/cash if the client ignores the field.

For MSC purposes, making it required I think means:
* New endpoint version for `/create`, to encourage clients to opt-in to the field's presence
* Deprecating the old endpoint version, further pushing clients towards the new one
* On `PUT /upload`, an error code to denote that uploads aren't welcome for that URI. This does not require a new endpoint version because the client already opted-in to using the new upload URL-supported endpoint earlier in the process.

proposals/3870-media-async-upload-url.md
```diff
@@ -0,0 +1,87 @@
...
+
+`POST /_matrix/media/v3/create`
```
turt2live, line 26: If keeping the optional behaviour:
```suggestion
`POST /_matrix/media/v1/create`
```

If making it mandatory per https://github.com/matrix-org/matrix-spec-proposals/pull/3870/files#r1488692873 :
```suggestion
`POST /_matrix/media/v2/create`
```

(the v3 endpoint has never existed in the spec)

proposals/3870-media-async-upload-url.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Example response**
+
+```json
+{
+  "content_uri": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+  "unused_expires_at": 1647257217083,
+  "upload_url": "https://cdn.example.com/media-repo/upload/XAPw4CtrzArk?signed=h4tGOHvCu"
```
turt2live, line 34: I assume the URL expires at the same time as `unused_expires_at` - just need to document that in the MSC.

proposals/3870-media-async-upload-url.md
Fizzadar: I think we should use a separate field here just to be explicit, `upload_method`, will update.

proposals/3870-media-async-upload-url.md
Fizzadar: I think we should leave this optional for now, and defer deprecation or blocking of direct upload to a later MSC. Allowing both in theory makes switching over to the upload URLs easier for client dev, and then we avoid all introducing additional complexity as part of this MSC.

turt2live: > Wouldn't clients still be able to avoid using the CDN by using the old synchronous upload endpoint?
> 
> If we're going through the effort to insist that clients use the `upload_url`, should we also start making plans tokill the ability to do synchrous uploads direct to the media server at all?

The proposed plan deprecates the endpoint, making it eligible for removal in a later version of the spec.



> I think we should leave this optional for now, and defer deprecation or blocking of direct upload to a later MSC. Allowing both in theory makes switching over to the upload URLs easier for client dev, and then we avoid all introducing additional complexity as part of this MSC.

Clients can still use deprecated endpoints, but without a reason to change there's not much motivation for them to implement this. I do feel quite strongly that if the server provides a URL, the client *must* use that URL.

proposals/3870-media-async-upload-url.md
Fizzadar: https://github.com/matrix-org/matrix-spec-proposals/pull/3870/commits/3c9e240a949b4f9228639983a38b356e88c37767

proposals/3870-media-async-upload-url.md
Fizzadar: https://github.com/matrix-org/matrix-spec-proposals/pull/3870/commits/3c9e240a949b4f9228639983a38b356e88c37767

proposals/3870-media-async-upload-url.md
Fizzadar: https://github.com/matrix-org/matrix-spec-proposals/pull/3870/commits/af3d0fcb8dbbd9e2faf179f6e157a00d1c797d41

proposals/3870-media-async-upload-url.md
Fizzadar: I just pushed up fixes for the other things. I think deferring the new endpoint version and requiring clients use the URL provided is the right call - it'd be a quick follow up MSC, or might tie in with replacement plans for the old upload endpoints. Either way I don't see any real benefit to bundling it into this MSC.

Fizzadar:
Re: implementation: Beeper's custom HS/media-repo both already support this as well as all our clients. I don't recall if this is sufficient though?

PR #2285, opened 2019-09-06
MSC2285: Private read receipts

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/hidden-read-receipts/proposals/2285-hidden-read-receipts.md)

Author: @SimonBrandner

<hr>

#### synapse

+ https://github.com/matrix-org/synapse/pull/10413
+ https://github.com/matrix-org/synapse/pull/12168

#### matrix-js-sdk

+ https://github.com/matrix-org/matrix-js-sdk/pull/1028
+ https://github.com/matrix-org/matrix-js-sdk/pull/1788
+ https://github.com/matrix-org/matrix-js-sdk/pull/2221

#### matrix-react-sdk

+ https://github.com/matrix-org/matrix-react-sdk/pull/6390
+  https://github.com/matrix-org/matrix-react-sdk/pull/7993

----

[FCP Comment](https://github.com/matrix-org/matrix-spec-proposals/pull/2285#issuecomment-1154098132)

PR #3871, opened 2022-08-18
MSC3871: Gappy timelines

MadLittleMods:
A proposal for letting homeservers give feedback to clients where there is a gap in the timeline from a `/messages` response.

*Spawned from a discussion with @erikjohnston about making `/messages` faster*

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/madlittlemods/gappy-timelines/proposals/3871-gappy-timelines.md)

Matrix Live demo: https://www.youtube.com/watch?v=Bo7GaLRhtc0&t=409s
 
Implementations:

 - Synapse: https://github.com/element-hq/synapse/pull/18873

### DIFF
--- a/proposals/3871-gappy-timelines.md
+++ b/proposals/3871-gappy-timelines.md
@@ -38,7 +38,7 @@ timeline are as defined below.
 
 ### 200 response
 
-Thid describes the new `gaps` response field being added to the `200 response`
+This describes the new `gaps` response field being added to the `200 response`
 of `/messages`:
 
 Name | Type | Description | required
--- a/proposals/3871-gappy-timelines.md
+++ b/proposals/3871-gappy-timelines.md
@@ -6,11 +6,10 @@ events. This could be because the homeserver hasn't fetched them yet or because
 it failed to fetch the events because those homeservers are unreachable and no
 one else knows about the event.
 
-Currently, there is an unwritten expectation(TODO: better word) between the
-server and client that the server will always return all contiguous events in
-that part of the timeline. But the server has to break this promise(TODO: match
-word above) sometimes when it doesn't have the event and is unable to get the
-event from anyone else. This MSC aims to change the
+Currently, there is an unwritten rule between the server and client that the
+server will always return all contiguous events in that part of the timeline.
+But the server has to break this rule sometimes when it doesn't have the event
+and is unable to get the event from anyone else. This MSC aims to change the
 dynamic so the server can give the client feedback and an indication of where
 the gaps are.
 
--- a/proposals/3871-gappy-timelines.md
+++ b/proposals/3871-gappy-timelines.md
@@ -30,17 +30,27 @@ the client about the gap that they can retry fetching a little later.
 
 ## Proposal
 
-Add a `?gaps_allowed=true` query parameter flag to `GET
-/_matrix/client/v3/rooms/{roomId}/messages` which allows usage of a
-`m.timeline.gap` indicator, that can be used in the `response` `chunk` list of
-events. There can be multiple gaps per response.
+Add a `gaps` field to the response of [`GET
+/_matrix/client/v3/rooms/{roomId}/messages`](https://spec.matrix.org/v1.1/client-server-api/#get_matrixclientv3roomsroomidmessages).
+This field is an array of `GapEntry` indicating where missing events in the
+timeline are as defined below.
 
 
-### `m.timeline.gap`
+### 200 response
+
+Thid describes the new `gaps` response field being added to the `200 response`
+of `/messages`:
+
+Name | Type | Description | required
+--- | --- | --- | ---
+`gaps` | `[GapEntry]` | A list of gaps indicating where events are missing in the `chunk` | no
+
+
+#### `GapEntry`
 
 key | type | value | description | required
 --- | --- | --- | --- | ---
-`gap_start_event_id` | string | Event ID | The event ID that the homeserver is missing where the gap begins | yes
+`next_to_event_id` | string | Event ID | The event ID that the homeserver is missing where the gap begins | yes
 `pagination_token` | string | Pagination token | A pagination token that represents the spot in the DAG after the missing `gap_start_event_id`. Useful when retrying to fetch the missing part of the timeline again via `/messages?dir=b&from=<pagination_token>` | yes
 
 Pagination tokens are positions between events. This already an established
@@ -51,21 +61,94 @@ concept but to illustrate this better, see the following diagram:
 <oldest-in-time> [0]<--[1] <gap> [gap_start_event_id]▼<--[4]<--[5]<--[6] <newest-in-time>
 ```
 
-`m.timeline.gap` has a similar shape to a normal event so it's still easy to
-iterate over the `/messages` response and process but has no `event_id` itself
-so it should not be mistaken as a real event in the room.
 
-A full example of the `m.timeline.gap` indicator:
+### `/messages` response examples
 
-```json
+`/messages?dir=f` response example with a gap (`chunk` has events in
+chronoligcal order):
+
+```json5
 {
-  "type": "m.timeline.gap",
-  "content": {
-    "gap_start_event_id": "$12345",
-    "pagination_token": "t47409-4357353_219380_26003_2265",
-  }
+  "chunk": [
+    {
+      "event_id": "$foo",
+      "type": "m.room.message",
+      "content": {
+        "body": "foo",
+      }
+    },
+    // <the `GapEntry` indicates a gap here>
+    {
+      "event_id": "$baz",
+      "type": "m.room.message",
+      "content": {
+        "body": "baz",
+      }
+    },
+  ]
+  "gaps": [
+        {
+          "next_to_event_id": "$foo",
+          "pagination_token": "t47402-4357353_219380_26003_2265",
+        }
+  ]
 }
 ```
+
+
+`/messages?dir=b` response example with a gap (`chunk` has events in
+reverse-chronoligcal order):
+
+```json5
+{
+  "chunk": [
+    {
+      "event_id": "$baz",
+      "type": "m.room.message",
+      "content": {
+        "body": "baz",
+      }
+    },
+    // <the `GapEntry` indicates a gap here>
+    {
+      "event_id": "$foo",
+      "type": "m.room.message",
+      "content": {
+        "body": "foo",
+      }
+    }
+  ]
+  "gaps": [
+        {
+          "next_to_event_id": "$baz",
+          "pagination_token": "t47403-4357353_219380_26003_2265",
+        }
+  ]
+}
+```
+
+
+## Potential issues
+
+Lots of gaps/extremities are generated when a spam attack occurs and federation
+falls behind. If clients start showing gaps with retry links, we might just be
+exposing the spam more.
+
+
+## Alternatives
+
+As an alternative, we can continue to do nothing as we do today and not worry
+about the occasional missing events. People seem not to notice any missing
+messages anyway but they do probably see our slow `/messages` pagination.
+
+
+### Synthetic `m.timeline.gap` event alternative
+
+Another alternative is using synthetic events (thing that looks like an event
+without an `event_id`) which the server inserts alongside other events in the
+`chunk` to indicate where the gap is. But this has detractors since it's harder
+to implement in strongly typed SDK's and easy for a client to naively display
+every "event" in the `chunk`.
 
 `/messages` response example with a gap:
 
@@ -96,35 +179,20 @@ A full example of the `m.timeline.gap` indicator:
 ```
 
 
-## Potential issues
-
-Lots of gaps/extremities are generated when a spam attack occurs and federation
-falls behind. If clients start showing gaps with retry links, we might just be
-exposing the spam more.
-
-
-## Alternatives
-
-As an alternative, we can continue to do nothing as we do today and not worry
-about the occasional missing events. People seem not to notice any missing
-messages anyway but they do probably see our slow `/messages` pagination.
-
-
-
 ## Security considerations
 
-Only your own homeserver controls whether a `m.timeline.gap` indicator is added to the
-message response and it isn't an event of the room so there shouldn't be any weird
-edge case where the gap is trying to get you to fetch spam or something.
+Only your own homeserver controls whether a gap is added to the `/messages`
+response so there shouldn't be any weird edge case where someone else can
+control whether you to fetch something.
 
 
 ## Unstable prefix
 
-Servers will indicate support for the new endpoint via a true value for feature
+Servers will indicate support for this MSC via a `true` value for feature
 flag `org.matrix.msc3871` in `unstable_features` in the response to `GET
 /_matrix/client/versions`.
 
-While this feature is in development, it can be used as `GET
-/_matrix/client/unstable/org.matrix.msc3871/rooms/{roomId}/messages?gaps_allowed=true`
+While this feature is in development, the `gaps` field can be used as
+`org.matrix.msc3871.gaps`
 
 
--- a/proposals/3871-gappy-timelines.md
+++ b/proposals/3871-gappy-timelines.md
@@ -401,11 +401,29 @@ control whether you to fetch something.
 
 ## Unstable prefix
 
-Servers will indicate support for this MSC via a `true` value for feature
-flag `org.matrix.msc3871` in `unstable_features` in the response to `GET
-/_matrix/client/versions`.
-
 While this feature is in development, the `gaps` field can be used as
 `org.matrix.msc3871.gaps`
 
+### While the MSC is unstable
 
+During this period, to detect server support clients should check for the
+presence of the `org.matrix.msc3871` flag in `unstable_features` on `/versions`.
+Clients are also required to use the unstable prefixes (see [unstable
+prefix](#unstable-prefix)) during this time.
+
+### Once the MSC is merged but not in a spec version
+
+Once this MSC is merged, but is not yet part of the spec, clients should rely on
+the presence of the `org.matrix.msc3871.stable` flag in `unstable_features` to
+determine server support. If the flag is present, clients are required to use
+stable prefixes (see [unstable prefix](#unstable-prefix)).
+
+### Once the MSC is in a spec version
+
+Once this MSC becomes a part of a spec version, clients should rely on the
+presence of the spec version, that supports the MSC, in `versions` on
+`/versions`, to determine support. Servers are encouraged to keep the
+`org.matrix.msc3871.stable` flag around for a reasonable amount of time
+to help smooth over the transition for clients. "Reasonable" is intentionally
+left as an implementation detail, however the MSC process currently recommends
+*at most* 2 months from the date of spec release.
--- a/proposals/3871-gappy-timelines.md
+++ b/proposals/3871-gappy-timelines.md
@@ -50,32 +50,49 @@ Name | Type | Description | required
 
 key | type | value | description | required
 --- | --- | --- | --- | ---
-`next_to_event_id` | string | Event ID | The event ID that the homeserver is missing where the gap begins | yes
+`next_to_event_id` | string | Event ID | The event ID indicating the position in the `/messages` `"chunk"` response where the gap starts after that position. This field can be `null` or completely omitted to indicate that the gap is at the start of the `/messages` `"chunk"` | no
 `pagination_token` | string | Pagination token | A pagination token that represents the spot in the DAG after the missing `gap_start_event_id`. Useful when retrying to fetch the missing part of the timeline again via `/messages?dir=b&from=<pagination_token>` | yes
 
-Pagination tokens are positions between events. This already an established
-concept but to illustrate this better, see the following diagram:
-```
-                                   pagination_token
-                                   |
-<oldest-in-time> [0]<--[1]<-- <gap>▼ <--[4 (next_to_event_id)]<--[5]<--[6] <newest-in-time>
+
+### `/messages` response examples
+
+The following mermaid diagram represents the room DAG snapshot used for the following
+`/messages` responses. The slightly transparent events with no background are events
+that the homeserver does not have and are in the gap.
+
+Pagination tokens are positions between events. This already an established concept but
+to illustrate this better, see the following `tX` pagination tokens in the following
+diagram.
+
+```mermaid
+flowchart RL
+    after[newest events...]:::gap-event -->|t10| fred -->|t9| waldo:::gap-event -->|t8| garply -->|t7| grault:::gap-event -->|t6| corge -->|t5| qux:::gap-event -->|t4| baz -->|t3| bar:::gap-event -->|t2| foo -->|t1| before[oldest events...]:::gap-event
+
+    classDef gap-event opacity:0.8,fill:transparent;
 ```
 
 The idea is to be able to keep paginating from `pagination_token` in the same
 direction of the request to fill in the gap.
 
 
-### `/messages` response examples
-
 #### `/messages?dir=b`
 
+`/messages?dir=b` response example with gaps (`chunk` has events in
+reverse-chronoligcal order since we're paginating backwards):
 
-`/messages?dir=b` response example with a gap (`chunk` has events in
-reverse-chronoligcal order):
-
+`/messages?dir=b&from=t6`
 ```json5
 {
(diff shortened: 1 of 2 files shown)


proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,125 @@
...
+`/messages` returns a linearized version of the event DAG. From any given
+homeservers perspective of the room, the DAG can have gaps where they're missing
+events. This could be because the homeserver hasn't fetched them yet or because
+it failed to fetch the events because those homeservers are unreachable and no
+one else knows about the event.
+
+Currently, there is an unwritten expectation(TODO: better word) between the
+server and client that the server will always return all contiguous events in
+that part of the timeline. But the server has to break this promise(TODO: match
+word above) sometimes when it doesn't have the event and is unable to get the
```
MadLittleMods, line 12: What's the right phrasing/language to describe this?

 - constraint
 - unwritten rule
 - unspoken agreement
 - agreement
 - promise
 - relationship
 - expectation
 - dynamic

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,125 @@
...
+    {
+      "type": "m.room.message",
+      "content": {
+        "body": "foo",
+      }
+    },
+    {
+      "type": "m.timeline.gap",
+      "content": {
+        "gap_start_event": "$12345",
```
MadLittleMods, line 84:
```suggestion
        "gap_start_event_id": "$12345",
```

proposals/3871-gappy-timelines.md
kegsay: I'd say unwritten rule personally.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,125 @@
...
+dynamic so the server can give the client feedback and an indication of where
+the gaps are.
+
+This way, clients know where they are missing events and can even retry fetching
+by perhaps adding some UI to the timeline like "We failed to get some messages
+in this gap, try again."
+
+This can also make servers faster to respond to `/messages`. For example,
+currently, Synapse always tries to backfill and fill in the gap (even when it
+has enough messages locally to respond). In big rooms like `#matrix:matrix.org`
```
kegsay, line 23: This just feels like a synapse bug? Why is it doing this?

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,125 @@
...
+Pagination tokens are positions between events. This already an established
+concept but to illustrate this better, see the following diagram:
+```
+                                                     pagination_token
+                                                     |
+<oldest-in-time> [0]<--[1] <gap> [gap_start_event_id]▼<--[4]<--[5]<--[6] <newest-in-time>
+```
+
+`m.timeline.gap` has a similar shape to a normal event so it's still easy to
+iterate over the `/messages` response and process but has no `event_id` itself
```
kegsay, line 55: This will horribly break clients as they assume events have an event ID. There's no opt-in flag for this behaviour either... I suggest you add an opt-in flag to the query parameters or change the format of this to not break client expectations.

proposals/3871-gappy-timelines.md
MadLittleMods: @kegsay What do you expect it to do?

If it never tries to backfill at an extremity, it will never get new messages. Right now, it is pretty over-zealous about doing it but it still needs to happen sometime.

We have no idea how big the gap actually is so just returning messages from before also seems wrong. Say we asked for `/messages?dir=b&limit=5`, we could return `Z, Y, X, B, A` but people probably expect `[Z, Y, X, W, V]`

```mermaid
flowchart BT
    A --- annotation1>"Note: older events are at the top"]
    Z --> Y --> X --> gap{{gap}} --> B --> A
```

proposals/3871-gappy-timelines.md
MadLittleMods: I had this controlled by a new room version but that doesn't make much sense. Moved to a `?gaps_allowed=true` query parameter 👍

proposals/3871-gappy-timelines.md
kegsay: > even when it has enough messages locally to respond

This sounds like a bug. AFAIK Dendrite does not randomly do federation backfill requests if it has enough events locally to satisfy the request.

erikjohnston: We specifically try and backfill the gap if we're trying to paginate over it so that we will at some point try and fetch stuff in the gap. If we *don't* do that then we'll never pull in the events in the gap? Given we can't tell clients there is a gap we try really hard to get the messages before moving on, as doing so is effectively *dropping* messages without telling the uer (which is bad!)

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Potential issues
+
+Lots of gaps/extremities are generated when a spam attack occurs and federation
+falls behind. If clients start showing gaps with retry links, we might just be
+exposing the spam more.
+
+
+## Alternatives
```
MadLittleMods, line 106: Another alternative is including the `prev_events` in the events that the client sees so they can figure out the DAG chain themselves and see if there is an missing event in the middle.

There is an [unspecced `/messages?raw=true` query parameter in Synapse](https://github.com/matrix-org/synapse/blob/20c76cecb9eb84dadfa7b2d25b436d3ab9218a1a/synapse/rest/client/room.py#L653) that returns the full raw event as seen over federation which means it will include the `prev_events`.

It looks like there is also [specced `event_format: federation` that you can make a filter](https://spec.matrix.org/v1.1/client-server-api/#post_matrixclientv3useruseridfilter) to return raw events as well. But it looks like that is only usable with `/sync` which accepts a "ID of a filter" whereas `/messages` accepts a JSON `RoomEventFilter`, not a filter ID.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,130 @@
+# MSC3871: Gappy timeline
```
ara4n, line 1: @benkuly - you downvoted this MSC; can you explain what your concern is please?

benkuly: I don't like the concept of synthetic events. On the one hand it is something, that does not follow the typing of message and state events, which makes it harder to implement (especially for strongly typed SDKs like Trixnity) and even break clients. On the other hand it is a whole new concept, which does not exist yet and should not be shipped within an MSC as a byproduct.

The information of this event is something, that is specific to this endpoint. It could be also implemented as top-level fields of this response.

MadLittleMods: > It could be also implemented as top-level fields of this response.

Can you give an example how that would look? I would want it to be easy for clients to know where in the `/messages` `chunk` that the gap is.

benkuly: I think of something like this:
```json
{
  "chunk": [
    {
      "type": "m.room.message",
      "content": {
        "body": "foo",
      }
    },
    {
      "type": "m.room.message",
      "content": {
        "body": "baz",
      }
    },
  ],
  "gaps": [
        {
          "before_event_id": "$12345",
          "after_event_id": "$12345",
          "pagination_token": "t47409-4357353_219380_26003_2265",
        }
  ]
}
```
This leaves the chunk untouched, but adds further information in the response as top level field `gaps`. Therefore it is not masked as a sythetic event. In fact it is no event at all, it is just metadata.

proposals/3871-gappy-timelines.md
kegsay: I don't quite understand the concern here. If my server was offline for months and is now back online, my client will hit /messages for the most recent N messages to render it's UI. If the user then scrolls back further, _the client will initiate more requests to fill the gap_. I'm guessing the counter point is "what if you were binged 5 months ago, you'd never know". This is inevitable with the way we get message events full stop, as we never try to obtain the entire room dag, and even if we did, there exists no machinery or APIs to inform the client of this "obtained late" events.

MadLittleMods: Your example isn't really relevant to the problem here.

The concern is unreachable events (backward extremities). We could fail to get an event for an unlimited amount of reasons. What should `/messages` do when there is an obvious missing message that the homeserver can't retrieve?

This MSC proposes we indicate the missing message with a "gap" so the client knows something is missing and can fetch that spot again later on.

proposals/3871-gappy-timelines.md
MadLittleMods: Seems viable

That `pagination_token` represents a position between events so we probably want to only specify `before` gaps when going backwards, and `after` gaps for forwards. Otherwise, we would have to have a `before_pagination_token` and `after_pagination_token` to not accidentally lose an event when someone naively paginates and then it gets confusing on which to use.


`/messages?dir=b`
```json5
{
  "chunk": [...],
  "gaps": [
        {
          "before_event_id": "$12345",
          "pagination_token": "t47409-4357353_219380_26003_2265",
        }
  ]
}
```

`/messages?dir=f`
```json5
{
  "chunk": [...],
  "gaps": [
        {
          "after_event_id": "$12345",
          "pagination_token": "t47409-4357353_219380_26003_2265",
        }
  ]
}
```

---

And I imagine, we just drop the `before_event_id`/`after_event_id` label and both forward and backward would use `next_to_event_id` where that logic is just implicit.

```json5
{
  "chunk": [...],
  "gaps": [
        {
          "next_to_event_id": "$12345",
          "pagination_token": "t47409-4357353_219380_26003_2265",
        }
  ]
}
```

dkasak: This seems a bit better than synthetic events. For synthetic events there's an additional complication of having to ensure we don't receive them over federation accidentally, which this design avoids.

MadLittleMods: Thanks for the feedback @benkuly and @dkasak ❤️ 

I've updated the MSC with the top-level `gaps` approach and left the synthetic events one as an alternative ⏩

MadLittleMods: Friendly ping @benkuly to adjust your 👍👎, 🙇

benkuly: > And I imagine, we just drop the before_event_id/after_event_id label and both forward and backward would use next_to_event_id where that logic is just implicit.

Looks good to me, but isn't it relevant for the client where the gap is? Before or after the event? It is relevant for my SDK Trixnity, because each event can have a gap after and before or both.

MadLittleMods: > Looks good to me, but isn't it relevant for the client where the gap is? Before or after the event? It is relevant for my SDK Trixnity, because each event can have a gap after and before or both.

@benkuly The proposed `next_to_event_id` solution solves this. Except for one case as explained below. I'll update this MSC with this example and with the proposed fix here.

If there is a gap between every event, it would look like the the following:

```mermaid
flowchart RL
    after[newest events...]:::gap-event -->|t10| fred -->|t9| waldo:::gap-event -->|t8| garply -->|t7| grault:::gap-event -->|t6| corge -->|t5| qux:::gap-event -->|t4| baz -->|t3| bar:::gap-event -->|t2| foo -->|t1| before[oldest events...]:::gap-event

    classDef gap-event opacity:0.8,fill:transparent;
```

`/messages?dir=b&from=t6` (notice `dir=b`, `backwards`)
```json5
{
  "chunk": [
    // there is no gap from `t6` to `$corge` as expected
    {
      "event_id": "$corge",
      "type": "m.room.message",
      "content": {
        "body": "corge",
      }
    },
    // <the first `GapEntry` indicates a gap here>
    {
      "event_id": "$baz",
      "type": "m.room.message",
      "content": {
        "body": "baz",
      }
    },
    // <the second `GapEntry` indicates a gap here>
    {
      "event_id": "$foo",
      "type": "m.room.message",
      "content": {
        "body": "foo",
      }
    }
    // <the third `GapEntry` indicates a gap here>
  ]
  "gaps": [
        {
          "next_to_event_id": "$corge",
          "pagination_token": "t5",
        },
        {
          "next_to_event_id": "$baz",
          "pagination_token": "t3",
        },
        {
          "next_to_event_id": "$foo",
          "pagination_token": "t1",
        }
  ]
}
```


`/messages?dir=f&from=t6` (notice `dir=f`, `forwards`)
```json5
{
  "chunk": [
    // Uh-oh!! There is no way to indicate that there is a gap here
    {
      "event_id": "$garply",
      "type": "m.room.message",
      "content": {
        "body": "garply",
      }
    },
    // <the first `GapEntry` indicates a gap here>
    {
      "event_id": "$fred",
      "type": "m.room.message",
      "content": {
        "body": "fred",
      }
    },
    // <the second `GapEntry` indicates a gap here>
  ]
  "gaps": [
        {
          "next_to_event_id": "$garply",
          "pagination_token": "t8",
        },
        {
          "next_to_event_id": "$fred",
          "pagination_token": "t10",
        }
  ]
}
```

In order to fix the **Uh-oh!!** from the response above with no way to indicate a gap at that spot, I think we just need a way to specify a gap at the beginning of the response. Either by specifying `next_to_event_id` as `null` or omiting(`undefined`).

`/messages?dir=f&from=t6` (notice `dir=f`, `forwards`)
```json5
{
  "chunk": [
    // <the first `GapEntry` indicates a gap here>
    {
      "event_id": "$garply",
      "type": "m.room.message",
      "content": {
        "body": "garply",
      }
    },
    // <the second `GapEntry` indicates a gap here>
    {
      "event_id": "$fred",
      "type": "m.room.message",
      "content": {
        "body": "fred",
      }
    },
    // <the third`GapEntry` indicates a gap here>
  ]
  "gaps": [
        {
          "next_to_event_id": null, // `null` or this field emitted
          "pagination_token": "t6",
        },
        {
          "next_to_event_id": "$garply",
          "pagination_token": "t8",
        },
        {
          "next_to_event_id": "$fred",
          "pagination_token": "t10",
        }
  ]
}
```

benkuly: Thank you for the clarification :slightly_smiling_face: 

For my SDK Trixnity it would be relevant, to have both tokens within a gap to fetch it from both directions.¹ If a gap has hundreds of events, it could be relevant, whether the client fetches it from one or the other direction, because he cannot fill it with one request. Is there a reason, why we should hide this information?

¹ Background: Trixnity does not save pages of events, which can have gaps. It saves each event on it's own, which points to the next and previous event and each one can have a gap before, after and both.

MadLittleMods: > Background: Trixnity does not save pages of events, which can have gaps. It saves each event on it's own, which points to the next and previous event and each one can have a gap before, after and both.

@benkuly Interesting! I think your background point provides the most compelling argument here. Clients store events differently and I think we can accommodate for that.

Does the following make sense? This even seems a bit more simple to reason about and flexible.

We can omit `prev_pagination_token` or `next_pagination_token` when there isn't a gap on the respective side.

```mermaid
flowchart RL
    after[newest events...]:::gap-event -->|t10| fred -->|t9| waldo:::gap-event -->|t8| garply -->|t7| grault:::gap-event -->|t6| corge -->|t5| qux:::gap-event -->|t4| baz -->|t3| bar:::gap-event -->|t2| foo -->|t1| before[oldest events...]:::gap-event

    classDef gap-event opacity:0.8,fill:transparent;
```

`/messages?dir=b&from=t6` (notice `dir=b`, `backwards`)
```json5
{
  "chunk": [
    {
      "event_id": "$corge",
      "type": "m.room.message",
      "content": {
        "body": "corge",
      }
    },
    {
      "event_id": "$baz",
      "type": "m.room.message",
      "content": {
        "body": "baz",
      }
    },
    {
      "event_id": "$foo",
      "type": "m.room.message",
      "content": {
        "body": "foo",
      }
    }
  ]
  "gaps": [
        {
          "prev_pagination_token": "t6",
          "event_id": "$corge",
          "next_pagination_token": "t5",
        },
        {
          "prev_pagination_token": "t4",
          "event_id": "$baz",
          "next_pagination_token": "t3",
        },
        {
          "prev_pagination_token": "t2",
          "event_id": "$foo",
          "next_pagination_token": "t1",
        }
  ]
}
```

`/messages?dir=f&from=t6` (notice `dir=f`, `forwards`)
```json5
{
  "chunk": [
    {
      "event_id": "$garply",
      "type": "m.room.message",
      "content": {
        "body": "garply",
      }
    },
    {
      "event_id": "$fred",
      "type": "m.room.message",
      "content": {
        "body": "fred",
      }
    },
  ]
  "gaps": [
        {
          "prev_pagination_token": "t7",
          "event_id": "$garply",
          "next_pagination_token": "t8",
        },
        {
          "prev_pagination_token": "t9",
          "event_id": "$fred",
          "next_pagination_token": "t10",
        }
  ]
}
```

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Unstable prefix
+
+Servers will indicate support for this MSC via a `true` value for feature
+flag `org.matrix.msc3871` in `unstable_features` in the response to `GET
+/_matrix/client/versions`.
+
+While this feature is in development, the `gaps` field can be used as
+`org.matrix.msc3871.gaps`
+
```
MadLittleMods, line 243: This language is derived/inspired by [MSC2285](https://github.com/matrix-org/matrix-spec-proposals/pull/2285)

```suggestion

### While the MSC is unstable

During this period, to detect server support clients should check for the
presence of the `org.matrix.msc3871` flag in `unstable_features` on `/versions`.
Clients are also required to use the unstable prefixes (see [unstable
prefix](#unstable-prefix)) during this time.

### Once the MSC is merged but not in a spec version

Once this MSC is merged, but is not yet part of the spec, clients should rely on
the presence of the `org.matrix.msc3871.stable` flag in `unstable_features` to
determine server support. If the flag is present, clients are required to use
stable prefixes (see [unstable prefix](#unstable-prefix)).

### Once the MSC is in a spec version

Once this MSC becomes a part of a spec version, clients should rely on the
presence of the spec version, that supports the MSC, in `versions` on
`/versions`, to determine support. Servers are encouraged to keep the
`org.matrix.msc3871.stable` flag around for a reasonable amount of time
to help smooth over the transition for clients. "Reasonable" is intentionally
left as an implementation detail, however the MSC process currently recommends
*at most* 2 months from the date of spec release.
```

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,244 @@
...
+instead be more intelligent about backfilling in the background and just tell
+the client about the gap that they can retry fetching a little later.
+
+
+## Proposal
+
+Add a `gaps` field to the response of [`GET
+/_matrix/client/v3/rooms/{roomId}/messages`](https://spec.matrix.org/v1.1/client-server-api/#get_matrixclientv3roomsroomidmessages).
+This field is an array of `GapEntry` indicating where missing events in the
+timeline are as defined below.
```
ShadowJonathan, line 36: In the spec room, the idea to also apply this effort to `/context` came up, is this something that's meant to be considered in this MSC, or does this MSC not apply to `/context`?

MadLittleMods: I think it definitely applies to `/context` because it's another endpoint that returns a linearized version of the DAG.

It might make the proposal less clear if we have to bulk it up by specifying the same details for `/context` though. Perhaps would be better as a follow-up but I can add a note for future considerations in any case.

proposals/3871-gappy-timelines.md
MadLittleMods: Added to the MSC ⏩

proposals/3871-gappy-timelines.md
MadLittleMods: ^ Updated the MSC with the latest idea ⏩

Thanks for talking this through and alternative viewpoint @benkuly!

proposals/3871-gappy-timelines.md
MadLittleMods: Added ✅

proposals/3871-gappy-timelines.md
MadLittleMods: Added a future considerations section ⏩

If you have more opinions on whether we should actually expand on it in this MSC, feel free to comment further and unresolve this thread.

proposals/3871-gappy-timelines.md
MadLittleMods: @kegsay Anything more I can clarify here? Good to resolve?

kegsay: Yep, happy to resolve.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,438 @@
...
+
+
+## Alternatives
+
+As an alternative, we can continue to do nothing as we do today and not worry
+about the occasional missing events. People seem not to notice any missing
+messages anyway but they do probably see our slow `/messages` pagination.
+
+
+### Expose `prev_events` to the client
```
andrewzhurov, line 187: That would simplify things dramatically, seems.
Same approach of gap-filling that is used for server-server could be employed for client-server.

Additionally, would make it possible for _clients_ to issue full events, as [wondered in MSC4080](https://github.com/matrix-org/matrix-spec-proposals/pull/4080/files#diff-01bd0fbeae293c6ef66c18b9d5b652458c39eda5caf84f44ff4c16659c523eb0R534), since it becomes possible for clients to detect forward extremeties.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,438 @@
...
+Add a `gaps` field to the response of [`GET
+/_matrix/client/v3/rooms/{roomId}/messages`](https://spec.matrix.org/v1.1/client-server-api/#get_matrixclientv3roomsroomidmessages).
+This field is an array of `GapEntry` indicating where missing events in the
+timeline are as defined below.
+
+
+### 200 response
+
+This describes the new `gaps` response field being added to the `200 response`
+of `/messages`:
```
MadLittleMods, line 42: We should probably also add a `gaps` field to the `timeline` in [`/sync`](https://spec.matrix.org/v1.14/client-server-api/#get_matrixclientv3sync). It's important for clients to know wherever they might be missing a message.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,438 @@
...
+        {
+          "prev_pagination_token": "t7",
+          "event_id": "$garply",
+          "next_pagination_token": "t8",
+        },
+        {
+          "prev_pagination_token": "t9",
+          "event_id": "$fred",
+          "next_pagination_token": "t10",
+        }
```
MadLittleMods, line 166: (this is in the context of paginating forwards)

I'm not sure whether we should consider the gap in front of the latest events to actually be a gap in the `gaps` list.

The problem is that the client has no way to know whether it's just a forward extremity (one of the latest events in the room) to know whether it should display the gap in the timeline. I.e. we probably don't want to display a "gap" after the latest events in the room.

We should either omit the gap for forward extremities in the room or add a field to indicate it's a forward extremity.

MadLittleMods: I'm currently thinking that we should just omit the gap after the latest messages in the room. The `/messages` endpoint already returns `start`/`end` pagination tokens for people wanting to continue off either side. And this way clients can simply just display a gap wherever we say there is one without needing complex logic.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,438 @@
...
+
+
+### 200 response
+
+This describes the new `gaps` response field being added to the `200 response`
+of `/messages`:
+
+Name | Type | Description | required
+--- | --- | --- | ---
+`gaps` | `[GapEntry]` | A list of gaps indicating where events are missing in the `chunk` | no
```
MadLittleMods, line 46: For ease of processing, we should declare that the order of the gaps is the in the same direction as the `chunk` of messages.

proposals/3871-gappy-timelines.md
```diff
@@ -0,0 +1,438 @@
...
+      "content": {
+        "body": "foo",
+      }
+    }
+  ]
+  "gaps": [
+        {
+          "prev_pagination_token": "t6",
+          "event_id": "$corge",
+          "next_pagination_token": "t5",
```
MadLittleMods, line 115: Per https://spec.matrix.org/v1.18/appendices/#pagination, we should probably align these fields to be `prev_batch` and `next_batch`

PR #3872, opened 2022-08-20
MSC3872: Organisation of Spaces

Alex-Sour:
A proposal for Spaces to have categories for Rooms and a specified order for Rooms and categories that clients can display.

### DIFF
--- a/proposals/3872-space-organisation.md
+++ b/proposals/3872-space-organisation.md
@@ -1,0 +1,14 @@
+# MSC3872: Organisation of Spaces
+
+Spaces are incredibly disorganized, unlike other platforms, such as [Revolt](https://revolt.chat), [Discord](https://discord.com) and [Slack](https://slack.com). Matrix Spaces are more similar to lists of Rooms.
+
+To solve this issue, this MSC proposes new features to Spaces.
+
+## Proposal
+
+This MSC proposes Spaces to have:
+
+1. A specific order of rooms that clients are sent upon joining, and that can be updated at any time with the right permissions
+2. Categories for rooms (also with their own order) to help improve organization
+
+This should greatly complement Spaces, and would make organisation and navigation simpler, especially since clients could implement features such as collapsing categories, drag-and-drop reordering and more.


proposals/3872-space-organisation.md
```diff
@@ -0,0 +1,14 @@
...
+
+Spaces are incredibly disorganized, unlike other platforms, such as [Revolt](https://revolt.chat), [Discord](https://discord.com) and [Slack](https://slack.com). Matrix Spaces are more similar to lists of Rooms.
+
+To solve this issue, this MSC proposes new features to Spaces.
+
+## Proposal
+
+This MSC proposes Spaces to have:
+
+1. A specific order of rooms that clients are sent upon joining, and that can be updated at any time with the right permissions
```
SimonBrandner, line 11: This is already a thing: https://spec.matrix.org/unstable/client-server-api/#ordering-1

proposals/3872-space-organisation.md
```diff
@@ -0,0 +1,14 @@
...
+2. Categories for rooms (also with their own order) to help improve organization
```
SimonBrandner, line 12: Would these also apply to sub-spaces?

Alex-Sour: I am unfamiliar with these, but I believe so.

proposals/3872-space-organisation.md
Alex-Sour: Editing the order does not seem to reflect on other clients. This proposal is for a custom order managed by Space administrators, that is propagated to all clients.

Alex-Sour: From what I am reading on the page, it seems like Rooms are ordered automatically, and not by an order defined by the Space's administrators

Alex-Sour: I also found no way to manually order them on Element

Alex-Sour: Added clarification in a second commit to this point

SimonBrandner: That is because Element doesn't support this (neither setting it or reading it in the room-list) and instead orders by name/notif count...

Alex-Sour: So to clarify, Matrix currently supports a custom ordering of Rooms that can be altered manually without a specific sorting method (like sorting by name) and be sent to clients, similarly to Discord, Slack and Revolt?

proposals/3872-space-organisation.md
SimonBrandner: These are sub-spaces and they are sort of meant to function as categories (if you want that)

Alex-Sour: Actually, these are perfect, I'll edit my PR to remove categories, I think these work

proposals/3872-space-organisation.md
SimonBrandner: Yep, if you're an admin of a spaces, you can define in what order the rooms and sub-spaces appear in the clients. The client is also free to override this order, if it prefers to order the rooms in a different way (name, notifs...). The user can also manually re-order their rooms (using tags I bealive) though the interaction of this with spaces is undefined, AFAIK

N.B. that both ways of custom ordering mentioned here are not supported by Element (at least not in the room-list - element will respect the `order` field in the explore rooms view of a space) and Element only supports the name and notif ordering

Alex-Sour: This proposal is for a manual order that is not defined by tags or anything of that kind, but instead just a custom order, not defined by any sorting method whatsoever

SimonBrandner: What I meant by _ordering using tags_ is that within a given tag a user can define in what order the rooms should appear, so in theory Matrix supports custom ordering of rooms/sub-spaces by a space admin and also custom ordering of rooms independently of what a space admin wants the order to be. 

Neither of these is implemented in Element and their interaction is undefined

Alex-Sour: Yes, but this system is still more complicated as a client will likely make users need to open a tags menu one at a time and enter something to set the order of rooms/subspaces, in contrast to other platforms in which there is simply an array of channels/categories that are displayed as in the order.

Alex-Sour:
Are there any further problems with this PR

proposals/3872-space-organisation.md
```diff
@@ -0,0 +1,11 @@
+# MSC3872: Order of rooms and subspaces in Spaces
+
+Spaces are incredibly disorganized, unlike other platforms, such as [Revolt](https://revolt.chat), [Discord](https://discord.com) and [Slack](https://slack.com).
+
+To solve this issue, this MSC proposes a custom, manual order for rooms and subspaces in Spaces.
+
+## Proposal
+
+This MSC proposes Spaces to have a specific, custom order of rooms and subspaces that clients are sent upon joining, and that can be updated at any time with the right permissions, which can then be reflected on other clients
```
turt2live, line 9: Lines should be wrapped to 120 characters.

proposals/3872-space-organisation.md
```diff
@@ -0,0 +1,11 @@
+# MSC3872: Order of rooms and subspaces in Spaces
```
turt2live, line 1: Not all commits are signed off on this PR: please rebase or add a comment with signoff.

turt2live, line 1: This MSC is missing key sections from the template which help answer questions like "why is https://spec.matrix.org/v1.3/client-server-api/#ordering-1 insufficient" and how this actually works.

I'm converting this to a draft (WIP) MSC pending these changes.

Alex-Sour:
Commit bc589d8fad66f954363bc72188cb71564c1922f2 Signed-off-by: Alex LeBlanc <alexsour@protonmail.com>

proposals/3872-space-organisation.md
Alex-Sour: Commit bc589d8fad66f954363bc72188cb71564c1922f2 Signed-off-by: Alex LeBlanc <alexsour@protonmail.com>

I also added an inline comment with signoff

Alex-Sour: I had not seen the template, but after some digging, I found it.

I have updated the file accordingly.

proposals/3872-space-organisation.md
Alex-Sour: Wrapped

proposals/3872-space-organisation.md
```diff
@@ -0,0 +1,42 @@
...
+
+To solve this issue, this MSC proposes a custom, manual order for rooms and subspaces in Spaces.
+
+## Proposal
+
+This MSC proposes Spaces to have a specific, custom order of rooms and subspaces that clients are sent upon joining,
+and that can be updated at any time with the right permissions, which can then be reflected on other clients.
+
+This should greatly complement Spaces, and would make organisation and navigation simpler, especially since clients
+could implement features such as drag-and-drop reordering.
```
t3chguy, line 14: Clients can already do this, Element Web/Desktop already have the appropriate algorithm for re-ordering top-level spaces, applying the same to space children would be trivial without considering dnd to drag between spaces

proposals/3872-space-organisation.md
clokep: After reading through this it still sounds almost identical to what is currently specced (as @turt2live already linked to).

I'm not sure if any clients allow you to manually change that ordering, but Element Web/iOS/Android and Synapse/Dendrite should respect the configured ordering.

turt2live: (Element doesn't have user-accessible ways of making the room list manually ordered, but the space panel respects the ordering)

t3chguy: Element Web/Desktop also *mostly* respect the order in the space home/hierarchy page, with the exception of ordering subspaces before rooms

PR #3873, opened 2022-08-21
MSC3873: event_match dotted keys

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/event-match-dotted-keys/proposals/3873-event-match-dotted-keys.md)

Implementations:

* mautrix-go: https://github.com/mautrix/go/commit/61b0553ae9883bf8d7ce8f099b7da10e7ecac402
* Synapse: matrix-org/synapse#15004
* matrix-js-sdk: matrix-org/matrix-js-sdk#3134

Fixes https://github.com/matrix-org/matrix-spec/issues/648

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3873#issuecomment-1421617428)

### DIFF
--- a/proposals/3873-event-match-dotted-keys.md
+++ b/proposals/3873-event-match-dotted-keys.md
@@ -4,7 +4,7 @@ The current specification of [`event_match`] describes the parameter
 `key` as
 
 > `key`: The dot-separated path of the property of the event to match,
-> e.g. `content.body`.
+> e.g. `content.body`.
 
 It does, however, not clarify how to handle collisions such as in
 
@@ -16,17 +16,28 @@ It does, however, not clarify how to handle collisions such as in
 where it is unclear which field the dot-separated path `m.foo` should
 match ([#648]).
 
-While collisions are often not a practical problem, the ambiguity in the
-specification leads to incompatible implementations as evidenced by
-[matrix-org/matrix-js-sdk#1454]. The current proposal resolves the
-ambiguity by leveraging the existing solution for the same problem in
-the [filter API].
+Previously collions are not often a practical problem, but as dotted-field names
+have become more common in Matrix, e.g. `m.relates_to` or [MSC1767]-style
+extensible events ,this ambiguity is no longer satisfactory.
+
+The ambiguity in the specification leads to incompatible implementations as
+evidenced by [matrix-org/matrix-js-sdk#1454]. The current proposal resolves the
+ambiguity by leveraging the existing solution for the same problem used by the
+`event_fields` of [filters]:
+
+> List of event fields to include. If this list is absent then all fields are included.
+> The entries may include ‘.’ characters to indicate sub-fields. So [‘content.body’]
+> will include the ‘body’ field of the ‘content’ object. A literal ‘.’ character
+> in a field name may be escaped using a ‘\’.
 
 ## Proposal
 
-The . character in the `key` parameter is changed to be exclusively
-reserved for field separators. Any literal key in field names is to be
-escaped using a \\.
+The dot (`.`) character in the `key` parameter is changed to be exclusively
+reserved for field separators. Any literal dot in field names are to be
+escaped using a backslash (`\.`) and any literal backslashes are also escaped with
+a backslash (`\\`). A backslash before any other character has no special meaning
+and is left as-is, but it is recommended that implementations do not redundantly
+escape characters, as they may be used for escape sequences in the future.
 
 Revisiting the example from above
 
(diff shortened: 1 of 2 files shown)


richvdh:
I feel like we should just do this, if we aren't going to set fire to pushrules any time soon. It's such a mess currently.

I see that @tulir has implemented it in mautrix, but to have confidence I'd rather like to see more implementations (including server-side ones). Any volunteers?

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,96 @@
...
+specification leads to incompatible implementations as evidenced by
+[matrix-org/matrix-js-sdk#1454]. The current proposal resolves the
+ambiguity by leveraging the existing solution for the same problem in
+the [filter API].
+
+## Proposal
+
+The . character in the `key` parameter is changed to be exclusively
+reserved for field separators. Any literal key in field names is to be
+escaped using a \\.
```
richvdh, line 29: Presumably a double-backslash (`\\`) in `key` should match a single backslash in a field name?

what does `\x` match? `x`? or `\x` ? Or will it never match?

clokep: It looks like the filter API which is referred to above is similarly vague:

>  The entries may include ‘.’ characters to indicate sub-fields. So [‘content.body’] will include the ‘body’ field of the ‘content’ object. A literal ‘.’ character in a field name may be escaped using a ‘\’.

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,96 @@
...
+As mentioned above, this exact solution is already employed in the
+[filter API]. Reusing it here, therefore, increases the
+specification’s coherence.
+
+## Backwards compatibility
+
+In order to provide partial backwards compatibility, implementations
+should continue supporting unescaped dot-separated paths in situations
+where they are collision-free. In other words, `"key": "m.foo"` should
+continue matching both
```
clokep, line 51: In the ambigious case what should be done? Match the "correct" case only?

You can fairly easily concoct examples which make this quite difficult to implement. E.g. for `"key": "m.foo.bar"`:

```json
{
  "m": { "foo.bar": "abc" },
  "m.foo": { "bar": "xyz" }
}
```

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,96 @@
...
+but not
+
+    {
+      "m": { "foo": "bar" },
+      "m.foo": "baz"
+    }
+
+It is recommend that implementations maintain this fallback for a
+certain transition period, e.g. one year, to give clients a chance to
+update any affected push rules.
```
clokep, line 72: There aren't any default push rules affected by this, I don't think? I'd be more tempted to pull the band-aid off and just declare you need to escape them. Are there particular rules we think would be affected?

proposals/3873-event-match-dotted-keys.md
Johennes: > Presumably a double-backslash (`\\`) in `key` should match a single backslash in a field name?

That seems sensible to me and would match how e.g. JavaScript treats backslashes in strings.

> what does `\x` match? `x`? or `\x` ? Or will it never match?

This is somewhat less clear to me. If we consider `\x` to mean "escape `x`", then I suppose `\x` and `x` should be identical since `x` doesn't require escaping?

proposals/3873-event-match-dotted-keys.md
Johennes: Right and I think that means we can only match _nothing_ in the ambiguous case?

proposals/3873-event-match-dotted-keys.md
Johennes: No, I think you're right. The [issue](https://github.com/matrix-org/matrix-js-sdk/issues/1454) that triggered this was about a custom push rule.

proposals/3873-event-match-dotted-keys.md
clokep: I'm not sure if @richvdh chose it on purpose or not, but `\x` is a rather strange choice as (at least) Python and JavaScript won't even let you put that in a string since it is an invalid hex escape sequence. 😄 

> This is somewhat less clear to me. If we consider `\x` to mean "escape `x`", then I suppose `\x` and `x` should be identical since `x` doesn't require escaping?

This is essentially what JavaScript does, Python you end up with both characters (essentially the `\` gets escaped automatically) `\x` -> `\\x`.

I think either is fine, we just need to define it.

proposals/3873-event-match-dotted-keys.md
clokep: > Right and I think that means we can only match _nothing_ in the ambiguous case?

Yeah, I think that's the only reasonable thing to do. 😢

proposals/3873-event-match-dotted-keys.md
clokep: Right! I have two proposed default push rules that would be affected (MSC3958 and MSC3752).

proposals/3873-event-match-dotted-keys.md
Johennes: If we don't have a use case, maybe we could decide based on what would be easier to implement?

clokep: It isn't quite the same, but it seems that [Synapse has tests](https://github.com/matrix-org/synapse/blob/bf82b56babc9e2cacba34f8878da3b3834914b3a/tests/push/test_push_rule_evaluator.py#L218-L238) for the `pattern` that the `\` is left as-is.

proposals/3873-event-match-dotted-keys.md
Johennes: So I guess the only thing to ensure here is that the MSCs you mention don't land _before_ this one? If we could achieve that, the whole backwards compatibility section could be removed.

proposals/3873-event-match-dotted-keys.md
richvdh: > So then could say `\` stays `\` except for when it's followed by `.` in which case `\.` is to interpreted as literal `.`? This sounds ok to me.

So double-backslash matches exactly that, double-backslash? How then does one match `\.` in a key? Not that that is a terribly likely thing to need to do but if we're designing an escaping system it would be nice if it could actually represent the complete pattern space.

IMHO it would b better to make a second exception for `\` followed by `\`, with a recommendation that clients always escape `\` if they are trying to match it, because the behaviour of `\<something else>` might change in the future.

Johennes: Hm, a literal `\.` couldn't be matched that way, you're right. Maybe instead of using two exceptions it would be easier to just say `\X` means a literal `X` – no matter what character `X` is?

richvdh: > Hm, a literal `\.` couldn't be matched that way, you're right. Maybe instead of using two exceptions it would be easier to just say `\X` means a literal `X` – no matter what character `X` is?

I have no strong opinion on whether `\X` should match `\X` or `X`, other than that:
 * we should pick one and spec it
 * we should recommend that clients do not redundantly escape characters, to leave us room to extend the range of escape sequences in future.

(I suggest we leave a decision to @clokep based on whatever turns out to be easiest to implement)

proposals/3873-event-match-dotted-keys.md
clokep: > So I guess the only thing to ensure here is that the MSCs you mention don't land _before_ this one? If we could achieve that, the whole backwards compatibility section could be removed.

That's my plan! 👍

proposals/3873-event-match-dotted-keys.md
Johennes: Oh, interesting! I think the only thing to be careful of is the order of replacements (`/` before `.`)?

clokep: > Oh, interesting! I think the only thing to be careful of is the order of replacements (`/` before `.`)?

Yes, I think that's the only tricky bit! 👍 I'll take a look at the matrix-js-sdk (and double check mautrix-go's implementation) to be sure they're all compatible.

This *doesn't* implement the backwards compatibility paths, but that *might* be doable.

proposals/3873-event-match-dotted-keys.md
clokep: 68542ef3563e0db184e98cce4cf6cc1807579c34 talks about this a bit.

proposals/3873-event-match-dotted-keys.md
clokep: I believe I clarified all of this in 03ddc0a4e8da0d00284f110bc5209e16cf32957f.

turt2live:
@clokep believes this is ready for FCP.

Implementations have not been checked as of this comment.

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,111 @@
+# MSC3873: event_match dotted keys
+
+The current specification of [`event_match`] describes the parameter
+`key` as
+
+> `key`: The dot-separated path of the property of the event to match,
+> e.g. `content.body`.
+
+It does, however, not clarify how to handle collisions such as in
```
richvdh, line 9:
```suggestion
It does not, however, clarify how to handle collisions such as in
```

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+    {
+      "m": { "foo": "bar" },
+      "m.foo": "baz"
+    }
+
+where it is unclear which field the dot-separated path `m.foo` should
+match ([#648]).
+
+Previously collions are not often a practical problem, but as dotted-field names
```
richvdh, line 19:
```suggestion
Previously collisions are not often a practical problem, but as dotted-field names
```

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,111 @@
...
+have become more common in Matrix, e.g. `m.relates_to` or [MSC1767]-style
+extensible events ,this ambiguity is no longer satisfactory.
```
richvdh, line 21:
```suggestion
extensible events, this ambiguity is no longer satisfactory.
```

richvdh:
Synapse and EW impls look good to me.

@mscbot fcp merge

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+    {
+      "m": { "foo": "bar" },
+      "m.foo": "baz"
+    }
+
+where it is unclear which field the dot-separated path `m.foo` should
+match ([#648]).
+
+Previously collisions are not often a practical problem, but as dotted-field names
```
KitsuneRal, line 19:
```suggestion
Previously collisions were not often a practical problem, but as dotted-field names
```

proposals/3873-event-match-dotted-keys.md
```diff
@@ -0,0 +1,111 @@
...
+
+This may break custom push rules that users have configured, but it is asserted
+that those are broken anyway, as mentioned above (see [matrix-org/matrix-js-sdk#1454]).
+
+## Alternatives
+
+Alternatives to the current proposal are to use [JSON pointers] or [JSONPath]. While
+being more versatile than the simplistic escaping proposed here, these are
+unnecessary and break backwards compatibility for *all* existing `event_match`
+conditions.
```
KitsuneRal, line 94: Ok, I tend to agree that JSONPath is a bit of an overkill. But how will we tackle arrays in MSC1767 `m.text` content blocks - for example, if someone would want to detect usage of a certain MIME type?

KitsuneRal (commented):
One question

proposals/3873-event-match-dotted-keys.md
clokep: The only use-case I know of for searching inside of arrays right now is to see if a value is contained with an array, which I wrote MSC3966 for; it doesn't work inside of an object inside of an array, however.

KitsuneRal: Thanks. Resolving this in anticipation that the arrays lookup will be enhanced in the context of further work on extensible events.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1464

turt2live:
Merged 🎉

PR #3874, opened 2022-08-23
MSC3874: Filtering threads from the /messages endpoint

justjanne:


proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
+# MSC3874 Loading Messages excluding Threads
+
+## Motivation
+
+In our beta deployment of threads, it's become a noticeable issue where a room may have activity in a long-running
```
gsouquet, line 5: In Element implementation of threads.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
...
+thread while the main timeline is inactive. If a user starts their client and tries to paginate through the main
+timeline, loading will feel sluggish, as lots of events have to be loaded that won’t be displayed.
```
gsouquet, line 7: I believe you're alluding to it, but the main motivation is to improve the efficiency of what is sent down the network and to avoid unecessery requests

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Allow filtering the /messages API to not include threaded messages
+
+This proposal recommends extending the existing [Event filters] are extended with a new filter, named
+`not_related_by_rel_types`, which acts exactly like the opposite of the existing `related_by_rel_types` filter.
+
+This means, if this filter is specified, only message which match none of the given relation types will be returned.
+
+```
+GET /_matrix/client/v3/rooms/!room_id:domain/messages?filter=...
```
gsouquet, line 19: I believe we might want to use `unstable` instead of `v3` ?

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
...
+formatted for legibility:
+
+```jsonc
+{
+  "types": ["m.room.message"],
+  "not_related_by_rel_types": ["m.thread"]
+}
+```
+
+Note that the newly added filtering parameters return events based on information  in related events. Consider the
```
gsouquet, line 32: double space after "information"

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
...
+following events in a room:
+
+* `A`: a `m.room.message` event sent by `alice`
+* `B`: a `m.room.message` event sent by `bob`
+* `C`: a `m.room.message` event sent by `charlie` which relates to `A` with type `m.thread`
+
+Using a filter of `"not_related_by_rel_types": ["m.thread"]` would return only event `B` as  it has no event which
```
gsouquet, line 39: double space after `B` as.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,75 @@
...
+```jsonc
+{
+  "unstable_features": {
+    "org.matrix.msc3874": true,
+    // ...
+  }
+}
+```
+
+## Potential issues
```
gsouquet, line 56: I think we need to add a section about limitations. This MSC will not exclude events that are related to an excluded `m.thread` event. So we might end up with zombie reactions that we will have to throw down the pipe.

As discussed previously, this is something that could have real performance impact. We could give a mention to multi relations as this would make this problem much simpler (there's an MSC related to that)

gsouquet (commented):
Great job for a first draft of an MSC!! Minor things to update, but this looks really good to me as a starting point

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,82 @@
+# MSC3874 Loading Messages excluding Threads
+
+## Motivation
+
+In Element's beta implementation of threads, it's become a noticeable issue where a room may have activity in a 
+long-running thread while the main timeline is inactive. If a user starts their client and tries to paginate through the 
+main timeline, loading will feel sluggish, as lots of events have to be loaded that won’t be displayed.
+
+This proposal would allow reducing the number of requests and amount of data transmitted.
```
clokep, line 9: Might be worth highlighting that this is especially critical for mobile apps!

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,82 @@
...
+main timeline, loading will feel sluggish, as lots of events have to be loaded that won’t be displayed.
+
+This proposal would allow reducing the number of requests and amount of data transmitted. 
+
+## Proposal
+
+### Allow filtering the /messages API to not include threaded messages
+
+This proposal recommends extending the existing [Event filters] are extended with a new filter, named
+`not_related_by_rel_types`, which acts exactly like the opposite of the existing `related_by_rel_types` filter.
```
clokep, line 16: I don't believe this is accurate: `related_by_rel_types` filters on the opposite end of the relationship than we care about (I think). The logic of `related_by_rel_types` is confusing, quoting from MSC3440:

> Note that the newly added filtering parameters return events based on information in related events. Consider the following events in a room:
>
> * `A`: a `m.room.message` event sent by `alice`
> * `B`: a `m.room.message` event sent by `bob` which relates to `A` with type `m.thread`
>
> Using a filter of `"related_by_rel_types": ["m.thread"]` would return event `A` as it has another event which relates to it via `m.thread`.

`not_related_by_rel_types` I think would mean "find all events which do not have another event related to them via a thread". So using the above this would return `B` only?

I think what we really want is all events which do not have a `rel_type` of `m.thread`, so we can add filtering by that: `"not_rel_types": ["m.thread"]`. In the above example this would filter out `B`, but keep `A` (or a theoreticaly `C` event which has no other events relating to it and doesn't related to any events).

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,82 @@
...
+
+This means, if this filter is specified, only message which match none of the given relation types will be returned.
+
+```
+GET /_matrix/client/unstable/org.matrix.msc3874/rooms/!room_id:domain/messages?filter=...
```
clokep, line 21: For MSCs we usually use the "final" URL endpoint and specify the unstable variants just at the bottom!

```suggestion
GET /_matrix/client/v3/rooms/!room_id:domain/messages?filter=...
```

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,82 @@
...
+- A suitable workaround, depending on the ratio of thread-messages compared to main timeline messages in a room, may be 
+  an increase of the page size
+
+## Dependencies
+
+- [MSC2674](https://github.com/matrix-org/matrix-doc/pull/2674)
+- [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675)
+- [MSC3567](https://github.com/matrix-org/matrix-doc/pull/3567)
+- [MSC3676](https://github.com/matrix-org/matrix-doc/pull/3676)
+- [MSC3440](https://github.com/matrix-org/matrix-doc/pull/344ß)
```
clokep, line 79: It is usually worth mentioning whether these are already accepted by the spec or not (as a way to identify risk with the proposal).

proposals/3874-messages-endpoint-threads-filter.md
justjanne: Uh so confusing! Thanks for the info, I had just followed a reviewer's suggestion blindly :see_no_evil:

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+this would allow clients significant usability improvements with threads.  
+
+## Proposal
+
+### Allow filtering the `/messages` API to not include threaded messages
+
+This proposal recommends extending the existing [Event filters] are extended with a new filter, named
+`not_rel_types`, which acts exactly like the opposite of the existing `rel_types` filter.
+
+This means, if this filter is specified, only message which match none of the given relation types will be returned.
```
clokep, line 19: `rel_types` isn't a filter:

```suggestion
This proposal recommends extending the existing [Event filters] are extended with a new filter, named
`not_rel_types`. If this filter is specified, only messages which match none of the given relation types
will be returned.
```

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+
+Note that the newly added filtering parameters return events based on information in related events. Consider the
+following events in a room:
+
+* `A`: a `m.room.message` event sent by `alice`
+* `B`: a `m.room.message` event sent by `bob`
+* `C`: a `m.room.message` event sent by `charlie` which relates to `A` with type `m.thread`
+
+Using a filter of `"not_rel_types": ["m.thread"]` would return only event `B` as it has no event which
+relates to it via `m.thread`.
```
clokep, line 43: I think this would return `A` and `B` (i.e. the thread root is part of the "main" timeline):

```suggestion
Using a filter of `"not_rel_types": ["m.thread"]` would return only events `A` and `B` as they do not
have a relation of `m.thread` in them.
```

proposals/3874-messages-endpoint-threads-filter.md
gsouquet:
```suggestion
- [MSC3440](https://github.com/matrix-org/matrix-doc/pull/344ß)
+ [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440)
```

turt2live:
This is actively undergoing implementation and needs sanity review to ensure it's on a good track. It is worth noting that this MSC is likely to accelerate quickly into FCP, so review which is more in line with that would be appreciated.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+following events in a room:
+
+* `A`: a `m.room.message` event sent by `alice`
+* `B`: a `m.room.message` event sent by `bob`
+* `C`: a `m.room.message` event sent by `charlie` which relates to `A` with type `m.thread`
+
+Using a filter of `"not_rel_types": ["m.thread"]` would return only events `A` and `B` as they do not have a relation of
+`m.thread` in them. Thread roots are returned in the same way as messages which are not part of threads at all.
+
+### Server capabilities
```
turt2live, line 45: This shouldn't be part of the proposal heading but rather its own "Unstable prefix" (or similar) heading, usually just above the dependencies heading.

We'd also need to mention that the filter key is `org.matrix.msc3874.not_rel_types` during this unstable period.

turt2live: For context: putting the unstable identifiers under "## Proposal" means we're saying that this bit needs to end up in the final spec prose, which is untrue: unstable identifiers/prefixes only go as far as the MSC document.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Potential issues
+
+This proposal moves the loading and processing of these hidden events onto the server. Depending on the server’s
+architecture, this may have a non-negligible performance impact. 
+
+## Limitations
+
+This proposal only considers events which have a direct relationship with the thread itself. Events such as reactions
+don’t, so they won’t be able to be filtered by this proposal.
```
turt2live, line 67:
```suggestion
While the client can effectively filter out noisy threads, it's not as easy to filter out events adjacent to threads such
as reactions. A more performant implementation is best left for a future MSC.
```

The threads mention caught me off guard, so have moved the understanding around a bit to tie it back to the intro a bit (and clarify that this MSC doesn't intend to solve the problem, which is a good thing imo).

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Alternatives
+
+- A suitable workaround, depending on the ratio of thread-messages compared to main timeline messages in a room, may be 
+  an increase of the page size
```
turt2live, line 72: This section can probably just be omitted from this MSC: the alternative would be to reinvent filtering, which is a non-starter for this MSC anyways.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,83 @@
...
+
+This proposal would allow reducing the number of requests and amount of data transmitted. Especially for mobile usage
+this would allow clients significant usability improvements with threads.  
+
+## Proposal
+
+### Allow filtering the `/messages` API to not include threaded messages
+
+This proposal recommends extending the existing [Event filters] are extended with a new filter, named `not_rel_types`.
+If this filter is specified, only messages which match none of the given relation types will be returned.
```
turt2live, line 17: It may be worth adding the complimentary `rel_types` to this MSC, or otherwise a sentence somewhere saying it's a future MSC's problem (and why).

turt2live (approved):
Generally this seems sane to me - looking forward to implementation :)

proposals/3874-messages-endpoint-threads-filter.md
clokep: I ended up implementing this with both `rel_types` and `not_rel_types` in matrix-org/synapse#14148 (the filtering classes in Synapse essentially require you to have both).

proposals/3874-messages-endpoint-threads-filter.md
clokep: I used `org.matrix.msc3874.rel_types` and `org.matrix.msc3874.not_rel_types` when implementing this for matrix-org/synapse#14148.

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,95 @@
...
+```jsonc
+{
+  "types": ["m.room.message"],
+  "not_rel_types": ["m.thread"],
+  "rel_types": ["m.edit"]
+}
+```
+
+Note that the newly added filtering parameters return events based on information in related events. Consider the
+following events in a room:
```
clokep, line 40:
```suggestion
The newly added filtering parameters return events based on information in the `m.relates_to` field. Consider the following events in a room:
```

proposals/3874-messages-endpoint-threads-filter.md
```diff
@@ -0,0 +1,95 @@
...
+
+* `A`: a `m.room.message` event sent by `alice`
+* `B`: a `m.room.message` event sent by `bob`
+* `C`: a `m.room.message` event sent by `charlie` which relates to `A` with type `m.thread`
+
+Using a filter of `"not_rel_types": ["m.thread"]` would return only events `A` and `B` as they do not have a relation of
+`m.thread` in them. Thread roots are returned in the same way as messages which are not part of threads at all.
```
clokep, line 47: I don't think we need to mention the thread root specifically here?

```suggestion
Using a filter of `"not_rel_types": ["m.thread"]` would return only events `A` and `B` as they do not have a relation of
`m.thread` in them.
```

PR #3879, opened 2022-09-05
MSC3879: Trusted key forwards

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/trusted_key_forwards/proposals/3879-trusted-key-forwards.md)

### DIFF
--- a/proposals/3879-trusted-key-forwards.md
+++ b/proposals/3879-trusted-key-forwards.md
@@ -19,15 +19,16 @@ property is set to `true` if the forwarder believes the session should be
 trusted.  If the property is set to `false` or is absent, this indicates that
 the session cannot be trusted.
 
-The recipient can trust the session if:
+The recipient should trust the session if and only if all of the following is true:
 
 - the `m.forwarded_room_key` event marks the key as trusted,
-- the recipient and forwarder belong to the same user, and
+- the recipient and forwarding devices belong to the same user, and
 - the forwarder has been verified (e.g. by verifying the device directly, or
   via cross-signing).
 
 Note that this depends on the forwarder being able to determine whether it
-trusts the session.  The forwarder can mark the session as trusted if:
+trusts the session.  The forwarder can mark the session as trusted if any of
+the following is true:
 
 - it created the session;
 - it received the session from the session creator via an `m.room_key` message;
@@ -35,7 +36,7 @@ trusts the session.  The forwarder can mark the session as trusted if:
   to the same user, and it was marked as trusted; or
 - it was received from a key backup that can be trusted (such as [Symmetric key
   backup](https://github.com/matrix-org/matrix-spec-proposals/pull/3270)), and
-  was marked as trusted.
+  was marked as trusted in the backup.
 
 ## Potential issues
 


proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,74 @@
...
+not to trust the session.
+
+## Proposal
+
+A new property, `trusted`, is added to the `m.forwarded_room_key` event.  The
+property is set to `true` if the forwarder believes the session should be
+trusted.  If the property is set to `false` or is absent, this indicates that
+the session cannot be trusted.
+
+The recipient can trust the session if:
```
dkasak, line 22:
```suggestion
The recipient should trust the session iff all of the following is true:
```

Let's use a bit more normative language ("should" rather than "can") since we believe these conditions to be correct. Also "iff" or "if and only if" (should the former be too obscure).

proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,74 @@
...
+
+- the `m.forwarded_room_key` event marks the key as trusted,
+- the recipient and forwarder belong to the same user, and
```
dkasak, line 25:
```suggestion
- the recipient and forwarding devices belong to the same user, and
```

I think this is the intended meaning and I find it clearer to be explicit here.

proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,74 @@
...
+Note that this depends on the forwarder being able to determine whether it
+trusts the session.  The forwarder can mark the session as trusted if:
+
+- it created the session;
+- it received the session from the session creator via an `m.room_key` message;
+- it was received via an `m.forwarded_room_key` from a trusted device belonging
+  to the same user, and it was marked as trusted; or
+- it was received from a key backup that can be trusted (such as [Symmetric key
+  backup](https://github.com/matrix-org/matrix-spec-proposals/pull/3270)), and
+  was marked as trusted.
```
dkasak, line 38:
```suggestion
  was marked as trusted in the backup.
```

proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,74 @@
...
+
+The recipient can trust the session if:
+
+- the `m.forwarded_room_key` event marks the key as trusted,
+- the recipient and forwarder belong to the same user, and
+- the forwarder has been verified (e.g. by verifying the device directly, or
+  via cross-signing).
+
+Note that this depends on the forwarder being able to determine whether it
+trusts the session.  The forwarder can mark the session as trusted if:
```
dkasak, line 30:
```suggestion
trusts the session.  The forwarder can mark the session as trusted if any of
the following is true:
```

proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,75 @@
+# MSC3879: Trusted key forwards
+
+When a device does not have a megolm session, it can [request the session from
+another device](https://spec.matrix.org/unstable/client-server-api/#key-requests).
+However, when receiving sessions in this way, the device currently does not
+know whether the session should be trusted.  As a result, some clients indicate
+that the events encrypted with this session cannot be trusted, which causes
+confusion for users.
+
+We propose adding a flag to the `m.forwarded_room_key` event to indicate
```
BillCarsonFr, line 10: Maybe clarify that it's added to the event content

proposals/3879-trusted-key-forwards.md
```diff
@@ -0,0 +1,75 @@
+# MSC3879: Trusted key forwards
```
uhoreg, line 1: As mentioned elsewhere, the term "trusted" isn't a great term to use.  It is used to mean various things in various places.  We'll probably switch to another term.

The basic premise of this MSC is to provide a method for the key forwarder to indicate whether it believes that the room key was actually sent by the given sender_key.  It is still up to the recipient of the key forward to determine whether the sender_key is "trusted" (e.g. by checking whether they have been verified)

richvdh: > As mentioned elsewhere, the term "trusted" isn't a great term to use. It is used to mean various things in various places. We'll probably switch to another term.

My objection to "trusted" isn't particularly about the term itself, but rather that it tends to get used without any additional context (what does it mean for a device to be "trusted"?). Provided it's clear from the context what we actually mean by "trusted", it's fine.

uhoreg: I think that "authenticated" would be a better term than "trusted".  I used "authenticated" for MSC4048, and I think we should use it here too.  In either case, both should use the same name.

PR #3881, opened 2022-09-05
# MSC9999: Remotely toggle push notifications for another client

kerryarchibald:


proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+## Proposal
+
+### Pusher-dependent clients
+
+#### Disabling pushers
+A new nullable field `is_disabled` is added to the `Pusher` model.
+
+| Name | Type | Description |
+|------|------|-------------|
+| `is_disabled` | boolean | Whether the pusher should actively create push notifications
```
babolivier, line 25: It would be more consistent with the rest of the spec to call this attribute `enabled` (similar to what already exists for push rules). This would also better match the description (imo).

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+
+In [POST /_matrix/client/v3/pushers/set](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3pushersset)
+the value is optional and if omitted, defaults to `false`.
+
+In [GET /_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers) the value
+is always returned.
+
+Pushers that are disabled do not send
+[`/notify`](https://spec.matrix.org/v1.3/push-gateway-api/#post_matrixpushv1notify) requests to push providers, or
+produce email notifications.
```
babolivier, line 35: > or produce email notifications.

Isn't this proposal supposed to only focus on HTTP pushers?

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+/_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers).
+
+| Name | Type | Description |
+|------|------|-------------|
+| `device_id` | string | **Required.** The device_id of the session that registered the pusher
+
+
+To be able to remove Pushers when sessions are deleted home servers must have some existing way to link a session to
+pusher, so exposing the `device_id` on http pushers should be trivial. (Synapse, for instance, stores the [access
+token](https://github.com/matrix-org/synapse/blob/develop/synapse/storage/databases/main/pusher.py#L487) when adding a
```
babolivier, line 48:
```suggestion
token](https://github.com/matrix-org/synapse/blob/3d201151152ca8ba9b9aae8da5b76a26044cc85f/synapse/storage/databases/main/pusher.py#L487) when adding a
```

(this way the link doesn't break next time that file is changed)

proposals/3881-remote-push-notification-toggling.md
Johennes: Could this complicate backwards compatibility for existing pushers (which should all be enabled)? I think part of the reason for designing it this way around may have been that a missing `disabled` field can be interpreted as falsy.

babolivier: Maybe (but not in a big way imo). Part of the rationale is also that field names that describe a negative increases the complexity of handling it (because you're eg checking if the pusher is not not enabled instead of checking that it's enabled, if that makes sense), so we try to avoid those.
In terms of backwards compatibility, it wouldn't be the first time implementations such as Synapse need to backfill a new field for a given data structure (Synapse does that by updating database rows in the background), so I think it's fine.

proposals/3881-remote-push-notification-toggling.md
kerryarchibald: The proposal solves issues unique to HTTP pushers, but the generic `enabled` flag should apply to all notifications as it will be set on all pushers.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,115 @@
...
+
+| Name | Type | Description |
+|------|------|-------------|
+| `device_id` | string | **Required.** The device_id of the session that registered the pusher
+
+
+To be able to remove Pushers when sessions are deleted home servers must have some existing way to link a session to
+pusher, so exposing the `device_id` on http pushers should be trivial. (Synapse, for instance, stores the [access
+token](https://github.com/matrix-org/synapse/blob/3d201151152ca8ba9b9aae8da5b76a26044cc85f/synapse/storage/databases/main/pusher.py#L487)
+when adding a pusher)
```
babolivier, line 49: Some more info on this. Synapse does store access tokens, but even though an access token is usually associated with a device, it's not always the case. For example, the [login as user admin API](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#login-as-a-user) does not create a new device and as such the access token is not associated with an existing device.

Currently it doesn't look like there's any pusher with an access token that's not associated with an existing device on matrix.org:

```
matrix=> SELECT COUNT(*) FROM pushers INNER JOIN access_tokens ON pushers.access_token = access_tokens.id WHERE access_tokens.device_id IS NULL;
 count 
-------
     0
(1 row)
```

But it's technically possible so the above does not mean it doesn't exist.

It also looks like other server implementation such as Dendrite don't necessarily store a device ID or access token for pushers.

So as a __TL;DR__ is that I think this should be a `null`-able field.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,115 @@
...
+To be able to remove Pushers when sessions are deleted some home servers have some existing way to link a session to
+pusher, so exposing the `device_id` on http pushers should be trivial. (Synapse, for instance, stores the [access
+token](https://github.com/matrix-org/synapse/blob/3d201151152ca8ba9b9aae8da5b76a26044cc85f/synapse/storage/databases/main/pusher.py#L487)
+when adding a pusher, which is usually associated with a device)
+
+In [GET /_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers) the value
+is required when `kind` is `http`. If `kind` is _not_ `http`, the `device_id` field is null.
+
+In [POST /_matrix/client/v3/pushers/set](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3pushersset)
+`device_id` is an invalid parameter and should raise an invalid request error.
```
babolivier, line 55: Currently the Matrix spec does not mention what a homeserver should do if it receives a request parameter it does not understand. It is left as an implementation detail - for instance Synapse will just ignore such parameters. It feels weird to do it for this specific parameter on this specific endpoint, so I would just drop this paragraph.

babolivier (commented):
Some drive-by comments as I'm implementing the `device_id` field in Synapse.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,115 @@
...
+| `device_id` | string | **Nullable.** The device_id of the session that registered the pusher
+
+
+To be able to remove Pushers when sessions are deleted some home servers have some existing way to link a session to
+pusher, so exposing the `device_id` on http pushers should be trivial. (Synapse, for instance, stores the [access
+token](https://github.com/matrix-org/synapse/blob/3d201151152ca8ba9b9aae8da5b76a26044cc85f/synapse/storage/databases/main/pusher.py#L487)
+when adding a pusher, which is usually associated with a device)
+
+In [GET /_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers) the value
+is required when `kind` is `http`. If `kind` is _not_ `http`, the `device_id` field is null.
```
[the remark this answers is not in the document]
Johennes: I think maybe there was a semantic reason. For `http` pushers, the client will treat `device_id` as the device receiving the notifications whereas for `email` pushers that doesn't work.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,115 @@
...
+
+The current proposal solves this problem by making the connection between HTTP pushers and Matrix devices explicit and
+assigning an enabled state to every pusher.
+
+## Proposal
+
+### Pusher-dependent clients
+
+#### Disabling pushers
+A new nullable field `enabled` is added to the `Pusher` model.
```
DMRobertson, line 21: `s/nullable/optional/ here?

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+### Pusher-less clients
+
+Pausing notifications for clients that create notifications outside of the Push Gateway will not be addressed in this
+MSC.
+
+## Migration
+
+Clients that connect to a home server that doesn't yet support this proposal should interpret a missing `enabled` value
```
gsouquet, line 65: How can client determine whether a home server supports this MSC or not?

I believe adding `org.matrix.msc3881` and `org.matrix.msc3881.stable` in the response of `/_matrix/client/versions` would be fairly appropriate here

gsouquet:
This spec proposal now has a client implementation too, https://github.com/matrix-org/matrix-js-sdk/pull/2686

proposals/3881-remote-push-notification-toggling.md
babolivier: Very good point. I've preemptively opened https://github.com/matrix-org/synapse/pull/13860 so that the Synapse implementation includes this flag.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+[`/notify`](https://spec.matrix.org/v1.3/push-gateway-api/#post_matrixpushv1notify) requests to push providers for
+`http` pushers or otherwise.
+
+#### Explicitly linking device and pusher
+A new optional field `device_id` is added to the `Pusher` model as returned from [GET
+/_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers).
+
+| Name | Type | Description |
+|------|------|-------------|
+| `device_id` | string | **Optional** The device_id of the session that registered the pusher
```
gsouquet, line 43: I am currently failing to link a pusher with a device on a client. I originally thought `device_id` would be what I need, but the description here is making me perplexed. 

My current understanding of the `device_id` field: 
Given I am logged in on web `device_id=laptop` and create a pusher for my iPhone `device_id=mobile`. The pusher that will be returned to me will have `device_id=laptop` as the pusher was registered from my web session.


Is my understanding of the description correct? If it is, I'm not sure we can reconcile a pusher with a device client-side.

Johennes: For HTTP pushers, the pusher itself can usually only be created by the device that is supposed to receive the notifications because it requires the push key (e.g. an APNS token) which normally the client receives from the platform / operating system.

So in your example, the iPhone would create the pusher and set `device_id=mobile` and the web client would merely toggle the pusher's `enabled` flag.

Johennes: A corollary of this is that (unless you sync the pushkey across devices) a client can only toggle on push notifications for another client if that other client has previously registered for push notifications itself.

gsouquet: Ah right! I was missing that piece of information, makes sense
Thank you

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+is optional.
+
+### Pusher-less clients
+
+Pausing notifications for clients that create notifications outside of the Push Gateway will not be addressed in this
+MSC.
+
+## Migration
+
+Home servers should indicate their support of this MSC by adding `org.matrix.msc3881` and `org.matrix.msc3881.stable` in
```
mnaturel, line 61: I guess we should also explain the difference between the 2 values: with or without `stable` prefix. Or add a link to a spec/doc explaining this difference.

The `stable` suffix seems to indicate that stable fields for the feature are now exposed from the server and can now be used in clients. This enables client to know whether it should use the stable or unstable fields.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+#### Explicitly linking device and pusher
+A new optional field `device_id` is added to the `Pusher` model as returned from [GET
+/_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers).
+
+| Name | Type | Description |
+|------|------|-------------|
+| `device_id` | string | **Optional** The device_id of the session that registered the pusher
+
+
+To be able to remove Pushers when sessions are deleted some home servers have some existing way to link a session to
```
turt2live, line 46:
```suggestion
To be able to remove Pushers when sessions are deleted some homeservers have some existing way to link a session to
```

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+
+Pausing notifications for clients that create notifications outside of the Push Gateway will not be addressed in this
+MSC.
+
+## Migration
+
+Home servers should indicate their support of this MSC by adding `org.matrix.msc3881` and `org.matrix.msc3881.stable` in
+the response of `/_matrix/client/versions`.
+
+Clients that connect to a home server that doesn't yet support this proposal should interpret a missing `enabled` value
```
turt2live, line 64:
```suggestion
Clients that connect to a homeserver that doesn't yet support this proposal should interpret a missing `enabled` value
```

turt2live: applies throughout

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+
+In [GET /_matrix/client/v3/pushers](https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3pushers) the value
+is optional.
+
+### Pusher-less clients
+
+Pausing notifications for clients that create notifications outside of the Push Gateway will not be addressed in this
+MSC.
+
+## Migration
```
turt2live, line 59: This section should not be under the "Proposal" heading, but rather the "Unstable prefix" heading: compatibility is determined by spec version support, not by unstable flags

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+as `true`.
+
+Home servers should migrate pushers that were registered before this proposal so that `enabled` is `true`
+
+## Potential issues
+
+Adding an enabled state to pushers increases the complexity of the push notification API. In addition to a pusher
+existing or not existing, implementations now have to also evaluate the pusher's `enabled` field.
+
+## Alternatives
```
turt2live, line 74: Per-device push rules would also be interesting here, as it would mean disabling (enabling? I can never remember) the master rule for the device.

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
+# MSC3881: Remotely toggling push notifications for another client
```
turt2live, line 1: This MSC describes that there is a problem, but not neccessarily how the problem affects users: would it be possible to get some words to describe what a client might want to do with this feature?

proposals/3881-remote-push-notification-toggling.md
```diff
@@ -0,0 +1,114 @@
...
+
+The [push notification API](https://spec.matrix.org/v1.3/client-server-api/#push-notifications) allows clients to
+register HTTP pushers so that they can receive notifications. An HTTP pusher is always tied to a specific Matrix device
```
turt2live, line 4: It also supports email pushers 😇

turt2live (commented):
overall the MSC seems fine at a technical level - just some added context/alternatives need to be provided before this can move forward, I think.

ftr: the author believes this is ready for FCP, and with the changes mentioned in this review I would agree.

proposals/3881-remote-push-notification-toggling.md
Johennes: I believe this is covered in the "Profile tags" section as these can, among others, be used to implement per-device push rule sets.

PR #3882, opened 2022-09-05
MSC3882: Allow an existing session to sign in a new session

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/login-token-requests/proposals/3882-login-token-request.md)

Implementations of initial revision:

- Homeservers:
   - [x] Synapse: https://github.com/matrix-org/synapse/pull/13722
- Clients:
   - [x] Element Web:
       - [x] matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/pull/2687
       - [x] matrix-react-sdk: https://github.com/matrix-org/matrix-react-sdk/pull/9403
   - [x] Element iOS: used as part of the Sign In with QR capability
   - [x] Element Android: used as part of the Sign In with QR capability

Implementations of revision 1:

- Homeservers:
   - [x] Synapse: https://github.com/matrix-org/synapse/pull/15388
- Clients:
   - [x] Element Web:
       - [x] matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/pull/3228
       - [x] matrix-react-sdk: https://github.com/matrix-org/matrix-react-sdk/pull/10443
   - [ ] Element iOS:
   - [x] Element Android: https://github.com/vector-im/element-android/pull/8299
 
[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3882#issuecomment-1467406297)

### DIFF
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -116,7 +116,8 @@ The capability should use the unstable prefix:
 - `org.matrix.msc3882.get_logintoken`
 
 For reference - an earlier revision of this proposal used an unstable prefix of
-`/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised as `org.matrix.msc3882`
+`/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised 
+in the response to `GET /_matrix/client/versions` as `org.matrix.msc3882`
 set to `true`. This may be referred to as "revision zero" in existing implementations.
 
 ## Dependencies
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -79,41 +79,12 @@ by default for the new endpoint. The purpose being that consent is obtained from
 [spec](https://spec.matrix.org/v1.6/client-server-api/#rate-limiting). It may be appropriate for the homeserver admin to
 to configure a low limit ("low" relative to other enforced limits). For example, a rate of once per minute could be appropriate.
 
-If a homeserver admin deems that they have a suitable alternative to UIA in place, they can make use of
-["dummy auth"](https://spec.matrix.org/v1.6/client-server-api/#dummy-auth) and have the homeserver respond with a response similar to:
+n.b. A homeserver admin may deem that they have suitable protections in place and offer the endpoint without UIA auth as described
+in the existing spec:
 
-```http
-HTTP/1.1 401 Unauthorized
-Content-Type: application/json
-```
-
-```json
-{
-  "flows": [
-    {
-      "stages": ["m.login.dummy"]
-    }
-  ],
-  "params": {},
-  "session": "xxxxxx"
-}
-```
-
-The client can then complete UIA without requiring action from the user by sending the following JSON request body:
-
-```http
-POST /_matrix/client/v1/login/get_token HTTP/1.1
-Content-Type: application/json
-```
-
-```json
-{
-  "auth": {
-    "type": "m.login.dummy",
-    "session": "xxxxxx"
-  }
-}
-```
+> A request to an endpoint that uses User-Interactive Authentication never succeeds without auth. Homeservers may allow requests
+> that don’t require auth by offering a stage with only the m.login.dummy auth type, but they must still give a 401 response to
+> requests with no auth data.
 
 ## Unstable prefix
 
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -25,6 +25,10 @@ Add a new POST endpoint to the Client-Server API that issues a time limited `m.l
 This new endpoint should be protected by user interactive authentication (UIA). However the the homeserver admin may
 choose to disable UIA if they deem suitable alternative protections are in place.
 
+It should also be protected by rate-limiting in accordance with the existing
+[spec](https://spec.matrix.org/v1.6/client-server-api/#rate-limiting). It may be appropriate for the homeserver admin to
+to configure a low limit ("low" relative to other enforced limits). For example, a rate of once per minute could be appropriate.
+
 The values returned are:
 
 - `login_token` - required, the token to use with `m.login.token`
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -1,11 +1,13 @@
 # MSC3882: Allow an existing session to sign in a new session
 
-In [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) a proposal is made to allow a user to login on a new device using an existing device by means of scanning a
-QR code.
+In [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) a proposal is made to allow a user to login
+on a new device using an existing device by means of scanning a QR code.
 
-In order to support the above proposal a mechanism is needed where by the new device can obtain a new access token that it can use with the Client-Server API.
+In order to support the above proposal a mechanism is needed where by the new device can obtain a new access token that
+it can use with the Client-Server API.
 
-It is proposed that the current `m.login.token` mechanism is extended to allow the issuance of a login token by an existing client session.
+It is proposed that the current `m.login.token` mechanism is extended to allow the issuance of a login token by an
+existing client session.
 
 ## Proposal
 
@@ -16,11 +18,12 @@ Add a new POST endpoint to the Client-Server API that issues a time limited `m.l
 ```json
 {
     "login_token": "<login token>",
-    "expires_in": 3600
+    "expires_in": 120
 }
 ```
 
-This new endpoint MAY be protected by user interactive authentication.
+This new endpoint should be protected by user interactive authentication (UIA). However the the homeserver admin may
+choose to disable UIA if they deem suitable alternative protections are in place.
 
 The values returned are:
 
@@ -49,8 +52,8 @@ could use the device authorization grant flow which allows for a new device to b
 
 ## Security considerations
 
-A malicious client could use the mechanism to spawn more than one session. For this reason the endpoint can be placed
-behind user interactive authentication.
+A malicious client could use the mechanism to spawn more than one session. For this reason the endpoint should by default
+be placed behind user interactive authentication.
 
 ## Unstable prefix
 
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -22,6 +22,11 @@ Add a new POST endpoint to the Client-Server API that issues a time limited `m.l
 
 This new endpoint MAY be protected by user interactive authentication.
 
+The values returned are:
+
+- `login_token` - required, the token to use with `m.login.token`
+- `expires_in` - required, the expiry time for the token in seconds
+
 This token can then be used as per the existing Login spec of the Client-Server API as follows:
 
 `POST /login`
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -1,4 +1,4 @@
-# MSCxxxx: Allow an existing session to sign in a new session
+# MSC3882: Allow an existing session to sign in a new session
 
 todo
 
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -11,7 +11,7 @@ existing client session.
 
 ## Proposal
 
-Add a new POST endpoint to the Client-Server API that issues a single-use, time-limited `m.login.token` token:
+Add a new optional POST endpoint to the Client-Server API that issues a single-use, time-limited `m.login.token` token:
 
 `POST /_matrix/client/v1/login/get_token`
 
@@ -56,6 +56,24 @@ Content-Type: application/json
 }
 ```
 
+The availability of the new API endpoint should be determined via a new `m.get_logintoken`
+[capability](https://spec.matrix.org/v1.6/client-server-api/#capabilities-negotiation).
+
+This capability has a single flag, `enabled`, to denote whether the `/login/get_token` API is available or not.
+Cases for disabling might include security restrictions imposed by the homeserver admin.
+
+An example of the capability API’s response for this capability is:
+
+```json
+{
+  "capabilities": {
+    "m.get_logintoken": {
+      "enabled": true
+    }
+  }
+}
+```
+
 ## Potential issues
 
 None identified.
@@ -93,23 +111,13 @@ While this feature is in development the new endpoint should be exposed using th
 
 - `/_matrix/client/unstable/org.matrix.msc3882/login/get_token`
 
-Additionally, the feature is to be advertised as unstable feature in the `GET /_matrix/client/versions` response, with
-the key `org.matrix.msc3882.r1` set to `true`. (The `r1` in the feature name is used to indicate that the server implements the first revision of this proposal)
+The capability should use the unstable prefix:
 
-So, the response could look then as following:
+- `org.matrix.msc3882.get_logintoken`
 
-```json
-{
-    "versions": ["r0.6.0"],
-    "unstable_features": {
-        "org.matrix.msc3882.r1": true
-    }
-}
-```
-
-For reference - an earlier iteration of this proposal used an unstable prefix of
+For reference - an earlier revision of this proposal used an unstable prefix of
 `/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised as `org.matrix.msc3882`
-set to `true`.
+set to `true`. This may be referred to as "revision zero" in existing implementations.
 
 ## Dependencies
 
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -1,6 +1,11 @@
 # MSC3882: Allow an existing session to sign in a new session
 
-todo
+In MSCyyyy (to be published) a proposal is made to allow a user to login on a new device using an existing device by means of scanning a
+QR code.
+
+In order to support the above proposal a mechanism is needed where by the new device can obtain a new access token that it can use with the Client-Server API.
+
+It is proposed that the current `m.login.token` mechanism is extended to allow the issuance of a login token by an existing client session.
 
 ## Proposal
 
@@ -30,7 +35,7 @@ This token can then be used as per the existing Login spec of the Client-Server 
 
 ## Potential issues
 
-todo
+None identified.
 
 ## Alternatives
 
@@ -44,7 +49,7 @@ behind user interactive authentication.
 
 ## Unstable prefix
 
-tbd
+None.
 
 ## Dependencies
 
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -1,6 +1,6 @@
 # MSC3882: Allow an existing session to sign in a new session
 
-In MSCyyyy (to be published) a proposal is made to allow a user to login on a new device using an existing device by means of scanning a
+In [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) a proposal is made to allow a user to login on a new device using an existing device by means of scanning a
 QR code.
 
 In order to support the above proposal a mechanism is needed where by the new device can obtain a new access token that it can use with the Client-Server API.
--- a/proposals/3882-login-token-request.md
+++ b/proposals/3882-login-token-request.md
@@ -49,7 +49,22 @@ behind user interactive authentication.
 
 ## Unstable prefix
 
-None.
+While this feature is in development the new endpoint should be exposed using the following unstable prefix:
+
+- `/_matrix/client/unstable/org.matrix.msc3882/login/token`
+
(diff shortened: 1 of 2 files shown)


proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+## Potential issues
+
+todo
+
+## Alternatives
+
+If Matrix was already using OIDC as per [MSC3861](https://github.com/matrix-org/matrix-spec-proposals/pull/3861) then we could use the device authorization grant flow which allows for a new device to be signed in using an existing device.
```
uhoreg, line 37: Please wrap your lines to ~80 or ~120 characters

turt2live:
@matrix-org/security - authentication scares me: would you be able to take a look at this for technical design and security considerations please?

@matrix-org/spec-core-team - the author is looking for general technical design review on this one for the moment. It looks like there's also implementations provided: might be able to remove `needs-implementation`?

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,76 @@
...
+## Proposal
+
+Add a new POST endpoint to the Client-Server API that issues a time limited `m.login.token` token:
+
+`POST /login/token`
+
+```json
+{
+    "login_token": "<login token>",
+    "expires_in": 3600
```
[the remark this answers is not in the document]
davidegirardi: The login needs to be verified by checking some digits right?

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,76 @@
...
+}
+```
+
+This new endpoint MAY be protected by user interactive authentication.
```
davidegirardi, line 23: Should it be default?

proposals/3882-login-token-request.md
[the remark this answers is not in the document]
hughns: Yes, I would expect the token to be valid for a shorter time period and I will incorporate this.

The login token can be used to sign in without any further checks. Any protections (such as UIA) should happen before this point.

proposals/3882-login-token-request.md
hughns: Having had some other discussions in recent weeks on this I concur with you that it should be the default. I will update to reflect this.

proposals/3882-login-token-request.md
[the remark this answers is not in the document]
davidegirardi: OK, since we [check](https://github.com/matrix-org/synapse/blob/develop/synapse/storage/databases/main/registration.py#L1875-L1876) for token reuse we have a good posture against shoulder-scanning where someone tries to sneak behind you and login.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+This new endpoint should be protected by user interactive authentication (UIA). However the the homeserver admin may
+choose to disable UIA if they deem suitable alternative protections are in place.
+
+The values returned are:
+
+- `login_token` - required, the token to use with `m.login.token`
+- `expires_in` - required, the expiry time for the token in seconds
```
turt2live, line 31:
```suggestion
- `expires_in` - required, how long until the token expires in seconds
```

to differentiate from a timestamp that uses seconds instead of milliseconds

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,79 @@
...
+
+This token can then be used as per the existing Login spec of the Client-Server API as follows:
```
turt2live, line 33:
```suggestion
This token can then be used as per the existing [Login spec](https://spec.matrix.org/v1.6/client-server-api/#login) as follows:
```

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,79 @@
...
+
+```json
+{
+    "login_token": "<login token>",
+    "expires_in": 120
+}
+```
+
+This new endpoint should be protected by user interactive authentication (UIA). However the the homeserver admin may
+choose to disable UIA if they deem suitable alternative protections are in place.
```
turt2live, line 26: In addition to regular authentication? Should the endpoint be rate limited?

turt2live (approved):
overall seems fine enough to move forward: just a few things to touch up concurrent to FCP.

@mscbot fcp merge

turt2live:
hmm, alright then.

@mscbot fcp merge

proposals/3882-login-token-request.md
hughns: I had imagined that it would fall under the default Client Server API [rate-limiting](https://spec.matrix.org/v1.6/client-server-api/#rate-limiting) but I will call this out explicitly.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,83 @@
...
```
richvdh, line 26: as per msc3967: define "disable UIA" please.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,83 @@
...
+it can use with the Client-Server API.
+
+It is proposed that the current `m.login.token` mechanism is extended to allow the issuance of a login token by an
+existing client session.
+
+## Proposal
+
+Add a new POST endpoint to the Client-Server API that issues a time limited `m.login.token` token:
+
+`POST /login/token`
```
richvdh, line 16:
```suggestion
`POST /_matrix/client/v1/login/token`
```

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,83 @@
...
+
+```json
+{
+    "login_token": "<login token>",
+    "expires_in": 120
+}
+```
```
richvdh, line 23: is this the response or the request? As currently set out, it looks like the request, but on re-reading I think it might be the response?

proposals/3882-login-token-request.md
richvdh: Maybe a bit of a bikeshed issue, but I'm unconvinced this is the clearest name. 

`.../login/token` gives me the impression that we are logging in using a token (ie, the functionality currently offered by `/login` with `type: m.login.token`). Would it be better to call it `.../login/get_token`?

hughns: I've made it `POST /_matrix/client/v1/login/get_token` in the latest iteration.

Should it definitely be on `/_matrix/client/v1/login/get_token` not `/_matrix/client/v3/login/get_token` which would be consistent with the other existing `/_matrix/client/v3/login*` endpoints?

proposals/3882-login-token-request.md
hughns: I have hopefully sufficiently clarified this in the latest revision?

proposals/3882-login-token-request.md
hughns: I've clarified this in the latest revision.

proposals/3882-login-token-request.md
richvdh: > Should it definitely be on `/_matrix/client/v1/login/get_token` not `/_matrix/client/v3/login/get_token` which would be consistent with the other existing `/_matrix/client/v3/login*` endpoints?

I think so. I'm sure the rest of the SCT will tell me I'm wrong during calls for FCP if not!

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,144 @@
...
+It is proposed that the current `m.login.token` mechanism is extended to allow the issuance of a login token by an
+existing client session.
+
+## Proposal
+
+Add a new POST endpoint to the Client-Server API that issues a single-use, time-limited `m.login.token` token:
+
+`POST /_matrix/client/v1/login/get_token`
+
+The client should send an empty JSON body for the `POST` request.
```
richvdh, line 18:
```suggestion
The client should send an empty JSON object for the body of the `POST` request (apart from
the `auth` property used in user-interactive authentication).
```

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Unstable prefix
+
+While this feature is in development the new endpoint should be exposed using the following unstable prefix:
+
+- `/_matrix/client/unstable/org.matrix.msc3882/login/token`
+
+Additionally, the feature is to be advertised as unstable feature in the `GET /_matrix/client/versions`
+response, with the key `org.matrix.msc3882` set to `true`. So, the response could look then as
+following:
```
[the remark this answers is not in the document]
uhoreg: FWIW, Rich's concern here is different from the issue that I brought up.  His concern is how to detect the presence of the feature between the acceptance of this MSC and the spec version being fully implemented.  My issue is about whether or not admins/servers can selectively disable the feature for users.  I'll start a new thread for my concern.

proposals/3882-login-token-request.md
uhoreg: That's my understanding.  Each endpoint has its own version number.  All the r0 endpoints got migrated to v3 due to reasons, but new endpoints get to start from v1.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+## Unstable prefix
+
+While this feature is in development the new endpoint should be exposed using the following unstable prefix:
+
+- `/_matrix/client/unstable/org.matrix.msc3882/login/get_token`
+
+Additionally, the feature is to be advertised as unstable feature in the `GET /_matrix/client/versions` response, with
+the key `org.matrix.msc3882.r1` set to `true`. (The `r1` in the feature name is used to indicate that the server implements the first revision of this proposal)
```
uhoreg, line 126: My understanding is that this is needed for [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906/files#diff-89279ef5721f17a005de8529325f6b7ef2aecec72dea442572f844f769a54304R99) to check whether it can go ahead with the login.  However, I'm not sure if this (or a spec version check after this gets released in the spec) is sufficient, as a server/admin may want to disable the feature for certain users, or disable the feature entirely.  So even though the server may implement the feature, it may not be available to the users.  We could say that servers are not allowed to disable the feature, but that sounds overly restrictive. Or we could say that MSC3906 will assume that if the server implements the feature, then the user can use it and the login will just fail if it can't, but I'd prefer to avoid having misleading UI if possible.

My first thought was to add a field to `GET /login` to indicate the availability of the feature, but we want the request to be authenticated because the feature might be enabled/disabled on a per-user basis. Maybe the best place to put it is in [Capabilities negotiation](https://spec.matrix.org/unstable/client-server-api/#capabilities-negotiation).

proposals/3882-login-token-request.md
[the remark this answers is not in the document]
richvdh: For more context on my concern: jump-to-date is currently suffering from the exact failure mode I outlined: https://github.com/matrix-org/synapse/pull/15283.

Though also, [a discussion thread on that PR](https://github.com/matrix-org/synapse/pull/15283#discussion_r1142071333) appears to be heading in the direction of: "let's avoid using stable feature flags".

So I'm not quite sure what to do here. Probably best to leave it as-is for now at least.

proposals/3882-login-token-request.md
richvdh: yes, this is the sort of thing that `GET /capabilities` is intended to support.

hughns: Okay, I've re-worked the proposal to make it an optional capability that is discoverable via `GET /capabilities`.

@uhoreg @richvdh I wasn't sure what to call it exactly - please can you let me know if the proposed `m.get_logintoken` is suitable before I create PRs for the updated implementations?

proposals/3882-login-token-request.md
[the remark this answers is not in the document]
hughns: @richvdh I think the revised proposal of using a capability addresses your concerns about rollout. The sequence could be:

1. Pre-FCP completed (i.e. now): clients check for the `org.matrix.msc3882.get_logintoken` capability (and optionally the unstable `org.matrix.msc3882` feature from the revision 0 of this MSC)
2. Once FCP is complete:
  - server support for feature can move to stable `m.get_logintoken` prefix
  - clients can check for `m.get_logintoken` capability and fall back to pre-FCP checks from (1)
3. Once the feature is in a release of the spec (say `1.8`):
  - if a client is talking to a server that supports `1.8` then it knows it only needs to check the stable `m.get_logintoken` capability
  - if a client is talking to an earlier server then it could follow the (2) FCP complete logic
4. At a future point that clients think that `1.8` is supported widely enough they can stop implementing (2) and only use the stable `m.get_logintoken` capability

proposals/3882-login-token-request.md
```diff
@@ -93,23 +111,13 @@ While this feature is in development the new endpoint should be exposed using th
...
-    "unstable_features": {
-        "org.matrix.msc3882.r1": true
-    }
-}
-```
+- `org.matrix.msc3882.get_logintoken`
 
-For reference - an earlier iteration of this proposal used an unstable prefix of
+For reference - an earlier revision of this proposal used an unstable prefix of
 `/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised as `org.matrix.msc3882`
```
richvdh, line 119:
```suggestion
`/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised 
in the response to `GET /_matrix/client/versions` as `org.matrix.msc3882`
```

proposals/3882-login-token-request.md
richvdh: `m.get_logintoken` sgtm. I think we need to mention the `org.matrix.msc3882.get_logintoken` unstable capability too, in the "unstable prefix" section.

hughns: However, whilst re-working the iOS and Android implementation I've realised that a signed-out client needs a way to know whether a Homeserver *might* support the capability.

This is because in the current UX if a client is not yet logged in it will only show the "Sign in with QR code" button if the server indicates support.

So, I need an unauthenticated way for a client to check the capability. Should this be somewhere in `/versions` or in the `/.well-known/matrix/client`? or somewhere else?

richvdh: gah, I think it has to be in `/versions`.

hughns: @uhoreg I had forgotten that both check them, not just the logged in client.

hughns: I'm not sure what is the best option:

1. `GET /versions`

The doesn't appear to be an existing way to slot this in. So perhaps add a top-level `features` field something like:

```json
{
  "features": {
    "m.login.get_logintoken": true,
  },
  "unstable_features": {
    "org.example.my_feature": true
  },
  "versions": [
    "r0.0.1",
    "v1.1"
  ]
}
```

2. `GET /login`

Define a "special" (similar to the existing `m.auth.dummy`) auth type that is returned as a possible login flow:

```json
{
  "flows": [
    {
      "type": "m.login.get_logintoken"
    }
  ]
}
```

Option (1) feels slightly cleaner to me albeit adding a new section to the versions response. Which way do you think I should go?

uhoreg: I don't really have a strong preference either way.  The first option has the advantage that it can be used by other features in the future.  Though I'm not sure how useful it would actually be.

Just to throw another option into the mix, we can also modify `GET /login` in a different way:  since we already know that we'll have an `m.login.token` item in the `flows`, we can add an extra field to that part, something like

```json
{
  "flow": [
    {
      "type": "m.login.token",
      "get_login_token": true
    }
  ]
}
```

But again, I don't have a strong preference either way.

richvdh: To be clear, my understanding is that the feature may be:
 1. unavailable to *any* user on the homeserver, or
 2. available for some users but not others.

An *authenticated* client can ask "is this feature available for my user" via `/capabilities`, but we also have a need for an *unauthenticated* client to ask "is this feature available for any user", and we are discussing how best to expose that.

This is all setting aside any "unstable" feature flags, which add complexity and we can figure out afterwards.

Is that correct?

---

> Option (1) feels slightly cleaner to me albeit adding a new section to the versions response. Which way do you think I should go?

I am reluctant that we extend `/versions` in this way - `/versions` is really about "how recent is my homeserver implementation". For me, it feels a much better fit for `GET /login`.

@uhoreg:

> ```json
> {
>   "flow": [
>     {
>       "type": "m.login.token",
>       "get_login_token": true
>     }
>   ]
> }
> ```

Yes, I like this suggestion.

uhoreg: > To be clear, my understanding is that the feature may be:
> 
>     1. unavailable to _any_ user on the homeserver, or
> 
>     2. available for some users but not others.

In theory, yes, though I don't know if # 2 will be a common-enough situation.  My initial thought was that if only the already-logged-in user was checking the capability, we might as well have the flexibility of allowing it to be selectively enabled.  But given that we have to expose the feature request to unauthenticated users, it would mean putting the feature query on two different endpoints, which may be more trouble than it's worth.  So unless someone else feels strongly about # 2, I'm fine, personally, to just have the unauthenticated query.

hughns: >To be clear, my understanding is that the feature may be:
>
> - unavailable to any user on the homeserver, or
> - available for some users but not others.
>
> An authenticated client can ask "is this feature available for my user" via /capabilities, but we also have a need for an unauthenticated client to ask "is this feature available for any user", and we are discussing how best to expose that.
>
> This is all setting aside any "unstable" feature flags, which add complexity and we can figure out afterwards.
>
> Is that correct?

@richvdh Yes, this is correct.

I have updated the proposal to use `GET /login` for the unauthenticated client and `GET /capabilities` for the definitive answer for the authenticated client.

It comes out that no feature flag is needed. However, there are unstable versions of the new fields.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,177 @@
...
+In this case the mechanism could be available and so the client could prompt the user to try using it.
+
+#### Scenario 2: The user wishes to use an already signed in Matrix client to sign in another client
+
+The client is already authenticated. The client can determine whether it is able and allowed to sign in
+another client by checking the
+[capabilities](https://spec.matrix.org/v1.6/client-server-api/#capabilities-negotiation)
+advertised by the homeserver.
+
+Where the client is authenticated the client can determine whether the new API endpoint is available
```
anoadragon453, line 109:
```suggestion
The unauthenticated client can also determine whether the new API endpoint is available
```

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,177 @@
...
+
+```json
+{
+  "capabilities": {
+    "m.get_login_token": {
+      "enabled": true
+    }
+  }
+}
+```
```
anoadragon453, line 129: MSC structure nit: I would define the `/capabilties` response addition here first, then the use cases afterwards. Otherwise I was reading the beginning of this section wondering what capability the clients would use.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,177 @@
...
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+```
+
+```json
+{
+    "login_token": "<login token>",
+    "expires_in_ms": 120000
```
anoadragon453, line 43: Is there any reason to make this `expires_in_ms` vs. `expires_ts`, i.e. handing the client the ms timestamp at which the token expires? I just wonder whether network latency could cause the client to compute a timestamp for when the token expires that's different enough than what the server has.

proposals/3882-login-token-request.md
```diff
@@ -0,0 +1,177 @@
...
+## Unstable prefix
+
+While this feature is in development the following unstable prefixes should be used:
+
+- API endpoint `/_matrix/client/v1/login/get_token` => `/_matrix/client/unstable/org.matrix.msc3882/login/get_token`
+- capability `m.get_login_token` => `org.matrix.msc3882.get_login_token`
+- login flow field `get_login_token` => `org.matrix.msc3882.get_login_token`
+
+For reference - an earlier revision of this proposal used an unstable prefix of
+`/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised
```
anoadragon453, line 171:
```suggestion
For reference - an earlier revision of this proposal used an unstable endpoint of
`/_matrix/client/unstable/org.matrix.msc3882/login/token` with an unstable feature advertised 
```

anoadragon453 (commented):
lgtm, other than the timestamp question.

proposals/3882-login-token-request.md
hughns: The reason for `expires_in_ms` being proposed is that it is consistent with the other parts of authentication in the client-server API.

anoadragon453: That's a good point. Plus while network latency could be a factor, it's less likely than the client and server having slightly different clock values - which a relative value such as `expires_in_ms` does defend against.

anoadragon453 (approved):

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1530

turt2live:
Merged 🎉

Issue #3298, opened 2021-07-27
device_one_time_keys_count: difficulty in differentiating between "keys exhausted" and "no change since last sync" states


anoadragon453:
As found in https://github.com/matrix-org/synapse/issues/10484, the spec is currently vague with regards to when the "device_one_time_keys_count" field on both [`GET /_matrix/client/r0/sync`](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-sync) and [`POST /_matrix/client/r0/keys/upload`](https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-keys-upload) responses.

The field is currently marked as Optional, yet it's not clear when a homeserver should include the field in a response, and when it does what should be put in it. The confusion lies in how to differentiate between telling the client that "there's been no change in the key count since last sync" and "there are no OTKs left".

For a long time Synapse has simply omitted the contents of the field (`"device_one_time_keys_count": {}`) in the case of OTKs being exhausted. However this has lead to confusion over some clients interpreting this to mean that there has been no change. https://github.com/matrix-org/synapse/pull/10485 fixes the issue by always including the key algorithm name; thus an exhausted pool would look like `"device_one_time_keys_count": {"signed_curve25519": 0}`.

You then might say that `"device_one_time_keys_count": {"algorithm_name": 0}` means an exhausted pool, and the lack of an algorithm name in the `device_one_time_keys_count` dictionary would mean that no change has occurred. But what if a new algorithm is introduced down the line? Then clients would interpret the lack of the new algorithm name as no change has occurred. Perhaps this can be rectified by tightly coupling possible algorithm names with supported spec versions advertised by the homeserver though.

Either way, the spec as it stands is currently vague. It would be beneficial to initially at least agree on something more explicit.

uhoreg:
The description for that field says:

> For each key algorithm, the number of unclaimed one-time keys of that type currently held on the server for this device.

To me, that means that it's an absolute count, and not meant to indicate a change, thus a missing algorithm would imply that no keys exist for that algorithm, rather than no change.

richvdh:
That might be a reasonable interpretation, but it does imply that every single sync response contains `{"device_one_time_keys_count": {"signed_curve25519": N},`, which feels pretty redundant when nothing is changing.

In any case, there's certainly been some confusion around it, so it would be good to clarify this.

PR #3884, opened 2022-09-06
MSC3884: Sliding Sync Extension: E2EE

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/msc3884/proposals/3884-sliding-sync-e2ee.md)

[Client Implementation](https://github.com/matrix-org/matrix-js-sdk/blob/v19.5.0-rc.4/src/sliding-sync-sdk.ts#L44)
[Server Implementation](https://github.com/matrix-org/sliding-sync/blob/main/sync3/extensions/e2ee.go)

### DIFF
--- a/proposals/3884-sliding-sync-e2ee.md
+++ b/proposals/3884-sliding-sync-e2ee.md
@@ -51,7 +51,7 @@ Changing this behaviour may break clients which expect these fields to always ex
 ## Alternatives
 
 The alternative is to not include this extension at all, making it impossible to include reliable
-E2EE support in Sliding Sync. As this extension is opt-in, as all Sliding Sync extensions are, this
+E2EE support in Sliding Sync. As this extension is opt-in, as all Sliding Sync extensions are, it
 feels undesirable to not provide this extension.
 
 ## Security considerations
--- a/proposals/3884-sliding-sync-e2ee.md
+++ b/proposals/3884-sliding-sync-e2ee.md
@@ -64,7 +64,7 @@ No unstable prefix as Sliding Sync is still in review. To enable this extension,
 your request JSON:
 ```js
 {
-    "extensions: {
+    "extensions": {
         "e2ee": {
             "enabled": true
         }
--- a/proposals/3884-sliding-sync-e2ee.md
+++ b/proposals/3884-sliding-sync-e2ee.md
@@ -31,6 +31,9 @@ the `e2ee` extension response:
 }
 ```
 
+All keys are optional and clients MUST check if they exist prior to use. If there are zero changes to
+every field then the server MAY omit sending back an `e2ee` extension entirely.
+
 The semantics of these fields is exactly the same as the current `/sync` implementation, as implemented
 in v1.3 of the Client-Server Specification. `device_lists` may be omitted if there are no users who
 have changed/left.
--- a/proposals/3884-sliding-sync-e2ee.md
+++ b/proposals/3884-sliding-sync-e2ee.md
@@ -38,12 +38,12 @@ have changed/left.
 For sliding sync, an initial response will include all fields. When there are updates, only the
 _changed_ fields are returned. That is to say, if `device_one_time_keys_count` has not changed between
 requests, it will be omitted which means to use the previous value. This deviates from the current
-`/sync` implementation which alwyas includes this field. Likewise for `device_unused_fallback_key_types`.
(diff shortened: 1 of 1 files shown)


proposals/3884-sliding-sync-e2ee.md
```diff
@@ -0,0 +1,80 @@
...
+{
+    "device_one_time_keys_count": {
+        "signed_curve25519": 50
+    },
+    "device_lists": {
+        "changed": ["@alice:example.com"],
+        "left": ["@bob:example.com"]
+    },
+    "device_unused_fallback_key_types": [
+        "signed_curve25519"
```
deepbluev7, line 29: This field has to always be included, otherwise there is no way for the client to tell, if the fallback key got used.

HOWEVER:

You can actually probably solve that with the sliding sync API, by having the client explicitly tell the server about what fallback keys it wants to know about, when they are **used**. Then the server would have to include them everytime when they are used, but could not include them in the common case of them being unused. The same could be done for the one time keys count, where you specify a minimum number you want to have and if you have those, the server doesn't update you about it.

This is however a bit racy, as a client might be told about a used key, while it is still uploading a key. But I think that is already an issue we have today.

Basically the request parameters would then look something like this:

```json5
{
  "enabled": true,
  "minimum_device_one_time_keys_count": {
    "signed_curve25519": 50 // tell the server to notify the client, if the OTKs fall below this limit
  },
  "fallback_key_types": [ "signed_curve25519" ] // tell the server to notify the client, if no such key is present
}
```

proposals/3884-sliding-sync-e2ee.md
```diff
@@ -0,0 +1,80 @@
...
+```
+If `enabled` is `true`, then the sliding sync response MAY include the following response fields in
+the `e2ee` extension response:
+```json
+{
+    "device_one_time_keys_count": {
+        "signed_curve25519": 50
+    },
+    "device_lists": {
+        "changed": ["@alice:example.com"],
```
deepbluev7, line 25: What are the semantics for this? When do I get a "changed" event? Only for device lists that changed during this connection? Or does this work on the delta token? Maybe this should have its own since and next batch token? Key data is actually a LOT of data, so invalidating that on every reconnect would be pretty scary. And not being told your device list is possibly missing changed events, because the delta token expired is equally scary.

proposals/3884-sliding-sync-e2ee.md
deepbluev7: This probably requires the server to tell the client, if a key type is unsupported though.

deepbluev7: Related spec issue: https://github.com/matrix-org/matrix-spec-proposals/issues/3298

Unsupported keys are hard to distinguish from "no change".

proposals/3884-sliding-sync-e2ee.md
kegsay: The semantics are the same as sync v2 regarding getting a "changed" or "left" event.

Currently, the proxy will aggregate changed/left responses from sync v2 then deliver this to the client, only purging the arrays server-side once the client has acknowledged the response. This does mean that if you're offline for a while, these arrays can be large. We don't currently have delta token support in this extension, but it might be worthwhile to add it if we can get the semantics right.

Also worth noting that in practice the size of these arrays isn't a large part of the overall sync response, even if you have been offline for a while.

proposals/3884-sliding-sync-e2ee.md
kegsay: I like this idea overall, especially with the "min" values. I'd be interested to hear if this makes things any easier on clients, or just saves a bit of bandwidth.

proposals/3884-sliding-sync-e2ee.md
deepbluev7: My concern is mostly, that it would be easy to get this wrong on the client side, if you don't treat all devices as changed after a reconnection, where the server doesn't send you change lists from before you reconnected. Currently clients mostly rely on getting all change events as long as they don't do a new initial sync, but with sliding sync the "initial sync" could possibly happen quite often and by accident and in that case you would need to treat ALL keys as stale. You can probably implement that correctly in a client, but some more explicit notification, that your delta token wasn't used and your whole lists should be treated as stale might be harder to get wrong.

erikjohnston:
Superseded by https://github.com/matrix-org/matrix-spec-proposals/pull/4543

PR #3885, opened 2022-09-07
MSC3885: Sliding Sync Extension: To-Device messages

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/msc3885/proposals/3885-sliding-sync-to-device.md)

[Client Implementation](https://github.com/matrix-org/matrix-js-sdk/blob/v19.5.0-rc.4/src/sliding-sync-sdk.ts#L92)
[Server Implementation](https://github.com/matrix-org/sliding-sync/blob/main/sync3/extensions/todevice.go)

### DIFF
--- a/proposals/3885-sliding-sync-to-device.md
+++ b/proposals/3885-sliding-sync-to-device.md
@@ -34,6 +34,9 @@ the `to_device` extension response:
 }
 ```
 
+The client MUST persist the `next_batch` value to persistent storage between requests in case the client is
+killed by the OS.
+
 The semantics of the `events` field is exactly the same as the current `/sync` implementation, as implemented
 in v1.3 of the Client-Server Specification. The server MUST NOT send more than `limit` events: it may send
 less than `limit` events if the value provided was too large.
--- a/proposals/3885-sliding-sync-to-device.md
+++ b/proposals/3885-sliding-sync-to-device.md
@@ -1,0 +1,81 @@
+# MSC3885: Sliding Sync Extension: To-Device
+
+This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575)
+which includes support for to-device messages in the `/sync` response.
+
+## Proposal
+
+MSC3575 does not include support for to-device events. This extension will add support for
+to-device events. There is a **critical** problem in the current to-device implementation, which is
+that events are implicitly acknowledged when the user advances the `/sync` token. This causes problems
+when clients need to have 2 or more sync streams open at a time, e.g a push notification process _and_
+a main process. This can cause the two processes to race to fetch the to-device events, resulting in
+the need for complex synchronisation rules to ensure the token is correctly and atomically exchanged
+between processes. Sliding Sync's implementation of to-device events removes this requirement by
+associating an _explicit_ token **just for to-device events**. This token uses the same terminology
+of the current `/sync` implementation: `since` and `next_batch`.
+
+The proposal is to introduce a new extension called `to_device` with the following request parameters:
+```js
+{
+    "enabled": true, // sticky
+    "limit": 100, // sticky, max number of events to return, server can override
+    "since": "some_token" // optional, can be omitted on initial sync / when extension is enabled
+}
+```
+If `enabled` is `true`, then the sliding sync response includes the following response fields in
+the `to_device` extension response:
+```json
+{
+    "next_batch": "some_token", // REQUIRED: even if no changes
+    "events": [ // optional, only if there are events since the last `since` value.
+        { ... to device event ... }
+    ]
+}
+```
+
+The semantics of the `events` field is exactly the same as the current `/sync` implementation, as implemented
+in v1.3 of the Client-Server Specification. The server MUST NOT send more than `limit` events: it may send
+less than `limit` events if the value provided was too large.
+
+The `events` are treated as "acknowledged" when the server receives a new request with the `since`
+value set to the previous response's `next_batch` value. When this occurs, acknowledged events are
+permanently deleted from the server, and MUST NOT be returned to the client should another request
+with an older `since` value be sent.
+
+If a response is lost, the client will retry the request with the same `since` value. When this happens,
+the server MAY bundle together more events if they arrived in the interim. 
+
+## Potential issues
+
+This proposal introduces a new sync token type for clients to manage. It is done for good reasons but
+it does slightly increase the complexity of designing a correct client implementation. 
+
+## Alternatives
+
+The alternative is to include to-device events like normal events in a different section, without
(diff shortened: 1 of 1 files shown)


proposals/3885-sliding-sync-to-device.md
```diff
@@ -0,0 +1,81 @@
...
+
+This proposal introduces a new sync token type for clients to manage. It is done for good reasons but
+it does slightly increase the complexity of designing a correct client implementation. 
+
+## Alternatives
+
+The alternative is to include to-device events like normal events in a different section, without
+making use of dedicated `since` and `next_batch` tokens, instead relying on the `pos` value. This
+would revert to-device events to be implicitly acknowledged, which has caused numerous [issues](https://github.com/vector-im/element-ios/issues/3817) in
+the past.
```
poljar, line 59: I'm not sure I understand how this setup differs, in a meaningful way, with regards to the linked issue. Not that the issue defines the problem completely. Matter of fact the following definition sounds exactly the same as for sync v2:

> The `events` are treated as "acknowledged" when the server receives a new request with the `since` value set to the previous response's `next_batch` value. When this occurs, acknowledged events are permanently deleted from the server, and MUST NOT be returned to the client should another request with an older `since` value be sent.

We're still going to have two processes that will try to fetch the to-device events. One of them might advance further then the other, resulting in the events being acknowledged and thus removed from the server.

If we have a bunch of tokens denoted by `Tₙ`. Where `n` is used to indicate the order in our sync token sequence. Once we use the token `Tₙ₊₁` for a sync request, to-device events, that were part of the previous `Tₙ` sync response, will be deleted from the server.

This means if process `A` uses token `Tₙ₊₁` before process `B` manages to use token `Tₙ`, then process `B` will miss all events that were part of the sync response for which token `Tₙ` was used by process `A`. This leads to a discrepancy in state of cryptographic objects of process `A` and process `B`. In the linked issue process `B` overwrites the newer state inserted by process `A` with an older and incomplete version of it.

Doesn't this scenario happen with this MSC as well?

kegsay: Yes of course, but that has never been the point it has been trying to solve.

This MSC solves the problem by allowing a process to choose to opt in to to-device messages, and explicitly decide when to acknowledge said messages whilst still getting other unrelated events. Processes will need to agree which one will be in charge of processing them. You cannot safely have multiple sync streams at all in Sliding Sync.

I consulted the iOS folks when designing this MSC and this was all they needed from their end when I asked.

proposals/3885-sliding-sync-to-device.md
```diff
@@ -0,0 +1,84 @@
...
+{
+    "next_batch": "some_token", // REQUIRED: even if no changes
+    "events": [ // optional, only if there are events since the last `since` value.
+        { ... to device event ... }
+    ]
+}
+```
+
+The client MUST persist the `next_batch` value to persistent storage between requests in case the client is
+killed by the OS.
```
deepbluev7, line 38: Why MUST? In the worst case the server will just send the last 100 or so to_device messages again, wouldn't it?

kegsay: Server-side, yes this is the worst case.

Cilent-side it's more pernicious, as the client may have processed some but not all the to-device events before being killed, resulting in processing duplicate to-device events, the effects of which will depend on the kind of event.

deepbluev7: Sounds more like it is the clients fault then, if it gets that wrong. I was a bit surprised to see that called out as explicitly, but I guess that is fine.

proposals/3885-sliding-sync-to-device.md
```diff
@@ -0,0 +1,84 @@
...
+between processes. Sliding Sync's implementation of to-device events removes this requirement by
+associating an _explicit_ token **just for to-device events**. This token uses the same terminology
+of the current `/sync` implementation: `since` and `next_batch`.
+
+The proposal is to introduce a new extension called `to_device` with the following request parameters:
+```js
+{
+    "enabled": true, // sticky
+    "limit": 100, // sticky, max number of events to return, server can override
+    "since": "some_token" // optional, can be omitted on initial sync / when extension is enabled
```
bnjbvr, line 23: Uh?

```suggestion
    "since": "some_token" // optional, can be omitted on initial sync / when extension is disabled
```

MadLittleMods, line 23: We could get away with not having a separate token.

For example, in Synapse, the `pos` sync token is a [vector clock and already has the to-device stream position](https://github.com/element-hq/synapse/blob/e8ee784c75d161d1634a60b7c4e419e8f08273b8/synapse/types/__init__.py#L909-L925) in it. Since the Sliding Sync extension interface allows selectively enabling an extension, you can choose to acknowledge to-device messages by whether you enable the `to_device` extension (`"enabled": true`).

reivilibre: probably means when the extension is ***just*** enabled

reivilibre: We have since discussed some times in Synapse that stopping the vector clock partially from advancing is one of the (anti-)patterns that made oldschool sync brittle and hard to reason about, which may explain why the extra token was added to this MSC.

proposals/3885-sliding-sync-to-device.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+If `enabled` is `true`, then the sliding sync response includes the following response fields in
+the `to_device` extension response:
+```js
+{
+    "next_batch": "some_token", // REQUIRED: even if no changes
```
reivilibre, line 30: concern: this being REQUIRED even with no changes seems like a mistake.
The client can surely be trusted to remember what it sent in `since` and we could have it just preserve that.

By requiring this field, we also have to always serialise the response extension, even when nothing happens. This is maybe fine in isolation, but it sets a bloaty pattern as other extensions will want to be consistent with this pattern.

proposals/3885-sliding-sync-to-device.md
```diff
@@ -0,0 +1,84 @@
...
+        "to_device": {
+            "enabled": true
+        }
+    }
+}
+```
+
+## Dependencies
+
+This MSC builds on MSC3575, which at the time of writing has not yet been accepted into the spec.
```
reivilibre, line 84: Despite being in production, this MSC still depends on an obsolete MSC. This MSC ought to be updated to meet MSC4186.

Indeed, some of the 'sticky' aspects don't appear to make sense anymore and don't match the production implementation in Synapse.

erikjohnston:
Superseded by https://github.com/matrix-org/matrix-spec-proposals/pull/4538

PR #3886, opened 2022-09-07
MSC3886: Simple client rendezvous capability

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/simple-rendezvous-capability/proposals/3886-simple-rendezvous-capability.md)

- Servers:
   - [x] Standalone Rust server: https://github.com/matrix-org/rust-http-rendezvous-server
   - [x] Synapse homeserver module: https://pypi.org/project/matrix-http-rendezvous-synapse/
   - [x] Standalone prototype Node.js server: https://github.com/matrix-org/node-http-rendezvous-server
   - [x] Synapse experimental feature to delegate rendezvous to external server via an HTTP `307` redirect: https://github.com/matrix-org/synapse/pull/14018
- Clients:
   - [x] Element Web:
       - [x] matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/pull/2747
       - [x] matrix-react-sdk: https://github.com/matrix-org/matrix-react-sdk/pull/9403
   - [x] Element iOS:
       - [x] https://github.com/vector-im/element-ios/pull/6806

### DIFF
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -1,4 +1,4 @@
-# MSC3886: Simple rendezvous capability
+# MSC3886: Simple client rendezvous capability
 
 In MSCyyyy a proposal is made to allow a user to login on a new device using an existing device by means of scanning a
 QR code.
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -10,6 +10,8 @@ information such as:
 - the user ID
 - facilitation of issuing a new access token
 - device ID for end-to-end encryption
+- data for establishing a secure communication channel (e.g. via
+  [MSC3903](https://github.com/matrix-org/matrix-spec-proposals/pull/3903))
 
 To enable [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) and support any future proposals this MSC proposes a simple HTTP based protocol that can be used to
 establish a direct communication channel between two IP connected devices.
@@ -296,7 +298,9 @@ following:
 
 ## Dependencies
 
-None, although it's intended to be used with [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906).
+None, although it's intended to be used with
+[MSC3903](https://github.com/matrix-org/matrix-spec-proposals/pull/3903) and
+[MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906).
 
 ## Credits
 
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -33,21 +33,21 @@ sequenceDiagram
   Note over A: Device A determines which rendezvous server to use
 
   A->R: POST / Hello from A
-  R->A: 201 Created Location: <rendezvous ID>
+  R->A: 201 Created Location: /abc-def-123-456
 
-  A-->B: Rendezvous URI between clients, perhaps as QR code: https://rendzvous-server/<rendezvous ID>
+  A-->B: Rendezvous URI between clients, perhaps as QR code: e.g. https://rendzvous-server/abc-def-123-456
 
   Note over A: Device A starts polling for contact at the rendezvous
 
-  B->R: GET /<rendezvous ID>
+  B->R: GET <rendezvous URI>
   R->B: 200 OK Hello from A
 
   loop Device A polls for rendezvous updates
-    A->R: GET /<rendezvous ID> If-None-Match: <ETag>
+    A->R: GET <rendezvous URI> If-None-Match: <ETag>
     R->A: 304 Not Modified
   end
 
-  B->R: PUT /<rendezvous ID> Hello from B
+  B->R: PUT <rendezvous URI> Hello from B
   R->B: 202 Accepted 
 
   Note over A,B: Rendezvous now established
@@ -76,7 +76,7 @@ HTTP response codes:
 
 HTTP response headers for `201 Created`:
 
-- `Location` - required, the allocated rendezvous ID represented as a URI safe **relative path**. e.g. `Location: abc-def-1234`
+- `Location` - required, the allocated rendezvous URI which can be on a different server
 - `X-Max-Bytes` - required, the maximum allowed bytes for the payload
 - `ETag` - required, ETag for the current payload at the rendezvous point as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.etag)
 - `Expires` - required, the expiry time of the rendezvous as per [RFC7233](https://httpwg.org/specs/rfc7234.html#header.expires)
@@ -85,14 +85,14 @@ HTTP response headers for `201 Created`:
 Example response headers:
 
 ```http
-Location: abcdEFG12345
+Location: /abcdEFG12345
 X-Max-Bytes: 10240
 ETag: VmbxF13QDusTgOCt8aoa0d2PQcnBOXeIxEqhw5aQ03o=
 Expires: Wed, 07 Sep 2022 14:28:51 GMT
 Last-Modified: Wed, 07 Sep 2022 14:27:51 GMT
 ```
 
-#### Update payload at rendezvous point: `PUT /<rendezvous id>`
+#### Update payload at rendezvous point: `PUT <rendezvous URI>`
 
 HTTP request headers:
 
@@ -107,7 +107,7 @@ HTTP request body:
 HTTP response codes:
 
 - `202 Accepted` - payload updated
-- `404 Not Found` - rendezvous ID is not valid (it could have expired)
+- `404 Not Found` - rendezvous URI is not valid (it could have expired)
 - `413 Payload Too Large` - the supplied payload is too large
 - `412 Precondition Failed` - when `If-Match` is supplied and the ETag does not match
 - `429 Too Many Requests` - the request has been rate limited
@@ -118,7 +118,7 @@ HTTP response headers for `202 Accepted` and `412 Precondition Failed`:
 - `Expires` - required, the expiry time of the rendezvous as per [RFC7233](https://httpwg.org/specs/rfc7234.html#header.expires)
 - `Last-Modified` - required, the last modified date of the payload as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.last-modified)
 
-#### Get payload from rendezvous point: `GET /<rendezvous id>`
+#### Get payload from rendezvous point: `GET <rendezvous URI>`
 
 HTTP request headers:
 
@@ -128,7 +128,7 @@ only return data if given ETag does not match
 HTTP response codes:
 
 - `200 OK` - payload returned
-- `404 Not Found` - rendezvous ID is not valid (it could have expired)
+- `404 Not Found` - rendezvous URI is not valid (it could have expired)
 - `304 Not Modified` - when `If-None-Match` is supplied and the ETag does not match
 - `429 Too Many Requests` - the request has been rate limited
 
@@ -140,12 +140,12 @@ HTTP response headers for `200 OK` and `304 Not Modified`:
 
 - `Content-Type` - required for `200 OK`
 
-#### Cancel a rendezvous: `DELETE /<rendezvous id>`
+#### Cancel a rendezvous: `DELETE <rendezvous URI>`
 
 HTTP response codes:
 
 - `204 No Content` - rendezvous cancelled
-- `404 Not Found` - rendezvous ID is not valid (it could have expired)
+- `404 Not Found` - rendezvous URI is not valid (it could have expired)
 - `429 Too Many Requests` - the request has been rate limited
 
 ### Maximum payload size
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -10,10 +10,12 @@ information such as:
 - the user ID
 - facilitation of issuing a new access token
 - device ID for end-to-end encryption
-- device keys for end-to-end encryption
 
 To enable [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) and support any future proposals this MSC proposes a simple HTTP based protocol that can be used to
 establish a direct communication channel between two IP connected devices.
+
+This channel SHOULD be considered untrusted by both devices, and SHOULD NOT be
+used to transmit sensitive data in plaintext.
 
 It will work with devices behind NAT. It doesn't require homeserver administrators to deploy a separate server.
 
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -17,7 +17,7 @@ To enable [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/390
 establish a direct communication channel between two IP connected devices.
 
 This channel [SHOULD be considered untrusted](#confidentiality-of-data)
-by both devices, and SHOULD NOT be used to transmit sensitive data in plaintext.
+by both devices, and SHOULD NOT be used to transmit sensitive data in cleartext.
 
 It will work with devices behind NAT. It doesn't require homeserver administrators to deploy a separate server.
 
--- a/proposals/3886-simple-rendezvous-capability.md
+++ b/proposals/3886-simple-rendezvous-capability.md
@@ -68,12 +68,12 @@ HTTP request body:
 
 - any data up to maximum size allowed by the server
 
-HTTP response codes:
+HTTP response codes, and Matrix error codes:
 
 - `201 Created` - rendezvous created
-- `403 Forbidden` - forbidden by server policy
-- `413 Payload Too Large` - the supplied payload is too large
-- `429 Too Many Requests` - the request has been rate limited
+- `403 Forbidden` (`M_FORBIDDEN`) - forbidden by server policy
+- `413 Payload Too Large` (`M_TOO_LARGE`) - the supplied payload is too large
+- `429 Too Many Requests` (`M_UNKNOWN`) - the request has been rate limited
 - `307 Temporary Redirect` - if the request should be served from somewhere else specified in the `Location` response header
 
 n.b. the relatively unusual [`307 Temporary Redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) response
@@ -111,13 +111,13 @@ HTTP request body:
 
 - any data up to maximum size allowed by the server
 
-HTTP response codes:
+HTTP response codes, and Matrix error codes:
 
 - `202 Accepted` - payload updated
-- `404 Not Found` - rendezvous URI is not valid (it could have expired)
-- `412 Precondition Failed` - when `If-Match` is supplied and the ETag does not match
-- `413 Payload Too Large` - the supplied payload is too large
-- `429 Too Many Requests` - the request has been rate limited
+- `404 Not Found` (`M_NOT_FOUND`) - rendezvous URI is not valid (it could have expired)
+- `412 Precondition Failed` (`M_DIRTY_WRITE`, **a new error code**) - when `If-Match` is supplied and the ETag does not match
+- `413 Payload Too Large` (`M_TOO_LARGE`) - the supplied payload is too large
+- `429 Too Many Requests` (`M_UNKNOWN`) - the request has been rate limited
 
 HTTP response headers for `202 Accepted` and `412 Precondition Failed`:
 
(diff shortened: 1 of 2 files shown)


proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,236 @@
+# MSC3886: Simple client rendezvous capability
```
ara4n, line 1: On one hand, this is a really simple and elegant standalone function. On the other hand, I'm a bit worried that it duplicates the semantics of to-device API (i.e. basic store & forward between devices), albeit with short-polling rather than long-polling.

I wonder how bad it would be if we opened up to-device messages to guests, and used the existing APIs for rendezvous?  So a new device would go and /login as a guest to get a temporary access token, and then publish its device ID & HS url in its QR code to let another device rendezvous with it.

My only reason for proposing this is to avoid having two store-and-forward APIs which look suspiciously similar, but have different semantics (short/long poll), and so require more code for client implementors.

ara4n (changes requested):

proposals/3886-simple-rendezvous-capability.md
hughns: Understood. I'll work up an alternative based on to-device messages and see how that feels.

hughns: I have started some discussion on the to-device based alternative as part of https://github.com/matrix-org/matrix-spec-proposals/pull/3903

turt2live:
This was proposed as ready for FCP by the author, however it appears to have some conflicts with to-device messaging. For now, I'm going to put it into "needs technical design review". This doesn't mean it's not ready for FCP, just that the review's tone should take a slightly different shape.

@matrix-org/spec-core-team please review this for technical design.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+
+  B->R: PUT <rendezvous URI> Hello from B
+  R->B: 202 Accepted 
+
+  Note over A,B: Rendezvous now established
+```
+
+Please note that it is intentional that this protocol does nothing to ensure the integrity of the data exchanged at a rendezvous.
+
+### Protocol
```
richvdh, line 58: As I understand it, none of the endpoints have any requirements for an Authentication header. I think that's fine, but could you make it explicit?

proposals/3886-simple-rendezvous-capability.md
richvdh: > I wonder how bad it would be if we opened up to-device messages to guests, and used the existing APIs for rendezvous? So a new device would go and /login as a guest to get a temporary access token, and then publish its device ID & HS url in its QR code to let another device rendezvous with it.

ugh, the complexity of this feels *horrible* to me.

> My only reason for proposing this is to avoid having two store-and-forward APIs which look suspiciously similar, but have different semantics (short/long poll), and so require more code for client implementors.

Sure, having two store-and-forward APIs is rather less than ideal, but this one is so simple and easy to use that I don't really buy that it's a meaningful amount of extra code for clients comparing to have to grab a temporary access token and then start /syncing.

For me, the simplicity of this proposal outweighs the fact it looks a bit like to-device messaging. (Or even matrix rooms, if you squint hard enough and invent "ephemeral rooms".)

The only thing I'd say here is that it would be good if the "Alternatives" section in this MSC said something about this idea (even if it's just a link to MSC3903's alternatives section).

richvdh (approved):
I love this. Clear, simple and effective.

Happy to put this forward for FCP, though we should avoid doing so unless we think MSC3906 is going to be ready in short order, otherwise we end up with a feature in the spec without any particular use.

turt2live:
@hughns fyi that this has also entered "not active SCT focus" as it has a soft dependency on https://github.com/matrix-org/matrix-spec-proposals/pull/3906, which is also currently back with you.

When you get a chance to update the MSCs, we'll push 3906 forward at the same time as this one - just let us know in the SCT office.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+```mermaid
+
+sequenceDiagram
+  participant A as Device A
+  participant R as Rendezvous Server
+  participant B as Device B
+  Note over A: Device A determines which rendezvous server to use
+
+  A->R: POST /rendezvous Hello from A
+  R->A: 201 Created Location: /abc-def-123-456
```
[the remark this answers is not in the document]
hughns: It appears I never added them in the first place. Now fixed. 😇

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+`Location` and `X-Max-Bytes` headers as:
+
+```http
+Access-Control-Allow-Headers: Content-Type,If-Match,If-None-Match
+Access-Control-Allow-Methods: GET,PUT,POST,DELETE
+Access-Control-Allow-Origin: *
+Access-Control-Expose-Headers: ETag,Location,X-Max-Bytes
+```
+
+These headers are different from the ones currently specified in the [spec](https://spec.matrix.org/v1.4/client-server-api/#web-browser-clients).
```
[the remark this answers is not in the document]
hughns: I don't recall intending to change them globally, so I have clarified that they just apply to the new API endpoint.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+
+HTTP response codes:
+
+- `201 Created` - rendezvous created
+- `403 Forbidden` - forbidden by server policy
+- `413 Payload Too Large` - the supplied payload is too large
+- `429 Too Many Requests` - the request has been rate limited
+- `307 Temporary Redirect` - if the request should be served from somewhere else specified in the `Location` response header
+
+HTTP response headers for `201 Created`:
```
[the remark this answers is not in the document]
hughns: As per notes inline they are required.

proposals/3886-simple-rendezvous-capability.md
hughns: I've added a section on this.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+
+- any data up to maximum size allowed by the server
+
+HTTP response codes:
+
+- `201 Created` - rendezvous created
+- `403 Forbidden` - forbidden by server policy
+- `413 Payload Too Large` - the supplied payload is too large
+- `429 Too Many Requests` - the request has been rate limited
+- `307 Temporary Redirect` - if the request should be served from somewhere else specified in the `Location` response header
```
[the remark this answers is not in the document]
hughns: Correct. I've added a note on why 307 is specifically used.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,254 @@
...
+  R->B: 202 Accepted 
+
+  Note over A,B: Rendezvous now established
+```
+
+Please note that it is intentional that this protocol does nothing to ensure the integrity of the data exchanged at a rendezvous.
+
+### Protocol
+
+#### Create a new rendezvous point: `POST /rendezvous`
```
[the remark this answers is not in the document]
hughns: I have been assuming that it will be part of the Client-Server API and so end up somewhere under `/_matrix/client`.

I supposed it could also be specced as a whole new top-level Matrix API. This might make the spec implementation simpler (where the CORS and some other aspects vary) in which case it could be `/_matrix/rendezvous`.

Or, this thought might be opening a can of worms.

proposals/3886-simple-rendezvous-capability.md
hughns: I've move the section from [MSC3903](https://github.com/matrix-org/matrix-spec-proposals/pull/3903) alternatives section into this proposal as there is much feedback here than on MSC3903 itself.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+  loop Device A polls for rendezvous updates
+    A->>+R: GET <rendezvous URI> If-None-Match: <ETag>
+    R->>-A: 304 Not Modified
+  end
+
+  B->>+R: PUT <rendezvous URI> Hello from B
+  R->>-B: 202 Accepted 
+
+  Note over A,B: Rendezvous now established
+```
```
uhoreg, line 54: My understanding of this is that A and B take turns writing to the same rendezvous URI until they're done.  So when it's B's turn to write, A keeps polling (using the ETag) until the server says the data has changed, and vice versa.

What happens if B tries to write, but gets some sort of network error, or an error from a proxy?  If the server got B's data, but B received a network error, then it seems to me what could happen is:
- A receives B's data, thinks it's now their turn to send data, so sends their data and gets a new ETag
- B retries the request, overwriting A's data (and never receiving it)
- A polls for new data, using the new ETag
- since B overwrote A's data, the data doesn't match the ETag, so A gets the data B sent, again

So B will miss a message from A, and A will get a duplicate message.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+- `ETag` - required, ETag for the current payload at the rendezvous point as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.etag)
+- `Expires` - required, the expiry time of the rendezvous as per [RFC7233](https://httpwg.org/specs/rfc7234.html#header.expires)
+- `Last-Modified` - required, the last modified date of the payload as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.last-modified)
+
+#### Get payload from rendezvous point: `GET <rendezvous URI>`
+
+HTTP request headers:
+
+- `If-None-Match` - optional, as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.if-none-match) server will
+only return data if given ETag does not match
```
uhoreg, line 133: Might be nice for servers to have the option to delay responding until it gets content that doesn't match the ETag, so we can do long-polling.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+1. If the client is already logged in: try and use current homeserver.
+1. If the client is not logged in and it is known which homeserver the user wants to connect to: try and use that homeserver.
+1. Otherwise use a default server.
+
+## Potential issues
+
+Because this is an entirely new set of functionality it should not cause issue with any existing Matrix functions or capabilities.
+
+The proposed protocol requires the devices to have IP connectivity to the server which might not be the case in P2P scenarios.
+
```
uhoreg, line 210: One potential issue here is that if A sends a message to B, then waits for a message from B using the ETag, but the message that B sends to A happens to be exactly the same as the message that A sent, then A will get the `304 Not Modified` response, and never realize that B sent a message.  So anything built on top of this needs to ensure that a message is never identical to the preceding message.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+- `307 Temporary Redirect` - if the request should be served from somewhere else specified in the `Location` response header
+
+n.b. the relatively unusual [`307 Temporary Redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) response
+code has been chosen explicitly for the behaviour of ensuring that the method and body will not change whilst the user-agent
+follows the redirect. For this reason, no other `30x` response codes are allowed.
+
+HTTP response headers for `201 Created`:
+
+- `Location` - required, the allocated rendezvous URI which can be on a different server
+- `X-Max-Bytes` - required, the maximum allowed bytes for the payload
```
turt2live, line 86: why are these headers and not response body parameters?

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+- `Content-Length` - required
+- `Content-Type` - optional, server should assume `application/octet-stream` if not specified
+- `If-Match` - optional, as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.if-match) server will assume `*`
+if not specified
+
+HTTP request body:
+
+- any data up to maximum size allowed by the server
+
+HTTP response codes:
```
turt2live, line 114: Matrix error codes here too please, and probably a `400` definition to cover missing headers and such

turt2live: (comment applies throughout remainder of proposal)

turt2live (commented):
overall this MSC seems like a good approach for the problem. I've left some comments about things that should be clarified before the MSC can go up for FCP.

proposals/3886-simple-rendezvous-capability.md
hughns: Perhaps we could mitigate against this by using a RFC7232 `If-Match` on the PUT requests?

proposals/3886-simple-rendezvous-capability.md
hughns: On reflection I would agree that if [it is](https://github.com/matrix-org/matrix-spec-proposals/pull/3886#discussion_r1135982763) going to be part of the C-S API then it would make sense to consider consistency with the rest of the C-S API where headers are not used.

proposals/3886-simple-rendezvous-capability.md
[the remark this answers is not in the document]
DMRobertson: Done in 931cf073.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+HTTP request headers:
+
+- `Content-Length` - required
+- `Content-Type` - optional, server should assume `application/octet-stream` if not specified
+
+HTTP request body:
+
+- any data up to maximum size allowed by the server
+
+HTTP response codes:
```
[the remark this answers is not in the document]
DMRobertson: See aee7d81c. I have proposed `M_DIRTY_WRITE` for the `HTTP 412 Precondition Failed` case where Alice attempts to write but hasn't seen Bob's latest write.

proposals/3886-simple-rendezvous-capability.md
DMRobertson: 90a8b495 defines a 400 response.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,302 @@
...
+- `429 Too Many Requests` (`M_UNKNOWN`) - the request has been rate limited
+- `307 Temporary Redirect` - if the request should be served from somewhere else specified in the `Location` response header
+
+n.b. the relatively unusual [`307 Temporary Redirect`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) response
+code has been chosen explicitly for the behaviour of ensuring that the method and body will not change whilst the user-agent
+follows the redirect. For this reason, no other `30x` response codes are allowed.
+
+HTTP response headers for `201 Created`:
+
+- `Location` - required, the allocated rendezvous URI which can be on a different server
```
DMRobertson, line 86: Presumably this needs URI needs to be not guessable, to prevent attackers from guessing this and impersonating the intended recipient?

proposals/3886-simple-rendezvous-capability.md
DMRobertson: Are we concerned with just these two headers? Or do we want _all_ of the response and request headers to be expressed via HTTP bodies?

proposals/3886-simple-rendezvous-capability.md
DMRobertson: ISTM that every PUT should be required to cite a previous ETag so that the rendezvous server can enforce a linear ordering. (The initial ETag is included in the POST and GET response, so both A and B should be fully aware of it.)

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,302 @@
...
+- `400 Bad Request` (`M_MISSING_PARAM`) - no `Content-Length` was provided.
+- `404 Not Found` (`M_NOT_FOUND`) - rendezvous URI is not valid (it could have expired)
+- `412 Precondition Failed` (`M_DIRTY_WRITE`, **a new error code**) - when `If-Match` is supplied and the ETag does not match
+- `413 Payload Too Large` (`M_TOO_LARGE`) - the supplied payload is too large
+- `429 Too Many Requests` (`M_UNKNOWN`) - the request has been rate limited
+
+HTTP response headers for `202 Accepted` and `412 Precondition Failed`:
+
+- `ETag` - required, ETag for the current payload at the rendezvous point as per [RFC7232](https://httpwg.org/specs/rfc7232.html#header.etag)
+- `Expires` - required, the expiry time of the rendezvous as per [RFC7233](https://httpwg.org/specs/rfc7234.html#header.expires)
```
DMRobertson, line 127: Is the intention that the expiry time is incremented every time the rendezvous payload is updated?

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,300 @@
...
+QR code.
+
+In order to facilitate this the two devices need some bi-directional communication channel which they can use to exchange
+information such as:
+
+- the homeserver being used
+- the user ID
+- facilitation of issuing a new access token
+- device ID for end-to-end encryption
+- device keys for end-to-end encryption
```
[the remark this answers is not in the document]
DMRobertson: dcbbcb0f and 74d90948 clarify that this is an untrusted communications channel.

proposals/3886-simple-rendezvous-capability.md
DMRobertson: The response body for GET ought to be just the payload itself. Since the payload data is an arbitrary byte sequence, it would be painful to embed this in JSON. Therefore I would encourage the current GET response headers (Content-Type, ETag, Expires, Last-Modified) to continue to be expressed via headers. For consistency it makes sense to do so in all other resposnes.

For POST this leaves the two highlighted headers: Location and X-Max-Bytes. We could present them as a JSON-encoded body, but it would seem odd to spread the POST response metadata in two places without any meaningful distinction to justify it. My vote would be to leave things as they are. But I neither have a vote, nor any strong opinions.

proposals/3886-simple-rendezvous-capability.md
DMRobertson: 58f1e86f does this.

proposals/3886-simple-rendezvous-capability.md
DMRobertson: I have assumed so in 65d697c2

proposals/3886-simple-rendezvous-capability.md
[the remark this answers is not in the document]
DMRobertson: Although I self-bikeshedded a different name in 3fecfcd5.

proposals/3886-simple-rendezvous-capability.md
dkasak: Is this still a problem? From the current text, it sounds like `304 Not Modified` will only be returned when a matching ETag is supplied in a `If-None-Match`. Given the resolution of [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/3886#discussion_r1177126767), clients will have to supply a previous ETag when doing a PUT, which means we no longer have to rely on the sameness of the content to decide whether the content has been modified. That is, a PUT request that specifies the previous ETag `A` should be regarded as altering the payload at `A`, even if the payload is unchanged, and therefore, it should be assigned a new ETag.

proposals/3886-simple-rendezvous-capability.md
```diff
@@ -0,0 +1,353 @@
...
+- facilitation of issuing a new access token
+- device ID for end-to-end encryption
+- data for establishing a secure communication channel (e.g. via
+  [MSC3903](https://github.com/matrix-org/matrix-spec-proposals/pull/3903))
+
+To enable [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) and support any future proposals this MSC proposes a simple HTTP based protocol that can be used to
+establish a direct communication channel between two IP connected devices.
+
+This channel [SHOULD be considered untrusted](#confidentiality-of-data)
+by both devices, and SHOULD NOT be used to transmit sensitive data in plaintext.
```
dkasak, line 20:
```suggestion
by both devices, and SHOULD NOT be used to transmit sensitive data in cleartext.
```

Technical nit: Cleartext is stuff that gets sent without encryption. Plaintext is stuff you get once you decrypt ciphertext.

proposals/3886-simple-rendezvous-capability.md
dkasak: > The response body for GET ought to be just the payload itself. Since the payload data is an arbitrary byte sequence, it would be painful to embed this in JSON.

We could base64 encode the payload, but I tend to agree. Not all things need to be shoehorned through JSON.

hughns:
It is proposed that [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108) supersedes this MSC.

hughns:
Closing this PR as https://github.com/matrix-org/matrix-spec-proposals/pull/4108 is now ready for review.

PR #3887, opened 2022-09-09
MSC3887: List matching push rules

Fizzadar:
[Rendered](https://github.com/Fizzadar/matrix-spec-proposals/blob/list-match-push-rules/proposals/3887-push-rule-list-matching.md)

### DIFF
--- a/proposals/3887-push-rule-list-matching.md
+++ b/proposals/3887-push-rule-list-matching.md
@@ -1,4 +1,4 @@
-# MSCXXXX: List matching push rules
+# MSC3887: List matching push rules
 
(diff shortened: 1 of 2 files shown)


clokep:
@Fizzadar Can we close this in-lieu of MSC3966 which is extremely similar, but defines a separate push rule condition?

Fizzadar:
> @Fizzadar Can we close this in-lieu of [MSC3966](https://github.com/matrix-org/matrix-doc/pull/3966) which is extremely similar, but defines a separate push rule condition?

Yes, that’s much better!

PR #3401, opened 2021-09-19
MSC3401: Native Group VoIP Signalling

ara4n:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/group-voip/proposals/3401-group-voip.md)

Obsoletes #2359

PR #3888, opened 2022-09-09
MSC3888: Voice Broadcast

weeman1337:


proposals/3888-voice-broadcast.md
```diff
@@ -0,0 +1,101 @@
...
+## Alternatives
+
+### WebRTC Based Approach
+
+Although WebRTC seems suitable for this MSC at first glance, there are several reasons not choosing it:
+
+- Voice messages already exist.
+- Hence out of the box fallback for clients that won't support this MSC.
+- Element Call is, at the time of writing this spec, not ready yet.
+- Custom WebRTC implementation has higher effort than building on something that already exist.
```
turt2live, line 81: Just ftr: this justification isn't strong enough, imo. Voice messages aren't really meant for this sort of thing and fallback support isn't always needed. Element Call is fairly irrelevant at this level, and custom WebRTC implementations shouldn't be needed thanks to voice/video rooms? (which also raises this MSC further into question: why are *recorded* voice rooms not feasible?)

HarHarLinks: Based on the requirement stated in the introduction, and assuming power levels can restrict (other) users from talking in the room, voice rooms would exactly fulfill this. 

The (implied) idea of being able to pause and rewind an ongoing stream is another question: video rooms/native calls may eventually want to support that (e.g. scroll back in a live video lesson), however it seems like a huge feature creep into becoming a streaming + vod service. In contrast to live (only) calls, storing all this media data could become an issue eventually?

proposals/3888-voice-broadcast.md
```diff
@@ -0,0 +1,119 @@
...
+A new state event `m.voice_broadcast_info` will be introduced. This state event
+identifies a broadcast and provides its state, such as "started" or "paused".
+In addition to that `m.voice` messages will receive a relation to the state
+event to mark them as voice broadcast chunks.
+
+`m.voice_broadcast_info` event example:
+
+```json
+{
+    "type": "m.voice_broadcast_info,
```
HarHarLinks, line 28:
```suggestion
    "type": "m.voice_broadcast_info",
```

proposals/3888-voice-broadcast.md
weeman1337: I've marked this MSC as on hold. We may go for an intermediate solution based on message chunks. See the [issue in Element Meta](https://github.com/vector-im/element-meta/discussions/632).

proposals/3888-voice-broadcast.md
```diff
@@ -0,0 +1,14 @@
...
+Compared to already existing voice messages it should be possible to listen to
+voice broadcast at the time they are being recorded.
+Think of this being a live-podcast over Matrix.
+
+Some use-case scenarios for this feature can be to verbally provide updates,
+trainings, or talk to a large number of people.
+
+## State of this MSC
+
+This MSC is on hold until Element Call with SFU is ready.
```
turt2live, line 14: It should be more than possible to outline a technical solution without Element Call existing. At a spec level, Element Call (with or without SFU) is just MSC3401 which means this MSC can start layering itself on top of that. As a work in progress it doesn't need to be complete, but reserving MSC numbers for arbitrary milestones isn't great :(

proposals/3888-voice-broadcast.md
```diff
@@ -0,0 +1,19 @@
+# MSC3888: Voice Broadcast
```
ara4n, line 1: For anyone lost like me, the actual voice broadcast behaviour that Element implements is documented at https://github.com/vector-im/element-meta/discussions/632.

For the record, it's definitely not obvious to me that voice broadcast should be done by via MSC #3401 VoIP.  For one thing, it'd require a stream recording/broadcasting SFU component for the feature to work (which doesn't currently exist, and could end up being very resource intensive and expensive to run (and thus paid) given you'd effectively have to run a headless brwoser).  Secondly, it would very likely break E2EE in the general case, in order to mix streams together serverside.  Specialcasing E2EE to still work for a single stream would be pretty painful, as we'd have to somehow match S-frame or SRTP encryption up with custom e2ee encryption for HLS segments.  Alternatively, if you dialled into the broadcast SFU via WebRTC rather than HLS (in order to keep E2EE and minimise latency), we'd have to implement a full mediaserver in order to scrub back and forth within the recorded stream.

So, my hunch is that the right way to implement VB is either via streaming file transfers or by the chunked file transfer approach [used here](https://github.com/vector-im/element-meta/discussions/632).  In order for streamed E2EE filetransfers to work, we'd have to switch to AES-GCM (given we can't calculate the hash on something which hasn't streamed yet).  On the plus side, we already have (in-house to Element) a streamed file repository in the form of `glow`.  However, in order to randomly seek within the file (via Range headers) without downloading the whole file first, we'd have to split the file into chunks anyway, given you can't randomly seek in an AES-GCM stream.  Alternatively, we could keep it as AES-CTR and handle integrity via a separate stream of hashes somehow (possibly publishing a new hash as a event relation every N KB blocks or something horrible, or devising a seekable variant of AES-GCM).  At this point, it begs the question as to why you wouldn't just send separate files around N KB blocks anyway, which is how we ended up with the current implementation in Element.  The only disadvantage is that you have latency of the size of one block in listening to the message, without to make vicious changes like streaming filetransfers or mediaserver SFUs.

weeman1337: I've incorporated the your comments into the spec change.

ara4n:
woo, thanks for populating this! (adds to infinite todo list to review)

PR #3890, opened 2022-09-13
MSCXXX: Remotely silence local notifications

kerryarchibald:


gsouquet:
This pull request now has an implementation https://github.com/matrix-org/matrix-js-sdk/pull/2700

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,63 @@
...
+- sounds, including noisy notifications and ringers
+
+In-app notifications will not be affected, with in-app notification center, badges, unread markers and counts remaining
+the same.
+
+*This proposal only refers to notifications that are generated on the client. Silencing of push notifications is covered
+by [MSC3881: Remotely toggle push notifications](https://github.com/matrix-org/matrix-spec-proposals/pull/3881)*
+
+## Proposal
+Add a new account data event `m.local_notification_settings.<device-id>` with content:
```
gsouquet, line 21: This account data event is created by clients that support this feature, but is only relevant for the lifecycle of the device.
I believe we should add a section about cleaning up these account data events server side when a session is deleted to avoid proliferation of account data events.

kerryarchibald: Added a section about this

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,67 @@
...
+The spec allows for pushers to be assigned a
+[`profile_tag`](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3pushersset) which can be used to
+define per-device push rule sets. In combination with the `notify_in_app` action proposed in
+[MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768) this would allow to toggle a pusher between the
+`global` push rule set and a push rule set where all rules with `notify` actions were overridden to use `notify_in_app`.
+Furthermore, the overrides could be simplified through cascading profile tags as proposed in
+[MSC3837](https://github.com/matrix-org/matrix-spec-proposals/pull/3837). Keeping the two sets in sync would, however,
+not be trivial. Additionally, profile tags are only partially spec'ed and there is active interest in
+[removing](https://github.com/matrix-org/matrix-spec/issues/637) them entirely. If feasible, this would allow for remote
+control of notifications for clients using http-pushers and local notification generation.
```
t3chguy, line 56: It is feasible, I think speccing device-scoped push rules even maybe just allowing `override` rules to manage this would be better as otherwise you have a "local" silencing, but still causing the device to get awoken via its pusher.

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,67 @@
+# MSC3890: Remotely silence local notifications
+Some clients (eg Element Web) do not use http pushers for notifications, but rather generate their own notifications in
+the client based on the `/sync` response. Users should be able to remotely toggle on and off these notifications (as
+well as [push notifications](https://github.com/matrix-org/matrix-spec-proposals/pull/3881)) to control interruptions
```
turt2live, line 4:
```suggestion
well as [push notifications (MSC3881)](https://github.com/matrix-org/matrix-spec-proposals/pull/3881)) to control interruptions
```

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,67 @@
...
+the same.
+
+*This proposal only refers to notifications that are generated on the client. Silencing of push notifications is covered
+by [MSC3881: Remotely toggle push notifications](https://github.com/matrix-org/matrix-spec-proposals/pull/3881)*
+
+## Proposal
+Add a new account data event `m.local_notification_settings.<device-id>` with content:
+```jsonp
+{
+    is_silenced: boolean
```
turt2live, line 24: generally we prefer to phrase things in the positive: `noisy_notifications: true` (for example)

proposals/3890-remote-local-notification-toggle.md
turt2live: ^the proposal, like MSC3881, should mention why per-device push rules aren't suitable

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,67 @@
...
+not be trivial. Additionally, profile tags are only partially spec'ed and there is active interest in
+[removing](https://github.com/matrix-org/matrix-spec/issues/637) them entirely. If feasible, this would allow for remote
+control of notifications for clients using http-pushers and local notification generation.
+
+## Security considerations
+N/A
+
+## Dependencies
+Communication of deletion of account data to clients is enabled by [MSC3391: Removing account
+data.](https://github.com/matrix-org/matrix-spec-proposals/pull/3391)
```
turt2live, line 63: This feels like a relatively arbitrary dependency which may slow the MSC down, ftr: it should be possible to just let the account data stack up for now and suggest in the MSC prose that "a server is to clean up deleted device account data when able to do so, such as with MSC3391"

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,67 @@
...
+`m.local_notification_settings.<device-id>` account_data events for the given device, and communicate these changes to
+clients as described in [MSC3391](https://github.com/matrix-org/matrix-spec-proposals/pull/3391).
+
+
+#### Support
+Clients that implement `m.local_notification_settings.<device-id>` notification filtering should ensure a
+`local_notification_settings` event exists for the active device to indicate support for that device. Clients should
+only expose local notification silencing controls for devices that have indicated their support.
+
+## Alternatives
```
turt2live, line 45: A consideration might be to have per-device account data similar to how we already have per-room account data: though awkward at first, the intention behind it would be to allow other devices to change the device's account data values, specifically for when cases like this matter. 

Another example might be to have per-device [client] settings that might be possible to change from another client, like breadcrumbs or theme. This would be part of the overal "granular settings" idea created years ago by element-web (then vector-web).

turt2live (commented):
the author has suggested that this might be ready for FCP, though I have questions/concerns about the approach. Specifically whether the idea of per-device account data would fit "better" for this particular usecase, and others, instead of account-data-with-subtypes.

proposals/3890-remote-local-notification-toggle.md
Johennes: > [...] as otherwise you have a "local" silencing, but still causing the device to get awoken via its pusher

This doesn't apply here as the proposal specifically covers clients _without_ pushers that generate notifications locally from `/sync` responses.

turt2live: Pushers are not the same as per-device push rules, confusingly. Per-device push rules are a theory that push rules (which are what clients use to calculate counts locally) could be scoped to a particular device, allowing for locally-silenced notification [counts].

Pushers also use push rules to determine counts/highlights, which also means per-device push rules could also affect how devices with pushers could work.

turt2live: right, but the underlying argument is still that this MSC might not be neccessary if per-device push rules were to actually exist. Account data feels like the wrong mechanism for this at the moment, regardless of having a pusher or not.

Johennes: Right, I'm not disagreeing with that. My point was just that needlessly waking up clients is not the proper reason to favor per-device push rules because that happens neither here nor in [MSC3881](https://github.com/matrix-org/matrix-doc/pull/3881).

Anyway, I think we're aligned. What's missing in both MSCs is a more elaborate description of why we think these proposals are favorable to per-device push rules. We'll be working an that. 👍

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,65 @@
+# MSC3890: Remotely silence local notifications
+Some clients (eg Element Web) do not use http pushers for notifications, but rather generate their own notifications in
+the client based on the `/sync` response. Users should be able to remotely toggle on and off these notifications (as
+well as [push notifications (MSC3881)](https://github.com/matrix-org/matrix-spec-proposals/pull/3881)) to control
+interruptions and reachability in private and professional contexts.
```
clokep, line 5: I think clients which use push also process notifications locally when doing `/sync` -- are they also expected to implement this? It seems like this is meant to be the matrix-js-sdk version of MSC3881, but it isn't really clear the interaction of them.

If you remote toggle a notifications for a client which supports push notifications does that mean setting the account data from MSC3890 (this MSC) & disabling the pusher (see MSC3881) for that device?

anoadragon453:
Synapse has gained experimental support for removing the `m.local_notification_settings.<device-id>` entry from a user's account data when the corresponding device is deleted: https://github.com/matrix-org/synapse/pull/14775

I've updated the MSC description to reflect this :slightly_smiling_face: cc @kerryarchibald

proposals/3890-remote-local-notification-toggle.md
```diff
@@ -0,0 +1,88 @@
...
+  trigger a system notification or sound.
+
+#####Displaying notification settings:
+
+Clients should endeavour to display notification settings that reflect the rules above.
+
+- While a registered pusher exists for a device, notifications should be displayed as enabled. While the client has the
+  ability to register/remove pushers for the device, toggling this setting should register or remove the pusher. For
+  servers that support [MSC3381](https://github.com/matrix-org/matrix-doc/pull/3881), notifications should be displayed
+  as enabled according to the the `enabled` setting of the pusher. Changing this setting should update the `enabled`
```
Johennes, line 47:
```suggestion
  as enabled according to the `enabled` setting of the pusher. Changing this setting should update the `enabled`
```

proposals/3890-remote-local-notification-toggle.md
kerryarchibald: Good point. I've added a section about how these settings should interact [here](https://github.com/matrix-org/matrix-spec-proposals/pull/3890/commits/10be7fbc1ac4d96313e243e960c3bb0d5296194d)

proposals/3890-remote-local-notification-toggle.md
kerryarchibald: I'm not familiar with the "granular settings" idea. I'm not sure how many client level settings it would be desirable/useful to have on the server and editable from other devices. Quick setup of new devices that mirror existing  ones could be one use case.

If read/write is allowed from any device in the account (which it would have to be to fit this use case) then the added utility of per device account data over profile account data seems quite small.

I've added a small note in the alternatives section which I can explore more fully if needed.

PR #2545, opened 2020-05-15
MSC2545: Image Packs (Emoticons & Stickers)

Sorunome:
> [!WARNING]
> This proposal is edited by one or more PRs post-acceptance. Use the "rendered" link instead of this PR's diff to see the final MSC text.

----

[Rendered](https://github.com/matrix-org/matrix-doc/blob/main/proposals/2545-emotes.md)

Signed-off-by: Sorunome <mail@sorunome.de>

Author: @Sorunome
Shepherd: @anoadragon453

## Future MSCs
 - Image pack cross-room referencing / suggesting (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r753830655)
 - Additional metadata for stickers and emotes to find their pack (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r753881475)
 - Image pack ordering (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r910188726)
 - Multiple shortcodes for a single image (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r598354072)
 - E2EE support (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r677009720)
 - Image-Pack rooms (needing a room type) (https://github.com/matrix-org/matrix-spec-proposals/pull/2545#discussion_r910190308)

## Differences between historical implementations and the current MSC

See [this comment](https://github.com/matrix-org/matrix-spec-proposals/pull/2545#issuecomment-4144315200) for a high-level overview of the changes since the previous set of implementations and today's MSC.

Implementation authors are recommended to re-read the MSC in general, as it has changed significantly in the past year as the spec was finalised.

### Emote rendering (rendering of the `` tag)
 - element-web
 - revolution (even handles `data-mx-emoticon`)
 - nheko
 - fluffychat (even handles `data-mx-emoticon`)
 - neochat
 - cinny
### Sending, using the mentioned events here
 - revolution (emoticons)
 - fluffychat (emoticons, stickers)
 - neochat (emoticons, at least partial)
 - nheko (emoticons, stickers)
 - cinny (emoticons, stickers)
 - kazv (stickers)

### Bridges
 - mx-puppet-discord
 - matrix-appservice-discord (partially, sending `data-mx-emoticon` only)

### FluffyChat
Data model: https://gitlab.com/famedly/company/frontend/libraries/matrix_api_lite/-/merge_requests/26
[...]

PR #3892, opened 2022-09-14
MSCXXX: Custom Emotes with Encryption

nmscode:
[Rendered](https://github.com/nmscode/matrix-spec-proposals/blob/emotes/proposals/xxxx-emotes.md)

Related MSCs: [MSC2545](https://github.com/matrix-org/matrix-spec-proposals/pull/2545) and [MSC1951](https://github.com/matrix-org/matrix-doc/pull/1951)

## Implementation PRs
Element web: https://github.com/matrix-org/matrix-react-sdk/pull/9240

## Implemented but not PR yet
Element android: https://github.com/nmscode/element-android

### DIFF
--- a/proposals/xxxx-emotes.md
+++ b/proposals/xxxx-emotes.md
@@ -53,3 +53,4 @@ This could be extended to spaces although the main current implementation does n
 Add custom emotes for reactions and the emoji picker. 
 <br/>
 Other MSCs introduced the idea of shareable packs. Those would be compatible with image packs except for encrypted rooms. It might be simpler to allow users to download a zip file of all the emotes in a room.
+<br/>
--- a/proposals/xxxx-emotes.md
+++ b/proposals/xxxx-emotes.md
@@ -1,11 +1,11 @@
 # MSCXXXX: Emotes
-##Proposal
+## Proposal
 Every emote proposal uses a shortcode and a way to store the image and make the client render it. Under this proposal they are not sent in the message source as <img> tags, which is what many non-Element clients use. In this proposal, the message source is not edited and it is sent as a shortcode. Rather it is up to the client to load it when rendering. In the current implementation/proposal it is rendered in the message as follows:
 ```html
  <img class="mx_Emote" title=":shortcode:" src="blob:https://example.com/blob-guid">
 ```
 This is similar to current implementations in structure, but rather than having clients render img tags from the message source they are added during construction of the message html.
-##Emote sources
+## Emote sources
 ###Room emotes
 Room emotes are per-room emotes that are defined in an m.room.emote event. 
 The emotes are defined inside of a dict which has the shortcode as a key and the either encrypted or unencrypted source for the emote as a value.
@@ -27,19 +27,19 @@ For non encrypted rooms it follows:
 
 Note: this format could be modified, but would have to be agreed upon so the clients can load them in properly.
 
-##Alternatives:
+## Alternatives:
 Other proposals regarding this issue: [MSC2545](https://github.com/matrix-org/matrix-spec-proposals/pull/2545) and [MSC1951](https://github.com/matrix-org/matrix-doc/pull/1951)
 
 Below is a comparison highlighting the differences between this and other proposals.
-###Advantages
+### Advantages
 Security: Emotes in private rooms can be encrypted so that potentially sensitive data is not publicly available to the server. Emotes in public rooms are necessarily unencrypted. One note is that although emotes are encrypted they are available to anyone in the private room, and server admins can join private rooms so they are not fully e2ee like messages. If server admins lose the ability to join any private room
 Message source is not edited. The client does not edit the actual sent message but renders emotes locally. Leaving the message source untouched could be helpful for future updates or deleting emotes/changing the way emotes are rendered.
 Rendering. Current implementations use data-mx-emoticon types in the img tag. This uses the mx_Emote class name although that could also be changed since it is rendered locally.
 Compatibility with other MSCs: Although certain aspects of other MSCs cannot be kept the same, many general ideas from those could be carried over.
-###Disadvantages
+### Disadvantages
 Lack of user-specific emotes. It might be possible to implement this in public rooms if the client pulls every user’s public emote data to load. Generally user specific emotes goes against the idea of emotes being a shared set of images.
 Might be more difficult for clients to implement. Requires existing clients to change their implementations.
-##Additional Points
+## Additional Points
 This could be extended to spaces although the main current implementation does not have this. Spaces would take priority over individual room emotes if there is a shortcode clash. 
 Add custom emotes for reactions and the emoji picker. 
 Other MSCs introduced the idea of shareable packs. Those would be compatible with image packs except for encrypted rooms. It might be simpler to allow users to download a zip file of all the emotes in a room.
--- a/proposals/xxxx-emotes.md
+++ b/proposals/xxxx-emotes.md
@@ -49,8 +49,8 @@ It might be possible to implement this in public rooms if the client pulls every
 Requires existing clients to change their implementations.
 ## Additional Points
 This could be extended to spaces although the main current implementation does not have this. Spaces would take priority over individual room emotes if there is a shortcode clash. 
-<br/>
+<br/><br/>
 Add custom emotes for reactions and the emoji picker. 
-<br/>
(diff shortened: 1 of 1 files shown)


proposals/3892-custom-emotes-with-encryption.md
```diff
@@ -0,0 +1,61 @@
...
+## Emote sources
+### Room emotes
+Room emotes are per-room emotes that are defined in an m.room.emotes event. They can be uploaded from the room settings by moderators or above.
+The emotes are defined inside of a dict which has the shortcode as a key and the either encrypted or unencrypted source for the emote as a value.
+### Storage format
+For encrypted rooms the m.room.emotes event follows the format:
+```
+{
+"content":{
+	"short": IEncryptedFile (a json object with information on how to decrypt the file)
```
deepbluev7, line 20: So you are storing the decryption key in plain text?

proposals/3892-custom-emotes-with-encryption.md
```diff
@@ -0,0 +1,61 @@
+# MSC3892: Emotes
+## Proposal
+Every emote proposal uses a shortcode and a way to store the image and make the client render it. Under this proposal they are not sent in the message source as img tags, which is what many non-Element clients use. In this proposal, the message source is not edited and it is sent as a shortcode. Rather, it is up to the client to load it when rendering messages. Emotes are added in message by sending a short piece of text called the shortcode between colons.
+<br/><br/>
+In the current implementation/proposal it is rendered in the message html as follows:
+
+```html
+ <img class="mx_Emote" title=":shortcode:" src="blob:https://example.com/blob-guid">
```
deepbluev7, line 8: Is this a blob encoded image or a link to an image on some website or should this be an mxc url? Currently the spec doesn't allow or at least discourages https:// as image links.

anoadragon453:
I've added a `needs-implementation` label as part of the typical "new MSC" flow. I have yet to look at the linked implementation to see if it accurately reflects the MSC.

proposals/3892-custom-emotes-with-encryption.md
MTRNord: From a security pov https should not be used as it leaks the client ip to a third party while an msc usually goes to the homeserver which already knows the ip.

nmscode: In my current implementation it uses an mxc in unencrypted and a blob in encrypted rooms. It could be changed though.

proposals/3892-custom-emotes-with-encryption.md
nmscode: Yes. I will edit the proposal to be more clear. Ideally this would be encrypted as well though for full E2EE.

proposals/3892-custom-emotes-with-encryption.md
```diff
@@ -0,0 +1,63 @@
+# MSC3892: Emotes
+## Proposal
+Every emote proposal uses a shortcode and a way to store the image and make the client render it. Under this proposal they are not sent in the message source as img tags, which is what many non-Element clients use. In this proposal, the message source is not edited and it is sent as a shortcode. Rather, it is up to the client to load it when rendering messages. Emotes are added in message by sending a short piece of text called the shortcode between colons.
```
tulir, line 3: This seems like a very bad idea.

The `body` field is plaintext, it has no further meaning in the spec (for example, `body` is not markdown, even though it is easy to incorrectly assume it is by looking at what clients usually put there). All rich text is currently done with HTML in the `formatted_body` field. Mixing HTML with a custom syntax makes it much more difficult for clients to implement, as they'd all need to go and parse custom emojis manually in addition to normal HTML rendering.

It would also create unnecessary bugs/edge cases, like where exactly do you parse the plaintext syntax? HTML isn't going anywhere, so you'd have to parse the shortcode syntax from HTML (you can't use `body` when the message has other formatting like bold, strikethrough, blockquotes, etc, as body doesn't have any of that). Element web has some plaintext parsing to add tooltips for unicode emojis, but that code is buggy: if you put an emoji in a HTML attribute, it'll expand it and corrupt the attribute value in the rendered version.

The proper solution would be using HTML to represent custom emoji. Something like `` could make sense, but it could even be an entirely custom tag, like `<mx-emoji shortcode="meow"/>`. Clients would likely need to convert those to whatever format they need internally for rendering, but it's much easier to do a correct and bug-free implementation with a HTML parser and walking the tree than parsing a custom plaintext syntax inside HTML.

If you really want to go with an entirely custom syntax, the MSC should have some very strong justification as to why it needs an exception to the standard way of formatting.

proposals/3892-custom-emotes-with-encryption.md
```diff
@@ -0,0 +1,63 @@
...
+<br/>
+#### Message source is not edited
+The client does not edit the actual sent message but renders emotes locally. Leaving the message source untouched could be helpful for future updates or deleting emotes/changing the way emotes are rendered.
+<br/>
+#### Rendering
+Current implementations use data-mx-emoticon types in the img tag. This uses the mx_Emote class name although that could also be changed since it is rendered locally.
+Compatibility with other MSCs: Although certain aspects of other MSCs cannot be kept the same, many general ideas from those could be carried over.
+### Disadvantages
+#### Lack of user-specific emotes. 
+It might be possible to implement this in public rooms if the client pulls every user’s public emote data to load. Generally, however, user specific emotes goes against the idea of emotes being a shared set of images.
```
tulir, line 54: This also seems like a deal-breaker, since you'd have to install emoji packs in all rooms where you want to use them in. Both Telegram and Discord allow using emoji packs that aren't installed in the chat/guild. Bridging would also be significantly harder if everything had to be added to the room state first.

Being meant for sharing isn't really a valid excuse either: sharing a set of images doesn't require the pack to be installed in every room, just one room that you share is enough for that. The ability to reference emojis in other rooms would already be much better, but that'd be hard to implement since we don't have proper peeking yet (MSC2444 and MSC2753 are the main proposals for that).

I think the best option would just be having the option to embed emoji metadata in the message itself. Anything embedded this way would also actually be encrypted without having to figure out encrypted state events (although if the same image is in another room in room state, that would still be unencrypted). Something like this (`<mx-emoji/>` being a potential solution to my other comment about the syntax):

```json
{
  "msgtype": "m.text",
  "format": "org.matrix.custom.html",
  "body": ":meow:",
  "formatted_body": "<mx-emoji shortcode=\"meow\"/>",
  "custom_emoji": {
    "meow": {
      "url": "mxc://...",
      "key": {"k": "...", "kty": "oct", "key_ops": ["encrypt", "decrypt"], "ext": true, "alg": "A256CTR"},
      "iv": "...",
      "hashes": {"sha256": "..."},
      "v": 2
    }
  }
}
```

nmscode: This could definitely be a solution for user-specific emotes. Room emotes should be updated when the room's emotes are updated so locking in the file at time of sending is not ideal.

proposals/3892-custom-emotes-with-encryption.md
nmscode: Formatted body does not solve the bugs/edge cases but rather just changes the point where you have to deal with them/parse the text (during composing/sending vs rendering). Since clients (element, cinny, fluffychat, etc.) already send and/or parse emojis in only the body field (with the exception of large emojis) it makes sense that custom emotes would do something similar. There isn't a great justification for adding formatted bodies to all messages with custom emotes while regular emojis do not.

uhoreg, line 3: Please define what you mean by "emote".  Also, please wrap your lines to ~80 or ~120 characters.

nmscode: Added the definitions and cut the line lengths down.

turt2live:
It's been mentioned OOB that this has an implementation and so should be safe to remove the `needs-implementation` label, however I believe that we're not quite there yet. It's great to see that this has active implementations in Element, though for a feature of this size, scale, and importance it's more valuable to have several clients *adopt* the MSC as an implicit show of support for the technical direction. Specifically, we'd want to see clients with implementation in front of users for enough time to reveal any usability issues.

For clarity, this does not need to be done in Element either. Element can serve as one of the implementations (if adopted), but generally would be a small percentage of the overall consideration.

PR #1597, opened 2020-08-19
Better spec for matrix identifiers

richvdh:
PRs: #1598 
Documentation: https://github.com/matrix-org/matrix-doc/blob/rav/proposals/id_grammar/proposals/1597-id-grammar.md

A proposal to clarify a bunch of rules about what is and is not allowed in various identifiers.

This is intended to fix #666, #667, #1190 and #1257.

PR #3896, opened 2022-09-24
MSCxxxx: Appservice media

tezlm:


rapiz1:
The duplication of media immediately came to my mind when thinking about building bridges for my use. I joined lots of rooms in an IM service and these rooms may upload various images, files, etc. But I only need a small portion of the files. And for some rooms I don't check often so even images I don't need them. Given that I run matrix on AWS for personal use, it will cause a considerable storage cost. So this is a big blocker for me to adopt matrix and I think this MSC is very useful.

rapiz1:
Also, I wonder if it's possible to solve the problem by adapting [MSC2246: Asynchronous media uploads](https://github.com/matrix-org/matrix-spec-proposals/pull/2246).

With MSC2246 implemented, I wonder if it can be done with:
1. `/create` a media, giving it an infinite lifetime
2. retrieve the media and `/upload` it when a `/download` request is received

Not familiar with the matrix specs though, just some thoughts

proposals/3896-appservice-media.md
```diff
@@ -0,0 +1,47 @@
...
+      regex: "foobar-.*"
+```
+
+In this case, fetching `mxc://server.tld/foobar-bazqux` leads to a request to
+`/_matrix/app/v1/media/foobar-bazqux`.
+
+Appservices may set `Cache-Control` on their response. Homeservers should cache the response, though
+they may remove cached remote media to save space.
+
+## Potential issues
```
Kladki, line 29: Since currently media ids are largely generated at random, it is very possible that existing media could collide with new namespaced media. Because of this, there needs to be some sort of way to disambiguate this media, either by adding some sort of prefix to media by appservices that homeservers cannot use, vice versa, or something completely different.

Kladki: Not to sure how much of an issue this is, considering:
- Dendrite uses hex for media IDs (`[0-9a-f]`)
- Conduit uses alphanumeric chars for media IDs (`[0-9a-zA-Z]`)
- Synapse uses alphabetical chars (`[a-zA-Z]`)

So perhaps something should be added to the spec stating that for automatically generated media IDs, only alphanumeric characters should be used, so that appservices can use other characters to not collide.

turt2live: This feels like a similar problem to users: if users already exist under the appservice namespace when the appservice is introduced, they are adopted into the namespace as well. For media, we'd do the same: media which matches the namespace is adopted by the appservice, regardless of when it was created.

The grammar concern is best handled by another MSC (https://github.com/matrix-org/matrix-spec-proposals/pull/1597, primarily)

Kladki:
I see. I will bring this up in the aforementioned MSC then, thanks!

proposals/3896-appservice-media.md
Kladki: I see. I will bring this up in the aforementioned MSC then, thanks!

proposals/3896-appservice-media.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Alternatives
+
+*none*
+
+## Security considerations
+
+*none*
+
+## Unstable prefix
```
Kladki, line 41: Should probably specify that for the endpoint, the unstable prefix is `/_matrix/app/unstable/org.eu.celery.msc3896/media/{mediaId}`

proposals/3896-appservice-media.md
```diff
@@ -0,0 +1,47 @@
...
+Appservices often need to bridge media, and do so by reuploading files to the homeserver. This
+can cause a lot of storage usage. This MSC proposes a way for appservices to register media
+(`mxc://`) namespaces.
+
+## Proposal
+
+This proposal adds a new key (`media`) to the namespaces block of appservice registration.
+
+Whenever the homeserver gets a request that matches the regex, it should make a http GET request
+to `/_matrix/app/v1/media/{mediaId}`.
```
Kladki, line 12: Would the `timeout_ms` parameter from the client request be forwarded to the appservice?

Kladki, line 12: IS the server expected to always follow redirects, or should it use the `allow_redirect` param like the C-S API?

proposals/3896-appservice-media.md
```diff
@@ -0,0 +1,47 @@
...
+`/_matrix/app/v1/media/foobar-bazqux`.
+
+Appservices may set `Cache-Control` on their response. Homeservers should cache the response, though
+they may remove cached remote media to save space.
+
+## Potential issues
+
+Media may not be able to load if the appservice is unable to reach wherever remote media is stored.
+
+## Alternatives
```
Kladki, line 33: mautrix-discord sets up what is essentially a homeserver which only serves media, so perhaps that could be mentioned, and then explain how that is sub-optimal.

proposals/3896-appservice-media.md
Kladki: Doesn't really make sense in this context, since this is for async uploads, which doesn't apply for media appservices. Not sure why I even asked this (this can be marked as resolved).

PR #3897, opened 2022-09-25
Rip out SFU bits out of MSC3401

SimonBrandner:
[MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/3401) is already large and there are a lot of SFU bits that are missing in it and would need to be added therefore it makes more sense to rip out the SFU bits into a separate MSC (see https://github.com/matrix-org/matrix-spec-proposals/pull/3898)

### DIFF
--- a/proposals/3401-group-voip.md
+++ b/proposals/3401-group-voip.md
@@ -15,7 +15,11 @@ Meanwhile we have no native signalling for group calls at all, forcing you to in
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


dbkr (approved):
Yeah, this seem fine, but I think I'd agree with Matthew in that we could probably just refer to this MSC for all the diagrams, if possible

SimonBrandner:
I've now moved all the diagrams into MSC3401 and link to it from MSC3898. @ara4n, does that look good to you?

dbkr (approved):

PR #3899, opened 2022-09-27
Misc clarifications to threading MSCs

turt2live:
***This is reviewable commit-by-commit***

In this PR we cover some of the nuances discovered while writing the actual spec for threads.

### DIFF
--- a/proposals/3440-threading-via-relations.md
+++ b/proposals/3440-threading-via-relations.md
@@ -51,12 +51,14 @@ would include additional information in the `unsigned` field:
 }
 ```
 
-* `latest_event`: The most recent event which relates to this event, with
-  `rel_type` of `m.thread`.
+* `latest_event`: The most recent event (topologically) which relates to this event, with
+  `rel_type` of `m.thread`. Events sent by [ignored users](https://spec.matrix.org/v1.3/client-server-api/#ignoring-users)
+  are not considered for the latest event.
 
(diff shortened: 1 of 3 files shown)


proposals/3440-threading-via-relations.md
```diff
@@ -51,12 +51,14 @@ would include additional information in the `unsigned` field:
 }
 ```
 
-* `latest_event`: The most recent event which relates to this event, with
-  `rel_type` of `m.thread`.
+* `latest_event`: The most recent event (topologically) which relates to this event, with
+  `rel_type` of `m.thread`. Events sent by [ignored users](https://spec.matrix.org/v1.3/client-server-api/#ignoring-users)
+  are not considered for the latest event.
```
clokep, line 56: We discussed this over Matrix, it makes sense to make this explicit here, but it really is just following the guidance of [MSC2675](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2675-aggregations-server.md#how-do-you-handle-ignored-users):

> Information about relations sent from ignored users must never be sent to the client, either in aggregations or discrete relation events. This is to let you block someone from harassing you with emoji reactions (or using edits as a side-channel to harass you). Therefore, it is possible that different users will see different aggregations (a different last edit, or a different reaction count) on an event.

clokep (approved):
Thanks! 👍

erikjohnston (approved):

PR #2444, opened 2020-02-24
MSC2444: peeking over federation via /peek

ara4n:
A new MSC for proposing how peeking over federation could work, hopefully replacing #1777  
This one introduces the idea of a `/peek` API in S2S which lets a peeking server subscribe to events from given server(s) so it can have a read-only peek view of peekable rooms.  This is based on the review of #1777 in Oct 2019.

[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/msc2444/proposals/2444-peeking-over-federation-peek-api.md)

Obsoletes #1777

PR #3901, opened 2022-09-28
MSC3900: Deleting state

andybalaam:


proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,329 @@
...
+Currently, state events are not encrypted, but
+[MSC3414](https://github.com/matrix-org/matrix-spec-proposals/pull/3414)
+proposes allowing them to be encrypted.
+
+If MSC3414 goes ahead, a deleted encrypted state event should contain
+`deleted: true` in its unencrypted content, as a sibling of e.g. `algorithm`
+and `ciphertext`.
+
+When the ciphertext is decrypted, the `content` in the plaintext JSON
+should also contain `deleted: true`.
```
MadLittleMods, line 182: Which one is is the source of truth if they differ?

I assume the the more top-level `deleted` outside of the encrypted contents is the source of truth. A server will only be able to interact with the unencrypted one but a client might also display something based on it.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,329 @@
...
+Based on our definition of "deleted" state, when sending room state to clients
+for an initial sync, do not include deleted state.
+
+TODO: specific spec wording change
+
+### Potential issues
+
+If clients actually need deleted state to render properly, this would imply
+that events have been marked as deleted when they should not have been. (Note:
+we are discussing room state here, not state events. Deleted state events
```
MadLittleMods, line 233: Might be more helpful to disambiguate? Not sure if "current room state" is coined anywhere else

```suggestion
we are discussing current room state here, not state events. Deleted state events
```

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,383 @@
...
+
+Currently, when a room is upgraded, the new room is only somewhat similar to
+the old one.
+
+We propose to expand the definition of a room upgrade to copy all useful
+information from the old to the new room.
+
+This involves copying all non-obsolete, non-user-scoped room state by creating
+state events in the upgraded room.
+
```
reivilibre, line 334: We may consider allowing the copying of user-scoped room state by the room founder in a restricted situation:
as an example: the imported state events could be permitted between the `m.room.create` and the `m.room.member` for the founder — needs reasoning through to ensure you can't fork off it later etc but just to paint a picture of what I have in mind.
The state events would be sent by the room founder but would be able to write into user-scoped state.
Some way of ensuring that the room founder doesn't 'fork' off this branch later on, as well as to ensure the founder can't somehow retain this right forever, would likely be needed, but it's a suggestion nonetheless.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+To mark some state as obsolete, a client sends a state event with
+`m.obsolete: true` in its content. To "unobsolete" some state later, the client
+sends another state event with no `m.obsolete` property (or with
+`m.obsolete: false`).
+
+### Redacted state events are obsolete
+
+We propose to update the definition of event redaction[^spec-redactions] to
+specify that all redacted state events contain `m.obsolete: true` in their
+content.
```
kegsay, line 174: In practice, does this mean that `content:{}` == obsolete?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+Some state is already, loosely speaking, "obsolete" in the sense that new
+members don't really care about it. For example, leaving events.
+
+It might be possible to define obsolete state as including these special cases,
+and this might allow us to avoid needing a new room version.
+
+However, we believe that we need to change the rules around redacted events,
+meaning that we can't avoid a new room version. Since we need a new room version
+anyway, we have gone for a simpler definition of obsolete state with no special
+cases.
```
kegsay, line 234: Extra boilerplate and bandwidth though having `membership: leave` always coupled with `m.obsolete: true`.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+However, this seems a relatively easy proposal, and we hope that implementing
+it will help us understand what we really mean by "obsolete" state, and flush
+out problems we have not yet considered.
+
+### Security considerations
+
+If security-critical events were not sent to clients, this could cause security
+problems, but since only events that are irrelevant to clients should be marked
+as obsolete, this should not happen.
```
kegsay, line 306: The problem is that the _sender_ determines this, not the _recipient_. What's stopping a `m.obsolete` appearing in the power levels in practice? We need more safeguards.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+The server should send invitations on behalf of the user performing the upgrade.
+These invitations should contain a `part_of` property in their content, whose
+value is the ID of the `m.room.create` event of the new room. (This makes a
+later step, automatically accepting these invitations, possible - see
+sub-proposal 3).
+
+This action should only be performed by the server if the user performing the
+upgrade is registered with this server. Otherwise, the server should not create
+any invitations or emit any membership events.
```
kegsay, line 348: ??? If the user isn't "registered" (read: access token?) with this server then it'll be 403d? Feels weird to mention this, as technically this applies to every CS API endpoint.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+### Proposal
+
+Relevant spec section:
+[11.33.3 Room upgrades - Server Behaviour](https://spec.matrix.org/v1.5/client-server-api/#server-behaviour-16).
+
+When a client requests to upgrade a room using `POST /rooms/{roomId}/upgrade`,
+this should be interpreted by the server as a request not only to create the
+room, but also to invite all members of the old room to the new one, with the
+same power level.
```
kegsay, line 338: This should only apply if the upgraded room version == the existing room version. By allowing actual room version bumps we risk breaking entire servers who aren't on the latest version of synapse.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+If the user upgrading the room is registered with this homeserver, create
+invitation events on behalf of the upgrading user for every user who is
+currently a member of the old room, inviting them to the new room. Also set the
+room power levels to give the same power level to each user that they had in the
+old room.
+
+Only members who are currently members of the room should be invited to the new
+ones.
+
+`m.room.membership` events should also be created for users who are banned from
```
kegsay, line 384:
```suggestion
`m.room.member` events should also be created for users who are banned from
```

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+This operation causes a homeserver to send out lots of invitations, which could
+be a cause of invite spam. It can only be caused by someone who is an admin of a
+room already containing the recipients, so that limits the scope.
+
+### Dependencies
+
+In order to exclude obsolete bans, the definition of obsolete from this proposal
+is required, but the main part of this sub-proposal does not depend on any
+others.
```
kegsay, line 426: I'm confused what an obsolete ban is _in practice_. Sure technically it is `membership: ban` with `m.obsolete: true` but when would you actually _do_ this? Wouldn't you just unban them if the ban is obsolete i.e `membership: leave`?

kegsay (commented):
Overall seems pretty thorough and an accurate assessment of the 2022 discussions.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+### Potential issues
+
+If clients actually need obsolete state to render properly, this would imply
+that events have been marked as obsolete when they should not have been. (Note:
+we are discussing room state here, not state events. Obsolete state events
+should be returned as normal when the events timeline is requested. This allows
+users to explore historical events.)
+
+The only time when an obsolete state event is needed to update room state is
+when a client has already received non-obsolete state for this `event_key`.
```
dkasak, line 289: `state_key`?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+be a cause of invite spam. It can only be caused by someone who is an admin of a
+room already containing the recipients, so that limits the scope.
+
+### Dependencies
+
+In order to exclude obsolete bans, the definition of obsolete from this proposal
+is required, but the main part of this sub-proposal does not depend on any
+others.
+
+## Sub-proposal 3: Auto-accept invitations to upgraded rooms
```
dkasak, line 428: What's stopping a HS admin from marking `m.room.encryption` as obsolete and then triggering an upgrade, semi-silently turning off encryption for the room?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+and create state events in the new room with the same `state_key` and
+`contents`, but with `sender` set to the mxid of the user performing the upgrade.
+
+The server should copy all state except:
+
+* Obsolete state, as defined earlier in this proposal
+* User-scoped state i.e. any state whose `state_key` is equal to the sender's
+  mxid. (If MSC3779 "Owned state events" is merged, user-scoped state will also
+  include anything with a `state_key` that starts with the user's mxid plus
+  underscore.
```
dkasak, line 508: Might be worth spelling out the logic on why these are excluded.

dkasak: Reading further: probably because the upgrading user resends all of the state events, so they wouldn't be able to write user-scoped state on behalf of another user?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+  relationships
+- push rules refer to room IDs
+- sync filters are based on room IDs
+- 3PID invitations include room ID
+
+We propose to make upgraded rooms keep the same room ID as the old version, by
+introducing a server-only sub-ID that represents the version of the room.
+
+Clients and external systems continue to use the existing room ID, and servers
+use room ID + room version to identify the real actual room.
```
dkasak, line 607: How does traversing history back into the old room work with this scheme? A lot of detail here is left out from the current proposal.

proposals/3901-deleting-state.md
kegsay: Yeah

anoadragon453: @andybalaam and I had a conversation regarding this, and figured that allowing room admins to send state events on behalf of other users (possibly via a new power level, or piggybacking off of `m.room.tombstone`) could solve this problem. There are special cases already to allow room admins to send user-scoped state events (inviting, kicking, banning a user).

proposals/3901-deleting-state.md
anoadragon453: You can currently do that today, by manually sending a `m.room.tombstone` event leading to an unencrypted room.

You can also remove encryption in a room by setting the `content` field of an existing `m.room.encryption` state event to `{}`, effectively deleting it. Clients are supposed to ignore this though - so perhaps they can be told to do the same for `m.room.encryption` events marked as obsolete?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+### Alternatives
+
+[MSC3325](https://github.com/matrix-org/matrix-spec-proposals/pull/3325)
+proposes that all users in the old room be _allowed_ to join the new room by
+using a `restricted` join rule.
+
+MSC3325 also mentions as an alternative that the room membership of each user
+could be set as `invited` without actually sending an invitation, to avoid
+invite spam.
```
anoadragon453, line 414: Another question is what should happen if more than one `m.room.tombstone` event is sent in a room. You would be able to invite spam a single user more than once.

I'm leaning towards extending the definition of the `invite` `join_rule` to mean "users joined to the predecessor room may join without an explicit invite". Essentially "restricted", but only for the `predecessor` room. That way you cut down on both bandwidth of sending invites over federation, as well as invite state events in the room.

The fact that a user has joined via this method should be known to other clients in the room, such that they can share historical message keys for the purposes of decrypting old room messages.

proposals/3901-deleting-state.md
andybalaam: Yes, that makes sense. Clarified in f82ea227.

It does lead me to ask whether duplicating this is helping. I only added it as a convenience, but given it could introduce ambiguity maybe it should not appear?

proposals/3901-deleting-state.md
andybalaam: Thanks. Done in 4728ceb8

proposals/3901-deleting-state.md
andybalaam: This would allow e.g. upgrading a room while someone was sharing their location, and having that state transferred across, so they are marked as sharing their location in the new room (I'm guessing the client would need to do some magic to notice and actually share there...)

If we want upgrades to be truly seamless this would help, but it also feels a little dangerous. It might also mean we copied state that we actually wanted to ignore, but this is hard to predict - it depends how user-scoped state is used in future.

I will add some words about this under alternatives and we can continue discussing whether we want to include it.

andybalaam: Added in 0dc4a782. I am thinking this should probably be a later refinement, because it sounds tricky.

proposals/3901-deleting-state.md
andybalaam: What do you mean by `==`? It does mean that all redacted events will have `content:{}` AND `obsolete:true` but I don't know whether there are other circumstances where `content:{}` is valid, especially in events defined outside the spec.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+
+### Leave events are obsolete
+
+We propose to update the definition of [membership
+events](https://spec.matrix.org/v1.5/client-server-api/#mroommember) so that
+every event with `membership: "leave"` must also have `m.obsolete: true` in its
+content.
+
+Note: `membership: "ban"` events are not considered obsolete since this
+information is needed in future to prevent bad actors from re-entering a room.
```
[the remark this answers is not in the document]
andybalaam: Added a note in f0eef7a8, thanks. I don't know exactly what an invite rejection looks like so maybe you could suggest wording to make it more explicit?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+Currently, state events are not encrypted, but
+[MSC3414](https://github.com/matrix-org/matrix-spec-proposals/pull/3414)
+proposes allowing them to be encrypted.
+
+If MSC3414 goes ahead, an obsolete encrypted state event should contain
+`m.obsolete: true` in its unencrypted content, as a sibling of e.g. `algorithm`
+and `ciphertext`.
+
+When the ciphertext is decrypted, the `content` in the plaintext JSON
+should also contain `m.obsolete: true`.
```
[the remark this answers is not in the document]
andybalaam: Yeah, see my replies to https://github.com/matrix-org/matrix-spec-proposals/pull/3901#discussion_r1019569373 . Can we continue the conversation there?

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+However, some existing obsolete state events such as leaving events (membership
+events indicating that someone left the room) contain useful content, and there
+is no reason to assume that future ones won't also want to do something similar.
+
+#### m.obsolete as a sibling of content
+
+We could say that the `m.obsolete` property is not inside `content`, but
+alongside it.
+
+This might make it easier for servers to find and index obsolete state.
```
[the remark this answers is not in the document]
andybalaam: Cool. Want me to update the MSC or just note this as a comment here?

proposals/3901-deleting-state.md
andybalaam: Yes, but I feel like it's worth it to avoid any chance of confusion. I assume the amount of bandwidth we use on `membership: leave` events is small compared with the other events in most rooms.

Noted in 96350bd6.

proposals/3901-deleting-state.md
florianjacob: An alternative would be to actually have some consistency checks of other servers, i.e., accept a new room as valid successor only if the state copied over is like it should have been, e.g., also copying m.room.encryption and more.

florianjacob:
I just had a chat with @anoadragon453 after the very helpful [Solving the Historical State Problem in Matrix
](https://cfp.gulas.ch/gpn21/talk/9Y3NFU/) talk on the topic - the [recording is already available](https://media.ccc.de/v/gpn21-202-solving-the-historical-state-problem-in-matrix) and presents the ideas of this MSC in a more current fashion than what is written down.

The fundamental issue here with ideas going for actually deleting things might be that garbage collection, from a computer science theory perspective, fundamentally requires coordination among servers, even consensus, which Matrix cannot and does not want to provide in its current form. Everything is fine with hiding obsolete events from clients - that is just a projection of the data still present on the server, but as soon as you actually want to delete them, you will probably run into problems that require hard drawbacks like requiring that servers cannot be offline for an arbitrary length of time.

Room upgrades potentially circumvent the fundamental issue because just starting a new room does not *actually* garbage-collect the old state, strictly-speaking. However, having a monotonically-increasing integer as sub-version to the room identifier would also require coordination / consensus, or requires hope that concurrent room upgrades do not happen in practice and network partitions are short enough, or actually ignoring everything that was written in the concurrent room instance that "lost". One way to circumvent the issue with monotonically-increasing sub-versions would be to switch to a partial order of sub-versions, i.e., a DAG, which would then map concurrent room upgrades to branches in the DAG. However, one would want have something like a "current" room version - which could be done by having a third room upgrade that merges the branched room upgrades again.

Or, the room sub-version DAG could be topologically sorted, like with state resolution - and that is where we were thinking, what about “room upgrades as rooms”? Have a space as a meta-room that contains pointers to the sub-rooms, and a room upgrade is pointer in the space to a new room? Then the sub-version DAG would exactly be the regular space DAG, and topological sorting would be the regular room topological sorting. Also, it would keep room ids as-is, and servers do not need to follow all room ugprades to end up with the current room version. However, this idea also has problems, like keeping the state in the meta-room minimal, i.e., not requiring normal users to join there to be able to take part in the sub-rooms - potentially requires #2444. On the other hand, it would also massively profit from other improvements around spaces, like improved access control / power level inheritance from space to sub-rooms.

proposals/3901-deleting-state.md
andybalaam: What is the attack vector you are concerned about? Different people interpreting the presence of `"m.obsolete": true` differently from each other and therefore having different views of the world? Did you have a specific case in mind?

proposals/3901-deleting-state.md
andybalaam: Thanks, removed in 203e7b23

proposals/3901-deleting-state.md
andybalaam: Why would we limit it to room "refactorings" (new version==old version)?

How will it break homeservers that don't support this? Won't they just see fairly normal-looking invitations, and then they themselves fail to do the part they are supposed to do? Isn't that "fine"?

proposals/3901-deleting-state.md
andybalaam: Yeah, I guess `membership: leave` (which would also include `obsolete: true` of course) would be sufficient to forget a ban. If there were a huge spam attack and state were full of bans that we believed were no longer needed, we could forget them by just unbanning and running a "refactor" (upgrade with old version==new version).

andybalaam: Fixed in 05e8e3a1

proposals/3901-deleting-state.md
andybalaam: Yeah, this part is quite forward-looking and not fully-defined.

proposals/3901-deleting-state.md
andybalaam: This does sound good to me. Any objections from anyone? Does anyone want to write it up in the MSC?

andybalaam:
> I just had a chat with @anoadragon453

Wow, really interesting stuff. I think this mostly makes me feel that the room sub-versions stuff is way harder than the rest, and should probably be separated out from it to allow the easier stuff to progress...

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,652 @@
...
+tombstone event refers to the new room as successor, and that the user was a
+member of the old room at the time the tombstone was created. If all these are
+true, the server should auto-join the user to the new room by emitting an
+`m.room.member` event on their behalf whose properties match their membership of
+the old room (excluding `join_authorised_via_users_server`, which should be
+omitted since the user is invited, so does not need additional authorisation).
+
+### Potential issues
+
+### Alternatives
```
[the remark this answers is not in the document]
andybalaam: Mentioned in 158146ed

proposals/3901-deleting-state.md
andybalaam: Validating is difficult though, because you need to identify a "time" at which you validate, and if you do that it's hard to guard against someone changing the state "after" that "time", and even harder to avoid state resets that bypass the "time" you chose...

andybalaam: Given what @anoadragon453 said, I think this is not too much of a problem, so I'm going to resolve this thread, but if someone thinks we should write this up as a potential issue (and/or fix it) I'd gladly accept proposed wording.

proposals/3901-deleting-state.md
```diff
@@ -0,0 +1,686 @@
...
+content.
+
+[^spec-redactions] https://spec.matrix.org/v1.4/rooms/v10/#redactions
+
+### Leave events are obsolete
+
+We propose to update the definition of [membership
+events](https://spec.matrix.org/v1.5/client-server-api/#mroommember) so that
+every event with `membership: "leave"` must also have `m.obsolete: true` in its
+content.
```
robintown, line 188: Interestingly, several folks have commented today [in a TWIM discussion](https://matrix.to/#/!QQpfJfZvqxbCfeDgCj:matrix.org/$_2uNRV6BxzXvDXsYdtBPj9jjpTelD-Ua5ODMVJjguGY?via=matrix.org&via=envs.net&via=element.io) that forgetting whether a member has left vs. whether they've simply never joined, could be inconvenient for moderation.

robintown (commented):
For visibility, I had some thoughts that are relevant to state deletion and wrote them up recently on my blog: https://robin.town/blog/distributed-access-control

Specifically, I had the thought that if Matrix room state, as a data structure, used version vectors instead of tombstones as in https://inria.hal.science/hal-00738680, then it seems that implementations could forget deleted state immediately, without the need for coordinated room upgrades.

PR #2391, opened 2019-12-19
MSC2391: Federation point-queries.

tulir:
Opening PR on behalf of @jevolk

[Rendered](https://github.com/matrix-construct/matrix-refreshed/blob/jevolk/fed_state__point_queries/proposals/2391-federation-point-query.md)

PR #3902, opened 2022-10-03
MSC3902: Faster remote room joins over federation (overview)

richvdh:


proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+      once resync completes, we send them out to any homeservers that were in
+      the room at any point since we started joining. ([Synapse
+      implementation](https://github.com/matrix-org/synapse/pull/13934))
+
+    * Fixes to incoming device list updates: normally we ignore device-list
+      updates from users who we don't think we share a room with. To ensure we
+      do not discard incoming device list updates, we keep a record of any
+      *remote* device list updates we receive, and replay them once resync
+      completes. ([Synapse
+      implementation](https://github.com/matrix-org/synapse/pull/13913)
```
H-Shay, line 125: Looks like there is a closing para theses missing here?

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+Once a server receives a "partial state" response to `/send_join`, it must then
+call [`/state/{room_id}`](https://spec.matrix.org/v1.4/server-server-api/#get_matrixfederationv1stateroomid),
+setting `event_id` to the ID of the join event returned by `/send_join`, to
+obtain a full snapshot of the state at that event. It can then update its database 
+accordingly.
+
+However, this process may take some time, and it is likely that other events
+have arrived in the meantime. These new events will also have been stored with
+"partial state", and will not have been subject to the full event authorisation
+process. The server must therefore work forward through the event DAG,
```
H-Shay, line 138: Q: is this the background job that is mentioned above?

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+recalculating the state at each event, and rechecking event authorisation,
+until it has caught up with "real time" and new events are being created with
+"full state".
+
+## Potential issues
+
+TBD
+
+## Alternatives
```
timokoesters, line 147: Alternatives: https://github.com/matrix-org/matrix-spec-proposals/pull/3883, or "interactive peeking" 

Less complicated state res => easier to verify correctness of spec and implementation

proposals/3902-faster-remote-joins.md
richvdh: indeed, thanks!

```suggestion
      implementation](https://github.com/matrix-org/synapse/pull/13913))
```

proposals/3902-faster-remote-joins.md
richvdh: It is, yes. Do you have any thoughts on if this needs clarifying?

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+   can reliably be answered (since we assume we have all non-membership state
+   in the room), whereas a request for the [list of joined
+   members](https://spec.matrix.org/v1.4/client-server-api/#get_matrixclientv3roomsroomidjoined_members) cannot be answered.
+
+   In the current implementation, requests that require knowledge of
+   `m.room.member` events for remote users will *block* until the
+   resynchronisation completes.
+
+   (Note that we can reliably answer requests that require knowledge only of
+   the membership state for local users.)
```
DMRobertson, line 81: Is this correct? Suppose `@alice:alice.com` partially joins a room via `bob.com`. Before the resync completes, she is banned from that room by `@chris:chris.com`. Call the ban event B.

Is there some way that B can be an invalid event whilst appearing invalid to `alice.com`? If so, that means we can't reliably answer requests about local users' membership(?)

richvdh: yes, that can certainly happen. "Reliably" here is doing a bit of heavy lifting: we actually don't know *any* of the state for *certain* during the resync (events may be accepted that should not have been, and they can also be rejected when they should not have been. I seem to remember writing complement tests for both cases). We may therefore tell clients about state that turns out to be wrong. In some ways this is an extension of https://github.com/matrix-org/matrix-spec/issues/1209.

But still, membership events for local users are in the same bracket as, for example, `m.room.topic` events: we assume that our view of the state is good enough, and return the events to clients without waiting for the resync to complete.

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+   server may or may not be able to accurately answer it. For example, a
+   request for the topic of the room via
+   [`/rooms/{roomId}/state/m.room.topic`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixclientv3roomsroomidstateeventtypestatekey)
+   can reliably be answered (since we assume we have all non-membership state
+   in the room), whereas a request for the [list of joined
+   members](https://spec.matrix.org/v1.4/client-server-api/#get_matrixclientv3roomsroomidjoined_members) cannot be answered.
+
+   In the current implementation, requests that require knowledge of
+   `m.room.member` events for remote users will *block* until the
+   resynchronisation completes.
```
timokoesters, line 78: An alternative would be to send the currently known members and send updates to this list over /sync's `state`. Then it also makes sense to include partial rooms in non-lazyloaded syncs. What do you think about that? I think that's better than blocking, especially when we don't know when or if the room ever fully loads.

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+     the (full) state before the event, we can no longer apply this
+     check. Instead, we perform a state-resolution between the limited state
+     that we do have, and the event's auth events; we then check that the
+     incoming event passes the authorization rules based on that resolved
+     state.
+
+     This process means that we are largely trusting remote servers not to send
+     invalid events (hence the need for a revalidation during the
+     resynchronisation process); however it does mean that if we have a ban for
+     a particular user, then their events will be rejected.
```
timokoesters, line 56: > if we have a ban for
     a particular user, then their events will be rejected.
     
 This seems to be the goal of soft failing and not the "state at event" check. Before fast remote joins, these banned user's events would still be accepted as outliers. To get the same behavior, we could remove the "state at event" auth check (unless there's a good reason to keep it?).

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+
+ * Processing incoming events received over federation:
+
+   * Currently, the
+     [spec](https://spec.matrix.org/v1.4/server-server-api/#checks-performed-on-receipt-of-a-pdu)
+     requires that an incoming event "Passes authorization rules based on the
+     state before the event, otherwise it is rejected". Since we do not know
+     the (full) state before the event, we can no longer apply this
+     check. Instead, we perform a state-resolution between the limited state
+     that we do have, and the event's auth events; we then check that the
```
timokoesters, line 49: This state resolution will be extremely expensive because nearly all events are conflicting

proposals/3902-faster-remote-joins.md
squahtx: In the current implementation in Synapse, non-lazy_load_members syncs don't block but instead pretend the room has not been joined until the homeserver has the full state of the room available. Other endpoints, like /members, still block. 

I'm not sure whether it's spec-compliant to initially send the known members over /sync (we can propose a change to the spec if it isn't) or how clients would interpret it. It might be that all clients are okay with it. We just haven't tested it.

proposals/3902-faster-remote-joins.md
squahtx: The case we had in mind was where the "true" state of the room after event P has a ban for user `@bad:some-homeserver.com` and `@bad:some-homeserver.com` sends an event E, with {P} as its prev_events and citing its membership prior to the ban as an auth event. The faster joining homeserver may or may not have the ban for that user in its idea of the state after P. If it does, we want event E to be _correctly_ rejected rather than trusting its auth events. To reject E, we have to run the "Passes authorization rules based on the (partial) state before the event" check. But the check would also reject most normal messages because their sender's memberships won't be in the partial state, so we combine the partial state with the auth events using state resolution.

In the soft fail case, where `@bad:some-homeserver.com` is banned in the current state of the room, but picks very old prev_events to send events with, the accept/reject behavior is unchanged.

proposals/3902-faster-remote-joins.md
squahtx: I'm not very familiar with state resolution, but if we take the expected case where the auth events match those in the partial state, we'd be calculating `state_res( partial_state={ create, power levels, name, topic, ... }, auth_events={ create, power levels, sender membership })`.

You're right that the conflicted state set would include most state events in the room: `{name, topic, ..., sender membership}`
but wouldn't this set be small, since most memberships aren't in the room's partial state? I'd also expect the auth difference to be small because both sets of state contain the same create and power levels events in the common case.

squahtx: I can't remember the rationale for doing state res here. I think a way to combine the partial state with the auth events was needed and state resolution was the usual way we combine state. Something simpler, like filling in the gaps in partial state with the auth events, might be appropriate too, but I may be overlooking some considerations.

richvdh: > I can't remember the rationale for doing state res here. I think a way to combine the partial state with the auth events was needed and state resolution was the usual way we combine state.

Yes, basically this. It was an attempt to enforce bans, as far as possible.

timokoesters: > since most memberships aren't in the room's partial state

Wouldn't this fill up with time and when the room is almost fully synced include a lot of events?

proposals/3902-faster-remote-joins.md
```diff
@@ -0,0 +1,171 @@
...
+     "include membership events for the `sender` of events being returned in
+     the response". Since we do not have the full state of the room, we may be
+     missing membership events for some senders. We resolve this by checking
+     the `auth_events` for affected events, which must include a reference to a
+     membership event.
+     
+   * If lazy-loading is *not* enabled, partial-state rooms are omitted from the
+     response (until the state synchronisation completes).
+
+     (This is [pending implementation](https://github.com/matrix-org/synapse/issues/12989) in Synapse.)
```
DMRobertson, line 98: I think this line is redundant, now that the Synapse [issue](https://github.com/matrix-org/synapse/issues/12989) is closed (by [this PR](https://github.com/matrix-org/synapse/pull/14870)).

```suggestion
```

proposals/3902-faster-remote-joins.md
Fizzadar: Could we "solve" this by adding a new federation API to get latest state by type/state_key (`_matrix/federation/v1/state/<type>/<state_key>` or similar)?

If we assume we trust the server used to join the room via we can then use that same server to fetch any current `m.room.member` state as needed when events are received. Wouldn't this allow a server to auth events completely without being fully synchronised?

richvdh:
Closing this because I'm tired of having an open PR on my account. The info is good, but I don't think it's a good fit for the MSC format, since it doesn't really propose changes to the spec. It probably wants to be turned into an implementation guide on https://matrix.org/docs or something.

proposals/3902-faster-remote-joins.md
jevolk: > Could we "solve" this by adding a new federation API to get latest state by type/state_key (`_matrix/federation/v1/state/<type>/<state_key>` or similar)?

Yes: https://github.com/matrix-org/matrix-spec-proposals/pull/2391

PR #3905, opened 2022-10-12
MSC3905: Application services should only be interested in local users

MadLittleMods:
Application services should only be interested in local users

Fix https://github.com/matrix-org/matrix-spec/issues/1272

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/madlittlemods/appservice-interested-in-local-users/proposals/3905-appservice-only-interested-in-local-users.md)

Homeserver implementations:

 - Synapse: https://github.com/matrix-org/synapse/pull/13958

----

[FCP Comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3905#issuecomment-1276532045)

### DIFF
--- a/proposals/3905-appservice-only-interested-in-local-users.md
+++ b/proposals/3905-appservice-only-interested-in-local-users.md
@@ -22,8 +22,6 @@ receive events in rooms that are shared with a local user. This leaves us with a
 behavior mismatch between remote and local users.
 
 
-
-
 ## Proposal
 
 Therefore the proposal is that the `users` namespace regex should only be applied
@@ -70,6 +68,24 @@ considerations introduced.
 
 With this MSC, an application service will be receiving less events than before.
 
+
+## Historical context
+
+According to @turt2live (SCT member), "the spec intended to [originally] say the
+namespace can make an appservice interested in remote users, though there's obviously no
+ability for the server to call `/user` on remote users (it's not like the appservice can
+create them)." (https://github.com/matrix-org/synapse/pull/13958#discussion_r988369446)
+
+This intention goes back further than `r0` (or `v1.0` in marketing versions speak) but
+this history is lost to time since there isn't really anything concrete to point to
+beyond the original spec
+[issue](https://github.com/matrix-org/matrix-spec-proposals/issues/1307) and
+[PR](https://github.com/matrix-org/matrix-spec-proposals/pull/1533) which don't mention
+these details.
+
+Since we're unable to come up with any valid use cases nowadays, it's unclear to
+outsiders from that time whether the original intention is actually true. In any case,
+we're clarifying it here and making an MSC to change it explicitly.
 
 
 ## Unstable prefix
--- a/proposals/3905-appservice-only-interested-in-local-users.md
+++ b/proposals/3905-appservice-only-interested-in-local-users.md
@@ -41,7 +41,7 @@ const interestingUsers = localRoomMembers.filter((localRoomMember) => localRoomM
 
 ---
 
-To avoid confusion, please note that the `rooms` and `aliases` namesapces are not
+To avoid confusion, please note that the `rooms` and `aliases` namespaces are not
 affected. You can still match whatever rooms and aliases to listen to all events
 that occur in them.
 
@@ -49,14 +49,14 @@ that occur in them.
 ## Potential issues
 
 There are use cases like moderation where an application service wants to hear all
-messages from remote users in rooms but these are are also covered by the `rooms`
+messages from remote users in rooms but these are also covered by the `rooms`
 namespace where all events in a matched room are considered "interesting".
 
 
 
 ## Alternatives
 
-The alternative is clarify that the `users` namespace should be matched against all
+The alternative is to clarify that the `users` namespace should be matched against all
 users (local and remote). This still leaves us with the behavior-difference footgun.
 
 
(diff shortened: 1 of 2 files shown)


proposals/3905-appservice-only-interested-in-local-users.md
```diff
@@ -0,0 +1,63 @@
...
+With this MSC, an application service will be receiving less events than before.
+
+
+
+## Unstable prefix
+
+If a homeserver wants to implement this functionality before this MSC merges, since it
+only affects application services local to the server, it can be implemented although it
+will be technically unspecced behavior until this MSC is merged. Also consider a config
+option for maximal compatibility with existing application services people may be using.
```
MadLittleMods, line 63: Sounds a bit cheeky but seems fine, right?

proposals/3905-appservice-only-interested-in-local-users.md
```diff
@@ -0,0 +1,63 @@
...
+receive events in rooms that are shared with a local user. This leaves us with a
+behavior mismatch between remote and local users.
+
+
+
+
+## Proposal
+
+Therefore the proposal is that the `users` namespace regex should only be applied
+against local users of the homeserver.
```
turt2live, line 30: We should probably mention that `rooms` and `aliases` aren't touched, and allow a `.*:.*` registration themselves. This is primarily to support the feature that most appservices with a `.*` namespace are interested in: listening to all rooms/users on the homeserver. It's extraordinarily rare that an appservice would want to listen to all Telegram users (for example), but it often wants to listen to local Telegram users or literally all events (in which case it'd be traditional for it to ask for a `.*` users registration, however the same is achieved with a `.*` rooms namespace).

turt2live, line 30: Explicit clarity on what servers should do with something like `@_telegram.+:.*` would be appreciated, for the legacy registration files out there. In practice, I believe the naive solution would be to do something like `sender.matches(regex) && sender.endsWith(":" + homeserver.domain)`

proposals/3905-appservice-only-interested-in-local-users.md
```diff
@@ -0,0 +1,63 @@
+# MSC3905: Application Service should only be interested in local users
```
turt2live, line 1: Somewhere in the MSC (the intro?) it'd be good to get the historical context captured: the spec has always meant to say that the regex is applied to all users it has visibility on (meaning one needs to specify something like `@_telegram.+:mydomain.org`), but has clearly done a bad job of documenting that. This goes back further than r0 (or v1.0 in marketing versions speak), which is why there's an MSC and not a clarification PR.

proposals/3905-appservice-only-interested-in-local-users.md
turt2live: Unstable behaviour of appservices is a bit annoying: we'd have to specify a flag in the registration file (since the registration file is strongly specified, which is irritating in its own right).

This description matches what we've done in the past (from memory):
> While this MSC is not considered stable, appservices can opt-in to this behaviour in either of two ways:
> 1. Specifying a regex which includes the local homeserver's domain in the regex (eg: `@_irc.*:example.org`)
> 2. Adding `"org.matrix.msc3905": true` to the root level of their registration file

One day we'll make appservices not so awful to do unstable behaviour on :(

turt2live (commented):
generally this lgtm - just a few clarifications and I think we can send this into FCP.

proposals/3905-appservice-only-interested-in-local-users.md
MadLittleMods: I don't think the regex needs to be changed at all. We just need to check that `sender` is a local user before applying the regex.

(Synapse uses `is_mine_id` to detect local user)
```js
is_mine_id(sender) && sender.matches(regex)
```

I can add this to the MSC 👍

proposals/3905-appservice-only-interested-in-local-users.md
MadLittleMods: I can document that this is was the intention from your memory. Just feels like it's based on nothing since there is nothing concrete to point to.

MadLittleMods: Added a historical context section 👍

proposals/3905-appservice-only-interested-in-local-users.md
MadLittleMods: Really appreciate the great wording here! Updated ✅

I think we could be even more loose and fast with this particular one but this sounds like the proper way 👍

turt2live (approved):

turt2live:
Let's get some opinions quickly, and push this through without undue delay :)

@mscbot fcp merge

anoadragon453 (approved):
cc @Half-Shot, who is likely to also be interested in this.

proposals/3905-appservice-only-interested-in-local-users.md
```diff
@@ -0,0 +1,98 @@
+# MSC3905: Application services should only be interested in local users
+
+Application services receive events that they are "interested" in.
+
+The current language in the spec describes it like this:
```
richvdh, line 5:
```suggestion
The [current language in the spec](https://spec.matrix.org/v1.4/application-service-api/#registration)
describes it like this:
```

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1305

PR #3906, opened 2022-10-13
MSC3906: Mechanism to allow sign in and E2EE set up via QR code

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/sign-in-with-qr/proposals/3906-sign-in-with-qr.md)

Implementations:

- [v1](https://github.com/matrix-org/matrix-spec-proposals/blob/87b0d7581b828b8cc99fe3760cd71728fe57f5a1/proposals/3906-sign-in-with-qr.md) of proposal
	- [x] Element Web: https://github.com/matrix-org/matrix-react-sdk/pull/9403 and https://github.com/matrix-org/matrix-js-sdk/pull/2747
	- [x] Element iOS: multiple PRs, last: https://github.com/vector-im/element-ios/pull/6866
	- [x] Element Android: https://github.com/vector-im/element-android/pull/7369
- [v2](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/sign-in-with-qr/proposals/3906-sign-in-with-qr.md) `org.matrix.msc3906.setup.additional_device.v2` flow implementations:
    - [ ] Element Web: https://github.com/matrix-org/matrix-js-sdk/pull/3193
    - [x] Element iOS: https://github.com/vector-im/element-ios/pull/7407
    - [ ] Element Android: https://github.com/vector-im/element-android/pull/8212

Related MSCs for convenience:
- MSC3882 to obtain a `m.login.token`
- MSC3886 to provide a open non-trusted rendezvous endpoint
- MSC3903 to secure the rendezvous channel

### DIFF
--- a/proposals/3906-sign-in-with-qr.md
+++ b/proposals/3906-sign-in-with-qr.md
@@ -21,7 +21,7 @@ By way of example this is what the QR for an X25519 based rendezvous via HTTP lo
 ```json
 {
     "rendezvous": {
-        "algorithm": "m.rendezvous.v1.curve25519-aes-sha256",
+        "algorithm": "m.rendezvous.v2.curve25519-aes-sha256",
         "key": "2IZoarIZe3gOMAqdSiFHSAcA15KfOasxueUUNwJI7Ws=",
         "transport": {
             "type": "http.v1",
--- a/proposals/3906-sign-in-with-qr.md
+++ b/proposals/3906-sign-in-with-qr.md
@@ -113,7 +113,7 @@ Note over A,B: Devices establish secure rendezvous and derive<br> confirmation c
                 B->>B: 16. Wait for new device to come online
             end
 
-            Note over B: 17. check that device key from Homeserver vs received in 15 match
+            Note over B: 17. check that device key from homeserver vs received in 15 match
             B->>B: 18a. Mark new device (AAABBBCCC) as verified locally
 
             alt Cross-signing active
@@ -124,22 +124,21 @@ Note over A,B: Devices establish secure rendezvous and derive<br> confirmation c
                 B->>A: 19b. { "type": "m.login.finish", "outcome": "verified", <br>"verifying_device_id": "XXXYYYZZZ", "verifying_device_key": "abcdef" }
             end
 
-            Note over A: 20. Check that verifying_device_id and verifying_device_key match data from Homeserver
-            A->>A: 21. Mark existing device (XXXYYYZZZ) as verified locally
-            A->>A: 22. If master_key was received then mark as trusted
-            alt Cross-signing active
-                A->>HS: m.secret.request for MSK/USK/SSK private keys from existing device XXXYYYZZZ
+            Note over A: 20. Check that verifying_device_id and verifying_device_key match data from homeserver. If no match then abort
+            A->>A: 21. If master_key was received then mark as trusted
+            alt `master_key` was received
+                Note over A: 21a. Check that it matches the master key<br>received from the homeserver. If it does not match then abort the process reporting a security error to the user.
+                A->>A: 21b. Mark master key as trusted
+            end
+            A->>A: 22. Mark existing device (XXXYYYZZZ) as verified locally
+
+            A->>HS: m.secret.requests for MSK/USK/SSK/backup secrets from existing device XXXYYYZZZ
+            loop
                 HS->>B: m.secret.request
                 B->>HS: m.secret.send
                 HS->>A: m.secret.send
-                A->>A: Check and store received private keys locally
+                A->>A: Check and store received secret locally
             end
-
-            A->>HS: m.secret.request for m.megolm_backup.v1 key
-            HS->>B: m.secret.request
-            B->>HS: m.secret.send
-            HS->>A: m.secret.send with key if known
-            A->>A: Connect to key backup
         end
     end
 ```
@@ -263,14 +262,18 @@ New device informs existing device of outcome and if E2EE is to be setup then in
 
 16. If doing E2EE then existing device then waits for up to X seconds for the `device_id` to become visible.
 
-17. If the device is visible within the time period then the existing device must first check that the `device_id` and
-`device_key` match those provided by the homeserver.
+17. If the device is visible within the time period then the existing device
+must first check that the `device_id` and `device_key` match those provided by
+the homeserver. If they do not match then abort the process reporting a security
+error to the user.
 
-18a. Assuming they match then locally mark the device as verified.
+18a. Mark the new device as locally verified on existing device
 
-18b. If cross signing is in use then cross-sign the new device and upload signature to Homeserver.
+18b. If cross-signing is in use then cross-sign the new device and upload
+signature to homeserver.
 
-19. The existing device notified the new device that verification has been completed on its end:
+19. The existing device notified the new device that verification has been
+completed on its end:
 
 ```json
     "type": "m.login.finish",
@@ -289,13 +292,23 @@ If cross-signing is in use then the public part of the master signing key `maste
     "master_key": "mmmmmmmm"
 ```
 
-20. The new device checks that `verifying_device_id` and `verifying_device_key` match those from the Homeserver
+20. The new device checks that `verifying_device_id` and `verifying_device_key`
+match those received from the homeserver. If they do not match then abort the
+process reporting a security error to the user.
 
-21. If they do match then locally mark the existing device as verified.
+21a. If the `master_key` was received then check that it matches the master key
+received from the homeserver. If it does not match then abort the process
+reporting a security error to the user.
 
-22. If the `master_key` was received then mark it as trusted
+21b. If it matched the new device marks the `master_key` as locally trusted.
 
-23. The new device requests that the verifying device share the `m.cross_signing.master`, `m.cross_signing.user_signing`, `m.cross_signing.self_signing`, `m.megolm_backup.v1` secrets by sending [`m.secret.request` device events](https://spec.matrix.org/v1.6/client-server-api/#sharing)
+22. Only after the above security checks, only now should the new device mark
+the verifying device as locally verified.
+
+23. The new device requests that the verifying device share the
+`m.cross_signing.master`, `m.cross_signing.user_signing`,
+`m.cross_signing.self_signing`, `m.megolm_backup.v1` secrets by sending
+[`m.secret.request` device events](https://spec.matrix.org/v1.6/client-server-api/#sharing)
 
 ### Anticipated errors
 
@@ -327,7 +340,7 @@ sequenceDiagram
     participant A as New device <br>wanting to sign in
     participant B as Existing device <br>already signed in
     participant OP as OIDC Provider
-    participant HS as Homeserver
+    participant HS as 
     rect rgba(240,240,240,0.5)
         Note over A,B: These steps are same as login.token flow
         alt Code was scanned on new device
--- a/proposals/3906-sign-in-with-qr.md
+++ b/proposals/3906-sign-in-with-qr.md
@@ -86,7 +86,7 @@ Note over A,B: Devices establish secure rendezvous and derive<br> confirmation c
         rect rgba(240,240,240,0.5)
             Note over A,B: These are from MSC3886
             A->>B: Rendezvous set up and ECDH public key encoded in QR
-            B->>A: New device responds with ECDH public key
+            B->>A: Existing device responds with ECDH public key
         end
         Note over B: 1. Check compatibility of intents
         alt Incompatible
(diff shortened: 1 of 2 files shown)


uhoreg:
I've added the needs-implementation label because we need to check the linked implementations.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,472 @@
...
+3. In the case that the new device scanned the code an empty progress payload is sent so that the the existing device
+knows it can proceed
+
+```json
+{
+    "type": "m.login.progress"
+}
+```
+
+4. The existing device determines if the homeserver has sufficient capabilities to support the request (i.e. support for MSC3882).
```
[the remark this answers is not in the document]
uhoreg: Doesn't the MSC3882 request always go to the existing device's own homeserver?

dbkr (commented):
Looks generally good as far as I can see, although I echo @uhoreg 's comment about m.login.progress having a few different purposes.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,472 @@
...
+    "master_key": "mmmmmmmm"
+```
+
+20. The new device checks that `verifying_device_id` and `verifying_device_key` match those from the Homeserver
+
+21. If they do match then locally mark the existing device as verified.
+
+22. If the `master_key` was received then mark it as trusted
+
+23. Request 4x secrets be gossiped from existing device and set them up as per normal verification
```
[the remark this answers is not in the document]
hughns: I meant the four secrets: MSK, USK, SSK, room key backup key. I will clarify

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,510 @@
...
+
+```json
+{
+    "type": "m.login.finish",
+    "outcome": "unsupported"
+}
+```
+
+8. A 12 numerical digit confirmation code derived from the shared key used by
+the rendezvous channel must be displayed on both devices.
```
deepbluev7, line 222: I doubt anyone actually checks those. What prevents us from instead showing a second QR code?

In theory, since device A shared its public key securely via a QR code, it should be able to trust, that anything sent to it can be trusted. As such device B should be able to send device A the 12 digit code over that channel and device A then decides to cancel the login or not. Device A could share the 12 digit code to B using a QR code before that, although I am not sure if that is necessary, since device B has direct line of sight to A.

I am probably missing something, but relying on users to manually verify 12 digits to ensure, that the channel is secure AFTER they already scanned a QR code sounds like something nobody would do, so making that automatic should improve security imo.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,472 @@
...
+    "type": "m.login.finish",
+    "outcome": "unsupported"
+}
+```
+
+6. Otherwise, the existing device acknowledges the request and indicates the protocols available:
+
+```json
+{
+    "type": "m.login.progress",
```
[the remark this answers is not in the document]
hughns: I've reworked the flow to try and make better use of the `type` fields.

proposals/3906-sign-in-with-qr.md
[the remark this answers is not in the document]
hughns: If the QR code is generated on the existing device then the check for MSC3882 can indeed be done ahead of time. This is what is done in the current Element Web implementation.

If the QR code is generated on the new device, then it is also the case that the MSC3882 check can be done ahead of offering the ability to scan the QR.

I've added a note about this in https://github.com/matrix-org/matrix-spec-proposals/commit/c7906ba1df3c9598362709b79b7d1ab548729814

turt2live:
The author believes this is ready for FCP. 

@matrix-org/spec-core-team : Please review the implementation and MSC with an aim of `@mscbot fcp merge`-ing it.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,528 @@
...
+There are currently two use cases for this proposal:
+
+- allowing a user to login and setup E2EE on an additional Matrix client by means
+of scanning a QR code
+- a mechanism to facilitate the launching of advanced E2EE aware "widgets" which
+are acting as full Matrix clients
+
+## Secure channel prerequisite
+
+This proposal relies on a secure "rendezvous" channel having been established
```
uhoreg, line 16: I think we need more of a description of what "secure" means here.  I believe that what we need here is secrecy (an attacker can't read the messages), integrity (an attacker can't modify messages), and authentication (we can determine that the message was sent from the other device).

Edit: also, it seems like there are some extra concepts ("flow" and "intent") that are part of the rendezvous channel establishment.  It would be helpful to have those explained here.  If they're explained in other MSCs, it would be helpful to have pointers to those MSCs here, and maybe give a summary of it here, or even just a list.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,528 @@
...
+
+The purpose of the `flow` field is to allow for future extensions such as updated
+flow semantics or other flows entirely.
+
+## Proposal
+
+This proposal defines the `m.setup.additional_device.v2` flow that is to be
+used once a secure channel has been established.
+
+The following `intent` values are defined:
```
uhoreg, line 54: I find it confusing that we're defining `intent` values before we say what the values are for.  I think this can be solved by adding some more words.  Maybe something along the lines of: "This proposal involves communication between two devices: one that is already logged in, and one that is not logged in.  We assign `intent` values, which are sent as part of the rendezvous channel establishment(?), to identify the two devices.  These values are..."

Or maybe it could be solved with more words about how the secure channel is established, if this is indeed part of the secure channel establishment?  Will add another comment above.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,528 @@
...
+This proposal should also work with a to_device based transport channel too.
+
+Furthermore it should be adaptable to work with OIDC based authentication in future.
+
+The purpose of the `flow` field is to allow for future extensions such as updated
+flow semantics or other flows entirely.
+
+## Proposal
+
+This proposal defines the `m.setup.additional_device.v2` flow that is to be
```
uhoreg, line 51: Similar to the comment below, I don't know what "flow" means in this context.  "Flow" is mentioned in the example above, but it's not clear how that example relates to the general case.  Is it defined in a different MSC?

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,528 @@
...
+    B->>+HS: POST /versions
+    HS-->>-B: supported features 
+    alt no protocol available
+        B->>A: 5. {"type":"m.login.failure", "reason":"unsupported"}
+    else
+        B->>A: 6. {"type":"m.login.protocols", "protocols":["login_token"]}
+    end
+
+    Note over A: 7. Check that suitable protocol available
+    A->>B: 10. {"type":"m.login.protocol", "protocol": "login_token"}
```
uhoreg, line 109:
```suggestion
    A->>B: 7. {"type":"m.login.protocol", "protocol": "login_token"}
```

?

uhoreg (commented):
Overall, I think the idea is sound, but could do with some clarifications.

proposals/3906-sign-in-with-qr.md
deepbluev7: But since Device B can trust that Device A's public key is the right one, if Device B sends it the confirmation key A sent over the QR code, then there should be no way for Device M to intercept the message, because it is encrypted using Device A's public key, which in theory could only have been intercepted by someone in the same room. So that way Device A should be able to establish this connection as trusted.

I don't think you can really put that into the rendevouz MSC, because that works with arbitrary channels, while in this MSC you are explicitly using an asymmetric, one directional, secure channel (the QR code). However automatically verifying the channel as secure, by sending a secret from B to A using A's public key should be able to mark the channel as trusted also in the direction A to B.

In general I just don't like the manual confirm step, because I don't think people will verify those numbers after they already scanned a QR code. And I don't think it should be necessary, if you don't restrict yourself to one QR code.

uhoreg: > But since Device B can trust that Device A's public key is the right one, if Device B sends it the confirmation key A sent over the QR code, then there should be no way for Device M to intercept the message, because it is encrypted using Device A's public key, which in theory could only have been intercepted by someone in the same room. So that way Device A should be able to establish this connection as trusted.

Right, if Device A scans a QR code generated by Device B that contains the confirmation key, then all is good.  However, we this requires that both devices are able to scan QR codes -- Device B scanning Device A initially to get its public key, and then Device A scanning device B to get the confirmation key.  Ideally, we would only need one device to do the scanning.

> I don't think you can really put that into the rendevouz MSC, because that works with arbitrary channels, while in this MSC you are explicitly using an asymmetric, one directional, secure channel (the QR code). However automatically verifying the channel as secure, by sending a secret from B to A using A's public key should be able to mark the channel as trusted also in the direction A to B.

My understanding is that the "secure channel" is bi-directional, and the QR code is only used to help initialize the channel.  Since we're reading this differently, it seems like this should be made clearer in the MSC, exactly how this works.

FSG-Cat: Ok so you are concerned over that users verify a 12 digit code. Why dont we just use the Emoji system we use for verification? Like UX wise numbers are unacceptable no matter what. The Emoji system is WAY more usable. Comparing 16 emojies is actually realistic compared to 12 digits (Yes its that bad i would argue). 

Like this is why the NATO alphabet exists. Because using distinct words is way more clear and the same goes for showing a set of Emoji in your UI instead of digits. 

Im not going to comment on the need or lack of need for this mechanism as im commenting from a prespective of someone who knows enough UX and enough matrix to conclude that dont we already have a system that solves this very type of problem. Users having to verify long blobs of data to prevent MITM and other malicious actions.

proposals/3906-sign-in-with-qr.md
```diff
@@ -0,0 +1,528 @@
...
+        Note over A: 1. Check compatibility of intents
+        alt Incompatible
+            A->>B: 2. {"type":"m.login.failure", "reason": "incompatible_intent", "intent": "login.start"}
+            B->>B: Cancel rendezvous
+        end
+    else Code was scanned on existing device
+        rect rgba(240,240,240,0.5)
+            Note over A,B: These are from MSC3886
+            A->>B: Rendezvous set up and ECDH public key encoded in QR
+            B->>A: New device responds with ECDH public key
```
richvdh, line 89: slightly confused. The text here says "New device responds with ECDH public key" but the arrow shows a message from the existing device to the new device?

turt2live:
@hughns fyi this has entered a "not active SCT focus" state, which is to say it feels like it's back with you for the moment. When this changes (or if the status is wrong), let us know in the SCT office.

proposals/3906-sign-in-with-qr.md
hughns: With regards to "UX wise numbers are unacceptable no matter what":

I would argue that the general human population are already much more familiar with handling a sequence of numbers in a security context - e.g. 4 digit PIN from a bank card or 6 digits from an SMS/authenticator OTP - than handling a sequence of emojis or words.

So, UX wise we can can make it easier for users by utilising something they are familiar with rather than having to educate them on an (entirely) new concept.

xyhhx: > I would argue that the general human population are already much more familiar with handling a sequence of numbers in a security context - e.g. 4 digit PIN from a bank card or 6 digits from an SMS/authenticator OTP - than handling a sequence of emojis or words.

I would agree if we were asking the user to type anything in anywhere; but that's not the case here. Emoji are much easier to distinguish from each other.

> So, UX wise we can can make it easier for users by utilising something they are familiar with rather than having to educate them on an (entirely) new concept.

It's not an entirely new concept? Matrix users are used to comparing emoji for cross signing already, and this is also what is used to compare secure comms in other platforms (like Telegram).

wj25czxj47bu6q: I agree with previous comments that virtually nobody is going to bother comparing numbers after already scanning a QR code, and I think the same will hold true for a comparison of emojis.

I propose instead that a 4&#8209; or 6&#8209;digit numeric code should be generated on one of the clients (doesn't matter which from a UX standpoint) that the user then has to _actively type in_ on the other client.

Requiring the user to actively input a code is a far more effective mitigation against interception than a system in which the user can simply press 'accept' without actually performing any verification. Google, for example, [employs a similar system where the user is asked to choose the matching number out of three choices](https://security.stackexchange.com/questions/255422/what-is-googles-match-the-number-verification-trying-to-achieve). The proposed comparison, either of emojis or numbers, is for the average user no better than [Discord's sorely inadequate accept&#8209;or&#8209;deny prompt](https://support.discord.com/hc/en-us/articles/360039213771) that has allowed ["Free Nitro"](https://discord.com/blog/common-scams-what-to-look-out-for) and other account-compromising scams to continue being effective.

(I do believe a similar change should be made to the cross&#8209;signing UX, but that is out of scope for this discussion.)

MTRNord: > I would argue that the general human population are already much more familiar with handling a sequence of numbers in a security context - e.g. 4 digit PIN from a bank card or 6 digits from an SMS/authenticator OTP - than handling a sequence of emojis or words.

You are aware that even at the invention of PIN codes, this UX was questionable, which is why we have only 4 numbers in a general banking pincode? https://en.wikipedia.org/wiki/Personal_identification_number#PIN_length

hughns:
It is proposed that [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108) supersedes this MSC.

uhoreg:
> It is proposed that [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108) supersedes this MSC.

@hughns As the author of this MSC, you can close this MSC if it is superseded by another one.

hughns:
Closing this PR as https://github.com/matrix-org/matrix-spec-proposals/pull/4108 is now ready for review.

PR #3907, opened 2022-10-13
MSCXXXX Mute Policy Recommendation

FSG-Cat:
Rendered FIX ME

MSC Co Authored by @FSG-Cat and @Gnuxie 


proposals/3907-mute-policy-recommendation.md
```diff
@@ -0,0 +1,135 @@
...
+
+Mutes will become less clumsy to use if they can be applied as policy. Especially when used together with
+MSC3907 to make them be able to expire on their own. MSC3909 will in the future also make them even better
+by making them a membership state if that MSC is implemented. This MSC is technically fully independent
+of these 2 MSCs.
+
+## Proposal
+
+This MSC will create a new Policy List recommendation of `m.mute`. Below is an example event for this MSC
+that displays how it would be used in conjunction with MSC3907.
```
MTRNord, line 29:
```suggestion
that displays how it would be used in conjunction with MSC3908.
```

seems like a typo

proposals/3907-mute-policy-recommendation.md
```diff
@@ -0,0 +1,135 @@
...
+  },
+  "sender": "@mjolnir:example.com",
+  "state_key": "_evil:example.com",
+  "type": "m.policy.rule.user",
+  "event_id": "exampleid",
+  "room_id": "!policylist:example.com"
+}
+```
+
+Below is the same event but without the expiry.
```
MTRNord, line 47: Does it even make sense to have this without an expiry? Isn't a mute without expiry essentially a ban, or what's the reasoning?

FSG-Cat: Yes a mute without a expiry is on the surface close to a ban but if one makes the assumption that policy application engines reverse policies that have been reversed then the assumption of it being a ban would no longer be true.

Even then a mute is in some ways not a ban since it forces you to be read only but does not eject you out of a room. Close but not equivalent and this lack of equivalence can actually end up being important for some communities.

proposals/3907-mute-policy-recommendation.md
```diff
@@ -0,0 +1,135 @@
...
+    or pings from them.
+  + Applied to a room: The user's power level should be lowered below the required level to send
+    `events_default`.
+  + Applied to a server: The user should be suspended from sending events and uploading media.
+    Effectively leaving the user's account in a read-only state.
+
+* is a `room` rule:
+  + Applied to a user: New events from the room are ignored, not displayed in notifications and
+    not shown to be active in a client.
+  + Applied to a room: No-op? Could be changing `events_default` to the same level as moderators
```
Kladki, line 87: Not sure why this is a question, you cannot ban a room from a room, same with mutes.

PR #3909, opened 2022-10-13
MSCXXXX: Membership based mutes

FSG-Cat:
Rendered FIX ME

MSC Co Authored by @FSG-Cat and @Gnuxie 


proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,88 @@
...
+exclusively send EDUs and no PDUs.
+
+## Proposal
+
+This proposal defines the `mute` membership type. If a user has this membership type in a room they
+are restricted to exclusively sending the read marker EDUs and nothing more except leaving via setting
+their membership to `ban` as defined later.
+
+The legal transitions for the `mute` membership type based on Room version 10 would be
+`join` -> `mute` -> `ban`, `join`, `leave`(Exclusively via kick)
```
uhoreg, line 19: Presumably, the `mute` -> `join` transition cannot be done by the user themselves, and can only be done by a user with sufficient power levels.

FSG-Cat: Yes that is something that should be clarified.

FSG-Cat: The new text should resolve this concern.

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,88 @@
...
+To be written. Contributions welcome.
+## Potential issues
+
+On the potential issue that this MSC requires rooms that want to employ it to be created after the room version
+its included in becomes default or their homeserver makes it default or they manually select it for an upgrade is
+deemed acceptable. The reason this is deemed acceptable is because as has been made clear in discussions with
+community moderation teams for multiple communities not all communities want to use mutes in their toolkit and therefore
+not all communities will feel the need to upgrade their room version to gain access to this feature. 
+
+Room version upgrades them self's are being worked upon making more smooth. There is work in the area of deleting state
```
Mikaela, line 43:
```suggestion
Room version upgrades themselves are being worked upon making more smooth. There is work in the area of deleting state

```

Mikaela: I think this was a spelling mistake as "...upgrades them self's are..." would expand to "...upgrades them self is are...", but English is weird and I am not a native speaker

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,89 @@
...
+To be written. Contributions welcome.
+## Potential issues
+
+On the potential issue that this MSC requires rooms that want to employ it to be created after the room version
+its included in becomes default or their homeserver makes it default or they manually select it for an upgrade is
+deemed acceptable. The reason this is deemed acceptable is because as has been made clear in discussions with
+community moderation teams for multiple communities not all communities want to use mutes in their toolkit and therefore
+not all communities will feel the need to upgrade their room version to gain access to this feature. 
+
+Room version upgrades themselves are being worked upon making more smooth. There is work in the area of deleting state
```
Mikaela, line 43: There is still something fishy on this line. "...worked upon making more smooth"? 

Missing *them* or *the upgrade experience* or similar?

FSG-Cat: Reworded to potentially resolve this concern.

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,88 @@
+# MSC3909: Membership based mutes
```
FSG-Cat, line 1: @luxaritas putting this as a thread as its the correct way to do things. 

A leave-mute membership state would actually bypass the issue that requires currently self banning. Because the rule could be that leave is only legal via kick and self leaves are instead leave-mutes and for leave mute the only legal self initiated transition is mute meaning you would essentially gain sticky mutes and bypass this whole self ban stuff. Its quite a good solution to the problem to be honest and it bypasses the whole concern of mine of having to look multiple layers deep in the chain to evaluate legality.

luxaritas: Oops yeah sorry I didn't comment in-line!

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,102 @@
...
+exclusively send Read markers and exempt events as defined later.
+
+A passive benefit of this MSC that is notable is that with the introduction of Partial joining of rooms the minimum
+state to join said room is important. As far as the authors understand this state is obligatorily going to be
+negatively effected by PL events climbing in number. This MSC mitigates this and this is a real world
+relevant example of why this MSC matters.
+## Proposal
+
+This proposal defines the `mute` membership type. If a user has this membership type in a room they
+are restricted to exclusively sending the read marker EDUs, leaving via setting
```
clokep, line 19:
```suggestion
are restricted to exclusively sending receipts, leaving via setting

```

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,103 @@
...
+negatively effected by PL events climbing in number. This MSC mitigates this and this is a real world
+relevant example of why this MSC matters.
+## Proposal
+
+This proposal defines the `mute` membership type. If a user has this membership type in a room they
+are restricted to exclusively sending receipts, leaving via setting
+
+their membership to `leave-mute` as defined later and `m.room.redaction` events.
+
+This proposal also introduces the new membership type of `leave-mute` this special membership
```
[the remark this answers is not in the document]
olivia-fl: Yeah, this seems like a simpler and more future-proof mechanism to me, especially if we want to add additional flags like this at some point. Combinatorial explosion of membership states sounds bad. A secondary benefit of a separate field is that it's backwards-compatible with clients that don't know about it.

proposals/3909-membership-based-mutes.md
```diff
@@ -0,0 +1,103 @@
...
+`ban`.
+
+The legal transitions for the `leave-mute` membership type based on Room version 10 would be
+`mute` -> `leave-mute` -> `invite`(Requires sufficient PL to set membership to `mute` and to set membership to `invite`) , 
+`leave-mute`, `leave`(Requires sufficient PL to set membership to `mute`) `ban`.
+
+### Authorisation rules.
+
+To be written. Contributions welcome.
+## Potential issues
```
turt2live, line 38: Like on MSC4106, I suspect this will be easier to achieve with policy servers - https://github.com/matrix-org/matrix-spec-proposals/pull/4106#discussion_r2770556984

This MSC doesn't require the same "unmute" mechanic that MSC4106 would need, but it does feel a bit difficult to get all the membership transitions right. The diagram at https://spec.matrix.org/v1.17/client-server-api/#room-membership is already quite complicated.

PR #3910, opened 2022-10-14
MSC3910: Content tokens for media

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/propsal/content_tokens_for_media/proposals/3910-content-tokens-for-media.md)

### DIFF
--- a/proposals/3910-content-tokens-for-media.md
+++ b/proposals/3910-content-tokens-for-media.md
@@ -1,0 +1,349 @@
+# MSC3910: Content tokens for media
+
+(Based on [MSC~~701~~3796](https://github.com/matrix-org/matrix-spec-proposals/issues/3796).)
+
+Currently, access to media in Matrix has the following problems:
+ 
+* The only protection for media is the obscurity of the URL, and URLs are
+  easily leaked (eg accidental sharing, access
+  logs). [synapse#2150](https://github.com/matrix-org/synapse/issues/2150)
+* Anybody (including non-matrix users) can cause a homeserver to copy media
+  into its local
+  store. [synapse#2133](https://github.com/matrix-org/synapse/issues/2133)
+* When a media event is redacted, the media it used remains visible to all.
+  [synapse#1263](https://github.com/matrix-org/synapse/issues/1263)
+* There is currently no way to delete
+  media. [matrix-spec#226](https://github.com/matrix-org/matrix-spec/issues/226)
+* If a user requests GDPR erasure, their media remains visible to all.
+* When all users leave a room, their media is not deleted from the server.
+
+## Proposal
+
+### Overview
+
+Each piece of uploaded media gains an associated `content_token`.  This is
+included alongside the `mxc://` uri whenever the content is referenced; for
+encrypted events it is included in the *cleartext* (unencrypted) part of the
+event. Critically, it is never embedded into a URL.
+
+Clients must present both their `access_token` (to prove they have the right to
+use the server) *and* the `content_token` (to prove they have the right to see
+the media in question).
+
+Servers can count the number of references to a given `content_token`, and
+expire any media whose refcount falls to zero.
+
+### Detailed spec changes
+
+1. [`/_matrix/media/v3/upload`](https://spec.matrix.org/v1.4/client-server-api/#post_matrixmediav3upload)
+   creates a (cryptographically-secure) random "content token" for the media,
+   and returns it alongside the `content_uri`. Result now looks like:
+   ```json
+   {
+     "content_uri": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+     "content_token": "hunter2"
+   }
+   ```
+2. Passing the content token on to other users:
+   * When sending `m.image` and `m.video` messages, the client includes a `content_token` property:
+     ```json
+     {
+       "type": "m.room.message",
+       "content": {
+         "body": "filename.jpg",
+         "msgtype": "m.image",
+         "url": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+         "content_token": "hunter2"
+       }
+     }
+     ```
+   * Encrypted media messages must include the `content_token` in the cleartext:
+     ```json
+     {
+       "type": "m.room.encrypted",
+       "content": {
+         "algorithm": "m.megolm.v1.aes-sha2",
+         "session_id": "<outbound_group_session_id>",
+         "ciphertext": "<encrypted_payload_base_64>",
+         "content_token": "hunter2"
+       }
+     }
+     ```
+   * Likewise `m.room.avatar_url` must include a `content_token`:
+     ```json
+     {
+       "type": "m.room.avatar",
+       "state_key": "",
+       "content": {
+         "url": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+         "content_token": "hunter2"
+       }
+     }
+     ```
+   * And `m.room.member` must include an `avatar_content_token`:
+     ```json
+     {
+       "type": "m.room.member",
+       "state_key": "@alice:example.org",
+       "content": {
+         "avatar_url": "mxc://example.org/AQwafuaFswefuhsfAFAgsw",
+         "avatar_content_token": "hunter2",
+         "displayname": "Alice Margatroid",
+         "membership": "join"
+       }
+     }
+     ```
+   * [`PUT /_matrix/client/v3/profile/{userId}/avatar_url`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3profileuseridavatar_url) 
+     requests should include an `avatar_content_token` alongside the `avatar_url`.
+
+3. New download endpoints
+     
+   The existing download endpoints are to be deprecated, and replaced with new
+   endpoints specific to client-server or federation requests:
+     
+   | Old                                                                                                                                                              | Client-Server                                                             | Federation                                                          |
+   | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------- | ------------------------------------------------------------------- |
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid)            | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}`            | `GET /_matrix/federation/v1/media/download/{serverName}/{mediaId}`  |
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}/{fileName}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid) | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}/{fileName}` | N/A                                                                 |
+   | [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3thumbnailservernamemediaid)          | `GET /_matrix/client/v1/media/thumbnail/{serverName}/{mediaId}`           | `GET /_matrix/federation/v1/media/thumbnail/{serverName}/{mediaId}` |
+
+   [Question: should we move `/config`, `/upload` and `/preview_url` while
+   we're at it, per [MSC1902](https://github.com/matrix-org/matrix-spec-proposals/pull/1902)?]
+
+   None of the new endpoints take an `allow_remote` query parameter. (For
+   `/_matrix/client`, servers should always attempt to request remote
+   media. For `/_matrix/federation`, servers should never attempt to request
+   remote media they do not already have cached.)
+
+   **Two** authorization headers must be given when calling these endpoints:
+     
+   * `Authorization` must be set the same as for any other CSAPI or federation
+     request (ie, `Bearer {accessToken}` for `/_matrix/client`, or the
+     signature for `/_matrix/federation`).
+   * `X-Matrix-Content-Token` must be set to the `content_token`.
+
+   `X-Matrix-Content-Token` should be omitted for legacy media which has no
+   `content_token`.  Otherwise, if no `X-Matrix-Content-Token` header is given,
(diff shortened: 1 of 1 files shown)


proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+  logs). [synapse#2150](https://github.com/matrix-org/synapse/issues/2150)
+* Anybody (including non-matrix users) can cause a homeserver to copy media
+  into its local
+  store. [synapse#2133](https://github.com/matrix-org/synapse/issues/2133)
+* When a media event is redacted, the media it used remains visible to all.
+  [synapse#1263](https://github.com/matrix-org/synapse/issues/1263)
+* There is currently no way to delete
+  media. [matrix-spec#226](https://github.com/matrix-org/matrix-spec/issues/226)
+* If a user requests GDPR erasure, their media remains visible to all.
+* When all users leave a room, their media is not deleted from the server.
```
richvdh, line 18: @turt2live says: fwiw, while the spec allows servers to GC rooms, Synapse currently doesn't (no server does, I believe)

@richvdh says: yes, this is https://github.com/matrix-org/synapse/issues/4720.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+       "type": "m.room.message",
+       "content": {
+         "body": "filename.jpg",
+         "msgtype": "m.image",
+         "url": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+         "content_token": "hunter2"
+       }
+     }
+     ```
+   * Encrypted media messages must include the `content_token` in the cleartext:
```
richvdh, line 60: @turt2live says:

> For clarity, would we do the same as relations and say that if the payload *also* has a `content_token` that it is ignored? ie: Client sends `content_token` in payload but not cleartext, or sends it in both.

@richvdh says:

> Same as relations, surely.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid)            | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}`            | `GET /_matrix/federation/v1/media/download/{serverName}/{mediaId}`  |
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}/{fileName}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid) | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}/{fileName}` | N/A                                                                 |
+   | [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3thumbnailservernamemediaid)          | `GET /_matrix/client/v1/media/thumbnail/{serverName}/{mediaId}`           | `GET /_matrix/federation/v1/media/thumbnail/{serverName}/{mediaId}` |
```
richvdh, line 108: @turt2live says:

> (origin,mediaId) tuple - media ID is not globally unique on its own.

... but I can't figure out what he's referring to.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+
+   [Question: should we move `/config`, `/upload` and `/preview_url` while
+   we're at it, per [MSC1902](https://github.com/matrix-org/matrix-spec-proposals/pull/1902)?]
```
richvdh, line 111: @turt2live says:

> yes please, though we should probably leave splitting endpoints out to its own MSC anyways - this to-be MSC could depend on MSC1902.
>
> Context: The title of the MSC imposes "problems" on what can be included in that MSC. Though relevant to the MSC itself, it's not obvious from the title that we'd also be changing where the endpoints live (and making the title a paragraph is really just crap). Suggestion is always to just split out functionality when one would add "and" to the title.

@richvdh says:

> The problem with having separate MSCs is that, if we land MSC1902 without the auth changes, then we need *another* way to distinguish clients that support auth from those that don't.
>
> In this proposal, we can tell when it's safe to turn off the backwards-compatibility "you can get media without auth" behaviour just by looking at the access logs and seeing who is hitting the old endpoint. If we extend existing endpoints, that gets way harder.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+
+   None of the new endpoints take an `allow_remote` query parameter. (For
+   `/_matrix/client`, servers should always attempt to request remote
+   media. For `/_matrix/federation`, servers should never attempt to request
+   remote media they do not already have cached.)
+
+   **Two** authorization headers must be given when calling these endpoints:
+     
+   * `Authorization` must be set the same as for any other CSAPI or federation
```
richvdh, line 120: @turt2live says:

> nit: clients can still auth with ?access_token, but obviously discouraged. We should probably just kill that feature in the spec (independent of this MSC)

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+   /_matrix/client/v1/media/download/{remoteServer}/{mediaId}`, it first
+   validates the `access_token` in the `Authorization` header.
+   
+   If the access token is valid, the server checks its cache for
+   `(remoteServer, mediaId)`. If the media is cached, we should also have a
+   record of the expected `content_token`.  The server therefore checks the
+   `X-Matrix-Content-Token` header, and returns either the media or a 403
+   error.
+   
+   If the media is uncached, the user's server makes a request to `GET
```
richvdh, line 153: @uhoreg says:

> Would it make sense to first check that you have an event that references that content_token before downloading remote media?

@richvdh says:

> 🤔 I guess it might

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+token in the `Authorization` header.
+
+This mechanism also works for a secondary use of the content repository in
+bridges: as a paste-bin. In this case, the bridge simply generates a link
+referencing its own media.
+
+The bridge might also choose to embed information such as the room that
+referenced the media, and the time that the link was generated, in the URL.
+(Such information would require an HMAC to prevent tampering.) This could be
+used to help control access to the media.
```
richvdh, line 234: @turt2live says:

> Sending a meta event into the room would be important for the server to flag a "usage" of the content, otherwise it'd be subject to expiration/GC.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+
+We also need to ensure that the icons are not deleted from the content
+repository even though their `content_token`s are not referenced anywhere.  It
+would be wise for servers to provide administrators with a mechanism to prevent
+cleanup of certain media items.
+
+(This was previously discussed in
+[MSC2858](https://github.com/matrix-org/matrix-spec-proposals/pull/2858#discussion_r543513811).)
+
+## Potential issues
```
richvdh, line 281: @turt2live says:

> With the proposal above, the origin server for a piece of media could count a piece of media (it's original copy) as "unreferenced" if all *visible* copies are dereferenced. Specifically, if a user on alice.com sends something to a room and a user from bob.com forwards that to a room visible to charlie.com, alice.com could eventually consider the media unreferenced and charlie.com users wouldn't be able to download it.

richvdh, line 281: @turt2live says:

> Issue 2: Custom emoji and such rely on `` tags in the message HTML - clients (and the spec) generally require that the img `src` be an MXC URI - presumably we just add a `mx_content_token` attribute or something to fix this case.

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,366 @@
...
+   
+   Servers must only do this if there is no `X-Matrix-Content-Token` in the
+   request (the fact that a content token is available implies that the origin
+   server must support the new endpoints).
+   
+   Clients should retry the request whether or not they have a content token.
+
+   b. Backwards compatibility with older clients and federating servers:
+   servers may for a short time choose to allow unauthenticated access via the
+   deprecated endpoints, even for media with an associated `content_token`.
```
richvdh, line 180: @turt2live says:

> I'm not sure this would be needed? The origin server (over federation) should still support the request if it also produced a content_token.

proposals/3910-content-tokens-for-media.md
turt2live: somewhere in the doc it said "look up the media in the cache by media ID" - I think HackMD was trying to be helpful and put the comment in a weird place.

proposals/3910-content-tokens-for-media.md
richvdh: Consider a situation where `@alice:orgin.com` is using a new server, and has uploaded some media. The `origin.com` server now needs to deal with two sets of legacy implementations:

 * `@bob:origin.com`, who is using an old client which doesn't know how to use content tokens.
 * `@charlie:remote.com`, where `remote.com` is an old server implementation. Whether or not charlie's client understands content tokens, `remote.com` isn't going to forward them.

In either case, `origin.com` will receive a request to the deprecated endpoint, and may choose to honour it (for now).

turt2live: Somewhere it was clarified that this was talking about old clients talking to old servers in rooms where "new" media is posted - this wasn't immediately clear from the text, so I think it just needs a relatively minor clarification?

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,367 @@
...
+       "type": "m.room.member",
+       "state_key": "@alice:example.org",
+       "content": {
+         "avatar_url": "mxc://example.org/AQwafuaFswefuhsfAFAgsw",
+         "avatar_content_token": "hunter2",
+         "displayname": "Alice Margatroid",
+         "membership": "join"
+       }
+     }
+     ```
```
richvdh, line 95: also `m.presence`

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,367 @@
...
+2. Passing the content token on to other users:
+   * When sending `m.image` and `m.video` messages, the client includes a `content_token` property:
+     ```json
+     {
+       "type": "m.room.message",
+       "content": {
+         "body": "filename.jpg",
+         "msgtype": "m.image",
+         "url": "mxc://example.com/AQwafuaFswefuhsfAFAgsw",
+         "content_token": "hunter2"
```
richvdh, line 56: Spotted by @uhoreg: a server doesn't, in general, know which event types reference media, and where in their structure the media url would be.

There might also be more than one media item. So instead, we have to define `content_token` to be a map at the top level of the event, with keys being the *hash* of an mxc url (to uniquely identify the media, without leaking the mxc uri outside encrypted events).

turt2live: this doesn't neccesarily solve the concern about media embedded into the content of an event, like custom emoji: https://github.com/matrix-org/matrix-spec-proposals/pull/3910#discussion_r996174591

(there's no key on the json for such media references)

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,367 @@
...
+    has a right to see it (as we would with its associated event), and if so,
+    include the associated event in the response. The requesting server then
+    filters further.
+  
+  (credit: based on ideas at
+  https://cryptpad.fr/code/#/2/code/view/oWjZciD9N1aWTr1IL6GRZ0k1i+dm7wJQ7juLf4tJRoo/)
+  
+  Problems with this approach:
+   * Forwarded media would require either a re-upload, or a server-side duplication.
+     (Note that matrix-media-repo has a [LocalCopy](https://github.com/turt2live/matrix-media-repo/blob/master/api/unstable/local_copy.go#L19) implementation for this.)
```
richvdh, line 312: I've now written up something much like this at MSC3911.

proposals/3910-content-tokens-for-media.md
deepbluev7, line 56: Since this is data returned by the upload request and events can reference multiple media events, this probably should be an array or so. I.e. inline images, thumbnails for e2ee images, thumbnails for video files, etc.

uhoreg: Yes, that's essentially https://github.com/matrix-org/matrix-spec-proposals/pull/3910#discussion_r996504784

proposals/3910-content-tokens-for-media.md
```diff
@@ -0,0 +1,367 @@
...
+  the query string; for example, encrypt the content token with a key based on
+  the access token? Tying it to the specific user means that the link isn't any
+  use to anyone else.
+  
+  Not sure this offers much advantage over a separate header, whereas it does
+  add complexity.
+  
+* Put the `content_token` within the encrypted content, for encrypted data.
+  This reduces the metadata leak, but means that none of the automated media
+  removal functionality works.
```
dkasak, line 334: I don't think putting the `content_token` in the cleartext actually solves the problem for automated media removal. And I don't think it can, fundamentally, because we can't force clients to cooperate in the metadata leak.

A non-cooperating client can always:

1. Implement support for `content_token` within the encrypted payload.
2. Upload the media.
3. Reference it from one or more junk catch-all rooms in the cleartext, to ensure the server considers it referenced.
4. Put `content_token` in the encrypted payload for all real references.

richvdh: > Implement support for content_token within the encrypted payload.

Well, that won't help the non-cooperating client if it wants to talk to cooperating clients.

richvdh: Guess so.

> Reference it from one or more junk catch-all rooms in the cleartext, to ensure the server considers it referenced.

this won't help unless both rooms have similar servers in, though?

... Anyway, I think MSC3911 was a better idea than this msc. It's probably not worth spending too much time on.

richvdh:
I'm going to close this as I think MSC3911/MSC3912 is a better solution.

PR #3911, opened 2022-10-16
MSC3911: Linking media to events

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/linking_media_to_events/proposals/3911-linking-media-to-events.md)


proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid)            | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}`            | `GET /_matrix/federation/v1/media/download/{serverName}/{mediaId}`  |
+   | [`GET /_matrix/media/v3/download/{serverName}/{mediaId}/{fileName}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3downloadservernamemediaid) | `GET /_matrix/client/v1/media/download/{serverName}/{mediaId}/{fileName}` | N/A                                                                 |
+   | [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3thumbnailservernamemediaid)          | `GET /_matrix/client/v1/media/thumbnail/{serverName}/{mediaId}`           | `GET /_matrix/federation/v1/media/thumbnail/{serverName}/{mediaId}` |
+
+   [TODO: we should probably move `/config`, and `/preview_url` while
+   we're at it]
```
richvdh, line 87: I think this MSC actually needs splitting into two:

 * one which moves all of the media endpoints *except* `/upload` to `/_matrix/{client,federation}`, and requires `Authorization` headers on them, but without worrying about what the requested media is
 * one which does the attachment to events/profiles

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+   The new media item can be attached to a new event, and generally functions
+   in every way the same as uploading a brand new media item.
+
+   This "copy" api is to be used by clients when forwarding events with media
+   attachments.
+
+5. Autogenerated `m.room.member` events
+
+   Servers will generate `m.room.member` events with an `avatar_url` whenever
+   one of their users joins a room, or changes their profile picture.
```
ShadowJonathan, line 173: A concern for what this might do in practice, is that older (or even newer) servers might download many many copies of the same backing file, if i'm reading this correctly.

If I am, it sounds like every single room will have different avatar URLs, requiring the server to fetch effectively the same file over and over again, if the user jumps between different rooms seeing the same chat members across multiple rooms.

This could bloat the "remote media" cache with many copies of the same media, which might be a concern.

richvdh: this is mentioned under the "Issues":

> * Since each `m.room.member` references the avatar separately, changing your
>  avatar will cause an even bigger traffic storm if you're in a lot of rooms.

ShadowJonathan: Not technically, it mentions a traffic storm, but not the duplication of media within servers' media repositories.

richvdh: fair, can you add that as a comment to the "issue" please.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+
+We also need to ensure that the icons are not deleted from the content
+repository even though they have not been attached to any event or profile. It
+would be wise for servers to provide administrators with a mechanism to upload
+media without attaching it to anything.
+
+(This was previously discussed in
+[MSC2858](https://github.com/matrix-org/matrix-spec-proposals/pull/2858#discussion_r543513811).)
+
+## Potential issues
```
ShadowJonathan, line 281: Additionally, due to the way that this changes member events and forwarded media, it might bloat the "remote media" cache on a local server, if that server implementation has a relatively simple media caching solution (mapping `mxc` -> blob in a local directory), without local deduplication (which, even if it is proposed as a solution, still might not scale).

---

_Re-stated from [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/3911#discussion_r996799838)_

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+   that is an implementation detail). 
+
+   The response is a json object with a required `content_uri` property,
+   giving a new MXC URI referring to the media.
+
+   The new media item can be attached to a new event, and generally functions
+   in every way the same as uploading a brand new media item.
+
+   This "copy" api is to be used by clients when forwarding events with media
+   attachments.
```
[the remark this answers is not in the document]
richvdh: If we allow multiple attachments to a single media item, then the list of "events with this media" can grow over time.

Which means that a server with a cached media item, cannot tell if it should return it to a user without checking all the servers in the federation that might possibly have added the media to a new event.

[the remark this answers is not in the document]
deepbluev7: Ah, because a remote server could add events to the media, makes sense, thanks!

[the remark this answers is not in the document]
richvdh: well, anyone could add events to the media. Even if you restrict it so that only the server that first created the media can add new events to it, any remote servers with the cached media still have to check back with the origin server for every single request in case the list has changed.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+* There is currently no way to delete
+  media. [matrix-spec#226](https://github.com/matrix-org/matrix-spec/issues/226)
+* If a user requests GDPR erasure, their media remains visible to all.
+* When all users leave a room, their media is not deleted from the server.
+
+## Proposal
+
+### Overview
+
+After an item of media is uploaded, it must be linked to an event (via
```
uhoreg, line 24: For MLS, we may need to reference media within a to-device event, but we might be able to link it to an in-room event.  I'll have to look into this more, but flagging this as something that will need to be considered.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+   the user that uploaded it). This is referred to as a "restricted" media item.
+
+   The existing endpoint is deprecated. Media uploaded via the existing endpoint
+   is "unrestricted".
+
+2. Attaching media
+
+   * The methods for sending events
+     ([`PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+     and [`PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid))
```
Half-Shot, line 43: Copy paste error, one of these is meant to be the non-state endpoint?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,351 @@
...
+https://<bridge_server>/media/{originServerName}/{mediaId}?mac={hmac}
+```
+
+When the bridge later receives a request to that URI, it checks the hmac,
+and proxies the request to the homeserver, using its AS access
+token in the `Authorization` header.
+
+This mechanism also works for a secondary use of the content repository in
+bridges: as a paste-bin. In this case, the bridge simply generates a link
+referencing its own media.
```
richvdh, line 231: wait, if the media gets uploaded without being attached to a matrix event, it won't be visible, unless we provide some other mechanism to make it so

proposals/3911-linking-media-to-events.md
richvdh: yes :). The link looks right, the text not.

Munna231 (changes requested):
- [ ] #

richvdh:
I've moved a bunch of the changes here out to MSC3716, in the hope of making them a little more digestible.

deepbluev7:
> I've moved a bunch of the changes here out to [MSC3716](https://github.com/matrix-org/matrix-doc/pull/3716), in the hope of making them a little more digestible.

You meant MSC3916 (if anyone else has issues following that links :D)

richvdh:
> > I've moved a bunch of the changes here out to [MSC3716](https://github.com/matrix-org/matrix-doc/pull/3716), in the hope of making them a little more digestible.
> 
> You meant [MSC3916](https://github.com/matrix-org/matrix-doc/pull/3916) (if anyone else has issues following that link :D)

Thank you, yes I did :facepalm:. Edited.

proposals/3911-linking-media-to-events.md
richvdh: added some text to the doc about this.

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
richvdh: Added a clarification about this in the doc.

proposals/3911-linking-media-to-events.md
turt2live: there's also ephemeral-ish things like user profiles which aren't *directly* related to events, though are duplicated/copied to membership events in most cases.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+Fixes [synapse#1263](https://github.com/matrix-org/synapse/issues/1263).
+
+## Potential issues
+
+* Since each `m.room.member` references the avatar separately, changing your
+  avatar will cause an even bigger traffic storm if you're in a lot of rooms.
+
+  In addition, this could cause duplication of media in the remote media cache,
+  if the implementation does not take steps to deduplicate (eg, storing media
+  by content hash rather than media id).
```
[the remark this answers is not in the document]
turt2live: De-duplication is already possible at an implementation level, and would probably be handled by a different MSC.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+   Conceptually, the API makes a new copy of a media item. (In practice, the
+   server will probably make a new reference to an existing media item, but
+   that is an implementation detail). 
+
+   The response is a json object with a required `content_uri` property,
+   giving a new MXC URI referring to the media.
+
+   The new media item can be attached to a new event, and generally functions
+   in every way the same as uploading a brand new media item.
```
turt2live, line 146: Can the user copy restricted media they don't have access to? (I assume so, but what is the appropriate error code?)

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+#### Redacting events
+
+Under this proposal, servers can determine which media is referenced by an
+event that is redacted, and add that media to a list to be cleaned up.
+
+This would also apply if all users in a room are deactivated (either via a GDPR
+section 17 request or by a self-requested "Deactivate account" request). In
+this case, all events in the room, and all media referenced by them, should be
+removed. Currently, Synapse does not support removing the events (see also
+[synapse#4720](https://github.com/matrix-org/synapse/issues/4720)); but if at
```
turt2live, line 222: Synapse was fixed a few days ago, apparently :)

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+   The
+   [`/preview_url`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3preview_url)
+   endpoint returns an object that references an image for the previewed
+   site.
+
+   It is expected that servers will continue to treat such media as unrestricted
+   (at least for local users), but it would be legitimate for them to, for example,
+   return a different `mxc:` URI for each requesting user, and only allow each user
+   access to the corresponding `mxc:` URI.
```
turt2live, line 180: I'm not sure I see the advantage of a server issuing distinct URIs for previews. Is there a motivation for putting this here?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+  provide it, per [MSC3910](https://github.com/matrix-org/matrix-spec-proposals/pull/3910).
+
+## Security considerations
+
+* Letting servers track the relationship between events and media is a metadata
+  leak, especially for e2ee rooms.
+
+## Unstable prefix
+
+TODO
```
turt2live, line 254:
```suggestion
While this MSC is not considered stable, implementations should use the following mapped values.

| Stable | Unstable |
|-|-|
| `/_matrix/client/v1/media/upload` | `/_matrix/client/unstable/org.matrix.msc3916/media/upload` |
| `/_matrix/client/v1/media/create` | `/_matrix/client/unstable/org.matrix.msc3916/media/create` | 
| `attach_media` | `org.matrix.msc3916.attach_media` |
| `restrictions` | `org.matrix.msc3916.restrictions` (`event_id` and `profile_user_id` are not prefixed) |
| `/_matrix/client/v1/media/copy/:serverName/:mediaId` | `/_matrix/client/unstable/org.matrix.msc3916/media/copy/:serverName/:mediaId` |
```

turt2live: (this is what MMR will be using)

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+     are extended to take a query parameter `attach_media`, whose value must be a complete `mxc:` URI.
+
+     The `attach_media` parameter may be used several times to attach several
+     pieces of media to the same event. The maximum number of pieces of media
+     that can be attached to a single event is implementation-defined by servers.
+
+     If any of the `attach_media` parameters do not correspond to known
+     restricted media items, or they refer to restricted media items that have
+     already been attached, the server responds with a 400 error with
+     `M_INVALID_PARAM`.
```
turt2live, line 58: Should there also be a limitation that one can only attach media that they uploaded?

proposals/3911-linking-media-to-events.md
richvdh: I'd say no, it's the same as `/download` and `/thumbnail`.  403 and `M_UNAUTHORIZED`.

proposals/3911-linking-media-to-events.md
richvdh: I think I was trying to reason about how we might apply access restrictions to media that is part of a URL preview (if we decided that was a sensible thing to do). It may well be that the extra text confuses more than it clarifies.

proposals/3911-linking-media-to-events.md
richvdh: erm... good question.

I think the answer is "yes, there should", to guard against the possibility that someone gets hold of the `mxc:` URI between it being uploaded and the media being attached to an event. (In which case, they could attach it to their own event and get access to the media.) Particularly in the case of `/create` rather than `/upload`, in which the `mxc:` URI would appear in access logs.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   that is an implementation detail). 
+
+   The response is a json object with a required `content_uri` property,
+   giving a new MXC URI referring to the media.
+
+   The new media item can be attached to a new event, and generally functions
+   in every way the same as uploading a brand new media item.
+
+   This "copy" api is to be used by clients when forwarding events with media
+   attachments.
```
richvdh, line 149: We'd need to use this for [`m.sticker`](https://spec.matrix.org/v1.8/client-server-api/#msticker) events too.

turt2live: (and custom emoji, when that exists)

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+   --gc0p4Jq0M2Yt08jU534c0p
+   Content-Type: text/plain
+
+   This media is plain text. Maybe somebody used it as a paste bin.
+
+   --gc0p4Jq0M2Yt08jU534c0p
+   ```
+
+5. New "media copy" API
```
turt2live, line 132: going down the rabbit hole of custom emoji, there's an interaction with [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027) we will have to figure out.

AndrewRyanChama:
How does this interact with m.replace?  Will the client be expected to /copy all of the medias or will it be handled automagically?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+section 17 request or by a self-requested "Deactivate account" request). In
+this case, all events in the room, and all media referenced by them, should be
+removed. Currently, Synapse does not support removing the events (see also
+[synapse#4720](https://github.com/matrix-org/synapse/issues/4720)); but if at
+some point in the future this is added, then this proposal makes it easy to
+extend to deleting the media.
+
+Fixes [synapse#1263](https://github.com/matrix-org/synapse/issues/1263).
+
+## Potential issues
```
turt2live, line 228: @AndrewRyanChama says:

> How does this interact with m.replace? Will the client be expected to /copy all of the medias or will it be handled automagically?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+We need to ensure that the icons are not deleted from the content
+repository even though they have not been attached to any event or profile. It
+would be wise for servers to provide administrators with a mechanism to upload
+media without attaching it to anything.
+
+#### Redacting events
+
+Under this proposal, servers can determine which media is referenced by an
+event that is redacted, and add that media to a list to be cleaned up.
```
AndrewRyanChama, line 216: The problem with this is that if the user who sent the media leaves the room, the homeserver may no longer get the reaction events.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   An example response:
+
+   ```
+   Content-Type: multipart/mixed; boundary=gc0p4Jq0M2Yt08jU534c0p
+
+   --gc0p4Jq0M2Yt08jU534c0p
+   Content-Type: application/json
+
+   { "restrictions": {
+       "event_id": "$Rqnc-F-dvnEYJTyHq_iKxU2bZ1CI92-kuZq3a5lr5Zg"
```
tenpura-shrimp, line 121: This means that the association between (encrypted) media and event ids will be plainly accessible to the homeserver, and federated servers.  I'd imagine that isn't something we actually care about but something that others may complain about.

AndrewRyanChama, line 121: This means that the association between (encrypted) media and event ids will be plainly accessible to the homeserver, and federated servers. I'd imagine that isn't something we actually care about but something that others may complain about.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+adds authentication to media download), to require that the authenticated
+user is *authorised* to access the requested media.
+
+## Proposal
+
+### Overview
+
+After an item of media is uploaded, it must be linked to an event (via
+parameters to the `/send` api). A given piece of media is only visible
+to a user if the user can see the corresponding event.
```
AndrewRyanChama, line 31: Since reference counting is required anyways, I feel like it's better if the server does it automatically.  I wrote a short doc on that here: https://github.com/matrix-org/matrix-spec-proposals/pull/4086

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
rom4nik: > De-duplication is already possible at an implementation level

I think that reivilibre wasn't talking about the way clients/homeservers are deduplicating files *after* download, but rather *before* they're downloaded. I agree that former is an implementation detail that doesn't require uniformity across software, and as such doesn't need to be covered in protocol spec. But if the intention here is to avoid the very process of downloading a file multiple times (from a different linked MXC each time), can this be done without protocol declaring one well known way to discover e.g. that hash of underlying file by HEAD request?

I'm specifically thinking of a situation where you've got a user looking at sticker selector and not-so-maliciously spamming them by clicking them 100s of times. Some people do that, it isn't really a bad behavior in some rooms and contexts (slightly distant example would be custom emote spam on Twitch chats). Since the stickers sent in each event are no longer sharing the same MXC, AFAIK client (and homeserver, if sender is remote) will have to download the sticker 100s of times, and thus generate a lot of unnecessary traffic. Only then it will be able to deduplicate file for local storage/cache. 

So my (and I assume reivilibre's) concern here is that with the proposal as it is right now, there's no way for client/homeserver to avoid that "traffic duplication", so to speak. Or is there? 😅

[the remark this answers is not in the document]
rom4nik: Hm, on the other hand, that choice of hash type used for deduplication before download, might eventually *have an effect* on future implementations of dedupe after download. For example, iirc currently in Synapse there's a mapping of MXC URI to local filename which is a random string. With this MSC, each server would have to store a mapping from the new generated IDs to the local file (in Synapse's case, again, a random name). If protocol adds a way to map MXC IDs to file hash early, is there a reason for servers to not store files named after their hashes (as received over e.g. federation)? This would save you one more lookup: MXC IDs -> file hash -> randomized local name.

Oh, and I haven't thought how/if this "early dedupe" could realistically work with files attached to E2EE events, and whether it would add any additional metadata leak other than what's caused by this MSC as is.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916) are
+   updated the server must check if the requesting user or server has
+   permission to see the corresponding event.  If not, the server responds with
+   a 403 error and `M_UNAUTHORIZED`.
+
+4. Federation API returns a `restrictions` property
+
+   The `/_matrix/federation/v1/media/download` and `/_matrix/federation/v1/media/thumbnail`
+   endpoints specified by [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916)
+   are extended: the returned json object may have a property `restrictions`.
```
Kladki, line 92: MSC3916 [doesn't specify a federation `/thumbnail` endpoint](https://github.com/matrix-org/matrix-spec-proposals/pull/3916/files#diff-6d5b43335cc0670d70d807b13c0ad9f5526eb97d64c5d2a427e27ef695c1be50R51-R53).

turt2live: there are several updates required to this MSC - the priority has been to get 3916 into working order, then the team will move onto this MSC afterwards.

marrobHD: MSC3916 was merged :partying_face:

proposals/3911-linking-media-to-events.md
HarHarLinks, line 228: please consider that widgets also use media, i.e. https://github.com/matrix-org/matrix-spec-proposals/pull/4039

to summarize, linking media is not a big issue for events of a well-known structure referencing them - with regards to widgets, that means that clients can parse the payload and handle it automatically. however especially for widgets it is often attractive to send nonstandard events (let alone ones with some custom encoding), that prohibit this.

there can be use cases (with widgets or without), where multiple media are attached to an event that represents a certain state at a point in time and which is edited collaboratively. however editing would usually involve changing only one of the referenced media items at most, resulting in an immense overhead when having to `copy` all the other attached media as well when sending an update ([example implementation](https://github.com/nordeck/matrix-neoboard/blob/main/docs/adrs/adr005-data-structure-for-storing-whiteboards-in-matrix-rooms.md)).

HarHarLinks (changes requested):
reviewing with my Nordeck hat on.

proposals/3911-linking-media-to-events.md
Johennes: If the server doesn't have a local user in the room anymore, I suppose it should remove the room, the events and any linked media anyway?

proposals/3911-linking-media-to-events.md
Johennes: If the replacing event contains the same media, `/copy` seems reasonable to me because it keeps with the idea of attaching media to a single event only.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   The new media item can be attached to a new event, and generally functions
+   in every way the same as uploading a brand new media item.
+
+   This "copy" api is to be used by clients when forwarding events with media
+   attachments.
+
+   (This mechanism, rather than just allowing clients to attach media to multiple
+   events, is necessary to ensure that the list of events attached a given piece of
+   media does not grow over time, which would make it difficult for servers to
+   reliably cache media and impose the correct access restrictions.)
```
timokoesters, line 154: Can you explain why this would be a problem? Don't servers have to do this anyway if they store copies as references?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   [`/preview_url`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixmediav3preview_url)
+   endpoint returns an object that references an image for the previewed
+   site.
+
+   It is expected that servers will continue to treat such media as unrestricted
+   (at least for local users), but it would be legitimate for them to, for example,
+   return a different `mxc:` URI for each requesting user, and only allow each user
+   access to the corresponding `mxc:` URI.
+
+### Applications
```
richvdh, line 182: Encrypted-history sharing (currently in progress, but tracked at https://github.com/element-hq/element-meta/issues/39, MSC4268) adds another application for media blobs which are not tied to a specific event. It might be that a separate mechanism which creates media specifically for sharing with one other user/device would help.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   * Alternatively, if a restricted media item is referenced in a call to
+     [`PUT /_matrix/client/v3/profile/{userId}/avatar_url`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3profileuseridavatar_url)
+     it is instead attached to the user's profile. 
+
+     Again, if the media is already attached, the server responds with a 400 error with
+     `M_INVALID_PARAM`.
+
+   If the media is not attached to either an event or a profile within a reasonable period
+   (say, ten minutes), then the server is free to assume that the user has changed their
+   mind (or the client has gone offline), and may clean up the uploaded media.
```
Johennes, line 78: "may clean up" means servers could decide to not do so. This would create dangling media that are not viewable by other users but still stored on the server which, I think, could still be an issue with respect to GDPR erasure. Should we make clean-up RECOMMENDED at least?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+3. Additional checks on `/download` and `/thumbnail` endpoints
+
+   The new `/download` and `/thumbnail` endpoints added in
+   [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916) are
+   updated the server must check if the requesting user or server has
+   permission to see the corresponding event.  If not, the server responds with
```
Johennes, line 85: The spec allows servers to restrict profile look-ups (the minimum being users that share rooms). A server that does this should probably also apply this check when user avatars are downloaded?

```suggestion
   permission to see the corresponding event or profile.  If not, the server responds with
```

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
kegsay: > a way to map MXC IDs to file hash early

This makes a lot of sense to me. There's no real benefit in making up garbage strings to then use as the MXC ID. This was useful pre-authorisation because it meant you could hypothetically drop references to the same piece of content which was used elsewhere, but if we're adding in authorisation for that link then there's little reason to keep the random MXC IDs around imo.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+removed. Currently, Synapse does not support removing the events (see also
+[synapse#4720](https://github.com/matrix-org/synapse/issues/4720)); but if at
+some point in the future this is added, then this proposal makes it easy to
+extend to deleting the media.
+
+Fixes [synapse#1263](https://github.com/matrix-org/synapse/issues/1263).
+
+## Potential issues
+
+* Since each `m.room.member` references the avatar separately, changing your
```
benkuly, line 230: This looks like a big issue to me, that should be adressed in this MSC. From a client perspecive it means, that a simple user avatar change would force clients to download the same file n times. Why not just allow multiple restrictions per mxc uri?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+   The existing endpoint is deprecated. Media uploaded via the existing endpoint
+   is "unrestricted".
+
+2. Attaching media
+
+   * The methods for sending events
+     ([`PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+     and [`PUT /_matrix/client/v3/rooms/{roomId}/send/{eventType}/{txnId}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid))
+     are extended to take a query parameter `attach_media`, whose value must be a complete `mxc:` URI.
```
kegsay, line 49: I disagree with this. We're going to be hitting URL length limits, query escaping issues, etc if we dump MXC URIs as query params. I would be much more happy if we just used the [media id](https://spec.matrix.org/v1.14/client-server-api/#matrix-content-mxc-uris) component, given the server name and `mxc://` are redundant. This makes it:
 - shorter
 - less likely to cause escaping issues

..at the cost of needing clients to split the MXC URI, though clients [already need to do some amount of introspection](https://spec.matrix.org/v1.14/client-server-api/#put_matrixmediav3uploadservernamemediaid) of the URI so this feels fine.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   (say, ten minutes), then the server is free to assume that the user has changed their
+   mind (or the client has gone offline), and may clean up the uploaded media.
+
+3. Additional checks on `/download` and `/thumbnail` endpoints
+
+   The new `/download` and `/thumbnail` endpoints added in
+   [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916) are
+   updated the server must check if the requesting user or server has
+   permission to see the corresponding event.  If not, the server responds with
+   a 403 error and `M_UNAUTHORIZED`.
```
kegsay, line 86: > the server must check if the requesting user or server has **permission to see** the corresponding event. If not, the server responds with a 403 error and `M_UNAUTHORIZED`.

This needs clarification. It's either:
 - based on history visibility, or
 - based on current room membership

I would advocate for it to be based on _current room membership_, not history visibility because:
 - history visibility makes no sense for room avatars, per-room profile avatars, etc which are _room state_. It would mean newly joined users would be unable to see the room avatar / etc because the room avatar event was sent prior to their join.
 - history visibility is not enforceable over federation, providing a false sense of security. At a federation level we rely on simply "was there at least 1 user on your server joined at that time", which is not the same as "were _you_ joined at that time". 
 - history visibility is hard to cache, as concurrent changes to that event can cause previously visible events to become hidden or vice versa. This is hard to explain to clients.
 - history visibility interacts poorly with E2EE rooms, where "visibility" is not just whether you see the event JSON, but whether you have the keys to decrypt said JSON. This poor interaction means you could be sent event JSON which you can't decrypt, or not be sent event JSON which you have the keys to decrypt.

In comparison, _current room membership_:
 - works for room avatar and per-room profiles.
 - is much easier to cache as it doesn't rely on historical room state, and is easy to explain to clients.
 - doesn't interfere with E2EE as badly, since it's a more permissive approach (we err on the side of letting clients fetch media and if they don't have the key too bad, but we never end up in a situation where we fail to let clients fetch media they can legitimately decrypt).

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+     ([`PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+     and [`PUT /_matrix/client/v3/rooms/{roomId}/send/{eventType}/{txnId}`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid))
+     are extended to take a query parameter `attach_media`, whose value must be a complete `mxc:` URI.
+
+     The `attach_media` parameter may be used several times to attach several
+     pieces of media to the same event. The maximum number of pieces of media
+     that can be attached to a single event is implementation-defined by servers.
+
+     If any of the `attach_media` parameters do not correspond to known
+     restricted media items, or they refer to restricted media items that have
```
kegsay, line 56: > If any of the `attach_media` parameters do not correspond to known restricted media items, **or they refer to restricted media items that have already been attached**, the server responds with a 400 error

This enforces a 1:1 relationship between MXC ID and event ID. This is problematic, and is acknowledged throughout this proposal:

 - > Since each m.room.member references the avatar separately, changing your avatar will cause an even bigger traffic storm if you're in a lot of rooms.
 - > this could cause duplication of media in the remote media cache, if the implementation does not take steps to deduplicate (eg, storing media by content hash rather than media id).
 - The copy API entirely.
 - The inability for clients to cache repeated content e.g stickers, resulting in suggestions like adding HTTP headers which contain the content hash.

It is poorly justified why it is worth it to keep a 1:1 mapping despite all these issues. The only relevant comment I see is justification for the copy API:

> (This mechanism, rather than just allowing clients to attach media to multiple events, is necessary to ensure that the list of events attached a given piece of media does not grow over time, which would make it difficult for servers to reliably cache media and impose the correct access restrictions.)

which doesn't make sense to me. In fact, the _inability_ to attach the same media to multiple events makes it hard to cache, hence people asking for a content hash.

To me, it makes far more sense to say:
 - The media ID in the MXC URI _is a content hash_ and we spec it so clients can introspect it.
 - When you `?attach_media=` a row is added in an `attached_media` table which has `event_id` and `content_hash` columns.  This table is `UNIQUE(event_id, content_hash)`, and indexed on both `event_id` and `content_hash` independently.
 - This gives you a 1:many mapping for content hashes, whilst still allowing servers to "impose the correct access restrictions".
 - Rows in this table will grow over time if there are genuinely more events referencing the same piece of content.
 - Illegal material can be nuked by content hash.
 - GDPR erasure can be achieved by `JOIN`ing on `event_id` to the event's `sender` and deleting those rows.
 - All content in a room can be deleted by `JOIN`ing on `event_id` to the event's `room_id` and deleting those rows.
 - Content can be deleted if there exists no content hash in this table.

It all just makes life significantly easier imo. The only reason I can think of why you wouldn't just do this is if you've outsourced your media up/downloading to a 3rd party who doesn't provide a way to add custom code. If you can only nuke individual URLs then that becomes your poor-man's "deleting rows from the attached media table". Whilst that may make it easier for a specific server setup, it negatively impacts everyone else who can no longer de-duplicate in a sane way.

proposals/3911-linking-media-to-events.md
TheArcaneBrony, line 146: How does this address cases where ie. I *intentionally* want to reuse the same media item for client deduplication reasons, ie. reusing the same 2MB room icon across all 100 rooms in my space? Could this be changed to some way where a client can *inherently* (without extra API calls) know it's downloading the same item 100 times for a /hierarchy listing or similar?
Similar case for custom emojis, where you might over time end up with thousands if not millions of copies of the same exact media item.

Idea: can we pass the event ID to the download/thumbnail endpoints instead, so it doesn't have to completely change semantics and break existing client media storage optimisations?

noelportillo (approved):

proposals/3911-linking-media-to-events.md
nico-famedly: Yes, the mxc:// prefix seems unnecessary, especially since clients use the mxc uri without the prefix for downloads anyway...

proposals/3911-linking-media-to-events.md
nico-famedly: current room membership wouldn't strictly be sufficient either. You should be able to download the avatar for a room, you want to join or see the avatar of a user, you want to invite, if you want to provide a good UX (although certain federations might want to limit this).

The problem with history visibility for me primarily is, that clients might get sent state events, that they can't see according to history visibility rules. For example your .state in a sync response will contain the current room members and avatar, but you will never see it in the timeline, if the room has history visibility set to "invite" and you weren't invited to the room, when that state event was sent.

For that reason we need somewhat of a mixed ruleset, that also covers current state events at any point of the room, where the user would have been able to see any other event as well as covers stripped state for potential joiners.

Current room membership would be too simplistic imo. It would allow users to download old media, just because they joined the room now. This doesn't feel right in security sensitive contexts, even if the problems with it would likely be minimal.

proposals/3911-linking-media-to-events.md
nico-famedly: How would you update the restrictions list on a remote server then? I think the reason why a 1:1 mapping was chosen, was because the remote media will currently never refetch the restrictions dictionary. To make it do that, we would likely need to attach the mxc uri in plain text with every event, so that servers could be notified, that their restrictions are out of date or similar. Such an approach would be quite fragile, imo and possibly leak even more metadata.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
+# MSC3911: Linking media to events
+
+(An alternative to [MSC3910](https://github.com/matrix-org/matrix-spec-proposals/pull/3910).)
+
+Currently, access to media in Matrix has the following problems:
+ 
+* The only protection for media is the obscurity of the URL, and URLs are
+  easily leaked (eg accidental sharing, access
+  logs). [synapse#2150](https://github.com/matrix-org/synapse/issues/2150)
+* Anybody (including non-matrix users) can cause a homeserver to copy media
```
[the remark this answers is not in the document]
nico-famedly: (I assume appservices and such are covered by this as well)

proposals/3911-linking-media-to-events.md
nico-famedly: Copying over all media might become somewhat expensive. Especially when using inline images you might need  to do a lot of network calls to send an edit now. I don't really see a good way around it right now though.

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
nico-famedly: We could also look into using the Repr-Digest: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Repr-Digest. It has some peculiarities around requests though, so a X-Matrix-Media-Hash might be better.

It would still require an extra HEAD request per media, but a lot better than downloading every file multiple times. Maybe in the future we can even include that in message metadata to skip network requests, especially in e2ee contexts. There are some risks regarding cache poisoning though.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+
+   * Alternatively, if a restricted media item is referenced in a call to
+     [`PUT /_matrix/client/v3/profile/{userId}/avatar_url`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3profileuseridavatar_url)
+     it is instead attached to the user's profile. 
+
+     Again, if the media is already attached, the server responds with a 400 error with
+     `M_INVALID_PARAM`.
+
+   If the media is not attached to either an event or a profile within a reasonable period
+   (say, ten minutes), then the server is free to assume that the user has changed their
```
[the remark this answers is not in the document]
nico-famedly: Imo this period should also be the same as for async uploads for simplicity (24hours). It is a similar use case and I see no reason why it needs to be a lot shorter.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   [`/_matrix/media/v3/upload`](https://spec.matrix.org/v1.4/client-server-api/#post_matrixmediav3upload)
+   endpoint, but media uploaded this way is not initially viewable (except to
+   the user that uploaded it). This is referred to as a "restricted" media item.
+
+   The existing endpoint is deprecated. Media uploaded via the existing endpoint
+   is "unrestricted".
+
+2. Attaching media
+
+   * The methods for sending events
```
nico-famedly, line 46: What about to_device messages and account data? Should those be supported? (I.e. adding personal avatar overrides for a room, widgets, etc)

nico-famedly, line 46: What about room creation as well as upgrades? How do they deal with room avatars, rich text topics, etc?

nico-famedly: (Also invites?)

proposals/3911-linking-media-to-events.md
nico-famedly, line 254: It would be good to have a feature flag for this on the server side, so that clients can enable support.

proposals/3911-linking-media-to-events.md
Johennes: I think the account data case might not need any treatment because account data is only visible to yourself and restriced media are always visible to the uploader?

Johennes: Oh wait, unattached media can be cleaned up. So some form of attachment would be needed for account data, too, then.

nico-famedly: Exactly. The problem is it getting cleaned up :D

proposals/3911-linking-media-to-events.md
Johennes: I think this was explained in [this resolved thread](https://github.com/matrix-org/matrix-spec-proposals/pull/3911/files#r997026590) but wasn't reproduced in the proposal text as clearly.

noelportillo (commented):
Ok I think it's the best way

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+   This media is plain text. Maybe somebody used it as a paste bin.
+
+   --gc0p4Jq0M2Yt08jU534c0p
+   ```
+
+5. New "media copy" API
+
+   A new endpoint is defined: `POST
+   /_matrix/client/v1/media/copy/{serverName}/{mediaId}`. The body of the
+   request must be a JSON object, but there are no required parameters.
```
nico-famedly, line 136: Should this have the `timeout_ms` parameter from the download API? Or should this only allow copying media, that is already stored locally?

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+     Servers should ensure that sending an event remains an idempotent operation: in
+     particular, if a client sends an event with a media attachment, and then
+     repeats the operation with identical parameters, a 200 response must be returned
+     (with the original event ID) even though the media has already been attached.
+
+   * Alternatively, if a restricted media item is referenced in a call to
+     [`PUT /_matrix/client/v3/profile/{userId}/avatar_url`](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3profileuseridavatar_url)
+     it is instead attached to the user's profile. 
+
+     Again, if the media is already attached, the server responds with a 400 error with
```
nico-famedly, line 73: What if the media is already attached and the user has no access to it? Shouldn't it return 404 then?

proposals/3911-linking-media-to-events.md
nico-famedly, line 86: Other authenticated media endpoints return 404, when it isn't accessible to prevent metadata enumeration. Should this really return 403?

proposals/3911-linking-media-to-events.md
nico-famedly, line 149: So clients should or should not copy media on edits? In theory it should only be necessary for new media, since edits are usually anchored where the original event was, but it could break the edit fallback in some clients.

proposals/3911-linking-media-to-events.md
nico-famedly: Strictly speaking you probably don't need to do a copy on an edit, if you ignore the fallback in clients not supporting edits. The old media should be visible to everyone, who saw the original event, so you could still reference it in theory, even if not ideal.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+* Use some sort of "content token" for each piece of media, and require clients to
+  provide it, per [MSC3910](https://github.com/matrix-org/matrix-spec-proposals/pull/3910).
+
+## Security considerations
+
+* Letting servers track the relationship between events and media is a metadata
+  leak, especially for e2ee rooms.
+
+## Unstable prefix
+
```
nico-famedly, line 253: Synapse implementation: https://github.com/famedly/synapse/pull/131
Client implementation in the matrix dart sdk: https://github.com/famedly/matrix-dart-sdk/pull/2134

The implementation needs basically no client side changes apart from updating the SDK and passing messages through `Event.copyMediaInContent` when forwarding events. That's actually kinda neat, since it means the media changes can be somewhat hidden in an SDK for the most part, significantly simplifying the implementation experience.

One open question is how to handle media when editing, especially inline media. In theory the media wouldn't have to be copied, since the edit would be visible to anybody who previously could see the event, but maybe there are arguments against that?

One downside of this MSC is, that it somewhat tightly couples media repo and the homeserver, so stuff like the matrix-media-repo might need additional server specific APIs.

(Disclosure: This implementation was done for the Gematik for the German TI-Messenger)

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
reivilibre: A HEAD request wouldn't necessarily be required, it would be also fine to start a GET and abort as soon as receiving the headers.

However I am happy to accept that this would be fine as a subsequent MSC without any particular risk in doing so, as far as I can tell.

proposals/3911-linking-media-to-events.md
reivilibre: This seems fundamentally impossible to work around, but good to spell it out loud

As for `/copy`: a `/copy` of encrypted media will also be obvious because the content will be the same across two pieces of media.

proposals/3911-linking-media-to-events.md
reivilibre: Good point. I think this will be OK to just acknowledge on the MSC for the time being; widgets could continue to use unrestricted media for the time being if it's a huge issue.

For the last use case, widgets could use dummy events to act as 'containers' for the linked/attached media. Not really ideal, but might be OK.

proposals/3911-linking-media-to-events.md
```diff
@@ -0,0 +1,281 @@
...
+remote network. This will become problematic, because servers are entitled
+to remove any media that does not get linked to an event.
+
+Solutions might include:
+ * the bridge hosting its own content repository for this usecase
+ * using an external service
+ * special-casing the bridge AS user to permit it to upload media without
+   expiry.
+
+#### Icons for "social login" flows
```
Saiv46, line 203: With MSC3861 this section became irrelevant. I never understanded why we should put social media icons in media repository, now it's on OIDC

Saiv46 (changes requested):

Saiv46:
Hi @richvdh, is there any progress or roadblocks of this proposal?

richvdh:
> Hi @richvdh, is there any progress or roadblocks of this proposal?

There hasn't been much progress that I'm aware of. I see there are some draft implementations (thank you @nico-famedly), but if I remember rightly, incompatibility with stickers is a bit of a road-block.

I think maybe @turt2live has some ideas in this area, but I don't know what they are

turt2live:
https://github.com/matrix-org/matrix-spec-proposals/pull/4396 is an attempt to figure out some of the roadblocks, but has other issues. 

The concept of media linking is currently with the Foundation's Trust & Safety team to figure out, though blocked by other work at the moment. Foundation T&S pushed the authenticated media dependency a couple of years ago.

proposals/3911-linking-media-to-events.md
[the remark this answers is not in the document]
Saiv46: We can re-use `ETag`s and its mechanism like `If-None-Match` conditional requests.

Make it something like `ETag: "org.matrix.msc3911/sha-256:5NdPJxfaZvKBSq5rRHB3svJ-YE_utPnyp8SF8_ASVPk"`

PR #3912, opened 2022-10-18
MSC3912: Redaction of related events

babolivier:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/babolivier/relation_redaction/proposals/3912-relation-based-redaction.md)

- Synapse implementation: 
   - https://github.com/matrix-org/synapse/pull/14260
   - https://github.com/matrix-org/synapse/pull/15705

Client side implementation:
- Web - JS SDK implementation:
   - https://github.com/matrix-org/matrix-js-sdk/pull/2954
   - https://github.com/matrix-org/matrix-js-sdk/pull/3420
- iOS SDK implementation:
   - https://github.com/matrix-org/matrix-ios-sdk/pull/1688
   - https://github.com/matrix-org/matrix-ios-sdk/pull/1793
- Android implementation:
   - https://github.com/vector-im/element-android/pull/7988
   - https://github.com/vector-im/element-android/pull/8532

### DIFF
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -100,14 +100,8 @@ If an event that matches the redaction criteria comes down federation after
 redaction requests has completed, the server must attempt to redact it on behalf
 of the redaction request's sender.
 
-Servers may wish to copy the `with_relations` property into the
-`m.room.redaction` event's `unsigned` object in order to facilitate watching for
-new events to redact once the redaction request has completed. Servers must
-ignore `with_relations` properties in the `unsigned` object of redaction events
-that were not sent by a local user.
-
-Since the response format to the `PUT
-/_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}` does not change as a
+Since the response format to `PUT
+/_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}` requests does not change as a
 result of this proposal, servers may respond to the request once the event
 referenced to by the `{eventId}` parameter is redacted, and redact events that
 relate to it in the background.
@@ -118,6 +112,17 @@ relate to it in the background.
 Homeservers which support this MSC should indicate it by adding
 `org.matrix.msc3912` and `org.matrix.msc3912.stable` in the response to `GET
 /_matrix/client/versions` requests.
+
+
+## Potential issues
+
+Since all the redactions are sent on behalf of the user who sent the original
+`PUT /_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}` request, if
+this user leaves the room the server is no longer able to send redactions on
+their behalf. However, the author considers this a small edge case which is
+unlikely to happen in practice, and can be worked around by another user (with
+permissions to delete the new events) sending another `PUT
+/_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}`.
 
 
 ## Alternatives
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -23,12 +23,16 @@ their other actions would also be prevented by rate limiters as a result).
 
 ## Proposal
 
-We add a new optional `with_relations` property to the body of the [`PUT
+We add a new optional `with_rel_types` property to the body of the [`PUT
 /_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}`](https://spec.matrix.org/latest/client-server-api/#put_matrixclientv3roomsroomidredacteventidtxnid)
 requests. This property is a list of relation types. If an event relates to the
 event specified in the `{eventId}` parameter with a relation type specified in
-this `with_relations` property, the server also redacts it on behalf of the user
+this `with_rel_types` property, the server also redacts it on behalf of the user
 performing the redaction. The response body of this endpoint does not change.
+
+The client may not be aware of the relation types associated with an event (for
+example in the case of moderation). That's why we introduce a catch-all `"*"` value,
+which if found in the list means "any relation type".
 
 For example, let's consider the following message:
 
@@ -70,7 +74,7 @@ Then the request:
 PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
 
 {
-    "with_relations": ["m.replace"]
+    "with_rel_types": ["m.replace"]
 }
 ```
 
@@ -81,7 +85,7 @@ A server may wish to leverage
 to redact all events targeted by the redaction request if supports it, however
 this is not a requirement.
 
-If the `with_relations` property is absent from the request body, only the event referred to by the `{eventId}` parameter (`$a` in the example above) is redacted. Same goes if `with_relations` is an empty list.
+If the `with_rel_types` property is absent from the request body, only the event referred to by the `{eventId}` parameter (`$a` in the example above) is redacted. Same goes if `with_rel_types` is an empty list.
 
 If an event that matches the redaction criteria (i.e. relates to the event
 that's being redacted with one of the relation types specified in the request)
@@ -109,7 +113,7 @@ a redaction of `$b` would not cause `$a` to be redacted. The following request:
 PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$b/foo
 
 {
-    "with_relations": ["m.replace"]
+    "with_rel_types": ["m.replace"]
 }
 ```
 
@@ -146,7 +150,7 @@ Then the request:
 PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
 
 {
-    "with_relations": ["m.replace"]
+    "with_rel_types": ["m.replace"]
 }
 ```
 
@@ -190,4 +194,4 @@ given they have more control over feature UX.
 ## Unstable prefixes
 
 Until this proposal is stabilised, `org.matrix.msc3912.with_relations` should be
-used instead of `with_relations`.
+used instead of `with_rel_types`.
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -1,4 +1,4 @@
-# MSC3912: Relation-based redaction
+# MSC3912: Redaction of related events
 
 There are cases where, when redacting an event, it would make sense to also
 redact all events that relate to it in a certain way. A few examples of this
@@ -37,11 +37,7 @@ For example, let's consider the following message:
     "type": "m.room.message",
     "content": {
         "body": "Hello",
-        "msgtype": "m.text",
-        "m.relates_to": {
-            "rel_type": "m.replace",
-            "event_id": "$some_event_id"
-        }
+        "msgtype": "m.text"
     },
     "event_id": "$a"
 }
@@ -106,6 +102,55 @@ result of this proposal, servers may respond to the request once the event
 referenced to by the `{eventId}` parameter is redacted, and redact events that
 relate to it in the background.
 
+This is important to note that we do not consider parent events. In the previous example,
+a redaction of `$b` would not cause `$a` to be redacted. The following request:
+
+```
+PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$b/foo
+
+{
+    "with_relations": ["m.replace"]
+}
+```
+
+Causes only event `$b` to get redacted.
+
+The redaction will be limited to the events which meet the validity requirements of the
+selected relationship types, for example [these requirements](https://spec.matrix.org/v1.6/client-server-api/#validity-of-replacement-events) for edited events.
+In case of the previous example, let's consider this additional event `$c`: 
+
+```json
+{
+    "type": "m.room.message",
+    "content": {
+        "body": "* Hello world2!",
+        "msgtype": "m.text",
+        "m.new_content": {
+            "body": "Hello world2!",
+            "msgtype": "m.text"
+        },
+        "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$b"
+        }
+    },
+    "event_id": "$c"
+}
+```
+
+This edit event is invalid because we must not edit an edit (We were supposed to refer to the
+original event: `$a`)
+Then the request:
+
+```
+PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
+
+{
+    "with_relations": ["m.replace"]
+}
+```
+
+Causes events `$a` and `$b` to get redacted, but no `$c`
 
 ### Unstable feature in `/version`
 
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -82,7 +82,7 @@ Causes events `$a` and `$b` to both get redacted.
 
 A server may wish to leverage
 [MSC2244](https://github.com/matrix-org/matrix-doc/pull/2244) (mass redactions)
-to redact all events targetted by the redaction request if supports it, however this is not a
+to redact all events targeted by the redaction request if supports it, however this is not a
 requirement.
 
 If the `with_relations` property is absent from the request body, only the event referred to by the `{eventId}` parameter (`$a` in the example above) is redacted. Same goes if `with_relations` is an empty list.
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -103,11 +103,14 @@ of the redaction request's sender.
 Servers may wish to copy the `with_relations` property into the
 `m.room.redaction` event's `unsigned` object in order to facilitate watching for
 new events to redact once the redaction request has completed. Servers must
-ignore `with_relations` properties in the `unsigned` object of redaction events that were not sent by a local user.
+ignore `with_relations` properties in the `unsigned` object of redaction events
+that were not sent by a local user.
 
 Since the response format to the `PUT
 /_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}` does not change as a
-result of this proposal, servers may respond to the request once the event referenced to by the `{eventId}` parameter is redacted, and redact events that relate to it in the background.
+result of this proposal, servers may respond to the request once the event
+referenced to by the `{eventId}` parameter is redacted, and redact events that
+relate to it in the background.
 
 
 ### Unstable feature in `/version`
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -80,7 +80,7 @@ PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
 
 Causes events `$a` and `$b` to both get redacted.
 
-If the `with_relations` property is absent from the request body, only the event refered to by the `{eventId}` parameter (`$a` in the example above) is redacted. Same goes if `with_relations` is an empty list.
+If the `with_relations` property is absent from the request body, only the event referred to by the `{eventId}` parameter (`$a` in the example above) is redacted. Same goes if `with_relations` is an empty list.
 
 If an event that matches the redaction criteria (i.e. relates to the event
 that's being redacted with one of the relation types specified in the request)
--- a/proposals/3912-relation-based-redaction.md
+++ b/proposals/3912-relation-based-redaction.md
@@ -156,7 +156,7 @@ PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
 
 Causes events `$a` and `$b` to get redacted, but no `$c`
 
-### Unstable feature in `/version`
+### Unstable feature in `/versions`
 
 Homeservers which support this MSC should indicate it by adding
 `org.matrix.msc3912` and `org.matrix.msc3912.stable` in the response to `GET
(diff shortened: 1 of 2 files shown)


proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,117 @@
...
+of the redaction request's sender.
+
+### Unstable feature in `/version`
+
+Homeservers which support this MSC should indicate it by adding
+`org.matrix.msc3912` and `org.matrix.msc3912.stable` in the response to `GET
+/_matrix/client/versions` requests.
+
+
+## Alternatives
```
clokep, line 105: I think another alternative is to define which relation types always get related events redacted. This has the limitation that generic things like `m.reference` can't really be handled well though.

🤷 Not sure it is worth mentioning, but that was what I initially assumed the solution would be upon reading the first few sentenences!

babolivier: Yeah, and also it doesn't allow as much flexibility in terms of UX/UI. For example, a client might want to show a check box to room moderators when deleting a root's thread to allow them to also delete the whole thread. I'll add a note.

clokep: Definitely. Thanks for considering! 👍

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,125 @@
+# MSC3912: Relation-based redaction
```
FSG-Cat, line 1: I personally think that this MSC could list MSC2244 as a soft dependency. 

What i mean by a soft dependency in this case is that this MSC is enhanced with mass redactions in the spec but is able to function without it.

babolivier: Yeah I wasn't pondering on that, I'll add a note for it.

babolivier: Done in https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/85b270a46201475ce006a778324ef82715f702af

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,143 @@
...
+delete other users' events) and display a warning to the user in this case.
+
+If an event that matches the redaction criteria comes down federation after
+redaction requests has completed, the server must attempt to redact it on behalf
+of the redaction request's sender.
+
+Servers may wish to copy the `with_relations` property into the
+`m.room.redaction` event's `unsigned` object in order to facilitate watching for
+new events to redact once the redaction request has completed. Servers must
+ignore `with_relations` properties in the `unsigned` object of redaction events
```
babolivier, line 106: Not sure if `unsigned` is the right place for the server to put it, or if it should just go in the event's `content`.

anoadragon453: If it's only useful to the local homeserver that sent it, then perhaps it shouldn't go in the event at all, but just tracked internally by the homeserver (in a database table, perhaps)?

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,143 @@
...
+their other actions would also be prevented by rate limiters as a result).
+
+## Proposal
+
+We add a new optional `with_relations` property to the body of the [`PUT
+/_matrix/client/v3/rooms/{roomId}/redact/{eventId}/{txnId}`](https://spec.matrix.org/latest/client-server-api/#put_matrixclientv3roomsroomidredacteventidtxnid)
+requests. This property is a list of relation types. If an event relates to the
+event specified in the `{eventId}` parameter with a relation type specified in
+this `with_relations` property, the server also redacts it on behalf of the user
+performing the redaction. The response body of this endpoint does not change.
```
anoadragon453, line 31: Are there/could there be a case where the client would like to redact one set of events with relation type X, but not another set of events with relation type X?

Essentially: are we sure that no group of two features or more will use the same relation type?

An alternative may be instead to either only - or additionally - specify event types you'd like to redact.

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,143 @@
...
+cannot be redacted by the user sending the redaction request (because of
+insufficient permissions), it is ignored. Clients may attempt to detect cases
+where a user would not be able to redact some events matching the redaction
+criteria (e.g. matching `m.reference` relations when redacting a `m.poll.start`
+event, matching `m.thread` relations, etc., with an insufficient power level to
+delete other users' events) and display a warning to the user in this case.
+
+If an event that matches the redaction criteria comes down federation after
+redaction requests has completed, the server must attempt to redact it on behalf
+of the redaction request's sender.
```
anoadragon453, line 101: Note that this will mean that anyone that tries to, say, respond to a thread that was redacted long in the past will have that message redacted immediately afterwards.

Also note that since only the homeserver that initially kicked off the redaction of the whole thread is responsible for redacting new related events; if that homeserver leaves the room then those messages will remain unredacted.

Just some edge cases to consider, but they might not warrant trying to fix.

proposals/3912-relation-based-redaction.md
babolivier: I added this paragraph after I tried implementing it in Synapse, which is quite non-trivial otherwise. The reason is that insertions to the redactions table is done in the event persistence path (where it's not trivial to add extra data which isn't part of the event), and the `/redact` endpoint can be handled by workers, which makes it potentially tricky for it to write into the database. So ideally it wouldn't be in the event at all, but I'm trying to balance that with not making it too difficult on server developers to implement it.

proposals/3912-relation-based-redaction.md
babolivier: I did a quick survey of the features using relations while writing the proposal. There are generic relation types (namely `m.reference`), but when it's used the semantics of the related messages can always be inferred by clients by looking at the type of the root event (`m.poll.start` for polls, `m.beacon_info` for live location sharing, etc). If that was not the case I believe there would be other annoying issues like clients not being able to figure out how to render an event. So I don't really see the point of also supporting also specifying event types. And it's something we can always add in a separate MSC if we realise we need it further down the line.

proposals/3912-relation-based-redaction.md
anoadragon453: With my spec hat on, that sounds quite Synapse specific... I'm not sure if it's a good enough reason on its own to spec an extra field in an event.

babolivier: My point is that if this is an issue with Synapse it can be an issue with other servers. And this paragraph is more meant to serve as an implementation hint than a field that should be spec'd - I'll reword to make it clearer.

proposals/3912-relation-based-redaction.md
babolivier: > Note that this will mean that anyone that tries to, say, respond to a thread that was redacted long in the past will have that message redacted immediately afterwards.

Yes, this is by design. When you want to remove a discussion from a room when moderating it, for example, you usually don't want other people to restart it (for the same reason some forums auto-lock threads after it considers the discussion has ended). Or if a client is slow to send an edit and you delete the root message with all of the edits before the slow client sends its edit, you don't want that to re-add data that should be deleted to the room.

> Also note that since only the homeserver that initially kicked off the redaction of the whole thread is responsible for redacting new related events; if that homeserver leaves the room then those messages will remain unredacted.

This is a good point, and I don't think there's a good answer to this, because all redactions are sent on behalf of the user who sent the `/redact` request. Technically it's an issue if the user who sent the redaction leaves the room (not even the whole server). I'll add it to the `Potential issues` section.

proposals/3912-relation-based-redaction.md
anoadragon453: OK - if you do store it in `unsigned`, be sure to strip it again before returning it to client/sending to other homeservers.

This has bitten Synapse in the past: https://github.com/matrix-org/synapse/pull/14064

babolivier: Done: https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/5823324ae696f13ab37e45b2b4f8ca138110153f

proposals/3912-relation-based-redaction.md
babolivier: Done: https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/5823324ae696f13ab37e45b2b4f8ca138110153f

proposals/3912-relation-based-redaction.md
anoadragon453: I assume clients must look at the `type` of the related event, otherwise they wouldn't be able to differentiate between stable and unstable versions of a related event (which may have different schemas).

On the flip side though, only specifying the relation type allows you to redact both the unstable and stable versions of a related event without needing to specify the unstable type (avoiding tech debt!).

You're right in that we can always add a field later if need, so let's just go with that for now.

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+### Unstable feature in `/version`
+
+Homeservers which support this MSC should indicate it by adding
+`org.matrix.msc3912` and `org.matrix.msc3912.stable` in the response to `GET
+/_matrix/client/versions` requests.
+
+
+## Potential issues
```
turt2live, line 117: `with_relations` also assumes the client is aware of the relation types associated with an event: in the case of moderation, the client might not be aware. This could be solved by allowing `with_relations` to also be a boolean (default `false`, similar to just not being present) - when `true`, *all known child events* are also redacted.

babolivier: That is a very good point, though I'm wondering about the impact making this a flexible type can have on implementations using strictly typed languages. How about introducing a catch-all `"*"` value, which if found in the list means "any relation type"?

jplatte: As maintainer of Ruma, I don't have a strong feeling here. We will want to use an enum with the known relation types + fallback variant anyways, whether one of the possible valued is represented as a bool in the serialized format doesn't matter that much.

My initial gut reaction was that the bool is cleaner but since `rel_type`s only have a handful of possible values, I guess `"*"` is perfectly fine too (and I'm pretty sure there's at least one case where that is already a special value).

weeman1337:
I am currently working on adding this to the Matrix JS SDK.

This is an extension to an existing endpoint. The same would probably apply to the function call in the SDK: Current plan is to add an option tot he redact call. How do you think the SDK should behave in this case?

- Send the event anyway. Means the caller is responsible to check for server side support before.
- Check for server side support on redaction with `with_relations` and raise an error if no support.

babolivier:
Hey @weeman1337, this is probably something the Element Web team would be better positioned to answer (since it's an implementation detail), but I would expect clients to check for server support and only show the feature to the user if there is server-side support for it.
Also, please use threads to comment on MSCs (by clicking on a line in the diff), since it makes the timeline more readable :slightly_smiling_face:

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
+
+{
+    "with_relations": ["m.replace"]
+}
+```
+
+Causes events `$a` and `$b` to both get redacted.
```
richvdh, line 81: To be explicit, the link only works one way? So in this example, `$some_event_id` will not be redacted, despite `$a` having an `m.replace` relationship to `$some_event_id` ?

babolivier: Can't update the MSC to clarify because I don't have write access anymore, but that is correct. I'm not sure I clearly remember the reason though (and I can't look it up since that was likely recorded in private documents/threads that I don't have access to anymore - maybe @giomfo can help), but I remember it being an explicit decision.

giomfo: @richvdh We didn't consider this use case in our initial discussion. My demand was focused on `m.reference` value of relation. I think your remark should be taken into account to update this MSC

giomfo: First the purpose of this MSC is to be able to redact some child events related to the redacted event. We don't target to redact the potential parent event. So here in this example, `$some_event_id` will not be redacted on purpose.

The `m.replace` relation is not the most suitable example to illustrate this change in real life. I consider `m.reference` is the actual targeted relation by this MSC. 
@richvdh: Should I update the MSC with another example, or just mention that `$some_event_id` will not be redacted on purpose?

On another hand, your remark inspired me an extension of this MSC. I consider we should not consider the possibility to spread the redaction in reverse the events tree here, but I think we may want to spread it down. I mean the user may want to redact all the child events which are related to the redacted child events. By considering the example: 
```
PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$some_event_id/foo
{
    "with_relations": ["m.replace"]
}
```
Causes events `$some_event_id` and `$a` to both get redacted, whereas the end user may want to redact `$b` too.

@richvdh @turt2live: Should I update this MSC to add a new optional `spread` property to the body (`false` by default)? `true` would mean to propagate the redaction through the child events respecting the provided relations. (Any suggestion to name this property is welcome)

proposals/3912-relation-based-redaction.md
giomfo: @turt2live In order to unblock this MSC, I will ignore this suggestion, it should be handled in another MSC if need. For the moment the moderator will have to list all knows relation types.

I let you think about the suggested boolean `spread` mentioned [here](https://github.com/matrix-org/matrix-spec-proposals/pull/3912/files#r1160552838). If we go for it, I plan to specify to ignore it if `with_relations` is missing or empty. But in order to satisfy your demand here, we may consider that the redaction is propagated to all child events whatever is the relation if no relation is provided and `spread` is `true`. wdyt?

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,148 @@
+# MSC3912: Relation-based redaction
```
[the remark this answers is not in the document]
richvdh: Though, tbh, every time I read this title, I think it must be some new way of sending a redaction request, that requires the redacting event to have an `m.relates_to` to the original event.

Maybe "Redaction of related events" would be be better?

proposals/3912-relation-based-redaction.md
richvdh: It has to be said that the current example is somewhat confusing. As https://spec.matrix.org/v1.6/client-server-api/#validity-of-replacement-events says: "you cannot edit an edit", so `$b` is not actually a valid edit event. (Though we should make sure that we consider this situation and make the behaviour explicit in the MSC.)

> The `m.replace` relation is not the most suitable example to illustrate this change in real life.
> I consider `m.reference` is the actual targeted relation by this MSC.

Why `m.reference`? Because that is used by MSC3489?  To be honest, the `m.replace` relation seems a fairly clear usecase to me - though as above, the example is confusing.

> @richvdh: Should I update the MSC with another example, or just mention that `$some_event_id` will not be redacted on purpose?

My recommendation would be:
 1. Remove the `m.relates_to` section of `$a`.
 2. Keep the example request as-is.
 3. Make an explicit mention that a redaction of `$b` would *not* cause `$a` to be redacted, because we do not consider parent events.
 4. Give an explicit statement on what happens to events that do not meet the [validity requirements](https://spec.matrix.org/v1.6/client-server-api/#validity-of-replacement-events) for edited events (and consider if there are similar issues for other relationship types)

> @richvdh @turt2live: Should I update this MSC to add a new optional `recursive` property to the body

I'm not sure we have a usecase for this?

giomfo: @richvdh Hi, I updated this MSC by following your recommendations (see https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/7fe8d6f83cda2f89829bd6962336d1a8da977d75). I let you resolve this conversation or add some new comments

proposals/3912-relation-based-redaction.md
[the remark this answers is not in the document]
giomfo: @richvdh FYI I changed the title (see https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/7fe8d6f83cda2f89829bd6962336d1a8da977d75)

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,148 @@
...
+Currently, the only way to do this is for the client to keep track of each
+individual event and redact them itself. This is cumbersome because it might
+require clients to work around their own limits on how much history is stored
+for a given room, and might cause them to hit rate limits when attempting to
+send redactions, which would translate into general slowness for the user (since
+their other actions would also be prevented by rate limiters as a result).
+
+## Proposal
+
+We add a new optional `with_relations` property to the body of the [`PUT
```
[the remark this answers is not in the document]
giomfo: @richvdh If we rename this property, we will have to update all the existing implementations (server and clients). I'm ok to go for this renaming, but can you confirm this is actually relevant. What should we do with the existing `org.matrix.msc3912.with_relations`?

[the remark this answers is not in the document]
richvdh: The existing, unstable, name, can stay: we can just fix this when we switch from unstable names to stable names.

Presumably we are going to have to update all of the implementations anyway when we switch from unstable to stable identifiers?

In general, I think it's going to be quite confusing if introduce a new term ("relations") to refer to event relationships. They are confusing enough as it is.

By the way, looking at this again, I much prefer `with_rel_types`.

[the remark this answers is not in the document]
giomfo: resolved by 55060fc3b4634215d417b4cf280e8cda1bb64613

proposals/3912-relation-based-redaction.md
giomfo: @turt2live FYI I added the catch-all `"*"` value in this commit 55060fc3b4634215d417b4cf280e8cda1bb64613

giomfo:
@richvdh @turt2live FYI I resolved all the pending discussions on this MSC with my 2 last commits.
I filed a new issue on each platform to update the current implementation according to the changes introduced by the [last commit](https://github.com/matrix-org/matrix-spec-proposals/pull/3912/commits/55060fc3b4634215d417b4cf280e8cda1bb64613)
- synapse: https://github.com/matrix-org/synapse/issues/15687
- web: https://github.com/vector-im/element-web/issues/25471
- ios: https://github.com/vector-im/element-ios/issues/7563
- android: https://github.com/vector-im/element-android/issues/8481

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,197 @@
...
+cannot be redacted by the user sending the redaction request (because of
+insufficient permissions), it is ignored. Clients may attempt to detect cases
+where a user would not be able to redact some events matching the redaction
+criteria (e.g. matching `m.reference` relations when redacting a `m.poll.start`
+event, matching `m.thread` relations, etc., with an insufficient power level to
+delete other users' events) and display a warning to the user in this case.
+
+If an event that matches the redaction criteria comes down federation after
+redaction requests has completed, the server must attempt to redact it on behalf
+of the redaction request's sender.
```
richvdh, line 101: The Synapse PR referenced in the PR description (https://github.com/matrix-org/synapse/pull/14260) says:

> This implementation is lacking a way to keep watch for incoming events that match the redaction criteria after the redaction request finishes, that'll come in a separate PR.

Is there any indication that this second part has happened? If not, then I don't think that counts as a complete implementation of the MSC and we should remove those parts from the MSC.

richvdh (changes requested):

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,197 @@
...
+}
+```
+
+Causes events `$a` and `$b` to get redacted, but no `$c`
+
+### Unstable feature in `/version`
+
+Homeservers which support this MSC should indicate it by adding
+`org.matrix.msc3912` and `org.matrix.msc3912.stable` in the response to `GET
+/_matrix/client/versions` requests.
```
richvdh, line 163: presumably there is some difference between  `org.matrix.msc3912` and `org.matrix.msc3912.stable`? Please make it explicit what that difference is.

proposals/3912-relation-based-redaction.md
```diff
@@ -0,0 +1,197 @@
...
+PUT /_matrix/client/v3/rooms/!someroom:example.com/redact/$a/foo
+
+{
+    "with_rel_types": ["m.replace"]
+}
+```
+
+Causes events `$a` and `$b` to get redacted, but no `$c`
+
+### Unstable feature in `/version`
```
richvdh, line 159:
```suggestion
### Unstable feature in `/versions`
```

proposals/3912-relation-based-redaction.md
richvdh: (I think https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3827-space-explore.md#unstable-prefix does a decent job of this -- in particular it explains what clients should do at each stage)

richvdh: are we sure we need `org.matrix.msc3912.stable` ? We should only spec a "stable"  `unstable_feature` if we absolutely need to get this feature into widespread use before homeserver implementations support the spec version which includes this.

babolivier:
Heads up I am formally abandoning this proposal since I don't currently have time for Matrix spec contributions (I'm not sure whether I should close it or wait for someone to adopt it).

PR #3510, opened 2021-11-20
MSC3510: Let users kick/ban/demote each other

ara4n:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/let-users-kick-each-other/proposals/3510-let-users-kick-each-other.md)

Preview: https://pr3510--matrix-org-previews.netlify.app

PR #3915, opened 2022-10-23
MSC3915: Owner power level

ara4n:


proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+transferring/sharing ownership.
+
+## Potential issues
+
+None?
+
+## Alternatives
+
+ * We could add an `owner` field or `owners` field to the PL event. However, this breaks compatibility, and has no
+   obvious benefit from simply having a higher PL number for owners.
```
ShadowJonathan, line 46: Referencing what @su-ex said in [their comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3510#discussion_r761836336), this may still have value, as then these are special users that will never be "excluded" from the room, becoming a definite lifetime. Having 1 owner (or a series of explicit owners) inside the powerlevel event would also give explicit atomicity of ownership in such way.

Furthermore, to expand that idea, is to rewrite stateres to *always* allow transitions to `join` membership by anyone that is in this array/key, to allow owners to always relinquish command/observation over the room, even if power levels have been messed with.

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+
+This closes the embarassingly long-standing https://github.com/matrix-org/matrix-spec/issues/165 spec issue, and
+replaces the previous flawed [MSC3510](https://github.com/matrix-org/matrix-spec-proposals/pull/3510) proposal.
+
+## Proposal
+
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
+ * The room creator defaults to PL150.
+ * We tighten the language in the spec which says "Clients may wish to assign names to particular power levels." to
+   specify the threshold names more concretely, to avoid divergent implementations.
```
ShadowJonathan, line 30: To better introduce this change, I suggest tying them to a specific version marker such as the matrix version this'll be introduced in, or a room version, to prevent fragmentation of client and server understanding of this mechanic, else an eager server or client will introduce this, while others might not, causing potential confusion (either of labels or programmatic assumptions)

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
+# MSC3915: Owner power level
```
FSG-Cat, line 1: So im just going to say. This Proposal unlike MSC3510 i actually can stand behind without issue. This MSC does not break any existing conventions it only helsp aleviate user confusion while not actually changing how matrix works in a core way. 

Since we already can create rooms with custom powerlevesl from my prespective this change should not be able to create issues because if it creates issues that is a bad implementation already. But ofc it can cause clients to be confused about the new 150 PL tier but that is about it. Current Element Web as of last i checked treats all 100+ as admin. (Checked against 9001 and 2^53 ish. Yes rooms exist that use PL that is close to PL MAX.)

This change vs the current 100 and 99 system leaves room for an admin hirachy and i actually like that a lot. Being able to define what admins should be able to act on each other is a good idea to let the community have access to in a easier way than the current way of custom powerlevel rooms that is hard to get access to without being a homeserver admin running Synapse that switches to [Tulir Synapse fork. ](https://mau.dev/maunium/synapse)

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+Everything else stays the same (e.g. the `trusted_private_chat` prefix would still give invited users Owner too, to
+share ownership of the DM with them).
+
+This clearly differentiates owners from their delegated admins, is compatible with existing rooms without a room version
+bump, and lets clients implement cmopletely different UI for appointing/deappointing admins to the special-case of
+transferring/sharing ownership.
+
+## Potential issues
+
+None?
```
Mikaela, line 41: How will existing rooms upgrade to this considering a room upgrade currently copies power levels from the previous room?

I think this proposal should require the upgrade process to maintain that the upgrader either gets this 150 or gets the new default power level depending on which is higher or otherwise I think Element Web would downgrade the administrator instantly, while I haven't tried this.

Mikaela (changes requested):
I have a concern on how upgrading current rooms work.

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+is no way to revoke admin permissions (short of coercing the other admins to demote themselves).
+
+This is particularly problematic for users migrating from other communities such as Discord or even IRC where room
+permissions distinguish owner from admin, and so apply the same terminology in Matrix, accidentally promoting users to
+admin thinking they can revoke it (despite the UI warnings to the contrary).
+
+Finally, the current workaround of having to manually assign 'admins' to PL99 and manually rewrite the PL event
+thresholds is incredibly uninuitive and bad UI (for instance, many clients will show the PL99 as 'Moderator' in the UI,
+confusing it with PL50).  Moreover, you can't apply the workaround in retrospect; the owner either has to create a
+whole new room with the correct permissions or anticipate the problem before you accidentally assign a user to PL100.
```
Gnuxie, line 20: An owner only makes sense in a situation where there is an owner to some related IP or product. Communities that spontaneously form around something don't have an owner. I don't think this solves a situation where a room creator doesn't trust admins they are giving permission to and it only worsens a situation where a community cannot trust its admins (because now there will be fewer with absolute power). It also arguably worsens the situation where there are hundreds of dead rooms because the original creators who didn't appoint anyone but moderators abandoned Matrix.

>Finally, the current workaround of having to manually assign 'admins' to PL99 and manually rewrite the PL event
thresholds is incredibly uninuitive and bad UI

So instead why not provide UI to template power levels on a room or server level?

proposals/3915-owner-power-level.md
Mikaela: Also how about Schalar and the bridges it manages? Willl Schalar/IRC integration start getting power level 150 to perform room takeover from the founder?

How will a bridge like LiberaChat's behave if it doesn't have the power to remove inactive users from the room? Will it silently break?

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
+ * The room creator defaults to PL150.
+ * We tighten the language in the spec which says "Clients may wish to assign names to particular power levels." to
+   specify the threshold names more concretely, to avoid divergent implementations.
+
+Everything else stays the same (e.g. the `trusted_private_chat` prefix would still give invited users Owner too, to
+share ownership of the DM with them).
+
+This clearly differentiates owners from their delegated admins, is compatible with existing rooms without a room version
+bump, and lets clients implement cmopletely different UI for appointing/deappointing admins to the special-case of
```
tezlm, line 36: "cmopletely"

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+This closes the embarassingly long-standing https://github.com/matrix-org/matrix-spec/issues/165 spec issue, and
+replaces the previous flawed [MSC3510](https://github.com/matrix-org/matrix-spec-proposals/pull/3510) proposal.
+
+## Proposal
+
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
+ * The room creator defaults to PL150.
+ * We tighten the language in the spec which says "Clients may wish to assign names to particular power levels." to
+   specify the threshold names more concretely, to avoid divergent implementations.
+
```
Mikaela, line 31: Just to clarify, PL100 will still be able to ACL ban the owner or tombstone the room pointing somewhere they have PL150 and the original owner nothing?

BurnyBoi:
So this is the same exact thing with the same exact risks as PL100, just with yet another PL boost that could be misused the same way? What's stopping the PL150 from giving other users PL150 and causing the same "problems" again? Why can't the PL100 owner just use PL99 for this purpose?

proposals/3915-owner-power-level.md
richvdh, line 30: for clarity, could you link to the parts of the spec you are proposing to change? I think it is just https://spec.matrix.org/v1.4/client-server-api/#permissions, where the current recommended mapping is given, and also a vaguely-worded "the room creator's power level defaults to 100". [1]

We might also want to update https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3createroom to recommend explicitly that a PL be sent giving the creator PL150 (currently there is no explicit PL given here).

[1] Does it mean the PL before or after the first PL event, or both?

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+confusing it with PL50).  Moreover, you can't apply the workaround in retrospect; the owner either has to create a
+whole new room with the correct permissions or anticipate the problem before you accidentally assign a user to PL100.
+
+This closes the embarassingly long-standing https://github.com/matrix-org/matrix-spec/issues/165 spec issue, and
+replaces the previous flawed [MSC3510](https://github.com/matrix-org/matrix-spec-proposals/pull/3510) proposal.
+
+## Proposal
+
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
+ * The room creator defaults to PL150.
```
richvdh, line 28: We need to be more explicit here. In particular: would we change the default power level if there is no `power_levels` event (defined to be 100 at https://spec.matrix.org/v1.4/client-server-api/#mroompower_levels)? I don't *think* this is strictly necessary (since the first PL event is special-cased), and doing so would require a room version bump, but it would be nice to keep things consistent.

ShadowJonathan: I also touched upon this in [my comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3915#discussion_r1002688933), if this is the intention, this new permissions structure could also be introduced with a new room version

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+thresholds is incredibly uninuitive and bad UI (for instance, many clients will show the PL99 as 'Moderator' in the UI,
+confusing it with PL50).  Moreover, you can't apply the workaround in retrospect; the owner either has to create a
+whole new room with the correct permissions or anticipate the problem before you accidentally assign a user to PL100.
+
+This closes the embarassingly long-standing https://github.com/matrix-org/matrix-spec/issues/165 spec issue, and
+replaces the previous flawed [MSC3510](https://github.com/matrix-org/matrix-spec-proposals/pull/3510) proposal.
+
+## Proposal
+
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
```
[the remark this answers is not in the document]
MTRNord: Does it really make any solution to the initial problem if the owner is PL1, 100, 1000, - 5 or a million? I feel like the whole msc just shifts the initial issue to another number. 🤔

proposals/3915-owner-power-level.md
kegsay: > It also arguably worsens the situation where there are hundreds of dead rooms because the original creators who didn't appoint anyone but moderators abandoned Matrix.

I agree with this. This MSC does not adequately give equal voice to the pros of the current behaviour, which is that having >1 admin/owner can be a beneficial thing when the original owner disappears. Furthermore, this MSC does not explain practically what changes here. Okay, so we have "Owner" now, what does this give a user by default over and above "Admin"? Which extra permissions are granted? Looking at Element-Web, this is the current list of Admin only operations _by default_:

Screenshot 2022-10-25 at 16 04 52

So in other words:
 - Enabling encryption: probably should be owner, you don't need this to moderate a room. It's also irreversible, so making this an owner thing makes sense.
 - Changing history visibility: keep this as admin? It's reasonable for an admin to decide to not expose room history (or change their minds) without the owner's permission?
 - Changing power levels (permissions): Definitely should be an admin thing.
 - Changing server ACLs: admin thing to help fight spam.
 - Upgrading a room: tricky as the upgrade could be for defensive purposes e.g to fix a protocol vulnerability, and needing permission from a hard-to-reach owner could be disasterous.

MTRNord: > Currently, the only effect this MSC would have would be to allow owner-admins to demote admins, so in other words they are a kind of super-admin. Other than that, normal admins can still wreck a room in a plethora of other ways (ACL banning, upgrading the room, etc)

It doesn't solve hostile takeover through demoting, though. There are issues where the room creator is the hostile person in a community. This is now not solved. Also in communities where there doesnt exist a single owner (For example a big community with a group of leaders) this model doesn't fit reality. If the room owner leader has to leave for some but refuses, you end up with the initial issue all over again.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
dbkr: The bike shed would look just fine in either of these colours ;)

proposals/3915-owner-power-level.md
dbkr: Considering this as a benefit of the current behaviour sounds a bit spurious. A room can still have multiple owners to get the same benefit. I think it's essentially saying that we want to force distribution of power because it's the only way to do `x`.

> So instead why not provide UI to template power levels on a room or server level?

I'm not quite sure what this means - it sounds like it might be something that warrants an alternative MSC though? This would help explain the idea.

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+Clients or servers which incorrectly assume that PLs range from 0 to 100 may get confused.  This would clearly be
+an implementation bug however.
+
+## Unstable prefix
+
+None
+
+## Dependencies
+
+None
```
dbkr, line 71: Have tried to address a few of the comments here. In general, this is a thing I've been slowly convincing myself of independently recently, so I think I agree it's the way to go. It doesn't magically solve all problems, but it gets us to a place where admin of rooms is a bit easier to understand, I think.

One note on what the actual power levels are (ie. 99/100 vs 100/150): if we use the former then existing rooms kind of get these semantics for free (ie. when a client got updated, members who were previously admins would show up as owners and one of them could demote others to admin level). However, having things appear to change when you're not expecting it could be quite scary: 100/150 would keep old rooms just as they are.

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Alternatives
+
+ * We could add an `owner` field or `owners` field to the PL event. However, this breaks compatibility, and has no
+   obvious benefit from simply having a higher PL number for owners.
+
+ * We could redeclare the Admin PL threshold to be 75 or 99 or similar, thus showing all existing Admins as being Owners
+   in the UI.  However, this will cause confusion with older clients, as well as confusion over users being seen to
+   apparently suddenly changing power levels.  It's better to reflect the current reality: that historically we haven't
+   had the concept of owners, rather that retconning the historical admins into being owners.
```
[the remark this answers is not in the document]
Gnuxie: Historically there's no distinction because the concept of an owner didn't exist, and there was no reason for it to. That's a personal opinion as much as this MSC's opinion that rooms indeed do have owners, ownership is not real. You and others have invented it only because there is a need for property holders to enforce their ownership here in Matrix too (as it doesn't exist and only exists through power).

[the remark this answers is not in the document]
Gnuxie: Specifically this is a response to the wording here suggesting ownership is a shared, canonical concept and the appeal to bikeshedding and personal opinions [here](https://github.com/matrix-org/matrix-spec-proposals/pull/3915#discussion_r1004875326), which feels to like it is aimed directly to dismiss this any issue with non-technical objections. Which isn't fair when the reason for this MSC existing is because of non-technical issue.

proposals/3915-owner-power-level.md
ara4n: yes, we could mandate a single owner. but life isn't like that: things are often equally owned by multiple people who mutually trust each other and can't demote each other (e.g. cofounders of a company). agreed that we could have a separate MSC which gives owners permission to always join a room. (`s/relinquish/retain/`)

proposals/3915-owner-power-level.md
ara4n: I've deliberately chosen an approach here which can be added *without* a specific version marker. Admins continue to be admins at PL 100; the labels and programmatic assumptions remain the same; meanwhile Owners get added above them. Obviously we can revisit this when the MSC is finished, but right now I don't see it being necessary.

proposals/3915-owner-power-level.md
ara4n: cool, thanks for the positive feedback.

It's worth noting that Cinny has been doing precisely this (albeit with 101 for Founders) for a while without any side-effects other than happy users.

Screenshot 2022-10-26 at 20 07 16

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
ara4n: @MTRNord it doesn't shift the problem to a different number - by explicitly having two levels in the UI to distinguish owner (who can set/unset admins) from admin (who have permission to administer the room) we force users to explicitly consider who they are sharing ownership with - and stops them from being obligated to share ownership with folks they merely want to delegate admin to.  It's the same reason that we have separate PLs for Mods & Admins already today.

Now, sure you can still have power struggles between owners, but that's *much* less likely when the only thing that makes them an owner is the fact they can own (as opposed to having other permissions mixed up in it, like the ability to admin the room).

@uranuspucksaxophone your suggestion of 1000 is noted, but in the end the difference between 150 and 1000 is very subjective and pretty arbitrary, and i'd rather just pick a number and stick with it.

proposals/3915-owner-power-level.md
ara4n: > members who were previously admins would show up as owners and one of them could demote others to admin level

they wouldn't, though. admins would still not be able to demote each other. (and owners would not be able to demote each other either).

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+(set PLs, ACLs, tombstones, history visibility, encryption, create/remove moderators)... and the 'owners' aka 'founders'
+of a room, who have overall responsibility and ownership for it and appoint (and deappoint) admins to help them run it.
+
+As a result, it's common for a room founder to promote a set of users to be admins at PL100 to share responsibility
+for administering the room... but then regret that they have effectively relinquished ownership of the room, given there
+is no way to revoke admin permissions (short of coercing the other admins to demote themselves).
+
+This is particularly problematic for users migrating from other communities such as Discord or even IRC where room
+permissions distinguish owner from admin, and so apply the same terminology in Matrix, accidentally promoting users to
+admin thinking they can revoke it (despite the UI warnings to the contrary).
```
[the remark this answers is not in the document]
ara4n: yes, "let space owners automatically be owners of rooms in that space" is a separate MSC (probably MSC2962). this MSC focuses purely on solving the "differentiate owners from admins" problem.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
ara4n: The reason to keep admins at 100 and add owners at 150 is simply so that current rooms don't simply rename their permissions (inconsistently, as clients gradually get updated).  Rather than confusing users by switching the terminology around, let's just keep the current terminology where it is and add the Owner level on top.  After all, the admins in the current rooms are primarily admins rather than owners; the clue is in the name :)

proposals/3915-owner-power-level.md
ara4n: This is a fair point; currently Admins can tombstone rooms, so a malicious admin could currently fake this today by creating a new room where they're the sole admin (or owner) and tombstoning the room into it.  I don't think that the act of upgrading a room should make you the owner, but perhaps we should make tombstones and other super-destructive actions (ACLs) owner-only for new rooms, as you propose in the other comment.

Meanwhile, existing ownerless rooms with multiple admins can break the impasse by tombstoning the room to point to a new room with a well-defined owner.

proposals/3915-owner-power-level.md
ara4n: > An owner only makes sense in a situation where there is an owner to some related IP or product. 

I don't think this is true. Any room ends up getting created by a single user, who defacto 'owns' the room they just created. Depending on their use case, they then either can explicitly share ownership of the room with other people, or simply give them admin rights, depending on what they want to do and what their community expects.

> I don't think this solves a situation where a room creator doesn't trust admins they are giving permission to

It does; the room creator can now give them admin rights without sharing ownership with them.  The creator can now demote them if desired.

> it only worsens a situation where a community cannot trust its admins

No, because the creator/owner(s) can now get rid of admins.  Of course, if the creator/owners turn out to be evil, there's no alternative but to fork a new room.   But that is a social problem that Matrix cannot solve.

> It also arguably worsens the situation where there are hundreds of dead rooms because the original creators who didn't appoint anyone but moderators abandoned Matrix.

I don't think it worsens it; in fact it makes it less likely to happen because it will be easier to appoint admins because you are no longer sharing ownership with them when you do so.

> So instead why not provide UI to template power levels on a room or server level?

Because applying different labels to different PLs on every different room sounds like it will cause more confusion than it helps, whereas at least today we have a relatively consistent set of labels.  Feel free to propose it as a separate MSC though!

proposals/3915-owner-power-level.md
ara4n: yes, I was thinking of https://spec.matrix.org/v1.4/client-server-api/#permissions and https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3createroom.

However, as @Mikaela points out, we might want to explicitly change the default PLs further - to stop admins using server ACLs or tombstones to hijack ownership of a room.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
ara4n: > Ownership is not real. You and others have invented it only because there is a need for property holders to enforce their ownership here in Matrix too (as it doesn't exist and only exists through power).

No, this is completely untrue. The reason to distinguish ownership from administration is the same reason that we distinguish admins from moderators; they can do different things.  Owners can appoint admins; Admins can appoint moderators.  By labelling the levels of ownership & administration differently in the UI, it gives us (pre-RBAC) the ability to distinguish the roles in a consistent manner.  If you prefer, you could call Owners 'Superadmins' or 'Founders' or 'root' or whatever; i really don't care. The important thing is that it provides an additional tier of control to make up for the lack of RBAC rather than being anything to do with "ownership doesn't exist" philosophy.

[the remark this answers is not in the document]
ara4n: > Which isn't fair when the reason for this MSC existing is because of a non-technical issue.

No, the reason the MSC exists is entirely a technical issue. Right now the UI of apps conflate permissions A and B together into a single role. This splits the role into two to avoid the incorrect conflation.

proposals/3915-owner-power-level.md
ara4n: yes, although as you imply, this was a thinko and should be addressed in this MSC. thanks for picking it up.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
ShadowJonathan: > it gives us (pre-RBAC) the ability to distinguish the roles in a consistent manner

If this is merely an MSC to label ownership, it does neither fix the fundamental problem (absence of RBAC or more granular controls) nor fix a user-education absence in matrix today. Trying to double down on the "owner" semantic here is not going to help, as the same stateres restrictions are still in place that lead to the same confusion, trying to add "ownership" to that mess of understanding is only going to make this more murky.

I think that, to fix this, there should be an MSC that addresses that fundamental problem, with RBAC or similar as you mentioned.

----

Also, please do not resolve threads after one response without a chance for the other party to voice their rebuttal, or agree with your stance, or otherwise state their thoughts about your response.

It comes over as silencing the debate, which to my eyes falls very poorly on your role as matrix.org guardian, whom is supposed to be above this kind of debate-forming.

I ask you to please respond in a different manner, or let conversations 'resolve themselves', it feels uncomfortable to see conversation take such a turn.

proposals/3915-owner-power-level.md
Mikaela: > perhaps we should make tombstones and other super-destructive actions (ACLs) owner-only for new rooms

I think this could make moderation tools have even higher threshold for use such as Mjølnir that needs access to manage ACL for server bans. The easiest action seems like making the bot a co-owner (which I understand is supposed to be somewhat hidden action in the UI(?)), which results to it being unmanageable by other owners unless they have access to the bot account in case of trouble.

Mikaela (commented):
I am not sure about moving power level requirements for owners

proposals/3915-owner-power-level.md
Ai-rin: > (and owners would not be able to demote each other either).

Then we have dead accounts problem again? Shoudn't we have an exception for PL150 ?

FSG-Cat: > > (and owners would not be able to demote each other either).
> 
> Then we have dead accounts problem again? Shoudn't we have an exception for PL150 ?


it would introduce the exact problem this MSC addresses when it replaces https://github.com/matrix-org/matrix-spec-proposals/pull/3510. Also in all current room versions the PL names are just visual and actually dont matter at all from the prespective of the authorisation rules so how this is solved im not sure.

proposals/3915-owner-power-level.md
jackmawer: > Meanwhile, existing ownerless rooms with multiple admins can break the impasse by tombstoning the room to point to a new room with a well-defined owner.

would it not make sense to, for example, read the room history and find the first user who created the room, and upgrade that one person to PL150? that would potentially allow an easier transition where owners can then be applied as needed.

bkil:
This will lead to even more unmoderated rooms being abandoned by all admins and moderators over time without a possibility for correction.

EmilJeppsson:
The issues of owners going missing could be solved by implementing an inheritance/succession for the owner role. The owner would specify an amount of time since their last room interaction after which other admins/members (in order of power level) would be able to initiate a claim on the owner role.

An example of how this could look:

Jeff (PL150) has specified 30 days as the countdown from his last event until claims are allowed. He has also specified the succession time as 10 days.
On day 33 since Jeff's AWOL, Bob (PL100) initiates a claim on the owner role. Unless Jeff returns or another admin with higher PL either aborts the claim or takes Bob's place as the claimee, Bob succeeds Jeff as the owner by day 43.

This way the power levels could also function as a line of succession.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
EmilJeppsson: That's fair.
I guess similar functionality could be delegated to a trusted bot/appservice instead. It wouldn't be decentralised (or part of the spec) but would still solve the problem of owner abandoned rooms.

proposals/3915-owner-power-level.md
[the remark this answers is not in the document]
BLumia: Might sound like bikesheding but I suggest using just 101 for the founder/owner. Since existing/regular client only allow the range of 0 to 100, so 101 would be special enough to make sure both protocol developer / specification reader know it's a special/magic value that should avoid be set via regular API calls. So, we can safely say the range is still `0-100` and treat 101 only as a special marker value.

proposals/3915-owner-power-level.md
```diff
@@ -0,0 +1,71 @@
...
+
+ * We expand the recommended range of PLs to 0-150, with 150 described as 'Owner'.
+ * The room creator defaults to PL150.
+ * We tighten the language in the spec which says "Clients may wish to assign names to particular power levels." to
+   specify the threshold names more concretely, to avoid divergent implementations.
+
+Everything else stays the same (e.g. the `trusted_private_chat` prefix would still give invited users Owner too, to
+share ownership of the DM with them).
+
+This clearly differentiates owners from their delegated admins, is compatible with existing rooms without a room version
```
BLumia, line 35: Do you mean we allows multiple owners per room? That sounds like a pretty bad idea since it can lead to the very same issue with the current 100-is-admin solution. We probably need to limit it from the spec so we can avoid mis-implementation. Something like:

- A room can only have *zero* or *one* owner/founder (for room that upgraded from earlier versions, there will be no founder, and they need to create a new room to make use of the founder feature)
- founder (PL > 100) cannot be set directly, it can only be transfered from the original founder to the new founder
- founder cannot leave the room directly (the founder needs to transfer the ownership/permission to someone else before they can leave)

BLumia (commented):
I strongly suggest we keep using the range `0-100` for power levels, and use `101` as a special power level with some additional server-side implementation conditions. See the comment for details.

For summary:

1. We should only allow max to 1 (one) owner per room. (No people have PL101 is fine so we keep 100% compatible with earlier versions. Either no owner or only one owner)
2. Role permission should stick to the 0~100 range.
3. There should be a standalone API to transfer the PL101 "foundership"/ownership.
4. We should avoid "generally" extending the original power level range (`101` as a very special case).

proposals/3915-owner-power-level.md
su-ex: I strongly disagree:
1. Sharing ownership is a common concept, see https://github.com/matrix-org/matrix-spec-proposals/pull/3510#discussion_r761836336.
2. There are already rooms that go beyond that scope.
3. That just adds complexity.
4. It's no special case, it gives maximum flexibility
    - to even step down as single owner after room creation if it suits the concept of the room more.
    - own a project together.
    - give bridges the possibility to administer rooms without sharing ownership with the bridge.

BLumia: I personally think all you want is the "zero owner" use-case, which is basically the current state.

Additionally maybe we can simply let the room creator to choose the room-creator's power level, can be either PL101 (one owner) or PL100 (zero/no owner).

> Sharing ownership is a common concept,

I don't get it. Since PL101 only gets the special power to meddle with other people's power level, all other permissions are the same as PL100. Why is this something that needs to be shared? If that's the case, can a PL101 user change another PL101 user's power level down to 100 or less?

IMO even for the huge-skyscraper room use-case, we should still keep the "owner" count max to one, so we can 100% ensure there can always have a people who can tweak other admin's power level, instead of accidently set two or more people to the max power level and then no-one can do anything to this status.

> There are already rooms that go beyond that scope.

I know since current spec doesn't forbid that. If we define the PL range clearly in the spec, then it will be a guideline so other client and server implementation will be trying to follow. As we currently only use the range 0~100 and there are still client/server use 101 or 150 for the "Founder" usage.

btw, if we ensure the spec clearly defines the power level range to 0~100, then we can safely use the `PL > 100` condition to check if the user is the owner of the server for maximized compatibility.

> That just adds complexity.

Yes, I agree, but it's the only way to address the "... but then regret that they have effectively relinquished ownership of the room, given there is no way to revoke admin permissions (short of coercing the other admins to demote themselves)" issue.

> it gives maximum flexibility...to even step down as single owner after room creation if it suits the concept of the room more.

I've mentioned we can transfer the ownership to someone else.

> ..own a project together.

Then consider PL100 is enough and never set a PL101 user from the beginning of the time. (the zero owner use-case)

BLumia: And to explain more about my suggestion:

> We should only allow max to 1 (one) owner per room.

To be clear, I mean, the room creator can choose to have the PL101 special power (suggested to be the default value) or PL100 when creating a room. As a "Do you want to ensure this room will always have one 'owner' level user?" option. If people intended to share the same permission equally, just choose the PL100 from the beginning of the time.

> The owner cannot leave the room directly (the owner needs to transfer the ownership/permission to someone else before they can leave)

Two options:

1. Transfer the ownership to someone else.
2. Demoting themself to PL<=100

And for client developers, we should warn the user about performing this action in both cases.

su-ex: > I personally think all you want is the "zero owner" use-case, which is basically the current state.

This is exactly ***not*** what I want. I want a clear distinction between _owning something_ and _administering something_, e.g. bots or bridges should be able to administer the room **without** taking co-ownership (which is still legitimate to share, but **only by intention**).

> If that's the case, can a PL101 user change another PL101 user's power level down to 100 or less?

Hell no, this was the whole point of https://github.com/matrix-org/matrix-spec-proposals/pull/3510#discussion_r761836336.

BLumia: > Hell no, this was the whole point of https://github.com/matrix-org/matrix-spec-proposals/pull/3510#discussion_r761836336.

So, there is basically no difference between go with the current max-to-PL100 solution with allow-multiple-PL101-user solution. Which will avoiding address the issue that the original proposal was trying to adress:

> ... but then regret that they have effectively relinquished ownership of the room, given there is no way to revoke admin permissions (short of coercing the other admins to demote themselves)

Allowing only a single PL101 exists at a time addressed this issue. There are no differences other than the PL101-user can naturally manage the PL<=100 user's power level. Why would that cause an issue? If it will, why not just both PL100 since apparently you want admins to share the very same rights.

> This is exactly not what I want. I want a clear distinction between owning something and administering something.

By definition, what's the difference between owning and administering?

In my suggestion, there is only one difference, which is PL101 is larger than PL100 so the PL101. Nothing else. All other things remain exactly the same as the current state. Curtain action's permission cannot be set to only allows power level 101 to perform (since the power level range sticks to 0~100), so the admins do own the room except PL101 can demoting PL100 user's power level just in case anything went wrong. If such a thing shouldn't happen, just choose PL100 when creating the room.

> e.g. bots or bridges should be able to administer the room without taking co-ownership (which is still legitimate to share, but only by intention).

Bots can use a lower power level than admins but share the very same permissions as admins! For example, make the bots use power level 90, and ensure power level 90 can do all the things that power level 100 can do. Then done! No need to extend the power level for this use-case.

Do remember, it is the admin unintentionally setting a bot or a user to the max power level (100) and then can nothing with it. This is the source of the issue. My suggestion only introduced the ability to stop this from happening and it's also opt-in.

So I actually don't think the change that I suggested could cause an issue for your use-case.

su-ex: You don't get the point at all: It's all about the semantic difference, not about being technically impossible.

> By definition, what's the difference between owning and administering?

... and everything else you mentioned, actually:

I could quote myself, this has all been covered in https://github.com/matrix-org/matrix-spec-proposals/pull/3510#discussion_r761836336, especially at the end.

BLumia: I'm sorry I'm not a native English speaker so I'm confused about your comment. So let me ask the question directly:

If you extended the range and then the current owner accidentally set another person (or even worse, a bot) to owner, what should they do? If they can do nothing, there might be another proposal to bump the room version and extend the range to 200. and later 250 and so on. The problem still exists if we just extend the PL range and do nothing else.

FSG-Cat: i should probably note. The true range is -2^53+1, 2^53-1. Its just that the default templates dont expose PL values above 100 currently or 150 under this proposal.

BLumia: > i should probably note. The true range is -2^53+1, 2^53-1. Its just that the default templates dont expose PL values above 100 currently or 150 under this proposal.

Yes, and I mean we should probably stick to using 0~100 when possible, and don't encourage the usage outside that range unless necessary.

The PL101 suggestion is basically the same as the PL150 one under this proposal but 101 sounds weird enough so it won't sound like a generic power level that we can apply it elsewhere. I actually think PL150 is also okay but PL150 sounds generic enough so it might still cause the same amount of confusion just like the current PL100 one has. There needs to be something in the spec or protocol to help the developer understand about the power level, and avoid the end-user accidently assigning the max power to another user without knowing what it actually means and can do nothing after the mistake.

su-ex: > If you extended the range and then the current owner accidentally set another person (or even worse, a bot) to owner, what should they do?

Well, nothing. If the ownership got shared, it got shared. So you're right, it's technically the same – but it's still different:
The important thing is the word "accidentally".
Right now, if someone wants to give appropriate permissions to manage a room to e.g. a bot and therefore wants to change its pl to current 100 admin, a sufficient warning message would sound something like that: "Do you really want to relinquish sole ownership of the room? If so, proceed. If not, pick an arbitrary value below 100, assign that to all the single permissions. And then assign that value to the user/bot account. Also note that clients will show that user/bot mislabeled as moderator, although that account will have administrative permissions for this room."
Administrative permissions and ownership a currently intermingled (at least UI-wise). Sole purpose of this change is to break that. With it, the warning could be simply "Do you really want to relinquish ownership?" with a red button "proceed" and a green button "give admin permissions only". And all that without giving up established values (leading to even more confusion) and homogenous across clients (have it well defined in the spec).

FSG-Cat: > > i should probably note. The true range is -2^53+1, 2^53-1. Its just that the default templates dont expose PL values above 100 currently or 150 under this proposal.
> 
> Yes, and I mean we should probably stick to using 0~100 when possible, and don't encourage the usage outside that range unless necessary.


Ok so the problem is that going above 100 is well essentially a custom at this point. Rooms all over the place do it and we have a whole client that creates all rooms with custom PL of 101 and can in theory as far as i understand go to any arbitrary number that is legal and that client is Cinny. Like we are well beyond the point of special casing specific PLs and should just go with a proposal like #3949 instead of this proposal.

BLumia: > The important thing is the word "accidentally".

Yes, and actually when I introduce Matrix to my workmates, almost all of them (who need to be the owner of a room) fall into this pit without realizing the problem. I also got into this pit the first time I used Matrix!

So quoted from this MSC:

>  * We could ignore it, and try to educate admins into understanding that admin = owner, and they should set custom PLs
>   if they want something else.  In practice this simply hasn't worked, hence this MSC.

In practice this simply hasn't worked.

My suggestion is just simply trying to avoid that from happening by introducing one and only PL>100 special case and all other places stay the same.

> we have a whole client that creates all rooms with custom PL of 101 and can in theory as far as i understand go to any arbitrary number that is legal and that client is Cinny

In fact, I think Cinny's current implementation is the most ideal way to do this as long as the user selects the "Founder" role when creating a new room. To assign another user's PL, the drop down keeps in the range 0, 50 and 100. They need to manually enter 101 to get a mistake but users won't do that unless they know what it means. The same applies to the Permission tab since all dropdowns have 0, 50 and 100 as default so 101 won't get used without explicitly entering the value.

And we can amend my suggestion as well. It's okay to allow multiple owners (and currently already possible) but all I mean is introduce a way to reduce the possibility that first-time users falling into the pit. The "stick to range 0~100" is one of the ways to do that. It's okay if the client could still explicitly assign a value but *we should document the suggested range* to let the client developers so they will know what should avoid to be implemented.

Since most clients follows the 0~100 default PL range thing, adding PL101 (or 150, or 999, or whatever the PL number is. Let's assume we use 101 for now) could be the simplest approach to do so, client developers who want to adapt this spec only needs to do:

1. Add an option to use PL101 when creating a room.
2. Keep all the permission parts the same as before (stick to 0~100 range).
3. Ensure there will be a special warning dialog when user trying to assign PL101 to another user.

If we do this, the server side won't need to do anything, and we only need to document this suggested limitation in the spec so developers could treat this as a sort of thing as an agreement to follow.

su-ex: It seems were not that far apart, all you want different is
- have 101 instead of 150 (this is indeed arbitrary, but having something "in between to 100" means additional flexibility for use cases we can't foresee right now)
- have clients mandatorily stick to 0 to 100 in the UI (if clients want to hide the owner level and show are warning or only show a warning or show an info and a warning etc. should be up to the client imo, the only important thing is that certain values are specced out so they mean the same and are named the same across clients)

BLumia: > (this is indeed arbitrary, but having something "in between to 100" means additional flexibility for use cases we can't foresee right now)

Agree! 101, 150 or 999 or any other values are okay for me (I choose 101 just because it's weird enough so other people won't think it's a regular PL value that should be applied elsewhere, but go with 150 is also okay), as long as...

> have clients mandatorily stick to 0~100 in the UI (if clients want to hide the owner level and show a warning or only show a warning or show an info and a warning etc. should be up to the client imo, the only important thing is that certain values are specced out so they mean the same and are named the same across clients)

Yeah that's my main suggestion :)

ara4n:
@mscbot fcp close

given this has been obsoleted by https://github.com/matrix-org/matrix-spec-proposals/pull/4289 which effectively introduces the concept of ownership by acknowledging the creator of the room as being the owner, letting them delegate to another owner via upgrades.

PR #3919, opened 2022-10-28
Placeholder MSC

turt2live:


proposals/3919-ietf-mimi-message-format.txt
```diff
@@ -0,0 +1,311 @@
...
+
+   *  Presence
+
+   *  1:1 VoIP signalling
+
+   *  Multiparty VoIP signalling
+
+   Matrix events are extensible, and proposals exist for additional
+   event formats ranging from attaching 3D world geometry to
+   conversations (for openly standardized metaverse communication)
```
ShadowJonathan, line 140: I would advice against using "metaverse" as a buzzword here, as it has a lot of negative connotation, and for my feeling does not properly define what thirdworld.io is "properly" doing.

I can understand trying to reclaim the term, but I would suggest something more generic like "virtualised social environments"

proposals/3919-ietf-mimi-message-format.txt
```diff
@@ -0,0 +1,311 @@
...
+
+   Events are JSON objects which by default follow the formal schemas
+   defined in the Matrix Client Server API [MxEvents], also available as
+   JSON Schema definitions [MxEventsSchema].  Events are extensible and
+   may contain additional off-schema prefixed fields, or use prefixed
+   event types not yet defined in the spec.  Events then get augumented
+   and signed by the server before being forwarded to other servers/
+   users in the room.
+
+   These JSON objects have a few key fields:
```
ShadowJonathan, line 153:
```suggestion
   These JSON objects have a few key top-level fields:
```

proposals/3919-ietf-mimi-message-format.txt
```diff
@@ -0,0 +1,311 @@
...
+
+2.  Matrix Events
+
+   Events are JSON objects which by default follow the formal schemas
+   defined in the Matrix Client Server API [MxEvents], also available as
+   JSON Schema definitions [MxEventsSchema].  Events are extensible and
+   may contain additional off-schema prefixed fields, or use prefixed
+   event types not yet defined in the spec.  Events then get augumented
+   and signed by the server before being forwarded to other servers/
+   users in the room.
```
ShadowJonathan, line 151: This section seems to allude that the top-level event field is extensible, while in reality only the `type` string and `content` object is truly user-extensible. All other changes require room versioning and whatnot.

proposals/3919-ietf-mimi-message-format.txt
turt2live: <small>*thirdroom</small>

I think for the larger audience of this draft the term "metaverse", and all its connotation, makes sense.

proposals/3919-ietf-mimi-message-format.txt
turt2live: we'd make this clarification as part of a more proper definition of events (also it's complicated to accept suggestions of minor edits)

proposals/3919-ietf-mimi-message-format.txt
turt2live: They are extensible at the top level though, without the need for a room version. As long as the minimum schema is met, you can add whatever other fields you want. It just might not work how you expect :)

proposals/3919-ietf-mimi-message-format.txt
ShadowJonathan: Fair enough then, I'm just afraid it wouldn't, given my own background on the definition.

proposals/3919-ietf-mimi-message-format.txt
ShadowJonathan: I feel like that doesn't meet the definition of "freely extensible", which this document (and the general premise of matrix) alludes to, thus my comment.

turt2live: how so? By being able to add information to all levels of the event it's extensible?

ShadowJonathan: Ah, yes, but I might've taken the definition of "user-extensible" over "change-process-extensible", as user-extensibility implies little involvement from matrix.org (or no involvement in some cases) for independently defined formats, which is what I took for "extensible" here.

turt2live: "Extensible" here fits all definitions: 
* User-extensible: You can define your own event types and reuse them, and other people's, to communicate with. You can also extend the top-level event fields with your own data if you so choose, such as by populating `unsigned` or putting server-specific information at the top level (given clients can't normally modify the top level, but a server or client which doesn't use the Client-Server API can).
* Change-process-extensible: Sufficiently reusable extensions to events (at any level) can be proposed for formal inclusion in the spec. In future, we'll also have a process for "third-party" specifications to handle purpose-specific event extensions (medical imaging, etc).
* Implementation-extensible: As mentioned above, it's possible for servers (and custom clients) to extend events however they like.

ShadowJonathan: Right, it seems i have confused [redaction](https://spec.matrix.org/v1.4/rooms/v1/#redactions) with restriction of top-level keys, apologies. `</res>`

PR #3922, opened 2022-11-01
Removing SRV records from homeserver discovery

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/del-srv/proposals/0000-remove-srv-discovery.md)

**Note**: This is not eligible for FCP until enough time has passed to allow legitimate uses of SRV records to be identified. This is considered the implementation requirement for this MSC.

### DIFF
--- a/proposals/0000-remove-srv-discovery.md
+++ b/proposals/0000-remove-srv-discovery.md
@@ -1,0 +1,84 @@
+# MSC0000: Removing SRV records from homeserver discovery
+
+Currently when [resolving server names](https://spec.matrix.org/v1.4/server-server-api/#resolving-server-names),
+homeservers (or any implementation trying to locate a server, such as integration managers or widgets
+using [OpenID Connect validation](https://spec.matrix.org/v1.4/server-server-api/#openid)) must support
+an ability to resolve SRV DNS records. Aside from this being difficult in the case of widgets (for example),
+SRV records typically cause deployment issues due to them not working "as expected" by server administrators.
+
+In addition to SRV records not "properly" supporting CNAMEs, TLS certificates are difficult to configure
+correctly and often lead to issues with the wrong certificate being presented. These sorts of issues
+come up often enough that [Synapse's documentation](https://matrix-org.github.io/synapse/v1.70/delegate.html#srv-dns-record-delegation)
+doesn't even explain how to use SRV records, instead referencing the specification itself and citing that
+.well-known is often what administrators are looking for. The documentation additionally calls it
+"SRV delegation", further indicating that the use of SRV records is complex (it's not true delegation,
+unlike what is possible with .well-known).
+
+This proposal removes all reference of SRV records from the homeserver discovery specification, and
+a plan to handle the rollout of such an invasive change.
+
+## Proposal
+
+In short, the [current rules](https://spec.matrix.org/v1.4/server-server-api/#resolving-server-names)
+which reference SRV records are deleted. This leads to the following discovery mechanism:
+
+*Note*: Some details, such as caching and certificate presentation, are excluded. They are unchanged.
+
+1. If the hostname is an IP literal, then that IP address should be used. If a port number is given then
+   it should be used, otherwise using port 8448. The `Host` header in the request is set to the server name
+   (which is the IP address), with port number if explicitly given.
+2. If the hostname is *not* an IP literal, but does have an explicit port, resolve the name using A or
+   AAAA records to an IP and use that with the explicit port. The `Host` header in the request is set to
+   the server name, with port number.
+3. If the hostname is *not* an IP literal, a regular HTTPS request is made to the .well-known endpoint
+   on that domain. The hostname presented by this endpoint is called the "delegated hostname" and repeats
+   discovery steps 1 & 2 above. It does not repeat step 3 (this step) as that could cause infinite loops
+   or needless delays in discovery.
+4. Server discovery fails and the server is presumed offline or invalid if it has not been resolved to
+   a usable IP and port by this step.
+
+Clearly this would cause disruption in the larger ecosystem as some servers might still be using SRV
+records to identify themselves. Readers of this proposal are encouraged to proactively change over to
+.well-known to identify if there are legitimate reasons for keeping SRV records, even if this proposal
+is still in a draft/unapproved state. 
+
+In order to not cause massive breaking changes in the ecosystem, this proposal first deprecates SRV
+discovery for a minimum of 1 calendar year from the time of the spec release itself. Afterwards, at the
+discretion of the Spec Core Team (SCT), SRV discovery can be removed without notice.
+
+Homeserver authors (Synapse, Dendrite, Conduit, etc) are encouraged to use the deprecation period to
+help their users transition to .well-known discovery, and if reading this proposal before it is accepted
+then to help identify any legitimate reason to keep SRV discovery in the specification (for example, a
+user of theirs is completely unable to switch to .well-known - the case would be discussed to determine
+if it's a reasonable blocker for this proposal). 
+
+The Matrix.org Foundation would also be engaged in helping users move over to .well-known through the
+normal channels (blog posts, changelog on Synapse, social media, etc). 
+
+## Potential issues
+
+As identified, this change could impact legitimate usage of SRV records for discovery - this proposal
+exists to give readers, homeserver authors, etc time to identify these cases before the proposal is put
+up for final comment period (thus proposing it for acceptance). If legitimate cases of SRV records are
+found, this proposal may be declined or rejected (per normal process).
+
+## Alternatives
+
+This may be a good time to design new discovery mechanisms, however that would have an even larger
+impact on the ecosystem. Additionally, .well-known appears to be the (current) industry standard
+for this mechanism.
+
+## Security considerations
+
+Removing SRV discovery could mean a higher rate of homeservers being delegated to third party providers
+or being targets of takeover attempts, however given Synapse (the most populous homeserver implementation)
+already strongly recommends .well-known over SRV, this issue is considered trivial in nature.
+
+## Unstable prefix
+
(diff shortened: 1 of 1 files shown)


neilalexander (commented):
(moved inline)

proposals/3922-remove-srv-discovery.md
```diff
@@ -0,0 +1,84 @@
+# MSC0000: Removing SRV records from homeserver discovery
```
FSG-Cat, line 1:
```suggestion
# MSC3922: Removing SRV records from homeserver discovery
```

This should fix the small omission of forgetting to put in the correct MSC identifier.

proposals/3922-remove-srv-discovery.md
```diff
@@ -0,0 +1,84 @@
...
+   AAAA records to an IP and use that with the explicit port. The `Host` header in the request is set to
+   the server name, with port number.
+3. If the hostname is *not* an IP literal, a regular HTTPS request is made to the .well-known endpoint
+   on that domain. The hostname presented by this endpoint is called the "delegated hostname" and repeats
+   discovery steps 1 & 2 above. It does not repeat step 3 (this step) as that could cause infinite loops
+   or needless delays in discovery.
+4. Server discovery fails and the server is presumed offline or invalid if it has not been resolved to
+   a usable IP and port by this step.
+
+Clearly this would cause disruption in the larger ecosystem as some servers might still be using SRV
```
neilalexander, line 40: On the whole I worry that this will create a *significant* headache for if the Matrix federation protocol changes sufficiently enough to no longer be HTTP-centric (which I really hope it will).

I have been advocating for some research into a binary federation protocol for ages now to get around the computational expense, wire bloat and signing difficulties that JSON has. This puts quite a roadblock in the way of that as it still mandates that the discovery part of the stack is HTTP+TLS even if no other part of the stack is.

DNS SRV might be awkward to configure but it's otherwise protocol-agnostic. I don't think it is wise at all to remove it.

proposals/3922-remove-srv-discovery.md
turt2live: Thanks - I always forget at least one of the 47 steps.

proposals/3922-remove-srv-discovery.md
turt2live: This MSC is part 1 of a fairly long series tied in with the IETF/MIMI work we're doing. Specifically, we're aiming to separate transport from the protocol by defining a "Matrix over HTTP+JSON" thing, which would naturally include discovery. The discovery mechanism itself might still be http-centric, however this is not a requirement that the entire federation protocol be over http.

richvdh: It feels to me like a different transport mechanism would necessitate a different discovery mechanism. SRV records don't provide a way for a server to say "actually you can talk to me over CoAP instead of HTTPs", so I don't really understand what benefit they provide in the current ecosystem.

On the other hand, one of the reasons SRV records currently suck is that they interact poorly with HTTPs. (Or at least, they don't interact in the intuitive way.) Everyone has expectations about how HTTPs and DNS interact, and SRV records don't follow the pattern.

With a different transport mechanism, perhaps those expectations will be different, in which case we can consider using SRV records for that transport. But I don't see why that means we need to keep SRV records for HTTP-based transports.

jhulkko: I would like to point out that SRV is currently problematic for e.g. Dendrite and all other implementations using golang. Go does not try to handle domain name compression gracefully, when resolvers and/or authoritative NSes act against the RFC with TXT and SRV records. I've run into this issue with multiple HSes in the past when debugging obscure cases of broken federation. Some of the relevant details of the behaviour can be found e.g. here: https://github.com/golang/go/issues/10622

Additionally I've seen cases where SRV works "by accident" because an admin didn't request separate certificates as they had originally planned to, thus having the required hostname included correctly while their intention was to use separate certificates instead. Point being, the SRV is hard to get right as it doesn't follow conventions familiar with the HTTPS.

Additional or differently formatted SRV record would be required for other than current HTTPS based federation service discovery anyway, so designing and adding such method back later is IMO rather obvious possibility, but not blocker for this proposal. Getting rid of SRV for HTTPS based federation would avoid many problems and time wasted, so the relevant consideration is, if all practical use-cases can be met with the well-known based discovery alone.

I would suggest adding configuration option to log HSes that were discovered over SRV, enabling easy gathering of real life usage information by participating HS admins. Additionally the log would help detecting those cases where Go dns resolution fails to serve its purpose and admin is debugging federations problems encountered by e.g. Dendrite. 

Unfortunately the mis-handling of domain name compression can happen at either end, so this cannot be fully solved by adding test functionality to federationtester only. Properly working caching resolver can hide the issue, or badly behaving one could introduce it regardless of the correct response from the authoritative NS. Unfortunately I don't have statistics or further information of those NSes that were involved in those cases that I have had to debug. Golang has decided not to mitigate the issue unlike practically every other implementation has seemed to do, so yes, it's (still) the DNS, again.

slumericanB (approved):

slumericanB (approved):

Saklad5:
I am using SRV records myself, mainly because HTTPS records are still a draft spec and not widely implemented.

I strongly encourage replacing this with HTTPS records before removing them.

Saklad5:
Here's my current setup, which relies on HTTPS records for client-server interaction (my client supports them) and SRV records for server-server interaction.

I have fallback A/AAAA records for <saklad5.com>, but for the moment I'm assuming/pretending they are not being relied on. Those aside, I also rely on HTTPS records for the well-known URIs.

---

# Resolution

```dns
;; QUESTION SECTION:
;saklad5.com.			IN	HTTPS

;; ANSWER SECTION:
saklad5.com.		172800	IN	HTTPS	1 xuahkwjssci42ywuenj5zvn5jdm4o5zcgrqqhbs25sd75dhmz6yyvmqd.onion. alpn="h2"
saklad5.com.		172800	IN	HTTPS	2 auk.saklad5.com. alpn="h3,h2"

;; ADDITIONAL SECTION:
auk.saklad5.com.	172800	IN	A	159.223.125.173
auk.saklad5.com.	172800	IN	AAAA	2604:a880:400:d0::2204:1
auk.saklad5.com.	172800	IN	HTTPS	0 .
```

## Client

https://saklad5.com/.well-known/matrix/client:

```json
{"m.homeserver":{"base_url":"https://matrix.saklad5.com"}}
```

```dns
;; QUESTION SECTION:
;matrix.saklad5.com.		IN	HTTPS

;; ANSWER SECTION:
matrix.saklad5.com.	86400	IN	HTTPS	1 45ebt6an43rs6btnrgwjlgrpohjgmvjftwvojzbaz542e2ml3cmrmzqd.onion. alpn="h2"
matrix.saklad5.com.	86400	IN	HTTPS	2 dove.saklad5.com. alpn="h2"

;; ADDITIONAL SECTION:
dove.saklad5.com.	172800	IN	A	45.33.37.128
dove.saklad5.com.	172800	IN	AAAA	2600:3c01::f03c:93ff:fe6a:83b3
dove.saklad5.com.	172800	IN	HTTPS	0 .
```

## Server

https://saklad5.com/.well-known/matrix/server:

```json
{"m.server":"matrix.saklad5.com"}
```

```dns
;; QUESTION SECTION:
;_matrix._tcp.matrix.saklad5.com. IN	SRV

;; ANSWER SECTION:
_matrix._tcp.matrix.saklad5.com. 86400 IN SRV	0 0 443 45ebt6an43rs6btnrgwjlgrpohjgmvjftwvojzbaz542e2ml3cmrmzqd.onion.
_matrix._tcp.matrix.saklad5.com. 86400 IN SRV	1 0 443 dove.saklad5.com.

;; ADDITIONAL SECTION:
dove.saklad5.com.	172800	IN	A	45.33.37.128
dove.saklad5.com.	172800	IN	AAAA	2600:3c01::f03c:93ff:fe6a:83b3
```

Saklad5:
I do encourage replacing SRV records with SVBC/HTTPS records: among other benefits, they aren't tied to TCP specifically. This is not a small issue, as servers with both HTTP/2 and HTTP/3 will be listening on UDP as well.

I also encourage including DANE support in the spec, since not everyone can use DNS-01 ACME challenges like I'm doing for this.

Saklad5:
> @Saklad5 please move your comments to threads on the diff for consideration.

You mean under the "Files changed" tab, in a comment?

turt2live:
> > @Saklad5 please move your comments to threads on the diff for consideration.
> 
> You mean under the "Files changed" tab, in a comment?

Yes.

Saklad5 (changes requested):
I am using SRV records myself, mainly because HTTPS records are still a draft spec and not widely implemented.

I strongly encourage replacing this with HTTPS records before removing them.

proposals/3922-remove-srv-discovery.md
```diff
@@ -0,0 +1,84 @@
+# MSC3922: Removing SRV records from homeserver discovery
```
turt2live, line 1: for abundance of clarity: this MSC is currently extremely low on the priorities list, and is leaning towards rejection rather than acceptance.

TheArcaneBrony: This MSC just came up in [#conduit:fachschaften.org](https://matrix.to/#/%21SMloEYlhCiqKwRLAgY%3Afachschaften.org/%24moBdp-5P1NvnxUyYwP3EjUxgsuuPV4dK1eyvJXT3Fak?via=flipdot.org&via=matrix.org&via=conduit.rs), and as an SDK & power user client developer, this is something that I would absolutely love to see getting removed as I have functionality that eg. explicitly depends on /_matrix/federation/v1/version. Getting this in a browser is currently a big pain point.

Adding that endpoint to the Client-Server API would help, but that doesn't solve eg. being able to look up server keys or other future extensions to the federation APIs.

PR #3924, opened 2022-11-03
MSC3924: Capability-style access control for Matrix media

cvwright:
[Rendered](https://github.com/cvwright/matrix-spec-proposals/blob/namespacing_media/proposals/3924-capability-access-control-for-media.md)

I saw MSC3910 and MSC3911, and those got me thinking about alternative approaches for protecting access to media in the Matrix content repository.

This proposal suggests a different way of looking at the problem, inspired by ideas from capability-based access control.


cvwright:
Note that this is still a draft, and focuses only on the client-server part of the problem for now.

proposals/3924-capability-access-control-for-media.md
```diff
@@ -0,0 +1,350 @@
...
+file sharing capability.
+
+### Listing the media for a room
+
+The client can get a list of all media in the room by sending
+* `GET /_matrix/client/{version}/rooms/{roomId}/media`
+* `GET /_matrix/client/{version}/rooms/{roomId}/media/{serverName}`
+
+The server should reply with a JSON object containing a single element,
+`media_ids`, an array if `media_id` strings.
```
deepbluev7, line 120: This sentence is a bit confusing. Wouldn't this always return an array? Also: 
```suggestion
The server should reply with a JSON object containing a single element,
`media_ids`, an array of `media_id` strings.
```

deepbluev7: Ah, you meant the object has only a single field, which is the media_ids array. You probably want to think about pagination.

turt2live:
@cvwright don't forget to sign off on the changes for if this gets accepted please :)

proposals/3924-capability-access-control-for-media.md
```diff
@@ -0,0 +1,352 @@
...
+
+The approach proposed above does not fit well with this requirement.
+When a user leaves a room, they lose their access to all media in that
+room.
+
+In the following section, this proposal outlines an alternative approach
+that addresses this issue by granting media access to events rather than
+to rooms.
+
+## Modified Approach: Granting Access via Events Instead of via Rooms
```
uhoreg, line 258: I think that we need to do something that's event-based because a user in a room may not be allowed to see all the events, and so should not be allowed to see all the media.  In addition, if access is only room-based, then new endpoint for listing media in a room may leak information that people should not have access to.

proposals/3924-capability-access-control-for-media.md
```diff
@@ -0,0 +1,352 @@
...
+The client can get a list of all media in the room by sending
+* `GET /_matrix/client/{version}/rooms/{roomId}/media`
+* `GET /_matrix/client/{version}/rooms/{roomId}/media/{serverName}`
+
+The server should reply with a JSON object containing a single element,
+`media_ids`, an array of `media_id` strings.
+
+```json
+{
+    "media_ids": ["abc", "def", "xyz"]
```
uhoreg, line 126: shouldn't these be `mxc` URIs rather than media IDs?

cvwright:
I think MSC3911 covers all the important bits of this, and that one seems to be getting traction, so I'm closing this one

PR #3925, opened 2022-11-03
MSC3925: m.replace aggregation with full event

benkuly:
[rendered](https://github.com/benkuly/matrix-spec-proposals/blob/main/proposals/3925-replace-aggregation-with-full-event.md)

Instead of always complaining, here is my first MSC :wink: 

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3925#issuecomment-1380150062)
Implementations: [synapse](https://github.com/matrix-org/synapse/pull/14811), [ele-web](https://github.com/matrix-org/matrix-js-sdk/pull/3045)


turt2live:
@benkuly don't forget to sign off on the changes for if this eventually gets accepted please :)

benkuly:
> @benkuly don't forget to sign off on the changes for if this eventually gets accepted please :)

Thank you, is done now.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
+# MSC3925: m.replace aggregation with full event
+
+The Matrix DAG has immutable PDUs forming an auth chain. The client representation of this PDUs are simply called events.
+While this events also have been immutable until v1.3, since v1.4 they aren't.
```
jplatte, line 4:
```suggestion
While these events also have been immutable until v1.3, since v1.4 they aren't.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+## Security considerations
+
+
+
+## Unstable prefix
+
+I'm not sure, if we need an unsable prefix, because the aggregation would just be extended by additional fields.
```
jplatte, line 38:
```suggestion
I'm not sure, if we need an unstable prefix, because the aggregation would just be extended by additional fields.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+but only `event_id`, `origin_server_ts` and `sender` are included.
+Theoretically this is enough to get the replacing content, but when the event with the `event_id` cannot be found locally it needs to be fetched from the server.
+To prevent this additional call to the server, the `m.replace` aggregation should just contain the complete replacing event.
+
+The additional server call is already needed for encrypted events and would be saved by this proposal too.
+
+## Potential issues
+
+* There could be clients, which rely on the current behavior.
+* It is not as easy for clients like as in the current spec to get the current content of an event by just looking into `content.body`. While this true, it is also a relatively inconsistent behavior. Future replacements of the event would be rendered as "* new content". So the event with the replaced event does look different (without "*") despite the fact, that it is also replaced.
```
richvdh, line 26: please wrap your text at 80 characters.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
+# MSC3925: m.replace aggregation with full event
+
+The Matrix DAG has immutable PDUs forming an auth chain. The client representation of this PDUs are simply called events.
```
richvdh, line 3: they are called events in the DAG as well. "PDU" isn't used much, but when it is, it's specific to transmitting individual events over the federation API.

The refererence to "auth chains" is irrelevant here too.

Generally I don't think this line is adding much, and I suggest you get rid of it.

```suggestion
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+While these events also have been immutable until v1.3, since v1.4 they aren't.
```
richvdh, line 4: I think this is overstating the case. For one thing, it's always been possible to mutate events via redaction. Second, they are still immutable: it's just the server's representation of them which is mutating. (That representation is defined at https://spec.matrix.org/v1.4/client-server-api/#room-event-format, and isn't quite the same as the "raw event": note the presence of `event_id` for example)

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+When a client sends a `m.replace` relation, [the server should replace the content of the original event](https://spec.matrix.org/v1.4/client-server-api/#server-side-replacement-of-content).
+
+There are some issues with this requirement:
+* Changing the fundamental concept of immutable events is confusing. The server can respond with different event contents for the same `event_id`.
+* If an event with `m.replace` relation is deleted, the client would need to detect, if the original content was replaced and possibly needs to fetch the original content.
```
richvdh, line 9: events can't be deleted. Do you mean `redacted`?

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+* There is an additional server call needed, when the replacing event is encrypted, because the server cannot replace the original event content.
```
richvdh, line 10: I don't see how you can claim that an "additional server call" is needed, when it's unclear how a server can implement the spec as it is currently written anyway.

```suggestion
* Server cannot replace the original event content for encrypted events (because the 
  replacement content is inside the encrypted body).
  See [matrix-spec#1299](https://github.com/matrix-org/matrix-spec/issues/1299).
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+* There are also some other issues with this spec paragraph, which are discussed [here](https://github.com/matrix-org/matrix-spec/issues/1299)
```
richvdh, line 11: Spell them out for us, please? I'm not sure what other issues you are referring to.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+In fact it is [already used](https://spec.matrix.org/v1.4/client-server-api/#server-side-aggregation-of-mreplace-relationships), 
+but only `event_id`, `origin_server_ts` and `sender` are included.
+Theoretically this is enough to get the replacing content, but when the event with the `event_id` cannot be found locally it needs to be fetched from the server.
+To prevent this additional call to the server, the `m.replace` aggregation should just contain the complete replacing event.
+
+The additional server call is already needed for encrypted events and would be saved by this proposal too.
+
+## Potential issues
+
+* There could be clients, which rely on the current behavior.
```
richvdh, line 25:
```suggestion
* There could be clients which rely on the current behavior.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Proposal
+
+Instead of replacing the original content of an event, servers should use the aggregation feature for it.
+In fact it is [already used](https://spec.matrix.org/v1.4/client-server-api/#server-side-aggregation-of-mreplace-relationships), 
+but only `event_id`, `origin_server_ts` and `sender` are included.
+Theoretically this is enough to get the replacing content, but when the event with the `event_id` cannot be found locally it needs to be fetched from the server.
+To prevent this additional call to the server, the `m.replace` aggregation should just contain the complete replacing event.
+
+The additional server call is already needed for encrypted events and would be saved by this proposal too.
```
richvdh, line 21: the additional server call is needed because Synapse isn't implementing the spec, and it's not specific to encrypted events: https://github.com/matrix-org/synapse/issues/12793.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Potential issues
+
+* There could be clients, which rely on the current behavior.
+* It is not as easy for clients like as in the current spec to get the current content of an event by just looking into `content.body`. While this true, it is also a relatively inconsistent behavior. Future replacements of the event would be rendered as "* new content". So the event with the replaced event does look different (without "*") despite the fact, that it is also replaced.
+
+## Alternatives
+
+
```
richvdh, line 30: we should probably mention @kegsay's alternative suggestion at https://github.com/matrix-org/matrix-spec/issues/1299#issuecomment-1290332433.

richvdh (changes requested):
Thanks! I agree with the proposal, but I'm not sure I agree with all the reasons, and the proposal could do with some clarifications.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Security considerations
+
+
+
+## Unstable prefix
+
+I'm not sure, if we need an unstable prefix, because the aggregation would just be extended by additional fields.
```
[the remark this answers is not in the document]
richvdh: the fact that we're just adding extra fields is no reason that we would not need an unstable prefix.

Still, it's hard to see how an unstable prefix could be used here.

proposals/3925-replace-aggregation-with-full-event.md
benkuly: yep

proposals/3925-replace-aggregation-with-full-event.md
benkuly: While I still think immutablilty is relevant, I understand what you mean. In fact, in Trixnity, I circumvented all the cases, where some fields are removed by just adding them to the json before the deserialization. I've always disliked this, because it breaks good typing in most languages.
I removed the sentence.

proposals/3925-replace-aggregation-with-full-event.md
benkuly: There are actually no more, thank you.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+When a client sends a `m.replace`
+relation, [the server should replace the content of the original event](https://spec.matrix.org/v1.4/client-server-api/#server-side-replacement-of-content).
+
+There are some issues with this requirement:
+
+* Changing the fundamental concept of mostly immutable events is confusing. The server
+  can respond with different event contents for the same `event_id`.
+* If an event with `m.replace` relation is redacted, the client would need to
+  detect, if the original content was replaced and possibly needs to fetch the
+  original content.
```
richvdh, line 12:
```suggestion
* If an event with `m.replace` relation is redacted, clients need to
  detect if the original content was replaced, and possibly need to fetch the
  original content.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
+# MSC3925: m.replace aggregation with full event
+
+When a client sends a `m.replace`
+relation, [the server should replace the content of the original event](https://spec.matrix.org/v1.4/client-server-api/#server-side-replacement-of-content).
```
richvdh, line 4:
```suggestion
As [currently specified](https://spec.matrix.org/v1.4/client-server-api/#server-side-replacement-of-content),
servers should replace the content of events that have been replaced via an `m.replace`
relation.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+* Changing the fundamental concept of mostly immutable events is confusing. The server
+  can respond with different event contents for the same `event_id`.
+* If an event with `m.replace` relation is redacted, the client would need to
+  detect, if the original content was replaced and possibly needs to fetch the
+  original content.
+* Server cannot replace the original event content for encrypted events (because the
+  replacement content is inside the encrypted body).
+  See [matrix-spec#1299](https://github.com/matrix-org/matrix-spec/issues/1299).
+
+There is a discussion about this spec paragraph [here](https://github.com/matrix-org/matrix-spec/issues/1299).
```
richvdh, line 17: this is mentioned just above
```suggestion
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+  // irrelevant fields not shown
+}
+
+```
+
+## Potential issues
+
+* There could be clients which rely on the current behavior.
+* It is not as easy for clients like as in the current spec to get the current
+  content of an event by just looking into `content.body`. While this is true, it
```
richvdh, line 112:
```suggestion
* It will be harder for clients to get the current content of an event: currently
  they can just looking at `content.body`. While this is true, it
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+  is also a relatively inconsistent behavior. Future replacements of the event
+  would be rendered as "* new content". So the event with the replaced event
+  does look different (without "*") despite the fact, that it is also replaced.
```
richvdh, line 115: I'm afraid I'm not following this. When you say "Future replacements", do you mean events that arrive (over `/sync`) *after* the client receives the original event?

The term "rendered" refers the way a client takes the json and turns it into user-visible text. Is that really what you mean, or are you talking about the way a server returns the event?

> So the event with the replaced event does look different (without "*") 

I would certainly expect events *without* a replacement not to have a `*`, so this makes it sound as though events with replacements will look the same as those without? I don't think that's what you mean though?

Basically: please could you rewrite these two sentences.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+implemented) then they see the right thing, and when someone does care about
+them, they
+can just inspect unsigned to present the "edits" dialog.
+
+## Security considerations
+
+## Unstable prefix
+
+I'm not sure if we need an unstable prefix, because the aggregation would just
+be extended by additional fields.
```
richvdh, line 132:
```suggestion
No unstable prefix is currently proposed.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+We could flip the event stack on it's head and have the outer event
+be the edit and then in unsigned have the base event being edited. Currently, it
+is the inverse where we have the original event and then in unsigned the newer
+event sits. That way, if someone doesn't care about edits (because not
+implemented) then they see the right thing, and when someone does care about
+them, they
+can just inspect unsigned to present the "edits" dialog.
+
+## Security considerations
+
```
richvdh, line 128:
```suggestion

None currently foreseen.

```

richvdh (changes requested):

proposals/3925-replace-aggregation-with-full-event.md
benkuly: see https://github.com/matrix-org/matrix-spec-proposals/pull/3925/commits/5bfffb4ddf7c4da863ca71bdd1eaab26dd272239

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,132 @@
...
+## Alternatives
+
+We could flip the event stack on it's head and have the outer event
+be the edit and then in unsigned have the base event being edited. Currently, it
+is the inverse where we have the original event and then in unsigned the newer
+event sits. That way, if someone doesn't care about edits (because not
+implemented) then they see the right thing, and when someone does care about
+them, they
+can just inspect unsigned to present the "edits" dialog.
```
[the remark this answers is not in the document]
benkuly: You are right. Thank you :)

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -108,25 +108,27 @@ This is how the original event would look like after the replacement:
 ## Potential issues
 
 * There could be clients which rely on the current behavior.
-* It is not as easy for clients like as in the current spec to get the current
-  content of an event by just looking into `content.body`. While this is true, it
-  is also a relatively inconsistent behavior. Future replacements of the event
-  would be rendered as "* new content". So the event with the replaced event
-  does look different (without "*") despite the fact, that it is also replaced.
+* It will be harder for clients, which do not support replacing events to get
+  the current content of an event: currently they can just look at `content.body`.
```
richvdh, line 112: Ok, point of English grammar: in English we don't use a comma here. If you include a comma, then "which" refers to the *whole* of the previous phrase ("It will be harder for clients") instead of just "clients". 

So, this is grammatically correct, though not what you mean here:

> "[Some action] will be harder for clients, which will [have some consequence]", eg "Displaying events will be harder for clients, which will be annoying for client developers"

In general, I think we use fewer commas in English than you think we do :)

```suggestion
* It will be harder for clients which do not support replacing events to get
  the current content of an event: currently they can just look at `content.body`.
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -108,25 +108,27 @@ This is how the original event would look like after the replacement:
...
+  However, this leads to a relatively inconsistent behavior:
+
+  Assuming that we have an event A with the body `1`, which is first replaced by
```
richvdh, line 115:
```suggestion
  Assume that we have an event A with the body `1`, which is first replaced by
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -108,25 +108,27 @@ This is how the original event would look like after the replacement:
...
+  the event A' with the new body `2` and then by the event A'' with the new body
+  `3`. The sync has been done right after A' und A''. Then the timeline
```
richvdh, line 117:
```suggestion
  `3`. A `/sync` is done right after A' and A''. Then the timeline
```

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -108,25 +108,27 @@ This is how the original event would look like after the replacement:
...
+  would look like this: `2` -> `*3` instead of `1` -> `*2` -> `*3`. Although `2`
+  is a replaced body, it does not have a `*`, which looks to the user as it has 
+  not been replaced.
```
richvdh, line 120: Sorry, still not following. Please can you break it down a bit more.  What do you mean by "the timeline would look like this: `2` -> `*3`" ? What does `->` represent here?

After A', the client receives a /sync like this:

```yaml
timeline:
  - { event_id: "A", content: {body: "1"} }
  - { event_id: "A'", content: {body: "* 2"} }
```

Right?

----

If I spent a long time staring at your proposal and thinking about what you are trying to say, I could probably figure it out.  But for a proposal to be accepted, everybody else needs to understand it, and you can't expect everyone to spend hours trying to reverse-engineer your proposal.

In other words, please remember your audience, and explain yourself clearly.

richvdh (changes requested):

proposals/3925-replace-aggregation-with-full-event.md
benkuly: Sorry for the effort and all the commas.

Is this better? I think I would put that into the introduction, because it is also an issue of the current spec.

* Replacing the content can lead to inconsistent behavior on clients which don't
  support replacing events.
  Assume that we have this timeline on the server:
    - `{"event_id": "E1", "content": {"body": "1"}}`
    - `{"event_id": "E2", "content": {"body": "* 2", "m.new_content": {"body": "2"}}` // replaces `E1`

  A `/sync` is done after `E2`. We will have this timeline on a client:
    - `{"event_id": "E1", "content": {"body": "2"}}`
    - `{"event_id": "E2", "content": {"body": "* 2"}`

  Although `2` is a replaced body it does not have a `*`. This looks to the
  user as if it is the original content of the event.

richvdh: Yes, that's much clearer, thanks. Good idea to put it in the introduction.

richvdh (approved):
looks good!

The next step is to get it implemented so that we can see it working.

richvdh:
This now has server-side (https://github.com/matrix-org/synapse/pull/14811) and client-side (https://github.com/matrix-org/matrix-js-sdk/pull/3045) implementations. I believe it is ready for FCP.

@mscbot fcp merge

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+## Proposal
+
+The following two changes are proposed:
+1. The [server-side aggregation of `m.replace` relationships](https://spec.matrix.org/v1.4/client-server-api/#server-side-aggregation-of-mreplace-relationships)
+   is extended to be the entire content of the most recent replacement event, formatted
+   as described in [Room Event Format](https://spec.matrix.org/v1.4/client-server-api/#room-event-format).
+   This ensures that the client will always have the most recent edit without having to
+   fetch it from the server.
```
turt2live, line 38: Sorry, I don't understand what is being described here: are we saying the replacement event applies to itself? (in which case, why bother with specifying new and old content separately?)

turt2live:
@mscbot concern Unclear what is being proposed

proposals/3925-replace-aggregation-with-full-event.md
richvdh: @turt2live the current wording seems pretty clear to me (though I had a hand in writing it, so maybe I'm biased).

The current spec, as linked, says:

> The aggregation format of `m.replace` relationships gives the `event_id`, `origin_server_ts`, and` sender` of the most recent replacement event.

We're changing it so that it is the *whole* of that replacement event, rather than just three fields.

turt2live:
@mscbot resolve Unclear what is being proposed

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,140 @@
...
+{
+  "event_id": "$edit_event",
+  "type": "m.room.message",
+  "content": {
+    "body": "* I really like *chocolate* cake",
+    "msgtype": "m.text",
+    "m.new_content": {
+      "body": "I really like *chocolate* cake",
+      "msgtype": "m.text",
+      "com.example.extension_property": "chocolate"
```
dbkr, line 70: Is this here to demonstrate that newly added properties appear on the related event and not on the original? It would b e good to clarify as it's not super obvious why it's here.

proposals/3925-replace-aggregation-with-full-event.md
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+  // irrelevant fields not shown
+}
+
+```
+
+## Potential issues
+
+* There could be clients which rely on the current behavior.
```
dbkr, line 122: Do we know of any in particular? What will the failure mode be in such clients and is it tolerable?

proposals/3925-replace-aggregation-with-full-event.md
richvdh: > Is this here to demonstrate that newly added properties appear on the related event and not on the original?

I think it's just an example, yes. Maybe it would be clearer just to remove it?

proposals/3925-replace-aggregation-with-full-event.md
richvdh: Well, element-web relied on it until https://github.com/matrix-org/matrix-js-sdk/pull/3045, and I suspect most other clients still do.

The failure mode is: suppose that some time ago, you have an event E, which was subsequently replaced by an event E'. Now, if you jump back to a bit of timeline that contains E but *not* E', then clients unaware of this change will show the original event E rather than the edited content E'.

My feelings are: (1) that is an edge-case; (2) it already happens for events other than `m.room.message` (including encrypted events) due to https://github.com/matrix-org/synapse/issues/12503.

All of this information should probably be in the MSC itself: @benkuly do you fancy updating it?

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1440

richvdh:
merged!

PR #3926, opened 2022-11-08
MSC3926: Disable server-default notifications for bot users by default

Half-Shot:


proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+
+In the case of the latter two, it's not useful to calculate notifications because the majority of
+application services and bots do not read notifications.
+
+Notifications are "cleared" when a read receipt is sent into the room, but neither bots or application
+services send read receipts in all cases. This can very easily lead to a lot of wasted disk space in some
+implementations.
+
+There is also the case that each notitification rule will take some amount of CPU time to process, which
+is entirely wasted if the account doesn't check notifications.
```
clokep, line 14:
```suggestion
In the case of the latter two, it's not useful to calculate notifications because the majority of
application services and bots do not read notifications. This leads to wasted processing time
as the homeserver will evaluate push rules to generate notification counts which will never
be read.

Notifications are "cleared" when a read receipt is sent into the room, but neither bots nor application
services send read receipts in all cases. This can easily lead toa wasted disk space or performance
issues if notifications are never cleared.
```

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+## Proposal
+
+A new body field `enable_predefined_push_rules` will be introduced to [`GET /_matrix/client/v3/register`](https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3register),
+which will allow users to opt-out of default notifications.
+
+When set, the homeserver MUST disable all rules defined in [predefined-rules](https://spec.matrix.org/v1.4/client-server-api/#predefined-rules),
+effectively thereby disabling the calculation of notifications. The user will still be able to enable these notification rules if they so wish.
+
+When the `type` of registration is `m.login.application_service`, the default value of `enable_predefined_push_rules`
+will be `false`.
```
clokep, line 28: This is not backwards compatible, technically.

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+Ultimately for these reasons, we should allow bots and appservices to opt-out of server-side notifications
+upon registration.
+
+## Proposal
+
+A new body field `enable_predefined_push_rules` will be introduced to [`GET /_matrix/client/v3/register`](https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3register),
+which will allow users to opt-out of default notifications.
+
+When set, the homeserver MUST disable all rules defined in [predefined-rules](https://spec.matrix.org/v1.4/client-server-api/#predefined-rules),
+effectively thereby disabling the calculation of notifications. The user will still be able to enable these notification rules if they so wish.
```
clokep, line 25:
```suggestion
When set, the homeserver MUST disable all [predefined push rules](https://spec.matrix.org/v1.4/client-server-api/#predefined-rules)
for this user by enabling the `.m.rule.master` rule. The user will still be able
to enable these notification rules if they so wish.
```

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+application service implementation, rather than a spec concern.
+
+## Alternatives
+
+### Only enable notifications on first /sync
+
+Instead of adding a flag, the homeserver could instead only start tracking notifications when the user
+/syncs. This might help with cases where users are registered but never used. It would also help with 
+application services where users never sync. However, this would still cause notifications to be calculated
+for traditional `/sync`ing bots. Overall an explicit option at registration time seems more preferable.
```
clokep, line 46:
```suggestion
Instead of adding a flag, the homeserver could instead only start tracking notifications when the user
makes a `/sync` request. This might help with cases where users are registered but never used. It would also help with 
application services where users never sync. However, this would still cause notifications to be calculated
for traditional `/sync`ing bots. Overall an explicit option at registration time seems more preferable.
```

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Bots and Appservices could explicitly disable notifications
+
+Bots and integrations could instead explicitly disable all rules on signup, rather than expecting the
```
clokep, line 50:
```suggestion
Bots and integrations could instead [explicitly enable the `.m.rule.master`](https://spec.matrix.org/v1.4/client-server-api/#predefined-rules)
rule on signup via a [`PUT /pushrules/global/override/.m.rule.master` request](https://spec.matrix.org/v1.4/client-server-api/#put_matrixclientv3pushrulesscopekindruleid),
rather than expecting the
```

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
...
+is considered more of a niche use case.
+
+
+## Security considerations
+
+None. The decision to enable or disable notifications is left up to the registering user.
+
+## Unstable prefix
+
+While this MSC is unstable, `enable_predefined_push_rules` should be called `org.matrix.mscXXXX.enable_predefined_push_rules`.
```
clokep, line 66:
```suggestion
While this MSC is unstable, `enable_predefined_push_rules` should be called `org.matrix.msc3926.enable_predefined_push_rules`.
```

proposals/3926-disable-bot-notifs-by-default.md
```diff
@@ -0,0 +1,74 @@
+# MSC3926: Disable server-default notifications for bot users by default
```
turt2live, line 1: fwiw, from title alone: this would be a very welcome change. t2bot.io, which offers largely bridge traffic, has [disabled](https://github.com/t2bot/synapse/commit/4ab280d29e77d7949c5b667e3b0d07875ff61973) push processing on synapse to speed up event handling paths - it feels like a natural next step to improve (at least) CPU usage for other massive bridge deployments.

Half-Shot: I understood from the Synapse team that they check users against the AS namespaces and short-circuit notification checks entirely. While that's *fine*, it's a bit divergent from the spec and we should seek to make it explicit/official.

PR #3927, opened 2022-11-13
MSC3927: Extensible Events - Audio

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/audio/proposals/3927-extensible-events-audio.md)

### DIFF
--- a/proposals/3927-extensible-events-audio.md
+++ b/proposals/3927-extensible-events-audio.md
@@ -25,7 +25,7 @@ An example is:
   "content": {
     "m.markup": [
       // Format of the fallback is not defined, but should have enough information for a text-only
-      // client to do something with the image, just like with plain file uploads.
+      // client to do something with the audio, just like with plain file uploads.
       {"body": "matrix.mp3 (12 KB, 1:30) https://example.org/_matrix/media/v3/download/example.org/abcd1234"}
     ],
     "m.file": {
--- a/proposals/3927-extensible-events-audio.md
+++ b/proposals/3927-extensible-events-audio.md
@@ -23,7 +23,7 @@ An example is:
 {
   "type": "m.audio",
   "content": {
-    "m.markup": [
+    "m.text": [
       // Format of the fallback is not defined, but should have enough information for a text-only
       // client to do something with the audio, just like with plain file uploads.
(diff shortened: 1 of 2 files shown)


proposals/3927-extensible-events-audio.md
```diff
@@ -0,0 +1,90 @@
...
+
+An example is:
+
+```json5
+{
+  "type": "m.audio",
+  "content": {
+    "m.markup": [
+      // Format of the fallback is not defined, but should have enough information for a text-only
+      // client to do something with the image, just like with plain file uploads.
```
HarHarLinks, line 28:
```suggestion
      // client to do something with the audio, just like with plain file uploads.
```

proposals/3927-extensible-events-audio.md
```diff
@@ -0,0 +1,90 @@
...
+      "url": "mxc://example.org/abcd1234",
+      "name": "matrix.mp3",
+      "mimetype": "audio/mp3",
+      "size": 12345
+    },
+    "m.audio_details": { // optional
+      "duration": 90
+    },
+    "m.caption": { // optional - goes above/below audio
+      "m.markup": [{"body": "Listen to the Matrix logo in all its glory"}]
```
HarHarLinks, line 41: 🤔

HarHarLinks (commented):
missed a mention of "image" 🙃

---
what! I used the view feature wrongly again, sorry.

proposals/3927-extensible-events-audio.md
```diff
@@ -0,0 +1,90 @@
...
+      {"body": "matrix.mp3 (12 KB, 1:30) https://example.org/_matrix/media/v3/download/example.org/abcd1234"}
+    ],
+    "m.file": {
+      "url": "mxc://example.org/abcd1234",
+      "name": "matrix.mp3",
+      "mimetype": "audio/mp3",
+      "size": 12345
+    },
+    "m.audio_details": { // optional
+      "duration": 90
```
yncyrydybyl, line 38: A bit more precision and using milliseconds would be good approach. That especially comfortable if the files are used to combine for notification sounds or other programmatic use cases.

HarHarLinks: I agree, and milliseconds is also what Element already implements

PR #3381, opened 2021-09-07
MSC3381: Polls (mk II)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3381-polls.md)

**Note**: the implementations below are for an earlier draft of this proposal. The latest draft does not technically have any implementations, but is not functionally different from the prior draft. That prior draft, called "v1", is available [here](https://github.com/matrix-org/matrix-spec-proposals/blob/95fdc44b904d2b4d2f227db99050e539e43f3509/proposals/3381-polls.md).

Element Web Implementation:
* https://github.com/matrix-org/matrix-react-sdk/pull/7088 (rendering)
* https://github.com/matrix-org/matrix-react-sdk/pull/7166 (limits)
* https://github.com/matrix-org/matrix-react-sdk/pull/7158 (voting)
* https://github.com/matrix-org/matrix-react-sdk/pull/7202 (complexities of multi-device voting)
* https://github.com/matrix-org/matrix-react-sdk/pull/7233 (above part 2)
* https://github.com/matrix-org/matrix-react-sdk/pull/7269 (mechanics of showing results)
* https://github.com/matrix-org/matrix-react-sdk/pull/7305 (ending polls)

Element iOS Implementation:
* https://github.com/vector-im/element-ios/pull/5202 (rendering)
* https://github.com/matrix-org/matrix-ios-sdk/pull/1295 (mechanics)
* https://github.com/vector-im/element-ios/pull/5100 (creation)
* https://github.com/vector-im/element-ios/pull/5378 (undisclosed and edits)

Further work:
* events-sdk: https://github.com/matrix-org/matrix-events-sdk/pull/1
  * Element Web: [React](https://github.com/matrix-org/matrix-react-sdk/pull/7517), [JS](https://github.com/matrix-org/matrix-js-sdk/pull/2102)
* ruma: https://github.com/ruma/ruma/pull/1131
 
----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3381#issuecomment-1597912475)

No MSC checklist.

Status: "BLOCKED" due to requiring Extensible Events to fully land, which (currently) requires a room version.

Next steps: Land extensible events, or merge an MSC which avoids the dependency.

PR #3930, opened 2022-11-14
MSC3930: Polls push rules/notifications

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/polls-notifs/proposals/3930-polls-notifications.md)

Implementations:
* Synapse: https://github.com/matrix-org/synapse/pull/14787
* Syncing behaviour (an implementation detail under this MSC):
  * Element Web/react-sdk: https://github.com/matrix-org/matrix-react-sdk/pull/10287
  * Element Android: https://github.com/vector-im/element-android/pull/8114 & https://github.com/vector-im/element-android/pull/8130

Note: implementation for this MSC was done before `event_property_is` was specified/adopted. The only difference between the implementations above and the current version of the MSC is the condition kind - the meaning and purpose is otherwise the same.

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3930#issuecomment-1659574910)

### DIFF
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -74,7 +74,7 @@ Note that the push rules are mirrored from those available to `m.room.message` e
 }
 ```
 
-Additionally, a new override rule is defined to supress poll responses by default:
+Additionally, a new override rule is defined to suppress poll responses by default:
 
 ```json
 {
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -27,7 +27,7 @@ order presented by this MSC.
     {"kind": "room_member_count", "is": "2"},
 
     // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
-    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.start"}
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.start"}
   ],
   "actions": [
     "notify",
@@ -43,7 +43,7 @@ order presented by this MSC.
   "enabled": true,
   "conditions": [
     // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
-    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.start"}
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.start"}
   ],
   "actions": [
     "notify"
@@ -60,7 +60,7 @@ order presented by this MSC.
     {"kind": "room_member_count", "is": "2"},
 
     // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
-    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.end"}
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.end"}
   ],
   "actions": [
     "notify",
@@ -76,7 +76,7 @@ order presented by this MSC.
   "enabled": true,
   "conditions": [
     // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
-    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.end"}
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.end"}
   ],
   "actions": [
     "notify"
@@ -127,6 +127,9 @@ preferred approach over this MSC's (hopefully) short-term solution.
 While the order within the MSC is deliberate for the new rules, the positioning relative to other rules
 already in the spec is fairly arbitrary. The new rules are placed at the end of each list (underride and
 override) for simplicity, but could realistically go anywhere in the list.
+
+See ["Dot-separated property paths"](https://spec.matrix.org/v1.7/appendices/#dot-separated-property-paths)
+for more information on why we escape the `event_property_is` condition values.
 
 ## Security considerations
 
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -18,14 +18,16 @@ Note that [order matters](https://github.com/matrix-org/matrix-spec/issues/1406)
 underride rules are to be inserted immediately after the `.m.rule.encrypted` underride push rule, in the
 order presented by this MSC.
 
-```json
+```jsonc
 {
   "rule_id": ".m.rule.poll_start_one_to_one",
   "default": true,
   "enabled": true,
   "conditions": [
     {"kind": "room_member_count", "is": "2"},
-    {"kind": "event_match", "key": "type", "pattern": "m.poll.start"}
+
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.start"}
   ],
   "actions": [
     "notify",
@@ -34,13 +36,14 @@ order presented by this MSC.
 }
 ```
 
-```json
+```jsonc
 {
   "rule_id": ".m.rule.poll_start",
   "default": true,
   "enabled": true,
   "conditions": [
-    {"kind": "event_match", "key": "type", "pattern": "m.poll.start"}
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.start"}
   ],
   "actions": [
     "notify"
@@ -48,14 +51,16 @@ order presented by this MSC.
 }
 ```
 
-```json
+```jsonc
 {
   "rule_id": ".m.rule.poll_end_one_to_one",
   "default": true,
   "enabled": true,
   "conditions": [
     {"kind": "room_member_count", "is": "2"},
-    {"kind": "event_match", "key": "type", "pattern": "m.poll.end"}
+
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.end"}
   ],
   "actions": [
     "notify",
@@ -64,13 +69,14 @@ order presented by this MSC.
 }
 ```
 
-```json
+```jsonc
 {
   "rule_id": ".m.rule.poll_end",
   "default": true,
   "enabled": true,
   "conditions": [
-    {"kind": "event_match", "key": "type", "pattern": "m.poll.end"}
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "pattern": "m\\.poll\\.end"}
   ],
   "actions": [
     "notify"
@@ -81,13 +87,14 @@ order presented by this MSC.
 Additionally, a new override rule is defined to suppress poll responses by default, inserted immediately
 after the `.m.rule.room.server_acl` override rule.
 
-```json
+```jsonc
 {
   "rule_id": ".m.rule.poll_response",
   "default": true,
   "enabled": true,
   "conditions": [
-    {"kind": "event_match", "key": "type", "pattern": "m.poll.response"}
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.response"}
   ],
   "actions": []
 }
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -117,4 +117,4 @@ None applicable - no new considerations need to be made with this proposal.
 ## Unstable prefix
 
 While this MSC is not considered stable, implementations should use `org.matrix.msc3930.*` as a prefix
-in place of `m.*`.
+in place of `m.*` for the push rule IDs. As of writing, polls are only implemented using the legacy `org.matrix.msc3381.poll.*` prefix rather than the newer `v2` prefix - implementations of this MSC should be aware of which version of MSC3381 they plan to support.
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -6,13 +6,17 @@ This proposal aims to address that specific feature gap.
 
 Readers should review MSC3381 to better understand how polls are intended to operate in a room.
 
-Push rules are currently defined [here](https://spec.matrix.org/v1.5/client-server-api/#push-rules) in
+Push rules are currently defined [here](https://spec.matrix.org/v1.7/client-server-api/#push-rules) in
 specification.
 
 ## Proposal
 
-In order to have polls behave similar to message events, the following underride push rules are defined.
-Note that the push rules are mirrored from those available to `m.room.message` events.
+Polls should behave similar to message events. To achieve this effect, we define the following underride
+push rules which mimic their `m.room.message` partners.
+
+Note that [order matters](https://github.com/matrix-org/matrix-spec/issues/1406) for push rules - these
+underride rules are to be inserted immediately after the `.m.rule.encrypted` underride push rule, in the
+order presented by this MSC.
 
 ```json
 {
@@ -74,7 +78,8 @@ Note that the push rules are mirrored from those available to `m.room.message` e
 }
 ```
 
-Additionally, a new override rule is defined to suppress poll responses by default:
+Additionally, a new override rule is defined to suppress poll responses by default, inserted immediately
+after the `.m.rule.room.server_acl` override rule.
 
 ```json
 {
@@ -110,6 +115,12 @@ term need while other MSCs work on improving the notifications system. Most impo
 believes future MSCs which aim to fix notifications for extensible events in general will be a more
 preferred approach over this MSC's (hopefully) short-term solution.
 
+## General considerations
+
+While the order within the MSC is deliberate for the new rules, the positioning relative to other rules
+already in the spec is fairly arbitary. The new rules are placed at the end of each list (underride and
+override) for simplicty, but could realistically go anywhere in the list.
+
 ## Security considerations
 
 None applicable - no new considerations need to be made with this proposal.
@@ -117,4 +128,6 @@ None applicable - no new considerations need to be made with this proposal.
 ## Unstable prefix
 
 While this MSC is not considered stable, implementations should use `org.matrix.msc3930.*` as a prefix
-in place of `m.*` for the push rule IDs. As of writing, polls are only implemented using the legacy `org.matrix.msc3381.poll.*` prefix rather than the newer `v2` prefix - implementations of this MSC should be aware of which version of MSC3381 they plan to support.
+in place of `m.*` for the push rule IDs. As of writing, polls are only implemented using the legacy
+`org.matrix.msc3381.poll.*` prefix rather than the newer `v2` prefix - implementations of this MSC
+should be aware of which version of MSC3381 they plan to support.
--- a/proposals/3930-polls-notifications.md
+++ b/proposals/3930-polls-notifications.md
@@ -1,5 +1,5 @@
 
-# MSC0000: Polls push rules/notifications
+# MSC3930: Polls push rules/notifications
 
 [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381) describes how chat polls can work,
 though deliberately leaves out details on how push rules or notifications could work for such a feature.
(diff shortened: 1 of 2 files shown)


proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,103 @@
...
+  "actions": [
+    "notify"
+  ]
+}
+```
+
+Servers should keep these rules in sync with the `m.room.message` rules they are based upon. For
+example, if the `m.room.message` rule gets muted in a room then the associated rules for polls would
+also get muted. Similarly, if either of the two poll rules were to be muted in a room then the other
+poll rule and the `m.room.message` rule would be muted as well.
```
turt2live, line 78: @Johennes says:

> @turt2live: From a client-UI perspective, I think the synchronization between message and polls rules only makes sense if the client has a single toggle for both. That's a fair choice but, I think, not one we should be making for _all_ clients. Being able to change polls notification settings separately from message notification settings seems like a valid use case that is denied by the current proposal.
> 
> Maybe we should weaken this a little? Instead of having the server hard-wire the polls and messages rules, we could simply _recommend_ clients to toggle the polls rules together with the messages rules unless they expose fine-grained notification settings.

@Johennes says:

> @mnaturel I think this point is also relevant for your upcoming live location sharing push rules MSC.

@turt2live says:

> If there's a strong usecase for separating them, sure. Currently the intention is to deliberately treat them as room messages, setting the stage for how future extensible events will work.

@Johennes says:

> I'm not sure if maybe that could be too limiting? For instance, it doesn't seem unlikely that somebody would want to _not_ follow messages in a room but still wants to be notified when there's a poll to take part in.

@turt2live says:

> it's slightly more of a question if the clients are willing/able to accept the complexity of push rules, and whether they'd actually diverge away from a linked toggle. It's easier to separate the push rules server-side than it is to join them, for instance.

turt2live: My modern update on this (thread is from August 3, 2022) is I'd rather leave this decision to a backend or product team to make a decision, as these requirements are not founded on ideals from that world.

slumericanB (approved):

proposals/3930-polls-notifications.md
jakewb-b: My reasoning for originally favouring syncing on the client was:

1. It is not obvious that syncing poll notifications with message notifications is definitely the right thing to do, universally, across the Matrix ecosystem.
2. As Element, I’m quite happy making the decision that, on our clients, we are going to sync those notifications anyway because Element’s bias against excessively granular settings outweighs a desire to give users the ability to receive poll notifications but not message notifications, or vice versa. 
3. But, I am much less happy making that decision for all clients, because I don’t think it’s universally right in all circumstances. If I were building a client that had a different established bias in regards to user choice vs settings proliferation, or entirely different goals as a product (e.g. a social app focused around polls, with chat as a side-feature) I might make a different decision. 

So, I think syncing on the client is not only sufficient, but desirable. Unless there’s a compelling counter-argument, in which case I’m very happy to hear it of course!

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,103 @@
+
+# MSC3930: Polls push rules/notifications
```
turt2live, line 2: This needs to account for https://github.com/matrix-org/matrix-spec-proposals/pull/3932

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,103 @@
+
```
turt2live, line 1:
```suggestion
```

proposals/3930-polls-notifications.md
alfogrillo: Hey @turt2live, is it worth also adding a default override rule to avoid notifications for all the `m.poll.response` messages?

I think it should be an `override` because we want clients to still be able to setup rules for `room` and `sender` (eg. notify all the messages in a room) without getting notification for all poll votes.

What do you think?

turt2live: I believe they already should be muted by the other push rules?

alfogrillo: I don't think so.
If the client adds a `room` rule for "all the messages in room" it will get sounds whenever another users votes.
An override rule just for poll votes will possibly fix this problem?

Screenshot 2022-12-06 alle 12 39 41

turt2live: That's somewhat intentional, I think 😅 

"All Messages" has had an awkward definition over the years I think we'll need to work out elsewhere, but if a dedicated rule for muting poll responses is needed then we should add one, yes.

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,102 @@
...
+[MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381) describes how chat polls can work,
+though deliberately leaves out details on how push rules or notifications could work for such a feature.
+This proposal aims to address that specific feature gap.
+
+Readers should review MSC3381 to better understand how polls are intended to operate in a room.
+
+## Proposal
+
+In order to have polls behave similar to message events, the following underride push rules are defined.
```
mnaturel, line 11: Should we also decribe the priority of these new rules compared to the existing ones?

turt2live: Priority is done through categorization, I believe. By having these be underride rules, they are executed at the same time as all the other underride rules (which is fine: we filter for event type anyways).

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,120 @@
...
+preferred approach over this MSC's (hopefully) short-term solution.
+
+## Security considerations
+
+None applicable - no new considerations need to be made with this proposal.
+
+## Unstable prefix
+
+While this MSC is not considered stable, implementations should use `org.matrix.msc3930.*` as a prefix
+in place of `m.*`.
```
anoadragon453, line 120: I've run into a slight issue while implementing an unstable version of this MSC alongside an older implementation of MSC3381. MSC3381 [initially defined](https://github.com/matrix-org/matrix-spec-proposals/blob/32ab69a9208c2de12383db98e0ded53e010a1d1f/proposals/3381-polls.md#unstable-prefix) an unstable prefix of `org.matrix.msc3381.*`, and this was implemented in the Element clients.

Today, MSC3381 [defines](https://github.com/matrix-org/matrix-spec-proposals/blob/169afe0bfa22f946dc4b3d1c1f6f01c9e4d1bff1/proposals/3381-polls.md#unstable-prefix) an unstable prefix of `org.matrix.msc3381.v2.*`. When trying to write an unstable implementation of this MSC (MSC3930) in Synapse, I've hit a snag where I'd like to support both the old and current unstable prefixes of MSC3381, in order to support the Element clients. This would effectively mean implementing two separate sets of push rules:

1. A set of push rules with an unstable prefix for MSC3930 (`org.matrix.msc3930.*`) and the **old** unstable prefix for MSC3381 (`org.matrix.msc3381.*`).
2. A set of push rules with an unstable prefix for MSC3930 (`org.matrix.msc3930.*`) and the **current** unstable prefix for MSC3381 (`org.matrix.msc3381.v2.*`).

Other than polluting the environment with lots of many push rules, this works. And supporting both sets of push rules at once makes a client transition from the older version of MSC3381 to the newer one simple.

The only issue now is that this MSC defines a single set of rule IDs for unstable push rules: `org.matrix.msc3930.rule.*`. I would need another unstable prefix defined that is specifically catered to the difference between old and new versions of MSC3381's unstable prefixes, in order to differentiate the rule IDs of both sets. One suggestion is to define `org.matrix.msc3930.*` and `org.matrix.msc3930.v2.*` prefixes here, to correspond to MSC3381's `org.matrix.msc3381.*` and `org.matrix.msc3381.v2.*` prefixes specifically.

I appreciate that this will also tie the two MSCs more closely together.

turt2live: There is no implementation planned for the v2 unstable poll events - MSC3381 only mentions the prefix in case someone does implement it without realizing it. For this MSC's implementation, it should rely on the v1 unstable prefixes *only*.

anoadragon453: I see, so the only known planned further implementation of polls will be going straight to stable prefixes?

If that's the case, it makes things a lot simpler. Thanks for the context!

turt2live: indeed - the v2 changes are insignificant for FCP purposes, so we'll just go v1 -> stable when it's time.

anoadragon453, line 120: ([ohai](https://github.com/matrix-org/matrix-spec-proposals/pull/3381#discussion_r1067172120).)

While implementing this MSC in turn with a not-yet-accepted MSC3381, I've had to use unstable prefixes for MSC3381 (`org.matrix.msc3381.poll.*`). The current text of this MSC _technically_ doesn't allow that for an unstable implementation yet.

```suggestion
in place of `m.*`.

While [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381) has not
been accepted, `org.matrix.msc3381.poll.*` may be used as a prefix in place of `m.poll.*`
throughout this proposal.
```

(I think this is still a valid concern for this MSC...)

clokep: I was less concerned about it being valid and more just that different folks would implement the unstable version differently. (Which isn't a huge deal since "unstable", but it would be nice if it was documented somewhere at least!)

turt2live: TLDR: The interaction with MSC3381 is intentionally vague in this MSC, though can be clarified. Skip to the suggestion below and justification below it if interested.

----

As MSC3381 is a dependency, the unstable prefixes for the event types are implied, though the text in this MSC is indeed a bit strict (it could be argued that it takes over the event type prefixes, which is untrue/not sensible). There's the added challenge that MSC3381 has a v1 which is implemented and v2 which has no planned implementation: this can make it unclear which unstable prefix to use for unstable push rules exactly, however implementations of MSCs are always done at point-in-time rather than forwards-looking, meaning the unstable implementation only needs to cover what is "currently" possible.

In short, the suggestion feels a bit too codependent on MSC3381 rather than treating it as a dependency, so would suggest something like this instead:

```suggestion
in place of `m.*` for the push rules. As of writing, polls are only implemented using the legacy `org.matrix.msc3381.poll.*` prefix rather than the newer `v2` prefix - implementations of this MSC should be aware of which version of MSC3381 they plan to support.
```

Technically speaking, "for the push rules" is the only required change here, however I do agree that some words to describe the MSC3381 interaction is worthwhile. 

This suggestion does leave the door open for different implementations to use different event types for polls, however that was already a possibility if MSC3381 were to be accepted and this MSC not: implementations would switch to stable poll events and retain the unstable prefix for the push rules. It is left as an implementation detail to pick which event type they are representing in the push rules, though the implied suggestion is to rely on market share (in a sense): most users on Synapse will want the MSC3381v1 event types, for instance.

anoadragon453: Thanks for clarifying that it's an implementation detail as to what unstable prefix to use for event types in the `pattern` field of push rules defined in this MSC. That was not immediately obvious to me on first read, but your suggested text helps clarify that.

I would just clarify it a bit first to say "in place of `m.*` for the push rule **IDs**. ...", assuming I understand you correctly.

proposals/3930-polls-notifications.md
anoadragon453: But won't `.m.rule.poll_start_one_to_one` and `.m.rule.poll_start` (for example) both match if I sent a `m.poll.start` event in a room with >2 members? Does the spec care about that?

Or does this not matter because the actions of one rule are a strict subset of the other?

proposals/3930-polls-notifications.md
anoadragon453: Clickable suggestion to make this change easier on your end:

```suggestion
in place of `m.*` for the push rule IDs. As of writing, polls are only implemented using the legacy `org.matrix.msc3381.poll.*` prefix rather than the newer `v2` prefix - implementations of this MSC should be aware of which version of MSC3381 they plan to support.
```

proposals/3930-polls-notifications.md
turt2live: The 1:1 rules have `{"kind": "room_member_count", "is": "2"}`, which should prevent the server from matching that rule in rooms with >2 members.

clokep: My understanding was there's a strict ordering to the rules that need to be defined -- that was what was done in MSC3786 at least and matches the Synapse implementation.

I don't think they all "run at the same time" -- what would that mean if different rules came to different conclusions?

anoadragon453: Oops. Consider the case of a room with 2 members then. Both push rules would match.

turt2live: The spec appears to mention that rules can have priority, but doesn't mention how to actually achieve that.

richvdh: > The spec appears to mention that rules can have priority, but doesn't mention how to actually achieve that.

https://github.com/matrix-org/matrix-spec/issues/638, I think

anoadragon453: Along with https://github.com/matrix-org/matrix-spec/issues/1401.

anoadragon453: The spec does not appear to strictly define an ordering of push rules. There is an order in which they are presented in the spec, but that at least does not match Synapse's implementation. Also note that only some rules *require* an ordering - though I don't see any reason not to define an order for the entire set to eliminate any edge cases.

In the case of MSC3786 mentioned above, an order was defined in the MSC (must be calculated after `m.room.tombstone`) but other than its listing in the spec, a note of that ordering was not preserved in [the spec PR](https://github.com/matrix-org/matrix-spec/pull/1190).
 
@turt2live it would be appreciated if you could define an ordering in this MSC to unblock implementation. I would suggest simply "The `.m.rule.poll_start` push rule must be evaluated after `.m.rule.poll_start_one_to_one`. Similarly, `.m.rule.poll_end` must be evaluated after `.m.rule.poll_end_one_to_one`".

Otherwise I'd like to propose that we define an ordering of push rules in the spec (as a separate endeavour).

richvdh: > The spec does not appear to strictly define an ordering of push rules. There is an order in which they are presented in the spec, but that at least does not match Synapse's implementation

The spec certainly *intends* to define an ordering. The mismatch with Synapse appears to be a spec bug: filed as https://github.com/matrix-org/matrix-spec/issues/1406.

turt2live: > @turt2live it would be appreciated if you could define an ordering in this MSC to unblock implementation. 

I can't keep this MSC at the top of my list at the moment, so will defer ordering to the implementation to decide.

Please leave a comment here for what was chosen for eventual inclusion in the MSC.

anoadragon453: I appreciate the response regardless :slightly_smiling_face: 

The [Synapse implementation](https://github.com/matrix-org/synapse/pull/14787) has gone with the following ordering for the underride rules:

* `.org.matrix.msc3930.rule.poll_start_one_to_one`
* `.org.matrix.msc3930.rule.poll_start`
* `.org.matrix.msc3930.rule.poll_end_one_to_one`
* `.org.matrix.msc3930.rule.poll_end`

...due to the `*_one_to_one` variants containing more specific match conditions than their counter parts.

clokep: @anoadragon453 Can you also define *where in relation to the existing rules* those go? That's the part that was missing in my mind! I think the ordering above is implicit in the original MSC that lists them in an order. 😄

anoadragon453: Aha, good point :smile:

I therefore propose the following full ordering of all known underride rules after this change:

* ...
* `.m.rule.encrypted_room_one_to_one`
* `.m.rule.poll_start_one_to_one`
* `.m.rule.poll_end_one_to_one`
* `.m.rule.room_one_to_one`
* `.m.rule.poll_start`
* `.m.rule.poll_end`
* `.m.rule.message`
* ...

Justification:

* `.m.rule.poll_start_one_to_one` must come before `.m.rule.room_one_to_one`.
    * As it is more specific than `.m.rule.room_one_to_one`, whereas both rules would match an `m.poll.start` event in a 1-1 room.
* `.m.rule.poll_end_one_to_one` must also come before `.m.rule.room_one_to_one`, for the same reason as above.
* `.m.rule.poll_start` must come before `.m.rule.message`.
    * As it is more specific than `.m.rule.message`, whereas both rules would match an `m.poll.start` event in any room.
* `.m.rule.poll_end` must also come before `.m.rule.message`, for the same reason as above.

If this looks good to you at a glance @turt2live, I'm happy to add it into the MSC for you?

turt2live: > If this looks good to you at a glance @turt2live, I'm happy to add it into the MSC for you?

Looks like ordering to me - have at it :)

anoadragon453: Correction on my comment above - the push rules defined in this MSC *do not need* to come before `.m.rule.room_one_to_one` or `.m.rule.message`, as those push rules ([link to spec](https://spec.matrix.org/v1.5/client-server-api/#default-underride-rules)) pattern match on events with type `m.room.message`. The push rules in this MSC pattern match on events with types `m.poll.start` and `m.poll.end`. So after all that, they won't conflict, and can pretty much go anywhere in the list of underride rules we have today :)

I noticed that after realising that the [Synapse implementation of this MSC](https://github.com/matrix-org/synapse/pull/14787) just plonks them at the end.

We may consider reordering them to make things more coherent, but purely functionally there's no reason to.

anoadragon453:
Synapse has gained experimental support for this MSC: https://github.com/matrix-org/synapse/pull/14787

I have updated this MSC's description to reflect this :slightly_smiling_face: cc @turt2live

proposals/3930-polls-notifications.md
mnaturel: I don't know if it is still planned but I think there is still a need to have ordering for the new introduced push rules for polls as they match the same event type. It means we should ensure to process the new rules for poll start event in the following order:

- `.m.rule.poll_start_one_to_one`
- `.m.rule.poll_start`

And same for poll end events:

- `.m.rule.poll_end_one_to_one`
- `.m.rule.poll_end`

Ideally the rules should be disjoint to avoid thinking about the ordering but unfortunately if we want to do so, we would need to introduce a new operator `!=` for the condition `room_member_count`. It seems it does not exist yet according to the Matrix API doc.

turt2live:
Once https://github.com/matrix-org/matrix-spec-proposals/pull/3381 has entered/completed FCP, this MSC can also enter FCP.

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,133 @@
...
+order presented by this MSC.
+
+```json
+{
+  "rule_id": ".m.rule.poll_start_one_to_one",
+  "default": true,
+  "enabled": true,
+  "conditions": [
+    {"kind": "room_member_count", "is": "2"},
+    {"kind": "event_match", "key": "type", "pattern": "m.poll.start"}
```
clokep, line 28: It might be clearer if this now uses [`event_property_is`](https://spec.matrix.org/v1.7/client-server-api/#conditions-1)? And likely needs the [proper escaping](https://spec.matrix.org/v1.7/appendices/#dot-separated-property-paths).

(Same goes for the other push rules.)

turt2live:
The dependent MSC has finished FCP enough for this to enter.

@mscbot fcp merge

turt2live:
This MSC has completed FCP per the [above comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3930#issuecomment-1680975554).

turt2live:
Spec PR is blocked on wider extensible events/polls efforts.

proposals/3930-polls-notifications.md
```diff
@@ -0,0 +1,143 @@
...
+```jsonc
+{
+  "rule_id": ".m.rule.poll_start_one_to_one",
+  "default": true,
+  "enabled": true,
+  "conditions": [
+    {"kind": "room_member_count", "is": "2"},
+
+    // Note: `.` is escaped once, but for valid JSON we need to escape the escape.
+    {"kind": "event_property_is", "key": "type", "value": "m\\.poll\\.start"}
```
zecakeh, line 30: The `value` does not need to escape dots (because it's not a path in the object), only the `key` does, right?

clokep: Correct, the value should not need to be escaped.

6543:
> Spec PR is blocked on wider extensible events/polls efforts.

what's the current state in that ?!?

any links to open issues/pulls would be nice :)

6543:
> Spec PR is blocked on wider extensible events/polls efforts.

cc @turt2live 

what's the current state in that ?!?

any links to open issues/pulls would be nice :)

PR #3932, opened 2022-11-15
Push rule condition for extensible events room version

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/rver-cond/proposals/3932-extensible-events-room-ver-push-rule-condition.md)


proposals/3932-extensible-events-room-ver-push-rule-condition.md
```diff
@@ -0,0 +1,50 @@
+# MSC3932: Extensible events room version push rule feature flag
```
turt2live, line 1: Per https://github.com/matrix-org/matrix-spec-proposals/pull/3952, we should require that the deprecated push rules be ignored in supported room versions. This would mean that clients can avoid sending empty/useless objects on their events.

It's unclear if we'd want to support `m.mentions` as a mixin or regular content block though - either works.

PR #3933, opened 2022-11-15
Core push rules for extensible events

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/push-rules/proposals/3933-extensible-events-push-rules-core.md)

### DIFF
--- a/proposals/0000-extensible-events-push-rules-core.md
+++ b/proposals/0000-extensible-events-push-rules-core.md
@@ -1,0 +1,246 @@
+# MSC0000: Core push rules for Extensible Events
+
+[MSC1767](https://github.com/matrix-org/matrix-spec-proposals/pull/1767) covers a system for extending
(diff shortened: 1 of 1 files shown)


proposals/3933-extensible-events-push-rules-core.md
```diff
@@ -0,0 +1,246 @@
+# MSC3933: Core push rules for Extensible Events
```
richvdh, line 1: I think you might need a special rule to replace `.m.rule.suppress_notices`.

richvdh: ... or this might be a topic for MSC3955? somewhere, anyway!

turt2live: I think this is yet another unwritten MSC, or maybe MSC3955 indeed. Here is a fine place for the thread at least.

PR #3939, opened 2022-11-24
MSC3939: Account locking

babolivier:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/babolivier/m_locked/proposals/3939-account-locking.md)

[Synapse implementation](https://github.com/matrix-org/synapse/pull/15870)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3939#issuecomment-1828611491)

### DIFF
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -27,14 +27,6 @@ a more useful error message about their account status, though their session dat
 may be deleted by the client if it doesn't recognize the error code. Soft logout
 aims to prevent total destruction of this data, however.
 
-Clients MAY prevent actually logging the user out until the error code or response
-changes. This is to allow the client to emit a few more requests after receiving
-the error, as may be the case with a very active `/sync` loop. Once the error code
-changes (but remains a 401 otherwise, regardless of soft logout), the client
-should proceed with the logout. Similarly, if the response changes from an error
-to a successful response, the client can assume the account has been unlocked and
-return to normal operation without needing to get a new access token.
-
 Upon receiving the `M_USER_LOCKED` error, clients SHOULD retain session information
 including encryption state and inform the user that their account has been locked.
 Details about *why* the user's account is locked are not formally described by
@@ -42,6 +34,18 @@ this proposal, though future MSCs which cover informing users about actions take
 against their account should have such details. Clients may wish to make use of
 [server contact discovery](https://spec.matrix.org/v1.10/client-server-api/#getwell-knownmatrixsupport)
 in the meantime.
+
+Clients SHOULD hide the normal UI from the user when informing them that their
+account is locked, preventing general use of the account.
+
+Clients SHOULD continue to `/sync` and make other API calls to more quickly detect
+when the lock has been lifted. If unlocked, the APIs will either return a different
+error code or a normal 200 OK/successful response. For example, `/sync` will return
+to working as though nothing ever happened. If the error code changes to
+`M_UNKNOWN_TOKEN`, the client should delete local session data as it normally
+would when seeing the error code (and use soft logout as it normally would). This
+is typically expected if the server admin logged the user out or the user logged
+themselves out.
 
 Locked accounts are still permitted to access the following API endpoints:
 
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -52,7 +52,7 @@ Locked accounts are still permitted to access the following API endpoints:
 * [`POST /logout`](https://spec.matrix.org/v1.9/client-server-api/#post_matrixclientv3logout)
 * [`POST /logout/all`](https://spec.matrix.org/v1.9/client-server-api/#post_matrixclientv3logoutall)
 
-Servers SHOULD NOT invalidate an account's access tokens in case the account becomes
+When a user's account is locked, servers SHOULD NOT invalidate an account's access tokens in case the account becomes
 unlocked: the user should be able to retain their sessions without having to log
 back in. However, if a client requests a logout (using the above endpoints), the
 associated access tokens should be invalidated as normal.
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -39,7 +39,7 @@ Clients SHOULD hide the normal UI from the user when informing them that their
 account is locked, preventing general use of the account.
 
 Clients SHOULD continue to `/sync` and make other API calls to more quickly detect
-when the lock has been lifted. If unlocked, the APIs will either return a different
+when the lock has been lifted.  However, clients should rate-limit their requests. If unlocked, the APIs will either return a different
 error code or a normal 200 OK/successful response. For example, `/sync` will return
 to working as though nothing ever happened. If the error code changes to
 `M_UNKNOWN_TOKEN`, the client should delete local session data as it normally
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -23,7 +23,9 @@ to make use of the [soft logout](https://spec.matrix.org/v1.9/client-server-api/
 semantics: keep encryption state, but otherwise render the account unusable. 401
 is used to support legacy clients by giving the user semantically meaningful
 experience: they may need to try logging in again, and when they do they may get
-a more useful error message about their account status.
+a more useful error message about their account status, though their session data
+may be deleted by the client if it doesn't recognize the error code. Soft logout
+aims to prevent total destruction of this data, however.
 
 Clients MAY prevent actually logging the user out until the error code or response
 changes. This is to allow the client to emit a few more requests after receiving
@@ -67,9 +69,17 @@ for details.
 Another similar concept would be "shadow banning", though this only applies to
 moderation use cases.
 
-Another option is to use 403 responses instead of 401 and `soft_logout`. We choose this
-so that existing apps provide some feedback to the user without explicit support for
-this MSC.
+A 403 HTTP status code was considered instead of 401 with a `soft_logout`. A 403
+would indicate that the given action is denied, but otherwise keep the user logged
+in. This could wrongly indicate [suspension](https://github.com/matrix-org/matrix-spec-proposals/pull/3823),
+confusing the user. Instead, we provide a semantically similar experience where
+the user gets soft logged out on legacy clients, preserving encryption and related
+session data (assuming the client also supports soft logout). This can result in
+some loss of other session data however, like device-specific settings. Users may
+also be differently confused when they try to log back in and get cryptic error
+messages (indicating wrong username/password), however as mentioned above in the
+Potential Issues section, communicating actions taken against an account is a
+concern for a future MSC.
 
 ## Unstable prefix
 
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Account locking
+# MSC3939: Account locking
 
 There are legitimate cases where server administrators might want the ability to
 temporarily lock users out of their account. For example, server administrators
@@ -11,11 +11,6 @@ to prevent a user from accessing their account is to deactivate it. However,
 this is a pretty destructive operation. Ideally, a locked account could be
 unlocked without any visible impact on the account itself (joined rooms,
 associated 3PIDs, etc).
-
-Note that
-[MSC3823](https://github.com/matrix-org/matrix-spec-proposals/pull/3823)
-addresses a similar concept coming from a moderation perspective. However,
-locking and suspending an account are semantically different.
 
 ## Proposal
 
@@ -52,4 +47,4 @@ two distincts error codes.
 ## Unstable prefix
 
 Until this proposal is accepted, implementations must use
-`ORG_MATRIX_MSCXXXX_USER_LOCKED` instead of `M_USER_LOCKED`.
+`ORG_MATRIX_MSC3939_USER_LOCKED` instead of `M_USER_LOCKED`.
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -53,6 +53,13 @@ unlocked: the user should be able to retain their sessions without having to log
 back in. However, if a client requests a logout (using the above endpoints), the
 associated access tokens should be invalidated as normal.
 
+## Potential issues
+
+This proposal does not communicate *why* a user's account is restricted. The human-readable `error`
+field may contain some information, though anything comprehensive may not be surfaced to the user.
+A future MSC is expected to build a system for both informing the user of the action taken against
+their account and allow the user to appeal that action.
+
 ## Alternatives
 
 [MSC3823](https://github.com/matrix-org/matrix-spec-proposals/pull/3823) covers
--- a/proposals/3939-account-locking.md
+++ b/proposals/3939-account-locking.md
@@ -1,54 +1,66 @@
 # MSC3939: Account locking
 
-There are legitimate cases where server administrators might want the ability to
-temporarily lock users out of their account. For example, server administrators
-might require users to go through a regular, out-of-bounds verification check in
-order to keep using their account, and temporarily lock a user out of their
-account if they do not complete this check in time.
+Account locking is a common safety and security tool where server administrators
+can prevent a user from usefully using their account. For example, too many failed
+login attempts, escalation in moderation action, or out of band verification
+needing to be completed.
 
-At the time of writing this proposal, the only option for a server administrator
-to prevent a user from accessing their account is to deactivate it. However,
-this is a pretty destructive operation. Ideally, a locked account could be
-unlocked without any visible impact on the account itself (joined rooms,
-associated 3PIDs, etc).
+Currently, Matrix only supports deactivating an account. This is a destructive
+action which often leads to the account leaving all joined rooms, among other
+details. With account locking, the effect of making the user unable to access
+their account is achieved without destroying that same account - the account
+can always be unlocked.
+
+This proposal covers account locking, though leaves the specifics of use as an
+implementation detail. Self-serve locking APIs are additionally not provided.
 
 ## Proposal
 
-A new `M_USER_LOCKED` is introduced, which is communicated to clients in
-`401 Unauthorized` HTTP responses with `soft_logout` set to `true`.
+When an account is locked, clients receive a `401 Unauthorized` error response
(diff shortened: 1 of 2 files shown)


aceArt-GmbH:
Seems to overlap a bit with https://github.com/matrix-org/matrix-spec-proposals/pull/3823 ;)

babolivier:
> Seems to overlap a bit with #3823 ;)

See the [Alternatives](https://github.com/matrix-org/matrix-spec-proposals/blob/babolivier/m_locked/proposals/3939-account-locking.md#alternatives) section :)

Also please write comments on MSCs in threads to prevent clottering the timeline :sweat_smile:

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,50 @@
...
+* homeservers must not automatically invalidate existing access tokens for the
+  user.
+* clients should keep data that has already been locally persisted, unless the
+  user manually logs out.
+
+When an account is unlocked, clients and the homeserver can start interacting
+again as if nothing happened, similarly to when a client recovers after loss of
+connection.
+
+## Alternatives
```
MadLittleMods, line 32: For moderation, another alternative to account deactivation is [shadow bans](https://matrix-org.github.io/synapse/v1.39/admin_api/user_admin_api.html#shadow-banning-users) to essentially lock the person out from sending any events although the user gets no feedback on purpose in this case.

Not a good candidate for other non-moderation cases through.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,50 @@
...
+
+When an account is locked:
+
+* homeservers must respond to any authenticated request from the user with
+  `M_USER_LOCKED`, _except_ for requests to `/logout` and `/logout/all`.
+* homeservers must not automatically invalidate existing access tokens for the
+  user.
+* clients should keep data that has already been locally persisted, unless the
+  user manually logs out.
+
```
MatMaul, line 27: We would like to hide locked users from the user directory, should it be added to this MSC or left as an implementation detail ?

If left as an implementation detail I am thinking of providing an option in synapse to tweak the behavior.

babolivier:
PSA: I don't really wish to spend time maintaining this MSC anymore (as it was written in the context of work related to a role I no longer hold); plus I don't have write access to this branch anymore. This MSC should be considered unmaintained/abandoned unless someone else wishes to take it over (in which case, be my guest).

proposals/3939-account-locking.md
babolivier: Yeah, if I remember correctly the initial reason that triggered this work was trying to make Synapse's [account validity](https://matrix-org.github.io/synapse/latest/modules/account_validity_callbacks.html) feature not use an [out-of-spec error code](https://github.com/matrix-org/synapse/blob/07d7cbfe69c239a7ffe5668c1166799370eef0d6/synapse/api/errors.py#L73).

A use case could be that an organisation might require its members to verify their activity within the org by clicking a link sent via email every few months. In which case it is desirable to visibly lock the user out of their account if they don't click the link before a set grace period ends, but it isn't to deactivate their account because they might e.g. be on holiday and should be able to recover it by clicking on said link when they come back.

I won't update the MSC with this clarification (since I'm abandoning it, see https://github.com/matrix-org/matrix-spec-proposals/pull/3939#issuecomment-1618816306), but if you think that would be helpful to add it in (which I would tend to agree with) feel free!

proposals/3939-account-locking.md
babolivier: My personal opinion is that this sounds like an implementation detail to me.

MatMaul:
Thanks for your work @babolivier. I am going to take this over.

MatMaul:
An implementation has been [merged](https://github.com/matrix-org/synapse/pull/15870) in synapse and will be available in synapse 1.91.

proposals/3939-account-locking.md
turt2live: This indeed sounds like an implementation detail, though per the [spec](https://spec.matrix.org/v1.8/client-server-api/#post_matrixclientv3user_directorysearch) the server would need to reveal locked users who share a room with the user doing the search - an MSC is required to change this.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
...
+At the time of writing this proposal, the only option for a server administrator
+to prevent a user from accessing their account is to deactivate it. However,
+this is a pretty destructive operation. Ideally, a locked account could be
+unlocked without any visible impact on the account itself (joined rooms,
+associated 3PIDs, etc).
+
+## Proposal
+
+A new `M_USER_LOCKED` is introduced, which is communicated to clients in
+`401 Unauthorized` HTTP responses with `soft_logout` set to `true`.
```
turt2live, line 18:
```suggestion
`401 Unauthorized` HTTP responses with [`soft_logout` set to `true`](https://spec.matrix.org/v1.8/client-server-api/#soft-logout).
```

turt2live:
Aside from some editorial comments, I think this is largely ready to go. 

SCT members should feel welcome to comment on the user directory thread.

@mscbot fcp merge

proposals/3939-account-locking.md
dbkr, line 18: Sorry, I don't think I really understand this paragraph. Firstly, "A new `M_USER_LOCKED` error code" presumably? And are you saying that this code will be used as the error code in any such response? Also, wouldn't it be 403 responses? A 401 would be asking the user to reauthenticate, which isn't what we want.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
...
+
+When an account is locked:
+
+* homeservers must respond to any authenticated request from the user with
+  `M_USER_LOCKED`, _except_ for requests to `/logout` and `/logout/all`.
+* homeservers must not automatically invalidate existing access tokens for the
+  user.
+* clients should keep data that has already been locally persisted, unless the
```
dbkr, line 26: I would phrase this in terms of what a client does when it gets that error code which might be clearer, as only the server implicitly knows that the account is locked.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
...
+  user manually logs out.
+
+When an account is unlocked, clients and the homeserver can start interacting
+again as if nothing happened, similarly to when a client recovers after loss of
+connection.
+
+Using 401 return code allows the user to be aware of the fact that something wrong is
+happening, since the app will be logged out.
```
dbkr, line 34: I'm left unclear as to whether the user is expected to have to log back in again after their account is unlocked. This would say yes, but the above paragraph implies that don't? It would be good to clarify, probably in the sectin about the client's behaviour (I don't think they'd need to log back in, necessarily?)

proposals/3939-account-locking.md
dbkr: Oh, I see the explanation in the alternatives section. A bit here about why 401 is used would probably be beneficial.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
...
+
+Another option for merging this proposal with MSC3823 would be adding a
+`suspended` boolean property and an optional `href` string property (which would
+be required if `suspended` is `true`) to errors bearing the `M_USER_LOCKED`
+error code. Opinions are welcome on whether this is a better solution than using
+two distincts error codes.
+
+Another option is to use 403 responses instead of 401 and `soft_logout`. We choose this
+so that existing apps provide some feedback to the user without explicit support for
+this MSC.
```
dbkr, line 55: It does mean that, assuming we believe 403 is the 'right' option, we'd have to make a further change to adjust it to this though? Also, wouldn't this be a bad UX because the user would try to log back in and not be able to? (Hopefully the client would show them the 'account locked' message once they tried to log in, but this is still rather too late). Honestly this feels to me like taking another shortcut that might get a solution faster but will add another paper cut and lead to a poorer experience.

dbkr:
@mscbot concern 401/403 behaviour

proposals/3939-account-locking.md
dbkr: Impl detail sounds fine. You could argue it would be non-spec compliant but by the same token you could argue that not returning deactivated users is too.

dbkr:
@msbot concern general clarity

https://github.com/matrix-org/matrix-spec-proposals/pull/3939#discussion_r1407543518 mostly

proposals/3939-account-locking.md
dklimpel: Are implementation details explicitly mentioned here in the spec so that it is clear to the reader and the topic is not regularly discussed because it is not clear?

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
+# MSC3939: Account locking
+
+There are legitimate cases where server administrators might want the ability to
+temporarily lock users out of their account. For example, server administrators
+might require users to go through a regular, out-of-bounds verification check in
+order to keep using their account, and temporarily lock a user out of their
+account if they do not complete this check in time.
+
+At the time of writing this proposal, the only option for a server administrator
+to prevent a user from accessing their account is to deactivate it. However,
```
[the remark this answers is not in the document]
babolivier: I can be wrong, but I believe this only works for password auth?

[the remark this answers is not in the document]
dklimpel: Yes, of course.

proposals/3939-account-locking.md
turt2live: There should be some text in the MSC to note that something has been declared an implementation detail, yes.

uhoreg: If it is to be left as an implementation detail, then the MSC can say something like "Servers may omit locked accounts from the user directory."

dbkr:
@mscbot concern general clarity

(duplicate because I spelt MSC wrong the first time and the bot doesn't understand edits)

turt2live:
This MSC is on a shortlist for considerations in Matrix 1.11's release cycle, and so I'm taking it on under my Element T&S hat in service of the Foundation.

proposals/3939-account-locking.md
turt2live: I believe I have clarified this

proposals/3939-account-locking.md
turt2live: I believe I have clarified this

proposals/3939-account-locking.md
turt2live: (added to text)

proposals/3939-account-locking.md
turt2live: moved up

turt2live: applied manually

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,60 @@
...
+this is a pretty destructive operation. Ideally, a locked account could be
```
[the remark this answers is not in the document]
turt2live: introduction clarified, I think

proposals/3939-account-locking.md
turt2live: brief mention has been added

proposals/3939-account-locking.md
turt2live: (I've avoided changing this line to ensure this thread stays visible).

Not sure the new text is any more convincing, but I'm curious if thoughts have changed here (or could be changed after reading the new proposal text).

turt2live:
@dbkr I believe I've addressed your concerns, but please feel free to raise new threads as needed.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,79 @@
...
+experience: they may need to try logging in again, and when they do they may get
+a more useful error message about their account status.
+
+Clients MAY prevent actually logging the user out until the error code or response
+changes. This is to allow the client to emit a few more requests after receiving
+the error, as may be the case with a very active `/sync` loop. Once the error code
+changes (but remains a 401 otherwise, regardless of soft logout), the client
+should proceed with the logout. Similarly, if the response changes from an error
+to a successful response, the client can assume the account has been unlocked and
+return to normal operation without needing to get a new access token.
```
dbkr, line 34: Sorry, I don't understand what this is trying to say. What/who is trying to log out that the client may be preventing them? 'Proceed with the logout' means call the logout endpoint? Delete the user's data from the device? How does the state change on the server when the logout API is called?

proposals/3939-account-locking.md
dbkr: As far as I can see, the deal is:
403:
 * Legacy apps just stop working on account lock, appearing more or less as if the server is offline
 * More correct(?)
401:
 * Legacy app deletes the user's session data on account lock so they potentially lose encrypted messages
 * User tries to log back in and them hopefully works out what's happened
 * Suggests that the credential (ie. access token) is incorrect when it isn't.

Does that feel accurate? Seems like mentioning that the user's session data will get nuked by legacy apps on a 401 would at least be useful to mention?

turt2live: Soft logout should prevent total destruction of session data, but some information would indeed be lost. 403 I don't think feels correct as it'd be more indicative of suspension (https://github.com/matrix-org/matrix-spec-proposals/pull/3823) than locking, at least as far as legacy apps are concerned.

I'll try to clarify this in the MSC (please feel free to open a new thread, as the line here will probably change)

proposals/3939-account-locking.md
turt2live: this is trying to say that the client doesn't need to delete anything, like a softer soft logout. Will try to clarify.

turt2live: hmm, this ended up being a duplicate paragraph in practice. I've distributed the useful bits to other places in the text - please open new threads if the wording is still unclear.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,91 @@
...
+needing to be completed.
+
+Currently, Matrix only supports deactivating an account. This is a destructive
+action which often leads to the account leaving all joined rooms, among other
+details. With account locking, the effect of making the user unable to access
+their account is achieved without destroying that same account - the account
+can always be unlocked.
+
+This proposal covers account locking, though leaves the specifics of use as an
+implementation detail. Self-serve locking APIs are additionally not provided.
```
richvdh, line 15: This took me several goes to grok

```suggestion
implementation detail. APIs for administration of locking are also not provided.
```

dbkr:
@mscbot resolve general clarity

dbkr:
@mscbot resolve 401/403 behaviour

turt2live:
I've manually resolved the concerns for now.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,91 @@
...
+would when seeing the error code (and use soft logout as it normally would). This
+is typically expected if the server admin logged the user out or the user logged
+themselves out.
+
+Locked accounts are still permitted to access the following API endpoints:
+
+* [`POST /logout`](https://spec.matrix.org/v1.9/client-server-api/#post_matrixclientv3logout)
+* [`POST /logout/all`](https://spec.matrix.org/v1.9/client-server-api/#post_matrixclientv3logoutall)
+
+Servers SHOULD NOT invalidate an account's access tokens in case the account becomes
```
uhoreg, line 55:
```suggestion
When a user's account is locked, servers SHOULD NOT invalidate an account's access tokens in case the account becomes
```

Having this sentence right after the bit about the logout endpoints makes it initially sound like servers shouldn't invalidate the access tokens when the logout endpoints are called, which is clearly not what is intended.

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,91 @@
...
+Details about *why* the user's account is locked are not formally described by
+this proposal, though future MSCs which cover informing users about actions taken
+against their account should have such details. Clients may wish to make use of
+[server contact discovery](https://spec.matrix.org/v1.10/client-server-api/#getwell-knownmatrixsupport)
+in the meantime.
+
+Clients SHOULD hide the normal UI from the user when informing them that their
+account is locked, preventing general use of the account.
+
+Clients SHOULD continue to `/sync` and make other API calls to more quickly detect
```
uhoreg, line 41: How should the server behave when `/sync` is called?  If it returns immediately with the `401 Unauthorized`, and then the client immediately retries, it sounds like this would cause unnecessary traffic.  So presumably the server should wait until the timeout is reached before sending the `401`, unless the account gets unlocked before?  Though I guess it should send the `401` immediately the first time that `/sync` gets called after the account gets locked, so that the user is notified immediately?

tulir: I also found this bit weird since it says `/sync` *and* make other API calls. I assumed it means the client should make *any* API calls periodically, which could be `/sync` or `/account/whoami` or whatever.

If /sync is getting a new kind of 401 long polling, it definitely needs to be documented explicitly. If not, the section should probably be clarified that clients should just periodically call any API (but not in a tight loop like /sync)

turt2live: I believe the missing language is that clients should back off for some time, as if they were rate limited. I don't think `/sync` neccessarily gets a new long poll mechanism here, but clients do need to do something to discover when an account is no longer locked. Element Web for example will back off on `/sync` errors, then loop a `/versions`, `/push_rules/`, `/sync` call order - if all of these endpoints return success codes, the client resumes normal operation.

The bit about "other API calls" is mostly to acknowledge that clients do not linearly send requests. An account may be locked in the middle of a user scrolling through backlog: there's `/messages`, `/send/m.receipt`, and `/sync` all happening at that point - the client should recognize that an error on any of them is assumed to be an error on all of them.

(if this is clearer, happy to translate it to the MSC)

proposals/3939-account-locking.md
```diff
@@ -0,0 +1,91 @@
...
+when the lock has been lifted. If unlocked, the APIs will either return a different
```
uhoreg, line 42:
```suggestion
when the lock has been lifted.  However, clients should rate-limit their requests. If unlocked, the APIs will either return a different
```

to address the thread above.

proposals/3939-account-locking.md
turt2live: addressed by https://github.com/matrix-org/matrix-spec-proposals/pull/3939#discussion_r1576886975

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1934

turt2live:
Merged 🎉

PR #3480, opened 2021-11-09
MSCxxxx: Make device names private

uhoreg:

PR #3945, opened 2022-12-06
MSC3945: Private device names

uhoreg:


Zocker1999NET:
I like the idea of hiding the device names in the future as I agree with they are no longer required to be public thanks to cross-signing.

However if others think public device names might still be useful to them, an easier idea could be that users can specify with a simple option if all their device names shall be public (which will be disabled by default). So techy aware users still can enable the option if required while knowing the risks. This would introduce less complexity than introducing separate public & private device names, especially because this does not require a change in the federation API, only in the client-server API.

proposals/3945-private-device-names.md
```diff
@@ -0,0 +1,73 @@
+# MSC3945: Private device names
```
anoadragon453, line 1: @Zocker1999NET please use threads to comment on suggestions for MSCs. Copying your comment here:

> I like the idea of hiding the device names in the future as I agree with they are no longer required to be public thanks to cross-signing.
> 
> However if others think public device names might still be useful to them, an easier idea could be that users can specify with a simple option if all their device names shall be public (which will be disabled by default). So techy aware users still can enable the option if required while knowing the risks. This would introduce less complexity than introducing separate public & private device names, especially because this does not require a change in the federation API, only in the client-server API.

anoadragon453: I'd like to see a valid use case for public device names before seeing an option added, just in case it ends up being unnecessary.

uhoreg:
Apparently, I already wrote an MSC for the same thing: https://github.com/matrix-org/matrix-spec-proposals/pull/3480

PR #3779, opened 2022-04-20
MSC0000: "Owned" State Events

andybalaam:

PR #1791, opened 2019-01-10
Add room version upgrades

turt2live:
Rendered: see 'docs' status check
Proposal: https://github.com/matrix-org/matrix-doc/issues/1501

PR #1502, opened 2018-08-10
MSC 1501: Room version upgrades

richvdh:
Proposal for #1501

Rendered: https://github.com/matrix-org/matrix-doc/blob/rav/room_upgrades/proposals/1501-room-version-upgrades.md

Discussion room: [#msc-1501:sw1v.org](https://matrix.to/#/#msc-1501:sw1v.org)

Issue #1425, opened 2018-07-17
Room Versioning


richvdh:
Documentation: https://docs.google.com/document/d/1urKgReoHqxX8R_XtySB17dPi-DZcKhqTEL2_s895Wz0
PRs: #1516

supercedes #1196.

richvdh:
Last call for comments on this before we go ahead and start implementation next week.

erikjohnston:
Are we using integers or strings for room versions? Benefits of strings is that it would make it a) clearer that you can't use ranges, and b) makes it easier to have versions like `test-version-foo`

OTOH, I'm not sure we want to be encouraging weird room versions in the wild

turt2live:
The error schema implies integers, which I think is a good thing. It's much easier to determine if a room is outdated when it's an integer.

KitsuneRal:
There's not a word about exposing the room version to clients; exposing the join error is only mentioned. Given that the room version may (potentially greatly, in case of event id formats etc.) affect the clients, should this proposal mention where to find the room version in CS API? Or are we deliberately keeping it a server-server thing for now, given that there no _actually immediate_ changes that would spill over to clients?

richvdh:
@erikjohnston / @turt2live : let's define them as integers.

richvdh:
@non-Jedi:

> The wording in this document makes it sound like the actual spec PR may have different "mechanics" than outlined in this document. Specifically, the phrase, "might look like"

Sorry, that phrasing was really just an artefact of the text being copied from an earlier proposal where things were less certain. Now updated.

> I just want to be clear that if the actual way that room versions work ends up being different than outlined in the mechanics section of this document, a new spec proposal will be made before a spec PR is merged, right?

It's likely that, as I implement this, I'll discover things that don't work out as planned. If that happens, I'll update the doc and highlight the changes here.

> Room versions are supposed to mark changes in mechanics for both clients and servers, right? But the way the proposal is worded, it seems to say that servers get to set the default version of rooms. 

Correct.

> Room versions are supposed to mark changes in mechanics for both clients and servers, right?

Yes, it's likely that some changes that might be supported by a room version bump will require client-side changes to support (an example would be: changing the way bans work so that bans are stored in a separate m.room.ban event rather than an m.room.member event. In order to figure out who has been banned from a room, a client would have to look at a different sort of event).

> Does this mean that outdated clients will no longer be able to create rooms that they can use on servers that set a default room version other than 1?
>
[...]

PR #3946, opened 2022-12-09
MSC0000: Dynamic room predecessor

MadLittleMods:
A proposal for specifying a room's predecessor after the fact.

Rendered (todo)

### DIFF
--- a/proposals/0000-dynamic-predecessor.md
+++ b/proposals/0000-dynamic-predecessor.md
@@ -1,0 +1,85 @@
+# MSC0000: Dynamic room predecessor
+
+Currently, the only way to specify the predecessor of a room is during room creation in
+the `m.room.create` state event which can't be changed after the room is created
+(the `m.room.create` event is immutable and can't be changed retrospectively).
+
+This is in contrast to specifying the successor of a room via the `m.room.tombstone`
+state event in the `replacement_room` content field which can be sent at any time to
+update the state.
+
+Traditionally, the predecessor use case has only been necessary for room upgrades where
+one room is replaced by another and since the old room exists at the time of the new
+room being created, it can easily be specified in the `m.room.create` event of the new
+room.
+
+But in history import cases like
+[MSC2716](https://github.com/matrix-org/matrix-spec-proposals/pull/2716), specifying the
+predecessor after the fact becomes important because the new historical room wasn't
+available to reference when the current room was originally created. It's important to
+note that history import is more than just
+[MSC2716](https://github.com/matrix-org/matrix-spec-proposals/pull/2716) since people
+can use [MSC3316 timestamp
+massaging](https://github.com/matrix-org/matrix-spec-proposals/pull/3316) `?ts=123`
+query parameter to send messages back in time, in general servers can craft events at
+any time which are accepted over federation.
+
+Replacing the predecessor of a room is also nice if you make a mistake in the import
+process and only notice after the fact. You can easily create a new historical room with
+the mistake corrected and splice it into the predecessor/successor chain as necessary.
+
+
+## Proposal
+
+Add a `m.room.predecessor` state event type where you can specify the room predecessor.
+This allows the predecessor state to be updated as necessary.
+
+Event type: State event
+State key: A zero-length string.
+
+**`m.room.predecessor` event `content` field definitions:**
+
+key | type | value | description | required
+--- | --- | --- | --- | ---
+`predecessor_room_id` | string | Room ID | A reference to the room that came before and was replaced by this room | yes
+
+The presense of `m.room.predecessor` state in the room should take priority over the
+`predecessor` specified in the `m.room.create` event.
+
+
+## Potential issues
+
+*None surmised so far*
+
+
+## Alternatives
+
+One possible alternative is to allow sending the `m.room.create` multiple times but this
+would require extensive [authorization
+rule](https://spec.matrix.org/v1.5/rooms/v10/#authorization-rules) changes and
+considerations.
+
+It would also be good to consider whether `predecessor` should be part of the
+`m.room.create` at all. Maybe we should consider removing it in favor of this new event
+type.
+
+
+## Security considerations
+
+`m.room.predecessor` is a state event that falls under the same power-level state
+restrictions/permissions as other state events so there shouldn't be any additional
+considerations here.
+
+Room upgrades are a known hairy area of Matrix and how clients deal with them (or lack
+of) but this doesn't change any of the existing semantics of predecessors. This is just
+a way to change the predecessor of where people should point to instead.
+
+
+## Unstable prefix
+
+While this feature is in development, the `m.room.predecessor` state event can be used
+as `org.matrix.msc0000.room_predecessor`.
+
+Clients can choose to honor the unstable event type at their discretion.
+
+


proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,85 @@
...
+This allows the predecessor state to be updated as necessary.
+
+Event type: State event
+State key: A zero-length string.
+
+**`m.room.predecessor` event `content` field definitions:**
+
+key | type | value | description | required
+--- | --- | --- | --- | ---
+`predecessor_room_id` | string | Room ID | A reference to the room that came before and was replaced by this room | yes
```
MadLittleMods, line 44: `predecessor_room_id` is a bit wordy but plainly describes what it is and what the value should be 🤷

proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,85 @@
...
+would require extensive [authorization
+rule](https://spec.matrix.org/v1.5/rooms/v10/#authorization-rules) changes and
+considerations.
+
+It would also be good to consider whether `predecessor` should be part of the
+`m.room.create` at all. Maybe we should consider removing it in favor of this new event
+type.
+
+
+## Security considerations
```
dkasak, line 67: There's another thing to consider here. If we allow a subsequent state event to change the predecessor, then a state reset can potentially reset it to an earlier value, or even make different homeservers observe different values. Moreover, this can happen even without a state reset in case of conditions such as a network split.

Obviously we should eradicate state resets, making this much rarer, but it's somewhat intrinsic to the MSC that situations like this can happen, as soon as we allow dynamically changing the predecessor.

Perhaps this isn't a problem, but I think it's worth mentioning.

turt2live, line 67: I'm a bit worried this MSC introduces higher risk of security issues in clients regarding room takeovers. Currently, the relationship between the previous and current room can be verified (baring state resets), giving the client confidence that room B is in fact the most recent edition of room A. With the relationship being mutable, clients are likely to not check the predecessor correctly and therefore allow someone to hide/take over non-tombstoned rooms.

The current relationship of rooms is very linear, and though this MSC says the mutable event takes precedence, not all clients will have updated to consider that point. This can lead to inconsistencies in what users see, or abusive scenarios where rooms are hidden from some people and not others.

dkasak: (I think we race conditioned a bit and share [more-or-less the same concern](https://github.com/matrix-org/matrix-spec-proposals/pull/3946#discussion_r1062703513), just worded a bit differently.)

turt2live: I'm slightly more concerned about clients getting the implementation wrong, despite warnings in the spec, though yes: the root issue is "mutable is scary", I think.

MadLittleMods: Resolving in favor of the other thread, https://github.com/matrix-org/matrix-spec-proposals/pull/3946#discussion_r1062714169

MadLittleMods: > Currently, the relationship between the previous and current room can be verified (baring state resets), giving the client confidence that room B is in fact the most recent edition of room A.

There is nothing "verifiable" about the `m.room.create` `predecessor` value besides that it came from someone with proper power levels. A Matrix room can created with whatever predecessor desired (even multiple rooms with the same predecessor). And same for the tombstone from some old room pointing to a new room. And as a note in case onlookers aren't as familiar, tombstones are already mutable.

To me, the chain of tombstone and predecessor are mere suggestions of where you can continue looking at history in the room. I'm not quite following the line of thought to get the severity with danger in any of this.

All of these problems like selectively sharing the state only to certain servers or state resets seem completely separate to this MSC. If we can't trust the state in a room, then everything falls apart.

turt2live: The presence of a tombstone pointing to the new room and a predecessor pointing to the tombstoned room forms a verifiable relationship - this was specifically designed in for room upgrades. 

With this MSC, we'd be breaking that relationship we previously went out of our way to make verifiable - this is where the concern is coming from. If we're breaking the verifiable nature of the relationship, it needs to be for justifiable/secure reasons. 

We already acknowledge that state resets and similar can cause the relationship to be broken, however the expectation in those cases is that clients actually split the room into two logical rooms again, even if half the relationship still exists. With both halves of the relationship, the client can merge the rooms together safely.

turt2live: Reference:
image

In this diagram, if a user is joined to all 3 rooms then what should happen is Room B and Room A are treated as a single room (Room B, specifically), and Room C is treated as its own room. With this MSC, the risk becomes that Room C "takes over" the relationship, thus making all 3 rooms represented by Room C, despite Room B never intending for that to happen.

State resets can also do this, but are a known problem and indeed irrelevant as far as this discussion is concerned.

MadLittleMods: Are there notes somewhere in the spec or wherever about what you're describing as "verifiable". From usage, it doesn't appear that way to me at all.

I can already create Room C with the predecessor of Room B without this MSC. Nothing seems "verifiable" about that.

`POST /_matrix/client/v3/createRoom`

```json
{
    "creation_content": {
        "predecessor": "!RFPsWIrSoCdxfOivXv:my.matrix.host"
    },
    "preset": "public_chat",
    "name": "predecessor_test1"
}
```

And this MSC doesn't change the linear linked-list relationship with tombstones and predecessors that already exists.

turt2live: It's possible to create such a room, just as it's possible to create a tombstone event that goes nowhere (or a room avatar with invalid media reference, or a message without `body`, etc). The current spec is incredibly vague because it was written in an era where that happened more than we would have liked: https://spec.matrix.org/v1.5/client-server-api/#client-behaviour-27

There might be some history in https://github.com/matrix-org/matrix-spec-proposals/pull/1791 (the spec PR) as it looks like there was some back and forth on the PR (because that was also legal in the spec process at the time), and the original MSC was https://github.com/matrix-org/matrix-spec-proposals/pull/1502 (note that it's technically MSC1501, for reasons of process at the time). There's also more discussion at https://github.com/matrix-org/matrix-spec-proposals/issues/1425 which might be useful.

I haven't looked in depth, but it appears the sentence you're after isn't described explicitly, but rather implicitly. Because the client is expected to make an effort to hide the old room from the user, it is also expected to ensure malicious predecessor structures the user has visibility on do not override rooms (as that would be a takeover). 

The js-sdk's implementation of this verification is togglable, as it somewhat depends on the scenario where it's being used: https://github.com/matrix-org/matrix-js-sdk/blob/424212cd6546a00b606380f10ae2585d1d43e518/src/client.ts#L5015

Similar filtering is done by the js-sdk when determining the true set of "visible" rooms (used by the room list algorithms in the react-sdk layer): https://github.com/matrix-org/matrix-js-sdk/blob/424212cd6546a00b606380f10ae2585d1d43e518/src/client.ts#L3768

In short, the verification happens client-side at the moment (which is regrettable, but there's much more inferred history to room upgrades that made this reality - I can dig into it internally if interested, though it doesn't affect this MSC).

----

fwiw, some of this concern is driven by the MSC's problem statement: it's not entirely clear what exact problem is being solved here, and some out of band poking has revealed a few more details about the real-world case for this. In practice, it looks like Gitter will be migrating chat history into Matrix and would like to do that in a clean way. The proposed solution appears to be that a MSC2716-compatible room will be created, history imported, tombstoned, and retroactive predecessor set on the already-existing room to point to the MSC2716-compatible room.

Assuming that's accurate (please correct me if not), there are a few concerns/questions/suggestions regarding the approach:
1. In order to guarantee any amount of durability on the imported history, Gitter would need to be using a stable room version to import the history into. Currently, MSC2716 is unstable and Synapse drops unstable room versions when the feature is included in a stable version, intentionally and regardless of usage. This will mean that eventually, when/if MSC2716 gets adopted into a stable room version, the history will be lost to time. It also means the history would only be accessible by Synapse currently, as no other server to my knowledge implements the unstable room version.
2. MSC2716 has a backwards compatibility case to allow importing history into existing rooms, as the room creator. From what I remember, all of the Gitter rooms were created by a service account: this service account could be used to import history into the already-existing room version, therefore not requiring a second room to be involved.
3. If the Gitter service account doesn't have creator access in enough of the rooms, but the Gitter server can upgrade the room, an option would be to upgrade the room in-place to a MSC2716-supported version (ideally stable, for reasons described in no. 1 above), import the history into the *new* room, and continue using the room normally (or upgrade it again for a clean slate): this gives the benefit of time-accurate history (not necessarily DAG-accurate, which clients don't care about anyways) while also requiring no changes to room upgrades. 

The value of a dynamic predecessor is definitely present, I think, though the security concern of room takeovers has me worried. Having written this short thesis of a comment, I don't really have many good ideas on how to fix it right now, but believe it can be done. In the meantime, hopefully one of the suggestions/considerations for the real-world usecase are useful.

MadLittleMods: > In short, the verification happens client-side at the moment (which is regrettable, but there's much more inferred history to room upgrades that made this reality - I can dig into it internally if interested, though it doesn't affect this MSC).

I don't see a problem where this MSC conflicts with any of this. The client can still do its merging and "verification" as necessary. And the client getting it wrong is not a dire scenario. It just results in two rooms being shown.


> fwiw, some of this concern is driven by the MSC's problem statement: it's not entirely clear what exact problem is being solved here, and some out of band poking has revealed a few more details about the real-world case for this. 

The MSC describes the import use case that I will be using this for but it's not with MSC2716.

For Gitter, we're using the single `/send?ts=123` endpoint with timestamp massaging. We're creating a separate historical room, importing the message history one by one, tombstoning to point to the "live" room and want to use this MSC to have the "live" room's predecessor point back to the historical room.

turt2live: > > In short, the verification happens client-side at the moment (which is regrettable, but there's much more inferred history to room upgrades that made this reality - I can dig into it internally if interested, though it doesn't affect this MSC).
> 
> I don't see a problem where this MSC conflicts with any of this. The client can still do its merging and "verification" as necessary. And the client getting it wrong is not a dire scenario. It just results in two rooms being shown. This "verification" is even toggleable as you mentioned so it's not core to anything.

A client getting it wrong means the room has been taken over, which is very much against our threat model for Matrix (for better or worse). This MSC allows for dynamic predecessors, but clients trying to apply their verification logic to the linked list will end up treating the proposed usecase as two isolated rooms instead of a single room. 



> > fwiw, some of this concern is driven by the MSC's problem statement: it's not entirely clear what exact problem is being solved here, and some out of band poking has revealed a few more details about the real-world case for this.
> 
> The MSC describes the import use case that I will be using this for but it's [not with MSC2716](https://gitlab.com/gitterHQ/webapp/-/merge_requests/2313#why-arent-we-using-msc2716).
> 
> For Gitter, we're using the single `/send?ts=123` endpoint with timestamp massaging. We're creating a separate historical room, importing the message history one by one, tombstoning to point to the "live" room and want to use this MSC to have the "live" room's predecessor point back to the historical room.

If this could be clarified in the MSC, it would be appreciated. The current intention of the MSC is unclear and appears to tell a different story than what is actually planned.

MadLittleMods: > A client getting it wrong means the room has been taken over, which is very much against our threat model for Matrix (for better or worse).

How is this a takeover? Above all, this is just a client display thing. And second, it looks like one room needs the tombstone and other one needs a predecessor for things to look replaced. Both sides acknowledged each other for this to happen. Anything other than this seems like a bad assumption on the clients part.

> If this could be clarified in the MSC, it would be appreciated. The current intention of the MSC is unclear and appears to tell a different story than what is actually planned.

I can update it to spell it out more plainly 👍

proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,85 @@
+# MSC3946: Dynamic room predecessor
+
+Currently, the only way to specify the predecessor of a room is during room creation in
+the `m.room.create` state event which can't be changed after the room is created
+(the `m.room.create` event is immutable and can't be changed retrospectively).
```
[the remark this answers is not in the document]
MadLittleMods: I'm not sure this is wrong 🤔 (more opinions/knowledge needed)

proposals/3946-dynamic-predecessor.md
turt2live: > How is this a takeover? Above all, this is just a client display thing. And second, it looks like one room needs the tombstone and other one needs a predecessor for things to look replaced. Both sides acknowledged each other for this to happen. Anything other than this seems like a bad assumption on the clients part.

The case I'm concerned about isn't when both rooms point at each other, but rather when a client developer inevitably makes a mistake in the implementation of this MSC. There's already significant risk of clients getting the implementation wrong, and making the predecessor dynamic opens the door wider for mistakes (despite warnings in the spec, which often go unheeded). This would result in a malicious room being allowed to overwrite/hide another room it shouldn't have right to, thus takeover.

proposals/3946-dynamic-predecessor.md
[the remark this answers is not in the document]
turt2live: "retroactively" is more correct as the subject is being looked at without consideration.

proposals/3946-dynamic-predecessor.md
MadLittleMods: > This would result in a malicious room being allowed to overwrite/hide another room it shouldn't have right to, thus takeover.

This seems like it would be a client doing something very wrong. We can't help a client that doesn't look at both `tombstone` and `predecessor` to see that they match. That's a bare minimum even with the current state of `tombstone`/`predecessor` if you're making rooms look replaced. A client can show the pointer forward/backward without this verification though.

Even a naive client implementation that only looks at the create event for the `predecessor` is safe in the new world that this MSC creates.

If a client implementation updates to look at `m.room.predecessor`, then that is also pretty fool-proof regardless of which era of state they look at. The only tricky thing I could think of is if we also lived in a [MSC3779](https://github.com/matrix-org/matrix-spec-proposals/pull/3779) world where someone doesn't check for the canoncial version with `state_key: ""` but that's not a concern yet.

proposals/3946-dynamic-predecessor.md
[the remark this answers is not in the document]
MadLittleMods: We seem to use `retrospectively` more often in the spec than `retroactively` in scenarios like this (just searching all of the specs in this repo)

Judging from https://english.stackexchange.com/a/148988/103469, it seems like `retroactively` fits better since an action is being taken to change things.

proposals/3946-dynamic-predecessor.md
turt2live: I agree that the client would be doing something wrong, however the risk of a client doing something wrong is higher with an MSC like this (in its current form) that overall leaves an impression that the protocol is insecure, even if the blame is actually with the clients. 

Unfortunately I don't recall all of the details, but there was also a reason why we embedded the predecessor into the create event instead of a dedicated state event: a dedicated state event would have been cleaner and "more proper" as far as Matrix is concerned, but we landed on putting it in the create event for a reason. Digging up that rationale would likely help this conversation, as it keeps coming to mind, though I sadly do not have the bandwidth to do that research.

MadLittleMods: > I agree that the client would be doing something wrong, however the risk of a client doing something wrong is higher with an MSC like this (in its current form) that overall leaves an impression that the protocol is insecure, even if the blame is actually with the clients.

I don't think this MSC increases the risk level. We have the exact same risk with the current `tombstone`/`predecessor` state of things.

And to address that one aspect, it sounds we should consider another MSC to verify this server-side so clients can more easily rely on eating whatever we feed it.


> Unfortunately I don't recall all of the details, but there was also a reason why we embedded the predecessor into the create event instead of a dedicated state event: a dedicated state event would have been cleaner and "more proper" as far as Matrix is concerned, but we landed on putting it in the create event for a reason. Digging up that rationale would likely help this conversation, as it keeps coming to mind, though I sadly do not have the bandwidth to do that research.

This would be interesting to find. We should start a new thread if we come up with anything here.

andybalaam: One datapoint around verifiability: I found a bug in matrix-js-sdk's code for verifying links: https://github.com/matrix-org/matrix-js-sdk/pull/3089 . The fact that this has remained undiscovered for a long time may imply that people are not paying a great deal of attention to verifiability.

proposals/3946-dynamic-predecessor.md
andybalaam, line 50: @MadLittleMods did you consider including event_id like in the create event? Some code in matrix-react-sdk expects to be able to have that. I'm not sure yet what the implications would be of not providing it when looking for a room's predecessor. (I imagine we can work around it, but I haven't looked into it yet)

MadLittleMods: > did you consider including event_id like in the create event?

Didn't know about it so no consideration was given to it so far. For my own reference: https://spec.matrix.org/v1.5/client-server-api/#mroomcreate

Adding a `last_known_event_id` field seems fine 👍

I assume it's not related to the upgrade verification part at all since we should be looking at the latest state in the room. But this sort of info could be nice as one option for a continuation point to jump to in the old room.

It terms of workaround fallbacks, we can just go to where the latest tombstone in the old room is or just the latest in the old room. One more complicated option is tor use the `/timestamp_to_event` endpoint to find the latest event from when the predecessor was sent but doesn't seem necessary or the right thing to do exactly.

t3chguy: Without a via list this is likely to be unjoinable over federation as room_ids are not routable. https://github.com/vector-im/element-web/issues/2925 for context.

justjanne:
One desired use case with this is to allow users to view and access a predecessor room they haven't been in before, and which their homeserver likely doesn't know about.

@t3chguy Just raised the issue with me that to join such rooms, the room id isn't enough, we'd also need the vias or a room alias. Which means the vias should be included at least in the dynamic predecessor message. As the static room predecessor message doesn't include vias either, use of that likely should be phased out.

proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Proposal
+
+Add a `m.room.predecessor` state event type where you can specify the room predecessor.
+This allows the predecessor state to be updated as necessary.
+
+Event type: State event
+State key: A zero-length string.
+
+**`m.room.predecessor` event `content` field definitions:**
```
justjanne, line 46: One desired use case with this is to allow users to view and access a predecessor room they haven't been in before, and which their homeserver likely doesn't know about.

@t3chguy Just raised the issue with me that to join such rooms, the room id isn't enough, we'd also need the vias or a room alias. Which means the vias should be included at least in the dynamic predecessor message. As the static room predecessor message doesn't include vias either, use of that likely should be phased out.

justjanne: (As requested, I've moved the comment to a separate thread)

MadLittleMods: > Without a via list this is likely to be unjoinable over federation as room_ids are not routable. https://github.com/vector-im/element-web/issues/2925 for context.
>
> @t3chguy, https://github.com/matrix-org/matrix-spec-proposals/pull/3946/files#r1100243871

(moving this comment to this thread)

MadLittleMods: This makes a lot of sense! 👍

Added a `via_servers` field to the `m.room.predecessor` event in the proposal

proposals/3946-dynamic-predecessor.md
MadLittleMods: So I understand the use cases, are my assumptions above correct? Are there other use cases? Is this useful?

proposals/3946-dynamic-predecessor.md
t3chguy: IMO this should be `via` for consistency with already specced shapes, e.g. `m.space.child`. https://spec.matrix.org/v1.5/client-server-api/#mspacechild

proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,96 @@
...
+
+key | type | value | description | required
+--- | --- | --- | --- | ---
+`predecessor_room_id` | string | Room ID | A reference to the room that came before and was replaced by this room | yes
+`via_servers` | [string] | List of server names | The servers to attempt to join the room through. These servers should be participating in the room to be useful. This option is necessary since room ID's are not routable on their own.
```
t3chguy, line 51: IMO this should be `via` for consistency with already specced shapes, e.g. `m.space.child`. https://spec.matrix.org/v1.5/client-server-api/#mspacechild

proposals/3946-dynamic-predecessor.md
```diff
@@ -0,0 +1,96 @@
...
+
+The presence of `m.room.predecessor` state in the room should take priority over the
+`predecessor` specified in the `m.room.create` event.
+
+An additional note is that since via servers are not specified alongside the
+`predecessor` in a `m.room.create` event, the `m.room.create` usage should probably be
+phased out in favor of this new event.
```
justjanne, line 58: The reasoning for this is that, even if the via servers were specified in the room.create event, they'd quickly go out of date, as the room.create event is immutable.

Whenever via servers change, the dynamic room predecessor event can be updated with new via information, though.

As result, even if the room.create event were updated to include via servers, phasing it out would still be recommended.

MadLittleMods: Added 👍

Thanks for the extra notes!

nyabinary:
So whats the status of this?

PR #3947, opened 2022-12-11
Allow Clients to Request Searching the User Directory Constrained to Only Homeserver-Local Users

marcusmueller:
Allow Clients to Request Searching the User Directory Constrained to Only Homeserver-Local Users

---
name: WIP Proposal
about: A proposal that isn't quite ready for formal review yet.
title: 'Allow Clients to Request Searching the User Directory Constrained to Only Homeserver-Local Users'
labels: proposal
assignees: ''

---

### Pull Request Checklist

* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/master/CONTRIBUTING.md#sign-off)
* [ ] Update the title and file name of your proposal to match this PR's number (after opening).
* [ ] Pull request includes a ['Rendered' link](https://matrix.org/docs/spec/proposals#process) above.

### DIFF
--- a/proposals/0000-restricting-user-directory-search.md
+++ b/proposals/0000-restricting-user-directory-search.md
@@ -1,0 +1,84 @@
+# MSC0000: Allow Clients to Request Searching the User Directory Constrained to Only Homeserver-Local Users
+
+Searching for local users is an extremely common task for users on
+organization-owned homeservers. Currently, the client-server protocol does not
+allow to constrain the search to local users. In practice, this often leads to
+irrelevant users from federated servers being returned.
+
+This proposal aims to fix that by allowing a client to ask for local users
+only.
+
+
+## Proposal
+
+### Proposed Protocol Extension
+
+Extend the
+[`client/v3/user_directory/search`](https://matrix.org/docs/api/#post-/_matrix/client/v3/user_directory/search)
+endpoint to take an optional argument that specifies whether search should be
+constrained to local users, federated users, or both.
+
+#### Syntactic Change
+
+`client/v3/user_directory/search` request body becomes
+
+```text
+{
+  limit: integer              The maximum number of results to return. Defaults to 10.
+  search_term*: string        The term to search for
+  exclude_sources: integer    Whether to exclude no users from directory search (0),
+                              exclude non-local users (1), exclude local users (2).
+			      Defaults to 0 (search and return all users)
+}
+```
+
+The difference here is purely the addition of the new `exclude_sources`
+optional parameter, which defaults to the current behaviour, so that legacy
+servers can just ignore the parameter and let the client filter results.
+
+The option to explicitly return only non-local users is especially meant to
+avoid returning duplicate results in case a client wants to separately request
+local users, and users known through federation.
+
+The usage of an integer allows for later extensibilty.
+
+#### Semantic Change
+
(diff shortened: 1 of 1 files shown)


proposals/3947-restricting-user-directory-search.md
```diff
@@ -0,0 +1,84 @@
...
+This proposal aims to fix that by allowing a client to ask for local users
+only.
+
+
+## Proposal
+
+### Proposed Protocol Extension
+
+Extend the
+[`client/v3/user_directory/search`](https://matrix.org/docs/api/#post-/_matrix/client/v3/user_directory/search)
```
turt2live, line 17: this should link through to spec.matrix.org/v1.5 (or whatever version is current at time of writing)

marcusmueller: Thanks! Fixed.

proposals/3947-restricting-user-directory-search.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### Syntactic Change
+
+`client/v3/user_directory/search` request body becomes
+
+```text
+{
+  limit: integer              The maximum number of results to return. Defaults to 10.
+  search_term*: string        The term to search for
+  exclude_sources: integer    Whether to exclude no users from directory search (0),
```
uhoreg, line 29: In general, in Matrix, we don't use integers to represent enumerations.  We usually use a set of string values instead.  e.g. "exclude_sources" can be one of "remote", "local", or "none", defaulting to "none".

While this doesn't mean that endpoints can't use integers, there would need to be a compelling reason for an endpoint to do so, and I don't think that the reason for using an integer is particularly strong in this case.  If you want to allow for the possibility of multiple things to be excluded in the future, the parameter can be made to be an array of things to exclude.

proposals/3947-restricting-user-directory-search.md
```diff
@@ -0,0 +1,84 @@
...
+  potential to reduce traffic and search effort, if not in meaningful ways.
+
+## Outlook
+
+Clients should implement an interface to allow for this feature to be used.
+Ideally, smartness is invested in allowing something like
+`Firstname:myownhomeserver.com`, `Firstname:local` to be decomposed into a
+search term and a term that specifies that only the local users should be
+searched.
+
```
uhoreg, line 81: There should be an ["Unstable Prefix"](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/0000-proposal-template.md#unstable-prefix) section

proposals/3947-restricting-user-directory-search.md
marcusmueller: oh, I wasn't aware of that convention. "none" makes little sense, IMHO: If it's a list / set of strings, just keep it empty.

proposals/3947-restricting-user-directory-search.md
marcusmueller: Sorry, I don't understand. Not having written an implementation so far, I did not add that section.

uhoreg: You don't need to have written any implementation yet to define an unstable prefix.  But you should define an unstable prefix for use by anyone who does want to write an implementation.

PR #3949, opened 2022-12-14
MSC3949: Power Level Tags

ajbura:
[Rendered](https://github.com/ajbura/matrix-spec-proposals/blob/power-level-tag/proposals/3949-power-level-tag.md)

implementations
- [Cinny](https://app.cinny.in)

### DIFF
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -17,30 +17,23 @@ level can be assigned to users, we can also display room member by associating i
 timeline and member list. This association will enhance the experience as tag will represent
 responsibility's of members at glance.
 
-### What is Tag?
+### Tag Definition
 
-A Tag have three properties: `"name"`, `"color"` and `"icon"`.
+A Tag have three properties: `name`, `color` and `icon` as described in table:
 
-```json
-{
-  "name": "Admin",
-  "color": "#000000",
-  "icon": {
-    "key": "",
-    "info": {
-      "h": 152,
-      "mimetype": "image/webp",
-      "size": 30001,
-      "w": 152
-    }
-  }
-}
-```
+#### Tag
+| Property | Type    | Description                                                                |
+|----------|---------|----------------------------------------------------------------------------|
+| name     | string  | Name of the tag. **Required**                                              |
+| color    | string  | Color as `hex` value. For graphical emphasis and distinction between tags. |
+| icon     | TagIcon | Graphical representation of tags name.                                     |
 
-- `"name"` For what a tag represent. **Required**.
-- `"color"` Color as `hex` value. To create graphical emphasis and distinction between tags. (_Optional_)
-- `"icon"` To display graphical representation of tags name. (_Optional_)
-  - `"key"` Key can have value as mxc or emojis. `info` object should can be considered valid if `key` is mxc.
+#### TagIcon
+| Property | Type      | Description                                                             |
+|----------|-----------|-------------------------------------------------------------------------|
+| key      | string    | This can be emoji similar to key in `m.reaction` or an `mxc` for image. |
+| info     | ImageInfo | Standard `info` object for image if key is `mxc`.                       |
+
 
 ### Attaching Tag to Power Level
 
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -24,10 +24,7 @@ A Tag have three properties: `"name"`, `"color"` and `"icon"`.
 ```json
 {
   "name": "Admin",
-  "color": {
-    "on_light": "#023020",
-    "on_dark": "#90EE90"
-  },
+  "color": "#000000",
   "icon": {
     "url": "mxc:abc",
     "info": {
@@ -41,9 +38,7 @@ A Tag have three properties: `"name"`, `"color"` and `"icon"`.
 ```
 
 - `"name"` For what a tag represent. **Required**.
-- `"color"` Color of tag. To create graphical emphasis and distinction between tags. (_Optional_)
-  - `"on_light"` Tag color for lighter background.
-  - `"on_dark"` Tag color for darker background.
+- `"color"` Color as `hex` value. To create graphical emphasis and distinction between tags. (_Optional_)
 - `"icon"` To display graphical representation of tags name. (_Optional_)
 
 ### Attaching Tag to Power Level
@@ -57,10 +52,7 @@ state key as value of power level.
   "state_key": "100",
   "content": {
     "name": "Admin",
-    "color": {
-      "on_light": "#023020",
-      "on_dark": "#90EE90"
-    },
+    "color": "#000000",
     "icon": {
       "url": "mxc://admin_icon",
       "info": {
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -43,7 +43,7 @@ empty string. The content will be a map of power level integer to Tag.
 ```json
 {
   "type": "m.room.power_level_tags",
-  "state_key": "100",
+  "state_key": "",
   "content": {
     "100": {
       "name": "Admin",
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -100,7 +100,7 @@ None?
 
 ## Unstable prefix
 
-TODO
+Implementations SHOULD refer to `m.room.power_level_tags` as `in.cinny.room.power_level_tags`.
 
 ## Dependencies
 
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -1,4 +1,4 @@
-### MSC0000 Power Level Tags
+### MSC3949 Power Level Tags
 
 Currently permissions in a certain room are handle via `m.room.power_levels`. A power level is an integer value, bigger the integer bigger the power. When we assign power level to a user we gave them power represented as integer value. This integer value is abstract and does not tell what kind of power it represent. Currently spec only suggest mapping abstract integer value to idea as: 0 - 49 to User, 50 - 99 to Moderator - 100 to Admin. This suggestion limits the wide range of power level integers to only three idea's.
 
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -26,7 +26,7 @@ A Tag have three properties: `"name"`, `"color"` and `"icon"`.
   "name": "Admin",
   "color": "#000000",
   "icon": {
-    "url": "mxc:abc",
+    "key": "",
     "info": {
       "h": 152,
       "mimetype": "image/webp",
@@ -40,6 +40,7 @@ A Tag have three properties: `"name"`, `"color"` and `"icon"`.
 - `"name"` For what a tag represent. **Required**.
 - `"color"` Color as `hex` value. To create graphical emphasis and distinction between tags. (_Optional_)
 - `"icon"` To display graphical representation of tags name. (_Optional_)
+  - `"key"` Key can have value as mxc or emojis. `info` object should can be considered valid if `key` is mxc.
 
 ### Attaching Tag to Power Level
 
@@ -54,7 +55,7 @@ state key as value of power level.
     "name": "Admin",
     "color": "#000000",
     "icon": {
-      "url": "mxc://admin_icon",
+      "key": "mxc://admin_icon",
       "info": {
         "h": 152,
         "mimetype": "image/webp",
--- a/proposals/3949-power-level-tag.md
+++ b/proposals/3949-power-level-tag.md
@@ -26,7 +26,7 @@ A Tag have three properties: `name`, `color` and `icon` as described in table:
 |----------|---------|----------------------------------------------------------------------------|
 | name     | string  | Name of the tag. **Required**                                              |
 | color    | string  | Color as `hex` value. For graphical emphasis and distinction between tags. |
-| icon     | TagIcon | Graphical representation of tags name.                                     |
+| icon     | TagIcon | Graphical representation of tag.                                     |
 
 #### TagIcon
 | Property | Type      | Description                                                             |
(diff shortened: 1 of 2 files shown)


proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Attaching Tag to Power Level
+Tag can be attached to a power level by send a state event with key as `m.room.power_level_tag` and state key as value of power level.
+```json
+{
+  "type": "m.room.power_level_tag",
+  "state_key": "100",
+  "content": {
+    "name": "Admin",
+    "color": {
```
MTRNord, line 43: I think this may not work as there are clients with more than just those 2 options. So it may not always work. What is speaking against letting the displaying client decide the color?

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### What if we have power level with undefined tag?
+If a power level is not defined client can consider to display it with the lower defined power level tag. If there is no lower defined tag, client can consider displaying tag defined with `m.room.power_level_tag` state event + state key as `"state_key": ""`, if no such event find client can display it power level itself or with the way of it's own choice.
+
+### What else it fix?
+This will also fix problem defined in [MSC3915: Owner power level](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/owner-pl/proposals/3915-owner-power-level.md).
+
+## Alternatives
+
+None?
```
MTRNord, line 74: Roles instead of powerlevels would be imho a note worthy alternative even though there is no msc for it that I am aware of.

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
+### MSC3949 Power Level Tags
```
MTRNord, line 1: Does this replace `m.room.power_levels`? Or does it add another layer to it? If it adds a layer why not add it within that state Event?

MTRNord, line 1: Who is allowed to set these? I assume only PL>50?

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+```
+* `"name"` For what a tag represent. **Required**.
+* `"color"` Color of tag. To create graphical emphasis and distinction between tags. (*Optional*)
+  - `"on_light"` Tag color for lighter background.
+  - `"on_dark"` Tag color for darker background.
+* `"icon"` To display graphical representation of tags name. (*Optional*)
+
+### Attaching Tag to Power Level
```
MTRNord, line 35: Does an event without associated PL grant the permission? How would it get mapped to those without the powerlevel? The current powerlevel event allows setting specific numbers to specific actions. This doesn't seem to have something like this.

proposals/3949-power-level-tag.md
FSG-Cat: I think its more of a suggestion. Discord has Roles with Colours and on there a colour is fine on dark theme and bad on light but this fixes that problem kinda but matrix also has infinite UI variants so that problem is more complicated on matrix.

proposals/3949-power-level-tag.md
FSG-Cat: This is as close as you get to roles without going for RBAC or hacking in full roles that dont matter auth wise. If you ask me. So its an alternative that should be mentioned but still.

proposals/3949-power-level-tag.md
FSG-Cat: This is clearly a PL Labels MSC and therefore drop in backwards compatible with all room versions and all existing homeserver implementations and clients that follow the spec due to that it will be ignored safely by non implementing stakeholders.

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
...
+
+This MSC propose tags for power levels. A tag represent a configurable information that can be tagged to any power level integer. This configurable information will give an idea of what a power level is when presented to user in graphical user interface. This will enhance the experience of managing permission as now power level can be able to represent the wide range of idea's. As power level can be assigned to users, we can also display room member by associating idea of power in room timeline and member list. This association will enhance the experience as tag will represent responsibility's of members at glance.
+
+## Proposal
+
+### What is Tag?
+A Tag have three properties: `"name"`, `"color"` and `"icon"`.
+```json
```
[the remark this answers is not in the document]
uhoreg: I don't think this is intended to be a full  event?  It looks like this is just the definition for the `content` of an event.

proposals/3949-power-level-tag.md
FSG-Cat: Not the author but i mean unless this MSC gives a recomendation to change the PL template your client uses this would be arbitrary state PL ofc.

proposals/3949-power-level-tag.md
FSG-Cat: As addressed earlier. This is clearly a PL tags MSC and therefore makes no authorisation related changes. It only makes it so you can define that users at the PL of 1337 are Elite Haxors or whatever name you want to give to a given powerlevel. This proposal does implicit ranges. I personally would have if this wasnt made written a MSC that does Explicit ranges only. But as mentioned in another comment i dont want to compete with this MSC with my MSC that has diffrences due to the lack of improvements that are worth competing over when they are of potentially dubious value compared to this approach.

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
...
+  - `"on_light"` Tag color for lighter background.
+  - `"on_dark"` Tag color for darker background.
+* `"icon"` To display graphical representation of tags name. (*Optional*)
+
+### Attaching Tag to Power Level
+Tag can be attached to a power level by send a state event with key as `m.room.power_level_tag` and state key as value of power level.
+```json
+{
+  "type": "m.room.power_level_tag",
+  "state_key": "100",
```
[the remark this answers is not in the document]
uhoreg: I think the idea is that this just gives names to the numerical power levels, rather than trying to introduce a new permissions system.  So, say I have PL100 in a room via the normal power levels mechanism.  Then the client can look up the appropriate `m.room.power_level_tag` and find out that in that room, with PL100, I am to be referred to as "The Absolute Ruler".

proposals/3949-power-level-tag.md
uhoreg, line 74: Rather than using one state event per state key, you could put all the labels in a single event, with a map from PL to label.  e.g.

``` jsonc
{
  "type": "m.room.power_level_tags",
  "state_key": "",
  "content": {
    "100": {
      "name": "Admin",
      // ...
    }
  }
}
```

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,86 @@
+### MSC3949 Power Level Tags
+
+Currently permissions in a certain room are handle via `m.room.power_levels`. A power level is an integer value, bigger the integer bigger the power. When we assign power level to a user we gave them power represented as integer value. This integer value is abstract and does not tell what kind of power it represent. Currently spec only suggest mapping abstract integer value to idea as: 0 - 49 to User, 50 - 99 to Moderator - 100 to Admin. This suggestion limits the wide range of power level integers to only three idea's.
```
uhoreg, line 3: Please wrap your lines to ~80-120 characters

proposals/3949-power-level-tag.md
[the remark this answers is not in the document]
MTRNord: Ah yeah it just seemed weird as the state event is in full

proposals/3949-power-level-tag.md
ajbura: My only concern is event size overflow. Maybe we should restrict the name value to like 50 character?

proposals/3949-power-level-tag.md
ajbura: Most of the client use white and black backgrounds so both these properties make sense to maintain good color contrast for most usecases.

MTRNord: My concern is just that dark and light is a gradient not a constant. You can have dark and light and yet have different bg colors in clients. And with that you end up running into accessibility issues where the contrast might not be enough. (see accessibility spec for html/css World which needs a surprising large amount of contrast. So it is easy to mot have enough.)

FSG-Cat: I personally think that we cant do too much to mittigate accessibility issues caused by user definable colours. I think that the obvious fix accessibility wise is an accessibility option that disables this customisation venue.

proposals/3949-power-level-tag.md
uhoreg: Possibly.  I'm not suggesting that the alternative that I gave is better (or worse) than what you suggested, but it should probably be listed as an alternative, so that the pros and cons can be discussed.  The reason I suggested that alternative is because it is more similar to how some other state events work (notably the power levels event itself), so it seems to be an obvious way to do things.  Though of course, obvious doesn't mean better.

proposals/3949-power-level-tag.md
ajbura: Alternatively we can only use a `hue` (0 - 360) property to supply a base color and let the client decide how much light and brightness they want to add so it stay accessible in their design system. But this will limit the color choices.

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Proposal
+
+### What is Tag?
+
+A Tag have three properties: `"name"`, `"color"` and `"icon"`.
+
+```json
+{
+  "name": "Admin",
```
Fox32, line 26: How about providing translated names? Right now clients like Element translate the names into the user locale, so making this a map would be interesting. For example

```json
{
  "name": {
    "en": "Guest",
    "de": "Gast"
  }
  ...
}
```

The key would be an ISO locale code. If the requested locale is missing, we would need some kind of fallback to other locales, in the end falling back to `"en"`.

kfiven: It looks like this is going to be user input so translation does not make sense here. A community can label users with certain power level anything.

FSG-Cat: Well multilingual communities should be supported since they might bring translations for all langauges they use. For example a community for the Swedish Speakers in finland might bring a Swedish, Finish and English translation for their roles. And in more corporate usecases this will also be able to be utilised properly.

Fox32: In our case it is a cooperate use case, where multi-language role names are important. Right now it is solved by forking clients and replacing the translations with custom ones.

FSG-Cat: > @FSG-Cat How do you propose managing these translations? Since this is user input, and only some users can edit state events then only those will be able to translate them. Plus this makes the UX of roles in a client very complex and on top of that since power levels are room-specific, translating those in each room will be an extra burden.
> 
> IMO, if a community wants to add a multi-lingual role for say administrator with PL 100, they can create a Tag with name administrateur and PL99 and add those accordingly to users.

@kfiven I fail to see how the UX would be too complicated because of this. But lets put that besides for now. The translations would be community managed because its a Bring your own Translations system. Also on the note that its room specific yes and no. Its room specific but you can template this and like tbh PL sync is going to be a thing at this rate. I have personally thought about writing the MSC for it.

Also as for your using diffrent PL fix i say that is a wholy unacceptable compromise compared to the slightly more complicated UX or requirement to use specialised tooling like the Feline Matrix Assistant that i made or Matrix Wrench etc. Like its not a compromise that gets you this specific feature at all. You get forced to give admins a rank list as in who can demote who and also you also make it so their tag name is only accurate in a single language per admin meaning you cant know what their role is supposed to be in your language even if said community can translate into it unless said language is the language used for that admins tag. 

Also just to clarify on the nature of these translations. They are not translations where you try to cover all languages on earth. They are translations where you cover the languages used in your community. Therefore it does not matter that its limited scope as to who can write.

kfiven: I mean how will the client developers know which community wants which language in their PL translation, so clients have to suggest ISO codes of all languages. If we left it empty then it's up to the user to find out those, which again is bad UX for not-so-technical users.

On top of those, this introduces a parallel translation option to the one client uses to translate its interface, which is again bad UX for translators.

Lastly, I feel translating a state event is fundamentally wrong because with this logic a room name also could be another candidate for such translation, and room topic too.

maltee1: Could we partially solve this by defining default tags that we can then refer to by for example instead of using `"name" : "Moderator"`, we use `"default_name" : "Moderator"` to instruct the client to localize? A selection of maybe 5 different tags will be plenty for most use cases and only the remainder would have to deal with translating the tags themselves.

proposals/3949-power-level-tag.md
ajbura: I have replaced those with the single `hex` value. Client can actually make the color accessible by tuning the `hex` value by *any* way. for example they can extract the hue out of it or check the lightness/darkness and tweak them.

proposals/3949-power-level-tag.md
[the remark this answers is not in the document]
ajbura: Thanks. I have change it with the table layout.

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,107 @@
...
+
+Tags can be saved by sending a state event of type `m.room.power_level_tags` with `state_key` as
+empty string. The content will be a map of power level integer to Tag.
+
+```json
+{
+  "type": "m.room.power_level_tags",
+  "state_key": "",
+  "content": {
+    "100": {
```
TheArcaneBrony, line 48: How would one indicate the creator powerlevel in v12+?

ajbura: This msc is only for numeric power levels and doesn't cover "Creator" since creators are special privileged user and has no power levels. As per this msc it is only for custom labeling for abstract power level integers, if the "Creator" label need customization a separate msc can be created using same `Tag` schema in a dedicated state event.

1vel: I don't see why we can't just have a `creator` key in the event here. I also think it would be better to get both of them in this MSC, rather than another MSC to amend this one.

TheArcaneBrony: Creator is not an integer, so it makes serialisation a pain

proposals/3949-power-level-tag.md
nushea: While a singular color that could be modified does allow for contrast in multiple situations, in reality it would limit the user choice.

The original fields of on_light and on_dark would enable a user to define more preferred colors to be chosen from. For a personal example, while I do like having an #ffd9f5 HSL(315, 100, 92) color over a dark theme, I would hate having a #29001e (315, 100, 8) dirty brown or any other dirty pink that adjusting purely the saturation and lightness would produce, or a dark fuchsia that adjusting the value would. And that is a general idea that when people choose a color it is because they like that color in particular not necessarily the entire series of colors within that hue

Accessibility could be achieved by only adjusting the color when it does not create the required amount of contract, or disregarding it entirely when it does not

proposals/3949-power-level-tag.md
```diff
@@ -0,0 +1,107 @@
...
+      "name": "Admin",
+      "color": "#000000",
+      "icon": {
```
nushea, line 51: Maybe instead of the Icon it could be useful to have a full formatted_text style field where the icon and other items could be described, in order to aid in establishing who is a moderator or bot within a room at first sight (by having a specific color and having a "Mod Neofox holding gun emoticon" or "Bot Roomba with a knife emoticon" mark next to their name

PR #3953, opened 2023-01-07
MSC0000: Server capability DAG

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/capability-dag/proposals/XXXX-capability-dag.md)

### DIFF
--- a/proposals/XXXX-capability-dag.md
+++ b/proposals/XXXX-capability-dag.md
@@ -1,12 +1,13 @@
-# MSC0000: Replacing power levels with capabilities.
+# MSC0000: Server capability DAG
 
 ## The authority of homeservers
 
 Within Matrix, homeservers inherit any authority that any of its users has because of the power to
 impersonate users. This is especially true when a user on a homserver posses the the power level to
-change `m.room.power_levels`, as this means the server can then transfer power to any of its users.
+change `m.room.power_levels`, as this means the server can then also transfer power
+it posses to any other user or server.
 
-Homeserver also have the authority to generate any number of `m.room.member` state events for any
+Homeservers also have the authority to generate any number of `m.room.member` state events for any
 Matrix room where the `m.room.join_rules` is `public`. On top of this, if we want to restrict who
 can generate joins, the `m.room.server_acl` event has to anticipate the existance of any malicious
 server (if the allow rule is `*`) and is an entirely reactive measure.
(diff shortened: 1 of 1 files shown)


proposals/3953-capability-dag.md
```diff
@@ -0,0 +1,410 @@
...
+Instead we propose that membership of users be forgoed entirely and that instead server
+membership is the only part of room membership that is relevant to event authorization.[^1]
+
+[^1]: I believe this idea was originally discussed by Kegan and Till before but I got the idea from Kegan. Not
+sure that either will be thrilled by me appropriating it in this way.
+
+If we take this view, then what happens to user membership?
+In order to keep users out of the auth dag, their membership has to be treated as entirely virtual.
+Servers that possess membership, and also possess the capability to send events therefore will
+also have the authority to send events with any sender local part.
```
Gnuxie, line 60: So some initial feedback provided by @FSG-Cat & Nyaaori (not on Github) is that the MSC is still too focused on the sender local part and the concept of senders. As a result the proposal still leans towards / is coupled to current ideas about how to represent users and room members, and it is desirable to leave this open to interpretation at this point.

As a result, it has been suggested that an `origin` field be added to the top level of PDUs. Granted this is provided from the transactions endpoint, but while backfilling servers currently have to infer the origin from the `sender`.

proposals/3953-capability-dag.md
```diff
@@ -0,0 +1,410 @@
...
+## Potential issues
+
+We are sure they exist, and welcome suggestions but have not listed them yet.
+
+- Additional checks may be required to ensure the server is representing its user's presence
+  in the room consistently
+- We don't provide clear semantics for user level bans or attenuation of their capabilities
+- What is the process for transferring a capability to a user via a server from the perspective
+  of a client?
+- This is a pretty big breaking change, no client is equipped to deal with this,
```
Gnuxie, line 381: I believe that it is possible to proxy this room version to clients without changing the client server api depending on how you implement users/members in a dependent MSC, even if sender is removed from the top level of PDUs. Still stands that it probably isn't desirable to emulate the existing behavior and it would just complicate things.

proposals/3953-capability-dag.md
```diff
@@ -0,0 +1,410 @@
+# MSC3953: Server capability DAG
```
kegsay, line 1: This can be seen almost as the opposite of [MSC4014](https://github.com/matrix-org/matrix-spec-proposals/pull/4014) because 4014 demotes the importance of servers in the protocol to merely routing information, and instead it's entirely cryptographic keys all the way down to identify users. I think I prefer that to this proposal.

Gnuxie:

PR #3954, opened 2023-01-13
MSC3954: Extensible Events - Text Emotes

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/emotes/proposals/3954-extensible-events-emotes.md)

### DIFF
--- a/proposals/3954-extensible-events-emotes.md
+++ b/proposals/3954-extensible-events-emotes.md
@@ -1,0 +1,49 @@
+# MSC3954: Extensible Events - Text Emotes
+
+[MSC1767](https://github.com/matrix-org/matrix-doc/pull/1767) describes Extensible Events in detail,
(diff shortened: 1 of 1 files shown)


TheArcaneBrony:
Perhaps we could rename this to something like "m.action"?
I've always found it kinda weird to call it an "emote" when 99% of the time this has nothing to do with emotions, but is rather used to represent mentioning an action in third person. (Eg. "Bob is thinking", "Alice is reading", ...)

proposals/3954-extensible-events-emotes.md
turt2live, line 1: @TheArcaneBrony says:

> Perhaps we could rename this to something like "m.action"? I've always found it kinda weird to call it an "emote" when 99% of the time this has nothing to do with emotions, but is rather used to represent mentioning an action in third person. (Eg. "Bob is thinking", "Alice is reading", ...)

PR #3956, opened 2023-01-13
[WIP] MSC3956: Extensible Events - Encrypted Events

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/encryption/proposals/3956-extensible-events-encryption.md)

### DIFF
--- a/proposals/3956-extensible-events-encryption.md
+++ b/proposals/3956-extensible-events-encryption.md
@@ -1,0 +1,64 @@
+# MSC3956: Extensible Events - Encrypted Events
+
+[MSC1767](https://github.com/matrix-org/matrix-doc/pull/1767) describes Extensible Events in detail,
+though deliberately does not include schemas for encrypted events. This MSC covers a replacement
+for `m.room.encrypted` in rooms.
+
+*Rationale*: Splitting the MSCs down into individual parts makes it easier to implement and review in
+stages without blocking other pieces of the overall idea. For example, an issue with the way encryption
+is represented should not block the overall schema from going through.
+
+## Proposal
+
+**Author's note**: There is fairly strong, and reasonable, opposition to having encryption be a content
+block. This could theoretically allow an event to be partially encrypted, which is undesirable.
+
+Like `m.room.message`, `m.room.encrypted` is also deprecated in favour of a new `m.encrypted` event
+type. `m.encrypted` expects an `m.encrypted` content block, which is the current `content` schema for
+an `m.room.encrypted` event:
+
+```json5
+{
+    // irrelevant fields not shown
+    "type": "m.encrypted",
+    "content": {
+        "m.encrypted": {
+            "algorithm": "m.olm.v1.curve25519-aes-sha2",
+            "sender_key": "<sender_curve25519_key>",
+            "ciphertext": {
+                "<device_curve25519_key>": {
+                    "type": 0,
+                    "body": "<encrypted_payload_base_64>"
+                }
+            }
+        }
+    }
+}
+```
+
+This allows the `m.encrypted` content block to be reused by other event types, if required.
+
+For clarity, this is *not* intended to allow unencrypted fallback on an encrypted event - doing
+so would be extraordinarily dangerous and is explicitly discouraged.
+
+## Potential issues
+
+***TODO: Address author's note in proposal body***
+
+***TODO***
+
+## Alternatives
+
+***TODO***
+
+## Security considerations
+
+***TODO***
+
+## Unstable prefix
+
+While this MSC is not considered stable, implementations should use `org.matrix.msc1767.*` as a prefix in
+place of `m.*` throughout this proposal. Note that this uses the namespace of the parent MSC rather than
+the namespace of this MSC - this is deliberate.
+
+Note that extensible events should only be used in an appropriate room version as well.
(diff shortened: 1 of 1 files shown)


proposals/3956-extensible-events-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+for `m.room.encrypted` in rooms.
+
+*Rationale*: Splitting the MSCs down into individual parts makes it easier to implement and review in
+stages without blocking other pieces of the overall idea. For example, an issue with the way encryption
+is represented should not block the overall schema from going through.
+
+## Proposal
+
+**Author's note**: There is fairly strong, and reasonable, opposition to having encryption be a content
+block. This could theoretically allow an event to be partially encrypted, which is undesirable.
```
clokep, line 14: I am not agreeing nor disagreeing with the sentiment here, but it is already possible to have partially encrypted events [since `m.relates_to` is always in cleartext](https://spec.matrix.org/v1.5/client-server-api/#forming-relationships-between-events); it does create complexity though of how that interacts with a decrypted plaintext version.

proposals/3956-extensible-events-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+This allows the `m.encrypted` content block to be reused by other event types, if required.
+
+For clarity, this is *not* intended to allow unencrypted fallback on an encrypted event - doing
```
MTRNord, line 41: I wonder if it would make sense to have the spec say that a server impl should refuse a cleartext fallback if `m.encrypt` is present.

While this doesn't protect against people modifying their servers this would protect against faulty client implementations.

MTRNord: On further thinking and talking to people: wouldn't it make sense to even enforce this in the auth rules? Is there any case where checking wouldn't be trivial or where you need plaintext next to encrypted?

MTRNord: Reasoning behind doing this via auth rules would be that this is a security concern. So this would prevent a security relevant issue which would justify having strict validation of the event imho as security outweighs most use cases I can think of.

proposals/3956-extensible-events-encryption.md
turt2live: indeed, though the concern is more that someone will stick `m.markup` as a fallback on an encrypted event, which defeats the purpose of encryption. This proposal as-is also theoretically allows for weird parts of the event to be encrypted, though that's not intended to be valid.

proposals/3956-extensible-events-encryption.md
turt2live: At this stage, auth rules feel overly strict for the general protocol: while I can agree with the principle that information should be enforced as encrypted, the general purpose nature of the protocol would call for a less constraining approach. 

Also at this stage, I currently believe that adding auth rules to specific event types would be a design failure. Ideally, there aren't exceptions.

proposals/3956-extensible-events-encryption.md
```diff
@@ -0,0 +1,64 @@
...
+
+```json5
+{
+    // irrelevant fields not shown
+    "type": "m.encrypted",
+    "content": {
+        "m.encrypted": {
+            "algorithm": "m.olm.v1.curve25519-aes-sha2",
+            "sender_key": "<sender_curve25519_key>",
+            "ciphertext": {
```
tulir, line 28: I don't love the variable-type `ciphertext` in the old `m.room.encrypted` schema. Could it be split up into separate fields or something?

proposals/3956-extensible-events-encryption.md
```diff
@@ -0,0 +1,64 @@
+# MSC3956: Extensible Events - Encrypted Events
+
+[MSC1767](https://github.com/matrix-org/matrix-doc/pull/1767) describes Extensible Events in detail,
+though deliberately does not include schemas for encrypted events. This MSC covers a replacement
+for `m.room.encrypted` in rooms.
```
tulir, line 5: The example event uses Olm rather than Megolm. Does this MSC also cover a replacement for `m.room.encrypted` in to-device events?

turt2live: toDevice is message-based, not event-based. 

Extensible Events does not consider toDevice in scope.

proposals/3956-extensible-events-encryption.md
turt2live: as mentioned in your other comment, it looks like I copy/pasted the wrong example. Not sure we can split up a string.

proposals/3956-extensible-events-encryption.md
```diff
@@ -0,0 +1,61 @@
...
+
+```json5
+{
+    // irrelevant fields not shown
+    "type": "m.encrypted",
+    "content": {
+        "m.encrypted": {
+            "algorithm": "m.megolm.v1.aes-sha2",
+            "sender_key": "<sender_curve25519_key>",
+            "device_id": "<sender_device_id>",
```
HarHarLinks, line 28: as i understand current spec, device_id is deprecated and if not already could be dropped by this MSC. sender_key is only used for olm (hence technically shouldn't be present in this example)

https://spec.matrix.org/v1.18/client-server-api/#mmegolmv1aes-sha2

PR #3958, opened 2023-01-24
MSC3958: Suppress notifications from message edits

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/suppress-message-edits/proposals/3958-suppress-notifications-of-message-edits.md)

Implementation:

* Beeper's Synapse fork: https://github.com/beeper/synapse/blame/beeper/rust/src/push/base_rules.rs#L98-L114
* Mainline Synapse: https://github.com/matrix-org/synapse/pull/14960
* Updated implementation: matrix-org/synapse#15992

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3958#issuecomment-1651847253)

### DIFF
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -89,6 +89,31 @@ when we generate notifications.
 If an event is edited and the new event (but not the original event) matches a keyword
 then the notification would erroneously be suppressed.
 
+### Edits of rooms set to "all messages"
+
+A room can be configured to be notify for "all messages" by creating the earliest
+[override push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules)
+possible which matches the room ID & has `actions` set to "notify" [^1][^2], e.g.:
+
+```json
+{
+    "rule_id" : "!abcdef:example.com",
+    "conditions" : [
+       {
+          "key" : "room_id",
+          "kind" : "event_match",
+          "pattern" : "!abcdef:example.com"
+       }
+    ],
+    "default" : false,
+    "enabled" : true,
+    "actions" : ["notify"]
+}
+```
+
+Since this push rule would be evaluated before the new `.m.rule.suppress_edits` it would
+still result in duplicate notifications for those rooms.
+
 ## Alternatives
 
 An alternative solution would be to add a push rule with no actions and a condition to
@@ -123,3 +148,9 @@ A previous version of this MSC used `.com.beeper.suppress_edits` with a differen
 ## Dependencies
 
 N/A
+
+[^1]: The [`.m.rule.master`](https://spec.matrix.org/v1.7/client-server-api/#default-override-rules)
+is *always* first, so this rule gets created right after it.
+
+[^2]: See the [Element Web](https://github.com/matrix-org/matrix-react-sdk/blob/da7aa4055e04f291be9b5141b704bd12aec03d0c/src/RoomNotifs.ts#L162-L170)
+implementation.
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -11,7 +11,7 @@ is re-notified). This contributes to notification fatigue as the additional
 notifications contain no new information.
 
 Additionally for users which have a room set to "all messages" then every event
-edit results in a new notification.
+edit results in an additional notification.[^1]
 
 ## Proposal
 
@@ -89,31 +89,6 @@ when we generate notifications.
 If an event is edited and the new event (but not the original event) matches a keyword
 then the notification would erroneously be suppressed.
 
-### Edits of rooms set to "all messages"
-
-A room can be configured to be notify for "all messages" by creating the earliest
-[override push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules)
-possible which matches the room ID & has `actions` set to "notify" [^1][^2], e.g.:
-
-```json
-{
-    "rule_id" : "!abcdef:example.com",
-    "conditions" : [
-       {
-          "key" : "room_id",
-          "kind" : "event_match",
-          "pattern" : "!abcdef:example.com"
-       }
-    ],
-    "default" : false,
-    "enabled" : true,
-    "actions" : ["notify"]
-}
-```
-
-Since this push rule would be evaluated before the new `.m.rule.suppress_edits` it would
-still result in duplicate notifications for those rooms.
-
 ## Alternatives
 
 An alternative solution would be to add a push rule with no actions and a condition to
@@ -149,8 +124,19 @@ A previous version of this MSC used `.com.beeper.suppress_edits` with a differen
 
 N/A
 
-[^1]: The [`.m.rule.master`](https://spec.matrix.org/v1.7/client-server-api/#default-override-rules)
-is *always* first, so this rule gets created right after it.
+<!-- Footnotes below -->
 
-[^2]: See the [Element Web](https://github.com/matrix-org/matrix-react-sdk/blob/da7aa4055e04f291be9b5141b704bd12aec03d0c/src/RoomNotifs.ts#L162-L170)
-implementation.
+[^1]: A room can be configured to be notify for "all messages" by creating a [room-specific push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules)
+with an `rule_id` of the room ID & has `actions` set to "notify" , e.g.:
+
+    ```json
+    {
+        "rule_id" : "!abcdef:example.com",
+        "default" : false,
+        "enabled" : true,
+        "actions" : ["notify"]
+    }
+    ```
+
+    See the [Element Web](https://github.com/matrix-org/matrix-react-sdk/blob/da7aa4055e04f291be9b5141b704bd12aec03d0c/src/RoomNotifs.ts#L162-L170)
+    implementation.
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -12,7 +12,7 @@ notifications contain no new information.
 
 ## Proposal
 
-A new default push rule is added to suppress notifications [due to edits](https://spec.matrix.org/v1.5/client-server-api/#event-replacements).
+A new default push rule is added to suppress notifications due to [edits](https://spec.matrix.org/v1.5/client-server-api/#event-replacements).
 
 ```json
 {
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -30,7 +30,8 @@ A new default push rule is added to suppress notifications [due to edits](https:
 }
 ```
 
-This rule should be placed as the [first override rule](https://spec.matrix.org/v1.5/client-server-api/#default-override-rules).
+This rule should be placed before the [`.m.rule.suppress_notices` rule](https://spec.matrix.org/v1.5/client-server-api/#default-override-rules)
+as the first non-master, non-user added override rule.
 
 It would match events such as those given in [event replacements](https://spec.matrix.org/v1.5/client-server-api/#event-replacements)
 portion of the spec:
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -1,0 +1,80 @@
+# MSC3958: Suppress notifications of message edits
+
+[Event replacement](https://spec.matrix.org/v1.5/client-server-api/#event-replacements)
+(more commonly known as message edits) signals that a message is intended to
+be replaced with new content.
+
+This works well for fixing typos or correcting information, but tends to cause
+spurious notifications if the event mentions a user. This contributes to notification
+fatigue as the additional notifications contain no new information.
+
+
+## Proposal
+
+A new default push rule is added to suppress notifications [due to edits](https://spec.matrix.org/v1.5/client-server-api/#event-replacements).
+
+```json
+{
+    "rule_id": ".m.rule.suppress_edits",
+    "default": true,
+    "enabled": true,
+    "conditions": [
+        {
+            "kind": "event_match",
+            "key": "content.m.relates_to.rel_type",
+            "pattern": "m.replace"
+        }
+    ],
+    "actions": []
+}
+```
+
+This rule should be placed as the [first override rule](https://spec.matrix.org/v1.5/client-server-api/#default-override-rules).
+
+It would match events such as those given in [event replacements](https://spec.matrix.org/v1.5/client-server-api/#event-replacements)
+portion of the spec:
+
+```json5
+{
+    "type": "m.room.message",
+    "content": {
+        "body": "* Hello! My name is bar",
+        "msgtype": "m.text",
+        "m.new_content": {
+            "body": "Hello! My name is bar",
+            "msgtype": "m.text"
+        },
+        "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$some_event_id"
+        }
+    },
+    // ... other fields required by events
+}
+```
+
+## Potential issues
+
+Some users may be depending on notifications of edits. If a user would like to
+revert to the old behavior they can disable the `.m.rule.suppress_edits` push rule.
+
+If the message edits were alloewd by other senders than it would be useful to
+know that your own message was edited, but this
+[is not currently allowed](https://spec.matrix.org/v1.5/client-server-api/#validity-of-replacement-events).
+
+## Alternatives
+
+None explored.
+
+## Security considerations
+
+None forseen.
+
+## Unstable prefix
+
+The unstable prefix of `.com.beeper.suppress_edits` should be used in place of
+`.m.rule.suppress_edits`.
+
+## Dependencies
+
+N/A
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -4,10 +4,11 @@
 (more commonly known as message edits) signals that a message is intended to
 be replaced with new content.
 
-This works well for fixing typos or correcting information, but tends to cause
-spurious notifications if the event mentions a user. This contributes to notification
-fatigue as the additional notifications contain no new information.
-
+This works well for fixing typos or other minor correction, but can cause
+spurious notifications if the event mentions a user's display name / localpart or
+if it includes `@room` (which is particularly bad in large rooms as every user
+is re-notified). This contributes to notification fatigue as the additional
+notifications contain no new information.
 
 ## Proposal
 
--- a/proposals/3958-suppress-notifications-of-message-edits.md
+++ b/proposals/3958-suppress-notifications-of-message-edits.md
@@ -90,6 +90,13 @@ due to the `.` in `m.relates_to` and could also match other, unrelated, events:
 
 (Note that `relates_to` being embedded inside of the `m`.)
 
+Another issues is that mobile clients would no longer receive push notifications for
+message edits, which are currently used to update the text of on-screen notifications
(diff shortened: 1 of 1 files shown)


proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,80 @@
...
+    // ... other fields required by events
+}
+```
+
+## Potential issues
+
+Some users may be depending on notifications of edits. If a user would like to
+revert to the old behavior they can disable the `.m.rule.suppress_edits` push rule.
+
+If the message edits were alloewd by other senders than it would be useful to
```
sumnerevans, line 61:
```suggestion
If the message edits were allowed by other senders than it would be useful to
```

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,80 @@
+# MSC3958: Suppress notifications of message edits
+
+[Event replacement](https://spec.matrix.org/v1.5/client-server-api/#event-replacements)
+(more commonly known as message edits) signals that a message is intended to
+be replaced with new content.
+
+This works well for fixing typos or correcting information, but tends to cause
+spurious notifications if the event mentions a user. This contributes to notification
```
sumnerevans, line 8: It may be worth mentioning that it's especially bad with `@room` messages that currently re-notify many people in a room (most of the time noisily as well).

sumnerevans (commented):
Thanks for writing this up!

proposals/3958-suppress-notifications-of-message-edits.md
clokep: Oh, yes that does make it worse. 👍

turt2live:
The author believes this is ready for formal review, potentially for FCP.

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,107 @@
...
+
+```json
+{
+    "rule_id": ".m.rule.suppress_edits",
+    "default": true,
+    "enabled": true,
+    "conditions": [
+        {
+            "kind": "event_match",
+            "key": "content.m.relates_to.rel_type",
```
deepbluev7, line 25: Since this matches on dotted keys, it might run into issues because of https://github.com/matrix-org/matrix-spec-proposals/pull/3873

Probably it is not an issue, since the m namespace is reserved and as such there should be no field with an "m" key. However, this might cause trouble when 3873 is merged, since this is the first pushrule matching on a dotted key, afaik.

It might be fine to ignore that on this MSC, I just wanted to mention it to be safe, if anyone has any smart ideas about it.

clokep: Thanks! I've included this under the potential issues section! 👍

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,107 @@
...
+        "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$some_event_id"
+        }
+    },
+    // ... other fields required by events
+}
+```
+
+## Potential issues
```
anoadragon453, line 57: Another potential issue is that this breaks the use case of editing a message to include an additional push target. For instance, if you write a message and then edit it to put another user pill in it, in this case the user would not be notified.

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    },
+    // ... other fields required by events
+}
+```
+
+(Note that `relates_to` being embedded inside of the `m`.)
+
+## Alternatives
```
anoadragon453, line 92: An alternative solution would be to add a condition to check whether a notification should have been generated for the user in the original message (following a capped number of `m.relates_to` jumps), and if so don't notify again.

Then modify the `.m.rule.contains_display_name` and the `.m.rule.roomnotif` [push rules](https://spec.matrix.org/v1.5/client-server-api/#push-rules) to include this condition.

clokep: This seems significantly more complicated for an edge-case (IMO). Is it that common to edit a message to add another person?

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,107 @@
...
+            "kind": "event_match",
+            "key": "content.m.relates_to.rel_type",
+            "pattern": "m.replace"
+        }
+    ],
+    "actions": []
+}
+```
+
+This rule should be placed as the [first override rule](https://spec.matrix.org/v1.5/client-server-api/#default-override-rules).
```
richvdh, line 33: It's worth noting that currently we have:

 1. `.m.rule.master`
 2. user-added override rules
 3. other default override rules.

Could you be more explicit about where in this ordering you think the new rule should go?

richvdh (approved):
generally seems good, but I think I'd like to hear from people testing it. Can we get an experimental implementation in synapse mainline that I can turn on on my dev homeserver?

proposals/3958-suppress-notifications-of-message-edits.md
clokep: Yep, that's where I'm thinking it should go! 👍  I'll re-word to be more explicit.

clokep:
> generally seems good, but I think I'd like to hear from people testing it. Can we get an experimental implementation in synapse mainline that I can turn on on my dev homeserver?

See matrix-org/synapse#14960 for an upstreamed version of Beeper's code behind a config flag.

proposals/3958-suppress-notifications-of-message-edits.md
clokep:
```suggestion
This rule should be placed before the [`.m.rule.suppress_notices` rule in the list of override rules](https://spec.matrix.org/v1.5/client-server-api/#default-override-rules) as the first non-master,
non-user added override rule.
```

richvdh (approved):

richvdh:
@mscbot fcp merge

proposals/3958-suppress-notifications-of-message-edits.md
richvdh: I agree this is an edge-case and I think we can descope it.

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,108 @@
...
+        "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$some_event_id"
+        }
+    },
+    // ... other fields required by events
+}
+```
+
+## Potential issues
```
richvdh, line 58: another issue seems to be around mobile clients, which might display a notification message when a DM is received, on the back of a push notification.

Currently, the mobile clients receive a push after each edit, which means they can update the text of an on-screen notification to show the updated content. The new push rule will of course inhibit the push, so the mobile client gets stuck with the old content.

I'm much of the opinion this is an unavoidable result of us trying to do too much magic with push rules, and that the new behaviour is probably a lesser evil. Still, worth calling out.

proposals/3958-suppress-notifications-of-message-edits.md
clokep: I added some info into the alternatives section about this.

proposals/3958-suppress-notifications-of-message-edits.md
clokep: I fleshed out the potential issues section a bit more. Thanks!

clokep: I fleshed out the potential issues section a bit more. Thanks!

proposals/3958-suppress-notifications-of-message-edits.md
clokep: Thinking on this more, I wonder if this MSC is even necessary if MSC3952 lands. In that MSC we wouldn't create mentions *unless* the user is explicitly in `m.mentions: {user_ids: [...]}`. If you edit a message the edited content goes into `m.new_content`. This does mean we could reserve the top-level `m.mentions` for mention information:

* If it isn't provided (or is empty, in compatibility mode) then no mentions.
* But a client could optionally provide it if e.g. they added a new person to the original message, which case they could mention *just* that person.

It does put a bit more onto the client generating the message, but is a bit more stateless.

WDYT @anoadragon453?

dbkr:
Yes please!

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,115 @@
...
+
+Some users may be depending on notifications of edits. If a user would like to
+revert to the old behavior they can disable the `.m.rule.suppress_edits` push rule.
+
+If the message edits were allowed by other senders than it would be useful to
+know that your own message was edited, but this
+[is not currently allowed](https://spec.matrix.org/v1.5/client-server-api/#validity-of-replacement-events).
+
+The rule is ambiguous (see [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873))
+due to the `.` in `m.relates_to` and could also match other, unrelated, events:
```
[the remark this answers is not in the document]
uhoreg: Yeah, I'd say it's probably worth adding a sentence saying what the `key` would be under MSC3873.

proposals/3958-suppress-notifications-of-message-edits.md
uhoreg: This sounds to me like a reasonable solution.  The editing client can set the `m.mentions` so that it pings people who have been added, and avoid re-pinging people who have already been pinged.  Or it could maybe have the option to re-ping someone if the edit is significant and the sender thinks the recipient needs to re-look at the message?

uhoreg: @mscbot concern does MSC3952 make this unnecessary?

clokep: Thanks @uhoreg! Was just coming to cross-link https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/intentional-mentions/proposals/3952-intentional-mentions.md#impact-on-edits

This *might* still be useful as a short-term fix until clients start sending (and understanding) the `m.mentions` property?

anoadragon453: I agree that MSC3952 could essentially replace this MSC, but that it would take longer to propagate.

I fear that there may be a likely scenario where a client ends up in a world where both intentional mentions and the `.m.rule.suppress_edits` push rule proposed by this MSC exist. Perhaps if the homeserver has added this push rule and the client itself supports intentional mentions.

In that case, the user of that client would not receive (unencrypted) notifications even if an edit explicitly added them to the `m.mentions` array, as the `.m.rule.suppress_edits` push rule would end up suppressing it.

Having both MSCs around at once may create a scenario where clients have to implement logic to explicitly disable the `.m.rule.suppress_edits` push rule in order to not break the use case of mentioning a user only in a message edit.

The only way I could see getting around this is by adding another condition to `.m.rule.suppress_edits` that the event does not contain an `m.mention` array in the `content`, effectively disabling this MSC if the sender of an event is using intentional mentions.

anoadragon453: Do any other @matrix-org/spec-core-team members have thoughts on this concern?

richvdh: Er, my thought is that short-term fixes to the push rules system end up making it even harder to reason about than it already is. If we think MSC3952 makes this redundant, then I would vote that we focus on getting that through. We've lived with spurious notifications from edits for four years; what's another few months?

clokep: Another argument in *favor* of this is that there may be other push rules that match the edit; mentions are just a subset of why you might get a duplicate notification, e.g. if you have keywords configured (or set a room to notify you for each message) then you would still get duplicate notifications without this MSC.

In light of @anoadragon453's concerns though, maybe this should be made a lower priority rule than some of the others.

turt2live: Playing with the priorities a bit to try and work with MSC3952 I think would make sense. MSC3952 doesn't have the formal confidence written down yet, but given the context from other places it seems likely to be successful enough to be worth considering as a complimenting MSC.

proposals/3958-suppress-notifications-of-message-edits.md
sumnerevans: I think that we will likely need to separate pushing vs. notifications in order to properly handle dismissing notifications on mobile platforms when the message is read on another device.

turt2live:
Given the direction of https://github.com/matrix-org/matrix-spec-proposals/pull/3958#discussion_r1153656716 I'm inclined to believe this MSC needs more time in the iteration stages (ie: FCP won't help it). 

Process-wise, this means the MSC is formally being moved off the Spec Core Team's list. Please raise the MSC in the [SCT office](https://matrix.to/#/#sct-office:matrix.org) when the SCT's attention would be helpful.

@mscbot fcp cancel

proposals/3958-suppress-notifications-of-message-edits.md
clokep: Taking into account intentional mentions (MSC3952), I think what we want to do is move this to be the first *under*ride rule (instead of the first *over*ride rule). With that the following could occur:

* Invites, tombstones, user, and room mentions should continue causing highlight notifications. ✅ 
* Edits to invites and tombstones would still result in a highlight notification (are edits of state events even valid?) ❌ 
* Legacy mentions would still result in highlight notifications. 🤷 
	* Edits to intentional mentions would not result in notifications due to MSC3952 logic. ✅ 
* Keywords would still result in notifications ✅ 
	* Edits to events with keywords would also result in notifications. ❌ 
* Edits to calls, messages, encrypted messages, messages in 1-to-1s, encrypted messages in 1-to-1s would all no longer notify. ✅ 

That all sounds good except for the edits to events with keywords. To get around that we could put the rule as the *last* override rule, so after `.m.rule.room.server_acl`. With that change we would fix the edits to events with keywords being an issue. Edits to invites / tombstones might still cause a notification, but I think that would be very uncommon?

proposals/3958-suppress-notifications-of-message-edits.md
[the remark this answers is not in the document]
clokep: That landed so I've updated this. 👍

proposals/3958-suppress-notifications-of-message-edits.md
clokep: @anoadragon453 I'd be curious if the current reading of the MSC sounds good? I think it hits all your concerns.

turt2live: > (are edits of state events even valid?)

technically yes, but silly and does nothing of value. The `m.new_content` isn't applied for things like event auth, so you're still dependent on the top level `content` keys existing, making it less effort to just `PUT` a new state event instead of crafting an edit.

clokep: > > (are edits of state events even valid?)
> 
> technically yes, but silly and does nothing of value. The `m.new_content` isn't applied for things like event auth, so you're still dependent on the top level `content` keys existing, making it less effort to just `PUT` a new state event instead of crafting an edit.

The spec says it is [not valid](https://spec.matrix.org/v1.7/client-server-api/#validity-of-replacement-events):

> The replacement and original events must not have a `state_key` property (i.e. you cannot edit state events at all).

turt2live: 👍 (/me flips table as Ctrl+F failed to find that)

richvdh: (just in case anyone is interested: https://github.com/vector-im/element-web/issues/21851 demonstrates the sort of mess you get into if you allow state events to be edited)

anoadragon453: The current reading of the MSC looks good to me indeed!

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,125 @@
...
+rules may still cause spurious notifications if events which match those rules
+are edited. Both of those are state events and
+[not subject to valid edits](https://spec.matrix.org/v1.7/client-server-api/#validity-of-replacement-events).
+
+### Keeping notifications up-to-date
+
+Mobile clients currently depend on the push notifications of edited events to update the
+text of on-screen notifications. The proposed push rule would result in mobile clients no
+longer receiving these edits; but showing slightly outdated text on a notification screen
+is only a minor impact and it would be better to separate when (& why) we send pushes vs.
```
anoadragon453, line 84: A small oddity here.

```suggestion
longer receiving these edits; but showing slightly outdated text on a notification screen. That
is only a minor impact and it would be better to separate when (& why) we send pushes vs.
```

anoadragon453 (approved):

proposals/3958-suppress-notifications-of-message-edits.md
```diff
@@ -0,0 +1,125 @@
...
+be replaced with new content.
+
+This works well for fixing typos or other minor correction, but can cause
+spurious notifications if the event mentions a user's display name / localpart or
+if it includes `@room` (which is particularly bad in large rooms as every user
+is re-notified). This contributes to notification fatigue as the additional
+notifications contain no new information.
+
+Additionally for users which have a room set to "all messages" then every event
+edit results in a new notification.
```
clokep, line 14: The downside of moving this below the the `is_user_mention` / `is_room_mention` rules is that if a room is set to *all* notifications than edits will still result in notifications. I don't think it is possible to handle both those situations, unfortunately.

I think the options are:

1. The old version which would not allow an edit to add a mention of a user.
2. The new version which will result in duplicate notifications for edits when a room is set to "all messages" for notifications.

clokep: https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/suppress-message-edits/proposals/3958-suppress-notifications-of-message-edits.md#edits-of-rooms-set-to-all-messages might explain this better than my comment.

anoadragon453: Forced to make a choice, I would go for 1., as I think a user would understand a mention added in an edit not actually mentioning a user, whereas a notification for edits in rooms set to "All Messages" does come across as a bug.

---

But would it not be possible to have the push rules set by "All Messages" (which, to be clear, is an Element-only construct correct?), include a condition to not match on message edits?

Unfortunately, when trying to come up with an example, I don't see an appropriate condition that's already spec'd. You would want something like:

```json5
    "conditions": [
        {
            "kind": "event_property_is_not", // note the "_not"
            "value": "m.replace"
        }
    ],
```

clokep: That would be doable, we would need to create an `event_property_is_not` condition, however.

clokep: Actually...looking again at https://github.com/matrix-org/matrix-react-sdk/blob/da7aa4055e04f291be9b5141b704bd12aec03d0c/src/RoomNotifs.ts#L162-L170, it seems to be *room-specific*, not global. So maybe I'm wrong at where that push rule is placed.

/me double checks some stuff

clokep: OK, so I completely botched my testing, this works as expected since the "all message" rule is a *room-specific* rule, while the "mute" rule is an override rule and I got them confused. 😢

anoadragon453:
The author believes this is once again ready for FCP. After looking at the updated implementation in the PR description, so do i!

@mscbot fcp merge

turt2live:
This MSC has completed FCP per the [above comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3958#issuecomment-1671589593).

clokep:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1617

anoadragon453:
Merged 🎉

PR #3967, opened 2023-02-15
MSC3967: Do not require UIA when first uploading cross signing keys

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/device-signing-upload-uia/proposals/3967-device-signing-upload-uia.md)

Feb 2024: this MSC has received updates, please re-review.

Implementations:

- [x] Synapse: https://github.com/matrix-org/synapse/pull/15077
- [x] All Element clients already support the scenario of no UIA challenge being received
- [x] Idempotency patch https://github.com/element-hq/synapse/pull/16943

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3967#issuecomment-2059104784)


proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,55 @@
...
+[configurable grace period](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#ui_auth)
+(`ui_auth`.`session_timeout`) in Synapse whereby UIA will not be required for uploading cross-signing keys where
+authentication has taken place recently.
+
+This proposal aims to provide for a standard way to address this UIA usability issue with respect to setting up cross-signing.
+
+## Proposal
+
+For the `POST /_matrix/client/v3/keys/device_signing/upload` endpoint the Homeserver should only require User-Interactive
+authentication (UIA) if a user has an existing cross-signing master key uploaded to the Homeserver.
```
dkasak, line 23:
```suggestion
For the `POST /_matrix/client/v3/keys/device_signing/upload` endpoint, the Homeserver should only require User-Interactive
authentication (UIA) if the already user has an existing cross-signing master key uploaded to the Homeserver.
```

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,55 @@
...
+## Potential issues
+
+The Client-Server API [spec](https://spec.matrix.org/v1.6/client-server-api/#user-interactive-api-in-the-rest-api) states:
+
+> A request to an endpoint that uses User-Interactive Authentication never succeeds without auth. Homeservers may allow
+requests that don’t require auth by offering a stage with only the m.login.dummy auth type, but they must still give a
+401 response to requests with no auth data.
+
+Does this mean that the endpoint can't simply give back a `200` response in the case that `auth` was not given as an
+input? If this is case then the existing Synapse behaviour of allowing a UIA grace period is probably also non-compliant.
```
dkasak, line 34:
```suggestion
input? If this is the case then the existing Synapse behaviour of allowing a UIA grace period is probably also non-compliant.
```

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,55 @@
...
+
+This change could be viewed as a degradation of security at the point of setting up cross-signing in that it requires
+less authentication to upload cross-signing keys on first use.
+
+However, this degradation needs to be weighed against the typical real world situation where a Homeserver will be
+applying a grace period and so allow a malicious actor to bypass UIA for a period of time after each authentication.
+
+## Unstable prefix
+
+Not applicable as a client behaviour need not change.
```
dkasak, line 51:
```suggestion
Not applicable as client behaviour need not change.
```

dkasak (approved):
Looks good to me. Left a few minor style fixes.

proposals/3967-device-signing-upload-uia.md
hughns: I've applied this and made some more tweaks. Thank you.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,55 @@
...
+## Potential issues
+
+The Client-Server API [spec](https://spec.matrix.org/v1.6/client-server-api/#user-interactive-api-in-the-rest-api) states:
+
+> A request to an endpoint that uses User-Interactive Authentication never succeeds without auth. Homeservers may allow
+requests that don’t require auth by offering a stage with only the m.login.dummy auth type, but they must still give a
+401 response to requests with no auth data.
+
+Does this mean that the endpoint can't simply give back a `200` response in the case that `auth` was not given as an
+input? If this is the case then the existing Synapse behaviour of allowing a UIA grace period is probably also non-compliant.
```
richvdh, line 34: in theory, yes. And yes, synapse is non-compliant. The idea is that making a request without an `auth` property is a way to find out what authentication is required.

Could we mandate that clients hoping to take advantage of this exemption proactively supply an `m.login.dummy` auth dict?

hughns: Okay. I've made https://github.com/matrix-org/synapse/issues/15166 about the Synapse compliance. I don't think it needs to impact this MSC so I will update this section.

BillCarsonFr (approved):
LGTM

turt2live:
@matrix-org/spec-core-team : the author is seeking feedback which relates to the overarching mission of OIDC-for-Matrix. Please provide some thoughts on whether this approach is workable before we get too far along in the OIDC transition to fix it :)

For the author's own words on this request, see [this message](https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$4E9SUEMKTovQ8yEuPx5zUIFIkgZ4qzfIC4wKgewdTHk?via=matrix.org&via=element.io&via=envs.net).

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Security considerations
+
+This change could be viewed as a degradation of security at the point of setting
+up cross-signing in that it requires less authentication to upload cross-signing
+keys on first use.
+
+However, this degradation needs to be weighed against the typical real world
+situation where a Homeserver will be applying a grace period and so allow a
+malicious actor to bypass UIA for a period of time after each authentication.
```
turt2live, line 49: Also chances are the user has just entered their password on the registration form, so they'll likely remember it anyhow.

A possible security issue around this is accounts which are ancient (created more than an hour in the past) and under control of an attacker: they could use the user's existing non-encrypted reputation to potentially get people to trust a new master key. However, this is adequately covered by the "should" in the proposal - the homeserver can decide it's been too long and enact UIA.

turt2live (approved):

turt2live:
@mscbot fcp merge

proposals/3967-device-signing-upload-uia.md
hughns: Yes, there could indeed be scenarios where the homeserver admin wishes to implement some alternative logic depending on the account/user state or other policy.

This could be clarified in the actual spec changes such that a client should be prepared to complete UIA even if it is the first time keys are being uploaded.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,57 @@
...
+(`ui_auth`.`session_timeout`) in Synapse whereby UIA will not be required for
+uploading cross-signing keys where authentication has taken place recently.
+
+This proposal aims to provide for a standard way to address this UIA usability
+issue with respect to setting up cross-signing.
+
+## Proposal
+
+For the `POST /_matrix/client/v3/keys/device_signing/upload` endpoint the
+Homeserver should only require User-Interactive Authentication (UIA) if the user
```
dbkr, line 27: I think this is unquestionably a good idea and is entirely sensible, but I'm also not sure you need an MSC for this? I think it depends on the 'should' in this sentence. If it were a 'must', clients could rely on any server claiming to support a recent enough version of the spec allowing this request without auth. If it's just a recommended behaviour, aren't servers free to just allow the request with no auth? (Although also rich's comment below about being more specific about no auth dict vs `m.login.dummy`). That said, an MSC to set intentions could still be useful, but there may be no need for this to block you.

turt2live: It's hard to tell from the spec, but I think we do actually require an MSC to at least make the intention clear. What "should" actually equates to, as has been pointed out, is a question, though the actual behaviour of whether it's legal to return no flows, etc is undetermined at present. This MSC would codify that.

turt2live:
@mscbot concern unclear what "not requiring UIA" actually means

turt2live:
Cancelling FCP while this isn't considered a priority. 

@mscbot fcp cancel

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,57 @@
...
+already has an existing cross-signing master key uploaded to the Homeserver.
```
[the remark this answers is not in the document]
kegsay: I would prefer not sending the `auth` dict at all, so literally it does not need anything to do with UIA until the x-signing key has been uploaded. This would not require a spec wording change so long as it's clear that the first upload is _not treated as a UIA endpoint_.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,57 @@
+# MSC3967: Do not require UIA when first uploading cross signing keys
```
kegsay, line 1: This has come up again in the context of rust SDK crypto, whereby if the initial x-signing upload request fails, rust SDK currently shrugs and gives up because it cannot retry due to UIA blocking it (so would need a whoooole bunch of machinery to pass through UIA, show UI prompts, etc). The merge was blocked on a very small technicality: can we agree what we mean by "not needing UIA" and then land this, so SDK authors can actually do retries without worrying about hitting a UIA wall?

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,57 @@
...
+
+This proposal aims to provide for a standard way to address this UIA usability
+issue with respect to setting up cross-signing.
+
+## Proposal
+
+For the `POST /_matrix/client/v3/keys/device_signing/upload` endpoint the
+Homeserver should only require User-Interactive Authentication (UIA) if the user
+already has an existing cross-signing master key uploaded to the Homeserver.
+
```
kegsay, line 29: We also need to be clear what "an existing cross-signing master key" means. It is critical for reliability that the endpoint is idempotent and doesn't suddenly UIA when the client tries to upload the exact same set of x-signing keys. Otherwise, a client that dies mid-way through uploading the keys (so the response is lost) will retry on startup and then get UIA'd, desyncing the client. This is part of the [state synchronisation](https://github.com/matrix-org/complement-crypto/blob/main/TEST_HITLIST.md#state-synchronisation) tests in complement-crypto.

kegsay: Note the synapse implementation does not do this, merely checking for a key existing and then UIA'ing https://github.com/matrix-org/synapse/pull/15077/files

kegsay: https://github.com/matrix-org/matrix-spec-proposals/pull/3967 landed.

richvdh: > #3967 landed.

Is that the PR you meant to link?

richvdh: @kegsay did you mean https://github.com/element-hq/synapse/pull/16943?

richvdh:
This looks good to me.

@mscbot fcp merge

proposals/3967-device-signing-upload-uia.md
clokep: > A possible security issue around this is accounts which are ancient (created more than an hour in the past) and under control of an attacker

It is unclear from the above whether it is expected if servers will apply limitations such that this only applies right after registration or not. I'd really like to see this security consideration explored more before merging.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,70 @@
...
+uploading cross-signing keys where authentication has taken place recently.
+
+This proposal aims to provide for a standard way to address this UIA usability
+issue with respect to setting up cross-signing.
+
+## Proposal
+
+For the `POST /_matrix/client/v3/keys/device_signing/upload` endpoint, the
+Homeserver MUST require User-Interactive Authentication (UIA) _unless_:
+ - there is no existing cross-signing master key uploaded to the Homeserver, OR
```
turt2live, line 28: This feels a bit risky - if the user's account is quite old, they're suddenly able to take over the account via cross-signing verification. I'd suggest language to limit this to time since login specifically. Ideally we'd tie it to account age, but if someone logs in with the user's password, a second UIA prompt for that very same password won't help anyone.

turt2live: (new thread as the condition appears to have been upgraded from SHOULD to MUST)

turt2live (changes requested):

proposals/3967-device-signing-upload-uia.md
turt2live: I've raised a thread in context, given the MUST in the MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3967#discussion_r1567696484

turt2live:
For https://github.com/matrix-org/matrix-spec-proposals/pull/3967#discussion_r1567696484

@mscbot concern Language appears to change security conditions

proposals/3967-device-signing-upload-uia.md
kegsay: Closing in favour of the new thread.

proposals/3967-device-signing-upload-uia.md
kegsay: I don't think it is a good idea to tie this to time-since-login because:
 - it introduces temporal variability which becomes difficult to debug. With the current form of the MSC, it's very clear that if you get an HTTP 401 back from this endpoint it's because you are trying to replace keys -> something went very wrong. Speaking from first hand experience, rageshakes often elide the response bodies so we rely heavily on the HTTP status code. If we introduce a time delay, it's possible that an HTTP 401 is due to this instead. We could maybe provide a different status code, but this breaks the general form of UIA which mandates 401.
 - it introduces configuration variability which becomes difficult to debug. It's not clear what the grace period should actually be. Anything less than 1 hour risks catching initial x-signing requests from users who are on particularly awful networks. However, even increasing this to 1 hour poses the risk that we incorrectly catch the initial upload (e.g you login on a bad GSM connection, then give up waiting for it to login and close the app, only reopening it the next day). This becomes difficult to debug in rageshakes, as we will just see HTTP 401s and won't know what the HS configuration is for the time delay. We already see this due to the use (or non-use) of `ui_auth.session_timeout`.

> they're suddenly able to take over the account via cross-signing verification

I would like you to explain in more detail the attack you're trying to protect against here. It is my understanding that uploading MSKs merely allows you to manipulate _some_ aspects of E2EE, it does _not_ allow account takeover as the device logout and login endpoints are protected via UIA.

> the condition appears to have been upgraded from SHOULD to MUST

I am reluctant to allow implementations to introduce variability which will inevitably become inconsistencies. I would much rather we were prescriptive in the spec, as that's kinda the whole point of a specification. That being said, I'm not going to push back strongly on this if this ends up being a blocker.

richvdh: I guess the threat model here is something like:

 * Room with sensitive content contains old user who never published cross-signing keys.
 * Attacker gets hold of access token for said user and publishes *their own* cross-singing keys
 * Attacker creates a device belonging to said user (how?) and signs it with the bad cross-signing key.
   * Alternatively, if the compromised access token is for an old device with no published device keys, they could use that.
 * Everyone in the room thinks the new device belongs to $old_user

But I don't think that's a viable attack path. (Mostly because, if people are actually relying on cross-signing to stop them leaking sensitive content, their client will have been complaining about the non-cross-signed user for a long time anyway. Also, if the attacker can create a new device, they can complete UIA anyway.)

clokep: At the very least the attack vector should be detailed in the MSC, even if the author(s) feel the impact is negligible.

turt2live: (I feel a bit more strongly about it being a blocker, ftr)

kegsay: > Users won't know they're talking to an attacker, and instead see the legitimate owner as the attacker instead, due to their device being explicitly unverified

If an attacker has pwned this user's access token then the attacker already wins because users won't know they're talking to an attacker. The owner hasn't uploaded x-signing keys for a long time so users will not expect the owner to start doing so. The attack you outline feels unlikely to be a real world concern. Is there another way this can go wrong?

I'll add this attack to the security considerations.

uhoreg: > If an attacker has pwned this user's access token then the attacker already wins because users won't know they're talking to an attacker.

That's not exactly true. If an attacker has only pwned the user's access token, they won't be able to send or read encrypted messages messages since they don't have the private keys. Even if they temporarily pwn a user's device, they will only be able to send/receive encrypted messages up to a limited time after they have control of the device, due to Olm/Megolm's ratcheting.

> The owner hasn't uploaded x-signing keys for a long time so users will not expect the owner to start doing so.

I don't think any clients currently warn of users who weren't cross-signed who recently became cross-signed. That said, I agree with Rich that people who are worried about security would probably have done something about a non-cross-signed user.   And also Kegan's point that people should be doing key verification.

I think it's worthwhile outlining the attacks in the Security Considerations section, but I don't think they're blockers.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -61,6 +61,37 @@ However, this degradation needs to be weighed against the typical real world
...
+
+Existing user accounts who do not already have cross-signing keys set up will
+now be able to upload keys without UIA. If such a user has their access token
+compromised, an attacker will be able to upload a `master_key`, `self_signing_key`
+and `user_signing_key`. This is a similar threat model to a malicious server admin
+replacing these keys in the homeserver database.
+
+This does not mean:
+ - the attacker can "take over the account". Device login/logout endpoints are
+   still protected via UIA.
```
turt2live, line 74: Linking to the endpoints in question here would be good. Device *deletion* is covered by UIA, but no forms of login use UIA. Additionally, *logout* is not protected by UIA.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -61,6 +61,37 @@ However, this degradation needs to be weighed against the typical real world
...
+ - the device will appear as verified. This requires out-of-band verification e.g
+   emoji comparison, which will correctly detect the MITM'd key.
```
turt2live, line 76: if the key is used to verify the device, surely it'd appear as verified? What level of verification is preventing this from being a concern?

proposals/3967-device-signing-upload-uia.md
```diff
@@ -61,6 +61,37 @@ However, this degradation needs to be weighed against the typical real world
...
+
+The main usability issue around this endpoint is requiring UIA, so it is critical
+that we only require UIA when absolutely necessary for the security of the account.
+In practice, this means requiring UIA when keys are _replaced_. There have been
+suggestions to reintroduce a grace period (e.g after initial device login) or just
+mandate it entirely for these old existing accounts. This would negatively impact
+usability because:
+ - it introduces temporal variability which becomes difficult to debug.
```
turt2live, line 84: This could do with more words. Why is it hard to debug? What is being debugged?

proposals/3967-device-signing-upload-uia.md
```diff
@@ -61,6 +61,37 @@ However, this degradation needs to be weighed against the typical real world
...
+ - it introduces configuration variability which becomes difficult to debug. It's not
+   clear what the grace period should actually be. Anything less than 1 hour risks
+   catching initial x-signing requests from users who are on particularly awful networks.
+   However, even increasing this to 1 hour poses the risk that we incorrectly catch the
+   initial upload (e.g the client logs in on a bad GSM connection, then give up waiting
+   for it to login and close the app, only reopening it the next day). This becomes
+   difficult to debug in bug reports, as they just report HTTP 401s and it is unknown what
+   the HS configuration is for the time delay. This is seen already due to the use (or non-use)
+   of `ui_auth.session_timeout`.
```
turt2live, line 93: I'm not sure this statement is true. 1 hour is indeed pretty constrained, but why is 1 day bad? Why is it difficult to debug?

The suggestion has never been to make this a configuration option either, as that indeed negatively harms user experience. If we were to pick a grace period, we'd write it into the spec. This should be clarified in the text.

turt2live (changes requested):
Thanks for adding the additional security consideration text. It appears as though some (most?) of the conversation context didn't make it into the MSC though - this is important detail that should be in the MSC's text. After an MSC is accepted we stop referencing the original PR for the most part, instead preferring to use the rendered Markdown on the main branch.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -0,0 +1,121 @@
...
+replacing these keys in the homeserver database.
+
+This does not mean:
+ - the attacker can "take over the account". It does not allow the attacker to
+   [login](https://spec.matrix.org/latest/client-server-api/#login) as they need to
+   know the password to the account. Likewise, an attacker cannot [logout all devices](https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3logoutall)
+   nor can they [logout specific devices](https://spec.matrix.org/latest/client-server-api/#delete_matrixclientv3devicesdeviceid)
+   as these also go through UIA prompts.
+ - the device will appear as verified to other users. Other users need to verify the
+   public key [out-of-band](https://spec.matrix.org/latest/client-server-api/#short-authentication-string-sas-verification).
```
turt2live, line 79: We make no guarantees that `latest` will continue to work as time goes on:

```suggestion
   [login](https://spec.matrix.org/v1.10/client-server-api/#login) as they need to
   know the password to the account. Likewise, an attacker cannot [logout all devices](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3logoutall)
   nor can they [logout specific devices](https://spec.matrix.org/v1.10/client-server-api/#delete_matrixclientv3devicesdeviceid)
   as these also go through UIA prompts.
 - the device will appear as verified to other users. Other users need to verify the
   public key [out-of-band](https://spec.matrix.org/v1.10/client-server-api/#short-authentication-string-sas-verification).
```

turt2live (approved):

proposals/3967-device-signing-upload-uia.md
turt2live: (considering this resolved as author - other SCT members are welcome to re-raise the concern)

turt2live:
@mscbot resolve Language appears to change security conditions

kegsay:
@turt2live I have added a caveat to the MSC around interactions with MSC3882. I don't think this materially changes things so I don't think it acts as a blocker, but I do want you to check it and agree/disagree before this lands.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -112,6 +112,25 @@ usability because:
 For these reasons, this MSC does not specify a grace period or treat some user accounts
 without existing cross-signing keys as special.
 
+#### Interactions with MSC3882
```
turt2live, line 115: This section feels out of place, and should be removed from the MSC in my opinion. It describes a process problem that's irrelevant to the subject of the MSC and an attack vector that's already discussed elsewhere in the proposal. 

The specific vector described by the section also appears to be heavily based on assumptions which shouldn't be picked apart here (even if we had time to do so).

turt2live (dismissed):

proposals/3967-device-signing-upload-uia.md
kegsay: Removed.

proposals/3967-device-signing-upload-uia.md
```diff
@@ -70,18 +70,31 @@ and `user_signing_key`. This is a similar threat model to a malicious server adm
...
 This does not mean:
- - the attacker can "take over the account". Device login/logout endpoints are
-   still protected via UIA.
- - the device will appear as verified. This requires out-of-band verification e.g
-   emoji comparison, which will correctly detect the MITM'd key.
+ - the attacker can "take over the account". It does not allow the attacker to
+   [login](https://spec.matrix.org/v1.10/client-server-api/#login) as they need to
+   know the password to the account. Likewise, an attacker cannot [logout all devices](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3logoutall)
+   nor can they [logout specific devices](https://spec.matrix.org/v1.10/client-server-api/#delete_matrixclientv3devicesdeviceid)
+   as these also go through UIA prompts. **See notes on MSC3882**.
```
richvdh, line 77: which notes on MSC3882?

```suggestion
   as these also go through UIA prompts.
```

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1828

turt2live:
Merged 🎉

PR #3968, opened 2023-02-18
MSC3968: Poorer features

progval:
[Rendered](https://github.com/progval/matrix-spec-proposals/blob/poorer-features/proposals/3968-poorer-features.md)

This MSC is meant primarily for bridges to signal to clients which "richer features" are not supported, allowing clients to hide or prioritize UI elements so their users do not unwittingly send events that would be bridged either badly or not at all.

This addresses the largest part of problems listed by https://github.com/matrix-org/matrix-spec/issues/836 that cannot be solved with power levels. MSC3969 will address the remaining (size limits)

### DIFF
--- a/proposals/3968-poorer-features.md
+++ b/proposals/3968-poorer-features.md
@@ -105,7 +105,7 @@ as of 2023-02 (ie. outside RFC1459 and sending files), that is:
             "a": 0,
             "b": 0,
             "code": 0,
-            "dive": 0,
+            "div": 0,
             "font": 0,
             "p": 0,
             "pre": 0,
--- a/proposals/3968-poorer-features.md
+++ b/proposals/3968-poorer-features.md
@@ -1,0 +1,233 @@
+# MSC3968: Poorer features
+
+Matrix owes its name to its aim to interoperate with many protocols, by being supporting a superset
+of their individual features. While this is great to bridge events from other protocols to Matrix,
+this causes friction in the other direction, as events need to be either converted to a fallback
+(that may not be natural to the target platform) or silently dropped altogether.
+
+Additionally, room administrators may also prefer to forbid some features in order to ease moderation,
+cater to specific needs of their community, or provide a different experience of Matrix.
+
+This MSC aims to provide a way for bridges and room administrators to signal which features senders
+should avoid using. This allows clients to hide elements from their UI, provide an error or warning
+early to the user, or themselves fallback using any extra context they may have.
+
+
+## Proposal
+
+This MSC introduces a new `m.room.event_features` state event type, which associates a desirability
+level to:
+
+- attachment mimetypes for `m.file`, `m.audio`, `m.video`, ... events;
+- `msgtype` values
+- (for clients implementing [MSC1767][MSC1767]) to mimetypes used in extensible events;
+- keys in `"content"` objects;
+- HTML elements
+
+Mimetypes, msgtype values, keys, and HTML element names are jointly referred to as "entities" below.
+
+
+### Desirability level
+
+This level is an integer, whose values are defined as:
+
+- under -100: this entity is forbidden, and must not be sent in any case.
+  Clients receiving events containing them should hide them from their users
+  (or display these events without as if they didn't contain the entity)
+- from -100 to -1: this entity is discouraged, and should not be sent unless there is
+  no alternative.
+  Clients receiving these events may hide them from their users
+  (or display these events without as if they didn't contain the entity)
+- 0 to 100: this entity is acceptable
+- over 100: currently undefined, and should be interpreted as 100
+
+
+Unlike `m.room.power_levels`, `m.room.event_features` acts on event content, so it is enforced
+primarily by receiving clients.
+
+Servers may apply restrictions before sending unencrypted events, and return `M_FORBIDDEN`
+in case the event contains an entity with a level under 0.
+Receiving servers may hide such events from their clients.
+
+
+### Format of `m.room.event_features`
+
+`m.room.event_features` events may have these keys:
+
+- `attachment_mimetypes` (object with integer values): This is a mapping from file mimetypes
+  to desirability level
+- `attachment_mimetypes_default` (integer): the default desirability level for file mimetypes
+  Can be overridden by the `content_mimetypes` key. Defaults to 0 if unspecified.
+- `content_mimetypes` (object with integer values): This is a mapping from content mimetype
+  (as per MSC1767) to desirability level
+- `content_mimetypes_default` (integer): the default desirability level for content mimetypes
+  (as per MSC1767). `*` may be used as a wildcard.
+  Can be overridden by the `content_mimetypes` key. Defaults to 0 if unspecified.
+- `keys` (object with integer values): This is a mapping from key name to desirability level
+- `keys_default` (integer): the default desirability level for keys.
+  Can be overridden by the `keys` key. Defaults to 0 if unspecified.
+- `html_elements` (object with integer values): This is a mapping from HTML element name to desirability level
+- `html_elements_default` (integer): the default desirability level for HTML elements.
+  Can be overridden by the `html_elements` key. Defaults to 0 if unspecified.
+- `msgtypes` (object with integer values): This is a mapping from `msgtype` of `m.room.message`
+  events to desirability level
+- `msgtypes_default` (integer): the default desirability level for values of `msgtype`
+  in `m.room.message` events.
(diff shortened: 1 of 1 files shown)


progval:
Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

proposals/3968-poorer-features.md
```diff
@@ -0,0 +1,233 @@
...
+            "file": -100,
+            "thumbnail_file": -100
+        },
+        // discourage HTML elements with no counterpart on IRC:
+        "html_elements_default": -1,
+        "html_elements": {
+            "a": 0,
+            "b": 0,
+            "code": 0,
+            "dive": 0,
```
uhoreg, line 108:
```suggestion
            "div": 0,
```

presumably

proposals/3968-poorer-features.md
```diff
@@ -0,0 +1,248 @@
...
+
+Additionally, room administrators may also prefer to forbid some features in order to ease moderation,
+cater to specific needs of their community, or provide a different experience of Matrix.
+
+This MSC aims to provide a way for bridges and room administrators to signal which features senders
+should avoid using. This allows clients to hide elements from their UI, provide an error or warning
+early to the user, or themselves fallback using any extra context they may have.
+
+
+## Proposal
```
sumnerevans, line 16: One of the biggest features that needs restricted on bridges is the set of allowed reactions. For example, iMessage only allows 5 reactions. I would be interested to see if this proposal could be adapted to handle that use-case.

progval: Hmm, this should probably be its own MSC, because it sounds like it's going to be its own can of worms (a whitelist like this, blacklist, only allowing emojis, unicode versions, ...).

proposals/3968-poorer-features.md
```diff
@@ -0,0 +1,248 @@
+# MSC3968: Poorer features
```
Half-Shot, line 1: At the risk of bikeshedding too much, I'm wondering if we should call this "reduced featureset" modes or something? I'm just wondering if "poorer" is a good multi-lingual phrase.

Anyway, my 2c.

bleonard252: Maybe "feature availability preferences," although that uses some long words

proposals/3968-poorer-features.md
```diff
@@ -0,0 +1,248 @@
...
+in case the event contains an entity with a level under 0.
+Receiving servers may hide such events from their clients.
+
+If more than one `m.room.event_features` state events are present in a room, clients should
+consider the lowest value of each of the desirability levels they define.
+
+
+### Format of `m.room.event_features`
+
+`m.room.event_features` events may have these keys:
```
cloudrac3r, line 58: Could this proposal be expanded to consider read receipts?

It is desirable for some rooms to discourage sending and displaying read receipts, such as rooms bridged to platforms that don't support read receipts, or for announcement rooms with thousands of members where each member doesn't need to know if other members have read a particular message.

proposals/3968-poorer-features.md
```diff
@@ -0,0 +1,248 @@
...
+## Alternatives
+
+- giving a name to every feature (or use MSC number) and attribute a desirability level to each
+  of these names. While this makes `m.room.event_features` more readable from developer tools,
+  it adds an extra burdens on clients and servers to keep track of these names and map them to
+  entities anyway
+- using `"forbidden"`, `"discouraged"`, and `"acceptable"` as levels instead of integers.
+  This prevents finer grained display rules (eg. receiving clients may choose to hide events matching
+  levels under -50, because they are "very discouraged" by the administrator) and future
+  extendability (eg. values over 100 meaning an entity is required).
```
cloudrac3r, line 204: It makes more sense to me to use these names instead of integers. If level 0 is considered acceptable, what would a level of e.g. +37 even mean?

With integers, clients would end up choosing arbitrary number thresholds at which to activate or deactivate a certain feature in the interface. Non-human clients would also end up with arbitrary number thresholds programmed into them at which they'd change their behaviour. Among other problems, this would lead to inconsistency between clients when they each choose different numbers.

I think it makes more sense to directly specify the semantic meaning and intent, so I highly support using labels like "discouraged".

cloudrac3r:
Hi @progval , if you closed this because you are no longer interested in seeing this MSC to completion, could I try to continue it?

progval:
I closed it because Element is blocking implementation so I don't think it's going to go anywhere, but sure go ahead

ara4n:
to be clear, Element isn't blocking any implementation, and this is a desirable MSC from my pov, and i'd suggest reopening it.

progval:
@ara4n These PRs have been stuck in review by Element employees for five months while accumulating merge conflicts: https://github.com/matrix-org/matrix-appservice-bridge/pull/463 https://github.com/matrix-org/matrix-appservice-irc/pull/1683 https://github.com/matrix-org/matrix-appservice-irc/pull/1686

joshsimmons:
@progval thank you so much for your work on this, this is an important proposal that would make all manner of bridge implementations better – and I'm sorry this has been hanging.

I was glad to sit in on this month's Spec Core Team retro and we took some time to discuss this. There's broad agreement that we'd love to see this MSC through. Part of the discussion was acknowledgment that Element hasn't allocated the resources to keep this moving, and that we may need other contributors to step in. It's not yet clear where that would come from, but everyone is aligned: this is valuable and we'd love to see this re-opened so that potential contributors see the opportunity.

I'm sorry I can't offer concrete resolution, but did want to at least make sure you know this is something we value and want to see land in the spec even if it takes longer than we'd like.

Mikaela:
> we may need other contributors to step in.

Could you clarify how other contributors can step in and merge their own pull requests? Is this documented somewhere, I think this would be important information to document.

uhoreg:
> > we may need other contributors to step in.
> 
> Could you clarify how other contributors can step in and merge their own pull requests? Is this documented somewhere, I think this would be important information to document.

In general, from a completely spec standpoint, PRs don't need to be merged to count as implementation for MSCs.  In addition, this could be implemented in other bridges.  Though I recognize that it can be demotivating to have a PR sitting unreviewed for a long time, and it can be a pain to re-implement something for a different bridge.  However, since the spec process doesn't require PRs to be merged to count as an implementation, nothing on the spec side is being blocked by the PRs not being merged.

Skimming through the MSC, it looks like it's pretty much ready for review, so the main things that would be needed here are a client implementation (showing how a client would behave with an `event_features` event), and then letting us know in the Spec Office room that it's ready for FCP.

progval:
> the main things that would be needed here are a client implementation

I was ready to start working on Eleweb as soon as my AS-irc PRs were accepted. But if it's also never merged in Eleweb, then I don't see the point.

classabbyamp:
maybe it would be good to document requirements like this in CONTRIBUTING.md, so that contributors aren't left guessing for months

uhoreg:
> maybe it would be good to document requirements like this in CONTRIBUTING.md, so that contributors aren't left guessing for months

The proposal process is documented at https://spec.matrix.org/proposals/ and the implementation requirement specifically is at https://spec.matrix.org/proposals/#implementing-a-proposal  The problem isn't that the requirements are undocumented, because there is documentation.  But the proposal process documentation is quite long.  But in general, if contributors are wondering what the next step is for their MSC, the best thing to do is to ask in the #sct-office:matrix.org room.

uhoreg:
> > the main things that would be needed here are a client implementation
> 
> I was ready to start working on Eleweb as soon as my AS-irc PRs were accepted. But if it's also never merged in Eleweb, then I don't see the point.

So from a purely spec perspective, having such a PR would help this MSC get accepted.  But in this particular case, I suspect that you care less about what the spec says, and you care more about the implementations getting merged so that Element users stop making a mess of IRC rooms.  So I can understand the frustration of not having the PRs merged, and I can understand that having such a long delay can be demotivating in doing any further work in this topic.

I can't say much about the merging of the Application Service PRs, because I don't have anything to do with that.  But since this issue is in a Spec repository, I can reiterate what others have said, that from the Foundation's point of view, we would like to see something like this in the Spec.  And if @progval doesn't want to continue working on it, maybe someone else can push it forward.  As I said earlier, it looks like we would need an implementation in some client, and then asking for FCP to be called.  (Though if this MSC remains closed, then someone would have to open a new MSC.)

cloudrac3r:
I am writing a new proposal based on the review feedback here and I will link it here when I create the PR.

cloudrac3r:
I have created [MSC4110 Fewer Features](https://github.com/matrix-org/matrix-spec-proposals/pull/4110).

PR #2346, opened 2019-11-05
[WIP] MSC2346: Bridge information state event

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/hs/msc-bridge-inf/proposals/2346-bridge-info-state-event.md)

PR #3969, opened 2023-02-18
MSC3969: Size limits

progval:
[Rendered](https://github.com/progval/matrix-spec-proposals/blob/size-limits/proposals/3969-size-limits.md)

This MSC is meant to signal size limits of platforms bridged to a room to clients in the room, allowing them to display warnings and/or prevent their users from unwittingly sending events that would be bridged either badly or not at all.

This addresses some of the issues mentioned in https://github.com/matrix-org/matrix-spec/issues/836 ; while the rest can be addressed with power levels and MSC3968.

### DIFF
--- a/proposals/3969-size-limits.md
+++ b/proposals/3969-size-limits.md
@@ -68,8 +68,8 @@ Values not matching these types and undefined keys should be ignored.
 
 RFC1459 IRC messages are limited to 512 bytes, including message metadata overhead, whose
 length is per-user.
-IRC bridges may therefore consider the maximum possible size of the overhead, and substract
-it from the total limit, and advertize this in `m.room.size_limits`.
+IRC bridges may therefore consider the maximum possible size of the overhead, and subtract
+it from the total limit, and advertise this in `m.room.size_limits`.
 
 Sending long messages (over 3-5 lines) is usually considered poor style, but pastebining long ones
 occasionally is tolerated.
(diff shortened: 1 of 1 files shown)


progval:
Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

HarHarLinks:
Please fix the Rendered link in your OP:
```diff
- https://github.com/progval/matrix-spec-proposals/blob/size-limits/proposals/proposals/3969-size-limits.md
+ https://github.com/progval/matrix-spec-proposals/blob/size-limits/proposals/3969-size-limits.md
```

julianfoad:
@progval Have you seen the proposal [Bridge limitation warnings](https://github.com/matrix-org/matrix-react-sdk/pull/4642) and could you add to your Alternatives section a comparison with that?

progval:
I hadn't. The two main differences are that it counts by character, which I mention in the alternatives (and is also not great for IRC), and uses the bridge's state key instead of an empty state key.

Using the bridge's key is probably better as it allows stacking limitations of multiple bridges, in addition to explaining to users which bridge is causing the limit. I'll update the MSC to do the same unless someone objects.

progval:
@Half-Shot Do you have a spec of the bridge limitation format and/or of `uk.half-shot.bridge` somewhere, so I can point to it?

clokep:
> @Half-Shot Do you have a spec of the bridge limitation format and/or of `uk.half-shot.bridge` somewhere, so I can point to it?

It seems to be [MSC2346](https://github.com/matrix-org/matrix-spec-proposals/pull/2346).

PR #3861, opened 2022-08-04
MSC3861: Next-generation auth for Matrix, based on OAuth 2.0/OIDC

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3861-next-generation-auth.md)

Core sub-proposals:

- https://github.com/matrix-org/matrix-spec-proposals/pull/2964
- https://github.com/matrix-org/matrix-spec-proposals/pull/2965
- https://github.com/matrix-org/matrix-spec-proposals/pull/2966
- https://github.com/matrix-org/matrix-spec-proposals/pull/2967
- https://github.com/matrix-org/matrix-spec-proposals/pull/4254

Additional proposals:
- https://github.com/matrix-org/matrix-spec-proposals/pull/3824
- https://github.com/matrix-org/matrix-spec-proposals/pull/4108
- https://github.com/matrix-org/matrix-spec-proposals/pull/4190
- https://github.com/matrix-org/matrix-spec-proposals/pull/4191
- https://github.com/matrix-org/matrix-spec-proposals/pull/4198

----

SCT stuff:

[checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/3861#issuecomment-2603442893)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3861#issuecomment-2614112375)

PR #3970, opened 2023-02-23
MSC3970: Scope transaction IDs to devices

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/device-scope-txnid/proposals/3970-device-scope-txnid.md)

For implementations we have:

- Conduit: https://gitlab.com/famedly/conduit/
- Synapse: https://github.com/matrix-org/synapse/pull/15318

Complement tests: https://github.com/matrix-org/complement/pull/632

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3970#issuecomment-1478390006)

### DIFF
--- a/proposals/3970-device-scope-txnid.md
+++ b/proposals/3970-device-scope-txnid.md
@@ -61,16 +61,15 @@ with it, this proposal would widen the scope of the transaction ID.
 Therefore it could potentially lead to a request being treated as "new" where
 before it would have been identified as a retransmission and deduplicated.
 
-However, the evidence suggests that nothing would be impacted in reality.
+However, some evidence has been collated which suggests that nothing would be impacted
+in reality:
 
-Some evidences have been collated to support this claim:
+#### 1. Data from the matrix.org homeserver suggests the change would have no impact
 
-#### 1. Data from the matrix.org Homeserver suggests the change would have no impact
+The `matrix.org` homeserver is a reasonable size deployment and could be considered
+reasonably representative of the diversity of Matrix clients.
 
-The matrix.org Homeserver is a reasonable size deployment and could be considered
-reasonably representative of the diversity of Matrix client.
-
-The Synapse homeserver that runs matrix.org maintains a `event_txn_id` table
+Synapse maintains a `event_txn_id` table
 that contains a rolling 24 hour window of
 (`user_id`, `token_id`, `room_id`, `txn_id`) tuples.
 
@@ -80,7 +79,7 @@ transaction IDs for a given user, token and room.
 n.b. not all `PUT` endpoints contribute to the table but I think the high-volume
 ones do
 
-As such, the widening of the scope from token to device should not have caused
+This suggests that the widening of the scope from token to device would not have caused
 any issues during the periods sampled.
 
 For reference the following is the schema of the `event_txn_id` table:
--- a/proposals/3970-device-scope-txnid.md
+++ b/proposals/3970-device-scope-txnid.md
@@ -1,22 +1,28 @@
 # MSC3970: Scope transaction IDs to devices
 
-Transaction identifiers in the Client-Server API are currently scoped to the concept of a "client session" which, when
-refresh tokens are used, can span a sequence of access tokens.
+Transaction identifiers in the Client-Server API are currently scoped to the
+concept of a "client session" which, when refresh tokens are used, can span a
+sequence of access tokens.
 
 The spec [reads](https://spec.matrix.org/v1.6/client-server-api/#transaction-identifiers):
 
-> The scope of a transaction ID is a “client session”, where that session is identified by a particular access token.
-> When refreshing an access token, the transaction ID’s scope is retained. This means that if a client with token `A`
-> uses `TXN1` as their transaction ID, refreshes the token to `B`, and uses `TXN1` again it’ll be assumed to be a
-> duplicate request and ignored. If the client logs out and back in between the `A` and `B` tokens, `TXN1` could be used
+> The scope of a transaction ID is a “client session”, where that session is
+> identified by a particular access token. When refreshing an access token, the
+> transaction ID’s scope is retained. This means that if a client with token `A`
+> uses `TXN1` as their transaction ID, refreshes the token to `B`, and uses
+> `TXN1` again it’ll be assumed to be a duplicate request and ignored. If the
+> client logs out and back in between the `A` and `B` tokens, `TXN1` could be used
 > once for each.
 
 The transaction IDs appear in two parts of the Client-Server API spec:
 
-1. As a identifier to allow the homeserver to make some `PUT` endpoints [idempotent](https://spec.matrix.org/v1.6/client-server-api/#transaction-identifiers)
-2. An unsigned field in the event data model for a client to tell if it sent an event or not. a.k.a. solving the ["local echo"](https://spec.matrix.org/v1.6/client-server-api/#local-echo) problem
+1. As a identifier to allow the homeserver to make some `PUT` endpoints
+[idempotent](https://spec.matrix.org/v1.6/client-server-api/#transaction-identifiers)
+2. An unsigned field in the event data model for a client to tell if it sent an
+event or not. a.k.a. solving the
+["local echo"](https://spec.matrix.org/v1.6/client-server-api/#local-echo) problem
 
-For reference, the `PUT` endpoitns that have the a `{txnId}` param are:
+For reference, the `PUT` endpoints that have the a `{txnId}` param are:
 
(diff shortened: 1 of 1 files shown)


proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,198 @@
...
+choose transaction identifiers that are unique to that client session.
+
+## Alternatives
+
+### Do nothing
+
+We could leave the transaction ID scope as is.
+
+However, it makes it difficult to implement a features like [MSC3861: Matrix architecture change to delegate authentication via OIDC](https://github.com/matrix-org/matrix-spec-proposals/pull/3861)
+as the concept of a "client session" doesn't really exist in OIDC.
```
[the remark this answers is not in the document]
hughns: There are two parts to this in the [OIDC proposal](https://github.com/matrix-org/matrix-spec-proposals/pull/3861):

1. The client generates their own device ID and specifies it in the OIDC scope that is requested at the point of sign in. This scope described in [MSC2967](https://github.com/sandhose/matrix-doc/blob/msc/sandhose/api-scopes/proposals/2967-api-scopes.md#device-id-handling)
2. When the client signs out the OpenID Provider is responsible for informing the Homeserver that the device should be deleted. This process is described in [MSC3964](https://github.com/sandhose/matrix-doc/blob/msc/sandhose/oauth2-profile/proposals/2964-oauth2-profile.md#logging-out)

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Potential issues
+
+### This is technically a breaking change to the spec
+
+The main "issue" I see with this proposal is that this is technically a breaking
+change to the spec.
+
+Because a device ID could have multiple sequences of access tokens associated
+with it, this proposal would widen the scope of the transaction ID.
```
richvdh, line 59: I'm somewhat struggling to follow this explanation. I think an example of how it might happen in practice is illustrative:

```suggestion
A device ID could have multiple sequences of access tokens associated
with it (since device ID can be specified as a parameter to 
[`/login`](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3login)). 

For example, a "bot" implementation might masquerade as the same "device" despite
calling `/login` on every run by passing the same device ID. Such a device might
also use the same transaction ID on each run.
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+Therefore it could potentially lead to a request being treated as "new" where
+before it would have been identified as a retransmission and deduplicated.
```
richvdh, line 62: the opposite, no?

```suggestion
Therefore it could potentially lead to a request being treated as a duplicate
where previously it would have been treated as a distinct request.
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+However, the evidence suggests that nothing would be impacted in reality.
+
+Some evidences have been collated to support this claim:
```
richvdh, line 66:
```suggestion
However, some evidence has been collated which suggests that nothing would be impacted
in reality:
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### 1. Data from the matrix.org Homeserver suggests the change would have no impact
```
richvdh, line 68:
```suggestion
#### 1. Data from the matrix.org homeserver suggests the change would have no impact
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+The matrix.org Homeserver is a reasonable size deployment and could be considered
+reasonably representative of the diversity of Matrix client.
```
richvdh, line 71:
```suggestion
The `matrix.org` homeserver is a reasonable size deployment and could be considered
reasonably representative of the diversity of Matrix clients.
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+The Synapse homeserver that runs matrix.org maintains a `event_txn_id` table
+that contains a rolling 24 hour window of
+(`user_id`, `token_id`, `room_id`, `txn_id`) tuples.
```
richvdh, line 75:
```suggestion
Synapse maintains a `event_txn_id` table
that contains a rolling 24 hour window of
(`user_id`, `token_id`, `room_id`, `txn_id`) tuples.
```

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,210 @@
...
+
+Having analysed the contents of the table, it appears that there are no repeated
+transaction IDs for a given user, token and room.
+
+n.b. not all `PUT` endpoints contribute to the table but I think the high-volume
+ones do
+
+As such, the widening of the scope from token to device should not have caused
+any issues during the periods sampled.
```
richvdh, line 84:
```suggestion
This suggests that the widening of the scope from token to device would not have caused
any issues during the periods sampled.
```

richvdh (commented):
A few suggestions for clarification

proposals/3970-device-scope-txnid.md
hughns: You are quite right 👍

turt2live:
@matrix-org/spec-core-team : the author is seeking feedback which relates to the overarching mission of OIDC-for-Matrix. Please provide some thoughts on whether this approach is workable before we get too far along in the OIDC transition to fix it :)

For the author's own words on this request, see [this message](https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$4E9SUEMKTovQ8yEuPx5zUIFIkgZ4qzfIC4wKgewdTHk?via=matrix.org&via=element.io&via=envs.net).

richvdh:
removing needs-implementation label, as conduit's implementation is cited. I don't think the Synapse one does (yet), because there are quite a few comments which look like they need resolving.

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,214 @@
...
+
+- (`device_id`, `HTTP path of request which includes the transaction ID`)
+
+For local echo, the homeserver would now include the `transaction_id` in the
+event data when it is serving a sync request from the same `device_id` as
+determined from the access token.
+
+## Potential issues
+
+### This is technically a breaking change to the spec
```
richvdh, line 53: @turt2live I think you are the most likely person on the SCT to have concerns over this approach - could you have a look?

richvdh:
This seems to make a lot of sense, and basically seems to be ready for FCP.

Accordingly:

@mscbot fcp merge

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Clarification on transaction ID time scope
+
+I also suggest that the spec be clarified over what time periods the transaction
+ID is scoped for such that clients can be aware. This cloud simply be to say
+that the time period is not defined and so may vary by implementation.
+
+### Suggest a lease naive transaction ID format
+
+I think we should sense to update the recommendation on the format of
```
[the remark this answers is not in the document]
hughns: @uhoreg you are not the only one. I attempted to re-read this morning and couldn't make sense of it either!

I've reworked it into something that hopefully reads better.

proposals/3970-device-scope-txnid.md
```diff
@@ -0,0 +1,222 @@
+# MSC3970: Scope transaction IDs to devices
```
babolivier, line 1: As someone with very little context on what triggered this MSC to be proposed, I find it quite difficult to figure out the problem it's trying to solve just from reading the MSC's introduction. The "Do nothing" part of the Alternatives section has a bit to say about that, but I think it would be helpful to also have this information here, so by the time the reader gets to the proposal it's clear what problem it's trying to solve.

hughns: I've had a re-read and I agree. I've added some more info in the intro section to try and make it clearer sooner: [`6fc4975` (#3970)](https://github.com/matrix-org/matrix-spec-proposals/pull/3970/commits/6fc4975c52ec3ed72b9234aa01c81551850c840d)

babolivier: This is very helpful, thanks!

richvdh:
@hughns mentioned he hoped to get a chance to write the spec for this in the first half of May.

richvdh:
I'm going to have a crack at writing the spec for this

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1526

richvdh:
... and merged.

hughns:
Thank you ⭐️

PR #3971, opened 2023-02-25
Sharing image packs

AndrewRyanChama:
create a proposal for sharing image packs.


proposals/3971-sharing-image-packs.md
```diff
@@ -0,0 +1,75 @@
...
+When the image pack is part of a public room which supports guest access, then the client must include
+the pack url in the event when sending an image from that pack.  If the room is private or does not
+have guest access, then the client must not include the pack url.  Of course, when sending an emoji or
+sticker that is not part of a pack, the client does not have to include a pack url either.
+
+The pack url is included in the following way:
+
+1. For stickers, the pack url is included as an additional string field in the event content ```pack_url```
+2. For emojis, the emoji should be sent as an html image inside a hyperlink.  An example representation is:
+```[<img data-mx-emoticon height="18" src="mxc://matrix.org/cqiMCZgfXMWgDQqqFYnhjFPk" title=":abc:" alt=":abc:">](https://matrix.to/#/!hvgkwxdxoAwOmJvvUH:matrix.org/$_KxiZYeOMJDthWxj3m_lRiW_twVmOMJo_afvRRmnkCM)```
```
deepbluev7, line 46: I would probably prefer to use a data-mx-pack-url parameter on the img tag. That way only clients, that support custom emoji would render it as a link and they can know, that this is intended to share a custom emoji set. Otherwise users clicking the link might be confused and you can't really use emoji inside of link descriptions anymore.

proposals/3971-sharing-image-packs.md
```diff
@@ -0,0 +1,75 @@
...
+might see the message "this pack can't be previewed: Add Pack" and still get the option to add the pack in one click.
+
+## Security considerations
+
+There are no major security implications of this change.  Clients may only share room links of public rooms
+with guest access.
+
+## Unstable prefix
+
+N/A.
```
deepbluev7, line 70: You probably want to specify some prefix for the pack_url field, so that you don't reserve the name, while this MSC is still unmerged. It would add some more overhead to each event, but I think it is nicer for the ecosystem to namespace independent MSCs. :)

(Once that exists, I would probably start implementing it)

proposals/3971-sharing-image-packs.md
```diff
@@ -0,0 +1,75 @@
...
+url is the "pack url".  The room id must be used, not the room alias, as the alias may change.
+
+If a client not supporting this MSC clicks this link, then the deeplink will simply point to the state
+event, if that client chooses to display it.
+
+### Stickers and emojis
+
+When the image pack is part of a public room which supports guest access, then the client must include
+the pack url in the event when sending an image from that pack.  If the room is private or does not
+have guest access, then the client must not include the pack url.  Of course, when sending an emoji or
```
deepbluev7, line 39: I think I would prefer this to be a bit more lenient. A user should be able to disable this, but also I would like it, if a client is allowed to share a link to a pack, as long as every recipient would be able to access the pack. For example in a closed community, if there is a room with sticker packs, that any of the community members could join using "restricted joins" or "knocking", then sharing a link to the pack would still be nice. For invite only rooms it of course never makes sense.

proposals/3971-sharing-image-packs.md
```diff
@@ -0,0 +1,75 @@
...
+on third party sites.
+
+In this MSC, we propose a way to provide the following:
+
+1. Sharable url link to the image pack.
+2. Clickable emojis and stickers.
+
+With the following requirements:
+
+1. Links work inside and outside Matrix clients.
```
deepbluev7, line 20: What's the use case for having this work outside of Matrix clients?

AndrewRyanChama: The links should be sharable outside of Matrix, such as over other social media.  This can allow users to post lists and collections of image packs on those platforms.

I'll update the doc to say that links should be sharable outside of Matrix clients - the link itself will still lead to opening Matrix.

deepbluev7: That means I should be able to share a link outside matrix, but I don't need the embedding in the message to be accessible outside matrix, right? I.e. it doesn't need to be a link embedded into the img tag for that to work, but a data attribute fulfills that in the same way, since the attribute would still be a matrix.to link?

This means you probably want to add some metadata to the matrix.to link, so that it can be distinguished from a normal event link, I guess? I.e. appending a ?image_pack or so?

proposals/3971-sharing-image-packs.md
AndrewRyanChama: I figure that for closed communities, the community organizers already have a way to distribute their image packs to their members, so the utility is a little bit less.  The hope of this MSC is to allow public image packs to "go viral" and spread far beyond their original source.

Still, it makes sense that it doesn't hurt to allow these links.  In this case, I will leave it open as an optional feature.

proposals/3971-sharing-image-packs.md
AndrewRyanChama: Right, basically I just mean we want a matrix.to link, rather than some bespoke link that can only be clicked within a matrix client.  We don't need to expose any new data to the outside world.

AndrewRyanChama: Adding a query param won't automatically be passed through matrix.to to the matrix client.  This might be a useful expansion, but it adds a lot of complexity and I think it's functional as-is.

The person sharing the matrix.to link should be responsible for describing what it is - already a matrix.to event link doesn't give the room name or any other info, so sharers already have to do this.

deepbluev7: You probably want clients as well as intermediates (like the matrix.to page) to know what the link is for before or without resolving it to the actual event. Otherwise matrix.to will just show "you were invited to room X". You probably want to be able to show, that you were linked to a sticker pack instead, which needs additional metadata in the link. Similarly you probably want clients to suggest you to add the sticker pack instead of scrolling to the event. This means again, you want extra metadata. In general such parameters are passed to clients, that can handle matrix.to, but possibly it would need to be supported by the matrix: scheme as well.

You don't really want clients to have to do some deep inspection of every matrix.to target to be able to use it and you don't want to have users paste the link into some field, but in the best case have it be automatic when clicking a link, that the client does the right thing. It is less complex to just add metadata to the matrix.to link than have every client do deep inspection of the link target before prompting the user what to do. That is why matrix.to already distinguishes between room, user and event links and why the matrix URI format has a specific action parameter, describing the purpose of a link. This is handled before any request is sent to a server, so that the user can decide what to do.

proposals/3971-sharing-image-packs.md
deepbluev7: (resolved, but I can't do that myself)

proposals/3971-sharing-image-packs.md
deepbluev7: (resolved, but I can't do that myself)

proposals/3971-sharing-image-packs.md
deepbluev7: (resolved, but I can't do that myself)

deepbluev7:
(Small note, as far as I am aware, force pushes are generally frowned upon for MSCs, because it makes it harder to review changes to the MSC. Afaik it is suggested to just push new commits and they are squashed, once the MSC is merged)

proposals/3971-sharing-image-packs.md
AndrewRyanChama: Hmmm, action= seems like the right thing to do here.  Unfortunately, matrix.to and also Element don't honor this value.  You can see here that https://github.com/matrix-org/matrix.to/blob/main/src/open/clients/Nheko.js#L49 that action is harcoded depending on the resource type.  I think the only query params that work here are the ones on the matrix.to homepage.  Matrix.to always shows the same prompt "choose app to continue" regardless of the room type.

I think adding an additional metadata to the url does make sense, but it needs a bunch of additional things to move at the same time.  I'll document a bit more about what is wanted here though.

deepbluev7: You can just add an unrecognized action. Clients that want to use that metadata can then do that, but if they don't recognize it, they should just ignore it and handle it like normal events, which I would say is fine :)

AndrewRyanChama:
Dropping my take on the implementation here: https://github.com/SchildiChat/matrix-react-sdk/pull/15

PR #3008, opened 2021-02-15
MSC3008: Scoped access for widgets

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/widgets-scoped/proposals/3008-widgets-scoped-access.md)

PR #3973, opened 2023-03-01
MSC3973: Search users in the user directory with the Widget API

dhenneke:
[Rendered](https://github.com/nordeck/matrix-spec-proposals/blob/nic/feat/widgetapi-user-directory-search/proposals/3973-widgetapi-user-directory-search.md)

### DIFF
--- a/proposals/3973-widgetapi-user-directory-search.md
+++ b/proposals/3973-widgetapi-user-directory-search.md
@@ -1,0 +1,90 @@
+# MSC3973: Search users in the user directory with the Widget API
+
+[MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762) specifies a Widget API that
+is able to receive events from and write events to the room that is loaded in the client. This includes
+the support to invite users into a room by creating membership events. However, in order for this to
+work properly, the widgets needs to be able to discover users from the user directory of the server.
+That allows them to show the user a custom UI which might be useful in different applications such
+as a meeting planner.
+
+This proposal aims to bring the functionality of searching in the user directory into the widget
+specification. It should provide the same features that the
+[“User Directory search” endpoint](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3user_directorysearch)
+of the Client-Server API provides. It should further provide the same results as the invite dialog of
+the hosting client (e.g. Element) so the same search terms lead to the same results in all components.
+
+## Proposal
+
+We will add a new interface to the widget API to search users in the user directory. We will introduce
+a new capability that the user must manually approve:
+
+- `m.user_directory_search`: Let the widget access the user directory.
+
+To trigger the action, widgets will use a new `fromWidget` request with the action
+`user_directory_search` which takes the following shape:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "user_directory_search",
+  "data": {
+    "search_term": "foo",
+    "limit": 10
+  }
+}
+```
+
+Under `data`, all keys are a mirrored representation of the original `/_matrix/client/v3/user_directory/search`
+API. The `limit` field is optional and defaults to whatever the homeserver implementation uses.
+
+If the widget did not get approved for the capability required to send the event, the client MUST
+send an error response (as required currently by the capabilities system for widgets).
+
+The client SHOULD NOT modify the data of the request. The widget is responsible for producing valid
+events - the client MUST pass through any errors, such as permission errors, to the widget using the
+standard error response in the Widget API.
+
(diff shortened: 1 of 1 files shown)


proposals/3973-widgetapi-user-directory-search.md
```diff
@@ -0,0 +1,90 @@
...
+the support to invite users into a room by creating membership events. However, in order for this to
+work properly, the widgets needs to be able to discover users from the user directory of the server.
+That allows them to show the user a custom UI which might be useful in different applications such
+as a meeting planner.
+
+This proposal aims to bring the functionality of searching in the user directory into the widget
+specification. It should provide the same features that the
+[“User Directory search” endpoint](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3user_directorysearch)
+of the Client-Server API provides. It should further provide the same results as the invite dialog of
+the hosting client (e.g. Element) so the same search terms lead to the same results in all components.
```
dhenneke, line 14: There are two options to solve this:

1. Just return the response from the CS-API.
2. Return whatever the client is doing to provide the user list in their invite dialogs (and injecting them transparently into the results list without the widget noticing).

For 2., Element does many additional things that is currently implemented in the [`InviteDialog`](https://github.com/matrix-org/matrix-react-sdk/blob/e5291c195d3259b5905d72dc2945ed740fffc6d1/src/components/views/dialogs/InviteDialog.tsx). If we would want to provide the client-enhanced list instead of the raw server response, this would require refactoring in the react-sdk. But for the long term (if [MSC3008](https://github.com/matrix-org/matrix-spec-proposals/pull/3008) becomes a reality) it might be questionable how useful that would be. It could for example be an option to move the code to the js-sdk and expecting that future widgets will keep using that one. If the MSC is formulated openly enough, it could also leave other client authors the option to not do any additional processing here.

toger5: I agree with the thought, that the widget api should not be unnecassarly complex to implement for other clients. I think the better solution for a better user list would be to use the js-sdk for the widget and run the user list modifications to be done inside the widget.

proposals/3973-widgetapi-user-directory-search.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Proposal
+
+We will add a new interface to the widget API to search users in the user directory. We will introduce
```
toger5, line 18: I am wondering if using "we" is common in MSC's. I have not seen it before. I am not aware if there are any rules. But I think rephrasing without "we" would not hurt?

toger5: (this is very nitpicky indeed...)

dhenneke: Good point. I'm used to write ADRs for our own products which are written from our teams perspective. I will rewrite it later to better fit the MSC styling

proposals/3973-widgetapi-user-directory-search.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+Under `data`, all keys are a mirrored representation of the original `/_matrix/client/v3/user_directory/search`
+API. The `limit` field is optional and defaults to whatever the homeserver implementation uses.
+
+If the widget did not get approved for the capability required to send the event, the client MUST
+send an error response (as required currently by the capabilities system for widgets).
+
+The client SHOULD NOT modify the data of the request. The widget is responsible for producing valid
+events - the client MUST pass through any errors, such as permission errors, to the widget using the
```
andybalaam, line 46: This looks like it was copied from MSC2762 but doesn't quite fit here. Maybe the sentence "The widget is responsible..." should be removed completely?

proposals/3973-widgetapi-user-directory-search.md
```diff
@@ -0,0 +1,92 @@
...
+apply. This feature will allow the widget to be able to receive information about who is present on
+the homeserver / in the user directory. This information could be used by the widget to send it to
+a third-party server to store or misuse it. However, the access will only be possible when the user
+accepts the capability and grant access if the widget is trusted by the user.
+
+## Unstable prefix
+
+While this MSC is not present in the spec, clients and widgets should:
+
+- Use `org.matrix.msc3973.` in place of `m.` in all new identifiers of this MSC.
```
andybalaam, line 89: Is this needed? Doesn't the following bullet point cover everything? (Or is this a standard sentence we include in this section?)

proposals/3973-widgetapi-user-directory-search.md
dhenneke: This is is a copy from MSC2762 because it asks for the same behavior. MSC2762 is mainly defining the actions `read_event` and `send_event` but I would de-facto see it as a more general that kind-of defines the general behavior of the contract between the homeserver, the client, and the widget. But I'm fine with skipping it in this proposal.

PR #3974, opened 2023-03-07
Fix typo in example of MSC3758

turt2live:


turt2live:
@clokep [confirmed](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$MHsiu2lp1x0l9XwIScAW59N7FBV1xFkeeZbK7aDP7NI?via=matrix.org&via=libera.chat&via=element.io) that this is the expected behaviour, and it ultimately matches the spec the MSC is based upon, so merging without review.

cc @Fizzadar as author. If we've gotten this fix horribly wrong, please yell and we'll correct it.

PR #3976, opened 2023-03-08
Add mention of private sign off to contributing guidelines

turt2live:
also fix headings because I can never remember what the next symbol is when using the underline syntax

borrowed from Synapse: https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#private-sign-off


anoadragon453 (approved):
Thanks!

PR #3978, opened 2023-03-15
MSCxxxx: Deprecate room tagging

uhoreg:


proposals/3978-deprecate-room-tagging.md
```diff
@@ -0,0 +1,31 @@
...
+room tagging is not used by clients any more.
+
+## Proposal
+
+The ["Room Tagging"
+module](https://spec.matrix.org/unstable/client-server-api/#room-tagging) is
+deprecated.  It will be removed in a future version of the spec.  New clients
+should not implement support for it.
+
+## Potential issues
```
turt2live, line 13: Clients still use favourites and (to a degree) low priority tagging within Spaces.

proposals/3978-deprecate-room-tagging.md
```diff
@@ -0,0 +1,31 @@
+# MSC3978: Deprecate room tagging
+
+The functionality provided by room tagging can also be achieved by spaces, and
+room tagging is not used by clients any more.
```
deepbluev7, line 4: First of all spaces don't have namespacing. Currently clients can use namespaced tags to attach extra information to rooms, that they might need. For example they can distinguish between spaces that should be autojoined or not, if they should be rendered as a workspace or a community, use it to order top-level spaces, hide rooms from the roomlist, etc. All of that without coming up with their own name for some custom account data event. How do spaces achieve that? There is currently no way to tag a space as "implementation detail of client X".

Additionally room tags are much lighter weight and can be easily deleted. Spaces forever leave some state in the room, that this was either a child or a parent of room X. Tags do leave the tag event around, but only in your account data and the actual tag name and such is gone. Space editing is also heavily rate limited, which makes many, especially automated, modifications quite hard.

And lastly as Travis mentioned, they are still used for favourites, low priority rooms and server notices. How do spaces replace that?

I planned to rely much more heavily on tags as I improve the spaces support in my client. That way I can make spaces more usable, especialy when you can't edit a space, because it is maintained by someone else. I also frequently use it to attach some data to a room i.e. "Person I talked about X with" and such. While that can be achieved with spaces, instead of sending an account data event or 2, you now need to do: create space, add m.space.child, when leaving the room or untagging the room, remove the space child event, leave the space if it is empty and don't forget to forget it. That is waaaaaay more effort and easy to get wrong.

I do think there is some merit in slimming down the spec. Especially when new clients want to implement it, they might be confused why there are spaces and tags. But I think it is too early to remove them. Otoh, it doesn't matter for clients if tags are removed, since that is just an account data event with some rules attached. It can easily still be used by those clients.

clokep: I'm also skeptical that spaces can replace all of tags features -- it doesn't seem possible to use a space to segment the rooms into e.g. favorites vs. low priority using spaces without ignoring any public or corporate sort of hierarchies that exist. I'd really like to see this include a worked example of how all features of tags can be replaced with spaces.

proposals/3978-deprecate-room-tagging.md
uhoreg: Ah right, I forgot that tagging was used for favourites and low priority.  I was only thinking about user-defined tags.

PR #3981, opened 2023-03-19
MSC3981: `/relations` recursion

justjanne:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/justjanne/3981-relations-recursion/proposals/3981-relations-recursion.md)

----

Implementations:

* Homeserver: matrix-org/synapse#15315
* Client: matrix-org/matrix-js-sdk#3248

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3981#issuecomment-1860631322)

### DIFF
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -1,0 +1,63 @@
+# MSC3715: Add a pagination direction parameter to `/relations`
+
+[MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) introduced the
+`/relations` API to retrieve events which are directly related to a given event.
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
+
+Threads was one of the first usages of this API that allowed nested relations -
+an event may have an m.reaction or m.replace relation to another event, which 
+in turn may have an m.thread relation to the thread root.
+
+This forms a tree of relations, which is necessary for clients to retrieve to
+be able to correctly display threads and determine the latest event of a thread
+to be able to correctly send read receipts and determine the thread's 
+unread status.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API.
+
+> Whether to recursively include all nested relations of a given event. 
+>
+> If this is set to true, it will return the entire subtree of events related
+> to the specified event, directly or indirectly.
+> If this is set to false, it will only return events directly related to the 
+> specified event.
+>
+> One of: `[true false]`.
+
+In order to be backwards compatible with MSC2675 (and Synapse's legacy 
+implementation), the `recurse` parameter must be optional (defaulting to 
+`false`).
+
+Regardless of the value of the `recurse` parameter, events will always be 
+returned in the same order as they would be by the `/messages` API.
+
+If the API call specifies an `event_type` or `rel_type`, this filter will be
+applied to nested relations just as it is applied to direct relations.
+
+## Potential issues
+
+Naive implementations of recursive API endpoints frequently cause N+1 query 
+problems. Homeservers should take care to implementing this MSC to avoid 
+situations where a specifically crafted set of events and API calls could 
+amplify the load on the server unreasonably.
+
+## Alternatives
+
+1. Clients could fetch all relations recursively client-side, which would 
+   increase network traffic and server load significantly.
+2. A new, specialised endpoint could be created for threads, specifically 
+   designed to present separate timelines that, in all other ways, would
+   behave identically to `/messages`
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+Before standardization, `org.matrix.msc3981.recurse` should be used in place
+of `recurse`.
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -50,10 +50,10 @@ Regardless of the value of the `recurse` parameter, events will always be
 returned in topological ordering. Pagination and limiting shall also apply to 
 topological ordering.
 
-Filters specified via `event_type` or `rel_type` will only be applied to the 
-leaves of the graph formed by the related events, even if their only relation
-to the original given event is through an intermediate event that does not
-match these filters. 
+Filters specified via `event_type` or `rel_type` will be applied to leaves 
+and intermediate events on the graph formed by the related events.  
+Events that would match the filter, but whose only relation to the original 
+given event is through a non-matching intermediate event, will not be included.
 
 ## Potential issues
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -26,6 +26,9 @@ It is proposed to add the `recurse` parameter to the `/relations` API.
 > If this is set to false, it will only return events directly related to the 
 > specified event.
 >
+> It is recommended that at least 3 relations are traversed, implementations
+> should be careful to not infinitely recurse.
+>
 > One of: `[true false]`.
 
 In order to be backwards compatible the `recurse` parameter must be
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -1,4 +1,4 @@
-# MSC3715: Add a pagination direction parameter to `/relations`
+# MSC3981: `/relations` recursion
 
 [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) introduced the
 `/relations` API to retrieve events which are directly related to a given event.
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -20,15 +20,16 @@ reference as per [MSC3771] – chronological order is necessary.
 It is proposed to add the `recurse` parameter to the `/relations` API, defined
 as follows:
 
-> Whether to recursively include all nested relations of a given event. 
+> Whether to include events which relate only indirectly to the given event.
+> 
+> If set to false, only events which have direct a relation with the given 
+> event will be included.
+> 
+> If set to true, all events which relate to the given event, or relate to 
+> events that relate to the given event, will be included.
 >
-> If this is set to true, it will return the entire subtree of events related
-> to the specified event, directly or indirectly.
-> If this is set to false, it will only return events directly related to the 
-> specified event.
->
-> It is recommended that at least 3 relations are traversed. Implementations
-> should be careful to not infinitely recurse.
+> It is recommended that at least 3 levels of relationships are traversed. 
+> Implementations should be careful to not infinitely recurse.
 >
 > One of: `[true false]`.
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -126,7 +126,8 @@ return `[G, E]`.
 ### While the MSC is not yet part of a spec version
 
 During this period, to detect server support, clients should check for the
-presence of the `org.matrix.msc3981` flag in unstable_features on `/versions`.
+presence of the `org.matrix.msc3981` flag in the `unstable_features` map
+on [`/versions`](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientversions).
 
 Clients are also required to use `org.matrix.msc3981.recurse` in place
 of `recurse` at this time.
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -49,10 +49,12 @@ applied to nested relations just as it is applied to direct relations.
 
 ## Potential issues
 
-Naive implementations of recursive API endpoints frequently cause N+1 query 
-problems. Homeservers should take care to implementing this MSC to avoid 
-situations where a specifically crafted set of events and API calls could 
-amplify the load on the server unreasonably.
+Naive implementations might be tempted to provide support for this parameter
+through a thin shim which is functionally identical to the client doing 
+separate recursive `/relations` requests itself. This is ill-advised.
+
+Such an implementation would, given a specifically crafted set of events, 
+allow a client to cause unreasonable load.
 
 ## Alternatives
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -122,6 +122,11 @@ return `[G, E]`.
 Before standardization, `org.matrix.msc3981.recurse` should be used in place
 of `recurse`.
 
+## Unstable feature in `/version`
+
+Homeservers which support this MSC should indicate it by adding `org.matrix.msc3981` and
+`org.matrix.msc3981.stable` in the response to `GET /_matrix/client/versions` requests.
+
 [MSC2836]: https://github.com/matrix-org/matrix-spec-proposals/pull/2836
 [MSC3771]: https://github.com/matrix-org/matrix-spec-proposals/pull/3771
 [`/relations`]: https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv1roomsroomidrelationseventid
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -10,27 +10,23 @@ Threads was one of the first usages of this API that allowed nested relations -
 an event may have an [`m.reaction`] or [`m.replace`] relation to another event, 
 which in turn may have an `m.thread` relation to the thread root.
 
-This forms a tree of relations, which can currently only be traversed 
-efficiently in hierarchical, but not in chronological order. Yet, for some
-functionality – e.g., to determine which event a read receipt should 
-reference as per [MSC3771] – chronological order is necessary.
+This forms a tree of relations. A client wanting to display a thread will want
+to display reactions and edits to messages in the thread, and will therefore need
+the second-order related events in addition to just the events with a direct thread
+relation to the root.
 
-Previously, clients would be unable to obtain a consistent ordering of
-events in threads or related to threads.  
-Workarounds such as sending a separate `/relations` request per individual 
-message in each thread are only able to approximate the actual ordering,
-as they rely on timestamps.
-
-If a homeserver announces support for this feature, a client can stop this 
-fallback behavior and instead just append the recurse parameter to the initial 
-request.
+Clients can recursively perform the /relations queries on each event but this is
+very slow and does not give the client any information on how the events are ordered
+for the purpose of sending read receipts.
 
 ## Proposal
 
 It is proposed to add the `recurse` parameter to the `/relations` API, defined
 as follows:
 
-> Whether to include events which relate only indirectly to the given event.
+> Whether to additionally include events which only relate indirectly to the
+> given event,
+> ie. events related to the root events via one or more direct relationships.
 > 
 > If set to false, only events which have direct a relation with the given 
 > event will be included.
@@ -39,34 +35,31 @@ as follows:
 > events that relate to the given event, will be included.
 >
 > It is recommended that at least 3 levels of relationships are traversed. 
-> Implementations should be careful to not infinitely recurse.
+> Implementations may perform more but should be careful to not infinitely recurse.
 >
 > One of: `[true false]`.
 
 In order to be backwards compatible the `recurse` parameter must be
 optional (defaulting to `false`).
 
-In this situation, topological ordering is intended to refer to the same
-ordering that would be applied to the events when requested via the `/messages`
-api given the same `dir` parameter.
-
 Regardless of the value of the `recurse` parameter, events will always be 
-returned in topological ordering. Pagination and limiting shall also apply to 
-topological ordering.
+returned in topological ordering, ie. the same order in which the `/messages` API
+would return them (given the same `dir` parameter).
 
-Filters specified via `event_type` or `rel_type` will be applied to leaves 
-and intermediate events on the graph formed by the related events.  
-Events that would match the filter, but whose only relation to the original 
-given event is through a non-matching intermediate event, will not be included.
+Filters specified via `event_type` or `rel_type` will be applied to all events
+returned, whether direct or indirect relations. Events that would match the filter,
+but whose only relation to the original given event is through a non-matching
+intermediate event, will not be included. This means that supplying a `rel_type`
+parameter of `m.thread` is not appropriate for fetching all events in a thread since
+relations to the threaded events would be filtered out. For this purpose, clients should
+omit the `rel_type` parameter and perform any necessary filtering on the client side.
 
 ## Potential issues
 
 Naive implementations might be tempted to provide support for this parameter
 through a thin shim which is functionally identical to the client doing 
-separate recursive `/relations` requests itself. This is ill-advised.
-
-Such an implementation would, given a specifically crafted set of events, 
-allow a client to cause unreasonable load.
+separate recursive `/relations` requests itself. However this would allow a
+client to craft a set of events that would cause unreasonable load.
 
 ## Alternatives
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -7,13 +7,13 @@ This API has been used as basis of many other features and MSCs since then,
 including threads.
 
 Threads was one of the first usages of this API that allowed nested relations -
-an event may have an m.reaction or m.replace relation to another event, which 
-in turn may have an m.thread relation to the thread root.
+an event may have an `m.reaction` or `m.replace` relation to another event, 
+which in turn may have an `m.thread` relation to the thread root.
 
-This forms a tree of relations, which is necessary for clients to retrieve to
-be able to correctly display threads and determine the latest event of a thread
-to be able to correctly send read receipts and determine the thread's 
-unread status.
+This forms a tree of relations, which so far could only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
 
 ## Proposal
 
@@ -34,8 +34,13 @@ It is proposed to add the `recurse` parameter to the `/relations` API.
 In order to be backwards compatible the `recurse` parameter must be
 optional (defaulting to `false`).
 
+In this situation, topological ordering is intended to refer to the same
+ordering that would be applied to the events when requested via the `/messages`
+api given the same `dir` parameter.
+
 Regardless of the value of the `recurse` parameter, events will always be 
-returned in the same order as they would be by the `/messages` API.
+returned in topological ordering. Pagination and limiting shall also apply to 
+topological ordering.
 
 If the API call specifies an `event_type` or `rel_type`, this filter will be
 applied to nested relations just as it is applied to direct relations.
@@ -55,10 +60,44 @@ amplify the load on the server unreasonably.
    designed to present separate timelines that, in all other ways, would
    behave identically to `/messages`
 3. Twitter-style threads (see [MSC2836])
+4. Alternatively a `depth` parameter could have been specified, as in [MSC2836]  
+   We believe that a customizable depth would add unnecessary constraints to 
+   server implementers, as different server implementations may have different
+   performance considerations and may choose different limits. Additionally,
+   the maximum currently achievable depth is still low enough to avoid this
+   becoming an issue.
 
 ## Security considerations
 
 None.
+
+## Examples
+
+Given the following graph:
+
+```mermaid
+flowchart RL
+    subgraph Thread
+    G
+    E-->D
+    B-->A
+    end
+    B-.->|m.thread|A
+    G-.->|m.thread|A
+    E-.->|m.reaction|B
+    D-.->|m.edit|A
+    G-->F-->E
+    D-->C-->B
+```
+
+`/messages` with `dir=f` would return 
+`[A, B, C, D, E, F, G]`
+
+`/relations` on event `A` with `rel_type=m.thread` and `dir=f` would return 
+`[A, B, G]`. 
+
+`/relations` on event `A` with `recurse=true` and `dir=f` would return 
+`[A, B, D, E, G]`.
 
 ## Unstable prefix
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -90,14 +90,17 @@ flowchart RL
     D-->C-->B
 ```
 
-`/messages` with `dir=f` would return 
-`[A, B, C, D, E, F, G]`
+`/messages` with `dir=f` would 
+return `[A, B, C, D, E, F, G]`.
 
-`/relations` on event `A` with `rel_type=m.thread` and `dir=f` would return 
-`[A, B, G]`. 
+`/relations` on event `A` with `rel_type=m.thread` and `dir=f` would 
+return `[A, B, G]`. 
 
-`/relations` on event `A` with `recurse=true` and `dir=f` would return 
-`[A, B, D, E, G]`.
+`/relations` on event `A` with `recurse=true` and `dir=f` would 
+return `[A, B, D, E, G]`.
+
+`/relations` on event `A` with `recurse=true`, `dir=b` and `limit=2` would
+return `[G, E]`.
 
 ## Unstable prefix
 
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -28,9 +28,8 @@ It is proposed to add the `recurse` parameter to the `/relations` API.
 >
 > One of: `[true false]`.
 
-In order to be backwards compatible with MSC2675 (and Synapse's legacy 
-implementation), the `recurse` parameter must be optional (defaulting to 
-`false`).
+In order to be backwards compatible the `recurse` parameter must be
+optional (defaulting to `false`).
 
 Regardless of the value of the `recurse` parameter, events will always be 
 returned in the same order as they would be by the `/messages` API.
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -10,14 +10,15 @@ Threads was one of the first usages of this API that allowed nested relations -
 an event may have an `m.reaction` or `m.replace` relation to another event, 
 which in turn may have an `m.thread` relation to the thread root.
 
-This forms a tree of relations, which so far could only be traversed 
+This forms a tree of relations, which can currently only be traversed 
 efficiently in hierarchical, but not in chronological order. Yet, for some
 functionality – e.g., to determine which event a read receipt should 
 reference as per [MSC3771] – chronological order is necessary.
 
 ## Proposal
 
-It is proposed to add the `recurse` parameter to the `/relations` API.
+It is proposed to add the `recurse` parameter to the `/relations` API, defined
+as follows:
 
 > Whether to recursively include all nested relations of a given event. 
 >
@@ -26,7 +27,7 @@ It is proposed to add the `recurse` parameter to the `/relations` API.
 > If this is set to false, it will only return events directly related to the 
 > specified event.
 >
-> It is recommended that at least 3 relations are traversed, implementations
+> It is recommended that at least 3 relations are traversed. Implementations
 > should be careful to not infinitely recurse.
 >
 > One of: `[true false]`.
@@ -58,9 +59,9 @@ amplify the load on the server unreasonably.
    increase network traffic and server load significantly.
 2. A new, specialised endpoint could be created for threads, specifically 
    designed to present separate timelines that, in all other ways, would
-   behave identically to `/messages`
-3. Twitter-style threads (see [MSC2836])
-4. Alternatively a `depth` parameter could have been specified, as in [MSC2836]  
+   behave identically to `/messages`.
+3. Twitter-style threads (see [MSC2836]).
+4. Alternatively a `depth` parameter could have been specified, as in [MSC2836].  
    We believe that a customizable depth would add unnecessary constraints to 
    server implementers, as different server implementations may have different
    performance considerations and may choose different limits. Additionally,
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -119,13 +119,24 @@ return `[G, E]`.
 
 ## Unstable prefix
 
-Before standardization, `org.matrix.msc3981.recurse` should be used in place
-of `recurse`.
+### While the MSC is not yet part of a spec version
 
-## Unstable feature in `/version`
+During this period, to detect server support, clients should check for the
+presence of the `org.matrix.msc3981` flag in unstable_features on `/versions`.
 
-Homeservers which support this MSC should indicate it by adding `org.matrix.msc3981` and
-`org.matrix.msc3981.stable` in the response to `GET /_matrix/client/versions` requests.
+Clients are also required to use `org.matrix.msc3981.recurse` in place
+of `recurse` at this time.
+
+### Once the MSC is in a spec version
+
+Once this MSC becomes a part of a spec version, clients should rely on the
+presence of the spec version that supports this MSC in the `/version` response
+to determine support.
+
+Servers are encouraged to keep the `org.matrix.msc3827` flag around for a 
+reasonable amount of time to help smooth over the transition for clients.  
+"Reasonable" is intentionally left as an implementation detail, however the MSC
+process currently recommends at most 2 months from the date of spec release.
 
 [MSC2836]: https://github.com/matrix-org/matrix-spec-proposals/pull/2836
 [MSC3771]: https://github.com/matrix-org/matrix-spec-proposals/pull/3771
--- a/proposals/3981-relations-recursion.md
+++ b/proposals/3981-relations-recursion.md
@@ -1,13 +1,13 @@
 # MSC3981: `/relations` recursion
 
-[MSC2675] introduced the `/relations` API to retrieve events which are directly
-related to a given event.
+The [`/relations`] API allows clients to retrieve events which directly relate
+to a given event.
 
 This API has been used as basis of many other features and MSCs since then, 
 including threads.
 
 Threads was one of the first usages of this API that allowed nested relations -
-an event may have an `m.reaction` or `m.replace` relation to another event, 
+an event may have an [`m.reaction`] or [`m.replace`] relation to another event, 
 which in turn may have an `m.thread` relation to the thread root.
 
 This forms a tree of relations, which can currently only be traversed 
@@ -108,6 +108,8 @@ return `[G, E]`.
 Before standardization, `org.matrix.msc3981.recurse` should be used in place
 of `recurse`.
 
-[MSC2675]: https://github.com/matrix-org/matrix-spec-proposals/pull/2675
 [MSC2836]: https://github.com/matrix-org/matrix-spec-proposals/pull/2836
 [MSC3771]: https://github.com/matrix-org/matrix-spec-proposals/pull/3771
+[`/relations`]: https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv1roomsroomidrelationseventid
+[`m.reaction`]: https://github.com/matrix-org/matrix-spec-proposals/pull/2677
+[`m.replace`]: https://spec.matrix.org/v1.6/client-server-api/#event-replacements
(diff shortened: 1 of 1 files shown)


proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
+# MSC3715: Add a pagination direction parameter to `/relations`
```
turt2live, line 1:
```suggestion
# MSC3981: `/relations` recursion
```

justjanne: Oops, that's what I get for trying to write a draft on a weekend :see_no_evil: 

Thanks for spotting that, that'd have been quite embarrassing

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
...
+
+[MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) introduced the
+`/relations` API to retrieve events which are directly related to a given event.
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
+
+Threads was one of the first usages of this API that allowed nested relations -
+an event may have an m.reaction or m.replace relation to another event, which 
+in turn may have an m.thread relation to the thread root.
```
clokep, line 11:
```suggestion
Threads was one of the first usages of this API that allowed nested relations -
an event may have an `m.reaction` or `m.replace` relation to another event, which 
in turn may have an `m.thread` relation to the thread root.
```

Maybe include an example (MSC3771 has some useful things maybe?):

```mermaid
flowchart RL
    subgraph Thread
    D-->C-->B-->A
    end
    B-.->|m.thread|A
    C-.->|m.thread|A
    D-.->|m.thread|A
    E-.->|m.reaction|B
    F-.->|m.edit|C
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
...
+
+This forms a tree of relations, which is necessary for clients to retrieve to
+be able to correctly display threads and determine the latest event of a thread
+to be able to correctly send read receipts and determine the thread's 
+unread status.
```
clokep, line 16: For cross-referencing / documenting; MSC3771/MSC3773 do specify that relations should be recursed to find a thread. I think the issue this is trying to solve is understanding the *relative ordering of events* between those that are directly related to the thread and those that aren't?

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
...
+
+In order to be backwards compatible with MSC2675 (and Synapse's legacy 
+implementation), the `recurse` parameter must be optional (defaulting to 
+`false`).
+
+Regardless of the value of the `recurse` parameter, events will always be 
+returned in the same order as they would be by the `/messages` API.
+
+If the API call specifies an `event_type` or `rel_type`, this filter will be
+applied to nested relations just as it is applied to direct relations.
```
clokep, line 39: Are there any relations that are allowed to be applied to events of that same relation? It doesn't matter too much now, just curious.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
...
+It is proposed to add the `recurse` parameter to the `/relations` API.
+
+> Whether to recursively include all nested relations of a given event. 
+>
+> If this is set to true, it will return the entire subtree of events related
+> to the specified event, directly or indirectly.
+> If this is set to false, it will only return events directly related to the 
+> specified event.
+>
+> One of: `[true false]`.
```
clokep, line 29: I think this needs a similar warning to MSC3771/MSC3773:

```suggestion
> Whether to recursively include all nested relations of a given event. 
>
> If this is set to true, it will return the entire subtree of events related
> to the specified event, directly or indirectly.
> If this is set to false, it will only return events directly related to the 
> specified event.
>
> It is recommended that at least 3 relations are traversed, implementations
> should be careful to not infinitely recurse.
>
> One of: `[true false]`.
```

Those have a have a footnote of:

> Three relations is relatively arbitrary, but is meant to cover an edit or reaction to a thread (to an event with no relations, i.e. the root of a thread): `A<--[m.thread]--B<--[m.annotation]--C`. With an additional leftover for future improvements. This is considered reasonable since threads cannot fork, edits cannot modify relation information, and generally annotations to annotations are ignored by user interfaces.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,65 @@
...
+in turn may have an m.thread relation to the thread root.
+
+This forms a tree of relations, which is necessary for clients to retrieve to
+be able to correctly display threads and determine the latest event of a thread
+to be able to correctly send read receipts and determine the thread's 
+unread status.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API.
```
clokep, line 20: I suspect it is worth putting somewhere why a generic `recurse` (instead of something like `depth`) was used.

(I'm not suggesting `depth` would be better, but it is was MSC2836 proposed.)

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,65 @@
...
+> It is recommended that at least 3 relations are traversed, implementations
+> should be careful to not infinitely recurse.
+>
+> One of: `[true false]`.
+
+In order to be backwards compatible the `recurse` parameter must be
+optional (defaulting to `false`).
+
+Regardless of the value of the `recurse` parameter, events will always be 
+returned in the same order as they would be by the `/messages` API.
```
clokep, line 38: And per the pagination parameters (`dir`, `from`, `to`) I assume?

How does this all work if a `limit` is provided? I assume from the "returned in the same order" means the limiting would essentially just apply by topological order and nothing to do with how many relationships deep or anything like that.

Maybe a graph makes it clearer (the letter shows topological order):

```mermaid
flowchart RL
    E-->D-->C-->B-->A
    B-.->|m.thread|A
    E-.->|m.thread|A
    D-.->|m.reaction|B
    C-.->|m.edit|A
```

My assumption is that a limit of `3` would return `B`, `C`, and `D` regardless of how many layers of relations you have to traverse to get back to `A`.

(The other reasonable way I can think of this is `B`, `C`, `E` which are all closest to `A`.)

justjanne: > I assume from the "returned in the same order" means the limiting would essentially just apply by topological order and nothing to do with how many relationships deep or anything like that.

That's correct, that's exactly what I meant. I'll phrase this paragraph more clearly.

turt2live:
@matrix-org/spec-core-team the author is seeking technical review with the next stage being proposed FCP if possible. Please take a look with that in mind, and if the MSC is ready then `@mscbot fcp merge` it.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
+# MSC3981: `/relations` recursion
+
+[MSC2675] introduced the `/relations` API to retrieve events which are directly
+related to a given event.
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
+
+Threads was one of the first usages of this API that allowed nested relations -
+an event may have an `m.reaction` or `m.replace` relation to another event,
```
richvdh, line 10:
```suggestion
an event may have an [`m.reaction`](https://github.com/matrix-org/matrix-spec-proposals/pull/2677) or [`m.replace`](https://spec.matrix.org/v1.6/client-server-api/#event-replacements) relation to another event, 
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
+
+Threads was one of the first usages of this API that allowed nested relations -
+an event may have an `m.reaction` or `m.replace` relation to another event, 
+which in turn may have an `m.thread` relation to the thread root.
+
+This forms a tree of relations, which so far could only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
```
richvdh, line 14:
```suggestion
This forms a tree of relations, which can currently only be traversed 
efficiently in hierarchical, but not in chronological order. Yet, for some
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API.
```
richvdh, line 20:
```suggestion
It is proposed to add the `recurse` parameter to the `/relations` API, defined as follows:
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+
+> Whether to recursively include all nested relations of a given event. 
+>
+> If this is set to true, it will return the entire subtree of events related
+> to the specified event, directly or indirectly.
+> If this is set to false, it will only return events directly related to the 
+> specified event.
+>
+> It is recommended that at least 3 relations are traversed, implementations
+> should be careful to not infinitely recurse.
```
richvdh, line 30:
```suggestion
> It is recommended that at least 3 relations are traversed. Implementations
> should be careful to not infinitely recurse.
```

Could you define this better? What does it mean to "traverse" a relation?

Also, it seems to be in conflict with the line above which says "it will return the **entire** subtree of events related to the specified event"

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
+# MSC3981: `/relations` recursion
+
+[MSC2675] introduced the `/relations` API to retrieve events which are directly
+related to a given event.
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
```
richvdh, line 7: At this point, the spec is more useful than the MSC:

```suggestion
The [`/relations`](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv1roomsroomidrelationseventid)
API allows clients to retrieve events which directly
relate to a given event.
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+This forms a tree of relations, which so far could only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API.
+
+> Whether to recursively include all nested relations of a given event.
```
richvdh, line 22: I don't think this a great definition of what you are describing.

For me, the "relations" of a given event are those listed in the `m.relations` property of its content - which is to say, its *parents*. I think what you are talking about - and what `/relations` currently returns - are  its *children*. Note that the current spec is careful to avoid the term "relations" as it's quite ambiguous (and yes, if I had my way it wouldn't be called `/relations`.)

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+reference as per [MSC3771] – chronological order is necessary.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API.
+
+> Whether to recursively include all nested relations of a given event. 
+>
+> If this is set to true, it will return the entire subtree of events related
+> to the specified event, directly or indirectly.
```
richvdh, line 25:
```suggestion
> If this is set to true, it will return the entire subtree of events which relate
> to the specified event, directly or indirectly.
```

richvdh: likewise other clarifications below.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+In this situation, topological ordering is intended to refer to the same
+ordering that would be applied to the events when requested via the `/messages`
+api given the same `dir` parameter.
+
+Regardless of the value of the `recurse` parameter, events will always be 
+returned in topological ordering. Pagination and limiting shall also apply to 
+topological ordering.
+
+If the API call specifies an `event_type` or `rel_type`, this filter will be
+applied to nested relations just as it is applied to direct relations.
```
richvdh, line 46: could you be clear as to whether `event_type` and `rel_type` apply to the intermediate event, or just the leaves?

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Potential issues
+
+Naive implementations of recursive API endpoints frequently cause N+1 query 
+problems. Homeservers should take care to implementing this MSC to avoid
```
richvdh, line 51: I'm afraid I'm not familiar with the concept of an "N+1 query problem". Could you rephrase or clarify?

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+situations where a specifically crafted set of events and API calls could 
+amplify the load on the server unreasonably.
+
+## Alternatives
+
+1. Clients could fetch all relations recursively client-side, which would 
+   increase network traffic and server load significantly.
+2. A new, specialised endpoint could be created for threads, specifically 
+   designed to present separate timelines that, in all other ways, would
+   behave identically to `/messages`
```
richvdh, line 61:
```suggestion
   behave identically to `/messages`.
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+3. Twitter-style threads (see [MSC2836])
```
richvdh, line 62:
```suggestion
3. Twitter-style threads (see [MSC2836]).
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+4. Alternatively a `depth` parameter could have been specified, as in [MSC2836]
```
richvdh, line 63:
```suggestion
4. Alternatively a `depth` parameter could have been specified, as in [MSC2836].
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,112 @@
...
+    D-.->|m.edit|A
+    G-->F-->E
+    D-->C-->B
+```
+
+`/messages` with `dir=f` would 
+return `[A, B, C, D, E, F, G]`.
+
+`/relations` on event `A` with `rel_type=m.thread` and `dir=f` would 
+return `[A, B, G]`.
```
richvdh, line 97: why would `A` be returned here (or in the case below)?

richvdh (commented):
seems generally plausible from the technical point of view.

Apart from the editorial nits, I think the main thing this is missing before it can be considered for FCP is a demonstration implementation.

clokep:
> seems generally plausible from the technical point of view.

Thank so much for the review! 🎉 

> Apart from the editorial nits, I think the main thing this is missing before it can be considered for FCP is a demonstration implementation.

I added this information to the description, but matrix-org/synapse#15315 is a server implementation. matrix-org/matrix-js-sdk#3248 seems like a partial client implementation, but I'm unsure if there's more to do there and defer to @justjanne.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,118 @@
...
+
+Threads was one of the first usages of this API that allowed nested relations -
+an event may have an [`m.reaction`] or [`m.replace`] relation to another event, 
+which in turn may have an `m.thread` relation to the thread root.
+
+This forms a tree of relations, which can currently only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
+
```
MadLittleMods, line 17: Some context on what spawned this MSC and what kind of things it can fix would be nice to read about

proposals/3981-relations-recursion.md
clokep: Relationships create a tree of events, without the `recurse` flag you only look at the first layer of edges and nodes (so the `event_type` and `rel_type` have an obvious meaning of appyling to the nodes and edges, respectively).

With the `recurse` flag there's now multiple edges / nodes being traversed. Since we're returning a flatted tree, I think it only makes sense to apply these to the final node / edge being returned.

An example:


```mermaid
flowchart BT
    B[B: m.room.message]-->A[A: m.room.message]
    B-.->|m.thread|A
    C[C: m.file]-.->|m.thread|A
    G[G: m.reaction]-.->|m.annotation|F
    E[E: m.reaction]-.->|m.annotation|B
    D[D: m.file]-.->|m.edit|C
    F[F: m.room.message]-.->|m.reference|C
    G-->F-->E-->D-->C-->B
```

From my POV the following should be returned:

* `/relations/A?recurse=true`: `[B, C, D, E, F, G]`
* `/relations/A/m.thread?recurse=true`: `[B, C]`
* `/relations/A/m.thread/m.room.message?recurse=true`: `[B]`
* `/relations/A/m.annotation?recurse=true`: `[E, G]` (maybe this one is more interesting since they don't have a direct relationship to `A`?)

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,118 @@
...
+```mermaid
+flowchart RL
+    subgraph Thread
+    G
+    E-->D
+    B-->A
+    end
+    B-.->|m.thread|A
+    G-.->|m.thread|A
+    E-.->|m.reaction|B
```
clokep, line 91: `m.reaction` is the event type, `m.annotation` is the relation type:

```suggestion
    E-.->|m.annotation|B
```

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,129 @@
...
+> Whether to include events which relate only indirectly to the given event.
+> 
+> If set to false, only events which have direct a relation with the given 
+> event will be included.
+> 
+> If set to true, all events which relate to the given event, or relate to 
+> events that relate to the given event, will be included.
+>
+> It is recommended that at least 3 levels of relationships are traversed. 
+> Implementations should be careful to not infinitely recurse.
```
ShadowJonathan, line 38: This paragraph makes the reliability of the behaviour specified in this MSC weak, IMO, because the most pessimistic outcome of using this API would be that you only have 3 levels of recursively related events, while the behaviour of infinite recursion is not defined.

---

To address the levels problem, what about making this parameter (or another parameter) accept the amount of levels a client would see the server want to try to resolve? (The server would simply go up to that many levels of search, specifying `recurse=1000` to a single-relation event would not cause the server to do any extra work)

---

To address the issue of infinite recursion, while I'm currently not aware of any way an event can be edited to re-define its relation to another event, it doesn't mean that such a feature could exist in the future (to, for example, edit the message you've replied to, when you made a mistake).

As such, what about specifying that, if a server detects a cycle, it will simply cut it off, and return as-is? This way, the guarantee that it returns all recursively-related events at least once will still be satisfied.

proposals/3981-relations-recursion.md
```diff
@@ -122,6 +122,11 @@ return `[G, E]`.
 Before standardization, `org.matrix.msc3981.recurse` should be used in place
 of `recurse`.
 
+## Unstable feature in `/version`
+
+Homeservers which support this MSC should indicate it by adding `org.matrix.msc3981` and
+`org.matrix.msc3981.stable` in the response to `GET /_matrix/client/versions` requests.
```
richvdh, line 128: 1. could you explain why a feature flag is necessary? What does a client do when it comes across a server that does not support the the feature?

2. please could you make explicit what the difference between `org.matrix.msc3981` and `org.matrix.msc3981.stable` is? 

3. In addition, I think we should consider what happens once the feature reaches the final spec. That's a little hard to predict because we don't know which spec version it will be but it's possible to make some arrangements anyway.

I think https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3827-space-explore.md#unstable-prefix is quite a good example of how to handle  this.

richvdh:
> > Apart from the editorial nits, I think the main thing this is missing before it can be considered for FCP is a demonstration implementation.
> 
> I added this information to the description, but [matrix-org/synapse#15315](https://github.com/matrix-org/synapse/pull/15315) is a server implementation. [matrix-org/matrix-js-sdk#3248](https://github.com/matrix-org/matrix-js-sdk/pull/3248) seems like a partial client implementation, but I'm unsure if there's more to do there and defer to @justjanne.

@justjanne, @weeman1337: could we get a clarification on the situation here? My impression from the existence of https://github.com/matrix-org/synapse/pull/15558 is that this is still in testing?

weeman1337:
> > > Apart from the editorial nits, I think the main thing this is missing before it can be considered for FCP is a demonstration implementation.
> > 
> > 
> > I added this information to the description, but [matrix-org/synapse#15315](https://github.com/matrix-org/synapse/pull/15315) is a server implementation. [matrix-org/matrix-js-sdk#3248](https://github.com/matrix-org/matrix-js-sdk/pull/3248) seems like a partial client implementation, but I'm unsure if there's more to do there and defer to @justjanne.
> 
> @justjanne, @weeman1337: could we get a clarification on the situation here? My impression from the existence of [matrix-org/synapse#15558](https://github.com/matrix-org/synapse/pull/15558) is that this is still in testing?

Yes. Element Web adds support for it with 1.11.31 . But only if the server announces support for it as well. I've amended the MSC to add it to the `/versions` endpoint ([Synapse PR](https://github.com/matrix-org/synapse/pull/15558)).

proposals/3981-relations-recursion.md
justjanne: @weeman1337 added this section yesterday, after we noticed we didn't specify the feature flag in the MSC. You're right that this doesn't match the standard we should hold ourselves to, I’ll address it today.

justjanne: > could you explain why a feature flag is necessary? What does a client do when it comes across a server that does not support the the feature?

Currently, a client might be recursively fetching a relation tree via hundreds of relation requests. If this feature is supported, a client might skip this behavior and instead just append the `recurse` parameter to the initial request.

In fact, that's precisely what element-web does.

proposals/3981-relations-recursion.md
justjanne: > This paragraph weakens the reliability of the behaviour, IMO, because the most pessimistic outcome of using this API would be that you only have 3 levels of recursively related events

As there are currently no relations defined that allow you to get more than 3 levels of relations, that's an entirely academic exercise for the time being.

I'd suggest that we change this if and when we've defined types of relations that can nest deeper than this. At that point we'd also have a better understanding of what the edge cases might look like.

ShadowJonathan: My opinion is is that this adds tech debt, if the intent is to recurse a "reasonable" amount of times, and "reasonable" is considered to be 3 levels, due to the current set of relations there exists, then that wording should be used, not an exact number that'll surely become outdated when more relation types emerge, and would require an MSC to change, as it is specified behaviour.

The fact that this number is 3, given reasonable optimisation concerns and trade-offs, should be an implementation detail, and not a recommendation.

---

What I'm more worried about is that this sentence weakens the guarantee that not all recursively related events are returned, and that a client would have to call this endpoint again with some select events "just to be sure" that they got "all of them", if the client wants to be pedantic. (Which it must, if it wants to display consistent and correct behaviour)

justjanne: @clokep You originally suggested the number of 3, what's your opinion on this?

proposals/3981-relations-recursion.md
richvdh: please could you put some words to that effect in the MSC

justjanne: Added in [03e8d68](https://github.com/matrix-org/matrix-spec-proposals/pull/3981/commits/03e8d68670c013d09fb9791ce0d1d56455f31af4)

proposals/3981-relations-recursion.md
andybalaam: Yeah I am inclined to agree that it's easy to imagine needing >3 and having to change this.

Maybe a number like 10 could be a good minimum - then we'd "know" that usage patterns similar to what exists now would definitely be covered, and the only cases that would not be would be ones that allow arbitrary-length reference chains.

clokep: > @clokep You originally suggested the number of 3, what's your opinion on this?

This text was the recommended text (I think from @turt2live) in MSC3771 and MSC3773 and exists in a [similar form in the spec today](https://spec.matrix.org/v1.6/client-server-api/#threaded-read-receipts):

> Following the event relationships, it has a parent event which qualifies for one of the above. Implementations should not recurse infinitely, though: a maximum of 3 hops is recommended to cover indirect relationships.

tl;dr I'd leave it as is, but maybe the section on why a `depth` parameter was chosen needs to be expanded.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,149 @@
...
+This forms a tree of relations, which can currently only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
+
+Previously, clients would be unable to obtain a consistent ordering of
+events in threads or related to threads.  
+Workarounds such as sending a separate `/relations` request per individual 
+message in each thread are only able to approximate the actual ordering,
+as they rely on timestamps.
```
MadLittleMods, line 22: Expanding from [my previous comment asking for context](https://github.com/matrix-org/matrix-spec-proposals/pull/3981#discussion_r1173015449),

This touches a bit on the context of why which is good but technically why. Perhaps an example like below but where it breaks down with the current implementation. Or point out in that example how it breaks down without this MSC.

proposals/3981-relations-recursion.md
MadLittleMods: Discussion continued in https://github.com/matrix-org/matrix-spec-proposals/pull/3981#discussion_r1192740584

richvdh:
> > @justjanne, @weeman1337: could we get a clarification on the situation here? My impression from the existence of [matrix-org/synapse#15558](https://github.com/matrix-org/synapse/pull/15558) is that this is still in testing?
> 
> Yes. Element Web adds support for it with 1.11.31 . But only if the server announces support for it as well. I've amended the MSC to add it to the `/versions` endpoint ([Synapse PR](https://github.com/matrix-org/synapse/pull/15558)).
> 
> As far as I know we've activated the feature for `beta.ems.host`. Because of point_up we cannot test it.

So, to be absolutely clear: nobody has yet tested this?

turt2live:
@justjanne @weeman1337 : what's the plan for this MSC? The SCT has this tracked as needing fairly urgent attention, but the MSC doesn't have a *tested* implementation to validate it. Are you still interested in pursuing it?

weeman1337:
> @justjanne @weeman1337 : what's the plan for this MSC? The SCT has this tracked as needing fairly urgent attention, but the MSC doesn't have a _tested_ implementation to validate it. Are you still interested in pursuing it?

I would exclusively hand this over to @justjanne as I am busy with other things.

anoadragon453 (dismissed):
I can see why this would be useful from the proposal, but I was left a little confused about the justification in the introduction.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,149 @@
...
+
+In order to be backwards compatible the `recurse` parameter must be
+optional (defaulting to `false`).
+
+In this situation, topological ordering is intended to refer to the same
+ordering that would be applied to the events when requested via the `/messages`
+api given the same `dir` parameter.
+
+Regardless of the value of the `recurse` parameter, events will always be 
+returned in topological ordering. Pagination and limiting shall also apply to
```
Johennes, line 54: The problem statement at the beginning of this proposal says "chronological order is necessary" to correctly evaluate read receipts. How will this proposal help address this when the API always returns events in topological order?

MadLittleMods: It doesn't feel like "chronological order" is needed at all and I think it just might be a mistake in wording since `topological_ordering` is mentioned after. Additionally, `origin_server_ts` is an untrusted field (can be set to whatever integer) and is even designed to be deliberately [massaged by application services](https://spec.matrix.org/v1.6/application-service-api/#timestamp-massaging) so it's not useful to rely on for anything.

What read receipts probably care about is `topological_ordering` in the room DAG which `/messages` uses but the problem is the timeline in clients like Element mix in a bunch of messages from `/sync` which are in the order they were received on the server (`stream_ordering`). So the client has no clue what the latest message in the main timeline or thread timeline is from a `topological_ordering` perspective.

Good reading: https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#depth-and-stream-ordering

In terms of a solution to always choose the correct event for the read receipt, maybe we only construct timelines from `/messages` or ask for the PDU version of the events that include the `prev_events` in order to construct the DAG on the client and mix in messages appropriately. Maybe this is just a behind-the-scenes representation on the client we keep track of. Or some new endpoints which handle the magic for us, etc (many options).

AFAICT, this MSC improves read receipts on threads because it makes it easier to get the list of events for a given thread in `topological_order`.

Johennes: Thanks a ton Eric! 🙏

There's another recent proposal, [MSC4033](https://github.com/matrix-org/matrix-spec-proposals/pull/4033), that attempts to solve the ordering problem more holistically which might partially obsolete this one (as far as the ordering problematic is concerned).

MadLittleMods: :+1: Thanks for cross-linking! [MSC4033](https://github.com/matrix-org/matrix-spec-proposals/pull/4033) is definitely tackling the right problem and exposes the same confusion I had while trying to draft my last reply. The spec is not clear what order read receipts use, `topological_ordering` vs `stream_ordering`.

I assumed it was `topological_ordering` because you need to specify event ID's in some places and the [threaded receipts example](https://spec.matrix.org/v1.7/client-server-api/#threaded-read-receipts) shows a DAG but then also the receipts seem to also use timestamps (it's all very ambiguous).

In any case, the problem of `/messages` vs `/sync` using different orderings is the main thing to fight against. The good thing about that proposal is you could just take the highest `order` that you see in a given timeline for the read receipt (I think).

dbkr:
In my capacity of looking at stuck notifications (which turn out to be quite related to threads) and assuming the originals authors are not in a position to keep driving it, I'm going to attempt to rescue this MSC.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,63 @@
+# MSC3981: `/relations` recursion
+
+[MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) introduced the
+`/relations` API to retrieve events which are directly related to a given event.
+
+This API has been used as basis of many other features and MSCs since then, 
+including threads.
+
+Threads was one of the first usages of this API that allowed nested relations -
```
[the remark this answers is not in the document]
dbkr: I think this is saying only thread & reference relations can have second-order relations? In which case I think probably (although I'd be surprised if references ever did) but I'm not sure it makes a huge difference here.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,149 @@
...
+an event may have an [`m.reaction`] or [`m.replace`] relation to another event, 
+which in turn may have an `m.thread` relation to the thread root.
+
+This forms a tree of relations, which can currently only be traversed 
+efficiently in hierarchical, but not in chronological order. Yet, for some
+functionality – e.g., to determine which event a read receipt should 
+reference as per [MSC3771] – chronological order is necessary.
```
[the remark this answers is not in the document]
dbkr: I'm not 100% sure about this either and I don't see how ordering is a benefit of this MSC - I'm going to rewrite / remove this bit.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,149 @@
...
+If a homeserver announces support for this feature, a client can stop this 
+fallback behavior and instead just append the recurse parameter to the initial 
+request.
+
+## Proposal
+
+It is proposed to add the `recurse` parameter to the `/relations` API, defined
+as follows:
+
+> Whether to include events which relate only indirectly to the given event.
```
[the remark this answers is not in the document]
dbkr: Mmm, I think indirectly is what was intended (ie. do you include events that do not relate directly but do so indirectly?) but it could probably be clearer.

justjanne:
@dbkr awesome that you're offering to take over the MSC! If there's any way I can assist you, feel free to contact me.

proposals/3981-relations-recursion.md
dbkr: I think the spec is trying to ensure servers do the bare minimum level of recursion here: if they want to do more, they can.

dbkr:
I've committed some changes that hopefully should address the comments. NB. I've reduced the talk about ordering quite a bit and reduced it to a comparison to manual recursion where you'd get no ordering. For me, the main purpose of this is simply so we can get all the events in a thread in one fell swoop, relations and all, which is enough justification for its existence as far as I'm concerned.

@justjanne thanks so much for your comment. My main concern is whether I'm fundamentally missing something on the ordering stuff, but other than that I think it looks straightforward.

proposals/3981-relations-recursion.md
clokep: > I think the spec is trying to ensure servers do the bare minimum level of recursion here: if they want to do more, they can.

Yes, this is the goal. There are warnings not to infinitely recurse, but we wanted implementations to have a minimum required recursion amount.

dbkr: Thanks - in that case I'd say this is fine as-is then. Infinite recursion is warned against elsewhere and we don't need to tell people exactly how to write their servers. That said, I have to admit I'm struggling to understand the meaning of @ShadowJonathan 's original comment so please correct me if I've misunderstood.

proposals/3981-relations-recursion.md
dbkr: As in X relates to Y and Y relates to Z with the same rel type? Only replies I think. Either way I've clarified that a rel type filter + the recurse option is probably not useful for this reason.

proposals/3981-relations-recursion.md
dbkr: Hopefully I've made this a lot more straightforward now, ie. for me it's really just a question of needing a way to get all the events you need to render a thread, and that includes relations to the thread events.

proposals/3981-relations-recursion.md
dbkr: Likewise I've removed order-related stuff from the equation here, as I think it was a distraction, and in any case this flag should only change whether this API includes related events or not. It shouldn't affect the order.

dbkr:
Current status:
 * Implemented in js-sdk (and fixed with https://github.com/matrix-org/matrix-js-sdk/pull/3959)
 * Implemented in Synapse (behind a flag)
 * Impl checked by me and appears to be working and matches what's written here

dbkr:
On the belief that this is now ready as per my comment above...

@mscbot fcp merge

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,143 @@
...
+> given event,
+> ie. events related to the root events via one or more direct relationships.
+> 
+> If set to false, only events which have direct a relation with the given 
+> event will be included.
+> 
+> If set to true, all events which relate to the given event, or relate to 
+> events that relate to the given event, will be included.
+>
+> It is recommended that at least 3 levels of relationships are traversed.
```
uhoreg, line 37: Do we need a way for the server to communicate that it has stopped recursing?  e.g. hypothetically, if a client needs 4 levels, and the server only goes down 3 levels, the client needs to know if it is missing information and needs to fetch manually, or if all is good.

Could do something simple like add a `recursion_depth` parameter to the response.

uhoreg (commented):
@mscbot concern do we need to communicate recursion depth

uhoreg:
@mscbot concern do we need to communicate recursion depth

Maybe mscbot will like this one better?

proposals/3981-relations-recursion.md
turt2live: The recursion value comes from existing spec, where we thought the maximum number of indirect references was 2, so we added 1 and codified it. It may be worth converting the recommendation to a requirement: "minimum 3 hops" or something.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,143 @@
...
+4. Alternatively a `depth` parameter could have been specified, as in [MSC2836].  
+   We believe that a customizable depth would add unnecessary constraints to 
+   server implementers, as different server implementations may have different
+   performance considerations and may choose different limits. Additionally,
+   the maximum currently achievable depth is still low enough to avoid this
+   becoming an issue.
+
+## Security considerations
+
+None.
```
turt2live, line 81:
```suggestion
Security considerations are discussed inline throughout this proposal.
```

turt2live (approved):
generally lgtm

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,143 @@
...
+> If set to true, all events which relate to the given event, or relate to 
+> events that relate to the given event, will be included.
+>
+> It is recommended that at least 3 levels of relationships are traversed. 
+> Implementations may perform more but should be careful to not infinitely recurse.
+>
+> One of: `[true false]`.
+
+In order to be backwards compatible the `recurse` parameter must be
+optional (defaulting to `false`).
```
KitsuneRal, line 43: I wonder if existing client implementations are good enough to face the `recurse` parameter set to `true` by default? That would bring in the awesomeness of recursion to the existing fleet automagically (still with a caveat of limiting recursion to a sane value).

proposals/3981-relations-recursion.md
uhoreg: Even if we required (rather than recommended) a minimum of 3 levels, I think we still need to have a way for the server to indicate how many levels it actually traversed.  I know that the current maximum that clients would need is 2, but it's possible that in the future, or in some custom use cases, more would be needed, and so clients will need to know whether they have all the information that they need, or whether they need to fetch more.

If we don't want to change existing implementations, we could say that if the server doesn't include a `recursion_depth` parameter in the response, then clients must not depend on having more than 3 levels available.  Though, come to think of it, having a `recursion_depth` parameter in the response might be an easier way to detect server support -- clients can blindly set the `recurse` parameter in the request, and if the server includes a `recursion_depth` in the response, then they know that the server supports recursion.  If not (or if the `recursion_depth` is lower than what it needs), then it knows that it needs to do the recursion on its own.

dbkr: Mmm, it's possible it could bite us in the future if we did ever have something more recurse-y. I wouldn't want to spend too long chasing perfection, but the `recursion_depth` response param does sound like a fairly simple way of conveying this and involves very few changes.

proposals/3981-relations-recursion.md
dbkr: Maybe, but personally I'd rather just flip the flag in the client once I know it worked that go scouting for what clients implement it and try to make sure they all behave sensibly when the server breaks (old) spec.

proposals/3981-relations-recursion.md
dbkr: Updated and also implemented in synapse in https://github.com/element-hq/synapse/pull/16775

uhoreg:
@mscbot resolve do we need to communicate recursion depth

proposals/3981-relations-recursion.md
KitsuneRal: My main concern here is that we may extend the API surface for no actual use, with one option (`false`, that is) being inherently broken in some way. I mean, it's just not right when reactions don't get pulled along with other messages in the same thread, right? I'm totally fine to punt the removal of this into some future MSC, just in case.

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,161 @@
...
+4. Alternatively a `depth` parameter could have been specified, as in [MSC2836].  
+   We believe that a customizable depth would add unnecessary constraints to 
+   server implementers, as different server implementations may have different
+   performance considerations and may choose different limits. Additionally,
+   the maximum currently achievable depth is still low enough to avoid this
+   becoming an issue.
+
+## Security considerations
+
+Security considerations are discussed inline throughout this proposal.
```
ara4n, line 96: even though this is true, i suspect it's wise to re-summarise them here for ease of reference & indexing (e.g. https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/msc4016/proposals/4016-streaming-e2ee-file-transfer.md#security-considerations deliberately duplicates the security concerns to gather them into one place).  Not a blocker for FCP, though, more mentioning it for future reference

proposals/3981-relations-recursion.md
```diff
@@ -0,0 +1,161 @@
...
+client to craft a set of events that would cause unreasonable load.
+
+## Alternatives
+
+1. Clients could fetch all relations recursively client-side, which would 
+   increase network traffic and server load significantly.
+2. A new, specialised endpoint could be created for threads, specifically 
+   designed to present separate timelines that, in all other ways, would
+   behave identically to `/messages`.
+3. Twitter-style threads (see [MSC2836]).
```
ara4n, line 86: When I proposed aggregations (in the mess that is [MSC1849](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/msc1849/proposals/1849-aggregations.md)) the intention was basically to define a graph of relations linking events together:

> If it helps, you can think of relations as a "subject verb object" triple, where the subject is the relation event itself; the verb is the `rel_type` field of the `m.relationship` and the object is the `event_id` field.

I assumed that when users wanted to do complicated graph queries (e.g. "give me all the events in this thread, including their aggregations") or "give me the siblings of this twitter thread") we'd end up with a generic API for navigating the relations graph - and that's what kegan & I came up with in [MSC2836].

Personally, my hunch is still that a single generic API for navigating the relation graph would have been a better solution than having lots of incremental MSCs which gradually nudge the existing APIs in the right direction. However, we've ended up down the incrementalist path anyway now, and this MSC seems a good enough way to solve the immediate problem of "give me all the events which relate to this thread".

However, I have a horrible feeling that we're promptly going to hit another limitation for another relations use case (e.g. "I want to see the last 20 beacon locations for the 5 most recent location-shares in this thread. Including if the location-share event subsequently got edited" - or a similar one for VoIP signalling using aggregations, or even MSC4016-style file transfer updates via aggregations)... and find ourselves wanting a more generic API, again.

I'm also dubious about "the server may just ignore your recursion request, impose an arbitrary depth, and the client will have to figure it out".  I'd have thought it would be better for the client to state its requirements  (i.e. "i need depth=3 for my use case") and for the server to fulfil those... but to cap resource usage by the server then batching up the responses in whatever size it's most comfortable with (or by capping the pagination requests that the client makes).  So even if the client asks for some horribly inefficient query, the server can decline to calculate more than 10 results at a time or whatever.  Otherwise it's passing a lot of complexity onto the client to work around the server's limitations, and the design of Matrix has always been for the server to do the heavy-lifting rather than the client, in order to make clients as easy to write as possible.

So, TL;DR: I don't love this, but it's an acceptable solution for where we're at right now and I'm not going to block it, lest perfect be the enemy of good. In future I really hope we converge on a robust generic way in a new MSC (or MSC2836 gets dusted off) in order to walk the relations DAG rather than sprouting incremental tweaks.

ara4n (approved):
I have reservations on the overall strategy here, but they're not blocking - we should FCP this and revisit relations in future, imo.

proposals/3981-relations-recursion.md
dbkr: I feel fairly similarly about this really in that much more advanced uses of relations will likely require a more sophisticated API that can properly control recursion depth and what rel_types get recursively returned, etc. I also think the way we've chosen to implement features, ie. with threads as a live feature in Matrix now that, apart from not working very well itself is also causing other parts of the protocol to be flakey: we need to be fairly tactical here. Obviously we can introduce a more full-featured API separately but equally it's a shame that servers will have both APIs to maintain.

proposals/3981-relations-recursion.md
turt2live: @dbkr can this get folded into the MSC before it's merged please?

dbkr: Done in bdae5772768bbbd142f58ac0be0ea2a807c2f25d

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1746

turt2live:
Merged 🎉

PR #3983, opened 2023-03-23
MSC3983: Sending One-Time Key (OTK) claims to appservices

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/enc-as/otk-query/proposals/3983-sending-otk-claims-to-appservices.md)

Implementations:
* bot-sdk: https://github.com/turt2live/matrix-bot-sdk/pull/307
* Synapse: https://github.com/matrix-org/synapse/pull/15314
* Actual appservice which does something with this: ??? (possible effort would be https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/issues/29 )

### DIFF
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -125,9 +125,10 @@ being able to (securely) DM bridge users is a valuable consideration for this MS
 encryption for appservices might take the shape of an appservice-wide device of some sort.
 
 It could be argued that supporting a fallback key for appservices is too much considering their uptime,
-however in practice a protocol should not be designed around 100% uptime. This is additionally why the
-proposal doesn't suggest proxying device/signing key queries to the appservice: the appservice could be
-down, leaving encryption broken until it comes back online.
+however in practice appservices are not quite able to achieve 100% uptime. This proposal doesn't propose
+proxying device/signing key queries to the appservice for the same reliability concerns, though appservices
+which wish to opt to do so anyways could use [MSC3984](https://github.com/matrix-org/matrix-spec-proposals/pull/3984).
+
 ## Additional uses
 
 An appservice aiming to bridge two different encryption systems might use this endpoint to save on data,
@@ -150,3 +151,6 @@ which don't support the endpoint and thus engage in the behaviour described by t
 
 This MSC has no direct dependencies, however is of little use without being partnered with something
 like [MSC3202](https://github.com/matrix-org/matrix-spec-proposals/pull/3202).
+
+This MSC is additionally useful when paired with [MSC3984](https://github.com/matrix-org/matrix-spec-proposals/pull/3984),
+though has no direct dependency.
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -107,7 +107,7 @@ for the user.
 
 **TODO**: This is probably best as its own MSC.
 
-Independent of the appservice having `/keys/claim` proxied to it, it may be desireable for both the
+Independent of the appservice having `/keys/claim` proxied to it, it may be desirable for both the
 fallback and one-time key to be returned. Servers should *always* include the fallback key alongside
 the requested OTKs. When using this proposal's new endpoint, the server should use the fallback key
 from the appservice's response rather than a previously stored fallback key, if present (if the
@@ -120,6 +120,8 @@ do upon first (known) use.
 
 Clients can determine which of the keys returned is the fallback key by `fallback: true` on the returned
 keys.
+
+Servers MUST NOT mark the fallback key as "used" unless no other OTKs are returned.
 
 ## Potential issues
 
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -126,6 +126,12 @@ It could be argued that supporting a fallback key for appservices is too much co
 however in practice a protocol should not be designed around 100% uptime. This is additionally why the
 proposal doesn't suggest proxying device/signing key queries to the appservice: the appservice could be
 down, leaving encryption broken until it comes back online.
+## Additional uses
+
+An appservice aiming to bridge two different encryption systems might use this endpoint to save on data,
+though currently the encryption used on both sides of the bridge would need to be compatible (ie: signatures
+from device IDs and user IDs need to exist). In future, other MSCs might make encryption bridges easier to
+build.
 
 ## Security considerations
 
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -116,7 +116,11 @@ reliability concerns.
 
 ## Alternatives
 
-No major alternatives.
+Many encryption-capable bridges today can avoid uploading OTKs (and sometimes even device keys) because
+they have a bot user in the room. The bot user uploads its keys, but the remaining bridge users do not.
+This works if the bridge users don't need to be involved in rooms without the bot user present, though
+being able to (securely) DM bridge users is a valuable consideration for this MSC. In future, scalable
+encryption for appservices might take the shape of an appservice-wide device of some sort.
 
 It could be argued that supporting a fallback key for appservices is too much considering their uptime,
 however in practice a protocol should not be designed around 100% uptime. This is additionally why the
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -81,7 +81,9 @@ if desirable. This is an optional optimization for homeserver implementations. I
 keys are claimed for one device.
 
 If the appservice responds with an error of any kind (including timeout), the homeserver uses the
-fallback key, if known.
+fallback key, if known. The homeserver additionally uses the fallback key (if known) to fill in
+missing keys from the appservice. For example, if the homeserver requested 2 keys for Alice but
+the appservice only provided 1, the homeserver would use the fallback key to fulfill the second.
 
 In this case, the appservice is responsible for ensuring it doesn't use a key twice. The
 `device_one_time_keys_count` field for the appservice (over MSC3202, for example) would be zero. In
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -116,6 +116,10 @@ any keys for the user(s). This could be mitigated with an implementation-specifi
 and just do proxying, though for the general case in this MSC the fallback key consideration is kept for
 reliability concerns.
 
+Similarly, if an appservice doesn't intend on uploading keys (because it doesn't support encryption) and
+indicates the route is [unknown](https://spec.matrix.org/v1.6/application-service-api/#unknown-routes),
+the homeserver could avoid calling appservice with a backoff to prevent excessive calls.
+
 ## Alternatives
 
 Many encryption-capable bridges today can avoid uploading OTKs (and sometimes even device keys) because
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -89,9 +89,9 @@ many implementations, when this field falls below a threshold it is common for u
 happen: appservices intending on using the new API should not perform those uploads as it means,
 quite simply, not using the new API.
 
-Normally, a client (including appservices) would consider a OTK "used" once it sees it in a to-device
-message. Given the claim query is proxied directly to where the keys are generated in this proposal,
-appservices *can* consider a key "used" when it is claimed, before seeing it in a to-device message.
+Normally the homeserver would be [ensuring](https://spec.matrix.org/v1.6/client-server-api/#one-time-and-fallback-keys)
+OTKs are only used once, however with the appservice serving the endpoint it becomes the responsibility
+of the appservice to perform this check.
 
 If the homeserver uses the fallback key, that will be communicated in the traditional ways to the
 appservice (namely through `device_unused_fallback_key_types` in the case of MSC3202).
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -103,6 +103,24 @@ to have a namespace covering the user: instead of guessing or going around to ea
 user to be in an exclusive namespace. This guarantees that there's only one appservice responsible
 for the user.
 
+## Returning extra keys
+
+**TODO**: This is probably best as its own MSC.
+
+Independent of the appservice having `/keys/claim` proxied to it, it may be desireable for both the
+fallback and one-time key to be returned. Servers should *always* include the fallback key alongside
+the requested OTKs. When using this proposal's new endpoint, the server should use the fallback key
+from the appservice's response rather than a previously stored fallback key, if present (if the
+appservice doesn't respond with a fallback key then the server uses the stored fallback key instead,
+if known). 
+
+The server SHOULD NOT replace any uploaded fallback keys with ones returned by the appservice via
+this proposal. The appservice MUST re-upload the fallback key if it wants to replace it, as it would
+do upon first (known) use.
+
+Clients can determine which of the keys returned is the fallback key by `fallback: true` on the returned
+keys.
+
 ## Potential issues
 
 As described, the appservice could be offline or in fact experience a worse uptime than the homeserver.
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -72,6 +72,9 @@ API described below:
 }
 ```
 
+*Note*: Like other appservice endpoints, this endpoint should *not* be ratelimited and *does* require
+normal [authentication](https://spec.matrix.org/v1.6/application-service-api/#authorization).
+
 Multiple users, devices, and keys for those devices can be claimed in a single request. This is to
 allow homeservers to batch multiple client/federation requests into a single request on the appservice,
 if desirable. This is an optional optimization for homeserver implementations. In the example above, 2
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -109,6 +109,11 @@ will be bad), they can simply upload keys in advance, just like before this prop
 the appservice is trying to use the API but is offline, they *should* have a fallback key to continue
 using as, well, a fallback.
 
+For appservices which never intend to upload keys there is a bit of a wasted lookup to see if there are
+any keys for the user(s). This could be mitigated with an implementation-specific flag to skip the lookup
+and just do proxying, though for the general case in this MSC the fallback key consideration is kept for
+reliability concerns.
+
 ## Alternatives
 
 No major alternatives.
--- a/proposals/3983-sending-otk-claims-to-appservices.md
+++ b/proposals/3983-sending-otk-claims-to-appservices.md
@@ -1,0 +1,132 @@
+# MSC3983: Sending One-Time Key (OTK) claims to appservices
+
+Presently in Matrix, the public portion of OTKs are [uploaded](https://spec.matrix.org/v1.6/client-server-api/#uploading-keys)
+to the homeserver to ensure other devices can encrypt new messages without requiring the device to
+be online and responsive. This works for devices operating exclusively over the Client-Server API,
+however [appservices](https://spec.matrix.org/v1.6/application-service-api/) looking to support
+encryption (through [MSC3202](https://github.com/matrix-org/matrix-spec-proposals/pull/3202) or
+similar) could have millions or billions of users on them, which can easily translate to quite a
+few public keys needing to be uploaded to the homeserver.
+
+Given appservices *generally* have an uptime which is equivalent to the homeserver itself, and will
+have already stored the public portion of its OTKs somewhere, we can save a bit of duplication by
+having the homeserver delegate [`/keys/claim`](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3keysclaim)
+requests to the appservice.
+
+In numbers, a conservative estimate for an interoperable messaging bridge (appservice) would be
+500 million users. Each user generates between 50 and 100 OTKs, so we'll pick the low end at 50.
+That's 25 **billion** public keys. Currently in Matrix, that means the appservice stores 25 billion
+keys and the homeserver stores a copy of those 25 billion keys.
+
+This proposal introduces a mechanism for saving the homeserver from duplicating 25 billion keys.
+
+## Background
+
+Appservices can register a [namespace](https://spec.matrix.org/v1.6/application-service-api/#registration)
+of users either exclusively (no one else can register users matching the regex) or implicitly (the
+appservice receives events about those users, but can't prevent registration). Implicit namespaces
+can be shared across multiple appservices.
+
+## Proposal
+
+For users under an appservice's explicit namespace, if that user has no unused OTKs (excluding fallback
+keys) on the homeserver, the homeserver proxies the following APIs to the appservice using the new
+API described below:
+* [`/_matrix/client/v3/keys/claim`](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3keysclaim)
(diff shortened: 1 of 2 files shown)


proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,132 @@
...
+
+In this case, the appservice is responsible for ensuring it doesn't use a key twice. The
+`device_one_time_keys_count` field for the appservice (over MSC3202, for example) would be zero. In
+many implementations, when this field falls below a threshold it is common for upload requests to
+happen: appservices intending on using the new API should not perform those uploads as it means,
+quite simply, not using the new API.
+
+Normally, a client (including appservices) would consider a OTK "used" once it sees it in a to-device
+message. Given the claim query is proxied directly to where the keys are generated in this proposal,
+appservices *can* consider a key "used" when it is claimed, before seeing it in a to-device message.
```
uhoreg, line 91: Generally, clients don't need to care about whether a key has been claimed.  They only need to care about whether they receive a message using that key, since they need to remember the private key until that point.  So I don't think this paragraph is necessary.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,135 @@
...
+*Note*: Like other appservice endpoints, this endpoint should *not* be ratelimited and *does* require
+normal [authentication](https://spec.matrix.org/v1.6/application-service-api/#authorization).
+
+Multiple users, devices, and keys for those devices can be claimed in a single request. This is to
+allow homeservers to batch multiple client/federation requests into a single request on the appservice,
+if desirable. This is an optional optimization for homeserver implementations. In the example above, 2
+keys are claimed for one device.
+
+If the appservice responds with an error of any kind (including timeout), the homeserver uses the
+fallback key, if known.
```
turt2live, line 84: the server should also fill gaps in the response (ie: appservice misses a key/user) with the fallback key. 

see https://github.com/matrix-org/synapse/pull/15314#discussion_r1146977933

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,135 @@
...
+using as, well, a fallback.
+
+## Alternatives
+
+No major alternatives.
+
+It could be argued that supporting a fallback key for appservices is too much considering their uptime,
+however in practice a protocol should not be designed around 100% uptime. This is additionally why the
+proposal doesn't suggest proxying device/signing key queries to the appservice: the appservice could be
+down, leaving encryption broken until it comes back online.
```
turt2live, line 119: While uptime might not be a concern, the appservice might still not want to duplicate even 500 million device keys to the homeserver, so we should probably support that function to.

This sort of behaviour would likely be its own MSC, but when combined with this proposal would mean appservices have an option to upload *nothing* to the homeserver, instead proxying all the requests to the appservice itself. Appservices doing this would be accepting the implicit reliability concerns.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,135 @@
...
+
+Appservices can register a [namespace](https://spec.matrix.org/v1.6/application-service-api/#registration)
+of users either exclusively (no one else can register users matching the regex) or implicitly (the
+appservice receives events about those users, but can't prevent registration). Implicit namespaces
+can be shared across multiple appservices.
+
+## Proposal
+
+For users under an appservice's explicit namespace, if that user has no unused OTKs (excluding fallback
+keys) on the homeserver, the homeserver proxies the following APIs to the appservice using the new
```
clokep, line 33: One downside of the "no unused OTKs" approach is that the homeserver must:

1. Query internally for OTKs for each user.
2. Then query the appservice.
3. Then (maybe) query for fallback keys internally.

I find it likely that an appservice either always or never uploads the OTKs so steps 1 or 2 are always going to be a semi-expensive no-op. (Either a wasted database query OR a wasted HTTP request, both likely involve requests over a network too.)

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,135 @@
...
+
+As described, the appservice could be offline or in fact experience a worse uptime than the homeserver.
+This new API is optional for appservices: if they don't want to use it (because they know their uptime
+will be bad), they can simply upload keys in advance, just like before this proposal. Similarly, if
+the appservice is trying to use the API but is offline, they *should* have a fallback key to continue
+using as, well, a fallback.
+
+## Alternatives
+
+No major alternatives.
```
tulir, line 114: The alternative used by the majority of existing e2be-capable bridges is simply not creating ridiculous amounts of devices. One device with 50 OTKs is perfectly sufficient, creating multiple devices has no real advantages when all the keys are stored in the same place anyway.

I think the primary real use case for this MSC would be bridging OTK claims somewhere else, like for hypothetical e2ee over bridges (which doesn't exist yet and has several other blockers, but this would at least solve one of them). Non-bridge appservices could also have some use for proxying OTK claims.

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: The database query I think would normally happen anyways to serve the endpoint - it could also include a condition to get all keys then filter out the fallback keys within the application (temporarily caching them so it can use them after the http hit). The http hit still might be wasted, though this would be the appservice's "fault" for not uploading keys.

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: One large encryption-capable device does tend to mean the puppets can't do their own encryption, though. (Or they have to be the first to initiate)

Which other blockers are you thinking about for bridging encryption?

proposals/3983-sending-otk-claims-to-appservices.md
clokep: I guess worst-case is an appservice exclusive user for an AS which doesn't upload keys, this adds an additional no-op step.

Regardless, I don't think this is a big deal, just wanted to point it out.

proposals/3983-sending-otk-claims-to-appservices.md
tulir: I guess Element's new sharing keys before creating rooms would prevent starting encrypted chats with ghosts that don't have keys. In any other case, the ghost doesn't need to do its own encryption, in new rooms it can just invite the bot immediately to help with encryption. I think it'd be better to find solutions for those few problematic cases than to generate 25 billion OTKs and 500 million device keys.

Bridging without breaking encryption would require user and device IDs to match on both sides because they're signed in various places (like device keys). The encryption scheme and plaintext format of course have to match too, but the user ID signing likely even prevents matrix-matrix bridging with e2ee currently.

turt2live: I suppose it depends on the kind of bridge. Traditional bridges with a bot user can use the workaround, though bridges which want to allow DMs or similar would indeed need OTKs of some kind. A future MSC would almost certainly figure out a way to give the whole appservice a single device/set of keys, but for now this probably works well enough.

Signing is an interesting point. I'm not sure how it'd immediately be resolved, though this also starts to stretch into domains I'm not totally familiar with. This MSC would indeed remove at least one of the blockers though - others are probably needed to make an encryption bridge function at a larger scale (particularly if the other side isn't using something close enough to Olm/Megolm).

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: yea, it's a bit of a waste in that sense, though in theory the homeserver can cache the query or if it knows in advance then it could just not bother with the query (ie: an implementation-specific appservice registration flag to say the appservice will never upload keys worth considering)

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: covered in 0b9f6d9005710f0526a1825bae31208dd41d7668 and 8a1dd86c69b32d1717dde8765ca362a363ae9df4 I think

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: 5a4a6c64d1e5b2ee928464f8e84d4f14c24d2b59

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: a95bf4f55d02853da22c4c9dcb85d751282686dc

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,160 @@
...
+user to be in an exclusive namespace. This guarantees that there's only one appservice responsible
+for the user.
+
+## Potential issues
+
+As described, the appservice could be offline or in fact experience a worse uptime than the homeserver.
+This new API is optional for appservices: if they don't want to use it (because they know their uptime
+will be bad), they can simply upload keys in advance, just like before this proposal. Similarly, if
+the appservice is trying to use the API but is offline, they *should* have a fallback key to continue
+using as, well, a fallback.
```
dkasak, line 112: Who is offline in this sentence? And which API are we talking about? Also, who is *they*?

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,160 @@
...
+It could be argued that supporting a fallback key for appservices is too much considering their uptime,
+however in practice appservices are not quite able to achieve 100% uptime. This proposal doesn't propose
+proxying device/signing key queries to the appservice for the same reliability concerns, though appservices
+which wish to opt to do so anyways could use [MSC3984](https://github.com/matrix-org/matrix-spec-proposals/pull/3984).
+
+## Additional uses
+
+An appservice aiming to bridge two different encryption systems might use this endpoint to save on data,
+though currently the encryption used on both sides of the bridge would need to be compatible (ie: signatures
+from device IDs and user IDs need to exist). In future, other MSCs might make encryption bridges easier to
```
dkasak, line 140:
```suggestion
from device and user identities need to exist). In the future, other MSCs might make encryption bridges easier to
```

Device IDs and user IDs can't produce signatures, so the proposed change would make it a bit clearer to me.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,160 @@
...
+If the appservice responds with an error of any kind (including timeout), the homeserver uses the
+fallback key, if known. The homeserver additionally uses the fallback key (if known) to fill in
+missing keys from the appservice. For example, if the homeserver requested 2 keys for Alice but
+the appservice only provided 1, the homeserver would use the fallback key to fulfill the second.
+
+In this case, the appservice is responsible for ensuring it doesn't use a key twice. The
+`device_one_time_keys_count` field for the appservice (over MSC3202, for example) would be zero. In
+many implementations, when this field falls below a threshold it is common for upload requests to
+happen: appservices intending on using the new API should not perform those uploads as it means,
+quite simply, not using the new API.
```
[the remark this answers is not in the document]
turt2live: It's more of a warning to existing implementations of crypto: nearly all of them do a `if (count < 50) { generateAndUploadKeys() }` call, which would be a problem here.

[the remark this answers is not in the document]
dkasak: Hmm sure, but what are implementations supposed to do instead?

proposals/3983-sending-otk-claims-to-appservices.md
dkasak: I gathered as much 😄 But I'm still having trouble parsing that sentence in exact terms. e.g. If I interpret "but is offline" as "but the appservice is offline", how is the appservice able to do anything given it's offline?

I'd suggest rewriting the sentence with less implicit words / pronouns.

proposals/3983-sending-otk-claims-to-appservices.md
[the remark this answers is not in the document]
turt2live: "not that" :p 

Normally implementation details like this wouldn't be called out, however given the chance for everyone to have the exact same bug (intending to use new API, uploads keys by accident, new API isn't used), this is called out here.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,160 @@
...
+
+Multiple users, devices, and keys for those devices can be claimed in a single request. This is to
+allow homeservers to batch multiple client/federation requests into a single request on the appservice,
+if desirable. This is an optional optimization for homeserver implementations. In the example above, 2
+keys are claimed for one device.
+
+If the appservice responds with an error of any kind (including timeout), the homeserver uses the
+fallback key, if known. The homeserver additionally uses the fallback key (if known) to fill in
+missing keys from the appservice. For example, if the homeserver requested 2 keys for Alice but
+the appservice only provided 1, the homeserver would use the fallback key to fulfill the second.
```
clokep, line 86: I'm left wondering why the appservice would bother uploaded fallback keys when (to my reading) it could return them directly as part of the response.

I suppose uploading them would still be useful for error conditions, however.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -103,6 +103,24 @@ to have a namespace covering the user: instead of guessing or going around to ea
...
+## Returning extra keys
+
+**TODO**: This is probably best as its own MSC.
+
+Independent of the appservice having `/keys/claim` proxied to it, it may be desireable for both the
+fallback and one-time key to be returned. Servers should *always* include the fallback key alongside
+the requested OTKs. When using this proposal's new endpoint, the server should use the fallback key
+from the appservice's response rather than a previously stored fallback key, if present (if the
+appservice doesn't respond with a fallback key then the server uses the stored fallback key instead,
+if known).
```
clokep: Is this enough of a change that it need to be a `v4` of the endpoint, perhaps? (Or a query parameter which defaults to the current behavior?)

turt2live: I think it'd be fine, though open to thoughts. Currently clients appear to pick the "first" key returned, not validating if the extra data is useful to them. In theory this means we can include extra keys with no issues, though if there's a significant need to have it be a dedicated endpoint then let's do that.

We should probably only do this fallback stuff on an `/unstable/org.matrix.msc3983` endpoint though, just in case.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,180 @@
...
+## Returning extra keys
+
+**TODO**: This is probably best as its own MSC.
+
+Independent of the appservice having `/keys/claim` proxied to it, it may be desirable for both the
+fallback and one-time key to be returned. Servers should *always* include the fallback key alongside
+the requested OTKs. When using this proposal's new endpoint, the server should use the fallback key
+from the appservice's response rather than a previously stored fallback key, if present (if the
+appservice doesn't respond with a fallback key then the server uses the stored fallback key instead,
+if known).
```
clokep, line 115: This doesn't define if the appservice should be queried for fallback keys if a OTK is in the database.

proposals/3983-sending-otk-claims-to-appservices.md
clokep: matrix-org/synapse#15462 implements this as a separate endpoint.

proposals/3983-sending-otk-claims-to-appservices.md
turt2live: In lieu of MSC text: the appservice is supposed to be queried, similar to MSC3984 where the server uses what it knows if the appservice didn't provide it.

clokep: Actually I suppose the text in the MSC is enough as is:

> For users under an appservice's explicit namespace, if that user has no unused OTKs (excluding fallback keys) on the homeserver, the homeserver proxies the following APIs to the appservice using the new API described below:

This would *not* match what you just wrote though and would mean it is not possible for an appservice to only provide fallback keys. I'm not sure if that's a good thing or a bad thing though. 🤷 

If we really do want that then I think we want to use different endpoints or something. Otherwise the appservice wouldn't know if we truly want to query for OTKs or only for fallback keys (it could end up returning OTKs which go unused?)

proposals/3983-sending-otk-claims-to-appservices.md
clokep: Another piece of this I realized [during implementation](https://github.com/matrix-org/synapse/pull/15462#discussion_r1174099041) is if we need to also make a change on the federation side.

proposals/3983-sending-otk-claims-to-appservices.md
```diff
@@ -0,0 +1,180 @@
...
+to have a namespace covering the user: instead of guessing or going around to each, we require the
+user to be in an exclusive namespace. This guarantees that there's only one appservice responsible
+for the user.
+
+## Returning extra keys
+
+**TODO**: This is probably best as its own MSC.
+
+Independent of the appservice having `/keys/claim` proxied to it, it may be desirable for both the
+fallback and one-time key to be returned. Servers should *always* include the fallback key alongside
```
clokep, line 111: In addition to both fallback and OTKs, @dkasak suggested that it could be helpful to request multiple keys at once over the C-S API. In particular having 0 or 1 keys of multiple algorithms could be useful.

I [implemented this using](https://github.com/matrix-org/synapse/pull/15468) the same unstable endpoint by accepting a map of `{"<algo>": <count>}` instead of a single algorithm. I'm now seeing that the above has an array of algorithms to pass to the appservice instead though. I guess I should consolidate those to be the same though.

clokep: > I [implemented this using](https://github.com/matrix-org/synapse/pull/15468) the same unstable endpoint by accepting a map of `{"<algo>": <count>}` instead of a single algorithm. I'm now seeing that the above has an array of algorithms to pass to the appservice instead though. I guess I should consolidate those to be the same though.

The Synapse PR was updated to use the same form sent to appservices (i.e. a list of algorithms).

clokep: We'll probably want to add something saying that clients may need to deal with getting back fewer than the requested numbers of keys *and* that servers should protect against users request too many keys at once to avoid exhaustion.

jevolk:
We would like to ask that MSC3983 and MSC3984 be closed as superseded. We implemented both in Tuwunel in July, shipped them in 1.8.3, watched MSC3984 produce a request storm in production in September, and are removing both in matrix-construct/tuwunel#593. Thanks for the careful discussion on this thread in 2023; it is that discussion, more than our incident, that convinced us the ecosystem has moved on.

**The premise has been overtaken:** Both proposals optimise a world in which every bridged user carries its own device and key pool, so that 500 million ghosts imply 25 billion one-time keys. No encrypting bridge works that way. mautrix registers ghosts with `inhibit_login: true`, so a ghost never owns a device, never uploads a key and is left out of Megolm recipient lists; the bridge owns exactly one device, created through appservice login or MSC4190, holding the fifty or so one-time keys the whole bridge needs. Hookshot documents the same design. This thread reached that conclusion in its first week ([tulir, 2023-03-24](https://github.com/matrix-org/matrix-spec-proposals/pull/3983#discussion_r1147514776): "One device with 50 OTKs is perfectly sufficient"), and the reply agreed that "a future MSC would almost certainly figure out a way to give the whole appservice a single device/set of keys" ([2023-03-24](https://github.com/matrix-org/matrix-spec-proposals/pull/3983#discussion_r1147818685)). That MSC now exists: [MSC4350](https://github.com/matrix-org/matrix-spec-proposals/pull/4350) formalises the single bridge identity, and its alternatives section rejects per-ghost identities because they "would require users' devices to encrypt every megolm session for every ghost separately even though all the ghosts are in the same bridge process".

**The spec shipped the other design:** Matrix 1.13 carries MSC2409 with its `receive_ephemeral` opt-in. Matrix 1.17 carries MSC4190, which lets an appservice create and delete devices without `/login` or UIA, and MSC4326, the `device_id` identity assertion split out of MSC3202. MSC4203 and the remaining transaction extensions of MSC3202 deliver to-device events, device-list changes and key counts inside the transaction; mautrix-go, mautrix-python, hookshot and Synapse implement them, and MSC4203's checklist is posted. Under that stack the homeserver answers every `/keys/query` and `/keys/claim` for a bridge from its own tables, as for any other device, and its device-list stream stays authoritative. There is nothing left for a key proxy to save.

```mermaid
sequenceDiagram
    participant C as Client or remote server
    participant H as Homeserver
    participant B as Bridge
    Note over C,B: MSC3983 and MSC3984, where the bridge holds the keys
    C->>H: POST /keys/query for a ghost
    H->>B: POST /_matrix/app/unstable/org.matrix.msc3984/keys/query
    B->>H: 404, no shipping bridge serves the route
    H->>C: the local answer, after the round trip
    Note over C,B: The shipped stack, one bridge device on the homeserver
    B->>H: PUT /devices/DEVICEID (MSC4190), then POST /keys/upload
    H->>B: PUT /transactions with to_device (MSC4203), device_lists and key counts (MSC3202)
    C->>H: POST /keys/query for the bridge bot
    H->>C: answered from its own tables
```

**On their own terms the proposals do not hold up either:**

- *Discovery by error.* Neither defines a registration flag, and both state "There is no version compatibility check: homeservers implementing this functionality would receive an error from appservices which don't support the endpoint." Since no shipping bridge serves the route, every query is that error. A Tuwunel 1.9.0 deployment with four ordinary bridges produced 2,323 of them in twenty minutes, one per ghost queried, all 404. A device-list bug of ours inflated the count and is fixed, but the shape is the proposal's: a homeserver cannot tell an unsupported route from a broken reverse proxy or an outage, so it can neither cache the failure safely nor stop repeating it. Synapse's implementation has the same shape and swallows the 404 without a log line. A registration flag was floated here in 2023 ([link](https://github.com/matrix-org/matrix-spec-proposals/pull/3983#discussion_r1147820815)) and never reached the text; we suggested one on MSC3984 ourselves and withdraw that suggestion in favour of this comment, because there is no capability left to negotiate.
- *Hot-path coupling.* MSC3984 has the homeserver "always" proxy for exclusive users, so every client and every remote server that queries a ghost's keys waits on a bridge round trip, and, in the author's own words, "the appservice effectively needs higher uptime than the homeserver" ([2023-03-29](https://github.com/matrix-org/matrix-spec-proposals/pull/3984#discussion_r1152173391)). reivilibre's "a large cost to pay for what seems like a fairly small benefit" ([2023-03-30](https://github.com/matrix-org/matrix-spec-proposals/pull/3984#discussion_r1152937165)) was answered with "the MSC is less than a week old"; the potential issues section still reads "No major issues."
- *Keys outside the device-list stream.* Keys served by the appservice are never stored, so nothing emits `device_lists.changed` to clients, `m.device_list_update` over federation, or a new `stream_id` on `/user/devices`; a remote server that cached a ghost's device list has no way to learn it changed. Raised in May 2023 ([link](https://github.com/matrix-org/matrix-spec-proposals/pull/3984#discussion_r1200652285)) and still open, like the day-one cross-signing TODO and the "probably best as its own MSC" section on fallback keys.
- *One-time-key accounting leaves the homeserver.* The appservice must guarantee single use, and `device_one_time_keys_count` reads zero for such a device, which turns the universal "upload more when the count is low" loop into a bug the proposal can only warn about.

**Where the implementations stand after three and a half years:** The only appservice-side code anywhere is a hook in the author's own SDK that emits an event and returns `404 M_UNRECOGNIZED` unless a consumer listens, and no consumer does: hookshot, built on that SDK, documents its encryption as MSC3202 plus MSC4203 and answers both routes with the default 404. mautrix-go and mautrix-python contain no reference to either proposal. Synapse's support is two experimental flags, off by default, with no behavioural change since May 2023, while over the same period mautrix's author upstreamed and stabilised MSC4326 and the opt-in-free MSC4190 into Synapse 1.141. Both PR bodies still read "Actual appservice which does something with this: ???". The one operator configuration we found with both flags enabled runs them against mautrix bridges, which cannot answer.

The remaining case in the alternatives section, an encrypted DM with a ghost and no bot present, is real, and the ecosystem answered it the other way: MSC4350 takes bot membership as a given rather than adding a hidden key path, and a ghost that truly needs its own device can have one through MSC4190 with ordinary key uploads, at a scale that never approaches 25 billion keys. Closing both proposals with a pointer to MSC4190, MSC4326, MSC4203, MSC3202 and MSC4350 would let the remaining server-side implementations be removed and would spare the next homeserver our detour.

PR #3984, opened 2023-03-24
MSC3984: Sending key queries to appservices

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/enc-as/key-query/proposals/3984-sending-key-queries-to-appservices.md)

Implemenations:
* bot-sdk: https://github.com/turt2live/matrix-bot-sdk/pull/311
* Synapse: https://github.com/matrix-org/synapse/pull/15321
* Actual appservice which does something with this: ???

**No implementation of cross-signing bits yet.**

### DIFF
--- a/proposals/3984-sending-key-queries-to-appservices.md
+++ b/proposals/3984-sending-key-queries-to-appservices.md
@@ -8,7 +8,7 @@ avoid uploading any keys, it would use this MSC to proxy the `/keys/query` endpo
 *Note*: Readers are encouraged to read MSC3983 for background on this problem space.
 
 Appservices which implement this MSC would generally be expected to *not* upload fallback keys or
-OTKs - implementations can make decisions to optimize for this usecase if they choose.
+one-time keys - implementations can make decisions to optimize for this usecase if they choose.
 
 ## Proposal
 
--- a/proposals/3984-sending-key-queries-to-appservices.md
+++ b/proposals/3984-sending-key-queries-to-appservices.md
@@ -1,0 +1,201 @@
+# MSC3984: Sending key queries to appservices
+
+As mentioned in [MSC3983](https://github.com/matrix-org/matrix-spec-proposals/pull/3983), appservices
+can save some data by accepting OTK claims directly, however they would still need to upload device
+keys for their users. If an appservice wanted to be completely independent from the homeserver and
+avoid uploading any keys, it would use this MSC to proxy the `/keys/query` endpoint alongside MSC3983.
+
+*Note*: Readers are encouraged to read MSC3983 for background on this problem space.
+
+Appservices which implement this MSC would generally be expected to *not* upload fallback keys or
+OTKs - implementations can make decisions to optimize for this usecase if they choose.
+
+## Proposal
+
+Similar to MSC3983, the homeserver *always* proxies the following APIs to the appservice using the
+new API described below:
+* [`/_matrix/client/v3/keys/query`](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3keysquery)
+* [`/_matrix/federation/v1/user/keys/query`](https://spec.matrix.org/v1.6/server-server-api/#post_matrixfederationv1userkeysquery)
+
+**`POST /_matrix/app/v1/keys/query`**
+```jsonc
+// Request
+{
+  "@alice:example.org": ["DEVICEID"],
+  "@bob:example.org": [], // indicates "all device IDs"
+  // ...
+}
+```
+```jsonc
+// Response
+// This is the same response format as `/_matrix/client/v3/keys/query`
+{
+  "device_keys": {
+    "@alice:example.org": {
+      "DEVICEID": {
+        "algorithms": [
+          "m.olm.v1.curve25519-aes-sha2",
+          "m.megolm.v1.aes-sha2"
+        ],
+        "device_id": "DEVICEID",
+        "keys": {
+          "curve25519:DEVICEID": "3C5BFWi2Y8MaVvjM8M22DBmh24PmgR0nPvJOIArzgyI",
+          "ed25519:DEVICEID": "lEuiRJBit0IG6nUf5pUzWTUEsRVVe/HJkoKuEww9ULI"
+        },
+        "signatures": {
+          "@alice:example.org": {
+            "ed25519:DEVICEID": "dSO80A01XiigH3uBiDVx/EjzaoycHcjq9lfQX0uWsqxl2giMIiSPR8a4d291W1ihKJL/a+myXS367WT6NAIcBA"
+          }
+        },
+        "unsigned": {
+          "device_display_name": "Alice's mobile phone"
+        },
+        "user_id": "@alice:example.org"
+      }
+    },
+    // ... also a @bob:example.org record if appropriate
+  },
+  "master_keys": {
+    "@alice:example.org": {
+      "keys": {
+        "ed25519:base64+master+public+key": "base64+master+public+key"
+      },
+      "usage": [
+        "master"
+      ],
+      "user_id": "@alice:example.org"
+    }
+    // ... also a @bob:example.org record if appropriate
+  },
+  "self_signing_keys": {
+    "@alice:example.org": {
+      "keys": {
+        "ed25519:base64+self+signing+public+key": "base64+self+signing+master+public+key"
+      },
+      "signatures": {
+        "@alice:example.org": {
+          "ed25519:base64+master+public+key": "signature+of+self+signing+key"
+        }
+      },
+      "usage": [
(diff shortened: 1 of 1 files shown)


proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+        },
+        "unsigned": {
+          "device_display_name": "Alice's mobile phone"
+        },
+        "user_id": "@alice:example.org"
+      }
+    },
+    // ... also a @bob:example.org record if appropriate
+  },
+  "master_keys": {
```
turt2live, line 58: All of the cross-signing fields probably don't work as easily as we're portraying them here. 

TODO: Figure out how cross-signing works, and represent it accordingly.

proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+    // ... also a @bob:example.org record if appropriate
+  }
+}
+```
+
+*Note*: Like other appservice endpoints, this endpoint should *not* be ratelimited and *does* require
+normal [authentication](https://spec.matrix.org/v1.6/application-service-api/#authorization).
+
+If the appservice responds with an error of any kind (including timeout), the homeserver should treat
+the appservice's response as `{}`. If the appservice indicates that the route is
```
reivilibre, line 111: could we have some explanation for why this is safe to do?

This doesn't seem very convincing to me — it doesn't feel right that this functionality is so unimportant that we can just shrug off errors silently.
Given our pain with encryption bugs and the area this MSC surrounds, it'd be good to have this sort of stuff figured out before it lands in the spec and we have to try and work backwards

turt2live: in absence of having an editor open, the rationale is there's *a lot* which can go wrong so it's on the appservice to either implement proper error handling or not. For example, it may be possible to complete the request for one user/device, but not another: the appservice *should* return a partial response in this case, but it can also decide to throw a normal error instead.

The homeserver *could* retry the request if it wanted to, but ultimately for speed it's relatively important that the request fails fast. The vast majority of errors will be application errors that prevent *any* keys from being served, or the appservice being offline (in which case this API isn't able to be used). 

The MSC acknowledges that this is a high risk area. I'd even go as far as recommending people *don't* use it unless they're willing to take on the implied operational stability requirements.

reivilibre: I think I slightly see what you're saying, but it'd be good to have this written out and explained in the MSC, up to the point of what the actual impact on clients/users this is.

If the user triggers this to be queried and the AS is down, the response is assumed to be `{}` and then .... what? What happens? Does anything break?

I'm not super enthusiastic about having this, even with as much discouragement as possible, because eventually it's probably going to come down to one of us to dig into this and find out why things are buggy. However, if we must have it, let us at least understand what happens when it goes wrong.

turt2live: The short version is encryption as a whole breaks down: when clients don't know the device keys, they can't send to those devices. The appservice effectively needs higher uptime than the homeserver in this case.

It's similar to what would happen if you contact a new user while that user's homeserver is down (regardless of appservices being involved).

reivilibre: Breakdown of encryption sounds like it deserves a mention in the 'major problems' section.

To be frank, I'm not keen — I don't really like the idea of having to debug this when it goes wrong. On balance, it seems like a large cost to pay for what seems like a fairly small benefit (if I understand the MSC properly). Is there really no alternative here?

proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+avoid uploading any keys, it would use this MSC to proxy the `/keys/query` endpoint alongside MSC3983.
+
+*Note*: Readers are encouraged to read MSC3983 for background on this problem space.
+
+Appservices which implement this MSC would generally be expected to *not* upload fallback keys or
+one-time keys - implementations can make decisions to optimize for this usecase if they choose.
+
+## Proposal
+
+Similar to MSC3983, the homeserver *always* proxies the following APIs for an appservice's exclusive users
```
clokep, line 15: @reivilibre pointed out during the Synapse review that we might want to shout a bit louder that this only applies to *exclusive* users.

clokep: (And maybe some rationale as to why bad things would happen if you allow non-exclusive users?)

proposals/3984-sending-key-queries-to-appservices.md
turt2live: I suppose the potential issues section should say "No major issues not already mentioned" (with the assumption that this thread's context lands in the MSC).

As far as I'm aware, there's no viable alternative, though the MSC is less than a week old.

proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+to the appservice using the new API described below:
+* [`/_matrix/client/v3/keys/query`](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3keysquery)
+* [`/_matrix/federation/v1/user/keys/query`](https://spec.matrix.org/v1.6/server-server-api/#post_matrixfederationv1userkeysquery)
```
clokep, line 18: There's also a [`/user/devices`](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1userdevicesuserid) query which returns device information in a slightly different format (and some other information that we don't talk about in this MSC, I think).

Synapse uses this to "sync device lists" when it does not know the devices of a user (or if it thinks devices are out of sync). In order to properly pipe this information back and forth I think we need to somehow proxy this to appservices too.

In matrix-org/synapse#15539 I was able to "proxy" it by actually making the `/keys/query` request to the appservice and reformatting the returned data. I'm not sure this is the ideal way though.

proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+  "self_signing_keys": {
+    "@alice:example.org": { /* ... */  } // as per homeserver's existing knowledge
+  },
+  "user_signing_keys": {
+    "@alice:example.org": { /* ... */  } // as per homeserver's existing knowledge
+  }
+}
+```
+
+## Potential issues
```
clokep, line 178: This doesn't provide a way for the appservice to communicate to the homeserver that there's a new device (or that a device has changed). This would need to be used to apply the [device management](https://spec.matrix.org/v1.6/server-server-api/#device-management) section of the federation spec so that remote servers actually pull new device information.

(This is akin to what needs to happen when `/keys/upload` is called on the Client-Server API.)

proposals/3984-sending-key-queries-to-appservices.md
clokep: See also matrix-org/matrix-spec#1539

proposals/3984-sending-key-queries-to-appservices.md
clokep: I'm not sure if it is fully as speced or not, but for synapse at least it seems you can do a `PUT /_matrix/client/v3/devices/<device ID>` to force Synapse to send a `m.device_list_update` EDU for that device.

proposals/3984-sending-key-queries-to-appservices.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Security considerations
+
+No major considerations.
+
+## Unstable prefix
+
+While this MSC is not considered stable, implementations should use
+`/_matrix/app/unstable/org.matrix.msc3984/keys/query` as the endpoint instead. There is no version
+compatibility check: homeservers implementing this functionality would receive an error from appservices
```
jevolk, line 194: The MSC says there is "no version compatibility check" and suggests backing off after an unknown-route error. That signal is not reliable enough for capability discovery: an unsupported endpoint, reverse-proxy routing error, transient bridge outage, application error, and timeout can be indistinguishable at the homeserver. Caching those failures as "unsupported" can disable a capable appservice, while not caching them repeats the request and log amplification on every key query.

Could MSC3984 require an explicit, default-false `org.matrix.msc3984: true` opt-in in `registration.yaml`, with the homeserver sending key queries only for exclusively owned users when the appservice advertises support? This follows the registration capability pattern used by MSC3202. Endpoint errors could then remain operational failures rather than feature negotiation. If a registration field is unsuitable, the protocol still seems to need some positive capability signal before stabilization.

jevolk (changes requested):

PR #3266, opened 2021-07-04
MSC3266: Room summary API

deepbluev7:
[Rendered.](https://github.com/deepbluev7/matrix-doc/blob/room-summaries/proposals/3266-room-summary.md)

Somewhat related to [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) this provides an API to get a summary for a specific room either from the local server or over federation.

Useful for a few cases, where you need to show a summary for a room like matrix.to, traveler bots, showing spaces, lightweight clients, etc.

~~**Open design questions looking for feedback**: see this comment chain: https://github.com/matrix-org/matrix-doc/pull/3266#discussion_r686213648~~ resolved as of 2021/10/06

Implementations:
- matrix.to: [matrix-org/matrix.to#219](https://github.com/matrix-org/matrix.to/pull/219)
- Synapse: [matrix-org/synapse#10394](https://github.com/matrix-org/synapse/pull/10394) (Only CS API) and [matrix-org/synapse#11507](https://github.com/matrix-org/synapse/pull/11507) (for the federation bits). The `allowed_room_ids` part is implemented here: https://github.com/deepbluev7/synapse/commit/37f4253bda4942adc6356e9ed73dc40582764c9b. 
- Element Android: https://github.com/vector-im/element-android/pull/3986
- Element Web (for member counts on space invites): https://github.com/matrix-org/matrix-react-sdk/pull/6982 and https://github.com/matrix-org/matrix-js-sdk/pull/1988
- Nheko (to have a prettier join room preview): https://github.com/Nheko-Reborn/nheko/commit/9d8d6b4bcaec0994776bf91a556a7e27cb862c30
- matrix-rust-sdk: https://github.com/matrix-org/matrix-rust-sdk/pull/3339
- shields.io: https://shields.io/badges/matrix, https://github.com/badges/shields/pull/10782
- Matrix Room Search: https://gitlab.com/etke.cc/mrs/api/-/commit/bc4a4a649cfbf98d850ea0248c78afc5d552e1be

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>

----

SCT stuff:

Shepherd: @richvdh
[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3266#issuecomment-2579239313)

**Checklist**: https://github.com/matrix-org/matrix-spec-proposals/pull/3266#issuecomment-2506672133

PR #3985, opened 2023-03-27
MSC: Break-out rooms

SimonBrandner:
Done with a community :tophat: on


proposals/3985-break-out.md
```diff
@@ -0,0 +1,66 @@
+# MSC3985: Break-out rooms
+
+VoIP applications such as Zoom or MS Teams support break-out rooms - a mechanism
+for splitting users into multiple calls. This proposal suggests a mechanism for
+break-out rooms in Matrix which would not be limited to VoIP use-cases.
```
deepbluev7, line 5: This sounds really useful for me personally to split out some discussions. I think what I would still need is a way to force users to only talk in the breakout room temporarily. Do you have some easy ideas for that, that might be easy to cover in a follow up MSC? :)

SimonBrandner: Hmm, that's a good point. I don't think I have anything in mind. The best I can think of atm is just temporarily changing permissions

proposals/3985-break-out.md
```diff
@@ -0,0 +1,74 @@
...
+```json5
+{
+    "type": "m.breakout",
+    "state_key": "cvsiu2813",
+    "content": {
+        "m.breakout": {
+            "!roomId1": {
+                "via": ["example.org", "other.example.org"],
+                "users": ["userId1", "userId2"]
+            },
```
uhoreg, line 22: Would it be helpful to add some more information about the room?  Such as a [stripped state](https://spec.matrix.org/unstable/client-server-api/#stripped-state) event list, so that users can more easily choose which room to join?

proposals/3985-break-out.md
```diff
@@ -0,0 +1,74 @@
...
+their client's choosing. A few common examples are:
+
+- The "child" room has a `join_rule` of `restricted` indicating that only the
+  members of the "parent" room can join.
+- The "child" room has `join_rule` if `invite` indicating that only the invited
+  users can join. In this case, the room information in the `m.breakout` event
+  should include the list of users who are suggested to join the given room.
+- The "child" room has a `join_rule` of `public` indicating anyone can join.
+  This would usually happen if the "parent" room was also public.
+
```
uhoreg, line 52: It might be good to say something about how breakout rooms are ended.

proposals/3985-break-out.md
MTRNord: Having a topic or a name might indeed be helpful. In University, we had them for example named by Groups. "Group 1", "Group 2"... Something like that might be nice to have.

clokep: I wonder if storing that in the breakout rooms and then using something like the `/hierarchy` API would alleviate duplication...

SimonBrandner: I was thinking we'd use [MSC3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266) to avoid the duplication and the info getting out of sync

Using the `/hierarchy` is definitely an interesting idea... I wonder if `m.space.child` should be used if we'd go that route but I don't really like how it would mean the info is spread over multiple events

clokep: Could also modify that API to return different events or something for different room types 🤷 There's a lot of ways to cut it...

SimonBrandner: I think we can just use the room name and topic and so don't have to modify the API?

proposals/3985-break-out.md
```diff
@@ -0,0 +1,88 @@
...
+`m.breakout` event:
+
+```json5
+{
+    "type": "m.breakout",
+    "state_key": "",
+    "content": {
+        "m.breakout": {
+            "!roomId1": {
+                "via": ["example.org", "other.example.org"],
```
MTRNord, line 20: I guess this technically isn't the fault of the MSC itself but the via process in itself how it currently is handled, but how would this get updated if the via servers get stale? Are we expecting the event to be short-lived enough for this to be no concern? Or should they get updated if some condition is met (like last person on an HS left)?

proposals/3985-break-out.md
```diff
@@ -0,0 +1,88 @@
...
+  members of the "parent" room can join.
+- The "child" room has `join_rule` if `invite` indicating that only the invited
+  users can join. In this case, the room information in the `m.breakout` event
+  should include the list of users who are suggested to join the given room.
+- The "child" room has a `join_rule` of `public` indicating anyone can join.
+  This would usually happen if the "parent" room was also public.
+
+### VoIP use-case
+
+If a user is participating in a call when the event is sent, they should either
```
MTRNord, line 61: Mostly a UX question that might be beyond the scope of MSC process, but should this be semi-automatic (like Zoom does it) or would the expected behavior be that the user has to reconfirm joining it? (This comment is not a concern against the MSC. So imho it should not be considered as part of the process. It's just an implementation question)

SimonBrandner: I feel like this could be client-specific

proposals/3985-break-out.md
SimonBrandner: I don't think the space MSC solves this either?

MTRNord: Yeah, I don't think it does. But it's just another area where this "bug" happens. And in theory, this MSC still has the chance to do something about it. :) Hence, I was pointing it out. Though of course it may be better to solve that in another dedicated MSC? Not sure what's the best tbh.

proposals/3985-break-out.md
```diff
@@ -0,0 +1,88 @@
...
+start a call or join an existing call in the "child" room, if there is one.
+
+### Security
+
+In public rooms, special precautions should be taken. In most cases, regular
+users shouldn't be allowed to send `m.breakout` events. But if they are, the
+client should warn the user, if the event is coming from an unknown user (one
+with whom we don't have a DM) who is not an admin/moderator either.
+
+## Potential issues
```
robintown, line 71: If you're in a large call and want to create, say, 10 breakout rooms, rate limiting could become a problem. A separate MSC could make it possible to bulk-create rooms, perhaps?

SimonBrandner: Hmm, good point...

clokep: I had written MSC3192 while doing the spaces stuff....

clokep: Oh, but you're concerned about *creating* the rooms... yeah I'm not sure how to differentiate it from spam. 😢

proposals/3985-break-out.md
```diff
@@ -0,0 +1,88 @@
...
+"parent" no longer links to the "children".
+
+### "Child" rooms' `join_rule`s
+
+The "child" rooms may have `join_rules` of the `m.breakout` event's creator's or
+their client's choosing. A few common examples are:
+
+- The "child" room has a `join_rule` of `restricted` indicating that only the
+  members of the "parent" room can join.
+- The "child" room has `join_rule` if `invite` indicating that only the invited
```
Ugzuzg, line 53:
```suggestion
- The "child" room has `join_rule` of `invite` indicating that only the invited
```

PR #3991, opened 2023-04-04
MSCXXXX: Power level up! Taking the room to new heights

MadLittleMods:
MSCXXXX: Power level up! Taking the room to new heights

Rendered (TODO)

Server implementations:

 - *Implementations needed*

### DIFF
--- a/proposals/0000-increase-power-level.md
+++ b/proposals/0000-increase-power-level.md
@@ -1,0 +1,111 @@
+# MSC0000: Power level up! Taking the room to new heights
+
+Once a room is created, the highest power level is set in stone. Even if you're the only
+admin in the room, if you try to raise your own power level above the initial `100`, it
+just throws a `403 Forbidden` error:
+
+`PUT /_matrix/client/r0/rooms/{roomId}/state/m.room.power_levels` -> `403 Forbidden`
+```json
+{
+    "errcode": "M_FORBIDDEN",
+    "error": "You don't have permission to add ops level greater than your own"
+}
+```
+
+This is dictated by the [authorization rules (auth
+rules)](https://spec.matrix.org/v1.5/rooms/v10/#authorization-rules) around
+`m.room.power_levels` in a room:
+
+>  9. If type is `m.room.power_levels`:
+>     [...]
+> 
+>      8. For each entry being changed in, or removed from, the `users` property, other than the `sender`’s own entry:
+>          1. If the current value is greater than or equal to the `sender`’s current power level, reject.
+>      9. For each entry being added to, or changed in, the `users` property:
+>          1. If the new value is greater than the `sender`’s current power level, reject.
+
+It's possible to have power levels greater than the default `100` (Admin) power level
+but this has to be specified at the time of room creation.
+
+Not being able to adjust the max power level of `users` after the fact means that any
+mistake is baked into the room forever and requires a room upgrade to rectify the
+situation. Sometimes, more flexibility is needed in the power level ranges and this only
+becomes obvious with hindsight.
+
+For example with the [Gitter
+migration](https://blog.gitter.im/2023/02/13/gitter-has-fully-migrated-to-matrix/), we
+synced all Gitter room admins over as users with a power level of `90` and set the power
+level to do any action at `90` so the bridge bot user could maintain a higher power
+level than the rest of the users while giving the room admins autonomy over everything.
+Then later after the migration, we wanted to clean up the power levels and grant room
+admins the true `100` power level to avoid clients labeling `90` as a "Moderator"
+instead of "Admin". But discovered this was impossible to get right because we couldn't
+raise the power level of the bridge bot above `100` to still maintain a higher power
+level. This condundrum is tracked by https://gitlab.com/gitterHQ/gitter.im/-/issues/4
+
+Something like [MSC3915: Owner power
+level](https://github.com/matrix-org/matrix-spec-proposals/pull/3915) could help in a
+situation like this to formalize the position of owner but doesn't help with the
+flexibility of an existing room. For example, imagine this MSC lands before MSC3915, it
+would be nice to just upgrade your room power levels to reflect the new owners role. And
+going beyond MSC3915, it can be useful to set a user/bot above `150` at a later date.
+Imagine wanting a central company bot to maintain control of every room at power level
+`200`. It would be nice to just update your room power levels to achieve this than have
+to upgrade every room.
+
+
+
+## Proposal
+
+This MSC proposes updating the room event auth rules to allow for raising the `sender`'s
+own `users` power level above the current max power level as long as you update all
+others at the same level to the new max level.
+
+This means that if you're a solo admin in the room, you can arbitrarily raise your own
+power level however you want.
+
+If there are multiple admins in the room, then you must raise all other admins to the
+new max power level.
+
+Propsed new auth rule language:
+
+>  9. If type is `m.room.power_levels`:
+>     [...]
+> 
+>      8. For each entry being removed from the `users` property, other than the `sender`’s own entry:
+>          1. If the current value is greater than or equal to the `sender`’s current power level, reject.
+>      9. For each entry being added to, or changed in, the `users` property:
+>          1. If the new value is greater than the `sender`’s current power level and the `sender` *doesn't have* the highest power level in the room, reject.
+>          1. If the new value is greater than the `sender`’s current power level, the `sender` *has* the highest power level in the room, but doesn't raise everyone else with the current highest power level to the new value, reject.
+
+Because this MSC changes the authorization rules of a room, it requires a new room
+version to ensure all participating servers are authorizing events and state
+consistently.
+
+
+## Potential issues
+
+*None surmised so far*
+
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,111 @@
...
+
+Propsed new auth rule language:
+
+>  9. If type is `m.room.power_levels`:
+>     [...]
+> 
+>      8. For each entry being removed from the `users` property, other than the `sender`’s own entry:
+>          1. If the current value is greater than or equal to the `sender`’s current power level, reject.
+>      9. For each entry being added to, or changed in, the `users` property:
+>          1. If the new value is greater than the `sender`’s current power level and the `sender` *doesn't have* the highest power level in the room, reject.
```
MadLittleMods, line 78: Is "highest power level" ambiguous? For example, I'm trying to say if the `sender` has `100` along with other admins in the room or is the only admin

proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,111 @@
+# MSC3991: Power level up! Taking the room to new heights
```
MadLittleMods, line 1: Any better puns?

 - "Power levels to infinity and beyond"
 - "Breaking the power level ceiling"
 - "Turning power levels up to 11"

Phrases to work from:

 - "Power up"
 - "Fight the power"
 - "Raise the roof"
 - "Seize power"
 - "Power leveling"
 - "Rising power levels"

proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,111 @@
...
+flexibility of an existing room. For example, imagine this MSC lands before MSC3915, it
+would be nice to just upgrade your room power levels to reflect the new owners role. And
+going beyond MSC3915, it can be useful to set a user/bot above `150` at a later date.
+Imagine wanting a central company bot to maintain control of every room at power level
+`200`. It would be nice to just update your room power levels to achieve this than have
+to upgrade every room.
+
+
+
+## Proposal
```
FSG-Cat, line 58: Could someone please explain to me where this MSC dismisses the concern of that this introdudes the exact same type of problems like the Moderators can demote each other MSC. Like under this MSC admins can fork the room to raise the PL above each other in a race to the top. 

Like as far as i am concerned this smells to me like a MSC that is fundementally broken and CAN NOT be reworked to work and should have been killed in pre publication review. But please prove me wrong.

proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+## Alternatives
+
+Room upgrades allow for creating a new room where you can initially specify
+`m.room.power_levels` as desired. There is nothing restricting the integer range that
+`users` field of `m.room.power_levels` so all of the same end results can be achieved
+this way. This has all of the flexibility downsides mentioned for existing rooms in the
+intro/context paragraphs above though.
+
```
deepbluev7, line 98: You could also fix how clients calculate who is an admin or a moderator instead of changing the auth rules. One way to do that could be by introducing roles, that map a name to a power level, which would be only aesthetic. (A proper roles system would also allow non linear power levels, but that isn't needed here).

Alternatively one could spec how to figure out what is the admin PL and what is the mod PL. For example my client does that by considering people with the permission to change powerlevels to be admins and users with redaction permissions are moderators. Hardcoding random integers just doesn't work in the rooms I am in and I would strongly discourage clients from doing that.

proposals/3991-increase-power-level.md
MadLittleMods: @FSG-Cat With this MSC, you can't raise your power level above the others. You can only raise your power level alongside everyone else.

proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,111 @@
...
+`200`. It would be nice to just update your room power levels to achieve this than have
+to upgrade every room.
+
+
+
+## Proposal
+
+This MSC proposes updating the room event auth rules to allow for raising the `sender`'s
+own `users` power level above the current max power level as long as you update all
+others at the same level to the new max level.
```
deepbluev7, line 62: At the same powerlevel or higher or if they have the highest PL.

You also want to check, that there is no event, that requires a higher PL than any user entry. You also need to check that users_default is less than or equal the highest user PL, same with all the other possible powerlevels (redact, notify, etc).

proposals/3991-increase-power-level.md
MadLittleMods: @deepbluev7 👍 This is a good alternative to note for the Gitter situation. Related to [MSC3949](https://github.com/matrix-org/matrix-spec-proposals/pull/3949). I can add a note to the MSC about this.

I think this MSC makes sense on top of those kind of changes as well!

deepbluev7: Possibly, but auth changes are always a bit scary and I would prefer an actual roles system to assign roles, if we already touch the auth rules again. Specifying how clients recognize admins and mods would (possibly) work without a room version too, which might be better for existing rooms as well :)

But if there is enough merit for this MSC apart from the one Gitter example, then it makes sense to go this route, since it is quite a bit simpler than a full role system.

proposals/3991-increase-power-level.md
FSG-Cat: I was going to edit in this major fix because yes i overlooked this as i read this early in the morning and then now when i re read it i see that you adressed the exploit i was thinking about. Thanks for pointing this out.

proposals/3991-increase-power-level.md
MadLittleMods: > At the same powerlevel or higher or if they have the highest PL.

Can you propose a suggestion block? I don't think this is accurate or I can't quite pick it apart to see how it applies. But do think that paragraph could be clearer.

> You also want to check, that there is no event, that requires a higher PL than any user entry. You also need to check that users_default is less than or equal the highest user PL, same with all the other possible powerlevels (redact, notify, etc).

I assume you're trying to plug the hole where if you have an admin and moderator in a room and the admin removes their power level, the moderator could boost themselves to an admin or above?

Keep in mind, that only the top-most power level people can raise their power level further. If there is an admin and moderator in the room, the moderator can't go above their current power level. Only the admin can.

I think your language is an option but the other side also opens up a different possibility. It kinda relates to https://github.com/matrix-org/matrix-spec/issues/165. If we leave it as-is, people would be able to self-service and solve this problem:

> Perhaps, if there is no administrator left in the room, the administrator's rights should be given to the user (one or more) with the highest rank?
>
> *-- @slipeer, https://github.com/matrix-org/matrix-spec/issues/165#issuecomment-1055739882*

proposals/3991-increase-power-level.md
MadLittleMods: Noted this alternative in https://github.com/matrix-org/matrix-spec-proposals/pull/3991/commits/dec22ac804c469e2a562c38acc86183f3ae627fd

Thanks @deepbluev7 🙂

progval:
If you have a room with PL:

```json5
{
    "content": {
        // ...
        "users": {
            "alice:example.org": 100,
            "bob:example.com": 100,
        }
    },
    "sender": "anyone:example.net"
}
```

then simultaneously, alice replaces it with:

```json5
{
    "content": {
        // ...
        "users": {
            "alice:example.org": 101,
            "bob:example.com": 101,
        }
    },
    "sender": "alice:example.org"
}
```

and bob with:

```json5
{
    "content": {
        // ...
        "users": {
            "alice:example.org": 102,
            "bob:example.com": 102,
        }
    },
    "sender": "bob:example.com"
}
```

then after they each process each other's event, wouldn't they get stuck in inconsistent states? (alice seeing 102 and bob as 101)

MadLittleMods:
@progval Please re-comment on the diff instead so we can thread

proposals/3991-increase-power-level.md
```diff
@@ -0,0 +1,124 @@
...
+to upgrade every room.
+
+
+## Proposal
+
+This MSC proposes updating the room event auth rules to allow for raising the `sender`'s
+own `users` power level above the current max power level as long as you update all
+others at the same level to the new max level.
+
+This means that if you're a solo admin in the room, you can arbitrarily raise your own
```
Half-Shot, line 63: I'd be a bit careful with using the term "admin" here (and instead say PL50/PL100 or PLMax). Admin means PL100 in Element yes, but that's a dirty hack in the client and not represented by all of them.

MadLittleMods: "Admin" is just being used as the understandable example here. The situation can occur at any power level (whatever is the highest in the room) so "PLMax" is kinda the only alternative. Not sure it would make it more clear but it would be more correct. More trying to convey the idea in these sentences.

MadLittleMods: We could add some preface sentence  like [MSC3993](https://github.com/matrix-org/matrix-spec-proposals/pull/3993),

> For this proposal a "room admin" is a user with the highest power level in a room. For the use case this MSC addresses, a room must have two or more room admins.

Adapting that to say perhaps:

> In this proposal we use the term "admin" as stand-in for the user(s) with the highest power level in a room.

PR #3995, opened 2023-04-10
Linearized Matrix

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/linearized-matrix/proposals/3995-linearized-matrix.md) (recommended)

## ⚠⚠ This is not considered final and might change dramatically. This MSC exists to show what we're thinking about, but is not representative of the final proposal.


proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+  respond usefully to these endpoints), the following are also copied to Linearized Matrix's transport:
+
+  * The [backfill API](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1backfillroomid) is supported.
+  * The [get event API](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1eventeventid) is supported.
+  * The [get state API](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1stateroomid) and
+    [state IDs API](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1state_idsroomid) are supported.
+  * The [event auth API](https://spec.matrix.org/v1.6/server-server-api/#get_matrixfederationv1event_authroomideventid) is
+    supported.
+
+* [Joins](https://spec.matrix.org/v1.6/server-server-api/#joining-rooms), [invites](https://spec.matrix.org/v1.6/server-server-api/#inviting-to-a-room),
```
turt2live, line 417: Likely a future MSC, but we need to define how a LM participant server discovers the hub server without an invite (ie: joining a public room)

turt2live, line 417: Invites for `m.linearized` servers need to route through the current room's hub

turt2live (commented):
various comments from ad-hoc review

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+The hub server aggressively appends to its linked list, which in turn would be represented by a single
+branch of the DAG.
+
+This leads to the following roles for servers:
+
+* **DAG-capable**: Every homeserver which already exists today. These servers do not need to change how
+  they persist rooms (they continue using a DAG).
+
+* **Hub**: A server which accesses the room using a linked list on behalf of other *participant* servers.
+  This server can be DAG-capable, but does not need to be.
```
Cadair, line 61: Having given this a quick read, this line really confused me from a language perspective. I see that a hub server may or may not also be a DAG-capable server, but for clarity I would keep the roles separate and say something like:


```suggestion
  A single server can be both a hub server and a DAG-capable server, but does not need to be.
```

turt2live: Sorry, I'm struggling to parse the concern and suggestion here. The text is already keeping the roles separate and being clear which role ("DAG-capable") is optional? The suggestion feels like it's saying the hub server doesn't need to be a hub server, at least.

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+  * Otherwise, allow.
+
+Note that a feature of these auth rules is being able to change the room hub, though that requires both the
+current hub server and the new server to be online for the signing (or at least online enough to both sign
+the event).
+
+The [auth events selection algorithm](https://spec.matrix.org/v1.6/server-server-api/#auth-events-selection)
+is additionally changed. Currently the selection algorithm is roughly the same for all room versions, but the
+changes here will bring it formally into the room version specification. A new condition is added to state that
+any event with a `hub_server` top level field must reference the `m.room.hub` state event (if any).
```
turt2live, line 286: also the `m.room.hub` should reference the previous `m.room.hub` event, if any.

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+## This MSC is going through extremely rapid iteration
+
+As is implied throughout, this MSC is considered a work in progress and is subject to change wildly without
+notice. Implementations are welcome, though the author notes that due to the highly dynamic nature of this
+MSC's content it may be a frustrating experience. This notice will be removed when the MSC is stable enough
+for a wider range of (unstable) implementations.
+
+## Proposal
+
+### Room structure
```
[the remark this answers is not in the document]
turt2live: I think the heuristics in particular need work, but the general aim is to ensure there's at least 1 hub in the room for the participant servers. 

Part of the idea behind LM is you're theoretically able to implement a participant-only homeserver, though maybe we give up on that idea and instead require implementations to act as hub and participant, eliminating the concerns of not having a single hub (the fallback would be "become a hub" rather than "find a hub algorithmically"). However, that also takes us a step closer to just implementing state res considering that would be the only thing farmed out to a DAG-capable server, and all the LM servers would be federating through DAG-level fanout.

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+
+* `depth` is no longer protected from redaction.
+* `prev_state` is no longer protected from redaction, as per [MSC2176](https://github.com/matrix-org/matrix-spec-proposals/pull/2176).
+* `origin` is no longer protected from redaction, as per [MSC3989](https://github.com/matrix-org/matrix-spec-proposals/pull/3989).
+* `membership` (at the top level) is no longer protected from redaction, as per MSC2176.
+* `m.room.create` no longer allows `creator` in the `content`, as per [MSC2175](https://github.com/matrix-org/matrix-spec-proposals/pull/2175),
+  but does preserve all other fields as per MSC2176.
+* `hub_server` at the top level is newly protected.
+
+### Auth rule changes
```
turt2live, line 244: We should also talk about how rejection works. Currently in the I-D we assume that we can just not append the event, but is that realistic?

https://spec.matrix.org/v1.6/server-server-api/#rejection

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
...
+
+At the room level, Matrix currently operates with a [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph)
+(DAG) to persist messages. Servers wishing to contribute to the DAG broadcast their events (nodes) to
+every other server participating in the room, which all then run the exact same set of rules on the
+received message to ensure it is legal before appending it to their local copy of the DAG. Servers
+missing events can self-heal by contacting other servers, validating the response, and inserting the
+events appropriately.
+
+This model provides fully decentralized communications as a feature to Matrix, a feature we'd like to keep,
+however is complex when aiming to adopt Matrix as an interoperable chat protocol, such as with the emergence
```
joepie91, line 11: This sections seems *extremely* incomplete. The MSC implies that the current model is too complex for adopting as an interoperable protocol; but it fails to specify *why* that is. Where does this complexity lie, in what ways does it pose a problem for this usecase, and what makes this a problem of the protocol rather than of the implementers?

Given the context, this seems like a crucial thing to specify, and it's odd that a proposal (even if a draft) has been written *without* specifying that upfront; it seems like putting the cart before the horse. Government regulation is often an opportunity to force unwilling parties to implement something that they would rather not, but that would be beneficial to society as a whole, precisely *because* they are enforced regulations and you do not need to rely on economic motivations.

I feel like such a situation applies here as well; the DAG model of Matrix is a crucial part of its resiliency (towards both technical outages and misbehaving participant servers), and while it *is* complex to implement correctly, I feel that it is nevertheless important to maintain these properties. Therefore, if the proposal is to weaken these properties, then it should really be supported by a considerable stronger rationale than the current implied "large tech corps don't like it".

proposals/3995-linearized-matrix.md
```diff
@@ -0,0 +1,530 @@
+# MSC3995: Linearized Matrix
```
turt2live, line 1: Readers: This MSC is based on the 01 and some of the 02 drafts within the IETF. It does not reflect the 03 draft.

PR #3952, opened 2023-01-06
MSC3952: Intentional Mentions

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/intentional-mentions/proposals/3952-intentional-mentions.md)

Fixes matrix-org/matrix-spec#15
Fixes matrix-org/matrix-spec#16
Fixes matrix-org/matrix-spec#353
Fixes matrix-org/matrix-spec#550

Implementations:

* Synapse: matrix-org/synapse#14823
* Element Web: matrix-org/matrix-js-sdk#3092 and matrix-org/matrix-react-sdk#9983.

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3952#issuecomment-1477291795)

PR #3996, opened 2023-04-13
MSC3996: Encrypted mentions-only rooms.

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/encrypted-mentions-only-rooms/proposals/3996-encrypted-mentions-only-rooms.md)

### DIFF
--- a/proposals/3996-encrypted-mentions-only-rooms.md
+++ b/proposals/3996-encrypted-mentions-only-rooms.md
@@ -1,0 +1,189 @@
+# MSC3996: Encrypted mentions-only rooms
+
+It is currently not possible for mobile clients to properly implement a mentions-only
+room.
+
+Currently, every new event in an encrypted room might be pushed to mobile clients
+due to the [`.m.rule.encrypted` default underride rule](https://spec.matrix.org/v1.6/client-server-api/#default-underride-rules).
+
+However, a room can be configured to be mentions-only by creating a
+[room-specific push rule](https://spec.matrix.org/v1.6/client-server-api/#push-rules)
+with the `id` property set to the room ID & `actions` set to do nothing.[^1] Since
+this is evaluated *before* the `.m.rule.encrypted` rule (almost)
+**no events get pushed for a mentions-only room**.
+
+Additionally, a room can be configured to be muted by creating the earliest
+[override push rule](https://spec.matrix.org/v1.6/client-server-api/#push-rules)
+possible which matches the room ID & has `actions` set to do nothing[^2], e.g.:
+
+```json
+{
+    "rule_id" : "!abcdef:example.com",
+    "conditions" : [
+       {
+          "key" : "room_id",
+          "kind" : "event_match",
+          "pattern" : "!abcdef:example.com"
+       }
+    ],
+    "default" : false,
+    "enabled" : true,
+    "actions" : []
+}
+```
+
+## Proposal
+
+A new top-level property (`m.has_mentions`) is defined which contains a boolean
+value. It is sent in cleartext (i.e. it is not encrypted) and is set to `true` if
+the event mentions another user or the room per
+[MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952).[^3]
+
+A new push rule is added after the `.m.rule.is_room_mention` push rule to match
+encrypted mentions:
+
+```json
+{
+    "rule_id": ".m.rule.is_encrypted_mention",
+    "default": true,
+    "enabled": true,
+    "conditions": [
+        {
+            "kind": "event_property_is",
+            "key": "content.m\\.has_mentions",
+            "value": true
+        },
+        {
+            "kind": "event_property_is",
+            "key": "type",
+            "value": "m.room.encrypted"
+        }
+    ],
+    "actions": ["notify"]
+}
+```
+
+When this push rule matches then it would push the event to all users, similar to
+how the [`.m.rule.encrypted` default underride rule](https://spec.matrix.org/v1.6/client-server-api/#default-underride-rules)
+push rule works, but with the added context that *some clients will probably care*.
+Clients would [decrypt the event and re-run push rules](https://spec.matrix.org/unstable/client-server-api/#receiving-notifications)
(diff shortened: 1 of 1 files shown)


proposals/3996-encrypted-mentions-only-rooms.md
```diff
@@ -0,0 +1,189 @@
...
+4. A push notification which the client would "upgrade" to a highlight notification
+   if the (decrypted) event mentions the user.
+
+The overall tradeoff is that clients will receive extra pushes some of the time
+(which won't matter), but help hide the metadata of who, in particular, is being
+mentioned.
+
+If the decrypted ciphertext contains a `m.has_mentions` property it should be ignored.
+
+## Potential issues
```
clokep, line 88: This is missing a section about abuse potential -- what happens if someone just says every message has a mentions on it... wasting the bandwidth, power, and CPU of others in the room?

proposals/3996-encrypted-mentions-only-rooms.md
```diff
@@ -0,0 +1,189 @@
...
+
+```json
+{
+    "rule_id": ".m.rule.is_encrypted_mention",
+    "default": true,
+    "enabled": true,
+    "conditions": [
+        {
+            "kind": "event_property_is",
+            "key": "content.m\\.has_mentions",
```
anoadragon453, line 53: Shouldn't this be:

```suggestion
            "key": "content.m\.has_mentions",
```

If we're trying to match on `content` -> `m.has_mentions`?

proposals/3996-encrypted-mentions-only-rooms.md
anoadragon453: It would be nice if clients could surface this field being set (to moderators/admins only?) so the abuse wouldn't be so invisible.

anoadragon453 (commented):
This seems like a happy medium between metadata privacy and resource usage, and a step up over today's situation.

proposals/3996-encrypted-mentions-only-rooms.md
clokep: The field is unencrypted so yeah I agree -- this is mostly up to clients or moderation tooling exposing some aggregate information on this? I think I just need to add some text. 😄

proposals/3996-encrypted-mentions-only-rooms.md
```diff
@@ -0,0 +1,189 @@
...
+            "key": "content.m\.has_mentions",
+            "value": true
+        },
+        {
+            "kind": "event_property_is",
+            "key": "type",
+            "value": "m.room.encrypted"
+        }
+    ],
+    "actions": ["notify"]
```
clokep, line 62: @giomfo had wondered if a "push, but don't increase unread count" action would be more appropriate here. I think that we don't need that here, but it *could* be useful?

I'd be curious of thoughts from other client developers!

clokep: @erikjohnston [said](https://github.com/matrix-org/matrix-spec-proposals/pull/3952#discussion_r1169971606) on MSC3952:

> I think its worth noting that this means that servers (still) won't be able to differentiate between notifications and mentions. This mostly affects how often we push to devices, but also means that we can't make `/notifications?only=highlight` API be useful again

I think this applies to this conversation as well. If we added a `highlight` tweak to this rule then homeservers could push the event (and return it it as part of `/notifications?only=highlight`) and allow clients to reparse push rules and either downgrade the highlight -> notifications (thus ignoring the event).

This would potentially mean that `/notifications?only=highlight` would return a lot of not-useful data, but it could be interested to consider.

The other potential solution here is the ideas in vector-im/element-web#6874.

proposals/3996-encrypted-mentions-only-rooms.md
```diff
@@ -0,0 +1,193 @@
+# MSC3996: Encrypted mentions-only rooms
```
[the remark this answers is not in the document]
clokep: (This was mentioned in a private channel, but I'll mention it here for transparency since I did a poor job of giving context on why this MSC exists.)

This MSC was written to formalize the idea of whether it would make sense to expose any additional metadata (the bare minimum). The trade-offs might not make sense, but I wanted to *remove* that conversation from MSC3952 to avoid blocking that MSC further. Thanks for starting the conversation here! 👍 

Currently I have no plans to implement this MSC, as you said it has downsides of more metadata leakage and doesn't necessarily give us all the features we want.

I was hoping those who were advocating for exposing all the mentions information in MSC3952 might be part of the conversation here, but so far this MSC has generated little interest.

In the meantime we're exploring other ways to properly push all E2EE messages while accounting for rooms that are muted or set to mentions & keywords only, etc.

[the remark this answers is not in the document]
sumnerevans: I finally got some time to actually read through this and understand what it's saying. I think part of why I at least haven't chimed in is that the MSC is somewhat confusingly named. I think that "Unencrypted Mention Indicator in Encrypted Rooms" or something would be more clear as to what this is about.

As for what Beeper is planning on doing to resolve the concerns we had with MSC3952, we are just putting the `mentions` in plaintext anyway. (We are using the presence of that field to determine whether or not the intentional mentions code path should be used to evaluate push rules.) Currently, only our bridges are utilizing this MSC to ensure that our users get notifications for mentions from other networks.

Example of what we've done in mautrix-go: https://github.com/mautrix/go/blob/3e840e962e24e90a439f91ce0cdf5e7251008413/crypto/encryptmegolm.go#LL133-L135. Whether or not to include the mentions in plaintext is configurable. We have that enabled for at least the whatsapp bridge right now.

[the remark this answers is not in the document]
clokep: > I think part of why I at least haven't chimed in is that the MSC is somewhat confusingly named. I think that "Unencrypted Mention Indicator in Encrypted Rooms" or something would be more clear as to what this is about.

I suppose it should really be "Supporting mentions-only encrypted room" or something.

PR #3316, opened 2021-08-07
MSC3316: Add timestamp massaging to the spec

tulir:
[Rendered](https://github.com/tulir/matrix-doc/blob/appservice-timestamp-massaging/proposals/3316-appservice-timestamp-massaging.md)

Synapse implementation: [original `/send`-only impl](https://github.com/matrix-org/synapse/blob/v1.53.0/synapse/rest/client/room.py#L264-L265) + state events in https://github.com/matrix-org/synapse/pull/11866

FCP proposal: https://github.com/matrix-org/matrix-doc/pull/3316#issuecomment-1055286692

PR #3998, opened 2023-04-13
MSCXXXX: Add proposal for `/join `and `/knock` timestamp massaging

MadLittleMods:
MSCXXXX: Add proposal for `/join `and `/knock` timestamp massaging

---

Rendered (TODO)

Server implementations:

 - *Implementations needed*
 
Other related references:

 - Original MSC where timestamp massaging was introduced: [MSC3316](https://github.com/matrix-org/matrix-spec-proposals/pull/3316)
    - https://github.com/matrix-org/synapse/pull/2754
    - https://github.com/matrix-org/synapse/pull/11866
 - Related [MSC3997](https://github.com/matrix-org/matrix-spec-proposals/pull/3997) to  add timestamp massaging to `/createRoom`
    - Synapse issue tracking `/createRoom` timestamp massaging proposed in this MSC and mentions the `/join` problem: https://github.com/matrix-org/synapse/issues/15346

### DIFF
--- a/proposals/0000-join-timestamp-massaging.md
+++ b/proposals/0000-join-timestamp-massaging.md
@@ -1,0 +1,98 @@
+# MSC0000: Add timestamp massaging to `/join` and `/knock`
+
+As mentioned in the original
+[MSC3316](https://github.com/matrix-org/matrix-spec-proposals/pull/3316) for timestamp
+massaging,
+
+> We consciously don't support the `ts` parameter on the various helper
+> syntactic-sugar APIs like `/kick` and `/ban`. If a bridge/bot is smart enough to
+> be faking history, it is already in the business of dealing with raw events,
+> and should not be using the syntactic sugar APIs.
+
+While it's possible to mimic a join/invite/knock for a room that the server already
+knows about, this falls apart for a federated action for a room that the server doesn't
+know about since it's not possible to specify any `via` servers with the `/state`
+endpoint. Currently, if you try with Synapse, it will throw a `404` with the following
+error response body:
+
+```json
+{
+  "errcode": "M_UNKNOWN",
+  "error": "Can't join remote room because no servers that are in the room have been provided."
+}
+```
+
+When writing end-to-end tests, it's useful to have your room appear as though it was
+created back in time before your messages were sent and to have stable/consistent
+timestamps. If you start using timestamp massaging when sending messages, it can appear
+as though those events occurred before the `m.room.member` events used to join federated
+rooms. We specifically run into this with the [Matrix Public
+Archive](https://github.com/matrix-org/matrix-public-archive/) end-to-end tests.
+
+In real-life scenarios, practically, this hasn't mattered much for content because the
+DAG is ordered topologically and not by timestamp but is a semantic inconsistency that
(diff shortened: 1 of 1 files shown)


proposals/3998-join-timestamp-massaging.md
```diff
@@ -0,0 +1,98 @@
+# MSC3998: Add timestamp massaging to `/join` and `/knock`
```
turt2live, line 1: It's not entirely clear to me why `/createRoom` is isolated from these endpoints at the moment. To ease discussion, it might make sense to merge the MSCs into a single place to catch what other endpoints might need this behaviour then break it up (if needed) prior to FCP.

MadLittleMods: They need timestamp massaging for different reasons and alternatives vary. I think it makes more sense to split them up but it does make sense to keep the bigger picture in mind.

proposals/3998-join-timestamp-massaging.md
```diff
@@ -0,0 +1,98 @@
...
+
+Add timestamp massaging to the `/join` and `/knock` endpoints to be able to override the
+`origin_server_ts` of sent events. We do this by adding a `ts` querystring parameter
+that specifies the value to apply to `origin_server_ts` on the event (UNIX epoch
+milliseconds).
+
+ - `POST /_matrix/client/v3/join/{roomId}?ts=123`
+ - `POST /_matrix/client/v3/knock/{roomIdOrAlias}?ts=123`
+
+Also related: [MSC3997](https://github.com/matrix-org/matrix-spec-proposals/pull/3997)
```
MadLittleMods, line 49:
```suggestion
This functionality is restricted to the application service (AS) API to be consistent
with [MSC3316](https://github.com/matrix-org/matrix-spec-proposals/pull/3316). There
could be future considerations to opening this up to any client as it's kinda arbitrary
to restrict it this way and just seems like friction to try to get only people with good
intentions using it.

---

Also related: [MSC3997](https://github.com/matrix-org/matrix-spec-proposals/pull/3997)
```

proposals/3998-join-timestamp-massaging.md
```diff
@@ -0,0 +1,106 @@
...
+timestamps. If you start using timestamp massaging when sending messages, it can appear
+as though those events occurred before the `m.room.member` events used to join federated
+rooms. We specifically run into this with the [Matrix Public
+Archive](https://github.com/matrix-org/matrix-public-archive/) end-to-end tests.
+
+In real-life scenarios, practically, this hasn't mattered much for content because the
+DAG is ordered topologically and not by timestamp but is a semantic inconsistency that
+is becoming more important with API's like `/timestamp_to_event` which find events by
+their `origin_server_ts`. And makes things tricky for the Matrix Public Archive to
+navigate history by date seamlessly assuming good intentions.
```
MadLittleMods, line 36: It also mattered to us in real life with historical Gitter rooms where we imported history and ran into these messy client issues:
 
  - https://github.com/matrix-org/matrix-public-archive/issues/182
  - https://github.com/matrix-org/matrix-public-archive/issues/198

PR #3030, opened 2021-02-25
MSC3030: Jump to date API endpoint

MadLittleMods:
MSC3030: Jump to date API endpoint

[Rendered](https://github.com/matrix-org/matrix-doc/blob/eric/msc-jump-to-date/proposals/3030-jump-to-date.md)
[FCP proposal](https://github.com/matrix-org/matrix-doc/pull/3030#issuecomment-1013856557)

Homeserver implementations:

 - Synapse: https://github.com/matrix-org/synapse/pull/9445

Client implementations:

 - `/jumptodate` slash command, https://github.com/matrix-org/matrix-react-sdk/pull/7372
 - Jump to date headers (with date picker) in timeline: https://github.com/matrix-org/matrix-react-sdk/pull/7339
 - Matrix Public Archive: https://github.com/matrix-org/matrix-public-archive

---

Related issues:

 - https://github.com/vector-im/element-web/issues/7677
 - https://github.com/matrix-org/synapse/issues/7164

PR #3999, opened 2023-04-14
MSCXXXX: Add causal parameter to `/timestamp_to_event`

MadLittleMods:
MSCXXXX: Add causal `?event_id=$abc` query parameter to `/timestamp_to_event` to signal that we want to keep progressing from this event regardless of what timestamp shenanigans are going on.

---

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/madlittlemods/timestamp-to-event-causality-parameter/proposals/0000-timestamp-to-event-causality-parameter.md)

Server implementations:

 - *Implementations needed*
 
Other related references:

 - Original MSC where `/timestamp_to_event` (jump to date) was introduced, https://github.com/matrix-org/matrix-spec-proposals/pull/3030

### DIFF
--- a/proposals/0000-timestamp-to-event-causality-parameter.md
+++ b/proposals/0000-timestamp-to-event-causality-parameter.md
@@ -1,0 +1,24 @@
+# MSC0000: Template for new MSCs
+
+
+
+
+## Proposal
+
+
+## Potential issues
+
+
+
+
+## Alternatives
+
+
+
+
+## Security considerations
+
+
+
+## Unstable prefix
+
--- a/proposals/0000-timestamp-to-event-causality-parameter.md
+++ b/proposals/0000-timestamp-to-event-causality-parameter.md
@@ -1,24 +1,129 @@
-# MSC0000: Template for new MSCs
+# MSCXXXX: Add causal parameter to `/timestamp_to_event`
 
+Causality just means being able to know "A happened before B" or vice versa.
 
+Because `origin_server_ts` on an event is untrusted and can be set to whatever value,
+using `/timestamp_to_event` can give you rogue results or even make you paginate in
+loops. `/timestamp_to_event` was originally introduced in
+[MSC3030](https://github.com/matrix-org/matrix-spec-proposals/pull/3030/) where this
+rogue result problem was first mentioned.
+
+For example even in a simple good intentioned case, you create a new room which puts all
+of the primordial creation events at time X. You then send some messages using timestamp
+massaging (`/send?ts=123`) at a time before X (this is exactly what we did with the
+Gitter historical import). The problem occurs when you use
+`/timestamp_to_event?dir=f&ts=123` because you will end up looping back to the
+`m.room.create` that comes after.
+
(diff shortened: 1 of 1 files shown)


proposals/3999-timestamp-to-event-causal-parameter.md
```diff
@@ -0,0 +1,129 @@
...
+These loops make it hard properly navigate from page to page in the [Matrix Public
+Archive](https://github.com/matrix-org/matrix-public-archive) where you view a given
+date in the room and use `/timestamp_to_event` to jump to the next page. 
+
+
+## Proposal
+
+Instead of only defining a timestamp, we add an optional `event_id` query parameter
+which represents a topological spot in the DAG that we can easily determine what comes
+before or after from (a casual relationship).
```
MadLittleMods, line 30: We should clarify that this should be added to the `client` and `federation` (server) API's

proposals/3999-timestamp-to-event-causal-parameter.md
```diff
@@ -0,0 +1,129 @@
...
+
+These loops make it hard properly navigate from page to page in the [Matrix Public
+Archive](https://github.com/matrix-org/matrix-public-archive) where you view a given
+date in the room and use `/timestamp_to_event` to jump to the next page. 
+
+
+## Proposal
+
+Instead of only defining a timestamp, we add an optional `event_id` query parameter
+which represents a topological spot in the DAG that we can easily determine what comes
```
MadLittleMods, line 29: Event ID's are globally unique but we're after a topological spot in the DAG which is per-room so the event lookup should be constrained to the room.

MadLittleMods, line 29: What happens when `event_id` can't be found?

Should the homeserver try to backfill the event from other homeservers?

What error code should we throw if we can't find it in any case?

proposals/3999-timestamp-to-event-causal-parameter.md
```diff
@@ -0,0 +1,129 @@
...
+For example even in a simple good intentioned case, you create a new room which puts all
+of the primordial creation events at time X. You then send some messages using timestamp
+massaging (`/send?ts=123`) at a time before X (this is exactly what we did with the
+Gitter historical import). The problem occurs when you use
+`/timestamp_to_event?dir=f&ts=123` because you will end up looping back to the
+`m.room.create` that comes after.
+
+In less scrupulous scenarios or with bad intentioned actors, these timestamp loops can
+occur throughout the room timeline.
+
```
MadLittleMods, line 20: Another case why the causal parameter is useful: Events can occur at the same exact time and it's not possible to get the next one with `/timestamp_to_event` atm

proposals/3999-timestamp-to-event-causal-parameter.md
cloudrac3r, line 1: If the purpose is to allow Matrix Public Archive to go back through events in the room, why not use [/messages](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3roomsroomidmessages) and the from/to/start/end fields for pagination?

PR #4000, opened 2023-04-14
MSCXXX: Forwards fill (`/backfill` forwards)

MadLittleMods:
MSCXXX: Forwards fill (`/backfill` forwards)

---

Rendered (TODO)

Server implementations:

 - *Implementations needed*
 
Other related references:

 - TODO

### DIFF
--- a/proposals/0000-forwards-fill.md
+++ b/proposals/0000-forwards-fill.md
@@ -1,0 +1,148 @@
+# MSCXXXX: Forwards fill (`/backfill` forwards)
+
+Currently, `/messages` only backfills events when you paginate backwards with `?dir=b`.
+The limiting factor here is `/backfill` which only looks backwards to find preceding
+events.
+
+The spec just says this:
+
+> Since users in that room are able to request the history by the `/messages` client API
+> endpoint, it’s possible that they might step backwards far enough into history before
+> the homeserver itself was a member of that room.
+> 
+> To cover this case, the federation API provides a server-to-server analog of the
+> `/messages` client API, allowing one homeserver to fetch history from another. This is
+> the `/backfill` API.
+> 
+> *-- https://spec.matrix.org/v1.5/server-server-api/#backfilling-and-retrieving-missing-events*
+
+This proposal is after adding a way to `/backfill` forwards, also known as *forwards fill*.
+
+This also opens the door to adding backfill support to `/context` which returns events
+around a given event in both directions. Please note, that this MSC doesn't dictate when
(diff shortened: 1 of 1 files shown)


proposals/4000-forwards-fill.md
```diff
@@ -0,0 +1,148 @@
...
+While this feature is in development, the endpoint can be used with the unstable prefix
+at `GET /_matrix/federation/org.matrix.msc4000/forwards_fill/{roomId}` instead of it's
+stable `/v1` location as described above.
+
+### While the MSC is unstable
+
+During this period, to detect server support, other servers should check for the
+presence of the `org.matrix.msc4000` flag in `unstable_features` on `/versions`. Servers
+are also required to use the unstable prefixes (see [unstable prefix](#unstable-prefix))
+during this time.
```
ShadowJonathan, line 129: `/versions` does not exist for the Federation API, do you mean MSC3723?

MadLittleMods: I was thinking the normal `GET /_matrix/client/versions` endpoint but didn't realize it had `client` in the name. Perhaps it's an overloaded use but seems fine? I don't see anything in the spec about only using for client-facing things and it doesn't require authentication to probe.

I guess you touch on this in MSC3723 ⏩

ShadowJonathan: Keep in mind that some servers might hide their client API, and not announce it in `.well-known/matrix/client`, for privacy reasons or else. The only assured thing is that you know which endpoint the federation one is.

MadLittleMods: > Usually for federation we just attempt to hit the endpoint and deal with the unimplemented error codes.

Updated to use this 👍

proposals/4000-forwards-fill.md
```diff
@@ -0,0 +1,150 @@
...
+> To cover this case, the federation API provides a server-to-server analog of the
+> `/messages` client API, allowing one homeserver to fetch history from another. This is
+> the `/backfill` API.
+> 
+> *-- https://spec.matrix.org/v1.5/server-server-api/#backfilling-and-retrieving-missing-events*
+
+This proposal is after adding a way to `/backfill` forwards, also known as *forwards fill*.
+
+This also opens the door to adding backfill support to `/context` which returns events
+around a given event in both directions. Please note, that this MSC doesn't dictate when
```
MadLittleMods, line 22: Add some context on why good:

This will better facilitate backfill for `/context` which means you can actually visit a permalink for section of history your server doesn't know about yet and actually be able to see the message.

PR #4001, opened 2023-04-21
MSCxxxx Return start of room state at context endpoint

BramvdnHeuvel:


proposals/4001-context-state-at-start.md
```diff
@@ -0,0 +1,31 @@
+# MSC4001: Return start of room state at context
+
+The endpoint `GET /_matrix/client/v3/rooms/{roomId}/context/{eventId}` describes that the `state` field in the 200 response returns the state of the room at the **last** event returned. This is the case for all six spec versions released so far:
```
uhoreg, line 3: Housekeeping comment: please wrap your lines at around 80 to 120 characters.

proposals/4001-context-state-at-start.md
```diff
@@ -0,0 +1,31 @@
...
+- [v1.6](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3roomsroomidcontexteventid)
+
+Earlier in these specs, the following rationale is described:
+
+> [...] consider a returned timeline [M0, S1, M2], where M0 and M2 are both messages sent by the same user, and S1 is a state event where that user changes their displayname. If the state list represents the room state at the end of the timeline, the client must take a copy of the state dictionary, and rewind S1, in order to correctly calculate the display name for M0.
+
+
+## Proposal
+
+For the same reason that this behaviour was altered for the `GET /_matrix/client/v3/sync` endpoint, the endpoint `GET /_matrix/client/v3/rooms/{roomId}/context/{eventId}` should use the `state` field in its 200 response to return the room state at the **first** event returned.
```
uhoreg, line 19: Under the assumption that this isn't a spec bug, this would be be an incompatible change and could break clients that rely on the current behaviour.  This could be solved by adding a query parameter to indicate that the server should use this behaviour.  Alternatively, the version number on this endpoint could be bumped.  Or you could show that clients don't rely on the current behaviour.

PR #4003, opened 2023-04-25
MSCXXXX: Semantic table attributes

AndrewKvalheim:
[Rendered](https://github.com/AndrewKvalheim/matrix-spec-proposals/blob/semantic-table-attributes/proposals/XXXX-semantic-table-attributes.md)

re matrix-org/matrix-spec#1498, vector-im/element-web#25179


proposals/4003-semantic-table-attributes.md
turt2live, line 1: The implementation requirements for this MSC were asked out of band. In my opinion, I think it's best implemented by showing the need for the attributes rather than a client's capability to render them (we already know that clients can render HTML tables if they want to). Specifically, a feature in a bot, bridge, or client demonstrates a measurable consequence by continuing to suggest stripping the table metadata.

PR #4004, opened 2023-04-26
Draft: MSC-4003: unified view of identity service

guimard:
Centralized chat platforms are able to automatically detect which entries of
the phone address book are known by the platform, then automatically propose
those names when user clicks on "new discussion".

The Matrix-Identity-Service has already a secured mechanism able to be used
for this _(lookup)_. However, the identity service is only able to answer
with data it knows _(ie submitted lookups)_.

This MR proposes a mechanism to give an unified view of the identity service


turt2live:
@guimard please note that in order for us to accept this (a much later step in the spec process), we'll need [Sign Off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes.

guimard:
> @guimard please note that in order for us to accept this (a much later step in the spec process), we'll need [Sign Off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes.

Hi @turt2live, I just added the "signed-off". Thanks !

proposals/4004-unified-identity-service-view.md
```diff
@@ -0,0 +1,73 @@
...
+    "matrix.domain.com:8448": [
+      "nlo35_T5fzSGZzJApqu8lgIudJvmOQtDaHtr-I4rU7I"
+    ]
+  }
+}
+```
+
+## Potential issues
+
+To avoid conversation hijacking, only trusted server should be allowed to push
```
MTRNord, line 67: How are these trusted servers going to be determined? How are pushing servers supposed to know if they are trusted?

guimard: Hi @MTRNord,
there are already some "trusted servers" parameters in Matrix-Synapse, I don't know if the way to trust servers should be part of this MSC, shouldn't it ?

PR #4005, opened 2023-05-01
MSCXXXX: Explicit read receipts for sent events

bradtgmurray:


proposals/XXXX-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,89 @@
...
+
+https://spec.matrix.org/v1.5/client-server-api/#receiving-notifications says (emphasis mine)
+
+> When the user updates their read receipt (either by using the API or by sending an event),
+> notifications prior to and including that event MUST be marked as read.
+
+This leads to overly complicated implementations on both the client and the server. Each codebase
+needs to comapre the sender of events as well as the read receipts in order to handle marking the
+room as read, clearning notification counts, and displaying read markers for other users. This code
+is complex and leads to bugs, such as https://github.com/matrix-org/synapse/issues/14837
```
clokep, line 22:
```suggestion
needs to compare the sender of events as well as the read receipts in order to handle marking the
room as read, clearing notification counts, and displaying read markers for other users. This code
is complex and leads to bugs, such as https://github.com/matrix-org/synapse/issues/14837
```

proposals/XXXX-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,89 @@
...
+https://github.com/matrix-org/matrix-spec/issues/1410):
+
+The client behavior section of https://spec.matrix.org/v1.5/client-server-api/#receipts says
+
+> Clients SHOULD NOT send read receipts for events sent by their own user.
+
+https://spec.matrix.org/v1.5/client-server-api/#receiving-notifications says (emphasis mine)
+
+> When the user updates their read receipt (either by using the API or by sending an event),
+> notifications prior to and including that event MUST be marked as read.
```
clokep, line 17: I'm guessing you wanted to emphasize the second bit:

```suggestion
https://spec.matrix.org/v1.5/client-server-api/#receiving-notifications says (emphasis mine)

> When the user updates their read receipt (either by using the API *or by sending an event)*,
> notifications prior to and including that event MUST be marked as read.
```

proposals/XXXX-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,89 @@
...
+"where the user has read up to" is simplified to just a read receipt, where previously it was the
+greater value of the user's read receipt and their last sent event.
+
+Clients should receive this receipt at the same time as receiving the event itself back from the
+homeserver (in the same /sync payload). The read receipt can't be sent prior to the event, as the
+receipt would then be pointing to an event that the client doesn't know about yet, and if the event
+is sent prior to the receipt there's a risk of the client attempting to generate a duplicate read
+receipt for the same event.
+
+The client logic for marking a room as read can be simplified to moving the m.read to the latest
```
clokep, line 38:
```suggestion
The client logic for marking a room as read can be simplified to moving the `m.read` (or `m.read.private`) receipt to the latest
```

proposals/XXXX-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,89 @@
...
+room as read, clearning notification counts, and displaying read markers for other users. This code
+is complex and leads to bugs, such as https://github.com/matrix-org/synapse/issues/14837
+
+## Proposal
+
+The homeserver should generate a public (`m.read`) read receipt for the user whenever the user
+sends an event to the room. The receipt should point at the newly created event. This receipt should
+be generated prior to calculating notification counts or push rules. This means the concept of
+"where the user has read up to" is simplified to just a read receipt, where previously it was the
+greater value of the user's read receipt and their last sent event.
```
clokep, line 30: How would this handled threaded read receipts?

I suspect saying something like:

> The event relationships should be followed up to 3 jumps to attempt to find a thread that the newly created event belongs to. If no thread is found, then the main thread is assumed.

This might not work well for unthreaded clients though.

bradtgmurray: Hrm, I haven't read the threaded read receipt part of the spec yet, so my thoughts are fresh here:

Can we assume that if a client is participating in a thread, they're a threaded client and want to generate a threaded read receipt? I guess participating in a thread is a tricky concept. New messages in the thread will have the m.thread relation, but a reaction to an existing message in a thread won't.

How do we infer the clients preference for a threaded vs unthreaded read receipt?

proposals/XXXX-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,89 @@
+# MSC4005: Explicit read receipts for sent events
```
turt2live, line 1: (don't forget to rename your file)

proposals/XXXX-explicit-read-receipts-on-event-send.md
bradtgmurray: Another way of thinking about this is "what is the current behaviour"?

I think today a message sent to a non-main thread won't clear a notification_count on the main thread, correct? If that's the case, we could make it so that messages to a non-main thread generate a threaded receipt, as this is no worse than the behaviour we already have today.

clokep: > Another way of thinking about this is "what is the current behaviour"?

Well the current behavior of Synapse is that it doesn't do this... 😄  But clients do show whether threads are unread or not...

> I think today a message sent to a non-main thread won't clear a notification_count on the main thread, correct? If that's the case, we could make it so that messages to a non-main thread generate a threaded receipt, as this is no worse than the behaviour we already have today.

Yeah, I think that makes sense.

Implementation wise I think you would just call [`get_thread_id_for_receipts`](https://github.com/matrix-org/synapse/blob/6a671ea5a7d3620fa5bc06286812972285772f93/synapse/storage/databases/main/relations.py#L1025) in synapse. (Also see [`_is_event_in_thread`](https://github.com/matrix-org/synapse/blob/6a671ea5a7d3620fa5bc06286812972285772f93/synapse/rest/client/receipts.py#L116) when creating receipts.)

bradtgmurray: Specified in https://github.com/matrix-org/matrix-spec-proposals/pull/4005/commits/311e8e0649fd56ace649c03e241232e04f4b0ea5

bradtgmurray:
FYI, I'm going to take a shot at implementing this in Synapse over the next few weeks when I find some time.

proposals/4005-explicit-read-receipts-on-event-send.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Proposal
+
+The homeserver should generate a public (`m.read`) read receipt for the user whenever the user
+sends an event to the room. The receipt should point at the newly created event. This receipt should
+be generated prior to calculating notification counts or push rules. This means the concept of
+"where the user has read up to" is simplified to just a read receipt, where previously it was the
+greater value of the user's read receipt and their last sent event. If the event is "in a thread"
+(using the definition from [threaded read receipts](https://spec.matrix.org/v1.7/client-server-api/#threaded-read-receipts))
+the generated read receipt should be a threaded read receipt.
```
andybalaam, line 32: If I read this correctly, I think it is saying that if an event is in the `main` timeline, the server should generate an unthreaded read receipt.

I think this would produce unexpected behaviour in a thread-aware client: if I send a message in the main timeline in my thread-aware client, I would not expect all unread threads to become read.

It's possible we need the client to tell the server whether the event it is sending is within the `main` timeline in a threaded context, or in an unthreaded context.

PR #4007, opened 2023-05-02
Fix typos in MSC2746.

dbkr:
This was causing the spell checker to fail on every other proposal.


turt2live (approved):
This sort of stuff can be pushed direct to main fwiw

PR #4009, opened 2023-05-03
MSC4009: Expanding the Matrix ID grammar to enable E.164 IDs

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/e164-ids/proposals/4009-e.164-matrix-ids.md)

Implementations:

* matrix-org/synapse#15536

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4009#issuecomment-1534839758)

### DIFF
--- a/proposals/4009-e.164-matrix-ids.md
+++ b/proposals/4009-e.164-matrix-ids.md
@@ -53,6 +53,14 @@ of an E.164 identifier to an actual Matrix ID, that's best left to the current
 [identity service `/lookup` endpoint](https://spec.matrix.org/v1.6/identity-service-api/#post_matrixidentityv2lookup)
 or future discovery services.
 
+----
+
+If a service uses E.164 identifiers as Matrix IDs then it must be careful to avoid
+leaking history when reassigning IDs to new users. This is related to [#234](https://github.com/matrix-org/matrix-spec/issues/234),
+but only applies to the localpart of the Matrix ID, not the entire domain. The
+solution remains the same: using portable identifiers ([MSC1228](https://github.com/matrix-org/matrix-spec-proposals/pull/1228)
+or [MSC4014](https://github.com/matrix-org/matrix-spec-proposals/pull/4014)).
+
 ## Alternatives
 
 The `+` could continue to be disallowed and left off the beginning of the Matrix
--- a/proposals/4009-e.164-matrix-ids.md
+++ b/proposals/4009-e.164-matrix-ids.md
@@ -1,0 +1,70 @@
+# MSC4009: E.164 Matrix IDs
+
+[E.164](https://www.itu.int/rec/T-REC-E.164) is a set of international standards
+for telephone numbering. This defines a phone number as a `+` followed by a country
+code (e.g. `1` for the US, `44` for the UK) followed by a local phone number.
+For example, a E.164 phone number would look like: `+15558675309`.
+
+It is somewhat common[^1] for social networks or messaging applications to use phone
+numbers as identifiers instaed of relying on users remembering separate usernames
+(and needing to tell other users how to find them).
+
+[Matrix user identifierss](https://spec.matrix.org/v1.6/appendices/#user-identifiers)
+are *almost* compatible with E.164:
+
+> The `localpart` of a user ID is an opaque identifier for that user. It MUST NOT
+> be empty, and MUST contain only the characters `a-z`, `0-9`, `.`, `_`, `=`, `-`,
+> and `/`.
+
+## Proposal
+
+Add `+` to the list of allowed characters in a Matrix user identifier. This would
+allow a full E.164 phone number as the localpart of a Matrix ID: `@+15558675309:example.com`.
+
+It could additionally lower the impedance mismatch when brdiging if the identifiers
+use the full E.164 phone numbers as identifiers.
+
+## Potential issues
+
+Homeservers and clients must already be
+[more tolerant of Matrix user IDs](https://spec.matrix.org/v1.6/appendices/#historical-user-ids):
+
+> Older versions of this specification were more tolerant of the characters
+> permitted in user ID localparts. [...] clients and servers MUST accept user IDs
+> with localparts from the expanded character set:
+>
+> `extended_user_id_char = %x21-39 / %x3B-7E  ; all ASCII printing chars except :`
+
+Thus it is expected to not cause any issues, although clients will need to identify
+that the `+` character is valid for a homeserver. This could be from the supported
+Matrix versions which the homeserver advertises.
+
+## Alternatives
+
+The `+` could continue to be disallowed and left off the beginning of the Matrix
+IDs Note that Synapse reserves digit-only usernames for guest users, although this
+seems to be an implementation detail and not mandated by the specification.
+
+Another option would be to [map from other character sets](https://spec.matrix.org/v1.6/appendices/#mapping-from-other-character-sets).
+This would generate a Matrix ID like `@=2b15558675309:example.com`, which would
+dramatically impact usability for homeservers which use phone numbers as
+identifiers for all users.
+
+----
+
+Although E.164 limits the `+` character to the initial character there seems to
+be no reason to limit that for Matrix identifiers.
+
+## Security considerations
+
+None noted.
+
+## Unstable prefix
+
+N/A
+
+## Dependencies
+
+N/A
+
+[^1]: E.g. Signal, WhatsApp, etc.
--- a/proposals/4009-e.164-matrix-ids.md
+++ b/proposals/4009-e.164-matrix-ids.md
@@ -9,7 +9,7 @@ It is somewhat common[^1] for social networks or messaging applications to use p
 numbers as identifiers instead of relying on users remembering separate usernames
 (and needing to tell other users how to find them).
 
-[Matrix user identifierss](https://spec.matrix.org/v1.6/appendices/#user-identifiers)
+[Matrix user identifiers](https://spec.matrix.org/v1.6/appendices/#user-identifiers)
 are *almost* compatible with E.164:
 
 > The `localpart` of a user ID is an opaque identifier for that user. It MUST NOT
@@ -21,7 +21,7 @@ are *almost* compatible with E.164:
 Add `+` to the list of allowed characters in a Matrix user identifier. This would
 allow a full E.164 phone number as the localpart of a Matrix ID: `@+15558675309:example.com`.
 
-It could additionally lower the impedance mismatch when brdiging if the identifiers
+It could additionally lower the impedance mismatch when bridging if the identifiers
 use the full E.164 phone numbers as identifiers.
 
 ## Potential issues
--- a/proposals/4009-e.164-matrix-ids.md
+++ b/proposals/4009-e.164-matrix-ids.md
@@ -45,10 +45,11 @@ The `+` could continue to be disallowed and left off the beginning of the Matrix
 IDs Note that Synapse reserves digit-only usernames for guest users, although this
 seems to be an implementation detail and not mandated by the specification.
 
-Another option would be to [map from other character sets](https://spec.matrix.org/v1.6/appendices/#mapping-from-other-character-sets).
-This would generate a Matrix ID like `@=2b15558675309:example.com`, which would
-dramatically impact usability for homeservers which use phone numbers as
-identifiers for all users.
+Another option would be to [map from other character sets](https://spec.matrix.org/v1.6/appendices/#mapping-from-other-character-sets)
+or prefix the Matrix ID with something (e.g. `msisdn`).
+This would generate a Matrix ID like `@=2b15558675309:example.com` or
+`@msisdn_5558675309:example.com`, which would dramatically impact usability
+for the users on homeservers which use phone numbers as identifiers.
 
 ----
 
--- a/proposals/4009-e.164-matrix-ids.md
+++ b/proposals/4009-e.164-matrix-ids.md
@@ -1,4 +1,4 @@
-# MSC4009: E.164 Matrix IDs
+# MSC4009: Expanding the Matrix ID grammar to enable E.164 IDs
 
 [E.164](https://www.itu.int/rec/T-REC-E.164) is a set of international standards
 for telephone numbering. This defines a phone number as a `+` followed by a country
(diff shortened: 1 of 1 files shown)


proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
+# MSC4009: E.164 Matrix IDs
```
FSG-Cat, line 1: Sounds like a sensible proposal and since we dont enforce the gramar anywhere outside of the CS API and because of all the fuckery with "illegal" MXIDs aka MXIDs that contain Unicode etc that has been done there is proof in the real world that MXIDs containing + are not going to magically break clients as this has already been deployed by the bot `@+:jae.fi` as an example of one application of MXID that contains +.

clokep: Thanks for the example that this isn't breaking anything in real life!

FSG-Cat: The community has also confirmed that clients dont break in any real way if you where to make a MXID like @🐈:domain. But it does break if you make a MXID that is in excess of 511 bytes i think it is. nheko breaks in that situation. Since we are using 255 max after the overflow drama a single Emoji or 2 is also not breaking clients. Now a MXID with a + in it wont be harder to type so that problem is not there and it wont have problems related to neeeding to be punycode for browsers to be happy in URL paths for stuff like the matrix URI.

clokep:
Synapse implementation at matrix-org/synapse#15536

turt2live:
I don't think this would have needed an implementation in hindsight:

@mscbot fcp merge

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
...
+[E.164](https://www.itu.int/rec/T-REC-E.164) is a set of international standards
+for telephone numbering. This defines a phone number as a `+` followed by a country
+code (e.g. `1` for the US, `44` for the UK) followed by a local phone number.
+For example, a E.164 phone number would look like: `+15558675309`.
+
+It is somewhat common[^1] for social networks or messaging applications to use phone
+numbers as identifiers instead of relying on users remembering separate usernames
+(and needing to tell other users how to find them).
+
+[Matrix user identifierss](https://spec.matrix.org/v1.6/appendices/#user-identifiers)
```
anoadragon453, line 12:
```suggestion
[Matrix user identifiers](https://spec.matrix.org/v1.6/appendices/#user-identifiers)
```

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
...
+
+Another option would be to [map from other character sets](https://spec.matrix.org/v1.6/appendices/#mapping-from-other-character-sets).
+This would generate a Matrix ID like `@=2b15558675309:example.com`, which would
+dramatically impact usability for homeservers which use phone numbers as
+identifiers for all users.
+
+----
+
+Although E.164 limits the `+` character to the initial character there seems to
+be no reason to limit that for Matrix identifiers.
```
anoadragon453, line 56: One could also just represent the number using a [`msisdn`](https://en.wikipedia.org/wiki/MSISDN), which are already in use in the Matrix spec (see [POST /_matrix/identity/v2/validate/msisdn/requestToken](https://spec.matrix.org/v1.6/identity-service-api/#post_matrixidentityv2validatemsisdnrequesttoken). However, I understand that a drawback is that Synapse already reserves all-digit localparts for user IDs.

But why couldn't one just use `@msisdn_44123456789:domain.com`?

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
...
+> The `localpart` of a user ID is an opaque identifier for that user. It MUST NOT
+> be empty, and MUST contain only the characters `a-z`, `0-9`, `.`, `_`, `=`, `-`,
+> and `/`.
+
+## Proposal
+
+Add `+` to the list of allowed characters in a Matrix user identifier. This would
+allow a full E.164 phone number as the localpart of a Matrix ID: `@+15558675309:example.com`.
+
+It could additionally lower the impedance mismatch when brdiging if the identifiers
```
anoadragon453, line 24:
```suggestion
It could additionally lower the impedance mismatch when bridging if the identifiers
```

proposals/4009-e.164-matrix-ids.md
turt2live: because that's a bit ugly :p

anoadragon453: That is a valid complaint!

clokep: Consider the POV of someone who *doesn't know Matrix at all*. If it is *just* a phone number you should be able to easily find people.

This does overlap a bit with the current `msisdn` support, but not really from the POV of a service provider which wants to use them as unique identifiers for users.

I'll mention something about this though.

anoadragon453 (approved):

proposals/4009-e.164-matrix-ids.md
reivilibre, line 1: Pedantically this MSC seems like a misnomer: you seem to be allowing `+` in any position in the localpart, so this doesn't really do anything specific to E.164s.

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
...
+> permitted in user ID localparts. [...] clients and servers MUST accept user IDs
+> with localparts from the expanded character set:
+>
+> `extended_user_id_char = %x21-39 / %x3B-7E  ; all ASCII printing chars except :`
+
+Thus it is expected to not cause any issues, although clients will need to identify
+that the `+` character is valid for a homeserver. This could be from the supported
+Matrix versions which the homeserver advertises.
+
+## Alternatives
```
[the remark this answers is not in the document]
anoadragon453: I did think about that when reviewing that PR. But that PR explicitly talks about historical Matrix IDs. It does not touch the [sanctioned list of what should be allowed](https://github.com/tulir/matrix-spec/blob/cdbf44eef0810d2229bdfb900471c88062262f55/content/appendices.md?plain=1#L547) when creating new Matrix IDs today, while this MSC does.

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,71 @@
...
+This would generate a Matrix ID like `@=2b15558675309:example.com` or
+`@msisdn_5558675309:example.com`, which would dramatically impact usability
+for the users on homeservers which use phone numbers as identifiers.
+
+----
+
+Although E.164 limits the `+` character to the initial character there seems to
+be no reason to limit that for Matrix identifiers.
+
+## Security considerations
```
HarHarLinks, line 59: What's to prevent someone from creating a phone-number-based account manually of a number they don't own, impersonating the owner of that number?

What's to prevent someone from seizing the account or at least impersonating the former owner of some phone number?

clokep: Both of those sound like issues for a service operator to deal with, but I can include something about it in here.

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,70 @@
...
+are *almost* compatible with E.164:
+
+> The `localpart` of a user ID is an opaque identifier for that user. It MUST NOT
+> be empty, and MUST contain only the characters `a-z`, `0-9`, `.`, `_`, `=`, `-`,
+> and `/`.
+
+## Proposal
+
+Add `+` to the list of allowed characters in a Matrix user identifier. This would
+allow a full E.164 phone number as the localpart of a Matrix ID: `@+15558675309:example.com`.
```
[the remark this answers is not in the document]
clokep: > Can you give more elaborate rationale and some examples how these MXIDs can be used?

They're just MXIDs -- if a service provider wishes to use a full E.164 as an MXID they now can. That's all this MSC is attempting to do.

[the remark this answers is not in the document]
maltee1: Given that as of today, there are at least two gatekeepers-to-be using E.164 IDs for identification (WhatsApp and iMessage, though I don't know what imessage uses internally), anyone registered on both services has a good chance of being represented by the same E.164 Matrix ID on different domains (assuming Matrix will be chosen for the DMA). This isn't something to be avoided, but rather to be dealt with. Maybe the matrix spec isn't the best place for that anyway.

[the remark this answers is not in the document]
clokep: > iMessage, though I don't know what imessage uses internally

You can use iMessage without a phone number, so I think they use them more as a mapping service back to an account?

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,71 @@
+# MSC4009: E.164 Matrix IDs
+
+[E.164](https://www.itu.int/rec/T-REC-E.164) is a set of international standards
+for telephone numbering. This defines a phone number as a `+` followed by a country
+code (e.g. `1` for the US, `44` for the UK) followed by a local phone number.
+For example, a E.164 phone number would look like: `+15558675309`.
+
+It is somewhat common[^1] for social networks or messaging applications to use phone
+numbers as identifiers instead of relying on users remembering separate usernames
```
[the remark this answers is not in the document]
clokep: Identity servers are about discovery -- this is about the identifier of the user. It is quite common for services to use a phone number (for better or worse) as a login. This is generally done because it is easy for users to remember, yes at the expense of privacy. For some users this is a tradeoff they're happy to make and we should not limit that technically.

proposals/4009-e.164-matrix-ids.md
clokep: I renamed this, hopefully it is a bit clearer now?

proposals/4009-e.164-matrix-ids.md
clokep: Although it does bring up whether services should *not* allow these by default unless they truly are phone numbers. It is easy enough to skip that check or something for spoofing though, so that's likely not very helpful.

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,95 @@
...
+>
+> `extended_user_id_char = %x21-39 / %x3B-7E  ; all ASCII printing chars except :`
+
+Thus, it is expected to not cause any issues, although clients will need to identify
+that the `+` character is valid for a homeserver. This could be from the supported
+Matrix versions which the homeserver advertises.
+
+----
+
+A user have two account with the same identifier on different services is not
```
uhoreg, line 49:
```suggestion
A user having two accounts with the same identifier on different services is not
```

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,95 @@
...
+ideal, but is not different than today where a user may log into both WhatsApp
+and Signal with their phone number. This MSC does *not* attempt to help with mapping
+of an E.164 identifier to an actual Matrix ID, that's best left to the current
+[identity service `/lookup` endpoint](https://spec.matrix.org/v1.6/identity-service-api/#post_matrixidentityv2lookup)
+or future discovery services.
+
+## Alternatives
+
+The `+` could continue to be disallowed and left off the beginning of the Matrix
+IDs Note that Synapse reserves digit-only usernames for guest users, although this
```
uhoreg, line 59:
```suggestion
IDs. Note that Synapse reserves digit-only usernames for guest users, although this
```

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,95 @@
...
+This would generate a Matrix ID like `@=2b15558675309:example.com` or
+`@msisdn_5558675309:example.com`, which would dramatically impact usability
+for the users on homeservers which use phone numbers as identifiers.
+
+----
+
+Although E.164 limits the `+` character to the initial character there seems to
+be no reason to limit that for Matrix identifiers.
+
+## Security considerations
```
richvdh, line 73: MSISDNs are frequently reassigned to new users. If I come along and convince a server that my phone number is the number that you had last week, does that mean I get access to your history?

It's probably fine to leave this problem up to individual servers/deployments, but it might be worth calling out.

(vaguely related: https://github.com/matrix-org/matrix-spec/issues/234)

clokep: I added a note and pointed to some potential solutions in MSCs.

proposals/4009-e.164-matrix-ids.md
```diff
@@ -0,0 +1,103 @@
...
+Add `+` to the list of allowed characters in a Matrix user identifier. This would
+allow a full E.164 phone number as a user's username on a server, which is common
+practice for many applications. This should hopefully ease bridging of those
+services to Matrix or help them natively speak Matrix in the future. Users would
+not need to learn a new Matrix ID, but can continue using their phone number, as
+today.
+
+Although E.164 IDs are meant to be globally unique they are still namespaced under
+a domain under this proposal, e.g. `@+15558675309:example.com`, as the same user may
+already be registered on multiple service providers.
```
KitsuneRal, line 30: I don't have specific technical objections against supporting `+` in MXIDs but I feel that this MSC basically tries to circumvent the somewhat complex 3PID machinery we have today. Instead of identifying users however they prefer but keeping MXID "an implementation detail", we (provide a means to) hardwire a user to some specific way of identification that is not remotely permanent or inherent to the user identity. If a user changes their number and wants to migrate the MXID, this is not possible. Your truly has changed his phone numbers four times in 15 years because of living in four countries; 3 of those numbers are obviously not connected with me any more. And as I see it, this can complicate fulfilling "the right to be forgotten" even after .

The right thing, from my perspective, would be to have first-class support of phone numbers in user profiles and provide a means in clients to use the phone number as 3PID for Matrix login instead of MXID.

KitsuneRal: For the record: after a discussion elsewhere, the driver/intention of this MSC is not as far-reaching as I anticipated. The idea is to merely enable usage of phone numbers for MXIDs, exactly when using phone numbers for MXIDs is warranted. There is no intention to incorporate phone numbers as a means of identification of specific human users in Matrix.

richvdh:
merged!

PR #4013, opened 2023-05-10
MSC4013: Poll history cache

alfogrillo:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/alfogrillo/poll_history_cache/proposals/4013-poll-history-cache.md)

### DIFF
--- a/proposals/4013-poll-history-cache.md
+++ b/proposals/4013-poll-history-cache.md
@@ -4,13 +4,18 @@ The purpose of this MSC is to allow clients to quickly fetch all the polls in a 
 
 ## Problem
 
-Clients sometimes may want to access quickly to all the polls sent in a given room (similarly to how they may want to access all the media being sent in a room).
-Today clients don’t have an efficient way to make such an operation. Clients that want to implement this feature for encrypted rooms on the status quo may need to backward paginate on the room’s timeline until all the polls are found or they reach the beginning of the timeline.
-For unencrypted rooms the problem doesn't exist since clients can call the [messages API](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3roomsroomidmessages) filtering by event type (e.g. by `m.poll.start`).
+Clients sometimes may want to access quickly to all the polls sent in a given room (similarly to how they may want to
+access all the media being sent in a room).Today clients don’t have an efficient way to make such an operation. 
+Clients that want to implement this feature for encrypted rooms on the status quo may need to backward paginate on the 
+room’s timeline until all the polls are found or they reach the beginning of the timeline. For unencrypted rooms the 
+problem doesn't exist since clients can call the 
+[messages API](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3roomsroomidmessages) 
+filtering by event type (e.g. by `m.poll.start`).
 
 ## Proposal
 
-Introduce a new state event `m.room.poll_history`. This state event is supposed to be referenced by any `m.poll.start` that will be sent next in the room. The new state event must have an empty string as `state_key`.
+Introduce a new state event `m.room.poll_history`. This state event is supposed to be referenced by any `m.poll.start` 
+that will be sent next in the room. The new state event must have an empty string as `state_key`.
 More information on polls can be found on the [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381).
 
 This is how the new state event would look like:
@@ -30,7 +35,8 @@ This is how the new state event would look like:
 - [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/338)
 
 ## Client behaviour
-First of all a room creator sends the new `m.room.poll_history` state event in the `initial_state` when calling the [createRoom API](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3createroom).
+First of all a room creator sends the new `m.room.poll_history` state event in the `initial_state` when calling the 
+[createRoom API](https://spec.matrix.org/v1.6/client-server-api/#post_matrixclientv3createroom).
 
 For example the body when creating a new encrypted room would look like this:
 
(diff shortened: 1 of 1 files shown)


proposals/4013-poll-history-cache.md
```diff
@@ -0,0 +1,180 @@
...
+   "content": {},
+   "state_key": "m.poll.start",
+   "type": "m.room.history",
+   "event_id": "poll_history_id"
+}
+
+```
+
+## Alternatives considered
+
```
uhoreg, line 126: It seems like using [`/search`](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3search) with `"type": "m.room.poll_history"` would be an alternative.

proposals/4013-poll-history-cache.md
```diff
@@ -0,0 +1,180 @@
...
+## Problem
+
+Clients sometimes may want to access quickly to all the polls sent in a given room (similarly to how they may want to access all the media being sent in a room).
+Today clients don’t have an efficient way to make such an operation. Clients that want to implement this feature for encrypted rooms on the status quo may need to backward paginate on the room’s timeline until all the polls are found or they reach the beginning of the timeline.
+For unencrypted rooms the problem doesn't exist since clients can call the [messages API](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3roomsroomidmessages) filtering by event type (e.g. by `m.poll.start`).
+
+## Proposal
+
+Introduce a new state event `m.room.poll_history`. This state event is supposed to be referenced by any `m.poll.start` that will be sent next in the room. The new state event must have an empty string as `state_key`.
+More information on polls can be found on the [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/polls/proposals/3381-polls.md).
```
turt2live, line 14:
```suggestion
More information on polls can be found on the [MSC3381](https://github.com/matrix-org/matrix-spec-proposals/pull/3381).
```

(also MSC3381 should be listed as a dependency as this MSC cannot land without it)

proposals/4013-poll-history-cache.md
alfogrillo: Do you mean an alternative to the [/state](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv3roomsroomidstateeventtypestatekey)? 
If yes, does it make any difference?

uhoreg: I mean that, rather than creating a new state event and searching for relations on that state event, why not just use the /search endpoint to search for polls?

alfogrillo: I guess because search doesn't work on encrypted room. Poll events are actually sent as `m.room.encrypted`. 😕
Fwiw in the main proposal just one state event is required.

uhoreg: That's a good reason.  It might be worth mentioning that in the MSC.

clokep: For unencrypted room `/messages` with a filter on event type would be better than `/search`, FWIW. But again doesn't work with encrypted events.

proposals/4013-poll-history-cache.md
```diff
@@ -0,0 +1,183 @@
+# MSC4013: Poll history cache
```
FSG-Cat, line 1: I have to ask plain and simple where in this MSC do we dismiss the problem that users do not want metadata leaks like this. Or atleast a large group of quite loud users do not want metadata leaks like this. 

Now This concern could be dismissed by the fact that polls already use relations, but atleast these relations cant tell you for sure its a poll you can only tell that the events are related. I personally would argue that its somewhat dubious why we even need relations to not respect encryption considering their current use. 

Still do we need to add more metadata leaks? Matrix already gets bad articles/blog posts written about it for metadata leaking and MSCs like this one doesnt help that.

Johennes: Encrypting the state event (https://github.com/matrix-org/matrix-spec-proposals/pull/3414) would help mitigate this in future.

alfogrillo: > I have to ask plain and simple where in this MSC do we dismiss the problem that users do not want metadata leaks like this. Or atleast a large group of quite loud users do not want metadata leaks like this.

This feature should considered optional. 
In theory room's creators aren't obliged to send the `m.room.poll_history` state event for a room. 
At the same time clients sending a poll may choose to opt out from this behaviour.
At the same time structures like this are very useful on clients that want navigate content by type (polls in this case) and willing to pay the price of a small metadata leak.

> I personally would argue that its somewhat dubious why we even need relations to not respect encryption considering their current use.

I guess the may reason is the server being able to group related events together efficiently. Eg. the same problem is solved in the same way for the "threads" feature.

> Encrypted state would mittigate a bit of the concerns but with plain text relations it would still leave a massive tell. as no other encrypted state event would be the target of this many relations, essentially undoing the whole point in encrypting the cache state event.

Yes, probably people could still guess what would be the `m.poll_history` state event could be in the end, but at least they couldn't be 100% sure I guess.

turt2live, line 1: Please wrap lines to ~120 characters when you get a chance

turt2live: If the state event is optional, the MSC should note that (I don't believe it does currently).

Aside from encrypting the whole state event though, the other option is encrypting the event relationship (like in [MSC4019](https://github.com/matrix-org/matrix-spec-proposals/pull/4019) or similar). I personally don't believe that encrypting the relationship on either end is a blocker for this MSC, however. Ideally we figure out how to encrypt both sides and then use both to protect metadata.

In the meantime, considering polls is quite far away from entering FCP itself, starting implementation unencrypted should be fine (as long as the client is comfortable with the risks). It may need to change in the future as other, adjacent, MSCs make forward progress though.

turt2live (approved):
I actually quite like how this is structured. I was concerned going in that this might end up being a `m.room.pinned_events` clone of sorts, and was pleasantly surprised to see relations being used like this. 

Hopefully it works well in practice :)

proposals/4013-poll-history-cache.md
alfogrillo: done 👍🏻

proposals/4013-poll-history-cache.md
```diff
@@ -0,0 +1,219 @@
...
+                {"id": "id_b", "m.text": "No"}
+            ],
+            "question": { "m.text": "Do you like polls?" },
+            "max_selections": 1,
+        },
+        // reference to the state event of type "m.room.poll_history"
+        "m.relates_to": {
+            "event_id": "poll_history_id",
+            "rel_type": "m.reference"
+        }
```
jonnyandrew, line 87: Could it be an issue that the poll start event would no longer be a root event and therefore could not be the root of a thread?

From https://spec.matrix.org/v1.8/client-server-api/#threading
> Threads are established using a rel_type of m.thread and reference the thread root (the first event in a thread). It is not possible to create a thread from an event which itself is the child of an event relationship (i.e., one with an m.relates_to property).

clokep: Yes, it looks like this would preclude a poll from being a reply in a thread.

jonnyandrew: > Yes, it looks like this would preclude a poll from being a reply in a thread.

This is also a potential problem, but I was actually referring to the fact that a thread _root_ should not be related to any other event. What it means is that there is no way to discuss a poll in a thread, which I imagine is a natural use case for threads.

jonnyandrew: That was my understanding from [the spec which states](https://spec.matrix.org/v1.8/client-server-api/#threading)

> It is not possible to create a thread from an event which itself is the child of an event relationship (i.e., one with an m.relates_to property).

I don't see why it shouldn't be possible though. Perhaps that part of the spec could be updated to accommodate this type of use case.

alfogrillo: Perhaps the documentation meant that you cannot create a thread from an event having already a `rel_type` of type `m.thread`?
@clokep do you know is the home server has checks that may reject this events (events starting a thread from an event with a `rel_type` equal to `m.reference`)?

clokep: > Perhaps the documentation meant that you cannot create a thread from an event having already a `rel_type` of type `m.thread`?

The documentation is correct AFAIK.

> @clokep do you know is the home server has checks that may reject this events (events starting a thread from an event with a `rel_type` equal to `m.reference`)?

They will get rejected: https://github.com/matrix-org/synapse/blob/928e9648578d24dc9f5ed3476d629fca9a64b22a/synapse/handlers/message.py#L1382-L1386

alfogrillo: > They will get rejected: https://github.com/matrix-org/synapse/blob/928e9648578d24dc9f5ed3476d629fca9a64b22a/synapse/handlers/message.py#L1382-L1386

Looking into the code the rejection seems to happens just if the relation type is `m.thread`?
If this is the case, polls having a `m.reference` rel_type can be still be used as root for a thread?

clokep: > Looking into the code the rejection seems to happens just if the relation type is `m.thread`?

If the new event's relation is a thread then it checks that the parent event does not have any relations. I think this situation would be rejected, but I did not check it manually.

PR #4014, opened 2023-05-10
MSC4014: Pseudonymous Identities

kegsay:


proposals/4014-pseudonymous-identities.md
```diff
@@ -0,0 +1,72 @@
...
+### Hard requirements
+ - Clients MUST work without any changes to their code.
+ - Clients MUST have enough data sent to them to be able to migrate to being key-aware.
+
+### Proposal
+
+Please read MSC1228 first, as it is extremely similar to this proposal.
+
+This proposal creates a new room version where the following changes are made:
+- Events no longer have a `sender` field. Events now have a `sender_key` field, which is a base-64 encoded ed25519 public key (encoded the same way as other binary event data in Matrix).
```
ara4n, line 31: where does the private half of `sender_key` live?

proposals/4014-pseudonymous-identities.md
```diff
@@ -0,0 +1,72 @@
...
+
+##### Addressing the "Problems" section in MSC1228
+ - The state keys in the room aliases event is unimportant since v6 rooms removed them from the specification.
+ - matrix.to links remain as they are, with the `via=` params formed using verified mxid mappings.
+ - Redacting an `m.room.member` event should not remove the `mxid_mapping` field in the case where the `displayname` or `avatar_url` is malicious. In order to remove the `mxid_mapping` field (e.g for PII removal), the account in question should be deactivated or the room in question should be "purged", which would cause the mapping to be removed. This more severe form of redaction has side-effects as deleting the `mxid_mapping` potentially alters routing information as that user may be the last user in the room for that server, and hence should not be triggered by a CSAPI `/redact` call. How this functions at a protocol level is undetermined at this point. The redaction algorithm will remain the same for the purposes of signing events / hashes (so the `mxid_mapping` is removed in this case). This implies 2 kinds of redaction for `m.room.member` events.
```
ara4n, line 57: This doesn't sound good - shouldn't we allow users to remove the mxid_mapping of their messages without deactivating their account or nuking the room?

proposals/4014-pseudonymous-identities.md
kegsay: Please make a separate comment for each point, as I've now addressed M's comment but your one is outstanding, which I'm going to resolve now.

proposals/4014-pseudonymous-identities.md
kegsay: The semantics are unclear here. If you remove the mxid mapping you not only remove routing information but you effectively are no longer part of the room in that E2EE will break as other users won't know your user ID, which has ramifications beyond PII removal. If we allow clients to remove the mxid mapping then this is roughly equivalent to _leaving the room_, in which case just do that?

franzitrone:
> There's not a lot of work going into this at the moment. Historical investigation into this back in 2023 was done by a few people, including myself. This resulted in the following MSCs:
> 
> * **MSC4014: Pseudonymous Identities: where events use public keys instead of user IDs for the sender property in events.**
> * MSC4080: Cryptographic Identities: where the responsibility for those keys is shifted to the clients (which requires clients to be cryptographically aware).
> 
> The point of these changes is to lay the groundwork1 by moving identity to client-controlled keys. Assuming these landed, the final piece of this was to make an MSC for "Portable Accounts", which would declare how users would migrate themselves and their data between homeservers. Unfortunately, this work never got funding to see it to completion, hence why nothing has happened for a few years.

Considering that https://github.com/matrix-org/matrix-spec/issues/246 has stagnated, is this proposal still being considered? What is missing to get this into the spec and into the servers?

proposals/4014-pseudonymous-identities.md
JadedBlueEyes: FWIW we rarely want to redact a member event for a user that is still in the room based on that member event - PII remover the user is either gone or has updated their membership, moderation wise the user is banned (and we really want to remove the MXID, as seen recently)

BloodyIron:
Will this enable ACTUAL user deletion and not _only_ deactivation? There are scenarios where fully deleting a user from everything (including the database) is desirable. In one particular case, a fully isolated Matrix ecosystem, fully self-hosted no federation to public Matrix. Wherein at times a user account needs to get recreated, as authentication is exclusively external (LDAP/AD/SSO/whatever) and local accounts are fully disabled. So to trigger a recreation it would be desirable to fully delete the account (not deactivate) and then when the user logs in next the account gets recreated. I don't have specific sample details this moment, but if accounts were to get corrupted or malformed in some way this would be a preferred method to simply recreate the account instead of attempt repair which may not succeed.

Hence the above linking to element-hq/synapse#18902

kegsay:
Yes it mostly would. I say "mostly" because there's still a few places which will be using the previous identity. The main one is encryption. Clients attest to owning a particular user ID, so when you recreate it anyone who previously trusted the old user's keys would not trust your new user's keys. This is solvable, it'll just appear as "@ alice:example.coms identity was reset". If some users have explicitly verified the old user there may be more aggressive warnings of a man in the middle attack though.

BloodyIron:
> Yes it mostly would. I say "mostly" because there's still a few places which will be using the previous identity. The main one is encryption. Clients attest to owning a particular user ID, so when you recreate it anyone who previously trusted the old user's keys would not trust your new user's keys. This is solvable, it'll just appear as "@ alice:example.coms identity was reset". If some users have explicitly verified the old user there may be more aggressive warnings of a man in the middle attack though.

Ahh the key and encryption aspect is certainly a tricky one. I'm not sure myself what the best way to address this would be as I can somewhat see scenarios where recreating the account you may or even may not want to recover/retain the old keys/decryption. Not all implementations of Matrix want to leverage the robust encryption aspects, so I think accounting for both wanting and not wanting to recover them might be prudent.

Eddudos:
I agree that this is indeed needed feature for deleting users. For me it's one of the things keeping me tied from switching from ejabberd to synapse!

wizzeh:
is this still being considered? i will put together a synapse implementation if it gets things moving again

JadedBlueEyes:
I believe it's likely to be superseeded by a similar idea that has a single key per user, although I can't remember the MSC ID

wizzeh:
Are you talking about [4243?](https://github.com/matrix-org/matrix-spec-proposals/pull/4243)

JadedBlueEyes:
Yep I think so

kegsay:
Yeah, I'm currently looking into MSC4243 as an "easier" way to provide much the same benefit as MSC4014. The main difference is that MSC4243 has a single key per user, whereas MSC4014 has per-room keys per user.

The main cost to taking the easier route is that it makes server-side metadata correlation very easy. The main benefit is that it presents a static identity, making it less likely to encounter unverified users, and much more similar to the existing ecosystem. There is a 3rd option involving [key blinding](https://cfrg.github.io/draft-irtf-cfrg-signature-key-blinding/draft-irtf-cfrg-signature-key-blinding.html) but this is significantly more difficult to land due to complexities with how it may interact with E2EE.

wizzeh:
Understood. Well, I'll be very interested to help however I can once an implementation is decided upon :) I'd really like to see nomadic identity landed eventually. I understand it's not element's focus but I'm happy to contribute free labor.

ara4n:
(element would also like nomadic identity, hence funding the work on msc4014, msc1228, msc4243, etc :) once we finally converge on a design then help implementing across the board will be much appreciated. it's worth noting that we're also missing another possible design for nomadic identity - simply indirecting identity like ATproto does)

PR #4015, opened 2023-05-14
MSC4015: Voluntary Bot indicators

MTRNord:
[Rendered](#)

Signed-off-by: Marcel Radzio <support@nordgedanken.dev>

### DIFF
--- a/proposals/4015-bot-indicator.md
+++ b/proposals/4015-bot-indicator.md
@@ -7,6 +7,11 @@ Other networks commonly have indicators that a user is a Bot. In most cases, thi
 a dedicated API that is used by Bots. Matrix doesn't have any dedicated API. However, it can have
 an indicator instead.
 
+A flag would allow to more easily identify bots both in the timeline but crucially also allow them
+to be quickly visible within the room member list.
+Additionally, the flag allows bridges to mark bots from networks like discord, IRC or ActivityPub
+to be marked as bots and therefore retaining the UX they have on the origin network.
+
 
 ## Proposal
 
@@ -54,6 +59,13 @@ the values `true` or `false` but also can be omitted. In this case, the user sho
 People might wrongly assume that this flag is always present and use it either for moderation tools
 or get blinded by it and assume that it's always set when you are a bot.
 These are currently limitations that are beyond the scope of this MSC.
+
+We already have `m.notice` events these days. However, contrary to this MSC they are more invasive in
+behaviour of the bots. Not only do they mark messages as bot messages commonly, but they also change
+the notification settings. While this might make sense for most bots, there are bots where it is
+reasonable to send the messages as normal messages instead. This applies for example to files,
+reminder bots or similar bots that either have no way to make their event m.notice due to the event type
+or are requiring notifications to behave normally due to the function of the bot.
 
 
 ## Alternatives
--- a/proposals/4015-bot-indicator.md
+++ b/proposals/4015-bot-indicator.md
@@ -1,0 +1,76 @@
+# MSC4015: Voluntary Bot indicators
+
+Knowing if a user is a bot is currently hard in Matrix. It currently relies heavily on the Displayname,
+mxid or operating server.
+
+Other networks commonly have indicators that a user is a Bot. In most cases, this is done by having
+a dedicated API that is used by Bots. Matrix doesn't have any dedicated API. However, it can have
+an indicator instead.
+
+
+## Proposal
+
+Adding a marker to the member event content if a user is a bot. This can be set by the User itself
+and is voluntary.
+
+This flag cannot be enforced technically since there is no difference in a bot or a normal user on
+the technical side. So we need to rely on the user. Furthermore, this flag is not supposed to be used
+to detect spambots. It doesn't come with guarantees.
+
+### Example
+
+Such an example member event should look like this:
+
+```json5
+{
+  "content": {
+    "avatar_url": "mxc://example.org/SEsfnsuifSDFSSEF",
+    "displayname": "Alice Connector",
+    "membership": "join",
+    "reason": "Looking for support",
+    // The new flag.
+    "bot": true
+  },
+  "event_id": "$143273582443PhrSn:example.org",
+  "origin_server_ts": 1432735824653,
+  "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
+  "sender": "@example:example.org",
+  "state_key": "@alice:example.org",
+  "type": "m.room.member",
+  "unsigned": {
+    "age": 1234
+  }
+}
+```
+
+### Details of the flag
+
+The flag consists of a boolean flag in the `content` of the event. It can therefore take either
+the values `true` or `false` but also can be omitted. In this case, the user should be considered
+`not a bot`.
+
+## Potential issues
+
+People might wrongly assume that this flag is always present and use it either for moderation tools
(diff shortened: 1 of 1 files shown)


proposals/4015-bot-indicator.md
```diff
@@ -0,0 +1,88 @@
...
+A flag would allow to more easily identify bots both in the timeline but crucially also allow them
+to be quickly visible within the room member list.
+Additionally, the flag allows bridges to mark bots from networks like discord, IRC or ActivityPub
+to be marked as bots and therefore retaining the UX they have on the origin network.
+
+
+## Proposal
+
+Adding a marker to the member event content if a user is a bot. This can be set by the User itself
+and is voluntary.
```
tulir, line 19: How is the flag supposed to be set? Manually managing `m.room.member` events in each room is way too much effort

MTRNord: I must admit, I came from the flag user (UI side) side. And not from the bot creator side of things. Meaning, I might have missed some details for bot implementors here.

tulir: I think the only solution (excluding proper extensible profiles) is allowing setting it in the global profile using the /profile endpoints, and have the server automatically copy it to member events like it does for displaynames and avatar urls. That way it also doesn't get overridden when the name or avatar changes.

(an MSC to allow arbitrary fields in the global profile would be nice :3)

MTRNord: Yep that makes sense. I will later today update the MSC for that and mention extensible profiles as alternative :)

TheArcaneBrony: > (an MSC to allow arbitrary fields in the global profile would be nice :3)

I have to wonder whether something like proper extensible profiles would work here?
Why aren't profiles extensible already, wasn't this a major design goal for Matrix as a whole?

MTRNord: Resolved in a9cc5d9e6834293298dd3cca32fe9c71a93d880f

proposals/4015-bot-indicator.md
```diff
@@ -0,0 +1,115 @@
...
+reminder bots or similar bots that either have no way to make their event m.notice due to the event type
+or are requiring notifications to behave normally due to the function of the bot.
+
+
+## Alternatives
+
+An alternative to using a single flag for `bots` it could make sense to have member types similar to
+how room types work. Here a user could be `m.bot`, `m.puppet` or similar.
+
+Another alternative is [MSC1769](https://github.com/matrix-org/matrix-spec-proposals/pull/1769) which
```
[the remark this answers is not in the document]
MTRNord: (Also closing, since it is listed as alternative by now)

turt2live:
For our future selves trying to determine if this is ready for FCP, I believe the implementation requirement for this MSC is effectively answering the question "Do users of X client want to distinguish between humans and bots, and does that need to be a strong guarantee?" multiple times (ideally against popular clients). 

As of writing, this MSC is trivial to write code for, so I don't think it needs too many links to PRs on the subject. The implementation requirement is to fulfill two purposes: first, that the MSC is technically achievable (already proven implicitly by this MSC), and second that there's interest in the idea (even if not agreement on the technical aspects). The second point feels more vital for this MSC, as adding bot-specific flair isn't great if no one wants it.

For clarity, I believe this amounts to opening issues/questions against several other clients to ask the question above in some way, proving that they're interested in the feature and thus passing implementation. Whether those clients go the extra mile to actually write code to support this feature is great, but I don't think accomplishes the purpose of implementation here.

proposals/4015-bot-indicator.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Proposal
+
+Adding a marker to the member event content if a user is a bot. This can be set by the User itself
+and is voluntary.
+
+This flag cannot be enforced technically since there is no difference in a bot or a normal user on
+the technical side. So we need to rely on the user. Furthermore, this flag is not supposed to be used
+to detect spambots. It doesn't come with guarantees.
+
```
FSG-Cat, line 24: Not exactly sure where in the MSC wording that makes this clear should be slotted but noted here so its not forgotten. 

Implementations of the MSC have to decide them selfs if they trust a given remote network and their bot markers enough to bridge them. Bridged users on discord for example and users of the popular in some groups on the platform accessibility tool Pluralkit will be incorrectly flagged as bots due to both examples using webhooks.

If a implementation ignores the bot markers from networks that are unreliable with their bot markers or implements special rules the bridge will be able to preserve a higher degree of usefulness of this voluntary bot marker.

proposals/4015-bot-indicator.md
```diff
@@ -0,0 +1,115 @@
+# MSC4015: Voluntary Bot indicators
```
MTRNord, line 1: Not sure if this is a thread for here or not but a possible User for this could be the TI-Messenger/Gematik since that spec requires marking Bots as "Chatbot" in the name currently.

image
(Page 39 of the "Spezifikation TI-Messenger-Client" Version 1.1.1 Document)

For english readers the screenshot says (translated using DeepL):

> It MUST be possible to enter function accounts into the VZD-FHIR directory as endpoint of a "HealthcareService" resource of an organisation via the Org-Admin-Client. When configuring the endpoint by the actor in the role "Org-Admin", the display name MUST contain the marker "Chatbot" if the function account is realised via a chatbot. The following formation rule shall be used for the display name: [name of the function account] (chatbot).

HarHarLinks, line 1: As a reference, this is how I achieve this on my server for my community currently with emoji:
image
image

uhoreg:
> The implementation requirement is to fulfill two purposes: first, that the MSC is technically achievable (already proven implicitly by this MSC), and second that there's interest in the idea (even if not agreement on the technical aspects). The second point feels more vital for this MSC, as adding bot-specific flair isn't great if no one wants it.
> 
> For clarity, I believe this amounts to opening issues/questions against several other clients to ask the question above in some way, proving that they're interested in the feature and thus passing implementation. Whether those clients go the extra mile to actually write code to support this feature is great, but I don't think accomplishes the purpose of implementation here.

Some existing issues:

- https://github.com/element-hq/element-meta/issues/1468
- https://github.com/element-hq/element-meta/issues/446
- https://github.com/matrix-org/matrix-spec/issues/27 and https://github.com/matrix-org/matrix-spec/issues/22 (https://github.com/matrix-org/matrix-spec/issues/27#issuecomment-1055636005 proposes a similar solution to this MSC)

proposals/4015-bot-indicator.md
```diff
@@ -0,0 +1,115 @@
...
+Adding a marker to the member event content if a user is a bot. This can be set by the User itself
+and is voluntary.
+
+This flag cannot be enforced technically since there is no difference in a bot or a normal user on
+the technical side. So we need to rely on the user. Furthermore, this flag is not supposed to be used
+to detect spambots. It doesn't come with guarantees.
+
+### Details of the flag
+
+The flag consists of a boolean flag in the `content` of the event. It can therefore take either
```
uhoreg, line 27:
```suggestion
The flag consists of a boolean `bot` flag in the `content` of the event. It can therefore take either
```

The name of the flag doesn't seem to actually be mentioned until the example.

MTRNord: Thanks :)

PR #4018, opened 2023-05-19
MSC4018: Reliable call membership

robintown:
[Rendered](https://github.com/robintown/matrix-doc/blob/reliable-call-membership/proposals/4018-reliable-call-membership.md)

### DIFF
--- a/proposals/4018-reliable-call-membership.md
+++ b/proposals/4018-reliable-call-membership.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Reliable call membership
+# MSC4018: Reliable call membership
 
 In [MSC3401](https://github.com/matrix-org/matrix-spec-proposals/pull/3401), clients announce their participation in a
 call by sending a state event of type `m.call.member`, with their user ID as the state key. This approach is generally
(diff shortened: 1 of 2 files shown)


dbkr (commented):
Seems nice & simple. Could potentially mention that homeservers would have a delay of presumably a few seconds of not having a stream for a device before deleting that device's member state, since it will just be one sync request followed by another rather than, for example, a websocket connection.

axelsimon:
This looks straightforward enough, but we might want to consider more in detail what happens if a homeserver goes offline, as this could easily become a large issue in P2P Matrix.

toger5:
Reding through this again I think the scope of this MSC should be bigger.
Introducing new endpoints for updating the membership, allows us to never grant any user permission to update `m.call.member` state events. Instead the HS will update those using the additional endpoints.
This frees us from the necassity of using the `key = user_id` permission system.

We could use keys of the format: `$user_id/$device_id` which are all handled by the HS.
This would make things much more efficient/simple (especially considering aggreagting sessions), since we dont update the whole state event with multiple memberships anymore but instead only update the data that is actually assocaited with the membership.

proposals/4018-reliable-call-membership.md
```diff
@@ -0,0 +1,128 @@
...
+
+- If a client goes offline while participating in a call, they will be unable to update their call membership state to tell others that they have left.
+  - This prevents clients from showing an accurate preview of who's connected to a call.
+  - In a product like Element Call, where users often join calls with burner accounts and leave them by closing the browser tab, stale call members can build up until a room becomes unusable, since clients joining the call will end up sending out many many to-device messages to invite stale users.
+- There's no way for clients to atomically update their call membership state.
+  - If a user tries to switch devices mid-call, the device that's hanging up may race with the device that's joining as they both try to update the same room state, resulting in one of the updates overwriting the other.
+
+In other group VoIP systems, this is typically solved by having a central application server which tracks each
+participant's connectivity and acts as the source of truth for who's on the call. In Matrix, the closest thing we have
+is a homeserver.
```
toger5, line 15: Even closer would be the SFU itself.
What do you think about the following idea:
The homeserver issues a token to update the memberships for one membership id. The client can then share this token with the SFU or hold keep it to itself.
In case its shared with the sfu, the sfu will update the homeserver on any membership changes (disconnects).

For the fullmesh case the client keeps the token to itself and uses it directly to disconnect.

If the client disconnects from the homeserver directly the homeserver will disconnect this participant independent of whether a explicit disconnect was sent.
(With the rational that a matrixRTC session only works if the user is also connected to matrix: for crypto and other matrixRTC applications that rely on both: matrix + matrixRTC updates at to work properly)

proposals/4018-reliable-call-membership.md
anoadragon453, line 1: This may have implications for anyone hoping to run a matrix homeserver in a "serverless" configuration.

PR #4019, opened 2023-05-20
Encrypted relationships

tusooa:


tusooa:
cc https://github.com/matrix-org/matrix-spec/issues/660

proposals/4019-encrypted-relationships.md
```diff
@@ -0,0 +1,55 @@
...
+Servers MAY track the existence of the state event `m.room.relationship_encryption` and, if it exists
+or existed in the room, respond to the relationships API as if no event relationships exists in the room.
+
+## Potential issues
+
+According to the existing specs, clients might ignore the `m.relates_to` in the encrypted payload of
+an event, causing rendering issues for them. Thus, clients that supports `m.room.relationship_encryption`
+MAY send the `m.relates_to` in both the cleartext part and the encrypted payload for rooms that has this state event.
+Clients that do so SHOULD inform their users of this, and SHOULD allow the users to choose whether they
+want to send the `m.relates_to` in the cleartext part.
```
richvdh, line 35: Obviously, server-side [aggregations](https://spec.matrix.org/v1.6/client-server-api/#aggregations) and the `/relationships` API will not work any more with this change.

These exist for a reason: clients are heavily reliant on them for things like finding the reactions and threads which relate to a particular event.

It seems to me that these are "Potential issues" which need discussing here.

proposals/4019-encrypted-relationships.md
```diff
@@ -0,0 +1,55 @@
...
+the cleartext part. Like `m.room.encryption`, once turned on, it cannot be turned off.
+
+### Client behaviour
+
+If the state event `m.room.relationship_encryption` exists or existed in the room, clients SHOULD send the
+`m.relates_to` that appears in any encrypted event in the encrypted payload. In received encrypted events,
+clients SHOULD ignore the `m.relates_to` that appears in the cleartext part of the event if an `m.relates_to`
+appears in the encrypted payload. Clients SHOULD NOT trust the results obtained from the
+[relationships API](https://spec.matrix.org/v1.6/client-server-api/#relationships-api)
+in the client-server API, and SHOULD NOT obtain event relationships from those API.
```
clokep, line 22: I think this will break any relations sent *before* the state event, e.g. consider:

```mermaid
flowchart TD
    A(Event A: m.room.encrypted) --> B(Event B: m.room.encrypted) --> C(Event C: m.room.relationship_encryption) --> D(Event D: m.room.encrypted)
    B -.->|Cleartext: m.thread| A
    D -.->|Encrypted: m.thread| A
```

Should this message from event `B` be discarded? In my opinion, no, but the text implies it should be.

Note that I think this issue might not exist for reactions since those aren't sent as encrypted events today.

proposals/4019-encrypted-relationships.md
tusooa: For my use case (1-to-1 chats and small groups of at most 10 people), aggregations from server side just do not help much. Most used relation is reply, second most editing, and tiny little bits of reactions. Reply and editing chains wouldn't be spanned across like 100s of events.

The sole purpose of this is to sacrifice the ability for the server to track relationships in order to achieve better privacy.

Made this explicit in the potential issues section.

proposals/4019-encrypted-relationships.md
tusooa: The cleartext part is ignored ONLY IF it has an encrypted `m.relates_to`. Clarified that in the doc.

clokep: Yeah, that seems fine then. 👍

tusooa:
Regarding implementation, [libkazv](https://lily-is.land/kazv/libkazv) currently sends all outbound relations encrypted in encrypted rooms.

proposals/4019-encrypted-relationships.md
FSG-Cat: @richvdh A somewhat recent discussion in Client Dev Room revealed no matrix clients use aggregations for Reactions. If a client does nobody was able to present one in that conversation. (Date of Discussion is Feburary 21nd of 2023. According to Central European Summer Time.)

So the concern that Reactions would break is if that conversation was accurate incorrect for most users. 

Due to that reactions wont break a MSC like this is safe to move forward for Reactions atleast if that conversation was accurate. As we are trying to prove a negative in this case its up to whoever wants to do the work to find a counterexample.

richvdh: You're right, reactions are not aggregated, I'd forgotten. Nevertheless the concern remains around other relation types (especially edits).

FSG-Cat: @clokep is correct.

deepadmax: The `key` field for reactions makes sense to encrypt in private conversations. However, I think that fields denoting the relationship itself should always be unencrypted. It doesn't help with just server aggregation. Bots and clients will have it way easier getting relevant events on demand rather than needing a full sync cached and searched through. It could be made the default and no extra toggle (`m.room.relationship_encryption`) would be needed; if a room is encrypted, all other relationship fields are stored inside the encrypted content, whereas the `event_id` and `rel_type` remain outside, unencrypted.

tusooa: > The `key` field for reactions makes sense to encrypt in private conversations. However, I think that fields denoting the relationship itself should always be unencrypted. It doesn't help with just server aggregation. Bots and clients will have it way easier getting relevant events on demand rather than needing a full sync cached and searched through. It could be made the default and no extra toggle (`m.room.relationship_encryption`) would be needed; if a room is encrypted, all other relationship fields are stored inside the encrypted content, whereas the `event_id` and `rel_type` remain outside, unencrypted.

I am talking about the *possibility* for clients to *not* rely on servers to keep record of event relationships. And it's *optional*. Clients that do want to rely on servers can simply not support it.

PR #2291, opened 2019-09-14
MSCxxxx: Configuration to Control Crawling

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/crawling/proposals/xxxx-configuration-to-control-crawling.md)

PR #4021, opened 2023-05-28
MSC4021: Archive client controls

jonaharagon:
This proposal solves these problems:

- https://github.com/matrix-org/matrix-public-archive/issues/47
- https://github.com/matrix-org/synapse/issues/14127

### DIFF
--- a/proposals/4021-archive-controls.md
+++ b/proposals/4021-archive-controls.md
@@ -1,0 +1,36 @@
+# MSC4021: Archive client controls
+
+The creation of archive.matrix.org indicates that search engine indexing of public Matrix rooms is a goal, but more
+granular control over how rooms should be indexed and displayed in search engine results should be granted to room
+admins.
+
+The current solution determines indexing eligibility based on `world_readable` `public` history visibility, but this is
+not an ideal solution because these settings only imply world readability within regular Matrix clients to most users,
+as opposed to the wider internet. Most alternative social media platforms provide separate settings for profile
+visibility and search engine indexing, for example.
+
+
+## Proposal
+
+Add an `m.room.archive_controls` state event where you can specify information about if and how you would like your
+room to be crawled. The room directory must relay this information to clients.
+
+| key | type | value | description | required
+|--|--|--|--|--
+| `robots` | [string] | Valid [robots meta rules](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives) | A list of rules which should be included in a `robots` meta tag and/or [HTTP header](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag-implementation) by public-facing clients. e.g. `["noarchive"]` or `["noindex", "nofollow"]`.
+| `via` | string | Hostname | A hostname which should be set as the canonical archive URL. e.g. `"archive.matrix.org"`.
+
+Public-facing clients like [matrix-public-archive](https://github.com/matrix-org/matrix-public-archive) should validate
+these rules before returning them in a response.
+
+When `via` is specified, the client should return a [rel=canonical link element](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method)
+and/or a [rel=canonical HTTP header](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-header-method)
+with the response pointing to the archive URL on the specified hostname. This prevents the Matrix.org public archive
+from returning duplicate content or taking precedence in search results over an organization's self-hosted archive.
+
+For example, if `via` is set to `"archive.example.net"` in `#main:example.net`, the page at
+https://archive.matrix.org/r/main:example.net/date/2023/05/28 should return this HTTP header:
+
+```
+Link: <https://archive.example.net/r/main:example.net/date/2023/05/28>; rel="canonical"
+```


proposals/4021-archive-controls.md
```diff
@@ -0,0 +1,36 @@
+# MSC4021: Archive client controls
```
MTRNord, line 1: https://github.com/matrix-org/matrix-spec-proposals/pull/2291

Should be added as an alternative.

Also this likely would need to depend on fed peeking since currently you need to join a room to access the info which some people may find bad.

MTRNord (commented):
I believe the alternative linked in my comment may be more useful and this should require fed peeking to exist to actually work as intended.

proposals/4021-archive-controls.md
uhoreg: I don't think that 2291 is an alternative;  I think the goals are different.  2291 indicates whether the bot is allowed to crawl the room, whereas it looks like the intent for this one is to communicate to search engines whether they are allowed to index the room.  For example, I might want my room available on archive.matrix.org, but I may not want Google to index it and present it in search results.

proposals/4021-archive-controls.md
```diff
@@ -0,0 +1,36 @@
...
+Public-facing clients like [matrix-public-archive](https://github.com/matrix-org/matrix-public-archive) should validate
+these rules before returning them in a response.
+
+When `via` is specified, the client should return a [rel=canonical link element](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method)
+and/or a [rel=canonical HTTP header](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-header-method)
+with the response pointing to the archive URL on the specified hostname. This prevents the Matrix.org public archive
+from returning duplicate content or taking precedence in search results over an organization's self-hosted archive.
+
+For example, if `via` is set to `"archive.example.net"` in `#main:example.net`, the page at
+https://archive.matrix.org/r/main:example.net/date/2023/05/28 should return this HTTP header:
```
uhoreg, line 32: This seems to assume that all archivers will have the same URL format, which may not be true.  If they all run matrix-public-archive, then that may be, but it's possible that some other archiving software may use a different format.

proposals/4021-archive-controls.md
MTRNord: Imho the indexing is already a form of crawling a room. That's my reasoning. And the other msc can also be used for this case imho. It's a little more generic than this one

jonaharagon: There is a desire (https://github.com/matrix-org/matrix-public-archive/issues/47#issuecomment-1536938601) to have the room directory API include this sort of information directly, which is why I'm not sure 2291 will work here. I edited the doc to expand upon this and add 2291 as an alternative.

proposals/4021-archive-controls.md
jonaharagon: An alternative could be to have `via` be a full URI, like `https://archive.example.net/r/main:example.net`, and then https://archive.matrix.org/r/main:example.net/date/2023/05/28 would return:

```
Link: <https://archive.example.net/r/main:example.net>; rel="canonical"
```

It would miss out on features like date pagination, although it now occurs to me that for the purposes of web indexing, this might actually be preferable behavior?

jonaharagon: The problem with this alternative is that it might be more difficult for the self-hosted client at archive.example.net to parse and _not_ include this canonical link header, because I don't think it would be ideal for the canonical archive to return this header. So I don't know, maybe that's something to leave up to client interpretation, maybe a standard URL format should be part of the spec? 🤷‍♂️

proposals/4021-archive-controls.md
```diff
@@ -0,0 +1,52 @@
...
+## Proposal
+
+Add an `m.room.archive_controls` state event where you can specify information about if and how you would like your
+room to be crawled. The [/publicRooms API](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientv3publicrooms)
+must relay this information to clients.
+
+| key | type | value | description | required
+|--|--|--|--|--
+| `archive` | boolean | | Whether the room should be included in room directory listings which are indended to be viewed by the public |
+| `robots` | [string] | Valid [robots meta rules](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives) | A list of rules which should be included in a `robots` meta tag and/or [HTTP header](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag-implementation) by public-facing clients. e.g. `["noarchive"]` or `["noindex", "nofollow"]`.
```
MadLittleMods, line 22: For the Matrix Public Archive, there are kind of two things to consider:

 - Whether you want to show up in the archive at all (display)
 - Whether you want to allow search engines to index that content (indexing)

The `robots` field definately covers the search engine indexing decision by being able to opt out with `noindex`

For the display decision, it's less clear whether `robots` can cover it. But `noarchive` sounds pretty decent just by name and also because of what it means:

> `noarchive`
>
> Requests the search engine not to cache the page content.
>
> *-- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name#other_metadata_names*

And the Matrix Public Archive really just allows you to view a public Matrix room with some potential caching on top (it doesn't store anything). But this might be an overloaded usage of `noarchive` since caching is not the same as displaying which the archive also does.


Depending on the answer here, the `archive` field may be redundant compared to what can be specified in `robots`

proposals/4021-archive-controls.md
```diff
@@ -0,0 +1,52 @@
...
+
+The current solution determines indexing eligibility based on `world_readable` `public` history visibility, but this is
+not an ideal solution because these settings only imply world readability within regular Matrix clients to most users,
+as opposed to the wider internet. Most alternative social media platforms provide separate settings for profile
+visibility and search engine indexing, for example.
+
+
+## Proposal
+
+Add an `m.room.archive_controls` state event where you can specify information about if and how you would like your
```
MadLittleMods, line 15: `m.room.archive_controls` feels very specific to the archive use case and we may want to be more generic.

For example, people building a blog or forum on Matrix would use similar `robots` controls (see other beyond chat applications for Matrix)

proposals/4021-archive-controls.md
MadLittleMods: If we wanted something specific to the Matrix Public Archive format, we could use an event type scoped to the sub-domain like `org.matrix.archive.canonical` to convey this information.

proposals/4021-archive-controls.md
jonaharagon: Maybe, but I didn't want this to be confused for controls over Matrix chat/integration bots, which this really isn't. It's more of a control over a specific class of clients in my mind, which I wasn't sure how to refer to.

bkil:
Neither MSC4021, nor MSC2291 copes with this to my liking.

As mentioned on matrix, I'd use more granular controls and it would be ideal to avoid the overloaded term `archiving`. I'd introduce the following levels in decreasing level of openness:

- Redistributable (specify exact copyright/copyleft terms)
- Allowed to process via statistical inference engines
- Long term archiving/mirroring/backup with anonymization
- Long term archiving/mirroring/backup (can store indefinitely and even versioned in systems such as MementoWeb, must document mechanism how a user can get their UGC removed from all such external archives),
- Ephemerally short term cacheable by a search engines (i.e., for search result preview purposes)
- Indexing and spidering allowed for search only
- Specific hotlinks indexable for search only, but not allowed to spider - **default level during room creation**
- Contextually hotlinkable preview while allowing for a bit of room chat context and interactive room browsing by humans (~statically rendered nojs client)
- Hotlinkable preview (~unlisted individual messages),
- For human eyes present in the room & over admin-authorized controlled bridges/relays
- For present human eyes-only
- ...

MadLittleMods:
@bkil Please use a thread off the diff for discussions and separate topics.

alphapapa:
IMHO this proposal is misguided, because a room with world-readable history can have its history read by any client, which would be free to ignore such "controls."  That is, these would not be controls, but merely expressed preferences.  They would likely give users or room admins a false sense of security, because while they may have set a "control" to prevent indexing of their room's history, any party could be doing so with just a few lines of code, even connecting it to their own `archive.matrix.org`-like instance and feeding it to search engines.

Instead, we should seek to make the consequences of room settings as clear as possible to users and admins.

proposals/4021-archive-controls.md
MadLittleMods: > For example, I might want my room available on archive.matrix.org, but I may not want Google to index it and present it in search results.

@uhoreg With MSC2291, I think this could be achieved with a mix of `messages` and `log`. Am I misinterpreting?

```json
{
  "*": {
    "messages": false,
    "log": true
  },
}
```

> - `messages`: (boolean) **whether the bot is allowed to index the room's
>   messages**.  Default: `true` if `m.room.history_visibility` is
>   `world_readable`, and `false` otherwise.
> - `log`: (boolean) **whether the bot is allowed to display logs of the room** to
>   users.  This will be `false` if `messages` is `false`.  Default: `true` if
>   `m.room.history_visibility` is `world_readable`, and `false` otherwise.

uhoreg: In 2291, `messages` is intended to indicate whether the bot itself is allowed to index messages, whereas this proposal, whereas this proposal is intended to communicate preferences to other crawlers that crawl the bot's logs.  This may be able to be done with an addition to 2291 (e.g. add a new property), but 2291 itself doesn't do this.

MadLittleMods: Feels like there might not be any difference between the bot itself (Matrix Public Archive) and a different crawler that crawls the bots logs (a search engine). They're both accessing the same information (same or derived) and feels like `messages` from MSC2291 to control indexing of messages covers that. In other words, if `messages: false`, the archive can't index messages and neither can search engines.

Basically, any bot preference should probably be passed down for other bots to follow?

I think if it's a wildcard `*`, it should apply to downstream bots. It's less clear how things should flow if someone specified an app. Perhaps it wouldn't flow in the specific app case but could use the `*` rules to govern how search engines look at it.

And maybe we want to define some generic "search_engines" key for example since it might be common. But not all of the preferences are applicable since we can't pass along all of this preference detail seamlessly (impedance mismatch).

PR #4022, opened 2023-05-31
MSC 3381: Added new attribute show_participants

arnfriedgriesert:
Added a new attribute informing clients that they not only present the result of polls but also how each individual participant has voted. This can be very useful in many situation where room members will start a vote.

### DIFF
--- a/proposals/3381-polls.md
+++ b/proposals/3381-polls.md
@@ -41,6 +41,7 @@ To start a poll, a user sends an `m.poll.start` event into the room. An example 
     "m.poll": {
       "kind": "m.disclosed",
       "max_selections": 1,
+      "show_participants": true,
       "question": {
         "m.text": [{"body": "What should we order for the party?"}]
       },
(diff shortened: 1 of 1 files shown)


proposals/3381-polls.md
```diff
@@ -41,6 +41,7 @@ To start a poll, a user sends an `m.poll.start` event into the room. An example
     "m.poll": {
       "kind": "m.disclosed",
       "max_selections": 1,
+      "show_participants": true,
```
Johennes, line 44: This feature generally seems valuable to me but I think we might need to reconsider naming – possibly holistically – because the meaning of this option seems easy to confuse with the function of `m.disclosed` / `m.undisclosed`. It's already slightly confusing to me right now that `m.undisclosed` also "discloses" results but only once the poll is ended and I feel like adding another option related to visibility into this might cause further confusion. @turt2live curious what you think about this.

turt2live:
Thanks for the suggestion! Thinking about this sort of change ended up being more involved/complicated than I originally thought - I've moved it to [a thread](https://github.com/matrix-org/matrix-spec-proposals/pull/3381/files#r1212261986) on the MSC itself to get wider input from people who might be subscribed to that MSC specifically. 

Largely, my concern here is that the MSC leaves this as an implementation detail, but that might not be the correct way to approach this. The leading question being whether it's a "problem" if some clients show per-voter results while others don't, in the same room. For many environments this might not be a concern as they control the clients being used, but for a general purpose room there could be dozens of different clients and therefore dozens of different behaviours. 

I'm going to close this for now, but please feel free to comment on the thread. I'd generally encourage implementing this change regardless of the outcome on the thread to help build some potential use cases (when it feels important to know who voted for what, and when it feels important to not know). If the thread ends up wanting this feature, we'll incorporate this PR or something like it.

proposals/3381-polls.md
turt2live: @Johennes the naming definitely needs work on the MSC. "Disclosed" and "undisclosed" were what the MSC settled on after many hours of discussion, but do indeed still feel 'wrong' in many respects. 

If folks have ideas on better naming to match the semantics, please leave them as comments on the MSC.

PR #4024, opened 2023-06-01
Clarify how to make use of unstable prefixes in your MSC

anoadragon453:
This PR adds a section to the simple spec guide that gives some background to MSC authors on how to effectively make use of unstable and stable prefixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,27 +184,29 @@ like help with writing spec PRs, feel free to join and ask questions in the
 
 #### Stable and unstable prefixes
 
-*Unstable* prefixes are the namespaces which are used by implementations while
-an MSC is not yet accepted. For instance, an MSC can propose that a `m.space`
+"Unstable prefixes" are the namespaces which are used by implementations while
+an MSC is not yet accepted.
+
+For instance, an MSC might propose that a `m.space`
 event type or an `/_matrix/client/v1/account/whoami` endpoint should exist.
-However, an implementation of that same MSC cannot use a *stable* prefix (in the
-case, `/_matrix/client/` or `m.`) until the MSC has been accepted, as the
-underlying design may change at any time; the design is
+However, implementations cannot use these *stable* identifiers until the MSC
+has been accepted, as the underlying design may change at any time; the design is
 *unstable*.
 
-Typically MSCs will define `org.matrix.msc1234` (using the real MSC number once
-known) as an *unstable* prefix. For the above examples, this would mean using
-`org.matrix.msc1234.space` and
-`/_matrix/client/unstable/org.matrix.msc1234/account/whoami` to allow for
-breaking changes between edits of the MSC itself, or to not clash with another
-MSC that's attempting to add the same stable identifiers. The unstable
-prefixes, in this case, are `org.matrix.msc1234` and `/_matrix/client/unstable`
-respectively. It is also fine to use more traditional forms of namespace
-prefixes, such as `com.example.*` (e.g. `com.example.space`).
+Instead, an MSC can define a namespace such as `org.matrix.msc1234` (using the real
+MSC number once known) which is added to the stable identifier, allowing for
+breaking changes between edits of the MSC itself, and preventing clashes with other
+MSCs that might attempt to add the same stable identifiers. 
 
-Note: not all MSCs need to make use of unstable prefixes. They are only needed
-when defining new endpoints, field names, etc. in your MSC, and serve to allow
-implementations of your MSC to exist in the wild before your MSC is accepted.
+For the above examples, this would mean using `org.matrix.msc1234.space` and
+`/_matrix/client/unstable/org.matrix.msc1234/account/whoami`. It is also fine to
+use more traditional forms of namespace prefixes, such as `com.example.*` (e.g. `com.example.space`).
+
+Note: not all MSCs need to make use of unstable prefixes. They are only needed if
+implementations of your MSC need to exist in the wild before your MSC is accepted,
+*and* the MSC defines new endpoints, field names, etc.
+
+#### Unstable feature flags
 
 It is common when implementing support for an MSC that a client may wish to check
 if the homeserver it is communicating with supports an MSC.
@@ -245,10 +247,8 @@ At this point, it may be best to wait until a new spec version is released with
 your changes. Homeservers that support the changes will eventually advertise
 that spec version under `/versions`, and your client can check for that.
 
-But if you really can't wait, then there is another option.
-
-This is often solved by having the homeserver tell clients that it supports
-stable prefixes, using yet another `unstable_features` flag:
+But if you really can't wait, then there is another option: the homeserver can
+tell clients that it supports *stable* indentifiers for the new MSC, using yet another `unstable_features` flag:
 
 ```json5
 {
@@ -260,12 +260,12 @@ stable prefixes, using yet another `unstable_features` flag:
 ```
 
 If a client sees that `org.matrix.msc1234.stable` is `true`, it knows that it
-can start using stable prefixes, and the homeserver will accept and act on
+can start using stable identifiers for the new MSC, and the homeserver will accept and act on
 them accordingly.
 
 While the general pattern of using the text ".stable" has emerged from previous
 MSCs, you can pick any name you like. You need only to clearly state their
-meaning, usually udner an "Unstable prefixes" header in your MSC.
+meaning, usually under an "Unstable prefixes" header in your MSC.
 
 See
 [MSC3827](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3827-space-explore.md#unstable-prefix)
--- a/README.md
+++ b/README.md
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
 
 ### Other useful information
 
-#### Unstable prefixes
+#### Stable and unstable prefixes
 
-*Unstable* prefixes are the namespaces which are used before an MSC has
-completed FCP (see above). While the MSC might propose that a `m.space` or
-`/_matrix/client/v1/account/whoami` endpoint should exist, the implementation
-cannot use a *stable* identifier such as `/v1/` or `m.space` prior to the MSC
-being accepted: it needs unstable prefixes.
+*Unstable* prefixes are the namespaces which are used by implementations while
+an MSC is not yet accepted. For instance, an MSC can propose an a `m.space`
+event type or an `/_matrix/client/v1/account/whoami` endpoint should exist.
+However, an implementation of that same MSC cannot use a *stable* prefix (in the
+case, `/_matrix/client/` or `m.`) until the MSC has been accepted, as the
+underlying design may change at any time; the design is
+*unstable*.
 
-Typically for MSCs, one will use `org.matrix.msc0000` (using the real MSC
-number once known) as a prefix. For the above examples, this would mean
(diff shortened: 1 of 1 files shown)


ShadowJonathan (approved):

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+clients that they can start using *stable* prefixes, before the homeserver
+advertises support for a spec release which contains the new endpoint? One case
+where this poses a problem is when an MSC proposes new event types or fields.
+
+Events are immutable data. A client may wish to introduce your new feature, but
+if it sending events with unstable prefixes, eventually those events won't
+be supported by future clients, and thus that data will be effectively lost!
+
+This is often solved by having the homeserver telling clients that it supports
+stable prefixes for an MSC with yet another `unstable_features` flag:
```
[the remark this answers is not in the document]
clokep: Isn't it really just about how quickly you want it rolled out though? I think *any* MSC *could* use it for clients to know earlier than new spec release time to use it.

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+respectively.
+
+Note: not all MSCs need to make use of unstable prefixes. They are only needed
+when defining new endpoints, field names, etc. in your MSC, and serve to allow
+implementations of your MSC to exist in the wild before your MSC is accepted.
+
+In the case of some MSCs, implementation types may need a way to check that another
+implementation type supports this feature. A common case is when a client that has
+implemented an MSC needs to check whether the homeserver it is talking to has
+also implemented support. Typically, this is handled by the MSC defining an
```
clokep, line 211: I found this too abstract: I was left wondering what an "implementation types" is?

```suggestion
It is common when implementing support for an MSC that a client may wish to check
if the homeserver it is communicating with supports an MSC.
Typically, this is handled by the MSC defining an
```

I guess theroetically the same thing could be done for homeserver to {homeserver, identity servers, push servers}, but those (mostly) don't define a way to check this.

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
-Typically for MSCs, one will use `org.matrix.msc0000` (using the real MSC
-number once known) as a prefix. For the above examples, this would mean
-`org.matrix.msc0000.space` and
-`/_matrix/client/unstable/org.matrix.msc0000/account/whoami` to allow for
-breaking compatibility changes between edits of the MSC itself, or indeed
-another competing MSC that's attempting to add the same identifiers.
+#### Stable and unstable prefixes
+
+*Unstable* prefixes are the namespaces which are used by implementations while
+an MSC is not yet accepted. For instance, an MSC can propose an a `m.space`
```
turt2live, line 188: > propose an a

should be "propose to add a" or something?

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+*unstable*.
+
+Typically MSCs will define `org.matrix.msc1234` (using the real MSC number once
+known) as an *unstable* prefix. For the above examples, this would mean using
+`org.matrix.msc1234.space` and
+`/_matrix/client/unstable/org.matrix.msc1234/account/whoami` to allow for
+breaking changes between edits of the MSC itself, or to not clash with another
+MSC that's attempting to add the same stable identifiers. The unstable prefixes
+in this case of `org.matrix.msc1234` and `/_matrix/client/unstable`
+respectively.
```
turt2live, line 202: might be good to mention here that you can invent your own, but that we also don't mind people littering the `org.matrix.msc*` namespace.

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+  "unstable_features": {
+    "org.matrix.msc1234": true
+  }
+}
+```
+
+With a value of `true` indicating that the feature is supported, and `false`
+or lack of the field altogether indicating the feature is not supported.
+
+According to the spec process, once an MSC has been accepted, implementations
```
turt2live, line 227:
```suggestion
According to the spec process, once an MSC has been accepted (passed a `merge` FCP), implementations
```

... or we should mention this somewhere in the doc if we don't already

README.md
```diff
@@ -182,20 +182,81 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+are allowed to switch to *stable* prefixes (i.e. `m.`). However, the MSC is
+still not yet part of a released spec version. How can a homeserver advertise to
+clients that they can start using *stable* prefixes, before the homeserver
+advertises support for a spec release which contains the new endpoint? One case
+where this poses a problem is when an MSC proposes new event types or fields.
```
turt2live, line 232: are event types the common case? I thought it was endpoints, generally speaking. Clients prefer to just know rather than rely on 404-style semantics.

There is a case where the event types mean something to the server, but aren't affected by the room version, though this is theoretically rare - an example being threads (which was nothing to do with event types...).

turt2live (commented):
generally lgtm with the verbiage changes noted already

README.md
anoadragon453: Well, we don't want them to start encroaching on `org.matrix.msc<not their MSC number>` ideally, right?

turt2live: it's more than people can use `com.example.feature` instead of `org.matrix.*` if they want to.

README.md
clokep: I suspect it should be event properties, not types? Anything that gets persisted into the DAG. I've seen this be an annoying issue for event types (polls, etc.); relation types (threads is the only example I know of since the others always used stable); or event properties (intentional mentions, etc).

anoadragon453: I've rewritten this in 681173c4, let me know how it looks!

README.md
anoadragon453: Changed in 3eba07f420a5cae1c6ea331ffaac9eda0813b792.

README.md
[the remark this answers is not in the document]
clokep: Seems reasonable to me!

README.md
```diff
@@ -182,21 +182,95 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+`/_matrix/client/unstable/org.matrix.msc1234/account/whoami` to allow for
+breaking changes between edits of the MSC itself, or to not clash with another
+MSC that's attempting to add the same stable identifiers. The unstable
+prefixes, in this case, are `org.matrix.msc1234` and `/_matrix/client/unstable`
+respectively. It is also fine to use more traditional forms of namespace
+prefixes, such as `com.example.*` (e.g. `com.example.space`).
+
+Note: not all MSCs need to make use of unstable prefixes. They are only needed
+when defining new endpoints, field names, etc. in your MSC, and serve to allow
+implementations of your MSC to exist in the wild before your MSC is accepted.
```
richvdh, line 207:
```suggestion
Note: not all MSCs need to make use of unstable prefixes. They are only needed if
implementations of your MSC need to exist in the wild before your MSC is accepted,
*and* the MSC defines new endpoints, field names, etc.
```

README.md
```diff
@@ -182,21 +182,95 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
-being accepted: it needs unstable prefixes.
-
-Typically for MSCs, one will use `org.matrix.msc0000` (using the real MSC
-number once known) as a prefix. For the above examples, this would mean
-`org.matrix.msc0000.space` and
-`/_matrix/client/unstable/org.matrix.msc0000/account/whoami` to allow for
-breaking compatibility changes between edits of the MSC itself, or indeed
-another competing MSC that's attempting to add the same identifiers.
-
+#### Stable and unstable prefixes
```
richvdh, line 185:
```suggestion
####  Unstable prefixes
```

README.md
```diff
@@ -182,21 +182,95 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+}
+```
+
+If a client sees that `org.matrix.msc1234.stable` is `true`, it knows that it
+can start using stable prefixes, and the homeserver will accept and act on
+them accordingly.
+
+While the general pattern of using the text ".stable" has emerged from previous
+MSCs, you can pick any name you like. You need only to clearly state their
+meaning, usually udner an "Unstable prefixes" header in your MSC.
```
richvdh, line 268:
```suggestion
meaning, usually under an "Unstable prefixes" header in your MSC.
```

README.md
```diff
@@ -182,21 +182,95 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+
+Typically MSCs will define `org.matrix.msc1234` (using the real MSC number once
+known) as an *unstable* prefix. For the above examples, this would mean using
+`org.matrix.msc1234.space` and
+`/_matrix/client/unstable/org.matrix.msc1234/account/whoami` to allow for
+breaking changes between edits of the MSC itself, or to not clash with another
+MSC that's attempting to add the same stable identifiers. The unstable
+prefixes, in this case, are `org.matrix.msc1234` and `/_matrix/client/unstable`
+respectively. It is also fine to use more traditional forms of namespace
+prefixes, such as `com.example.*` (e.g. `com.example.space`).
```
richvdh, line 203: This stuff applies to *all* identifiers, not just those with an identifiable "prefix". (Consider, in particular, properties within event bodies or other JSON structures).

So, I don't really think we should be talking about "stable prefixes" at all.  Instead, we can talk about "stable identifiers":

```suggestion
"Unstable prefixes" are the namespaces which are used by implementations while
an MSC is not yet accepted.

For instance, an MSC might propose that a `m.space`
event type or an `/_matrix/client/v1/account/whoami` endpoint should exist.
However, implementations cannot use these *stable* identifiers until the MSC
has been accepted, as the underlying design may change at any time; the design is
*unstable*.

Instead, an MSC can define a namespace such as `org.matrix.msc1234` (using the real
MSC number once known) which is added to the stable identifier, allowing for
breaking changes between edits of the MSC itself, and preventing clashes with other
MSCs that might attempt to add the same stable identifiers. 

For the above examples, this would mean using `org.matrix.msc1234.space` and
`/_matrix/client/unstable/org.matrix.msc1234/account/whoami`. It is also fine to
use more traditional forms of namespace prefixes, such as `com.example.*` (e.g. `com.example.space`).
```

Likewise, the other references to "stable prefixes" below need rephrasing.

README.md
```diff
@@ -182,21 +182,95 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+
+Note: not all MSCs need to make use of unstable prefixes. They are only needed
+when defining new endpoints, field names, etc. in your MSC, and serve to allow
+implementations of your MSC to exist in the wild before your MSC is accepted.
+
+It is common when implementing support for an MSC that a client may wish to check
```
richvdh, line 209:
```suggestion
#### Unstable feature flags

It is common when implementing support for an MSC that a client may wish to check
```

richvdh (changes requested):

README.md
```diff
@@ -184,19 +184,94 @@ like help with writing spec PRs, feel free to join and ask questions in the
...
+
+```json5
+{
+  "unstable_features": {
+    "org.matrix.msc1234": true
+  }
+}
+```
+
+With a value of `true` indicating that the feature is supported, and `false`
```
richvdh, line 227:
```suggestion
... with a value of `true` indicating that the feature is supported, and `false`
```

richvdh (approved):
looks great!

PR #4026, opened 2023-06-06
MSC4026: Allow `/versions` to optionally accept authentication

H-Shay:
A proposal for [`_matrix/client/versions`](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientversions) to optionally accept authentication in order to facilitate advertising to clients experimental features which are enabled per-user.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/shay/opt_auth_versions/proposals/4026-optional-authed-versions.md)

Related: https://github.com/matrix-org/matrix-spec/issues/1532

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4026#issuecomment-1682533812)


proposals/4026-optional-authed-versions.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Proposal
+
+The proposal to remedy this is to make `/_matrix/client/versions` optionally accept authentication, and ask clients
+to use the authenticated version when determining which experimental features are enabled. 
+
+## Potential issues
+
+This does raise the question of what the non-authenticated version of `/_matrix/client/versions` should return with 
+regard to unstable features if the proposal is accepted.
```
MadLittleMods, line 24: If the feature is only enabled for certain users, it seems like the unauthenticated version should just indicate no support for the given feature.

ara4n:
lgtm!

erikjohnston:
Since nobody thinks this is a *bad* idea......

@mscbot fcp merge

proposals/4026-optional-authed-versions.md
```diff
@@ -0,0 +1,35 @@
+## Introduction
+
+Synapse is implementing the ability to turn some unstable features on per-user. Once this is 
+implemented, certain experimental features will be available to be enabled per-user via the [Admin API](https://matrix-org.github.io/synapse/latest/usage/administration/admin_api/index.html).
```
uhoreg, line 4: While I don't object, in principle, to making `/versions` optionally accept authentication, I'm a bit concerned about the proposed usage.  If the usage is only for experimental features, how would clients detect support for stable features?  I'd expect that the answer to that would be through the `/capabilities` endpoint which seems kind of strange to use two completely different endpoints to detect whether a client can do something.  That means that when the client switches between unstable and stable implementations (or wants to support both), then it need to call two different endpoints.

I'd be more in favour of relaxing the restriction on `/capabilities` and distinguishing between server support for a feature and whether a given client is allowed to make use of the feature: `/capabilities` should not be used to indicate  server support for an unstable feature (which should still be indicated using `/versions`), but may be used to indicate whether a client can use it.

proposals/4026-optional-authed-versions.md
```diff
@@ -0,0 +1,35 @@
+## Introduction
```
uhoreg, line 1:
```suggestion
# MSC4026: Allow /versions to optionally accept authentication

## Introduction
```

Please include the MSC name in the document so that the title stays with the file after it gets merged.

proposals/4026-optional-authed-versions.md
clokep: > In practice, both clients and servers are taking more than 1 release cycle to get any given release implemented anyways, so it does not feel like it's materially helping to have stable features so early in the process.

Isn't this *why* it is important to have stable features before the spec version though? Since clients/servers aren't keeping up with the spec versions, but want to be able to use new features quickly.

turt2live: They don't appear to be using those new features earlier than release, or at least they can't if the support is 1 release cycle behind.

uhoreg: > but for truly stable (ie: merged) features I strongly prefer we don't have flags at all. Feature support is ultimately determined by the spec version.

But the point of this MSC is for selectively enabling/disabling a feature for certain users, not about whether the server has any support for a feature, and it's not about stable-but-not-merged features.

If a feature is being selectively enabled for certain users while it is unstable, my assumption is that it will continue to offer that feature only to certain users after it is stable/merged.  So clients will need some way of determining whether they have access to the feature.  If they just look at `/versions`, they can see if the server supports the given spec version, and know whether the server supports the feature, but they'd need to find out whether they have access, presumably by asking `/capabilities`.  And if clients will be using `/capabilities` in the stable version, then it seems strange to tell them to use something else in the unstable version.

turt2live: yea, for that particular case an unstable and stable capability pairing would indeed be better.

erikjohnston: > If a feature is being selectively enabled for certain users while it is unstable, my assumption is that it will continue to offer that feature only to certain users after it is stable/merged.

That was not the motivation behind this MSC TBH. The reason we want to selectively enable unstable features for certain users is so that we can test these unstable features with real accounts without risking other users finding it and relying on those features (i.e. we want to be able to enable stuff on matrix.org for certain users without risking the unstable features becoming "defacto stable").

uhoreg: Ah, OK.  That makes sense.  Thanks for the clarification.

Can this be clarified in the MSC as well?  Maybe something like:

```suggestion
implemented, certain experimental features will be available to be enabled per-user via the [Admin API](https://matrix-org.github.io/synapse/latest/usage/administration/admin_api/index.html).
The intention is to allow certain users to test the experimental feature without making it available to
all users before it is stable.
```

H-Shay: Sure, I have updated to reflect the suggestion.

proposals/4026-optional-authed-versions.md
H-Shay: Updated - thanks for pointing that out.

proposals/4026-optional-authed-versions.md
```diff
@@ -0,0 +1,39 @@
...
+changes being implemented this is no longer possible, as some experimental features may be enabled per-user. As the 
+`/_matrix/client/versions` endpoint does not require authentication there is no way to know which experimental features 
+are enabled - there is no access token that we can extract user info from to determine which unstable features are 
+currently enabled (as they may only be enabled for some users) - and thus there is no way to correctly communicate to 
+clients which experimental features are enabled.
+
+## Proposal
+
+The proposal to remedy this is to make `/_matrix/client/versions` optionally accept authentication, and ask clients
+to use the authenticated version when determining which experimental features are enabled.
```
anoadragon453, line 23: For clarity, nobody should walk away from this thinking that the correct solution is to not advertise *any* unstable features to unauthenticated users. Unstable features that affect clients before they are logged in, and older clients that will not know to authenticate to this endpoint, are still valid.

anoadragon453 (approved):

benkuly:
This would break server discovery, would't it?

From the current spec:

> Clients SHOULD validate that the URL points to a valid homeserver before accepting it by connecting to the [/_matrix/client/versions](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientversions) endpoint, ensuring that it does not return an error, and parsing and validating that the data conforms with the expected response format. If any step in the validation fails, then FAIL_ERROR.

anoadragon453:
@benkuly This MSC proposes *optional* authentication for this endpoint. Clients can still request `/_matrix/client/versions` without authentication to complete a server discovery check.

Please use a pull request comment when commenting on MSCs, to allow for collapsible threads of discussion.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1728

uhoreg:
Merged! :tada:

PR #3746, opened 2022-02-25
MSC3746: Render image data in reactions

AndrewRyanChama:
Rendered: https://github.com/AndrewRyanChama/matrix-doc/blob/archama/imagereacts/proposals/3746-image-reactions.md
Implementation: https://github.com/matrix-org/matrix-react-sdk/pull/7903
Signed-off-by: Andrew Ryan [andrewryanchama@clover.club](mailto:andrewryanchama@clover.club)

---

Please take a look and sanity check this approach is feasible.  I'm personally not worried about it, but decoupling what is displayed in the reaction with the aggregation key opens the door to potentially annoying situations.

If this isn't feasible, I can put together a proposal that rams everything into the aggregation key, but I don't like that very much either.

PR #4027, opened 2023-06-10
MSC4027: Custom Images in Reactions

sumnerevans:
A proposal to render MXC URIs in reaction keys as images.

[Rendered](https://github.com/beeper/matrix-spec-proposals/blob/custom-images-in-reactions/proposals/4027-custom-images-in-reactions.md)

### Clients

* Schildichat
  * Android: https://github.com/SchildiChat/SchildiChat-android/commit/85a26ae8bea7d1425019df31b05ff72541ad61dd
  * Desktop implementation (hasn't landed in a release yet): https://github.com/SchildiChat/matrix-react-sdk/commit/4d104fc87f8d32da679476967850c5f7a9c20c8a
* Beeper (all platforms)
* Element Web: https://github.com/matrix-org/matrix-react-sdk/pull/11087
* Element X Android: https://github.com/element-hq/element-x-android/pull/2160
* Cinny: https://github.com/cinnyapp/cinny/pull/686/commits/519f620cb7660b665602e3799bd63dff7e4382a6

### Bridges

* [mautrix-discord](https://github.com/mautrix/discord/): https://github.com/mautrix/discord/blob/00465bb71509b984614ed41fa4f147a9d6bd4b3a/example-config.yaml#L145-L147

-----

Supercedes #3746

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4027#issuecomment-1734153328)


proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,111 @@
...
+  "m.relates_to": {
+    "rel_type": "m.annotation",
+    "event_id": "$abcdefg",
+    "key": "mxc://matrix.org/VOczFYqjdGaUKNwkKsTjDwUa"
+  },
+  "shortcode": ":partyparrot:"
+}
+```
+
+## Potential issues
```
clokep, line 55: Can multiple MXC URLs point to the same image? Should those be deduplicated?

deepbluev7: It would be way too much effort to deduplicate this, so it shouldn't be required to do so.

clokep: Fair, but it should be acknowledged.

sumnerevans:
I don't think this deserves the `needs-implementation` label as it is already implemented across multiple clients and at least one bridge.

SpiritCroc:
Some extra details concerning SchildiChat:
- Android implementation: https://github.com/SchildiChat/SchildiChat-android/commit/85a26ae8bea7d1425019df31b05ff72541ad61dd
- Desktop implementation (hasn't landed in a release yet): https://github.com/SchildiChat/matrix-react-sdk/commits/4d104fc87f8d32da679476967850c5f7a9c20c8a
- We don't do anything with `shortcode` yet :upside_down_face:

proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,133 @@
...
+## Proposal
+
+This proposal suggests two changes to events with the `m.annotation` relation.
+
+1. If the `key` of an `m.annotation` relation is an MXC URI of an image, clients
+   should render the referenced image instead of the key text.
+
+   Detecting if the `key` is an MXC URI can be as sophisticated as the client
+   wants, but this proposal recommends checking if the string starts with
+   `mxc://`.
```
anoadragon453, line 33: I don't think this proposal needs to recommend validation for MXC URIs outside of what clients already need to do elsewhere. The spec already specifies the format of MXC URIs: https://spec.matrix.org/v1.7/client-server-api/#matrix-content-mxc-uris

If I were writing my own validator, I would additionally add a rule for checking for a `/` between the server name and media ID.

anoadragon453 (approved):
This LGTM at a high level.

Once we have a mechanism for image packs, it may be nice to link back to that pack from the reaction. But that change should live in the image packs MSC, instead of this one.

proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,133 @@
...
+Example custom image reaction event content
+
+```json
+"content": {
+  "m.relates_to": {
+    "rel_type": "m.annotation",
+    "event_id": "$abcdefg",
+    "key": "mxc://matrix.org/VOczFYqjdGaUKNwkKsTjDwUa"
+  },
+  "shortcode": ":partyparrot:"
```
anoadragon453, line 51: Should we enforce that shortcode's must begin and end with `:`? I think it'd be confusing to users if some clients started picking some other character.

proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. If the `key` of an `m.annotation` relation is an MXC URI of an image, clients
+   should render the referenced image instead of the key text.
+
+   Detecting if the `key` is an MXC URI can be as sophisticated as the client
+   wants, but this proposal recommends checking if the string starts with
+   `mxc://`.
+
+2. When the annotation's key is an MXC URI, a new (optional) `shortcode` key can
+   be added to the content of the event with a textual name for the image.
```
anoadragon453, line 36: We should put some constraints on allowed characters/length/type of this `shortcode` field.

proposals/4027-custom-images-in-reactions.md
anoadragon453, line 1: > ### Bridges
> 
> * [mautrix-discord](https://github.com/mautrix/discord/): https://github.com/mautrix/discord/blob/00465bb71509b984614ed41fa4f147a9d6bd4b3a/example-config.yaml#L145-L147

The [implementation in mautrix-discord](https://github.com/mautrix/discord/blob/09a94010d8f6ceb3d799ba68c03c87ceb261d1da/portal.go#L1527-L1529) doesn't actually appear to implement this MSC?

anoadragon453:
I played around with https://github.com/matrix-org/matrix-react-sdk/pull/11087 and it looks good! Removing the `needs-implementation` tag.

proposals/4027-custom-images-in-reactions.md
sumnerevans: The logic here is a bit weird. That code is for when the feature is *not* enabled (the fallback is to use the shortcode).

If it's enabled, then the MXC URI is used: https://github.com/mautrix/discord/blob/09a94010d8f6ceb3d799ba68c03c87ceb261d1da/portal.go#L1517

which is computed here: https://github.com/mautrix/discord/blob/09a94010d8f6ceb3d799ba68c03c87ceb261d1da/portal.go#L1471-L1483

proposals/4027-custom-images-in-reactions.md
Johennes: With just the MXC URI, how would clients verify that the media is actually an image and not a video?

sumnerevans: As I mentioned in this comment: https://github.com/matrix-org/matrix-react-sdk/pull/11087#discussion_r1248111093, the matrix-react-sdk seems to generally just check that the string starts with `mxc://`. I will just remove the recommendation from the MSC.

As far as verifying that the media is an image or not, I think that is not a concern of this MSC since it is possible to upload a video/file and put that MXC URI in a `m.room.message` with `msgtype=m.image` already.

proposals/4027-custom-images-in-reactions.md
sumnerevans: I'm not sure what appropriate constraints should be. It's unenforcable except by convention.

I think that we can possibly suggest that the shortcodes should only be printable ASCII characters, but that seems like about as far as we can go in restricting without potentially shooting ourselves in the foot with unanticipated edge cases.

proposals/4027-custom-images-in-reactions.md
sumnerevans: Added a "should" for that. It's not enforceable other than by convention.

proposals/4027-custom-images-in-reactions.md
anoadragon453: I see, thank you for the clarification!

proposals/4027-custom-images-in-reactions.md
greentore: Restricting shortcodes to ascii only is very anglocentric for no good reason.

sumnerevans: I'll just clarify that it must be a string, and call it good. I already cover that the shortcode should start and end with `:` as of https://github.com/matrix-org/matrix-spec-proposals/pull/4027/commits/5424ea803809d88cdd40585c5cff7ee34534b46c and I think that imposing any further restrictions should not be the responsibility of the spec.

anoadragon453: I'd like to point out that whatever shortcode grammar we come up with here will end up restricting MSCs that specify how emotes are defined and shared, like MSC4027. As we'll want to use the exact same shortcodes across both problem spaces, otherwise users will be confused.

MSC4027 is already loose enough to suggest that clients can define their own delimiters (e.g. `:`), which is already in conflict with this MSC, which requires a shortcode to **contain** a `:` at the beginning and end. Personally, I believe that if we're always requiring a shortcode begin and end with `:`, then we should just omit the `:`s from the shortcode, and require that shortcodes *not* contain a `:` character to reduce confusion.

I could also see the argument to reject `/` and `.` from being allowed, as that allows the shortcode to be URL-safe. If we ever include a shortcode in a URL, restricting these characters will reduce potential future headaches.

> Restricting shortcodes to ascii only is very anglocentric for no good reason.

@greentore indeed it would be nice to support Chinese, Japanese, Arabic, etc. characters in shortcodes. Though we do have to be careful about just allowing any unicode characters - in particular, "confusables", zero-width/invisible and control characters (such as right-to-left markers) can cause issues.

We could whitelist certain blocks of unicode to allow characters of different languages, but disallow potentially undesirable ones. Is such a whitelist already defined somewhere?

Finally, I feel that defining a reasonable length for this field to prevent spam/layout issues is still valid. Perhaps 64 characters? The latest standard emoji shortcode I'm aware of is `:hand_with_index_finger_and_thumb_crossed:` which - excluding the `:`s - is 40 characters long.

sumnerevans: I'm confused... MSC4027 is this MSC? What MSC are you referring to as being conflicting?

sumnerevans: Additionally, I would also like to know what you think would be a good way to enforce these restrictions. It seems somewhat of a burden to force servers to validate this as part of event auth, for example. And if it's not enforced at that level, it basically isn't enforced at all, and is all based on convention.

sumnerevans: I think that limiting it to something like 100 characters is reasonable enough.

sumnerevans: Does the following make sense?

> The `shortcode` must have a length of less than or equal to 100 bytes (including the colons at the beginning and end of the string). This restriction must be enforced by servers when sending reactions, but servers should not reject events coming across federation due to having too many bytes in the `shortcode` field.

anoadragon453: That sounds good to me.

> servers should not reject events coming across federation due to having too many bytes in the shortcode field.

I assume that homeservers would not reject these events, but may still opt to locally redact them.

sumnerevans: I added the above verbiage.

cloudrac3r:
This proposal is implemented in Out Of Your Element: https://gitdab.com/cadence/out-of-your-element/compare/c7ddf638dbd658057ec6a7a018a23b0693a97890~..044ccc08e06f11b48e16c7f7637d3c088008119a#diff-0f01ec9090ff5700793cc2f6399fb95cf3a7115f

It follows the spec and it works with Schildi and Cinny.

`shortcode` is not yet implemented. I will do that.

cloudrac3r (approved):

kimabjorkede:
It seems like Element now has this setting in Labs, and it works very well.
image

proposals/4027-custom-images-in-reactions.md
anoadragon453: A "must" snuck back in on line 33.

proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,138 @@
...
+## Proposal
+
+This proposal suggests two changes to events with the `m.annotation` relation.
+
+1. If the `key` of an `m.annotation` relation is an MXC URI of an image, clients
+   should render the referenced image instead of the key text.
+
+2. When the annotation's key is an MXC URI, a new (optional) `shortcode` key can
+   be added to the content of the event with a textual name for the image. This
+   field must be a string and should start and end with the `:` (colon)
```
anoadragon453, line 33: @mscbot concern Requiring two `:` characters in every shortcode string.

Could you explain why we should encourage clients to put `:`s in every shortcode instead of restricting `:` from being put in the string?

I know we can't enforce it over federation, but the spec could at least enforce it in the Client-Server API. I think the occasional extra `:` appearing in clients is worth the bandwidth saving.

For bridges, if they connect to a network that includes `:` on the ends of their custom emote shortcodes (or another character), then the bridge can just strip it, no?

anoadragon453 (approved):
Barring the below concern, I think this MSC is ready for review from the rest of the Spec Core Team.

@mscbot fcp merge

proposals/4027-custom-images-in-reactions.md
anoadragon453: Grr, technology.

@mscbot concern Requiring two : characters in every shortcode string.

anoadragon453: @mscbot fcp merge

proposals/4027-custom-images-in-reactions.md
sumnerevans: The type of the field is fairly easily enforceable by servers. We are already saying that they should verify the length. (Obviously if a rouge server doesn't validate these fields, we could have "bad" events come across federation.)

I think the question here is what part of the verification is "should" vs "must"? Right now, the servers "must" verify type and length of the shortcode, but the leading and trailing `:` is "should".

Maybe everything should be "must" and force servers to verify all three things, but leave it up to the server implementation if they want to locally redact invalid events that come across federation?

proposals/4027-custom-images-in-reactions.md
turt2live, line 1: @mscbot concern Needs to be reviewed against other custom emoji/sticker MSCs

proposals/4027-custom-images-in-reactions.md
anoadragon453: I think it would eliminate confusion in implementations for all three to be a MUST on the Client-Server API.

> leave it up to the server implementation if they want to locally redact invalid events that come across federation?

I would say yes, though I imagine most implementations would not redact such events - instead just allowing any UI bugs in the client that would surface.

proposals/4027-custom-images-in-reactions.md
anoadragon453: I think in particular, this MSC sets up grammar for emoji shortcodes in Matrix, which we'll want to double-check aligns with whatever MSC defines sharable image packs.

@turt2live did you have anything else in mind? The rest is fairly reaction-specific IMO.

turt2live: Shortcodes and feature as a whole are conflicting, imo. Specific impact to be determined once the remainder of the stack is looked at :)

(I plan to do this over the next 2ish weeks, barring other commitments - will hand over ~next week if things go poorly on my side)

proposals/4027-custom-images-in-reactions.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Proposal
+
+This proposal suggests two changes to events with the `m.annotation` relation.
+
+1. If the `key` of an `m.annotation` relation is an MXC URI of an image, clients
+   should render the referenced image instead of the key text.
+
+2. When the annotation's key is an MXC URI, a new (optional) `shortcode` key can
+   be added to the content of the event with a textual name for the image. This
```
tim-clifford, line 32: Why not use `key` for the shortcode and add something like `uri` as an optional field for the MXC URI? The current proposal seems like it overloads the meaning of `key` and breaks older clients unnecessarily.

sumnerevans: I addressed this here: https://github.com/matrix-org/matrix-spec-proposals/pull/4027/files#diff-c075783679f0bce493f0a817a6f71200aa434dd8ac13fef5265089a8dc433506R99-R105

Basically, if you put the shortcode in the key, then multiple people could upload different emojis with the same shortcode, and clients would improperly deduplicate them.

turt2live:
As this is tied to the media linking series and unable to progress otherwise, I'm cancelling FCP for the time being. 

This does not mean the MSC is unimportant, but rather the focus is elsewhere at the moment.

@mscbot fcp cancel

proposals/4027-custom-images-in-reactions.md
anoadragon453: It [was questioned](https://matrix.to/#/!0KNSXYXB_2xtEUkQ9MGBRy5oNIOfAKoq2uIqPZCJbI8/$3hhB5eDZU86eoEpYWrRpkPpacUSQLAgJ-lHbX0burSU?via=element.io&via=matrix.org&via=codestorm.net) what remains as a blocker on this MSC. Upon taking another look, I don't think we need to block this MSC on MSC2545. All I would do is define the `shortcode` field as a freeform text field (and remove the colon constraints). Clients will then look to MSC2545 for the shortcode grammar, which will proliferate to this field upon choosing images and shortcodes from an image pack.

sumnerevans: I made a PR against the branch in the Beeper org to remove the restriction on the shortcode here: https://github.com/beeper/matrix-spec-proposals/pull/3

q-TheDog:
Client ([Haven Desktop](https://github.com/Haven-Organization/haven-desktop)): https://github.com/Haven-Organization/haven-desktop/commit/da0cc6130184fbcd53174e7e3f7122d34f09e8c3

Bridge ([matrix-appservice-activitypub](https://github.com/Haven-Organization/matrix-appservice-activitypub)): https://github.com/Haven-Organization/matrix-appservice-activitypub/commit/288aaa8

PR #3013, opened 2021-02-17
MSC3013: Encrypted Push

Sorunome:
[Rendered](https://github.com/Sorunome/matrix-doc/blob/soru/encrypted-push/proposals/3013-encrypted-push.md)

Signed-off-by: Sorunome <mail@sorunome.de>
Signed-off-by: Sorunome <sorunome@famedly.com>

Synapse PR: https://github.com/matrix-org/synapse/pull/11512
FamedlySDK PR: https://gitlab.com/famedly/company/frontend/famedlysdk/-/merge_requests/868
Hedwig PR: https://gitlab.com/famedly/company/backend/services/hedwig/-/merge_requests/41

----

SCT Stuff:

~~[FCP call](https://github.com/matrix-org/matrix-doc/pull/3013#issuecomment-1002262322)~~

[FCP **close** tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3013#issuecomment-4076552387)

PR #4028, opened 2023-06-13
MSC4028: Push all encrypted events except for muted rooms

giomfo:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/giomfo/push_encrypted_events/proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md)

Implementations:
- Synapse: https://github.com/matrix-org/synapse/pull/16361
- Element Clients: EX/SDK are fully compatible with MSC4028. The feature is available on beta.matrix.org but not on matrix.org
    - https://github.com/matrix-org/matrix-rust-sdk/pull/3192
    - https://github.com/matrix-org/matrix-rust-sdk/pull/2840
    - https://github.com/element-hq/element-x-ios/issues/2078
    - https://github.com/element-hq/element-x-android/issues/1794
    - The suggested algorithm to manage the unread notification counts on the client side is ignored. It is optional and depends on the client UI.

### DIFF
--- a/proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
+++ b/proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
@@ -1,15 +1,22 @@
 # MSC4028: Push all encrypted events except for muted rooms
 
-About notifications handling, it is currently not possible for mobile clients which get push notifications to properly implement a mentions-and-keywords-only room when encryption is enabled.
+About notifications handling, it is currently not possible for mobile clients which get push notifications to properly
+implement a mentions-and-keywords-only room when encryption is enabled.
 
-Like it was already described in [MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996): 
+Like it was already described in
+[MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996): 
 
-Currently, every new event in an encrypted room might be pushed to mobile clients due to the default **underride** rule [`.m.rule.encrypted`](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules).
+Currently, every new event in an encrypted room might be pushed to mobile clients due to the default **underride**
+rule [`.m.rule.encrypted`](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules).
 
-However, a room can be configured to be mentions-and-keywords-only by creating a [room-specific push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules)
-with the `id` property set to the room ID & `actions` set to do nothing.[^1] Since this is evaluated *before* the `.m.rule.encrypted` rule (almost) **no events get pushed for a mentions-and-keywords-only room**.
+However, a room can be configured to be mentions-and-keywords-only by creating a
+[room-specific push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules)
+with the `id` property set to the room ID & `actions` set to do nothing.[^1] Since this is evaluated *before* the
+`.m.rule.encrypted` rule (almost) **no events get pushed for a mentions-and-keywords-only room**.
 
-Additionally, a room can be configured to be muted by creating the **override** [push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules) which matches the room ID & has `actions` set to do nothing, e.g.:
+Additionally, a room can be configured to be muted by creating the **override**
+[push rule](https://spec.matrix.org/v1.7/client-server-api/#push-rules) which matches the room ID & has `actions` set
+to do nothing, e.g.:
 
 ```json
 {
@@ -29,7 +36,8 @@ Additionally, a room can be configured to be muted by creating the **override** 
 
 ## Proposal
 
-A new predefined **override** push rule (`.m.rule.encrypted_event`) is added just after the `.m.rule.master` push rule to be able to push all encrypted events:
+A new predefined **override** push rule (`.m.rule.encrypted_event`) is added just after the `.m.rule.master` push rule
+to be able to push all encrypted events:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Listening notifications
+
+This new push rule will impact the existing endpoint [`GET /_matrix/client/v3/notifications`](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientv3notifications).
+
+When this rule will be present and enabled in the account push rules set, the clients which retrieve the list of events that it has been notified about with this endpoint, will receive most of the encrypted events (except for muted rooms). They will be able to decrypt and re-run push rules locally. This should fix the notifications panel observed in some web clients where currently the notifications of encrypted events are missing.
+
+## Potential issues
+
+Until a client is updated to support this MSC, it may display a high number of unread notifications for encrypted rooms. The first workaround would be to not display the count for the encrypted rooms when the rule `.m.rule.encrypted_event` is present and enabled in the account push rules set, until the suggested algorithm is implemented.
```
erikjohnston, line 94: Is there not a big potential issue with battery life that should be mentioned here? E.g. for my account I get an encrypted message roughly every 30s, and waking up my phone for each seems like it could be quite a drain?

giomfo: For my point of view, this may be a potential issue, but not a big one compared to clients which are polling all events in background like F-Droid client. I added a note on this point in my last commit (see [this sub-section](https://github.com/matrix-org/matrix-spec-proposals/blob/giomfo/push_encrypted_events/proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md#battery-life))

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Receiving notifications
+
+When this push rule matches then homeserver would push the event to all registered clients, similar to how other rules work. Clients would [decrypt the event and re-run push rules](https://spec.matrix.org/v1.7/client-server-api/#receiving-notifications) as normal. If the decrypted event results in no notification, the push is discarded [^2]. If the decryption failed, the new `.m.rule.encrypted_event` rule will match on client side, but there is no reason to notify the end user, so the push is discarded [^2] too. Globally the client must discard the push when the event stays encrypted locally because this use case should not happen except if the event was not encrypted for the current session, or if we are in front of an unexpected "Unable To Decrypt" (UTD) issue.
+
+The overall tradeoff is that clients will receive extra pushes some of the time.
+
+### Unread notification counts
```
clokep, line 58: This isn't really different from today for normal rooms, correct? If you wish to receive accurate counts it is complicated.

This should all definitely be documented in the spec though!

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,139 @@
...
+
+About the case 4, the back pagination should be mandatory for mentions-and-keywords-only rooms to detect if there are some mentions in the gap. The client may decide to not trigger back pagination for rooms in "all messages" mode, but in that case the local unread notification count is unknown. No badge can be displayed at the application side, only an unread marker.
+
+Note: some unencrypted events (for example `m.room.redaction`) may be taken into account or not in this unreads count k. This depends on the push rules set. If you force "all messages" mode on a room by creating a room-specific push rule, some unencrypted events (like redaction events) will be included in this count. This is not the case if "all messages" mode is obtained by the combination of server predefined rules. The client have locally all the information (= the push rules) to handle this edge case during the implementation of the algorithm.
+
+### Listening notifications
```
clokep, line 86:
```suggestion
### Listing notifications
```

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Server side
+
+#### Event fetching
+The clients with push services will request more frequently the server to retrieve the event contents. I'm mainly thinking about an encrypted room with a high number of members, each mobile client of these members will try to retrieve simultaneously the event content of each new encrypted event (pushed by the server). This scenario may happen today only when most of the members configure the room in "All messages" mode.
+
+This may be a potential performance issue if each client fetches separately each pushed event. But this should not be the case, because the clients have to run a sync if they want to be able to receive potential new encryption keys through to_devices. So they will retrieve more than one event at the time.
+
+#### Push sending
+Push is a huge bottleneck on the sending side. So calculating what to push and the unread counts (its the same process) is a big performance bottleneck currently when sending in large rooms. A bunch of optimisation has taken place to make that better, but its still quite expensive. We should plan an effort to limit potential issue server side here.
```
clokep, line 110: This won't cause the push rules *evaluation* to be more expensive (if anything it might be less expensive as we'll quickly match a rule and not need to iterate all the rules).

It will cause additional push notifications to be sent, however, which could increase load on the database, pushers, and the push gateways.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,139 @@
...
+         "kind": "event_match",
+         "key": "type",
+         "pattern": "m.room.encrypted"
+      }
+    ],
+    "actions": ["notify"]
+}
+```
+
+Note: The “user-defined rules” are evaluated with a higher priority than “server-default rules”, except for the `.m.rule.master` rule which has always a higher priority than any other rule (see [here](https://spec.matrix.org/v1.7/client-server-api/#predefined-rules)). So all the **override** push rules created to mute rooms will be evaluated before this new one. The new rule will be evaluated before all the other “server-default rules”. This new rule deprecates the existing underride rule [`.m.rule.encrypted`](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules), which will become useless.
```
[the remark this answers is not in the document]
giomfo: I consider my note is correct. According to the spec, it is not possible to add a rule relative to a predefined server rule, and `.m.rule.encrypted_event` is a predefined server rule. 
I would prefer to keep unchanged this note. I let you mark as resolved this point if you agree

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Receiving notifications
+
+When this push rule matches then homeserver would push the event to all registered clients, similar to how other rules work. Clients would [decrypt the event and re-run push rules](https://spec.matrix.org/v1.7/client-server-api/#receiving-notifications) as normal. If the decrypted event results in no notification, the push is discarded [^2]. If the decryption failed, the new `.m.rule.encrypted_event` rule will match on client side, but there is no reason to notify the end user, so the push is discarded [^2] too. Globally the client must discard the push when the event stays encrypted locally because this use case should not happen except if the event was not encrypted for the current session, or if we are in front of an unexpected "Unable To Decrypt" (UTD) issue.
```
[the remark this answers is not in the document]
giomfo: yes when the event stays encrypted locally

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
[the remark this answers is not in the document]
clokep: > I consider my note is correct. According to the spec, it is not possible to add a rule relative to a predefined server rule, and `.m.rule.encrypted_event` is a predefined server rule.

Hmmm, re-reading it it *does* seem to say that. Sorry for the noise. I did a quick test and synapse seems to reject this properly:

```bash
$ curl --header "Authorization: Bearer $TOK" https://matrix-client.matrix.org/_matrix/client/v3/pushrules/global/override/test\?after\=.m.rule.tombstone -X PUT --data '{"actions":["notify",{"set_tweak":"highlight"}],"conditions":[{"kind":"event_property_is","key":"type","value":"us.cloke.test"}]}'
{"errcode":"M_UNKNOWN","error":"Given priority class does not match class of relative rule"}
``` 

TIL there's two separate orderings inside of each priority based on default vs. non-default rules.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: I don't really want to reverse engineer the existing unread notification counts. I just [updated the intro of this section](https://github.com/matrix-org/matrix-spec-proposals/pull/4028/commits/132ec4ad15520f16b208c04ee1d525643ed8ec8f) to highlight that the suggested algorithm is similar the potential existing ones.
My wish is to have at least the same algorithm on each client, because this is not the case currently.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,141 @@
+# MSC4028: Push all encrypted events except for muted rooms
+
+About notifications handling, it is currently not possible for mobile clients which get push notifications to properly implement a mentions-and-keywords-only room.
```
Cadair, line 3: I think this is true?
```suggestion
About notifications handling, it is currently not possible for mobile clients which get push notifications to properly implement a mentions-and-keywords-only room when encryption is enabled.
```

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: I agree with your comment. I would definitely prefer to let someone from the backend describe this potential limitation about the push sending. I just tried here to summarize here a discussion that I had with @erikjohnston

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
clokep: > My wish is to have at least the same algorithm on each client, because this is not the case currently.

Yes, we definitely need this to be consistent!

> I don't really want to reverse engineer the existing unread notification counts.

For reference, have you seen https://gist.github.com/Half-Shot/f9501916363894761a1659250aa25181? I think it is pretty much the same as what you describe, but Element Web folks might know better. Regardless, I suppose it will be on each client to ensure they match it. (I'm fairly certain what you give as an algorithm is reasonable, but didn't try it with examples or try to find edge-cases.)

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,141 @@
...
+{
+    "rule_id": ".m.rule.encrypted_event",
+    "default": true,
+    "enabled": true,
+    "conditions": [
+      {
+         "kind": "event_match",
+         "key": "type",
+         "pattern": "m.room.encrypted"
+      }
```
clokep, line 44: This isn't really a change, but I find it clearer to use the newer [`event_property_is` condition](https://spec.matrix.org/v1.8/client-server-api/#conditions-1).

```suggestion
      {
         "kind": "event_property_is",
         "key": "type",
         "value": "m.room.encrypted"
      }
```

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Unexpected unread counts
+Until a client is updated to support this MSC, it may display a high number of unread notifications for encrypted rooms. The first workaround would be to not display the count for the encrypted rooms when the rule `.m.rule.encrypted_event` is present and enabled in the account push rules set, until the suggested algorithm is implemented, at least for the mentions-and-keywords-only encrypted rooms.
+
+#### Battery life
+The mobile client will wake up more frequently than currently. We will have to control any potential issue with battery life when we will implement this MSC. The impact on the battery life will be definitely lower than the one observed on clients without push service which poll for events in the background. The end user will be able to reduce the impact by muting some encrypted rooms. 
+
+### Server side
+
+#### Event fetching
```
clokep, line 106: [MSC3013](https://github.com/matrix-org/matrix-spec-proposals/pull/3013) could potentially mitigate this.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
DMRobertson, line 1: I have no strong opinions on this, but I did leave some comments on https://github.com/matrix-org/synapse/pull/16361#pullrequestreview-1642770233 which you might be interested in.

> My gut reaction is that pushing all encrypted events to clients sounds excessive and might mean the push gateway gets rate limited. But I have no actual data on that. Maybe this is what a mobile app with E2EE messaging has to do? (Are there any open source services like this (Signal?) we could compare to?)
>
>Does this only do push-gateway push, or does it also mean that we'll send out emails to people with email notifications turned on every time they get an encrypted message?

clokep: > Does this only do push-gateway push, or does it also mean that we'll send out emails to people with email notifications turned on every time they get an encrypted message?

This in particular sounds pretty bad. There's no way with push rules to separate these right now.

giomfo: @clokep what does the server implementation about the email notifications when the event matches only the existing underride rule [.m.rule.encrypted](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules)? 
I would expect that no email is sent.
I would generalise that no email notification is sent when the notification is triggered only because the condition "event type == `m.room.encrypted`" is satisfied. The server is not able to send relevant email notification in that specific case.
wdyt? the client should alert about this limitation when the user turns on the email notification option.
cc @DMRobertson

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: fixed by 2207b525d44d121e9be5e5ee113de19b51caffea

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
clokep: > what does the server implementation about the email notifications when the event matches only the existing underride rule [.m.rule.encrypted](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules)?

There's no special casing, Synapse sends emails whenever a "notification" is the result of push rules. (Note that it sends emails on a delay of ~10 minutes to give people time to see the notification and mark it as read.)

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,141 @@
...
+This new push rule will impact the existing endpoint [`GET /_matrix/client/v3/notifications`](https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientv3notifications).
+
+When this rule will be present and enabled in the account push rules set, the clients which retrieve the list of events that it has been notified about with this endpoint, will receive most of the encrypted events (except for muted rooms). They will be able to decrypt and re-run push rules locally. This should fix the notifications panel observed in some web clients where currently the notifications of encrypted events are missing.
+
+## Potential issues
+
+### Client side
+
+#### Unexpected unread counts
+Until a client is updated to support this MSC, it may display a high number of unread notifications for encrypted rooms. The first workaround would be to not display the count for the encrypted rooms when the rule `.m.rule.encrypted_event` is present and enabled in the account push rules set, until the suggested algorithm is implemented, at least for the mentions-and-keywords-only encrypted rooms.
```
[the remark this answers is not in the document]
giomfo: I don't know for the web, but the mobile clients don't update the unread counts when they evaluate the push rules. Their implementation trust naively the received unread counts for the moment. There are perhaps some exceptions, but when I introduced manually this new push rule in one of my test account for an encrypted room, the badge increased on each message in this room.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
clokep: > There's no special casing, Synapse sends emails whenever a "notification" is the result of push rules.

It is unclear to me if this [behavior is specced](https://spec.matrix.org/v1.8/client-server-api/#post_matrixclientv3pushersset) as part of a "pusher" (emphasis mine):

> `kind`: The kind of pusher to configure. `"http"` makes a pusher that sends HTTP pokes. `"email"` makes a pusher that *emails the user with unread notifications*. null deletes the pusher.

I think "unread notification" implies "events which resulted in an [`action` of `notify`](https://spec.matrix.org/v1.8/client-server-api/#actions) after processing push rules", but it is not 100% clear.

Things that we could do are:

1. Special-case this rule to not cause email notifications. This is likely *moderately difficult** due to needing to propagate *why* an event resulted in a push.
2. Define a new action which is "push but don't notify". This *might* not be too hard, but we would probably need to check a few code paths.
3. ...maybe other ideas, but I don't have any.

giomfo: I enabled email notifications for one of my test account, by enabling the underride rule [.m.rule.encrypted](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules) by default. Then I receive email notifications for a bunch of encrypted messages, I'm just able to know the room names and the sender names.
image

The existing email notifications are not really relevant in case of encrypted rooms. We should just list the encrypted room names in the email notification when the pushed events stayed encrypted server side. The users need at least to be notified by email when there is activity in these encrypted rooms without expecting more details (except unencrypted events). They should mute the room if they don't need this information. 

My suggestion: **The email notification mentions only the room name when all pushed events for this room are encrypted** (This will not be limited to this new push rule). 
BTW thanks to this new push rule the users will be able to receive an email notification for a potential mention in an encrypted room configured in "mention only" mode. Currently no email notification is triggered for this use case, so the user may miss a mention.

giomfo: @clokep FYI I added [a section](https://github.com/matrix-org/matrix-spec-proposals/pull/4028/files#diff-c7638211aa168fea4754c94985b770f1373649cc7ea1795563cc9088a8206454R94) in the MSC to summarise this conversation about Email notifications. I will mark it as resolved. We may open a new conversation directly from this section.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,155 @@
...
+
+The number of this email notifications may increase when this new rule will be present and enabled in the account push rules set. Indeed the email notifications will then include the encrypted rooms for which the notifications are configured in mentions-and-keywords-only mode. These rooms are not supported for the moment in the email notifications, so the users may miss some mentions.
+
+Currently the existing email notifications are not really relevant in case of encrypted rooms. The users may receive emails with a long bunch of encrypted messages (only the room names and the sender names are in plain text). This will happen more frequently when this new push rule will be enabled. **We should just list the encrypted room names in the email notification when the pushed events stayed encrypted server side**. The users need at least to be notified by email when there is activity in these unmuted encrypted rooms without expecting more details (except unencrypted events). They should mute the room if they don't need this information.
```
giomfo, line 100: @clokep What do you think about hiding encrypted events in the email notification body? This would be nice to limit the information to the room name in case of encrypted events (I hesitate to keep the sender names, but I want to keep the change simple). This will reduce strongly the impact of this new rule on the email notifications, and reduce the need to consider more drastic changes. Do you agree with me?

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,155 @@
...
+
+When this rule will be present and enabled in the account push rules set, the clients which retrieve the list of events that it has been notified about with this endpoint, will receive most of the encrypted events (except for muted rooms). They will be able to decrypt and re-run push rules locally. This should fix the notifications panel observed in some web clients where currently the notifications of encrypted events are missing.
+
+### Email notifications
+
+A user may set up a pusher to receive emails with unread notifications (see the spec [here](https://spec.matrix.org/v1.8/client-server-api/#post_matrixclientv3pushersset) with `kind` = "email"). Note that it sends emails on a delay of ~10 minutes to give people time to see the notification and mark it as read.
+
+The number of this email notifications may increase when this new rule will be present and enabled in the account push rules set. Indeed the email notifications will then include the encrypted rooms for which the notifications are configured in mentions-and-keywords-only mode. These rooms are not supported for the moment in the email notifications, so the users may miss some mentions.
```
clokep, line 98: > Note that it sends emails on a delay of ~10 minutes to give people time to see the notification and mark it as read.

This is implementation specific and describes what Synapse does, but... it looks like [Dendrite](https://github.com/matrix-org/dendrite/blob/933ae2db91c4c65dfc7b8a395567cd39f4c366c8/userapi/util/notify.go#L52); [Conduit](https://gitlab.com/famedly/conduit/-/blob/3bfdae795d4d9ec9aeaac7465e7535ac88e47756/src/service/pusher/mod.rs#L278); and [Construct](https://github.com/matrix-construct/construct/blob/0624b69246878da592d3f5c2c3737ad0b5ff6277/modules/m_pusher.cc#L432-L440) don't support email pushers though. So I guess this is probably fine.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
clokep: > This will reduce strongly the impact of this new rule on the email notifications, and reduce the need to consider more drastic changes. Do you agree with me?

I don't think I agree -- there will still be a lot of additional emails sent that just say "Go check room XYZ for new notifications." But the notifications are "just" normal messages, not notifications.

This is today's status quo unless you set a room to mentions & keywords (which I think is your point?), but will now cause a mentions & keywords room to essentially notify on every message.

> The users need at least to be notified by email when there is activity in these unmuted encrypted rooms without expecting more details (except unencrypted events).

I don't think this will meet user's expectations of setting a room to "mentions & keywords" where they really only expect to be notified of important things. I think they'll view this as spam.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,155 @@
...
+
+Currently the existing email notifications are not really relevant in case of encrypted rooms. The users may receive emails with a long bunch of encrypted messages (only the room names and the sender names are in plain text). This will happen more frequently when this new push rule will be enabled. **We should just list the encrypted room names in the email notification when the pushed events stayed encrypted server side**. The users need at least to be notified by email when there is activity in these unmuted encrypted rooms without expecting more details (except unencrypted events). They should mute the room if they don't need this information.
+
+If the email notifications become too noisy because of this new rule, we would have to work on them in another MSC by considering eventually one of the following options:
+1. Ignore in the email notifications the encrypted events which have been pushed only because the condition "event type == m.room.encrypted" was satisfied. The server is not able to send relevant email notification in that specific case.
```
clokep, line 103: I think it would probably make more sense to ignore them based on rule IDs, I suspect what to do with encrypted messages in email notifications has never really had a product review.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,155 @@
...
+2. Define a new action which is "push but don't notify", to exclude some pushed events in the email.
```
clokep, line 104: I wonder if we could use a new tweak for this `dont_email` or something? That way it would still increase the notification count, as we expect for backwards compatibility, but could selectively ignore emailing it?

(Alternately `http_only` or something if you don't want a negative.)

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: @clokep thanks for your feedbacks. I updated the Email notifications section [here](5f04a997c4e8720e9d88d8bf07759fb8c6ec56a6). In order to unblock this MSC, I would like to limit its impact on email notifications. Do you think possible to complete the implementation of this MSC in Synapse by ignoring in the email notifications the events pushed because of the new push rule?

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,154 @@
...
+
+Currently the existing email notifications are not really relevant in case of encrypted rooms. The users may receive emails with a long bunch of encrypted messages when the room is configured in "all messages" notification mode. This feature should be reviewed again with a Product point of view.
+
+In order to not disturb the existing email notifications mechanism, the server implementations which support it should ignore in the email notifications the events pushed because of the new push rule.
+Otherwise the number of emails will increase by including the encrypted rooms for which the notifications are configured in mentions-and-keywords-only mode. These rooms are not handled for the moment in the email notifications.
+
+As a first option, the server should ignore these pushed events by using the push rule Id: `.m.rule.encrypted_event` (`org.matrix.msc4028.encrypted_event`). We may consider to work later on a new MSC to introduce a new [action](https://spec.matrix.org/latest/client-server-api/#actions) _"push but don't notify"_ or a new tweak _"dont_email"_ to exclude some pushed events from the email.
```
giomfo, line 103: In order to unblock this MSC, I would like to limit its impact on email notifications.
@clokep Do you think possible to complete the implementation of this MSC in Synapse by ignoring in the email notifications the events pushed because of the new push rule (by following the first option)?

clokep: > Do you think possible to complete the implementation of this MSC in Synapse by ignoring in the email notifications the events pushed because of the new push rule (by following the first option)?

This will require significant rework to the way Synapse handles push notifications (in particular the database will need to be modified -- this is a massive table so modifying the database is non-trival). Adding either a new action or a new tweak wouldn't require any database modifications.

giomfo: I followed your advices. FYI I wrote [MSC4062](https://github.com/matrix-org/matrix-spec-proposals/pull/4062) and [update](b55877727c640a5a67e13e471eb5b4c3691ec49f) this MSC

clokep: Seems reasonable to me! 🎉

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,250 @@
...
+None
+
+## Unstable prefix
+
+During development the new push rule shall use `org.matrix.msc4028.encrypted_event` instead of `.m.rule.encrypted_event`.
+
+Caution: this unstable push rule will be enabled by default like the stable push rule, the server owners should wait for
+the clients to support a minimum this MSC before enabling the MSC server side.
+
+To ensure the server supports the functionality before a spec release, clients should look for the unstable feature `org.matrix.msc4028` in the response of the `/_matrix/clients/versions` endpoint. Once released in the specification, clients should be checking for server support through advertised spec versions instead.
```
anoadragon453, line 236: Just to clarify that a value of `true` is needed, rather than clients simply taking the presence of `org.matrix.msc4028` as whether this MSC's functionality is enabled.

```suggestion
To ensure the server supports the functionality before a spec release, clients should interpret an unstable feature `org.matrix.msc4028` with a value of `true` in the response of the `/_matrix/clients/versions` endpoint as the feature being supported by the homeserver. Once released in the specification, clients should be checking for server support through advertised spec versions instead.
```

PolynomialDivision:
Hi, why has there been no progress on this PR for over a year? Is there a discussion happening somewhere else? People are implementing workarounds because mentions in encrypted group chats are quite important. I guess matrix.org won't enable the existing experimental feature until this MSC is merged?

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,250 @@
...
+None
+
+## Unstable prefix
+
+During development the new push rule shall use `org.matrix.msc4028.encrypted_event` instead of `.m.rule.encrypted_event`.
+
+Caution: this unstable push rule will be enabled by default like the stable push rule, the server owners should wait for
+the clients to support a minimum this MSC before enabling the MSC server side.
+
+To ensure the server supports the functionality before a spec release, clients should interpret an unstable feature `org.matrix.msc4028` with a value of `true` in the response of the `/_matrix/clients/versions` endpoint as the feature being supported by the homeserver. Once released in the specification, clients should be checking for server support through advertised spec versions instead.
```
PolynomialDivision:
```suggestion
To ensure the server supports the functionality before a spec release, clients should interpret an unstable feature `org.matrix.msc4028` with a value of `true` in the response of the `/_matrix/client/versions` endpoint as the feature being supported by the homeserver. Once released in the specification, clients should be checking for server support through advertised spec versions instead.
```

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,250 @@
...
+
+Another alternative would be to define a new [`action`](https://spec.matrix.org/v1.7/client-server-api/#actions):
+`push_without_notify` or `silently_notify` in order to push all encrypted events without incrementing
+the notifications count. The client implementation to handle these counts would be then less complex.
+But this would need to implement more changes in Synapse. The current proposal is just to add a new push rules
+without changing the rules handling server side. This alternative should be discussed as soon as possible.
+
+## Security considerations
+
+None
```
PolynomialDivision:
```suggestion
This MSC does not reveal encrypted event contents, but it does change the amount and shape of push
traffic for encrypted rooms. Push gateways and application services involved in push delivery may
observe more wakeups for encrypted rooms than before, which can expose additional metadata about
encrypted-room activity.
```

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,250 @@
...
+“server-default rules”. When it is enabled, this new rule makes useless the 2 existing underride rules:
+`.m.rule.encrypted_room_one_to_one` and `.m.rule.encrypted` (predefined
+[here](https://spec.matrix.org/v1.7/client-server-api/#default-underride-rules)).
+
+### Receiving notifications
+
+When this push rule matches then homeserver would push the event to all registered clients, similar to how other rules
+work. Clients would
+[decrypt the event and re-run push rules](https://spec.matrix.org/v1.7/client-server-api/#receiving-notifications) as
+normal. If the decrypted event results in no notification, the push is discarded [^2]. If the decryption failed, the
```
PolynomialDivision:
```suggestion
When this push rule matches, the homeserver would push the event to all registered clients, similar to
how other rules work. This rule is intended to wake clients so they can
[decrypt the event and re-run push rules](https://spec.matrix.org/v1.7/client-server-api/#receiving-notifications)
locally. It is not by itself a reason to display a user-visible notification. If the decrypted event results in no notification, the push is discarded [^2]. If the decryption failed, the
```

PolynomialDivision:
BTW.: Fluffychat is also working now with MSC4028: https://github.com/krille-chan/fluffychat/commit/eca88e150268e9e6564dbe10194eee7cf51b0d0d

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: thanks for having fixed this typo

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Client side
+
+#### Unexpected unread counts
+Until a client is updated to support this MSC, it may display a high number of unread notifications for encrypted rooms.
+The first workaround would be to not display the count for the encrypted rooms when the rule `.m.rule.encrypted_event`
+is present and enabled in the account push rules set, until the suggested algorithm is implemented, at least for the
+mentions-and-keywords-only encrypted rooms.
+
+#### Battery life
```
clokep, line 182: Do we have any idea the impact here? It sounds like it would really impact battery life if you're in a few busy encrypted rooms.

PolynomialDivision: It probably will have some impact in busy encrypted rooms. But the alternative is that mentions in "mentions only" rooms simply don't work. Anyone who wants to avoid the extra processing can mute those rooms.

clokep: > It probably will have some impact in busy encrypted rooms

It _will_ have impact. I'm looking for some data to quantify whether this solution is the correct tradeoff or not.

> Anyone who wants to avoid the extra processing can mute those rooms.

Users won't know that their battery is draining fast and that they should mute certain rooms. That's not really a good workaround.

PolynomialDivision: Yes, but the current status is that when users set rooms to "mentions only," they are never notified. I see this as the more problematic issue. For me, it would be acceptable to simply add the room mentions to the outside of the encrypted message, allowing this metadata leakage, like I did here as a workaround for my bots: https://github.com/PolynomialDivision/matrix-rust-sdk/commit/fa305ba2c5d1f7b182d561b396d410bed83eff67
Alternatively, we could add a generic "wakeup" flag to the message (or something similar), so you regain some privacy and reduce leakage.

clokep: > For me, it would be acceptable to simply add the room mentions to the outside of the encrypted message, allowing this metadata leakage

This was the original design for `m.mentions`, which was deemed not OK. See MSC3952. [MSC3996](https://github.com/matrix-org/matrix-spec-proposals/pull/3996) was my attempted workaround that would be acceptable to everyone.

PolynomialDivision: If MSC3996 is the acceptable compromise, what is currently blocking it from moving forward?

clokep: It isn't moving forward because it probably isn't considered an acceptable compromise.

turt2live: It feels important that we measure the rate of change of battery usage for this proposal. Every device will respond differently to the actual battery usage, but we should be able to say "it uses X% more 95% of the time" to help guide a decision here (assuming reasonable sample size).

We should also measure how reliable the push delivery is, especially on large accounts. It's entirely possible that pushing *too much* data to the device actually slows down overall delivery to avoid overwhelming the user.

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,253 @@
...
+
+As a first option, we suggested to ignore these pushed events by using the push rule Id: `org.matrix.msc4028.encrypted_event`.
+But this would require significant rework to the way Synapse handles push notifications (in particular the database
+will need to be modified -- this is a massive table so modifying the database is non-trivial).
+To avoid database modifications, we wrote [MSC4062](https://github.com/matrix-org/matrix-spec-proposals/pull/4062).
+The new tweak `org.matrix.msc4062.dont_email` would let us disable the email notification in the push rule definition.
+
+## Potential issues
+
+### Client side
```
turt2live, line 174: I've been told in the past that Apple devices in particular *must* show a notification when they receive a push, and cannot cancel it. Has that been fixed/addressed? How?

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
```diff
@@ -0,0 +1,253 @@
...
+    "default": false,
+    "enabled": true,
+    "actions": []
+}
+```
+
+## Proposal
+
+A new predefined **override** push rule (`.m.rule.encrypted_event`) is added just after the `.m.rule.master` push rule
+to be able to push all encrypted events:
```
turt2live, line 40: Servers and web clients also use push rules to calculate badge counts, especially over gappy syncs. In the alternatives there's mention of additional `actions`, but can we instead just forward encrypted events to the push gateway without a dedicated rule or action?

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: This is addressed by using notification filtering entitlement (https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.usernotifications.filtering)

proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md
giomfo: I failed to understand the question

PR #4029, opened 2023-06-14
MSC4029: [WIP] Fixing `X-Matrix` request authentication

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/fix-x-matrix-auth/proposals/4029-current-practice-for-s2s-auth.md)

Issues relating to this problem space are:

* https://github.com/matrix-org/matrix-spec/issues/849
* https://github.com/matrix-org/matrix-spec/issues/561
* https://github.com/matrix-org/matrix-spec/issues/512
* https://github.com/matrix-org/matrix-spec/issues/1570
* https://github.com/matrix-org/matrix-spec/issues/1471
* https://github.com/matrix-org/matrix-spec/issues/510 (kinda)
* https://github.com/matrix-org/matrix-spec/issues/479

Related issues, while we're in the area:
* https://github.com/matrix-org/matrix-spec/issues/1569

### DIFF
--- a/proposals/4029-current-practice-for-s2s-auth.md
+++ b/proposals/4029-current-practice-for-s2s-auth.md
@@ -1,0 +1,57 @@
+# MSC4029: Fixing `X-Matrix` request authentication
+
+The spec inadequately describes how the [`X-Matrix` authentication scheme](https://spec.matrix.org/v1.7/server-server-api/#authentication)
+works in the real world. Details like how the query string for a request is handled, encoding
+considerations, error/failure conditions, how to handle multiple headers, when to *expect* multiple
+headers, etc are all lacking in the specification.
+
+This proposal covers those details, using what is currently implemented in servers as a starting
+principle. Normally, these sorts of changes would be considered bug fixes against the spec itself,
+however with an MSC it allows an opportunity to question the motivation for doing something. If
+this MSC ultimately diverges from what is done in practice, backwards compatibility becomes a major
+requirement.
+
+Issues relating to this problem space are:
+
+* https://github.com/matrix-org/matrix-spec/issues/849
+* https://github.com/matrix-org/matrix-spec/issues/561
+* https://github.com/matrix-org/matrix-spec/issues/512
+* https://github.com/matrix-org/matrix-spec/issues/1570
+* https://github.com/matrix-org/matrix-spec/issues/1471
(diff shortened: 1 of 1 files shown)


proposals/4029-current-practice-for-s2s-auth.md
```diff
@@ -0,0 +1,137 @@
+# MSC4029: Fixing `X-Matrix` request authentication
```
turt2live, line 1: Consider https://github.com/matrix-org/matrix-spec/issues/1817

Kladki, line 1: Probably not something that will be encountered with any sensible server, but it is possible to have [multiple Authorization headers with the same scheme if different realms are used](https://datatracker.ietf.org/doc/html/rfc7235#section-2.2):
> Note that a response can have multiple challenges with the same auth-scheme but with different realms.

So perhaps there could be a note stating that servers should only use the first X-Matrix header they encounter?

Kladki: Nvm, I see [this was mentioned](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/fix-x-matrix-auth/proposals/4029-current-practice-for-s2s-auth.md#required-keys).

Kladki: Hmm no, the RFC says this is the case for a **response**, so how exactly would multiple `X-Matrix` keys be available in a request while still following the RFC?

proposals/4029-current-practice-for-s2s-auth.md
```diff
@@ -0,0 +1,137 @@
...
+### Key retrieval
+
+*Issues: [#479](https://github.com/matrix-org/matrix-spec/issues/479)*
+
+The keys a server uses to sign its request are the same it uses to sign events. This also means that
+a server *verifying* a signature can use the normal [`GET /_matrix/key/v2/server`](https://spec.matrix.org/v1.9/server-server-api/#get_matrixkeyv2server)
+endpoint to fetch that server's keys.
+
+Servers SHOULD NOT [query keys through another server](https://spec.matrix.org/v1.9/server-server-api/#querying-keys-through-another-server)
+when validating a request signature. Instead, the server's local cache and direct requests to the
```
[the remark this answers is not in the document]
turt2live: However that cache was built, realistically.

PR #2654, opened 2020-06-24
MSC2654: Unread counts

babolivier:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/babolivier/unread_counts/proposals/2654-unread-counts.md)

Synapse implementation: https://github.com/matrix-org/synapse/pull/8059
Client: https://github.com/quotient-im/libQuotient/pull/521

PR #4033, opened 2023-07-04
MSC4033: Explicit ordering of events for receipts

### LINKED ISSUE #1167: Clarify if notifications are cleared by stream ordering or topological ordering
**Link to problem area**: https://spec.matrix.org/latest/client-server-api/#receiving-notifications

**Issue**:

It is not stated clearly which notifications get cleared when a read receipt is received, currently the spec reads:

> When the user updates their read receipt (either by using the API or by sending an event), notifications prior to and including that event MUST be marked as read.

It is unclear if "prior to" refers to stream ordering or topological ordering. Synapse clears this based on stream ordering.

(Note that because of the inconsistency in `/sync` and `/messages` either could give odd results to a user.)

See conversation around https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$sb6NPRyuiu4wN7egyY8SJ6YoubvQqZvJSOlMFsTzVSE?via=matrix.org&via=libera.chat&via=element.io

andybalaam:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/andybalaam/event-thread-and-order/proposals/4033-event-thread-and-order.md)

Points out ambiguities in the current spec about what a receipt means, because it is not clear on the order of events that means they are "before" or "after" a receipt. Proposes that events and receipts contain an `order` property that makes it explicit which receipt marks which event as read.

Fixes https://github.com/matrix-org/matrix-spec/issues/1167

### DIFF
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -57,7 +57,7 @@ event is read.
 ### Current definition of *after*
 
 The current spec (see
-[11.6](https://spec.matrix.org/latest/client-server-api/#receipts) is not clear
+[11.6 Receipts](https://spec.matrix.org/latest/client-server-api/#receipts)) is not clear
 about what it calls "read up to" means.
 
 Clients like Element Web make the assumption that *after* means "after in Sync
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -173,7 +173,7 @@ We propose that the definition of *in the same thread* should use this wording:
 
 An event is in a non-main-thread if:
 
-* it has a `thread_id` property in its content
+* it has a `thread_id` property in its cleartext content
 
 Otherwise, it is in the `main` thread.
 
@@ -219,6 +219,21 @@ Example event (changes are highlighted in bold):
   }
 }</pre>
 
+Example encrypted event (changes are highlighted in bold):
+
+<pre>{
+    "type": "m.room.encrypted",
+    "content": {
+        <b>"thread_id": "$fgsUZKDJdhlrceRyhj", // Optional</b>
+        "algorithm": "m.megolm.v1.aes-sha2",
+        "sender_key": "<sender_curve25519_key>",
+        "device_id": "<sender_device_id>",
+        "session_id": "<outbound_group_session_id>",
+        "ciphertext": "<encrypted_payload_base_64>"
+    }
+    // irrelevant fields not shown
+}</pre>
+
 Example receipt (changes are highlighted in bold):
 
 <pre>{
@@ -251,10 +266,10 @@ We propose:
 [^3]: This might make the `ts` property within receipts redundant. We are not
   actually sure what purpose this property is intended to serve.
 
-The creator of an event should include a `thread_id` property in the content of
-any event that has an ancestor relationship that includes an `m.thread`
-relationship. The value of `thread_id` is the event ID of the event referenced
-by the `m.thread` relationship.
+The creator of an event should include a `thread_id` property in the cleartext
+content of any event that has an ancestor relationship that includes an
+`m.thread` relationship. The value of `thread_id` is the event ID of the event
+referenced by the `m.thread` relationship.
 
 This makes it explicit how events are categorised, meaning clients and servers
 can always agree on which thread an event is in, so the meaning of a threaded
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -69,7 +69,8 @@ easily compare whether an event is before or after a receipt.
 
 This order should be a number that is attached to an event by the server before
 it sends it to any client, and it should never change. It should,
-loosely-speaking, increase for "newer" messages within the same room. It may be negative.
+loosely-speaking, increase for "newer" messages within the same room. It may be
+negative.
 
 The exact meaning of this order is not as important as the fact that it exists
 and is consistent between clients and servers, and events and receipts.
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -9,15 +9,14 @@
 - [x] Make it robust to redactions by always considering redacted messages read
 - [x] Remove thread roots from the thread
 - [x] ~~Move ordering into unsigned (maybe thread id too?)~~ No - unsigned is frowned upon
-- [ ] Events can never have the same stream order
+- [x] Events can never have the same stream order
 - [x] Reword to some generic order instead of stream order?
 - [x] Can't do m.is_thread_root because we don't know until a child exists
-- [ ] Example of inconsistent Sync Order:
-- [ ]   I guess that happens if one is doing an incremental sync and one is doing an initial sync and the event in question is the latest event by stream order but a much earlier event by topo order?
-- [ ] Explicitly say we should update the spec wording around what we mean by read-up-to
+- [x] Example of inconsistent Sync Order
+- [x] Explicitly say we should update the spec wording around what we mean by read-up-to
 - [x] Consider the thread root not being in the thread. Would need to think about whether it matters if the thread root is somehow later than a thread message in Stream Order.
-- [ ] Be really clear on the JSON changes, and the semantics changes
-- [ ] Suggest that `ts` in receipts might be redundant?
+- [ ] JSON examples
+- [x] Suggest that `ts` in receipts might be redundant?
 
 We argue that we have made it unnecessarily hard for clients and servers to
 decide whether a message is read or unread, and we can solve this problem by
@@ -67,13 +66,17 @@ Clients like Element Web make the assumption that *after* means "after in Sync
 Order", where "Sync Order" means "the order in which I (the client) received the
 events from the server via sync", so if a client received an event and another
 event for which it has a receipt via sync, then the event that was later in the
-sync or received in a later sync, is after the other one. We think this is
-similar to Stream Ordering, which is mentioned once in the spec without
-definition [7.6
-Syncing](https://spec.matrix.org/unstable/client-server-api/#syncing), but we
-are not certain that it is identical, because we believe it may be possible for
-different clients to receive events in a different order from each other for the
-same account.
+sync or received in a later sync, is after the other one [^1].
+
+[^1]: We think this is similar to Stream Ordering, which is mentioned once in
+  the spec without definition
+  [7.6 Syncing](https://spec.matrix.org/unstable/client-server-api/#syncing),
+  but we are not certain that it is identical, because we believe it may be
+  possible for different clients to receive events in a different order from
+  each other for the same account. For example, if one client is doing an
+  incremental sync, and another is doing an initial sync, recently-arrived
+  events that are "old "in Topological Order may be received in different orders
+  on the two clients.
 
 See also [Spec Issue #1167](https://github.com/matrix-org/matrix-spec/issues/1167).
 
@@ -135,7 +138,13 @@ We propose that the definition of *after* should be:
 We define *Stream Order* to be an immutable, unique number attached to an event
 on creation that defines the order of events in regard to receipts.
 
-Note: some home servers already have a concept of Stream Order, and we intend
+We propose updating the spec around receipts
+([11.6 Receipts](https://spec.matrix.org/latest/client-server-api/#receipts))
+to be explicit about what "read up to" means, using the above definition.
+
+#### Notes
+
+Some home servers already have a concept of Stream Order, and we intend
 that this definition is consistent with that as currently implemented. But, for
 the purposes of this proposal the only important aspect of Stream Order is that
 server and clients agree that receipts apply to events with a lower Stream
@@ -163,16 +172,17 @@ generally increase for "newer" messages. Clients may decide to re-order events
 into Stream Order, or they may decide to display unread messages higher up the
 timeline if the orders do not match the order they choose for display.
 
-Because Stream Order may be inconsistent across federation (and, in theory, across
-different users on the same home server, although we expect in practice this
-will not happen), one user may occasionally see a different unread status for
-another user from what that user themselves see. We regard this as impossible to
-avoid, and expect that in most cases it will be unnoticeable, since home servers
-with good connectivity will normally have similar Stream Order. When servers have
-long network splits, there will be a noticeable difference at first, but once
-messages start flowing normally and users start reading them, the differences
-will disappear as new events will have higher Stream order than the older ones
-on both servers.
+Because Stream Order may be inconsistent across federation[^2], one user may
+occasionally see a different unread status for another user from what that user
+themselves see. We regard this as impossible to avoid, and expect that in most
+cases it will be unnoticeable, since home servers with good connectivity will
+normally have similar Stream Order. When servers have long network splits, there
+will be a noticeable difference at first, but once messages start flowing
+normally and users start reading them, the differences will disappear as new
+events will have higher Stream order than the older ones on both servers.
+
+[^2]: In theory, Stream Order could also be inconsistent across different users
+  on the same home server, although we expect in practice this will not happen.
 
 ### Proposed definition of *in the same thread*
 
@@ -212,7 +222,10 @@ We propose:
 * all events should contain an `m.order` property.
 * all receipts should contain an `m.order` property alongside `m.read`
   and/or `m.read.private` inside the information about an event, which is a
-  cache of the `m.order` property within the referred-to event.
+  cache of the `m.order` property within the referred-to event [^3].
+
+[^3]: This might make the `ts` property within receipts redundant. We are not
+  actually sure what purpose this property is intended to serve.
 
 The server should include an `m.thread_id` property in any event that has an
 ancestor relationship that includes an `m.thread` relationship. The value of
@@ -310,7 +323,7 @@ ideas from @t3chguy, @justjanne, @germain-gg and @weeman1337.
 
 * 2023-07-04 Initial draft by @andybalaam after conversation with @ara4n.
 * 2023-07-05 Remove thread roots from their thread after conversation with @clokep.
-* 2023-07-05 Handle redactions after conversation with @t3chguy
-* 2023-07-05 Give a definition of Stream Order, caveat around its name
+* 2023-07-05 Make redactions never unread after conversation with @t3chguy
+* 2023-07-05 Give a definition of Stream Order
 * 2023-07-05 Be explicit about Stream Order not going over federation
 * 2023-07-05 Mention disagreeing about what another user has read
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -203,6 +203,32 @@ clarification:
 Further, it should be stated that events with negative order are always read,
 even if no receipt exists.
 
+### Order does not have to be unique
+
+If this proposal required the `order` property to be unique within a room, it
+might inadvertantly put constraints on the implementation of servers since some
+linearised process would need to be involved.
+
+So, we do not require that `order` should be unique within a room. Instead, if
+two events have the same `order`, they are both marked as read by a receipt with
+that order.
+
+Events with identical order introduce some imprecision into the process of
+marking events as read, so they should be minimised where possible, but some
+overlap is tolerable where the server implementation requires it.
+
+So, a server might choose to use the epoch millisecond at which it received a
+message as its order. However, if a server receives a large batch of messages in
+the same millisecond, this might cause undesirable behaviour, so a refinement
+might be the millisecond as the integer part and a fractional part that
+increases as the batch is processed, preserving the order in which the server
+receives the messages in the batch.
+
+If a server were processing multiple batches in parallel, it could implement
+this in each process separately, and accept that some events would receive
+identical orders, but this would be rare in practice and have little effect on
+end users' experience of unread markers.
+
 ### Redacted events
 
 Existing servers already include an `unsigned` section with redacted events,
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -9,18 +9,15 @@ which events are read.
 
 ## Motivation
 
-In order to decide whether a room is unread, a Matrix client must decide whether
-it contains any unread messages.
+To decide whether a room is unread, a Matrix client must decide whether it
+contains any unread messages.
 
-Similarly, in order to decide whether a room has notifications, a Matrix client
-or server must decide whether any of its potentially-notifying messages is
-unread.
+Similarly, to decide whether a room has notifications, we must decide whether
+any of its potentially-notifying messages is unread.
 
 Both of these tasks require us to decide whether a message is read or unread.
 
-To make this decision we have receipts.
-
-To decide whether an event is read, we use the following rule:
+To make this decision we have receipts. We use the following rule:
 
 > An event is read if the room contains an unthreaded receipt pointing at an
 > event which is *after* the event, or a threaded receipt pointing at an event
@@ -31,8 +28,6 @@ To decide whether an event is read, we use the following rule:
 
 (In both cases we only consider receipts sent by the current user, obviously. We
 consider either private or public read receipts.)
-
-We do not propose to change the above definition.
 
 To perform this calculation we need a clear definition of *after*.
 
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -249,3 +249,8 @@ None at this time.
 
 Formed from a discussion with @ara4n, with early review from @clokep. Built on
 ideas from @t3chguy, @justjanne, @germain-gg and @weeman1337.
+
+## Changelog
+
+* 2023-07-04 Initial draft by @andybalaam after conversation with @ara4n.
+* 2023-07-05 Remove thread roots from their thread after conversation with @clokep.
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -77,12 +77,10 @@ An event is in thread A if:
   in the thread), or
 * it has an ancestor event (found by traversing relationships) that fits either of the above.
 
-An event is in the main thread if:
-
-* its ancestor events do not include an `m.thread` relationship
-
-Note that thread root events, and their non-threaded children (e.g. reactions to
-thread root events) are in BOTH the main thread and another thread.
+The spec states that room events are in `main` if they are not in another thread, but
+clients such as Element Web treat thread roots, and non-thread descendants of thread roots
+(such as reactions to the thread root) as being in BOTH `main` and the thread branching
+from the thread root.
 
 ### Problems with the current definitions
 
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -52,7 +52,7 @@ event for which it has a receipt via sync, then the event that was later in the
 sync or received in a later sync, is after the other one [^1].
 
 [^1]: We think this is similar to Stream Ordering, which is mentioned once in
-  the spec without definition
+  the spec without definition in
   [7.6 Syncing](https://spec.matrix.org/unstable/client-server-api/#syncing),
   but we are not certain that it is identical, because we believe it may be
   possible for different clients to receive events in a different order from
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -64,8 +64,8 @@ information: the referenced event must be fetched and correctly ordered.
 
 ## Proposal
 
-We propose to add an explicit order number to events and receipts, so we always
-know what is before or after what.
+We propose to add an explicit order number to events and receipts, so we can easily
+compare whether an event is before or after a receipt.
 
 This order should be a number that is attached to an event by the server before
 it sends it to any client, and it should never change. It should,
--- a/proposals/4033-event-thread-and-order.md
+++ b/proposals/4033-event-thread-and-order.md
@@ -182,10 +182,12 @@ And the above definition of *after* should be included.
 
 ### Redacted events
 
-Since we propose that receipts should cache the `order` of their referred-to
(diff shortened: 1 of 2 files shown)


proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,375 @@
...
+is read.
+
+### Proposed definition of *after*
+
+We propose that the definition of *after* should be:
+
+* Event A is after event B if its *Stream Order* is larger.
+
+We define *Stream Order* to be an immutable, unique number attached to an event
+on creation that defines the order of events in regard to receipts.
```
Johennes, line 122: Where exactly does this happen between clients and home servers? Is this "on creation" on the receiver's home server?

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,375 @@
...
+
+Because it controls the meaning of read receipts, it is desirable that Stream
+Order be as close as possible to Sync Order, the order in which clients receive
+events via sync. However, since clients receive events in different orders
+depending on the APIs they use, it is not a goal that Stream Order exactly match
+the order in which clients receive events. Instead, it provides a canonical
+order that means we can be clear about what the user has read, and thus should
+generally increase for "newer" messages. Clients may decide to re-order events
+into Stream Order, or they may decide to display unread messages higher up the
+timeline if the orders do not match the order they choose for display.
```
Johennes, line 156: Would clients not always want to display messages in the same order that their unread status is maintained in? Otherwise, their read receipts could mark messages that haven't actually been read as read. I'm also struggling slightly to imagine a UI treatment where read and unread messages are mingled.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,375 @@
...
+thread branching from them. Non-thread children of thread roots (e.g.
+reactions to a thread root) are also in the `main` thread. This is a change to
+the current definition.
+
+### Proposed change in consideration of redacted events
+
+We propose that redacted events should never be considered unread.
+
+This avoids the need to identify which thread a redacted event belongs to, which
+will be difficult if its `thread_id` property has been stripped out.
```
Johennes, line 192: IIUC we could also mandate `thread_id` to not be stripped during redaction but this would require a new room version. If that's the case, it might be worth calling this out explicitly here.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,390 @@
...
+
+We propose to modify the definitions slightly, and include extra information in
+the events provided by the server to make it easy to calculate whether an event
+is read.
+
+### Proposed definition of *after*
+
+We propose that the definition of *after* should be:
+
+* Event A is after event B if its *Stream Order* is larger.
```
turt2live, line 119: Note that at least in Synapse it's possible to get negative stream ordering values if the server backfilled the event. Stream ordering itself is not send to clients, so they might not be aware of the "source" for an event.

richvdh: (synapse uses -ve stream orderings for such events mostly to ensure they are never sent down `/sync`, fwiw)

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,390 @@
...
+
+[^1]: We think this is similar to Stream Ordering, which is mentioned once in
+  the spec without definition in
+  [7.6 Syncing](https://spec.matrix.org/unstable/client-server-api/#syncing),
+  but we are not certain that it is identical, because we believe it may be
+  possible for different clients to receive events in a different order from
+  each other for the same account. For example, if one client is doing an
+  incremental sync, and another is doing an initial sync, recently-arrived
+  events that are "old" in Topological Order may be received in different orders
+  on the two clients.
```
richvdh, line 62: I think "Sync ordering" is exactly the same as "Stream ordering" -- at least for Synapse. Synapse will always send events down `/sync` response in stream ordering. Different clients should never see events in `/sync` responses in different orders.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,390 @@
...
+
+The current definition of *after* is ambiguous, and difficult for clients to
+calculate. It depends on only receiving events via sync, which is impractical
+due to backpagination, and the desire to fetch thread messages via the
+`relations` API.
+
+Further, we believe that the current working definition of "Sync Order" does not
+make sense, because different clients may receive events in different orders,
+meaning that a receipt from one client would be interpreted differently by
+another client.
```
richvdh, line 97: as above: I don't *think* this is possible.

proposals/4033-event-thread-and-order.md
andybalaam: How clients choose to display late-arriving old messages (e.g. after a network split) is considered out of scope of this MSC. Some clients might want to show them at the bottom of the timeline to make sure the user can see them, and some clients might want to mix them in higher up the timeline, perhaps with some kind of bold marking to show that they are unread.

The point in this MSC is that the client can't do that unless it can determine whether a specific message is unread. This means that the order needs to be explicit, and actually that it can't be topological order, since that would make newly-arriving old messages look like they had been read when they have not.

proposals/4033-event-thread-and-order.md
andybalaam: Yes correct. I will add something.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,375 @@
...
+* it has a `thread_id` property
+
+Otherwise, it is in the `main` thread.
+
+No events are in more than one thread.
+
+Note: this means that thread roots are in the `main` thread, and not in the
+thread branching from them. Non-thread children of thread roots (e.g.
+reactions to a thread root) are also in the `main` thread. This is a change to
+the current definition.
```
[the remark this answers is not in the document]
andybalaam: Yes, will add a sentence.

proposals/4033-event-thread-and-order.md
andybalaam: I will add a note that stream order can be negative. I am arguing that stream ordering should be exposed to clients explicitly.

proposals/4033-event-thread-and-order.md
andybalaam: OK. I am not making that assumption here for all home servers.

proposals/4033-event-thread-and-order.md
andybalaam: OK. I don't know whether this is important enough to investigate or whether to simply remove this comment from the MSC. I don't this it is critical to justifying why we need explicit order.

bradtgmurray:
Funnily enough we at Beeper have been using this exact approach (`com.beeper.hs.order` field in the `unsigned` portion) in our Synapse fork and I can say it's been useful for simplifying client development: https://github.com/beeper/synapse/pull/42

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+We propose that the definition of whether an event is read should be unchanged,
+but this note should be added:
+
+> (Because the receipt itself contains the `order` of the pointed-to event,
+> there is no need to examine the pointed-to event, so it is sufficient to
+> compare the `order` of the event in question with the `order` in the
+> receipt.)
+
+And the above definition of *after* should be included.
+
```
clokep, line 179: This doesn't clarify the issues in #1167 FTR -- the homeserver still needs to decide on this ordering (and we should define it in the spec). It is fine if this MSC doesn't attempt to do this though.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+### Redacted events
+
+Since we propose that receipts should cache the `order` of their referred-to
+event, this means we do not need to look within a redacted event for its order,
+because the receipt provides it. This avoids needing to preserve the `order`
+property when redacting events.
```
clokep, line 185: I think this is saying that the `order` field should be returned even for redacted events? I don't see why this needs to be called out specifically? (Is `age` not returned for redacted events?)

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+
+Some home servers such as Synapse already have a concept of Stream Order. We
+expect that the order defined here could be implemented using Stream Order.
+
+## Potential issues
+
+This explicitly allows receipts to be inconsistent across federation. In
+practice this is already the case in the wild, and is impossible to solve using
+Stream Order. The problems with using Topological Order (and Sync Order) have
+already been outlined.
```
clokep, line 242: Receipts over federation have some concept of being across multiple events. See how [User Read Receipt](https://spec.matrix.org/v1.7/server-server-api/#definition-mreceipt) has multiple event IDs available to it.

Synapse always sends a single event ID.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+MSC3981 would solve many problems for clients because the events in a thread
+would be returned in Sync Order, but this is not true: the proposal will return
+events in Topological Order, which is useless for determining which events are
+read.
+
+### The server could calculate unread status
+
+We could use the definitions within this proposal but avoid calculating what was
+unread on the client. Instead we could ask the server to figure out which rooms
+are unread.
```
clokep, line 264: Does the server already figure out which rooms are unread? Or are you differentiating between "notification count > 0" and "bold" in Element Web speak?

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+server providing messages in Sync Order, so we could happily fetch messages
+recursively and still be able to slot them into the right thread and ordering.
+
+Note that the expectation (from some client devs e.g. me @andybalaam) was that
+MSC3981 would solve many problems for clients because the events in a thread
+would be returned in Sync Order, but this is not true: the proposal will return
+events in Topological Order, which is useless for determining which events are
+read.
+
+### The server could calculate unread status
```
clokep, line 260: I expected this section to suggest adding a boolean flag to each event's `unsigned` that marked whether the event was before or after the receipt!

proposals/4033-event-thread-and-order.md
andybalaam: I removed this part of the MSC to simplify it down to just ordering.

proposals/4033-event-thread-and-order.md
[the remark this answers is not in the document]
andybalaam: Removed now this MSC is focused on ordering only.

proposals/4033-event-thread-and-order.md
andybalaam: Note added in bbf1c94ab

proposals/4033-event-thread-and-order.md
andybalaam: Removed this sentence.

andybalaam:
> Funnily enough we at Beeper have been using this exact approach (`com.beeper.hs.order` field in the `unsigned` portion) in our Synapse fork and I can say it's been useful for simplifying client development: [beeper/synapse#42](https://github.com/beeper/synapse/pull/42)
> 
> We were also planning on putting an order on read receipts as well, but haven't gotten to it yet.
> 
> Great minds and all that.

That is great news. We can use this as evidence of an implementation, which will help this MSC to progress smoothly.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+
+See also [Spec Issue #1167](https://github.com/matrix-org/matrix-spec/issues/1167),
+which calls out this ambiguity about the meaning of "read up to".
+
+### Problems with the current definition
+
+The current definition of *after* is ambiguous, and difficult for clients to
+calculate. It depends on only receiving events via sync, which is impossible
+when we want to do backpagination or to fetch thread messages via the
+`relations` API.
```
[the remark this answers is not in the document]
andybalaam: Yes, clarified in 74933916

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+it sends it to any client, and it should never change. It should,
+loosely-speaking, increase for "newer" messages. It may be negative.
+
+The exact meaning of this order is not as important as the fact that it exists
+and is consistent between clients and servers, and events and receipts.
+
+We do not require that order be consistent across federation. The only
+requirement is that all the clients for a user and the one server to which they
+connect agree. For this reason, we propose that `order` be included in an
+event's `unsigned` property.
```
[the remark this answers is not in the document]
andybalaam: Applied in 1b668ce9

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,318 @@
...
+  },
+  "type": "m.receipt"
+}</pre>
+
+We propose:
+
+* all events should contain an `order` property inside `unsigned`.
+* all receipts should contain an `order` property alongside `ts` inside the
+  information about an event, which is a cache of the `order` property within
+  the referred-to event.
```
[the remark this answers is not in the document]
andybalaam: Although other users' receipts are less important for the user experience than the current user's, they suffer from the same problems as the current user's, so yes, all receipts.

This means that we always have a consistent view of the read status of other users, but it does mean that we will sometimes see different status from what they themselves see, due to differences across federation. This is currently true, and this MSC does not fix it.

Do you think I should add a comment to this effect somewhere?

proposals/4033-event-thread-and-order.md
andybalaam: I think this MSC as a whole does clarify #1167 and chooses stream order, but with a caveat that if the answer were topological order, we could live with it. Do you disagree? Do I need to be more explicit?

proposals/4033-event-thread-and-order.md
andybalaam: No, it is saying that we _don't_ need order to be included with redacted events because if a receipt points at a redacted event, we already have its order within the receipt, so we can still see whether other events are read.

I guess that does leave the question of whether the redacted event itself is read, so it would be convenient if order were included.

andybalaam: `age` is returned within a redacted event.

andybalaam: The spec does not mention `unsigned` in https://spec.matrix.org/unstable/rooms/v10/#redactions and yet I see it in my redacted events, so I am guessing `unsigned` is considered outside of the purview of the redaction algorithm.

Given that, I will add a note saying that `order` should also be included in redacted events :-)

andybalaam: Done in d2fe0f41

proposals/4033-event-thread-and-order.md
andybalaam: I don't understand the relevance of your comment to this MSC. It feels like it could be important, but I'm not getting it, sorry!

proposals/4033-event-thread-and-order.md
andybalaam: Yes, I am differentiating between notifications and unreads. Currently the server provides no information about unreads.

proposals/4033-event-thread-and-order.md
andybalaam: Hmm, I guess it could, but I am thinking at the room level.

andybalaam: I will clarify that in the title.

andybalaam:
@bradtgmurray is the client code open source?

proposals/4033-event-thread-and-order.md
bradtgmurray: Note https://github.com/matrix-org/matrix-spec-proposals/pull/2654 does exist, but it's not yet merged and may not be.

bradtgmurray:
> @bradtgmurray is the client code open source?

Unfortunately no

proposals/4033-event-thread-and-order.md
clokep: I think I was attempting to say that *some* of the mechanisms exist to have consistent receipts over federation. They're not really implemented / finished, however.

tl;dr I don't think you need to make any changes, but wanted to point it out.

proposals/4033-event-thread-and-order.md
clokep: Maybe use "clarified" instead of "unchanged" (and mark the PR as fixing that issue).

proposals/4033-event-thread-and-order.md
clokep: > Note #2654 does exist, but it's not yet merged and may not be.

I don't think this is actually related -- I think @andybalaam is using "unread" to mean "bold" not having separate unread counts.

proposals/4033-event-thread-and-order.md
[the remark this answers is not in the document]
clokep: I guess I find that surprising due to all the other bits of the MSC talking about your own receipt. I assumed it was a typo.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,324 @@
+# MSC4033: Explicit ordering of events for receipts
```
jplatte, line 1: I've heard that Synapse already tracks the timestamp at which is first saw a given event (which is equivalent to `origin_server_ts` if the event came in over client-server API, but not equivalent if it came in over federation).

Wouldn't it make sense to just expose this to clients *as a timestamp* instead of introducing an opaque "order" that's pretty likely to just be that timestamp, but not necessarily across all homeservers so you can't actually rely on it?

andybalaam: Some messages reach the homeserver quite late, but are "known" to be "earlier" because they were explicitly requested by this server from another server in order to do backpagination (i.e. fetching older events). In that case, Synapse gives a negative stream order, and this stream order would be better than the timestamp at which Synapse first saw the event.

proposals/4033-event-thread-and-order.md
andybalaam: Clarified in e650a11a and PR marked as fixing #1167

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,327 @@
...
+This explicitly allows receipts to be inconsistent across federation. In
+practice this is already the case in the wild, and is impossible to solve using
+Stream Order. The problems with using Topological Order (and Sync Order) have
+already been outlined.
+
+## Alternatives
+
+### Solves the same problem MSC3981 Relations Recursion tried to solve
+
+This propsal would not replace
```
ara4n, line 256:
```suggestion
This proposal would not replace
```

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,327 @@
...
+order" where late-arriving messages tend to receive higher numbers, or
+"Topological Order" where late-arriving message tend to receive lower numbers.
+
+We believe that it is better to consider late-arriving messages as unread,
+meaning the client has the information that these newly arrived messages have
+not been read and can choose how to display it (or not). This is what leads us
+to suggest Stream Order as the correct choice.
+
+However, if servers choose Topological Order, this proposal still works - we
+just have what the authors consider undesirable behaviour regarding
```
ara4n, line 219: Worth noting that the undesirability of this could be mitigated by using Ranged Read Receipts (currently a protomsc lost at https://hackmd.io/Gxm8zuuSROeencoJ6gjgSg - i.e. you'd get told "you haven't read this range of delayed messages which just turned up yet".

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,327 @@
...
+  }
+}</pre>
+
+Example receipt (changes are highlighted in bold):
+
+<pre>{
+  "content": {
+    "$1435641916114394fHBLK:matrix.org": {
+      "m.read": {
+        "@rikj:jki.re": {
```
ara4n, line 139:
```suggestion
        "@erikj:jki.re": {
```

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,327 @@
...
+    "formatted_body": "&lt;b&gt;This is an example text message&lt;/b&gt;",
+    "msgtype": "m.text"
+  },
+  "event_id": "$143273582443PhrSn:example.org",
+  "origin_server_ts": 1432735824653,
+  "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
+  "sender": "@example:example.org",
+  "unsigned": {
+    "age": 1234,
+    <b>"order": 56764334543</b>
```
ara4n, line 108: This feels like a really nice solution, and I'm pretty confused on how we've got far without having an explicit ordering.  Fully supportive of this :) 🚀

ara4n (approved):

proposals/4033-event-thread-and-order.md
clokep: @andybalaam and I discussed this a bit, but there's lots of questions around the UX of handling that.

If it is easier to reference by me making a "real" MSC out of that, I can push the current content as a draft PR.

andybalaam: I agree that ranged receipts might help with this undesirable behaviour, but I consider this very much a side issue of this MSC, so happy to link to it if you do create it, but I don't see a burning need.

I do note that the current receipts are essentially a special case of ranged receipts, so without having read the proposal, I would expect that the same issues in terms of event order would apply to them, so I'd expect this MSC to become a dependency of that one...

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,346 @@
...
+Note that the `order` property for a particular event will probably be the same
+for every user, so will be repeated multiple times in an aggregated receipt
+event. This structure was chosen to reduce the chance of breaking existing
+clients by introducing `order` at a higher level.
+
+### Proposed definition of *after*
+
+We propose that the definition of *after* should be:
+
+* Event A is after event B if its order is larger.
```
bradtgmurray, line 176: How should we handle negative orders?

Synapse uses negative orders when it needs to backfill events. One case of this is a server purging history but then backfilling again from federation. When this occurs, these events get negative stream orders, but represent the order in which the backfilled events arrived at the homeserver, with more recent messages getting lower (less legative) values.

This unfortunately means that events with negative stream orders don't really have a relation to their order in the room, and just instead represent the order in which clients attempted to fetch events from a room that was purged. For example, if a client called /context in an arbitrary place in the history and then paginated forwards and backwards we'd get the /context events having the lowest stream ordering followed by the order the client paginated in.

One idea could just be to omit `order` from the events in the cases where the stream order is negative. We can just say that these events are unorderable, and read receipts are best effort at that point. This hopefully isn't too big of a problem, as these events are only likely to occur in rare cases or deep in the rooms history.

andybalaam: Good point. I'd be inclined to say that negative order just means "before all receipts" or "always read" so that these messages can never cause a new notification or unread room. Would that make sense?

bradtgmurray: That makes sense to me, but curious if others weigh in.

Note these events will never come down to clients over sync either or trigger push rules.

kegsay: So long as the ordering is consistent, I'm happy to have negative values, even if they don't correlate with actual event orderings.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,346 @@
+# MSC4033: Explicit ordering of events for receipts
```
zecakeh, line 1: This MSC only mentions receipts, but not fully read markers which are similar but a different section of the spec and have the same issue.

Is it intentionally out of scope of this MSC? If so, it would probably be worth mentioning it.

andybalaam: Good point. I added a note in 605eaddf . I don't want to expand this MSC to have a wider scope, but fixing fully-read markers to have the same `order` information seems like the way to go.

proposals/4033-event-thread-and-order.md
andybalaam: Added some notes about this in d749fb12

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,386 @@
...
+
+The [spec](https://spec.matrix.org/unstable/client-server-api/#receipts) states
+that receipts are "read-up-to" without explaining what order the events are in,
+so it is difficult to decide whether an event is before or after a receipt.
+
+We propose adding an explicit order number to all events, so that it is clear
+which events are read.
```
timokoesters, line 8: The problem with numbers is that you can't fill gaps between them, so if you already have 15 and 16, there's no place inbetween them. Instead, you could treat the order as a string of digits instead with the property that
15 < 151 < 16

For easier understanding, you can compare this to decimal numbers: 0.15 < 0.151 < 0.16

programmerjake: if you need to be able to insert a new value between any existing pair, you can use [Dyadic Rationals](https://en.wikipedia.org/wiki/Dyadic_rational), basically a value of the form `a * 2^b` where `a` and `b` are integers.

bnjbvr:
As a SDK implementer, I am *super* interested in this proposal, as it could be useful for other purposes too, in addition to making computing the unread badges very simple. For instance, I think this would allow figuring whether we have gaps when reconciling a local/cached timeline of events with the results of a /messages query.

proposals/4033-event-thread-and-order.md
toger5: Using an array as the index is also a concept that can be found in CRDT's.
`[1]`, `[1,1]`, `[1,2]`, `[1,2,1]`,`[2]`

bnjbvr: In terms of storage/memory used per event (thus bandwidth usage), dyadic rationals sound optimal: sufficiently large space, only require two u32 or two u64. Strings or array representations would require bigger allocations. (Nobody really proposed it, but FP numbers would be a waste in terms of storage space b/o all the `NaN`s + IEEE754 is hard to get right.)

programmerjake: dyadic rationals only require one byte for the exponent, unless you want to support a value with >256 bits (imo excessive), so they only need 5 bytes (32-bit mantissa) or 9 bytes (64-bit mantissa), though you can definitely use more for alignment or convenience.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,386 @@
+# MSC4033: Explicit ordering of events for receipts
```
erikjohnston, line 1: Just as a quick note: I have strong suspicions that: a) this is not possible to do in a way that doesn't limit scalability of servers, b) this isn't necessarily the order in which clients should render events, and c) trying to handle filling in gaps is complicated as you need to deal with edge cases.

I'm sure we can do something close here, but needs investigation from server teams to figure out viability. 

One change that might help a lot here is: instead of inserting an ordering on all events, instead have an opaque "receipt_key" field for events coming down `/sync` or `/messages`, which you can pass to the `/receipt` API. This *is* a lot less powerful than a full linearized history, but feels immediately more viable.

proposals/4033-event-thread-and-order.md
erikjohnston: Note that we will need to be able to handle edge cases where we run out of exponents, e.g. you have an event A and you keep inserting events just after it (which may very well end up a common case). If you only have a one byte exponent then you quickly run out of room.

programmerjake: if you run out of exponent bits, that also means the mantissa has to be at least 256-bits wide (or something like that), so you will need big-integer arithmetic for that...i think that's likely excessive.

maybe a better idea is to maintain a binary tree with one node per message and have the server send the list of changed nodes (if the message's contents doesn't change, all that needs to be sent is the node ids of the tree's children, since that's all that changes during tree balancing), this allows tree balancing to avoid the tree getting too deep.

messages would then be ordered by their position in an in-order traversal of that tree.

proposals/4033-event-thread-and-order.md
erikjohnston: Oh, also: the server sometimes doesn't know the correct ordering of two historical events immediately (e.g. it has two disconnected chunks of DAG). This is less of a problem for `/sync` and read receipts (as historical events don't go down `/sync` or count as unread, server side at least), but may be more of one for other use cases.

programmerjake: > Oh, also: the server sometimes doesn't know the correct ordering of two historical events immediately (e.g. it has two disconnected chunks of DAG).

this is easily handled by my [binary tree idea](https://github.com/matrix-org/matrix-spec-proposals/pull/4033#discussion_r1488348049), the server just picks an arbitrary spot in the tree and then changes it when it knows the correct spot. since the tree nodes' order can be changed without changing message/node ids, this makes it more flexible than using a numeric order.

proposals/4033-event-thread-and-order.md
```diff
@@ -0,0 +1,386 @@
...
+does not have the relevant event) it should not send the receipt. If a server
+later receives an event, allowing it to provide an order for this receipt, it
+should send the receipt at that time. Rationale: without the order, a receipt is
+not useful to the client since it is not able to use it to determine which
+events are read. If a receipt points at an unknown event, the safest assumption
+is that other events in the room are unread i.e. there is no receipt.
+
+If a receipt is received for an event with negative order, the server should set
+the order in the receipt to zero. All events with negative order are understood
+to be read.
```
erikjohnston, line 176: I think this rapidly runs into issues with threads. While we don't currently actively push for backfilled events, I think we *do* mark threads as unread client side if they have unread events (whether they're backfilled or not). In particular, either a) you incorrectly mark those threads as read, or b) make it impossible for threads to be marked as read (since you can't send a read receipt for the latest event in the thread). This even applies to the main timeline if the latest events in it are only backfilled events.

Separately, while we may be able to fudge something here for read receipts, this does limit the use of the `order` param for other use cases. In particular, it means that you cannot sensibly use the `order` param to order messages within the timeline, as half the messages won't have an ordering at all. As discussed elsewhere, calculating an `order` param for backfilled events is practically impossible without the ability to renumber things later, which adds a lot of complexity.

PR #4036, opened 2023-07-13
MSC4036: Room organization by promoting threads

C0ffeeCode:
[Rendered](https://github.com/C0ffeeCode/matrix-spec-proposals/blob/room-organization-by-promoting-threads/proposals/4036-room-organization-by-promoting-threads.md)

This MSC proposes standardization of a room state key to request clients to promote threads
for the purpose of organizing room discussions using threads.


turt2live:
@C0ffeeCode please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes so it is eligible for acceptance later in the process.

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
+# MSC436: Room organization by promoting threads
```
greentore, line 1: I think a good addition to this would be a power level rule regulating who can post messages in the main thread. The use case would be an announcement kind of room where regular users would only be allowed to comment in threads to reduce noise.

C0ffeeCode: Great idea.
But I think it would make more sense to propose it in a separate MSC as the ideas do not depend on each other.

InezMc:
> @C0ffeeCode please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes so it is eligible for acceptance later in the process.

Private sign off has been recieved.

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
...
+
+This MSC solely proposes client-side behavior.
+
+## Proposal
+
+Standardize a room state `m.promote_threads` (`dev.coffeeco.MSC4036.promote_threads` for non-finalized implementations of this MSC) with a Boolean value or an object, defaulting to `false`.
+
+The value SHOULD NOT be set to an object
+but clients MUST consider an object as implicating a value of `true` (possibly with additional behavior changes)
+for the purpose of forwards compatibility.
```
tulir, line 17: Event content must be an object. Even if it didn't have to be, varying types like this are not nice for statically typed languages, and should usually be avoided.

Also, the MSC doesn't explicitly say what it's proposing. This section talks about "a room state" and the unstable prefix says "state key", which is not the same as an event type. You should probably use "room state event" and "event type" respectively, assuming you're proposing a new state event with type `m.promote_threads`. State keys are a different thing, they're used for having multiple events of the same type.

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## Security considerations
+
+Not applicable
+
+## Unstable prefix
+
+For the proposed `m.promote_threads` state key,
+the unstable `dev.coffeeco.MSC4036.promote_threads` shall be used.
```
tulir, line 81: Identifiers like this should be lowercase

```suggestion
the unstable `dev.coffeeco.msc4036.promote_threads` shall be used.
```

(needs to be fixed in a few other places in the MSC too, but you could also just remove them: mentioning unstable prefixes in the main MSC text is not necessary)

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
...
+## Alternatives
+
+Instead of `m.promote_threads` being a value of `true` or `false`,
+it may be considerable that the value also may be an object.
+If the value is an object, it must be treated equally with a value of `true` with the following additions:
+The object may contain a strings,
+for example to specify the label of the button to access the message input box outside the scope of a thread
+(e.g. "Ask a new question", "Open a new issue").
+
+[MSC3088: room-subtyping](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/mutable-subtypes/proposals/3088-room-subtyping.md)
```
tulir, line 59:
```suggestion
[MSC3088: room-subtyping](https://github.com/matrix-org/matrix-spec-proposals/pull/3088)
```

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
+# MSC436: Room organization by promoting threads
+
+Sometimes, having a discussion in a busy room can be difficult.
+People are often simply replying with a message event which is technically not related (`m.relates_to`) using replies (`m.in_reply_to`) or threads (`m.thread`) with the previous messages of the conversation making separation of one conversation from others difficult.
```
tulir, line 4: Lines in proposal text should be limited to 80 or 100 characters

proposals/4036-room-organization-by-promoting-threads.md
```diff
@@ -0,0 +1,86 @@
...
+Hence, this MSC only proposes behavior.
+
+A potential issue may be that clients without support for threads or not following the proposed behavior will continue to reply as before,
+which may cause conversations in threads (as promoted by this MSC) to disappear faster due to new messages and replies of conversations not using threads.
+This may create difficulties with translations.
+
+Also, it may be considered to add the possibility to select specific behavior changes specified above.
+However, this might cause a too broad variation of the user interface, making it harder to understand and less predictable.
+
+## Alternatives
```
tezlm, line 50: Stumbled across this again, I think it would be a good idea to turn this from "promote threads" to "thread only rooms", and introduce a new room type. A big problem in rooms is when some clients support threads and some don't. Having a separate room type would be semantically better and could cause clients that don't support threads to not show thread-only rooms in the ui.

C0ffeeCode: So the thought is to specify clients not supporting threads not to show the room?
A possible concern is that as of now, many clients (and homeserver implementations?) still do not support threads, reducing the usefulness of rooms of such type.

PR #3771, opened 2022-04-15
MSC3771: Read receipts for threads

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/threads-read-receipts/proposals/3771-read-receipts-for-threads.md)

---

Implementations:

* Synapse: matrix-org/synapse#13782
* Element Web: matrix-org/matrix-js-sdk#2635 / matrix-org/matrix-react-sdk#9239

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3771#issuecomment-1238476257)

PR #4037, opened 2023-07-13
MSC4037: Thread root is not in the thread

andybalaam:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/andybalaam/thread-root-is-not-in-thread/proposals/4037-thread-root-is-not-in-thread.md)

Implementations:

* [Synapse](https://github.com/matrix-org/synapse/blob/v1.87.0/synapse/rest/client/receipts.py#L116-L154) has the behaviour specified in this proposal [^1]

[^1]: although it is slightly inconsistent in that it rejects receipts for thread roots that claim to be in the thread, but it accepts receipts for reactions/edits to thread roots that claim to be in the thread, although they do not affect the notification count in the thread.

* [matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L5004) was recently fixed to match the behaviour required by this proposal, as a bug fix for stuck notifications bugs resulting from inconsistency between Synapse and Element Web.

### DIFF
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -83,6 +83,26 @@ We propose that thread roots are in the main timeline, making the definition:
 > thread roots, and other events with non-thread relations to a thread root are
 > in the main timeline.
 
+## How we got here
+
+The MSC that introduced read receipts for threads is
+[MSC3771](https://github.com/matrix-org/matrix-spec-proposals/pull/3771).
+
+The relevant wording is in the
+[Proposal](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3771-read-receipts-for-threads.md#proposal)
+section:
+
+> notifications generated from events with a thread relation matching the
+> receipt’s thread ID prior to and including that event which are MUST be marked
+> as read
+
+Notably it only mentions things "**with a thread relation**", so it appears to
+match the wording of this proposal.
+
+It comes tantalisingly close to covering these issues in the example it uses,
+but unfortunately does not cover what would happen if we received a receipt for
+a thread root or for e.g. an edit of a thread root.
+
 ## Potential issues
 
 None known.
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -39,8 +39,8 @@ This is problematic because:
   they have only read the thread root.
 
 In practice, Synapse
-[ignores any request to mark the thread root as read](https://github.com/matrix-org/synapse/blob/v1.87.0/synapse/rest/client/receipts.py#L116-L154)
-within the thread, and accepts requests to mark it as read in the main timeline.
+[returns an error for any request to mark the thread root as read within the thread](https://github.com/matrix-org/synapse/blob/v1.87.0/synapse/rest/client/receipts.py#L116-L154),
+and accepts requests to mark it as read in the main timeline.
 When it reports thread notifications, it excludes thread roots (and e.g. edits
 to thread roots) from the thread count, only showing them in the main timeline
 count.
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -64,6 +64,9 @@ We propose that thread roots are in the main timeline, making the definition:
 >
 > Events not in a thread but still in the room are considered to be part of the
 > "main timeline": a special thread with an ID of `main`.
+>
+> Note: thread roots (events that are referred to in a `m.thread` relationship)
+> are in the main timeline.
 
 ## Potential issues
 
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -9,7 +9,7 @@ This is important for creating and interpreting read receipts.
 ## Motivation
 
 The current spec, in
-[11.6.2.2 Threaded read receipts](https://spec.matrix.org/unstable/client-server-api/#threaded-read-receipts)
+[11.6.2.2 Threaded read receipts](https://spec.matrix.org/v1.7/client-server-api/#threaded-read-receipts)
 says:
 
 > An event is considered to be "in a thread" if it meets any of the following
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -50,6 +50,9 @@ timeline: any message can become a thread root at any time, when a user creates
 a new threaded message pointing at it, so suddenly switching which receipts are
 allowed to apply to it would not be sensible.
 
+Similarly, it does not make sense for reactions to the thread root (or other
+related events such as edits) to be outside the main timeline.
+
 ## Proposal
 
 We propose that thread roots are in the main timeline, making the definition:
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -67,6 +67,7 @@ We propose that thread roots are in the main timeline, making the definition:
 
 ## Potential issues
 
+None known.
 
 ## Alternatives
 
@@ -77,10 +78,12 @@ been read. A thread cannot exist without at least one additional message.
 
 ## Security considerations
 
+Unlikely to have any security impact.
 
 ## Unstable prefix
 
+None needed.
 
 ## Dependencies
 
-
+No dependencies.
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -83,7 +83,7 @@ We propose that thread roots are in the main timeline, making the definition:
 >
 > Note: thread roots (events that are referred to in a `m.thread` relationship)
 > are in the main timeline. Similarly, reactions to thread roots, edits of
-> thread root, and other events with non-thread relations to a thread root are
+> thread roots, and other events with non-thread relations to a thread root are
 > in the main timline.
 
 ## Potential issues
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -43,7 +43,7 @@ the thread, and accepts requests to mark it as read in the main timeline.
 
 In consequence, Element Web exhibited bugs relating to unread rooms while its
 underlying library used spec-compliant behaviour, many of which were fixed by
-[adoptingthe behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
+[adopting the behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
 
 It really does not make sense to treat thread roots as outside the main
 timeline: any message can become a thread root at any time, when a user creates
--- a/proposals/4037-thread-root-is-not-in-thread.md
+++ b/proposals/4037-thread-root-is-not-in-thread.md
@@ -52,7 +52,13 @@ a new threaded message pointing at it, so suddenly switching which receipts are
 allowed to apply to it would not be sensible.
 
 Similarly, it does not make sense for reactions to the thread root (or other
-related events such as edits) to be outside the main timeline.
+related events such as edits) to be outside the main timeline, for similar
+reasons: the message we are reacting to can become a thread root at any time,
+making our previous receipt invalid retrospectively. (We could conceivably allow
+receipts to exist both within a thread and the main timeline, but this does not
+match the expected user mental model: I have either read a reaction/edit/reply,
+or I have not - I don't want to have to read it twice just because it appears in
+two places in the UI.)
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,92 @@
...
+* It normally does not make sense for a threaded read receipt to point at the
+  thread root, since the user has not really read anything in that thread if
+  they have only read the thread root.
+
+In practice, Synapse ignores any request to mark the thread root as read within
+the thread, and accepts requests to mark it as read in the main timeline.
+
+In consequence, Element Web exhibited bugs relating to unread rooms while its
+underlying library used spec-compliant behaviour, many of which were fixed by
+[adoptingthe behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
```
jplatte, line 46:
```suggestion
[adopting the behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
```

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,92 @@
...
+* It seems natural for messages that are displayed in the main timeline (as
+  thread roots are in most clients) to be considered read/unread when the user
+  reads them in the main timeline.
+
+* It normally does not make sense for a threaded read receipt to point at the
+  thread root, since the user has not really read anything in that thread if
+  they have only read the thread root.
+
+In practice, Synapse ignores any request to mark the thread root as read within
+the thread, and accepts requests to mark it as read in the main timeline.
```
clokep, line 42: Looking at the code I *think* it was intended to accept either, but I would need to do more testing to be sure about that. (And being not ambiguous is good. 😄 )

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,92 @@
...
+>   the thread root).
+> * Following the event relationships, it has a parent event which qualifies for
+>   one of the above. Implementations should not recurse infinitely, though: a
+>   maximum of 3 hops is recommended to cover indirect relationships.
+>
+> Events not in a thread but still in the room are considered to be part of the
+> "main timeline", or a special thread with an ID of `main`.
+
+This explicitly includes thread roots in the thread which branches off them, and
+implicitly _excludes_ those messages from being in the `main` thread.
```
clokep, line 29: I don't think it was meant to exclude this from being in the main *timeline* as the main *timeline* is not a *thread*.

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,92 @@
...
+In consequence, Element Web exhibited bugs relating to unread rooms while its
+underlying library used spec-compliant behaviour, many of which were fixed by
+[adopting the behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
+
+It really does not make sense to treat thread roots as outside the main
+timeline: any message can become a thread root at any time, when a user creates
+a new threaded message pointing at it, so suddenly switching which receipts are
+allowed to apply to it would not be sensible.
+
+## Proposal
```
clokep, line 53: It is implied but the changes, but just to call it out -- this also means events related to the thread root are no longer part of the thread.

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: But in terms of receipts, the spec talks about the main timeline as a thread. So I think it accidentally does exclude them. Which is clearly not intended in my opinion.

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: I thought I said this somewhere, but will add it here, thanks.

andybalaam: Note add here: 651824f5

proposals/4037-thread-root-is-not-in-thread.md
clokep: I don't think it was intentional from the design. 👍

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,95 @@
...
+The current spec implies that a thread root is considered within the thread, but
+we argue that this does not make sense, and a thread root is not "in" the thread
+branching from it.
+
+This is important for creating and interpreting read receipts.
+
+## Motivation
+
+The current spec, in
+[11.6.2.2 Threaded read receipts](https://spec.matrix.org/unstable/client-server-api/#threaded-read-receipts)
```
turt2live, line 12: Please link to a versioned edition of the spec, not unstable. Ie: v1.7

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,95 @@
+# MSC4037: Thread root is not in the thread
+
+The current spec implies that a thread root is considered within the thread, but
+we argue that this does not make sense, and a thread root is not "in" the thread
+branching from it.
```
turt2live, line 5: The proposal later goes into detail about how the implementation differs from the specification slightly, which has me concerned the spec is just wrong compared to the MSCs which lead to it. Normally the SCT aims to figure out whether the implementation or spec is wrong before opening an MSC, but we're already here, so: can we get words relating to the research of the threads MSCs to (dis)prove the spec being wrong? 

If it turns out that the spec is just outright wrong due to clear wording in the MSCs, we can bypass the FCP process and go to spec clarification, but will depend on the MSC research.

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,95 @@
...
+
+* It seems natural for messages that are displayed in the main timeline (as
+  thread roots are in most clients) to be considered read/unread when the user
+  reads them in the main timeline.
+
+* It normally does not make sense for a threaded read receipt to point at the
+  thread root, since the user has not really read anything in that thread if
+  they have only read the thread root.
+
+In practice, Synapse ignores any request to mark the thread root as read within
```
turt2live, line 41: permalink to code please :)

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,95 @@
...
+underlying library used spec-compliant behaviour, many of which were fixed by
+[adopting the behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
+
+It really does not make sense to treat thread roots as outside the main
+timeline: any message can become a thread root at any time, when a user creates
+a new threaded message pointing at it, so suddenly switching which receipts are
+allowed to apply to it would not be sensible.
+
+Similarly, it does not make sense for reactions to the thread root (or other
+related events such as edits) to be outside the main timeline.
```
turt2live, line 54: This needs a why - there's an argument for root messages to be in the main timeline, but from an API clarity perspective it does make sense to consider a reaction "in" the thread. The arguments made above don't appear to hold for reactions, etc.

turt2live (changes requested):

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: Thanks, fixed in 62d4f51c

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: I added an argument in cb06a85c - let me know whether it's convincing. If it's not, please DM me to help me understand: I find it almost conclusively convincing ;-)

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: The relevant MSC is [MSC3771: Read receipts for threads](https://github.com/matrix-org/matrix-spec-proposals/pull/3771)

Here is some wording:

> notifications generated from events with a thread relation matching the receipt’s thread ID prior to and including that event which are MUST be marked as read

Notably it only mentions things "**with a thread relation**", so it appears to match the wording of this proposal.

On the other hand, further experimentation with Synapse shows that it does allow and process receipts for edits to thread roots on BOTH the main timeline AND the thread, so Synapse does not fully align with the behaviour we propose.

andybalaam: Should I add this research to the MSC itself, or is reporting it here the right thing to do?

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -82,7 +82,9 @@ We propose that thread roots are in the main timeline, making the definition:
 > "main timeline": a special thread with an ID of `main`.
 >
 > Note: thread roots (events that are referred to in a `m.thread` relationship)
-> are in the main timeline.
+> are in the main timeline. Similarly, reactions to thread roots, edits of
+> thread root, and other events with non-thread relations to a thread root are
+> in the main timline.
```
jplatte:
```suggestion
> in the main timeline.
```

proposals/4037-thread-root-is-not-in-thread.md
turt2live: It should be in the MSC for now as we're in the information gathering stage - next step is to take all that information and figure out how we got here, and what the next steps are to fix it.

andybalaam: Added in 61765169.

> On the other hand, further experimentation with Synapse shows that it does allow and process receipts for edits to thread roots on BOTH the main timeline AND the thread, so Synapse does not fully align with the behaviour we propose

Note: the above was wrong, and the correct situation is now reflected in the MSC text. Synapse's behaviour basically matches this proposal.

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,129 @@
...
+two places in the UI.)
+
+## Proposal
+
+We propose that thread roots and their non-thread children are in the main
+timeline, making the definition:
+
+> An event is considered to be "in a thread" if:
+>
+> * It has a `rel_type` of `m.thread`, or it has an ancestor event with this
```
turt2live, line 74: Spec PR author: do not bring "ancestor" into the spec, please. We already have terminology to describe parent relationships.

proposals/4037-thread-root-is-not-in-thread.md
```diff
@@ -0,0 +1,129 @@
...
+In practice, Synapse
+[returns an error for any request to mark the thread root as read within the thread](https://github.com/matrix-org/synapse/blob/v1.87.0/synapse/rest/client/receipts.py#L116-L154),
+and accepts requests to mark it as read in the main timeline.
+When it reports thread notifications, it excludes thread roots (and e.g. edits
+to thread roots) from the thread count, only showing them in the main timeline
+count.
+
+In consequence, Element Web exhibited bugs relating to unread rooms while its
+underlying library used spec-compliant behaviour, many of which were fixed by
+[adopting the behaviour recommended by this proposal](https://github.com/matrix-org/matrix-js-sdk/pull/3600).
```
turt2live, line 51: [MSC3771](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3771-read-receipts-for-threads.md) (merged) originally introduced the threaded receipts behaviour into the spec in v1.4, but did not describe what actually constitutes what is "in a thread". Though, the implied wording of MSC3771 does *not* include the thread root in the thread for purposes of notifications. This is further backed up by Synapse being the primary cited implementation proof for MSC3771, which is backed up by this proposal's text in the above paragraph.

As mentioned elsewhere, usually when disagreements between implementation and specification happen we (the SCT) prefer to handle it via another medium, however since an MSC was already created we are using it as the coordination point. 

Given the above, and the references this MSC provides (including a bug fix for Element Web via the js-sdk), I'm confident that this would be best handled as a regular specification bug/clarification issue rather than as an MSC, given the spec is incorrect for not matching the original MSC3771 text/intent. 

I believe it's still valuable for the spec to reference the root being "in a thread" for representation purposes, however for notifications and specifically read receipts I strongly agree that it doesn't make sense to count the thread root as needing a dedicated read receipt. This is ratified by MSC3771 and clarified in depth within this MSC.

@andybalaam as author for this MSC, I'd ask that you take on the spec clarification work via a regular spec PR and close this MSC. @matrix-org/spec-core-team dissenting opinions to mine are welcome, up to and including the point where a spec PR is up for review on this. 

I don't think sending this through FCP would uncover much else in terms of opinions, and sending it through FCP to close it feels like excessive process burden.

turt2live (commented):
see contained comment thread

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: Great, thank you. I will work on this.

andybalaam:
Replaced by spec PR https://github.com/matrix-org/matrix-spec/pull/1677

proposals/4037-thread-root-is-not-in-thread.md
andybalaam: The spec PR is at https://github.com/matrix-org/matrix-spec/pull/1677

PR #4039, opened 2023-07-24
MSC4039: Add an MSC for a new Widget API action to upload files into the media repository

dhenneke:
[Rendered](https://github.com/nordeck/matrix-spec-proposals/blob/nic/feat/widgetapi-upload-files/proposals/4039-widget-api-media.md)

### DIFF
--- a/proposals/4039-widget-api-media.md
+++ b/proposals/4039-widget-api-media.md
@@ -1,0 +1,142 @@
+# MSC4039: Access the Content repository with the Widget API
+
+[MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/2762) specifies a Widget API that
+is able to receive events from and write events to the room that is loaded in the client. But beside
+the room events, homeservers also manage media content for the users. While widgets can already
+download media, given they acquired the URL of the homeserver, they can't upload new media into the
+room. This would be useful to post images or attachments, but also to upload larger files that can't
+be stored in a room event, like a whiteboard that can easily grow larger than 64kb.
+
+This proposal aims to bring the functionality of uploading media to the content repository into the
+widget specification. It should provide the same features that the
+[“Upload Content” endpoint](https://spec.matrix.org/v1.7/client-server-api/#post_matrixmediav3upload)
+of the Client-Server API provides. 
+
+## Proposal
+
+The widget API is extended with a new url parameter that can be requested and that will tell the
+widget the URL of the homeserver API the user is currently using:
+
+* `matrix_base_url`: The URL of the homeserver API.
+
+The widget API is extended with a two new interface to upload files into the content repository. The
+user must manually approve the following capability before the actions can be used:
+
+- `m.upload_file`: Let the widget access the user directory.
+
+### Get configuration
+
+To trigger the action to get the configuration, widgets will use a new `fromWidget` request with the
+action `get_media_config` which takes the following shape:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "get_media_config",
+  "data": {}
+}
+```
+
+If the widget did not get approved for the capability required to send the event, the client MUST
+send an error response (as required currently by the capabilities system for widgets).
+
+If the event is successfully sent by the client, the client sends the following response:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "get_media_config",
+  "data": {},
+  "response": {
+    "m.upload.size": 1000,
+  }
+}
+```
+
+The `response` is are a mirrored representation of the original `/_matrix/media/v3/config` API.
+
+### Upload file
+
+To trigger the action to upload a file, widgets will use a new `fromWidget` request with the action
+`upload_file` which takes the following shape:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "upload_file",
+  "data": {
+    "file": "some-content"
+  }
+}
+```
+
+Under `data`, all keys are a mirrored representation of the original `/_matrix/client/v3/user_directory/search`
+API. `file` is a `XMLHttpRequestBodyInit` that is supported as a data type in the `postMessage` API.
+
+If the widget did not get approved for the capability required to send the event, the client MUST
+send an error response (as required currently by the capabilities system for widgets).
+
+The client SHOULD NOT modify the data of the request.
+
+If the event is successfully sent by the client, the client sends the following response:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "user_directory_search",
+  "data": {
+    "file": "some-content"
+  },
+  "response": {
+    "content_uri": "mxc://..."
+  }
+}
+```
+
+## Potential issues
+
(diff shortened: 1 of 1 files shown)


robintown (commented):
Thanks for the nice write-up, I think this is very reasonable as an extension to the widget API as it currently exists (and it's an extension I was already expecting to be made sooner or later)

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,142 @@
...
+
+The widget API is extended with a new url parameter that can be requested and that will tell the
+widget the URL of the homeserver API the user is currently using:
+
+* `matrix_base_url`: The URL of the homeserver API.
+
+The widget API is extended with a two new interface to upload files into the content repository. The
+user must manually approve the following capability before the actions can be used:
+
+- `m.upload_file`: Let the widget access the user directory.
```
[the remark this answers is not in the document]
dhenneke: Yes, I corrected it. There were also other places I missed before.

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,141 @@
...
+of the Client-Server API provides. 
+
+## Proposal
+
+The widget API is extended with a new url parameter that can be requested and that will tell the
+widget the URL of the homeserver API the user is currently using:
+
+* `matrix_base_url`: The URL of the homeserver API.
+
+The widget API is extended with a two new interface to upload files into the content repository. The
```
Johennes, line 22:
```suggestion
The widget API is extended with two new interfaces to upload files into the content repository. The
```

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,360 @@
...
+
+[MSC3911][MSC3911] proposes a way to link media (files) to events such that there is a 1:n relationship for event ->
+file. Particularly, each file must be attached to exactly one event, and copied using the proposed copy API to attach it
+again to another event. Since the client is unaware of event content sent by widgets per [MSC2762][MSC2762], it cannot
+automatically handle copying any referenced media and the burden is hence on the widget, which would require a change to
+the API described above to add the `copy` endpoint. Additionally, the event sending payload from MSC2762 will need to be
+extended such that whenever a widget includes references to files in an event, it is required to provide them to the
+client in the `send_event` request so the client knows to attach the files.
+
+## Alternatives
```
HarHarLinks, line 298: we could of course also pass an access token to the widget by URI param or with a request, but that seems to contradict the whole idea of the widget API

HarHarLinks:

I am an active member of the Matrix FOSS community, member of the Matrix Foundation governing board, and employed by Nordeck. My parts of this proposal are written and published as a Nordeck employee.

(Similar applies to the other commits to this PR as indicated by the origin branch coming from the `nordeck` fork.)

HarHarLinks:
Implementations:
- `download_file` (authenticated media part)
  - widget API: https://github.com/matrix-org/matrix-widget-api/pull/98
  - client: https://github.com/matrix-org/matrix-react-sdk/pull/12927
  - widget: https://github.com/nordeck/matrix-widget-toolkit/pull/759
- `timeout_ms` (async media)
  - tbd
- `EncryptedFile` (e2ee)
  - tbd

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,360 @@
...
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "upload_file",
+  "data": {
+    "file": "some-content"
+  }
+}
+```
+
+`data.file` is a `XMLHttpRequestBodyInit` that is supported as a data type in the `postMessage` API.
```
HarHarLinks, line 136:
```suggestion
`data.file` is an `XMLHttpRequestBodyInit` that is supported as a data type in the `postMessage` API.
```

dbkr, line 136: This seems like quite a random type? Why not just blob or something? Do we explicitly want the API to support passing, say, FormData through directly?

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,360 @@
...
+`upload_file` which takes the following shape:
+
+```json
+{
+  "api": "fromWidget",
+  "widgetId": "20200827_WidgetExample",
+  "requestid": "generated-id-1234",
+  "action": "upload_file",
+  "data": {
+    "file": "some-content"
```
dbkr, line 131: 'file' here sort of implies that this is a https://developer.mozilla.org/en-US/docs/Web/API/File which it is not. This sort of goes with my other comment below on what type this is and what it's called.

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,360 @@
...
+  "requestid": "generated-id-1234",
+  "action": "download_file",
+  "data": <see above>,
+  "response": {
+    "file": "some-content"
+  }
+}
+```
+
+`response.file` mirrors `data.file` of the `upload_file` action and is a `XMLHttpRequestBodyInit` that is supported as a
```
dbkr, line 282: As above for 'file', plus it will be the widget's job to support whatever the client decides to send it of the union of different types, which might be an easy trap for widget authors to fall into. The impl in the react-sdk PR just always sends a blob anyway.

proposals/4039-widget-api-media.md
```diff
@@ -0,0 +1,360 @@
...
+## Potential issues
+
+[MSC3911][MSC3911] proposes a way to link media (files) to events such that there is a 1:n relationship for event ->
+file. Particularly, each file must be attached to exactly one event, and copied using the proposed copy API to attach it
+again to another event. Since the client is unaware of event content sent by widgets per [MSC2762][MSC2762], it cannot
+automatically handle copying any referenced media and the burden is hence on the widget, which would require a change to
+the API described above to add the `copy` endpoint. Additionally, the event sending payload from MSC2762 will need to be
+extended such that whenever a widget includes references to files in an event, it is required to provide them to the
+client in the `send_event` request so the client knows to attach the files.
+
```
dbkr, line 297: Also, large files will have to be sent over the postmessage API in one big lump, so this isn't going to scale up to anything over a few tens of MB very well.

proposals/4039-widget-api-media.md
HarHarLinks: This is not easy to answer, since the team members originally proposing that type have left the team. We did not change the type as we updated the MSC because the upload part is already implemented. The reason seems to be "Something that can be sent to XMLHttpRequest.send (typically a File)." https://github.com/matrix-org/matrix-widget-api/pull/86/files#diff-67e6f0676307a87283ca4533526de9b3ac4ff3bafdd0d016947e40e2b3574757R693-R694. However we can follow your reasoning that probably a Blob (which is one of the types forming this union type) would make more sense.

proposals/4039-widget-api-media.md
HarHarLinks: if decide to change the type to Blob, perhaps calling both fields blob would make the most sense.

proposals/4039-widget-api-media.md
HarHarLinks: One alternative already described could be letting widgets directly access (parts of) the C-S API in some way.

proposals/4039-widget-api-media.md
Half-Shot: I'd concur here that a Blob makes sense, and reduces the amount of uncertainty around a type.

For those playing at home.

`type XMLHttpRequestBodyInit = Blob | BufferSource | FormData | URLSearchParams | string;`

So in actual fact we *can* change the MSC to a Blob, and prior implementations shouldn't need to change.

BillCarsonFr:
Here is a rust-sdk implementation https://github.com/matrix-org/matrix-rust-sdk/pull/6354

PR #4040, opened 2023-08-04
MSC4040: Update SRV service name to IANA registration

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/iana/service_name/proposals/4040-update-srv-service-name.md)

Implementation: https://github.com/turt2live/matrix-media-repo/pull/427

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4040#issuecomment-1666078252)

### DIFF
--- a/proposals/4040-update-srv-service-name.md
+++ b/proposals/4040-update-srv-service-name.md
@@ -1,4 +1,4 @@
-# MSC4040: Update SRV service name to IANA registation
+# MSC4040: Update SRV service name to IANA registration
 
 The Internet Assigned Numbers Authority (IANA) holds a registry of all service names and ports in use
 by Internet applications. Unfortunately, the service name `matrix` was registered an eternity ago for
--- a/proposals/4040-update-srv-service-name.md
+++ b/proposals/4040-update-srv-service-name.md
@@ -34,7 +34,7 @@ The new steps thus become:
    and given port with a `Host` header of the original server name (with port). The target server must present
    a valid certificate for the hostname.
 
-3. (unchanged) If the hostname is not an IP literal, a regular HTTPS request is made to `https://<hostname>/.well-known/matrix/server`,
+3. (steps added/deprecated) If the hostname is not an IP literal, a regular HTTPS request is made to `https://<hostname>/.well-known/matrix/server`,
    expecting the schema defined later in this section. 30x redirects should be followed, however redirection
    loops should be avoided. Responses (successful or otherwise) to the `/.well-known` endpoint should be
    cached by the requesting server. Servers should respect the cache control headers present on the response,
--- a/proposals/4040-update-srv-service-name.md
+++ b/proposals/4040-update-srv-service-name.md
@@ -107,7 +107,8 @@ A potential way forward is to remove SRV resolution entirely, however prior effo
 
 ## Security considerations
 
-Increased
+There is increased DNS load with this change, due to having to potentially look up 6+ records in the worst case
+scenario. DNS servers should be more than capable of caching, and handling such a spike of requests.
 
 ## Unstable prefix
 
--- a/proposals/4040-update-srv-service-name.md
+++ b/proposals/4040-update-srv-service-name.md
@@ -1,0 +1,118 @@
+# MSC4040: Update SRV service name to IANA registation
+
+The Internet Assigned Numbers Authority (IANA) holds a registry of all service names and ports in use
+by Internet applications. Unfortunately, the service name `matrix` was registered an eternity ago for
+use in an unrelated project, which makes our usage of it improper at best. On August 4, 2023 our
+registration of port 8448 was accepted, with an included service name of `matrix-fed`. This MSC
+proposes a backwards-compatible change to use this newly registered service name.
+
+The registrations with keyword "matrix" can be found [here](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=matrix).
+
+This fixes https://github.com/matrix-org/matrix-spec/issues/400
+
+See also https://github.com/matrix-org/matrix-spec/issues/1576
+
+## Proposal
+
+The [current specification](https://spec.matrix.org/v1.7/server-server-api/#resolving-server-names) for
+resolving server names requires an implementation to look up `_matrix._tcp` SRV records in some steps.
+Those steps become *deprecated* and new steps which use `_matrix-fed._tcp` are added immediately before
+each existing step.
+
+Deprecation in context of this MSC is to discourage continued use, and to queue the affected steps for eventual
+removal from the specification under the [deprecation policy](https://spec.matrix.org/v1.7/#deprecation-policy).
+
+The new steps thus become:
+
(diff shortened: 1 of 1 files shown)


turt2live:
This MSC has no real need to linger.

@mscbot fcp merge

6543:
just as fun-fact note: its always `4`  4040, 400, ... D:

proposals/4040-update-srv-service-name.md
```diff
@@ -0,0 +1,118 @@
...
+      with a `Host` header containing the `<delegated_hostname>`. The target server must present a valid certificate
+      for `<delegated_hostname>`.
+
+   5. (unchanged) If no SRV record is found, an IP address is resolved using CNAME, AAAA or A records. Requests
+      are then made to the resolve IP address and a port of 8448, using a `Host` header of `<delegated_hostname>`.
+      The target server must present a valid certificate for `<delegated_hostname>`.
+
+4. **(ADDED)** If the `/.well-known` request resulted in an error response, a server is found by resolving an
+   SRV record for `_matrix-fed._tcp.<hostname>`. This may result in a hostname (to be resolved using AAAA or A
+   records) and port. Requests are made to the resolved IP address and port, using 8448 as a default port, with
```
richvdh, line 76: default port makes no sense here. SRV records always specify a port.

turt2live: this is copy/paste from the spec - issue tracked here: https://github.com/matrix-org/matrix-spec/issues/1568

proposals/4040-update-srv-service-name.md
```diff
@@ -0,0 +1,118 @@
...
+   given port number, or 8448 if no port is given. The target server must present a valid certificate for
+   the IP address. The `Host` header in the request should be set to the server name, including the port if
+   the server name included one.
+
+2. (unchanged) If the hostname is not an IP literal, and the server name includes an explicit port, resolve
+   the hostname to an IP address using CNAME, AAAA or A records. Requests are made to the resolved IP address
+   and given port with a `Host` header of the original server name (with port). The target server must present
+   a valid certificate for the hostname.
+
+3. (unchanged) If the hostname is not an IP literal, a regular HTTPS request is made to `https://<hostname>/.well-known/matrix/server`,
```
uhoreg, line 37: I think it's misleading to mark this as "unchanged" when steps have been modified.  Can you change the label to indicate this?  Maybe something like

```suggestion
3. (steps added/deprecated) If the hostname is not an IP literal, a regular HTTPS request is made to `https://<hostname>/.well-known/matrix/server`,
```

proposals/4040-update-srv-service-name.md
```diff
@@ -0,0 +1,118 @@
...
+
+Non-options for names include `mxfed`, `matrixf`, and similar.
+
+A potential way forward is to remove SRV resolution entirely, however prior effort via
+[MSC3922](https://github.com/matrix-org/matrix-spec-proposals/pull/3922) are expected to be *rejected* instead
+(as of writing, August 4th, 2023).
+
+## Security considerations
+
+Increased
```
uhoreg, line 110: ?

uhoreg (commented):
Seems fairly sensible, but the "Security considerations" sections looks like it has been truncated.

uhoreg:
Also, why is this marked as "No implementation is required"?  Shouldn't some server implement the new steps?

proposals/4040-update-srv-service-name.md
richvdh: can we just fix that, please, rather than propagating the brainfart further.

richvdh: (https://github.com/matrix-org/matrix-spec/pull/1615)

proposals/4040-update-srv-service-name.md
```diff
@@ -0,0 +1,119 @@
+# MSC4040: Update SRV service name to IANA registration
```
richvdh, line 1: > There's no unstable implementation required because there's no ability to prefix this. We don't require implementation on trivial MSCs either.

I'm not entirely convinced by either of these arguments.

1. Prefixing isn't the only way we have to prove an implementation. You could use a private federation or, in this case, I think it would be pretty safe to use a temporary deployment connected to the public federation.
2. I don't feel like these are trivial changes.

proposals/4040-update-srv-service-name.md
```diff
@@ -0,0 +1,119 @@
...
+descriptive and short. The author notes that `matrix-server` was attempted in registration, which was declined
+during Expert Review of the application.
+
+`matrix` cannot be taken over via IANA process. It was assigned before dates were put on records in 2000.
+
+Non-options for names include `mxfed`, `matrixf`, and similar.
+
+A potential way forward is to remove SRV resolution entirely, however prior effort via
+[MSC3922](https://github.com/matrix-org/matrix-spec-proposals/pull/3922) are expected to be *rejected* instead
+(as of writing, August 4th, 2023).
```
richvdh, line 106: I am somewhat :-1: on this (MSC4040) and would still prefer that we drop SRV support.

I continue to feel that the current SRV support serves almost entirely to confuse people; a significant portion of the traffic in `#synapse:matrix.org` continues to be people trying and failing to set up federation because they are using SRV records with the wrong certificates.

Erik raises good points at https://github.com/matrix-org/matrix-spec-proposals/pull/3922#discussion_r1022527789 but the fact remains that correctly configuring a deployment using SRV records instead of `.well-known` is so damn hard that virtually nobody does it.  Even `matrix.org` uses a .well-known file, and if anyone could benefit from avoiding the `.well-known`, it's `matrix.org`.

Adding *even more* complexity to it seems like, frankly, a retrograde step. If we deprecated SRV records now and then in future it turns out that we can't live without them, then re-introducing them based on `_matrix-fed._tcp` isn't going to be any harder than it is now.

(also: it's not entirely obvious to me that SRV records are the future: [SVCB](https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-12) records look like they might replace them for our purposes).

Still, if everyone else is in favour, I'll not stand in the way.

-0.5 from me on the [Apache Voting Scale](https://www.apache.org/foundation/voting.html)

richvdh:
@mscbot concern lacks an implementation

proposals/4040-update-srv-service-name.md
turt2live: I would argue they're extremely trivial: https://github.com/turt2live/matrix-media-repo/pull/427

turt2live: I've also deployed this to my own homeserver as a test:

image

richvdh:
@mscbot resolve lacks an implementation

turt2live:
This MSC has completed FCP per the [above comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3958#issuecomment-1671589593).

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1624

turt2live:
Merged 🎉

PR #4041, opened 2023-08-10
MSC4041: http header Retry-After for http code 429

ThomasHalwax:
[Rendered](https://github.com/syncpoint/matrix-spec-proposals/blob/main/proposals/4041-retry-after-header-rate-limiting.md)

This proposes the usage of the http header `Retry-After` in case a home server sends a http response 429 (rate limiting).

The `Retry-After` header handling was implemented in 
* [Synapse](https://github.com/matrix-org/synapse/pull/16136)
* [mautrix](https://github.com/mautrix/go/pull/44/commits/4f046814b89cc94ac0884291142bf52ef6f65b22)

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4041#issuecomment-1908591427)

### DIFF
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -24,26 +24,23 @@ of `Retry-After` (in __seconds__) should be the calculated in order to comply wi
 
 ## Potential issues
 
+With the introduction of the http header `Retry-After` the usage of the `retry_after_ms` property in the response body becomes deprecated.
+
 Existing SDKs may use client libraries that might be able to honor the http header `Retry-After`. Since 
 this header is currently not part of the response developers might have created purpose-built functions
 in order to handle rate-limiting correctly.
 
-If the http header `Retry-After` is introduced existing SDKs might behave differently thus leading to an
-unexpected behaviour.
-
 In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
-`retry_after_ms` property in the response body. 
+`retry_after_ms` property in the response body.
 
-In this case, client libraries are advised to use the values in this order:
+Client libraries are advised to use the values in this order:
 
 1) `Retry-After` http header.
 2) `retry_after_ms` property in the response body.
 
-
 ## Alternatives
 
 N/A
-
 
 ## Security considerations
 
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -32,6 +32,14 @@ in order to handle rate-limiting correctly.
 If the http header `Retry-After` is introduced existing SDKs might behave differently thus leading to an
 unexpected behaviour.
 
+In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
+`retry-after-ms` property in the response body. 
+
+In this case, client libraries are advised to use the values in this order:
+
+1) `Retry-After` http header.
+2) `retry-after-ms` property in the response body.
+
 
 ## Alternatives
 
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -17,20 +17,15 @@ potential that current http libraries offer in terms of automated retry handling
 ## Proposal
 
 In order to allow developers to make use of the automated retry handling capabilities of http libraries
-home servers should add an http header `Retry-After` in case they respond with an http error 429.
-Since the body of the response already contains a property `retry_after_ms` (in __milliseconds__) the value 
-of `Retry-After` (in __seconds__) should be the calculated in order to comply with the specification in 
-[RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+home servers shall use the http header `Retry-After` in case they respond with an http error 429.
+The value of `Retry-After` (in __seconds__) is meant to be a delay after receiving the response and must be 
+calculated in order to comply with the specification in [RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+
+With the introduction of the http header `Retry-After` the usage of the existing `retry_after_ms` property in the response body becomes deprecated.
 
 ## Potential issues
 
-With the introduction of the http header `Retry-After` the usage of the `retry_after_ms` property in the response body becomes deprecated.
-
-Existing SDKs may use client libraries that might be able to honor the http header `Retry-After`. Since 
-this header is currently not part of the response developers might have created purpose-built functions
-in order to handle rate-limiting correctly.
-
-In order to maintain backward compatibility with existing client libraries home servers must use both the `Retry-After` header and the
+In order to maintain backward compatibility with existing client libraries home servers shall use both the `Retry-After` header and the
 `retry_after_ms` property in the response body.
 
 Client libraries shall use the values in this order:
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -18,9 +18,8 @@ potential that current http libraries offer in terms of automated retry handling
 
 In order to allow developers to make use of the automated retry handling capabilities of http libraries
 home servers should add an http header `Retry-After` in case they respond with an http error 429.
-Since the body of the response already contains a property `retry_after_ms` (in milliseconds) the value 
-of `Retry-After` should be the calculated by rounding up `retry_after_ms / 1000` to the next integer. 
-Doing so would correspond to a (relative) delay (in seconds) as defined in 
+Since the body of the response already contains a property `retry_after_ms` (in __milliseconds__) the value 
+of `Retry-After` (in __seconds__) should be the calculated in order to comply with the specification in 
 [RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
 
 ## Potential issues
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -1,4 +1,4 @@
-# MSC0000: Use http header Retry-After to enable library-assisted retry handling
+# MSC4041: Use http header Retry-After to enable library-assisted retry handling
 
 The current Matrix Client-Server API (v1.7) recommends that home servers should protect themselves from
 being overloaded by enforcing rate-limits to selected API calls.
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -30,10 +30,10 @@ Existing SDKs may use client libraries that might be able to honor the http head
 this header is currently not part of the response developers might have created purpose-built functions
 in order to handle rate-limiting correctly.
 
-In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
+In order to maintain backward compatibility with existing client libraries home servers must use both the `Retry-After` header and the
 `retry_after_ms` property in the response body.
 
-Client libraries are advised to use the values in this order:
+Client libraries shall use the values in this order:
 
 1) `Retry-After` http header.
 2) `retry_after_ms` property in the response body.
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -18,14 +18,14 @@ potential that current http libraries offer in terms of automated retry handling
 
 In order to allow developers to make use of the automated retry handling capabilities of http libraries
 home servers should add an http header `Retry-After` in case they respond with an http error 429.
-Since the body of the response already contains a property `retry_after_ms` (in miliseconds) the value 
+Since the body of the response already contains a property `retry_after_ms` (in milliseconds) the value 
 of `Retry-After` should be the calculated by rounding up `retry_after_ms / 1000` to the next integer. 
 Doing so would correspond to a (relative) delay (in seconds) as defined in 
 [RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
 
 ## Potential issues
 
-Existing SDKs may use client libraries that migth be able to honor the http header `Retry-After`. Since 
+Existing SDKs may use client libraries that might be able to honor the http header `Retry-After`. Since 
 this header is currently not part of the response developers might have created purpose-built functions
 in order to handle rate-limiting correctly.
 
--- a/proposals/4041-retry-after-header-rate-limiting.md
+++ b/proposals/4041-retry-after-header-rate-limiting.md
@@ -33,12 +33,12 @@ If the http header `Retry-After` is introduced existing SDKs might behave differ
 unexpected behaviour.
 
 In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
-`retry-after-ms` property in the response body. 
+`retry_after_ms` property in the response body. 
 
 In this case, client libraries are advised to use the values in this order:
 
 1) `Retry-After` http header.
-2) `retry-after-ms` property in the response body.
+2) `retry_after_ms` property in the response body.
 
 
 ## Alternatives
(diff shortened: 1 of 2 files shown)


turt2live:
@ThomasHalwax please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes here for this MSC to be eligible for FCP in the future.

ThomasHalwax:
Signed-off-by: Your Name <thomas.halwax@syncpoint.io>

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Proposal
+
+In order to allow developers to make use of the automated retry handling capabilities of http libraries
+home servers should add an http header `Retry-After` in case they respond with an http error 429.
+Since the body of the response already contains a property `retry_after_ms` (in miliseconds) the value 
+of `Retry-After` should be the calculated by rounding up `retry_after_ms / 1000` to the next integer. 
+Doing so would correspond to a (relative) delay (in seconds) as defined in 
+[RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+
```
MTRNord, line 25: I am wondering if it makes sense to define backwards compatibility in the msc to define what to do if both are present?

I guess the obvious would be first checking for the header and if not present use the json (until that value is deprecated and removed.)

MTRNord (approved):
Lgtm apart from the one comment. Imho this makes sense and would ease the effort to implement the spec to some extend.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: @MTRNord I've added a note regarding backward compatibility as you suggested.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,59 @@
...
+The current Matrix Client Server API specification does not take this header into account. This wastes the 
+potential that current http libraries offer in terms of automated retry handling.
+
+## Proposal
+
+In order to allow developers to make use of the automated retry handling capabilities of http libraries
+home servers should add an http header `Retry-After` in case they respond with an http error 429.
+Since the body of the response already contains a property `retry_after_ms` (in miliseconds) the value 
+of `Retry-After` should be the calculated by rounding up `retry_after_ms / 1000` to the next integer.
```
Half-Shot, line 22: There is an issue here that you could unfortunately land on a worst case (i.e. `retry_after_ms` for `5001ms` becomes `6 seconds` under this rule.

A separate question is whether we realistically care about subsecond delays, as if anything it allows a service to balance out more. Perhaps leaving in `retry_after_ms` allows more advanced scripts to perfectly limit loads while "dumb" applications simply wait a little longer.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,59 @@
...
+N/A
+
+
+## Security considerations
+
+N/A
+
+## Unstable prefix
+
+N/A
```
Half-Shot, line 55: Do we think we need one? I'd argue not, given it's a stable bit of RFC9110. This spec just outlines how one should feel free to use the header in addition to the retry field.

Half-Shot:
Server side implementation in https://github.com/matrix-org/synapse/pull/16136.

Unsure if we need a client impl, given its a standard header. If needed, maybe @ThomasHalwax can help there.

proposals/4041-retry-after-header-rate-limiting.md
clokep: I agree, but should probably call it out.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,59 @@
...
+
+Existing SDKs may use client libraries that migth be able to honor the http header `Retry-After`. Since 
+this header is currently not part of the response developers might have created purpose-built functions
+in order to handle rate-limiting correctly.
+
+If the http header `Retry-After` is introduced existing SDKs might behave differently thus leading to an
+unexpected behaviour.
+
+In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
+`retry-after-ms` property in the response body.
```
clokep, line 36: I think this uses hyphens in a few spots when it should be underscores.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,59 @@
...
+and contains either a (relative) value for the delay in seconds or an (absolute) date.
+
+The current Matrix Client Server API specification does not take this header into account. This wastes the 
+potential that current http libraries offer in terms of automated retry handling.
+
+## Proposal
+
+In order to allow developers to make use of the automated retry handling capabilities of http libraries
+home servers should add an http header `Retry-After` in case they respond with an http error 429.
+Since the body of the response already contains a property `retry_after_ms` (in miliseconds) the value
```
clokep, line 21: I think we should officially deprecate this at the same time.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: The definition in the [RFC](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after) says it has to be in seconds. So there is no way of being more accurate without violating the RFC.

Do you know how the value for `retry_after_ms` is calculated by home servers?

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: Thank you! Fixed it!

ThomasHalwax:
> Server side implementation in [matrix-org/synapse#16136](https://github.com/matrix-org/synapse/pull/16136).
> 
> Unsure if we need a client impl, given its a standard header. If needed, maybe @ThomasHalwax can help there.

It is a standard header but not all clients libs take care about it. The current [matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/scheduler.ts) seems to ignore it which is OK since the current Matrix Spec does not mention it.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: Does anyone know of a client library that is so time critical that rounding up the value could cause it to malfunction?

Half-Shot: Oh yeah not disagreeing with your proposal, I think the RFC is perhaps a bit silly to use seconds.

turt2live: We can leave it as an implementation detail for how to populate a header. Some servers might decide to round up, down, or to nearest. 

For the MSC we should be calling out that it's seconds though, as most developers are used to milliseconds.

proposals/4041-retry-after-header-rate-limiting.md
turt2live: Errors on Matrix are a bit strange: deprecation I think would be inappropriate with their design, but personally I feel deprecation is a good thing. 

The backing context is errors are meant to be richer and provide more detail than the HTTP headers.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: > We can leave it as an implementation detail for how to populate a header. Some servers might decide to round up, down, or to nearest.

I'm not sure this will work. If the value of `retry_after_ms' is 5999 ms and server are rounding down the header value will be 5 seconds. In the worst case clients will be fast enough to make the request before the 5999 ms are over and run immediately into the next 429 error.

turt2live: The server is the one which decides both `Retry-After` and `retry_after_ms`, so it can just not make bad decisions. At the spec level, we are less concerned about preventing conflicts a server can self-resolve.

Half-Shot: A better way to resolve this might be a note within the spec to mention that your header value is in seconds, and therefore the header should be >= that value.

Then implementations have options. Synapse will round up, because we're not chagning the retry_after_ms behaviour.

proposals/4041-retry-after-header-rate-limiting.md
Half-Shot:
```suggestion
Since this MSC is using a standard HTTP header, it will not use a unstable prefix.
```

proposals/4041-retry-after-header-rate-limiting.md
Half-Shot: @ThomasHalwax are you happy to make this change?

proposals/4041-retry-after-header-rate-limiting.md
clokep: I suppose, it just seems odd to duplicate information when there's a standard way to do it. I don't feel strongly, however.

proposals/4041-retry-after-header-rate-limiting.md
anoadragon453: @Half-Shot I think your suggestion [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4041#discussion_r1301864312) equates to the original text of the proposal before this thread started...

proposals/4041-retry-after-header-rate-limiting.md
anoadragon453: In this case, I'm not sure it's worth

anoadragon453:
The Synapse and mautrix-go implementations listed in the PR description look sound. Removing the `needs-implementation` label.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: I'd prefer a clean cut and remove the `retry_after_ms` response. Why do it the matrix-way if there is an HTTP standard for the same purpose?

anoadragon453: We should deprecate instead of cut, as if servers immediately dropped sending `retry_after_ms`, most clients that have not yet implemented `Retry-After` would fail to back off correctly.

ThomasHalwax: Is there a policy in the spec on how to deal with deprecated structures? Like X deprecated in v1.10, will be removed in 1.15 and 2.0?

anoadragon453:
The author has indicated this MSC is ready for review. However, the remaining conversations should be resolved before FCP can be proposed.

Please write CC `@matrix-org/spec-core-team` in said thread(s) if you require guidance on resolving them, or ask for help in the [Office of the Matrix Spec Core Team](https://matrix.to/#/#sct-office:matrix.org) room. Thanks!

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: Updated proposal to deprecate `retry_after_ms`.

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: @anoadragon453 I think @turt2live has a very good point here. It's up to the homer server to deal with the values it sent to the clients. If they round up or down does not matter and should no be mentioned in the spec. The only condition that must be met is that the value for `Retry-After` is in seconds.

anoadragon453: The current text looks fine to me.

anoadragon453:
This proposal looks good to propose FCP merge on. All outstanding conversations have been resolved, and all items in the [MSC Checklist](https://github.com/matrix-org/matrix-spec-proposals/blob/fe587a7d33f9c6fb8122e7ddf73b17b48923df6c/MSC_CHECKLIST.md) pass.

@mscbot fcp merge

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,55 @@
...
+With the introduction of the http header `Retry-After` the usage of the `retry_after_ms` property in the response body becomes deprecated.
+
+Existing SDKs may use client libraries that might be able to honor the http header `Retry-After`. Since 
+this header is currently not part of the response developers might have created purpose-built functions
+in order to handle rate-limiting correctly.
+
+In order to maintain backward compatibility home servers should use both the `Retry-After` header and the
+`retry_after_ms` property in the response body.
+
+Client libraries are advised to use the values in this order:
```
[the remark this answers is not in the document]
KitsuneRal: I agree it should prescribe this order specifically (the other way around doesn't make much sense given the deprecation of `reply_after_ms`, assuming that all platforms we're usually dealing with can read raw HTTP response headers or manage retries directly.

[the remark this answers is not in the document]
ThomasHalwax: Servers MUST return both values and client libs SHALL use the given order.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,55 @@
...
+
+In order to allow developers to make use of the automated retry handling capabilities of http libraries
+home servers should add an http header `Retry-After` in case they respond with an http error 429.
+Since the body of the response already contains a property `retry_after_ms` (in __milliseconds__) the value 
+of `Retry-After` (in __seconds__) should be the calculated in order to comply with the specification in 
+[RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+
+## Potential issues
+
+With the introduction of the http header `Retry-After` the usage of the `retry_after_ms` property in the response body becomes deprecated.
```
dbkr, line 27: This doesn't feel like it belongs in the issues section. If the intention is to deprecate it, that should be part of the main proposal.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,55 @@
...
+this header is currently not part of the response developers might have created purpose-built functions
+in order to handle rate-limiting correctly.
+
+In order to maintain backward compatibility with existing client libraries home servers must use both the `Retry-After` header and the
+`retry_after_ms` property in the response body.
+
+Client libraries shall use the values in this order:
+
+1) `Retry-After` http header.
+2) `retry_after_ms` property in the response body.
```
dbkr, line 39: This also feels like part of the proposal, so I'd say it belongs in the proposal section.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,55 @@
...
+of `Retry-After` (in __seconds__) should be the calculated in order to comply with the specification in 
+[RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+
+## Potential issues
+
+With the introduction of the http header `Retry-After` the usage of the `retry_after_ms` property in the response body becomes deprecated.
+
+Existing SDKs may use client libraries that might be able to honor the http header `Retry-After`. Since 
+this header is currently not part of the response developers might have created purpose-built functions
+in order to handle rate-limiting correctly.
```
dbkr, line 31: There's no mitigation suggested for this problem, so it might be worth explaining why: presumably just not anticipated to be a big enough problem?

proposals/4041-retry-after-header-rate-limiting.md
ThomasHalwax: Moved it into the proposal section.

proposals/4041-retry-after-header-rate-limiting.md
```diff
@@ -0,0 +1,50 @@
...
+In order to allow developers to make use of the automated retry handling capabilities of http libraries
+home servers shall use the http header `Retry-After` in case they respond with an http error 429.
+The value of `Retry-After` (in __seconds__) is meant to be a delay after receiving the response and must be 
+calculated in order to comply with the specification in [RFC 9119 - HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).
+
+With the introduction of the http header `Retry-After` the usage of the existing `retry_after_ms` property in the response body becomes deprecated.
+
+## Potential issues
+
+In order to maintain backward compatibility with existing client libraries home servers shall use both the `Retry-After` header and the
```
uhoreg, line 28:
```suggestion
In order to maintain backward compatibility with existing client libraries homeservers shall use both the `Retry-After` header and the
```

minor nitpick: homeserver is one word

turt2live (approved):

proposals/4041-retry-after-header-rate-limiting.md
clokep: There's a couple of other occurrences of the incorrect spelling as well.

turt2live: Will apply manually after merging - can't edit the PR itself.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1737

turt2live:
Merged 🎉

PR #4042, opened 2023-08-10
MSC4042: Disabled Presence State

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-spec-proposals/blob/FSG-Cat-Disabled-Presence-State/proposals/4042-disabled-presence-state.md)

This simple MSC adds a new presence state to indicate if no data exists. Usually because presence is disabled.

### DIFF
--- a/proposals/4042-disabled-presence-state.md
+++ b/proposals/4042-disabled-presence-state.md
@@ -1,14 +1,14 @@
-# MSC0000: Disabled Presence State
+# MSC4042: Disabled Presence State
 
 In current matrix we have no way to tell clients that we simply do not have presence data at all
 for a given homeserver or user. This proposal addresses this fact by adding a `disabled` state.
 
 By adding a `disabled` state to presence it allows us to kill two birds with one stone. Its used for
-MSC_PRESENCE_OVERRIDE_PLACEHOLDER and for this proposal. In this proposal its simply used to indicate
-a lack of information about a given user. This is going to 9 times out of 10 be because of that presence
-is disabled somewhere in the chain and therefore you cant get data. Be that disabled by the other user
-or their server or your server. If you are on matrix.org for example all presence will return this value
-if this proposal is adopted until they reenable presence. Since they have presence disabled.
+MSC4043 and for this proposal. In this proposal its simply used to indicatea lack of information about 
+a given user. This is going to 9 times out of 10 be because of that presence is disabled somewhere in the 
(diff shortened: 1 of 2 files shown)


proposals/4042-disabled-presence-state.md
```diff
@@ -0,0 +1,52 @@
+# MSC4042: Disabled Presence State
+
+In current matrix we have no way to tell clients that we simply do not have presence data at all
+for a given homeserver or user. This proposal addresses this fact by adding a `disabled` state.
+
+By adding a `disabled` state to presence it allows us to kill two birds with one stone. Its used for
+MSC4043 and for this proposal. In this proposal its simply used to indicatea lack of information about
```
uhoreg, line 7:
```suggestion
[MSC4043](https://github.com/matrix-org/matrix-spec-proposals/pull/4043) and for this proposal. In this proposal its simply used to indicatea lack of information about 
```

proposals/4042-disabled-presence-state.md
```diff
@@ -0,0 +1,52 @@
...
+The alternatives for this proposal are well not that many if any? Like a way to indicate we don't have any data
+is useful. The only change I can think of is splitting intentional lack of data from we just don't have data yet.
+But I fail to find that distinction useful.
+
+
+## Security considerations
+
+Presence should not be security relevant as far as the author is aware. The only exception is the privacy
+discussion. This specific proposal should not have any impact on privacy because of the fact that this proposal
+does not it self change anything in practice. Presence being disabled is not a secret its a well known fact.
```
HarHarLinks, line 43: Doesn't this weaken the fact that one could mask offline as disabled or vice versa? I think generally for the existing kinds of presence there is deliberate difference in user hands, but this is slightly different.

FSG-Cat: This MSC is written for a context where disabled is used for a whole homeserver or for users who dont care about sending that message.

Considering that this MSC is intended to be used together with the ability to force presence states into whatever value you like to use in a perfect world the privacy argument becomes much weaker due to that factor too.

Like presence is not perfect in 2026 and it was not in 2023. This MSC is essentially just one of the parts of a greater presence redesign that Cat actually came to write a proposal for even if its imperfect.

Edit: And i should say that the whole text about that its no secret is more refering to that in practice its trivial to discover that presence is not enabled for a given matrix homeserver or atleast it was in 2023.

PR #4043, opened 2023-08-10
MSC4043: Presence Override API

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-spec-proposals/blob/FSG-Cat-Presence-Override/proposals/4043-presence-override-api.md)

This MSC adds a new API that lets you force a specific presence state. 

### DIFF
--- a/proposals/4043-presence-override-api.md
+++ b/proposals/4043-presence-override-api.md
@@ -71,7 +71,7 @@ positive effect is privacy related not strictly security in that users can cloak
 ## Unstable prefix
 
 While this proposal is unstable instead of using `/_matrix/client/v1/presence/{userId}/override` you use
-`/_matrix/client/unstable/support.feline.msc4043.v1/override`
+`/_matrix/client/unstable/support.feline.msc4043.v1/presence/{userId}/override`
 
 And unstable feature flag used is `support.feline.msc4043.v1`
 
--- a/proposals/4043-presence-override-api.md
+++ b/proposals/4043-presence-override-api.md
@@ -38,6 +38,14 @@ And you will get back a response that is like the example below if your override
 }
 ```
 
+And to disable override you send a payload with a empty `presence_override` key like the example below.
+
+```json
+{
+  "presence_override": ""
+}
+```
+
 As for error codes this is currently WIP but it follows a similar pattern to the current `/_matrix/client/v3/presence/{userId}/status`
 endpoint. With minor reasonable adaptations like how presence is replaced with presence override where sensible.
 
--- a/proposals/4043-presence-override-api.md
+++ b/proposals/4043-presence-override-api.md
@@ -5,7 +5,7 @@ for presence. Where more important states overrule less important ones. This pro
 
 This proposal proposes a new API that you can call to set a authoritative presence state for your account. This
 state is always used negating all other ways to set presence. So if you set your state to `disabled` from
-MSC4042 then well then your state would always be `disabled`.
+[MSC4042](https://github.com/matrix-org/matrix-spec-proposals/pull/4042) then well then your state would always be `disabled`.
 
 Using a new API for this instead of recycling some old method does come at the benefit of that this just works.
 
--- a/proposals/4043-presence-override-api.md
+++ b/proposals/4043-presence-override-api.md
@@ -1,11 +1,11 @@
-# MSC0000: Presence Override API
+# MSC4043: Presence Override API
 
 In current matrix there exists no mechanism that is reliable to override the supposed to exist totem pole
 for presence. Where more important states overrule less important ones. This proposal fixes that.
 
 This proposal proposes a new API that you can call to set a authoritative presence state for your account. This
 state is always used negating all other ways to set presence. So if you set your state to `disabled` from
-MSC_disabled_presence_state then well then your state would always be `disabled`.
+MSC4042 then well then your state would always be `disabled`.
 
 Using a new API for this instead of recycling some old method does come at the benefit of that this just works.
 
@@ -63,11 +63,11 @@ positive effect is privacy related not strictly security in that users can cloak
 ## Unstable prefix
 
 While this proposal is unstable instead of using `/_matrix/client/v1/presence/{userId}/override` you use
-`/_matrix/client/unstable/support.feline.mscAAAA.v1/override`
+`/_matrix/client/unstable/support.feline.msc4043.v1/override`
 
-And unstable feature flag used is `support.feline.mscAAAA.v1`
+And unstable feature flag used is `support.feline.msc4043.v1`
 
 ## Dependencies
 
-This MSC has MSC_disabled_presence_state as a soft dependency as its a very useful state to override to.
+This MSC has MSC4042 as a soft dependency as its a very useful state to override to.
 This MSC has no hard dependencies and can be used on its own and merged on its own.
--- a/proposals/XXXX-presence-override-api.md
+++ b/proposals/XXXX-presence-override-api.md
@@ -1,0 +1,81 @@
+# MSC0000: Presence Override API
+
+In current matrix there exists no mechanism that is reliable to override the supposed to exist totem pole
+for presence. Where more important states overrule less important ones. This proposal fixes that.
+
+This proposal proposes a new API that you can call to set a authoritative presence state for your account. This
+state is always used negating all other ways to set presence. So if you set your state to `disabled` from
+MSC_disabled_presence_state then well then your state would always be `disabled`.
+
+Using a new API for this instead of recycling some old method does come at the benefit of that this just works.
+
+No existing code can mess up and use this wrong to effectively ruin the whole point of this system because its new.
+
+Being able to set your authoritative presence state is seen as beneficial due to that it allows users to have full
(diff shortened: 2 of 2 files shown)


proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,73 @@
...
+## Proposal
+
+To set a new presence override you call the new `/_matrix/client/v1/presence/{userId}/override` endpoint
+using a PUT request. Using a payload that can look like the example below.
+
+```json
+{
+  "presence_override": "online"
+}
+```
```
clokep, line 27: Could this be an argument on the current `/presence` API? Either another body parameter or the keyword parameter. It could default to not-forcing to be backwards compatible.

FSG-Cat: Yes that is tbh one possible way to do it. use normal /presence but add a new argument. And ofc yes defaulting to not forcing is logical if we add new functionality since why break backwards compatibility for no reason.

The key problem to solve is some way to get us to a state where we can define a value that is set as the canonical value and therefore skipping the whole Online is more important than offline situation. 

And to disable it you send a override that is empty makes sense in this context. Because if the argument is missing that means backwards compatibility mode but argument present but empty means disable override.

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
+# MSC4043: Presence Override API
```
Half-Shot, line 1: I'm keen for something like this.  It would scratch a serious itch with current bridges where we have to repeatedly set the presence every N seconds (where N is a value that Synapse hard codes...) to keep presence "alive".

I don't think this is even what I'd call an override, but an alternative and valid way to update your presence for clients that do not fall under the traditional syncing system (a.k.a, anything that doesn't call /sync).

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+  "presence_override": ""
+}
+```
+
+As for error codes this is currently WIP but it follows a similar pattern to the current `/_matrix/client/v3/presence/{userId}/status`
+endpoint. With minor reasonable adaptations like how presence is replaced with presence override where sensible.
+
+## Potential issues
+
+Other than this functionality being duplicate the author does not foresee any potential issues other than
```
Half-Shot, line 54: This is adding additional complexity to an already complex system, so there is definitely a scare here that making presense **harder** to implement for homeservers might be tricky.

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+
+Using a new API for this instead of recycling some old method does come at the benefit of that this just works.
+
+No existing code can mess up and use this wrong to effectively ruin the whole point of this system because its new.
+
+Being able to set your authoritative presence state is seen as beneficial due to that it allows users to have full
+control over their presence status instead of having to rely on all clients on their account collaborating to not
+ruin the intention of the user.
+
+## Proposal
```
Half-Shot, line 18: My naive understanding of presence is that whenever a user updates their presence, the presence is then sent round to every interested server and user.

Obviously this API sets a static presence, so wouldn't be sent out more than once. How would this interaction work with other users and homeservers which presumably expect presence to be short lived? Does the homeserver need to repeatedly announce the presence, or does the federation API need extending?

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+
+To set a new presence override you call the new `/_matrix/client/v1/presence/{userId}/override` endpoint
+using a PUT request. Using a payload that can look like the example below.
+
+```json
+{
+  "presence_override": "online"
```
Half-Shot, line 25: Is there any reason not to include `status_msg`, and have that persist too? This would make it compatible with the existing presence system

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Security considerations
+
+This proposal should not as far as the author is aware have security implications that are negative. The primary
+positive effect is privacy related not strictly security in that users can cloak their presence state reliably.
+
+## Unstable prefix
+
+While this proposal is unstable instead of using `/_matrix/client/v1/presence/{userId}/override` you use
+`/_matrix/client/unstable/support.feline.msc4043.v1/override`
```
Half-Shot, line 74:
```suggestion
`/_matrix/client/unstable/support.feline.msc4043.v1/presence/{userId}/override`
```

maybe? You don't have a presence or userId field so we don't know who you are overriding :)

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+
+```json
+{
+  "presence_override": ""
+}
+```
+
+As for error codes this is currently WIP but it follows a similar pattern to the current `/_matrix/client/v3/presence/{userId}/status`
+endpoint. With minor reasonable adaptations like how presence is replaced with presence override where sensible.
+
```
Half-Shot, line 51: Unsure if this is relevant in this MSC, but presumably user deactivation voids the presence status?

Half-Shot (commented):
Lots of thoughts. As stated I have interest in having this for bridges to address our presence issues too, I think this could work but needs fleshing out.

proposals/4043-presence-override-api.md
Half-Shot: Sidenote: I think we should really have a key in presence updates that specify a `ttl`, with `0` being a valid way to say presence is valid until a new updates appears.

proposals/4043-presence-override-api.md
FSG-Cat: Thanks for catching this oversight.

proposals/4043-presence-override-api.md
FSG-Cat: I would agree that since Deactivation voids a lot of other data like this its only fair to include that deactivation voids this too.

proposals/4043-presence-override-api.md
FSG-Cat: Now that you mention it status has the exact same problem as presence it self currently so it makes sense to include. Yes its tbh a oversight and me thinking of a too narrow scope.

proposals/4043-presence-override-api.md
FSG-Cat: Yes that is indeed a problem how to solve remotes issue if we dont resend every X. A Presence TTL definetively makes sense and 0 being valid until otherwise noted makes sense. 

I mean i have a rough draft sitting in the drawer for a solution to this but i think its a bit too radical to move all of presence to be until i say otherwise like that draft is. 

For now i will note that a possible solution to the federation problem is to have the homeserver re emit presence for the user every X and leave X as an implementation detail. Because that way this would work over federation in a backwards compatible way. This also makes it so your presence heals from EDUs vanishing into the ether etc.

proposals/4043-presence-override-api.md
FSG-Cat: Isnt PUT /_matrix/client/v3/presence/{userId}/status the API that you use if you want to set presence in a non Syncing way? 

But if you mean that the API mentioned earlier has the problem with every N seconds then yes this would help bridges. 

My only concern is for some bridges where they access your own account to make a seamless experience those bridges would not be able to use this at the same time as the user is using it. Tho for those bridges that can easily be a user preference option so that concern is probably in implementation detail land. 

The bridges that are not sharing their accounts can use this API in the exact same way as anyone else without conflict concerns.

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
+# MSC4043: Presence Override API
+
+In current matrix there exists no mechanism that is reliable to override the supposed to exist totem pole
+for presence. Where more important states overrule less important ones. This proposal fixes that.
+
+This proposal proposes a new API that you can call to set a authoritative presence state for your account. This
+state is always used negating all other ways to set presence. So if you set your state to `disabled` from
+MSC4042 then well then your state would always be `disabled`.
```
uhoreg, line 8:
```suggestion
[MSC4042](https://github.com/matrix-org/matrix-spec-proposals/pull/4042) then well then your state would always be `disabled`.
```

proposals/4043-presence-override-api.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+As for error codes this is currently WIP but it follows a similar pattern to the current `/_matrix/client/v3/presence/{userId}/status`
+endpoint. With minor reasonable adaptations like how presence is replaced with presence override where sensible.
+
+## Potential issues
+
+Other than this functionality being duplicate the author does not foresee any potential issues other than
+that clients that don't support this feature will not be able to set the override status for the user.
+
```
stratself, line 56: At the time of writing, what was Cat's consideration with respect to stale presence due to having this endpoint never unset? For example a bridge could be improperly decommissioned and never send a `"presence_override": ""`, leaving its users to always be stuck online

FSG-Cat: Appservices are actually the one thing Cat did not consider how a bad decommissioning of them can cause their presence states to be stuck is the genuine answer when this was written.

PR #4044, opened 2023-08-12
MSC4044: Enforcing user ID grammar in rooms

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/strict-user-id-grammar/proposals/4044-strict-user-id-grammar-in-rooms.md)

No known unstable implementations.

### DIFF
--- a/proposals/4044-strict-user-id-grammar-in-rooms.md
+++ b/proposals/4044-strict-user-id-grammar-in-rooms.md
@@ -1,0 +1,88 @@
+# MSC4044: Enforcing user ID grammar in rooms
+
+The specification presently allows for a "historical" grammar for user IDs, described
+[here](https://spec.matrix.org/v1.7/appendices/#historical-user-ids) and clarified by
+[PR #1506](https://github.com/matrix-org/matrix-spec/pull/1506). The use of this historical
+character set has allowed for "weird" and generally not-great user IDs to appear in the wild,
+causing a variety of issues for software projects.
+
+This proposal uses a future room version to strictly use the
+[non-historical grammar](https://spec.matrix.org/v1.7/appendices/#user-identifiers) for user IDs,
+preventing historical IDs from joining and participating in those room versions.
+
+Matrix does not currently specify a way for a user with a historical ID to fix their user ID,
+however it is envisioned that account portability will exist around the time of this MSC's
+acceptance to provide an escape route. This is discussed in further detail in the "Potential Issues"
+section.
+
+## Proposal
+
+In a future room version, the [non-historical grammar](https://spec.matrix.org/v1.7/appendices/#user-identifiers)
+for user IDs is strictly enforced on all places a user ID can appear in an event. For example, the
+`sender` of all events and `state_key` of `m.room.member` MUST match the grammar. All user IDs which
+use the historical grammar, by any definition, are considered non-compliant.
+
+By making this change in a room version, non-compliant user IDs are slowly removed from the public
+federation. Several rooms will naturally upgrade to a room version which includes this MSC's change
+after it becomes available in servers, and eventually the specification will update the
+[recommended default room version](https://spec.matrix.org/v1.7/rooms/#complete-list-of-room-versions)
+to include this MSC's change as well to affect new room creation.
+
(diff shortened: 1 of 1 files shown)


proposals/4044-strict-user-id-grammar-in-rooms.md
```diff
@@ -0,0 +1,88 @@
...
+Account portability in Matrix is, as of writing, progressing under the pseudo IDs project. Prior
+art for such MSCs include:
+
+* [MSC1228: Removing MXIDs from events](https://github.com/matrix-org/matrix-spec-proposals/pull/1228)
+* [MSC2787: Portable Identities](https://github.com/matrix-org/matrix-spec-proposals/pull/2787)
+* [MSC3839: Primary-identity-as-key](https://github.com/matrix-org/matrix-spec-proposals/pull/3839)
+* [MSC4014: Pseudonymous Identities](https://github.com/matrix-org/matrix-spec-proposals/pull/4014)
+
+The general idea is that by breaking the association of user IDs with events, it becomes possible for
+the user ID to change without affecting the events themselves. In the interim, MSCs like this one
```
MTRNord, line 67: While I think I understand the point of view you are coming from, I believe this MSC is impossible to be used in any "default" room version until this migration path is existing, tested and consists of working UX. (To avoid another room upgrade UX nightmare in clients). Since we likely have thousands of people with historical IDs still. So a migration path should be a hard blocker for the MSC and not a "we fix that anyway with portable identities in the future" :)

proposals/4044-strict-user-id-grammar-in-rooms.md
```diff
@@ -0,0 +1,88 @@
...
+Matrix does not currently specify a way for a user with a historical ID to fix their user ID,
+however it is envisioned that account portability will exist around the time of this MSC's
+acceptance to provide an escape route. This is discussed in further detail in the "Potential Issues"
+section.
+
+## Proposal
+
+In a future room version, the [non-historical grammar](https://spec.matrix.org/v1.7/appendices/#user-identifiers)
+for user IDs is strictly enforced on all places a user ID can appear in an event. For example, the
+`sender` of all events and `state_key` of `m.room.member` MUST match the grammar. All user IDs which
```
clokep, line 22: Can we add the appropriate keys of power levels into this too?

olmari:
I otherhand am sad to see usernames would be forced to use such "7-bit ASCII" character set... I have no strong preferences for "shoulds and coulds and recommendations", but "must"ing to have such charset is so ancient now there does exist whole unicode system...

If specifically an abuse of using similar/same looking letters is specifically needed to be taken into consideration, then some normalization similar to ZFS formD for example could be very acceptable solution specifically for that aspect.

proposals/4044-strict-user-id-grammar-in-rooms.md
```diff
@@ -0,0 +1,88 @@
...
+* [MSC2787: Portable Identities](https://github.com/matrix-org/matrix-spec-proposals/pull/2787)
+* [MSC3839: Primary-identity-as-key](https://github.com/matrix-org/matrix-spec-proposals/pull/3839)
+* [MSC4014: Pseudonymous Identities](https://github.com/matrix-org/matrix-spec-proposals/pull/4014)
+
+The general idea is that by breaking the association of user IDs with events, it becomes possible for
+the user ID to change without affecting the events themselves. In the interim, MSCs like this one
+help limit the ability for non-compliant IDs to spread (and provide a useful way for association-breaking
+MSCs to only support compliant IDs).
+
+## Alternatives
```
turt2live, line 71: @olmari says:

> I otherhand am sad to see usernames would be forced to use such "7-bit ASCII" character set... I have no strong preferences for "shoulds and coulds and recommendations", but "must"ing to have such charset is so ancient now there does exist whole unicode system...
> 
> If specifically an abuse of using similar/same looking letters is specifically needed to be taken into consideration, then some normalization similar to ZFS formD for example could be very acceptable solution specifically for that aspect.
> 
> EDIT: I guess this thought-train is kind of late for the actual grammar set of choice, which my primary rant is about... Having "allow normalized unicode" and enforcing that would not change this MSC and in general would be okay with caveats @MTRNord resolved...

PR #4045, opened 2023-08-12
MSC4045: Deprecating the use of IP addresses in server names

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/drop-ips/proposals/4045-deprecate-server-ip-only-discovery.md)

No unstable implementations known.

### DIFF
--- a/proposals/4045-deprecate-server-ip-only-discovery.md
+++ b/proposals/4045-deprecate-server-ip-only-discovery.md
@@ -1,0 +1,72 @@
+# MSC4045: Deprecating the use of IP addresses in server names
+
+A server in Matrix must be named according to the [existing grammar](https://spec.matrix.org/v1.7/appendices/#server-name),
+which includes support for bare IPv4 and IPv6 addresses. These bare IP addresses are then further
+handled by the [server discovery](https://spec.matrix.org/v1.7/server-server-api/#server-discovery)
+algorithm when trying to make contact with the target server. However, a challenge arises due to
+the specification's [requirement to use TLS](https://spec.matrix.org/v1.7/server-server-api/#tls):
+getting a valid or useful certificate from a known Certificate Authority is effectively impossible.
+
+If a server *does* somehow get a certificate for an IP address, the operator of that server can *never*
+lose that IP address if they plan to continue communicating with the rest of the federated network.
+Worse, the operator can intentionally recycle the IP address as a [DoS vector](https://github.com/matrix-org/matrix-spec/issues/386).
+
+Further, [server ACLs](https://spec.matrix.org/v1.7/client-server-api/#server-access-control-lists-acls-for-rooms)
+already have a capability to ban bare IP address server names in rooms, and the vast majority of
+known public rooms using server ACLs already do so (largely thanks to tools like [Mjolnir](https://github.com/matrix-org/mjolnir)
+which non-optionally set the `allow_ip_literals` flag to `false`).
+
+This proposal deprecates the use of bare IP addresses, and entirely bans them from a future room
+version, preventing/discouraging further IP-only server names from appearing in the public federation.
+
+## Proposal
+
+In a future room version, the [grammar](https://spec.matrix.org/v1.7/appendices/#server-name) for
+a server name is adjusted to be:
+
+```
+server_name = hostname [ ":" port ]
+port        = 1*5DIGIT
+hostname    = 1*255dns-char
+dns-char    = DIGIT / ALPHA / "-" / "."
+```
+
+The special considerations for IPv4 and IPv6 addresses are removed.
+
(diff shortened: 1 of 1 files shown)


PHJArea217:
I guess this could be useful to allow Matrix servers to run under NAT64/DNS64 (since that breaks IP literals)

Also, there might be a way to still use IP literals if they use something like sslip.io.

I'm new to the Matrix spec change process, please be patient with me.

proposals/4045-deprecate-server-ip-only-discovery.md
```diff
@@ -0,0 +1,72 @@
+# MSC4045: Deprecating the use of IP addresses in server names
+
+A server in Matrix must be named according to the [existing grammar](https://spec.matrix.org/v1.7/appendices/#server-name),
+which includes support for bare IPv4 and IPv6 addresses. These bare IP addresses are then further
+handled by the [server discovery](https://spec.matrix.org/v1.7/server-server-api/#server-discovery)
+algorithm when trying to make contact with the target server. However, a challenge arises due to
+the specification's [requirement to use TLS](https://spec.matrix.org/v1.7/server-server-api/#tls):
+getting a valid or useful certificate from a known Certificate Authority is effectively impossible.
```
Kladki, line 8: This won't be the case in the near future: https://letsencrypt.org/2025/01/16/6-day-and-ip-certs/

turt2live: Honestly, I feel as though that may just further justify this MSC's existence 😅 if it's easier to get certificates for IP addresses like that, then it's easier to create throwaway servers too.

Most rooms which use a moderation bot of some kind additionally ban IP address hostnames via server ACLs, but that hardly accounts for rooms which haven't discovered the need for a moderation bot yet. 

I'm inclined to cite the blog post as a reason to land this MSC, and push it forward to FCP. Thoughts are welcome on the general direction of this MSC.

Kladki: I don't think I've ever seen a matrix user connecting from an IP address (likely due to the reasons mentioned in the MSC), so I don't really see any real benefit to banning raw IP addresses from participating. I wouldn't be surprised if this was the case for most people in the matrix federation, so maybe you should try to explain more what the benefits are to doing this.

CobaltCause: > I don't really see any real benefit to banning raw IP addresses from participating

One benefit would be eliminating parsing ambiguities: https://github.com/matrix-org/matrix-spec/issues/2090

Alternatively, what's the point in permitting it if it never happens in practice?

Kladki: > Alternatively, what's the point in permitting it if it never happens in practice?

Careful, just because it's not commonplace in most of the federation, doesn't rule out that there could be small groups that may find this very useful in other parts of the world once certs for IP addresses become readily available.

And of course, in rooms where IP addresses are banned (as is the case for all of matrix.org's rooms afaik), I'm not surprised that you don't see IP addresses connecting. :laughing:  (plus ofc that it's very difficult to get a valid cert for an IP address atm)

Saiv46:
I oppose this proposal for single reason: private networks.

Pinecone, Yggdrasil, ZeroTier and other solutions use own end-to-end encryption and IP addressation. And that doesn't include air-gapped networks with TLS termination on-edge, where getting valid TLS inside the network is problematic.

proposals/4045-deprecate-server-ip-only-discovery.md
1vel, line 8: Let's Encrypt now issues certs for IP addresses.

proposals/4045-deprecate-server-ip-only-discovery.md
```diff
@@ -0,0 +1,72 @@
...
+
+If a server *does* somehow get a certificate for an IP address, the operator of that server can *never*
+lose that IP address if they plan to continue communicating with the rest of the federated network.
+Worse, the operator can intentionally recycle the IP address as a [DoS vector](https://github.com/matrix-org/matrix-spec/issues/386).
```
1vel, line 12: Hundreds of FreeDNS providers exist, aren't going away any time soon, and have been used as DoS vectors in the wild. Additionally, this is the same case with domains; under this logic, regular domains should also be deprecated.

proposals/4045-deprecate-server-ip-only-discovery.md
```diff
@@ -0,0 +1,72 @@
...
+
+Further, [server ACLs](https://spec.matrix.org/v1.7/client-server-api/#server-access-control-lists-acls-for-rooms)
+already have a capability to ban bare IP address server names in rooms, and the vast majority of
```
1vel, line 15: So why not keep it that way? This seems like an unnecessary limitation, and is also only considering the public federation.

PR #4047, opened 2023-08-23
MSC4047: Send Keys

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/room-send-key/proposals/4047-room-send-keys.md)

Dependencies:
* [MSC4046](https://github.com/matrix-org/matrix-spec-proposals/pull/4046): Make & send PDU endpoints

### DIFF
--- a/proposals/4047-room-send-keys.md
+++ b/proposals/4047-room-send-keys.md
@@ -1,0 +1,310 @@
+# MSC4047: Send Keys
+
+A common feature request is a room structure where most/all members are not aware of each other. Matrix
+currently requires all senders to be aware of each other for routing purposes, making it difficult to hide
+a user's subscription to the room.
+
+The core principle of this problem is tracked as [issue #367](https://github.com/matrix-org/matrix-spec/issues/367), though the exact semantics described by that issue are different from this proposal. Instead of limiting
+membership visibility by power level, this MSC aims to introduce an ability for "external senders", with
+an assumption that [MSC2753-style](https://github.com/matrix-org/matrix-spec-proposals/pull/2753) peeking
+or other similar mechanism can be used to subscribe to the room without becoming a member.
+
+External senders are accomplished with a "send key" in this proposal. A public key is published to the room
+for event auth concerns, and a private key is sent to the desired senders. Anyone with that private key can
+then send events into the room.
+
+Dependencies:
+* [MSC4046](https://github.com/matrix-org/matrix-spec-proposals/pull/4046)
+
+## Proposal
+
+*This MSC is done entirely in context of a future room version, due to event auth and redaction algorithm
+changes.*
+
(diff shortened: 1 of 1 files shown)


proposals/4047-room-send-keys.md
```diff
@@ -0,0 +1,310 @@
...
+**Change 2**: Ensure events referencing old/regenerated send keys are soft-failed.
+
+Check 6 of "checks performed upon receipt of a PDU" is clarified as such:
+
+> With respect to Check 6, events using "send key" signatures must have a valid signature when using the current
+> `m.room.send_key` state event in the room, otherwise it is "soft failed".
+>
+> For example, if the current `m.room.send_key` event ID is `$current`, and the following signature is present:
+> ```json
+> {
```
turt2live, line 205: oh no, what have you done to my MSC, github.

proposals/4047-room-send-keys.md
```diff
@@ -0,0 +1,315 @@
...
+> 4. If `type` is `m.room.send_key`:
+>    1. If the event uses "send keys" (see Rules 2.5 and 2.6), reject.
+
+## Potential issues
+
+The potential issues with this MSC are primarily security considerations.
+
+## Alternatives
+
+Pseudo IDs of some variety, or senders-as-keys, might be an easier/different way to solve this MSC's
```
ara4n, line 252: I'm pretty concerned that this stack of MSCs collides almost directly with MSC #4014 and #4080. while aiming solve slightly different problems, the changes hit in almost precisely the same place. Once we finally get back to senders-as-keys I think we should reconcile the two together rather than having competing stacks of MSCs.

turt2live: This MSC is not meant to be conflicting, but indeed all the MSCs involved will have to consider each other's use cases before moving forward.

proposals/4047-room-send-keys.md
```diff
@@ -0,0 +1,315 @@
+# MSC4047: Send Keys
+
+A common feature request is a room structure where most/all members are not aware of each other. Matrix
+currently requires all senders to be aware of each other for routing purposes, making it difficult to hide
+a user's subscription to the room.
```
andrewzhurov, line 5: The direction of thought is interesting here.
Will I make stretch by extending it to "have rooms as topics _anyone_ can chat about"?
Akin to adding tags to messages.
Interestingly, this unlocks one to define what he wants to see about a topic. E.g., events about "politics" (tag) from my friends.
These interest can be expressed as events.
And there can be one personalized MEG per person, including all the event you care about, that been shared to you by others cordially (since they know you need them). Wondered about it [here](https://github.com/matrix-org/matrix-spec-proposals/pull/1769).

PR #2472, opened 2020-03-24
MSC2472: Symmetric SSSS

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/symmetric_ssss/proposals/2472-symmetric-ssss.md)

PR #4048, opened 2023-08-23
MSCxxxx: Signed key backup

uhoreg:


proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
+# MSC4048: Signed key backup
+
+The [server-side key
+backups](https://spec.matrix.org/unstable/client-server-api/#server-side-key-backups)
+allows clients to store event decryption keys so that when the user logs in to
+a new device, they can decrypt old messages.  The current algorithm encrypts
+the event keys using a symmetric algorithm, allowing clients to upload keys to
```
AndrewFerr, line 7: Doesn't the current key backup algorithm use asymmetric encryption?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
...
+the backup without necessarily giving them the ability to read from the
+backup.  For example, this allows for a partially-trusted client to be able to
+read (and save the keys for) current messages, but not read old messages.
+
+However, since the event decryption keys are encrypted using a symmetric
```
AndrewFerr, line 12: ditto

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
...
+algorithm, this allows anyone who knows the public key to write to the backup.
+As a result, keys loaded from the backup must be marked as unauthenticated,
+leading to [usability
+issues](https://github.com/vector-im/element-web/issues/14323).
+
+[MSC3270](https://github.com/matrix-org/matrix-spec-proposals/pull/3270) tries
+to fix this issue by using a symmetric, authenticated encryption algorithm,
```
AndrewFerr, line 19: This one really is symmetric (as it's the whole point of that MSC)

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
...
+which ensures that only someone who knows the secret key can write to the
+backup.  However this removes the ability for a client to be able to write to
+the backup without being able to read from it.
+
+We propose to continue using a symmetric encryption algorithm in the backup,
```
AndrewFerr, line 24: Continue using asymmetric as we do currently, right?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
...
+
+Being able to prove authenticity of keys may affect the deniability of
+messages: if a user has a Megolm session in their key backup that is signed by their
+backup signing key, and the session data indicates that it originated from one
+of their devices, this could be used as evidence that the Megolm session did in
+fact come from them.
+
+This is somewhat mitigated by the fact that obtaining the Megolm session
+requires the decryption key for the backup.  In addition, the deniability
+property is mainly refers to the fact that a recipient cannot prove the
```
AndrewFerr, line 92:
```suggestion
property mainly refers to the fact that a recipient cannot prove the
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,108 @@
...
+A user has a new signing key, referred to as the "backup signing key", used to
+sign key backups using the ed25519 signature algorithm.  The private key can be
+shared/stored using [the Secrets
+module](https://spec.matrix.org/unstable/client-server-api/#secrets) using the
+name `m.key_backup.signing`.
+
+The `AuthData` object for the [`m.megolm_backup.v1.curve25519-aes-sha2` key
+backup
+algorithm](https://spec.matrix.org/unstable/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2)
+has a new optional property called `signing_public_key`, contains the public
```
AndrewFerr, line 38: What's the rationale behind using a new key for backup signing instead of reusing the master signing key or device key?  Is it to make it possible to revoke the backup signing key without affecting any other keys?

AndrewFerr (changes requested):
This is a sensible idea.  Just have a few questions.

proposals/4048-signed-key-backup.md
BillCarsonFr: In several MSC we are seeing introduction of new keys (tofu, constraint membership), looks like we are going to soon have tons of such keys. It's mainly annoying because adding a new key requires to update the exisiting 4S in order to update the account recovery backup, and this will require users to enter their passphrase on upgrade.

Could we try to go with scoped signatures?

As per:

Given a scope named `my-scope`, a signature of a content would mean to take the canonical json of content then append `my-scope` to that string then sign it with an existing key (e.g MSK).

scoped_signature = sign(canonical_json_content + "my-scope", alice_msk).

And the signature should be uploaded with the ed25519.<scope> scheme:

"signatures": {
    "@alice:example.com": {
        "ed25519.myscope:base64+master+signing+public+key": "base64+scoped+signature+of+content"
    }
},

The scope is added in the string to sign so that the home server cannot switch signature from another scope (or no scope) of the same content.

In this case I suppose that using the scoped signature by the SSK makes sense

alice: Hi there! GitHub likes to notify me whenever someone uses my name with an @ in front of it as an example. I'll untag myself from this, but just a heads up :)

proposals/4048-signed-key-backup.md
uhoreg:
```suggestion
However, since the event decryption keys are encrypted using an asymmetric
```

proposals/4048-signed-key-backup.md
uhoreg:
```suggestion
We propose to continue using an asymmetric encryption algorithm in the backup,
```

proposals/4048-signed-key-backup.md
uhoreg: I've added a comment in the "Alternatives" section

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,116 @@
...
+for the key backup.  However, this is not backwards-compatible, and does not
+allow for clients that can write to the backup without reading.
+
+Rather than using a new signing key, we could use an existing signing key, such
+as one of the cross-signing keys.  This would remove the need for users to
+enter their Secret Storage key to add the new signing key.  However, this means
+that a user cannot create a key backup without also using cross-signing.  Using
+a separate key also allows the user to give someone else (such as a bot)
+permission to write to their backups without allowing them to perform any
+cross-signing operations.
```
AndrewFerr, line 88: Would it possible to use a device key as the backup signing key?  Unlike cross-signing keys, the device key is guaranteed to exist (since a device that cares about E2EE-related backups should also have prepared device keys for E2EE).

A blocker is that the current description of backup signing says that all backups are to be signed with the same key, thus requiring the private part of the signing key to be shared among all of an account's devices.  Device keys can't meet that requirement since their private part must remain private to its device.

However, this could be mitigated by allowing _multiple_ acceptable backup signing keys, where any of those keys could sign a backup (or even several could sign it, since Signed JSON allows multiple signatures on the same object), and a device knows & trusts the public part of the key with which to verify the backup's signature (which is / can be the case for device keys).

Also, this proposal still allows the `AuthData` object to be signed by a device key; is there anything preventing that from being done for the `SessionData` as well?

BillCarsonFr: The backup will outlive many devices. If you sign with a device key, you'll be unable to check the validity after the device has been deleted.

> requiring the private part of the signing key to be shared among all of an account's device

This is possible thanks to [4S](https://github.com/matrix-org/matrix-spec-proposals/pull/2472) that allows to share local secrets among the user trusted devices.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,116 @@
...
+name `m.key_backup.signing`.
+
+The `AuthData` object for the [`m.megolm_backup.v1.curve25519-aes-sha2` key
+backup
+algorithm](https://spec.matrix.org/unstable/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2)
+has a new optional property called `signing_public_key`, contains the public
+key of the backup signing key, encoded in unpadded base64.  If the `AuthData`
+is not signed by the user's master signing key or by a verified device
+belonging to the same user, the backup signing key must be ignored, and all
+keys in the backup must be treated as being unsigned.
```
davidegirardi, line 42: So verifying a backup requires to:
- Get the private key from the Secrets module
- Get the public key from `AuthData`
- Verify that the public key corresponds to the private key
- Verify the signature of `AuthData`

Correct?

I want to make sure an administrator of a homeserver cannot forge backups.

AndrewFerr: AIUI only _writing_ a backup requires reading from Secrets, since verification uses the public part of the signing key (which is present in `AuthData`), and verifying that the public & private keys correspond to each other is just a matter of verifying a backup's signature (which needs only the public key).

davidegirardi: I'm thinking about this scenario: a malicious admin could create a new backup with the corresponding `AuthData` signed with _their_ private key and containing its corresponding public key but encrypt the backups with the user's public key.

The client can verify the signature of `AuthData` and use the private key it already has to decrypt the forged backup.

dkasak: >  a malicious admin could create a new backup with the corresponding AuthData signed with their private key and containing its corresponding public key but encrypt the backups with the user's public key.

Note that there's actually three keys in the quoted text: the private and public parts of the key pair used to sign `AuthData` and the public key contained within the `AuthData` in a field called `signing_public_key`.

The key pair `AuthData` should be signed with is the user's MSK (or the device's device key, of the device which is trying to access the backup). These keys need to be validated via another mechanism, for instance via verification, so that a malicious admin cannot simply inject an `AuthData` signed with their own key pair.

The actual `SessionData` of the backup then needs to be signed with the `signing_public_key` contained within such a validated `AuthData` in order to be considered authenticated. This is required by the text of the MSC:

> If the `SessionData` is not signed by the backup signing key, then this flag must be treated as being `false`.

Do you still see a potential hole?

AndrewFerr: Agreed that device verification protects against this attack:

- For a malicious admin to forge backups, it has to make other clients trust its backup signature
- For a backup signature to be trusted, its `signing_public_key` must come from a trusted `AuthData`
- For an `AuthData` to be trusted, it must be signed by the account's MSK or the device key of a verified device

Thus, what the forgery needs is a way to get clients to trust the key it uses for signing its `AuthData`.  Aside from breaking device verification, a way to achieve that would be to steal the private part of the account's MSK & use it for signing, but that would require breaking the Secrets module: either by breaking the encryption of secrets stored on the server (in this case, the `m.cross_signing.master` secret), or by tricking a legitimate client into sharing secrets (in this case, the private MSK) with a malicious client.

On that note, using secret storage/sharing is another way an `AuthData` can be trusted (by storing/sharing the `public_key`'s corresponding private key), so a forgery would succeed if it could manage to inject secrets into a target account, or share secrets with a target client.  Achieving either of those tasks requires breaking components unrelated to backup signatures (and I believe much of it hinges on device verification too).

> a malicious admin could create a new backup with the corresponding `AuthData` ... containing its corresponding public key but encrypt the backups with the user's public key.

If the public key being referred to here is the backup encryption key, I don't think it impacts the forgery attack at all, because AIUI that key is never used for verification purposes.

AndrewFerr (dismissed):

proposals/4048-signed-key-backup.md
davidegirardi: Makes sense, so:
> Verify that the public key corresponds to the private key

Is correct, modulo that I didn't write anything about the device key.

proposals/4048-signed-key-backup.md
dkasak, line 88: @poljar, @davidegirardi, and myself discussed another alternative that avoids adding a new cryptographically-independent key while also avoiding all of the downsides mentioned above.

We propose deriving the backup signing key from the existing backup encryption key using a KDF. This retains the ability to have write-only bots, by sharing the derived signing key without sharing the encryption key. It also avoids having to set up cross-signing.

The derivation step could look something like this:

```
signing_key = HKDF("", encryption_key, "BACKUP_SIGNING_KEY", 32)
```

poljar: I think you meant:
`signing_key = HKDF("", decryption_key, "BACKUP_SIGNING_KEY", 32)`

Where the `decryption_key` is the [Backup recovery key] before it was clamped into a `Curve25519` key.

[Backup recovery key]: https://spec.matrix.org/v1.8/client-server-api/#recovery-key

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,116 @@
...
+
+The `SessionData` object for the `m.megolm_backup.v1.curve25519-aes-sha2` key
+backup algorithm has two new optional properties:
+
+- a `signatures` property: the `SessionData` is a [signed JSON
+  object](https://spec.matrix.org/unstable/appendices/#signing-json), signed
+  using the backup signing key, using the public key (encoded in unpadded
+  base64) as the key ID
```
dkasak, line 50: Given that `SessionData` is the plaintext, it sounds like we would have to decrypt the backup in order to find out whether the signature is correct. Given that we are also removing the outer MAC, that would leave us operating on unauthenticated cryptographic data, which is a no-no.

As an aside, I think this MSC would benefit from a "Mechanics" section describing concrete steps in a "You take THIS to perform THAT" kind of manner rather than leaving it implicit from the description of the components.

proposals/4048-signed-key-backup.md
dkasak: And then there's no reason to even keep using *signatures* as the auth method, since we could treat the derived key as a symmetric one and use the more efficient HMAC instead.

AndrewFerr: > we could treat the derived key as a symmetric one and use the more efficient HMAC instead.

Do you mean to reinstate the outer MAC, or to use the signing key to produce a MAC?

dkasak: > Do you mean to reinstate the outer MAC, or to use the signing key to produce a MAC?

The MAC *has* to be on the outside so that we don't even touch the data before we authenticate it. The "signing" key is then no longer a *signing* key, but a symmetric key used for HMAC:

    hmac_key = HKDF("", decryption_key, "BACKUP_AUTH_KEY", 32)

(The mechanics of it are the same, but the intention is different.)

If we'd like to avoid changing the backup algorithm version, we could always name the property differently to avoid collision with the old MAC property. But given it's likely we'll need a new backup algorithm version for other reasons, I'm not sure it makes sense to try to avoid it.

> I think you meant:
> 
> ```
> signing_key = HKDF("", decryption_key, "BACKUP_SIGNING_KEY", 32)
> ```
> 
> Where the `decryption_key` is the [Backup recovery key](https://spec.matrix.org/v1.8/client-server-api/#recovery-key) before it was clamped into a `Curve25519` key.

Right, I did. I mixed up the specifics of that but that's indeed the right solution. I'll edit.

uhoreg: It looks like deriving the signing/MAC key from the decryption key should be secure enough -- I can't think of a reason why someone who is allowed to decrypt the backup shouldn't also be allowed to write to the backup.  AFAICT, this will still require us to put the signing/MAC key in SSSS (so that devices that can't read can still write), but it allows getting both the decryption and signing/MAC key from a single key.

I had considered using a MAC instead of a signature, but had initially dismissed it because we already have a format defined for signed JSON, but not for MAC'ed JSON, and I didn't really see any major advantage (aside from efficiency) to using a MAC over a signature.  But if we encrypt-then-MAC, then we don't need to worry about a MAC'ed JSON because we just MAC the ciphertext.

I guess one advantage of using a MAC and this key derivation for the MAC key is that we don't need to change the auth_data for the backup, which means we don't need to create a new backup version (unless we do want to change the algorithm name).

uhoreg: I've pushed a new version of the MSC that uses a MAC and derives the key from the decryption key.  Let me know if I've mangled anything.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,155 @@
...
+
+We propose to continue using an asymmetric encryption algorithm in the backup,
+but to ensure authenticity by producing a MAC using a key derived from the
+backup's decryption key.
+
+## Proposal
+
+A user who has a key backup derives a new backup MAC key by performing HKDF on
+the backup decryption key (as raw unencoded bytes) with no salt and an info
+parameter if `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
```
dkasak, line 32:
```suggestion
parameter of `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
```

proposals/4048-signed-key-backup.md
dkasak: Just to add, the primary advantage here is that the auth mechanism should protect the outer ciphertext rather than the plaintext, to prevent chosen-ciphertext attacks. It being a MAC rather than a signature is just a more elegant approach and a better-fitting tool for the job, but both would work.

Will give it another read.

dkasak: > AFAICT, this will still require us to put the signing/MAC key in SSSS (so that devices that can't read can still write), but it allows getting both the decryption and signing/MAC key from a single key.

How would we give access to just the MAC key in the SSSS to such devices?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,155 @@
...
+not have a `mac2` property (see below), then this flag must be treated as being
+`false`.
+
+No changes are made to the `AuthData` for the
+`m.megolm_backup.v1.curve25519-aes-sha2` key backup algorithm.
+
+The following changes are made to the cleartext `session_data` property of the
+`KeyBackupData` object is deprecated:
+
+- a new `mac2` [FIXME: get a better name.  suggestions?] property is added,
```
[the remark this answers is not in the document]
dkasak: Note that we may have to change the backup algorithm name for other reasons. If we're doing that, we might get away with reusing the MAC field name.

proposals/4048-signed-key-backup.md
uhoreg: > How would we give access to just the MAC key in the SSSS to such devices?

In theory, Secret Storage can encrypt different sets of secrets with different keys, though in practice all clients just use the default key.

However, if the secrets are being shared from one device to another, it can share the MAC key without sharing the backup decryption key.

Hmm.  Maybe we don't have to store the MAC key, but just share it?

uhoreg: Actually, I think it's pretty simple.  If the Secret Storage key has encrypted the backup key, then we don't need to encrypt the MAC key, as we can just derive it.  If the Secret Storage key has not encrypted the backup key, but we want it to encrypt the MAC key, then we can store it using the name `m.key_backup.mac`.  The MAC key can also be shared using the same name.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Proposal
+
+A user who has a key backup derives a new backup MAC key by performing HKDF on
+the backup decryption key (as raw unencoded bytes) with no salt and an info
+parameter if `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
+
+    mac_key = HKDF("", decryption_key, "MATRIX_BACKUP_MAC_KEY", 32)
+
+The signing key can be shared/stored using [the Secrets
```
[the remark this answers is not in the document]
uhoreg:
```suggestion
The backup MAC key can be shared/stored using [the Secrets
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,155 @@
...
+but to ensure authenticity by producing a MAC using a key derived from the
+backup's decryption key.
+
+## Proposal
+
+A user who has a key backup derives a new backup MAC key by performing HKDF on
+the backup decryption key (as raw unencoded bytes) with no salt and an info
+parameter if `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
+
+    mac_key = HKDF("", decryption_key, "MATRIX_BACKUP_MAC_KEY", 32)
```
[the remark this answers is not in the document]
uhoreg:
```suggestion
    backup_mac_key = HKDF("", decryption_key, "MATRIX_BACKUP_MAC_KEY", 32)
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+
+A user who has a key backup derives a new backup MAC key by performing HKDF on
+the backup decryption key (as raw unencoded bytes) with no salt and an info
+parameter of `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
+
+    backup_mac_key = HKDF("", decryption_key, "MATRIX_BACKUP_MAC_KEY", 32)
+
+The backup MAC key can be shared/stored using [the Secrets
+module](https://spec.matrix.org/unstable/client-server-api/#secrets) using the
+name `m.megolm_backup.v1.mac`.  Note that if the backup decryption key (the
```
BillCarsonFr, line 38: Can we have a bit more detail on that, and maybe the use case?
Something like: 
It is possible in the protocol to only make part of the secrets available to certain devices. With that mecanism it's possible to have a session that is not authorised to access history (`m.megolm_backup.v1` is not shared), but that yet remains able to participate in the backup for new keys (needs then only the `m.megolm_backup.v1.mac`). It would also allow clients to not cache the backup decryption key, but still participate in the backup.

WDYT?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+c) the key was downloaded from the key backup, with the `authenticated`
+property set to `true` and was authenticated (for example using the method from
+this proposal).  If the `session_data` does not have a `mac2` property (see
+below), then this flag must be treated as being `false`.
+
+No changes are made to the `AuthData` for the
+`m.megolm_backup.v1.curve25519-aes-sha2` key backup algorithm.
+
+The following changes are made to the cleartext `session_data` property of the
+`KeyBackupData` object is deprecated:
```
BillCarsonFr, line 61: Notice that the server side is using some metadata of `KeyBackupData` to decide what to do when the same key is uploaded: keep or replace existing one.
The new meta data added to the `session_data` should be considered by this algorithm.

proposals/4048-signed-key-backup.md
dkasak: I would argue we never want to store `m.key_backup.mac` in SSSS, to simplify reasoning. Clients which have access to the SSSS should be expected to have full (write *and* read) access to the key backup anyway.

*Sharing* `m.key_backup.mac` using the Secret Sharing API makes sense though because it gives us an opportunity to support write-only clients.

> In theory, Secret Storage can encrypt different sets of secrets with different keys, though in practice all clients just use the default key.

I would avoid introducing this, so that we don't needlessly increase complexity and make everything harder to reason about. I don't think we even have any kind of key ID there, so a client would have to try decrypting with each SSSS key it has in succession?

proposals/4048-signed-key-backup.md
dkasak: See also https://github.com/matrix-org/matrix-spec-proposals/pull/4048#discussion_r1341236461 for some more nuance on this.

proposals/4048-signed-key-backup.md
dkasak, line 61: This sentence doesn't make much syntactic sense due to the "is deprecated" at the end. Editing error?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+  should also accept `session_data` that does not have the `mac` property if
+  the `mac2` property is present, as the `mac` property may become optional in
+  the future.
+
+The [construction of the `session_data`
+property](https://spec.matrix.org/unstable/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2)
+thus becomes:
+
+1. Encode the session key to be backed up as a JSON object using the
+   SessionData.
```
dkasak, line 78:
```suggestion
   `SessionData`.
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+2. Generate an ephemeral curve25519 key, and perform an ECDH with the ephemeral
```
dkasak, line 79:
```suggestion
2. Generate an ephemeral Curve25519 key, and perform an ECDH with the ephemeral
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+- a MAC is faster to calculate.  The main advantage of a signature is that it
+  allows one to verify the signature without knowing the private key, but in
+  this case, reading is a more privileged action than writing, and writers
+  already need to know the private/secret key.
+- since the MAC key is derived from the decryption key, two clients can be
+  upgraded at the same time without interfering with each other, as they will
+  derive the same MAC key
+- the MAC is calculated after encryption, and hence is verified before
+  decryption, so we know that it is authenticated before we do any processing
+  on it
```
dkasak, line 140:
```suggestion
- No changes to `AuthData` are necessary, so a new backup version is not
  required.
- A MAC is faster to calculate.  The main advantage of a signature is that it
  allows one to verify the signature without knowing the private key, but in
  this case, reading is a more privileged action than writing, and writers
  already need to know the private/secret key.
- Since the MAC key is derived from the decryption key, two clients can be
  upgraded at the same time without interfering with each other, as they will
  derive the same MAC key.
- The MAC is calculated after encryption, and hence is verified before
  decryption, so we know that it is authenticated before we do any processing
  on it.
```

proposals/4048-signed-key-backup.md
uhoreg: Yeah, you may be right about simplifying reasoning.  I'll leave it in for now, but will probably remove it later.

proposals/4048-signed-key-backup.md
uhoreg: Yes, looks like editing error

```suggestion
The following changes are made to the cleartext `session_data` property of the
`KeyBackupData` object:
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+below), then this flag must be treated as being `false`.
+
+No changes are made to the `AuthData` for the
+`m.megolm_backup.v1.curve25519-aes-sha2` key backup algorithm.
+
+The following changes are made to the cleartext `session_data` property of the
+`KeyBackupData` object:
+
+- a new `mac2` [FIXME: get a better name.  suggestions?] property is added,
+  which is a MAC of the `SessionData` ciphertext (priory to base64-encoding),
```
uhoreg, line 64:
```suggestion
  which is a MAC of the `SessionData` ciphertext (prior to base64-encoding),
```

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+below), then this flag must be treated as being `false`.
+
+No changes are made to the `AuthData` for the
+`m.megolm_backup.v1.curve25519-aes-sha2` key backup algorithm.
+
+The following changes are made to the cleartext `session_data` property of the
+`KeyBackupData` object:
+
+- a new `mac2` [FIXME: get a better name.  suggestions?] property is added,
+  which is a MAC of the `SessionData` ciphertext (prior to base64-encoding),
```
uhoreg, line 64: Should the MAC cover the `ephemeral` property too?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+messages: if a user has a Megolm session in their key backup that is MAC'ed by
+their backup MAC key, and the session data indicates that it originated from
+one of their devices, this could be used as evidence that the Megolm session
+did in fact come from them.
+
+This is somewhat mitigated by the fact that obtaining the Megolm session
+requires the decryption key for the backup.  In addition, the deniability
+property mainly refers to the fact that a recipient cannot prove the
+authenticity of the message to a third party, and usually is not concerned with
+preventing self-incrimination.  And in fact, a confiscated device may already
```
ara4n, line 160: Hmm. Just because I created a megolm session doesn’t mean that I was the one who encrypted the messages in it, as megolm is symmetric? So proving I own the creation of a key doesn’t achieve much in terms of deniability aiui; a given message could have been fabricated by the other party? (at least until you try to send a msg with the same ratchet key - but i guess the same would be true if the megolm session was entirely fabricated, in terms of happening at the wrong place relative to other megolm sessions)

uhoreg: The Megolm session has a signing key that only the creator knows the private part.  So while anyone can encrypt a message with the Megolm session, they won't be able to produce a correct signature, so the message won't be validated.

proposals/4048-signed-key-backup.md
dkasak: I don't see why we would want to leave it out.

uhoreg: Yeah.  I'm not sure that it's strictly necessary, but not much reason no to, and better to do so just in case.  I think the only reasons for leaving it out are: the original MAC left it out, and if we include it, we need to decide on a format.  But for the format, I think we can do the same thing as Signed JSON.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,177 @@
...
+parameter of `"MATRIX_BACKUP_MAC_KEY"` and generating 32 bytes (256 bits):
+
+    backup_mac_key = HKDF("", decryption_key, "MATRIX_BACKUP_MAC_KEY", 32)
+
+The backup MAC key can be shared/stored using [the Secrets
+module](https://spec.matrix.org/unstable/client-server-api/#secrets) using the
+name `m.megolm_backup.v1.mac`.  Note that if the backup decryption key (the
+secret using the name `m.megolm_backup.v1`) is stored/shared, then the backup
+MAC key does not need to be stored/shared as it can be derived from the backup
+decryption key.
```
davidegirardi, line 41: If we can derive the the MAC key in the first place, it means that the backup key is stored or input by the user. We already have the key in other words.
I think we can avoid a branch here if we just derive the MAC key in any case.

uhoreg: This is for situations in which a user wants a client to have the MAC key but not the decryption key.  So they can share the MAC key from another client that does have it, but without sharing the decryption key.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,275 @@
...
+algorithm, this allows anyone who knows the public key to write to the backup.
+As a result, keys loaded from the backup must be marked as unauthenticated,
+leading to [usability
+issues](https://github.com/vector-im/element-web/issues/14323).
+
+[MSC3270](https://github.com/matrix-org/matrix-spec-proposals/pull/3270) tries
+to fix this issue by using a symmetric, authenticated encryption algorithm,
+which ensures that only someone who knows the secret key can write to the
+backup.  However this removes the ability for a client to be able to write to
+the backup without being able to read from it.
```
richvdh, line 22: This MSC could really do with explaining why the ability to write to the backup but not read from it it is useful.

Something something bots, I gather?

dkasak: Not just bots, but also as a prerequisite component for less privileged clients which don't have the ability to read all of history while still being able to participate in a room.

proposals/4048-signed-key-backup.md
sumnerevans: I agree that it feels weird to store an intermediate value (the MAC key) in SSSS.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,275 @@
...
+   key and the backup’s public key to generate a shared secret. The public half
+   of the ephemeral key, encoded using unpadded base64, becomes the `ephemeral`
+   property of the `session_data`.
+3. Using the shared secret, generate 80 bytes by performing an HKDF using
+   SHA-256 as the hash, with a salt of 32 bytes of 0, and with the empty string
+   as the info. The first 32 bytes are used as the AES key, the next 32 bytes
+   are discarded, and the last 16 bytes are used as the AES initialization
+   vector.  (This is the same as the key generation for
+   `m.megolm_backup.v1.curve25519-aes-sha2`, except that the generated MAC key
+   is discarded since it is unused.)
```
sumnerevans, line 71: Is there any reason to derive 80 bytes when we are discarding 32 of them? It seems it's just for compatibility with the (broken) `m.megolm_backup.v1.curve25519-aes-sha2` algorithm?

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,275 @@
...
+  encode the source of the key forward?)  Clients may create other values to
+  specify other sources, using the Java package naming convention; clients
+  should treat unknown values as `m.undefined`.
+
+For the `m.megolm.v1.aes-sha2` algorithm, the algorithm-dependent properties
+are the `forwarding_curve25519_key_chain`, `sender_claimed_keys`, `sender_key`,
+and `session_key` properties defined for
+`m.megolm_backup.v1.curve25519-aes-sha2`.
+
+### `m.megolm_backup.v1.curve25519-aes-sha2`
```
sumnerevans, line 127: I just want to clarify my understanding here. We are modifying the definition of the existing `m.megolm_backup.v1.curve25519-aes-sha2` algorithm that is already specified [here](https://spec.matrix.org/v1.13/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2) to make it compatible with `m.backup.v2.curve25519-aes-sha2`?

Why? I don't see any reason for uploading new keys to an old backup.

proposals/4048-signed-key-backup.md
```diff
@@ -0,0 +1,275 @@
...
+   SHA-256 as the hash, with a salt of 32 bytes of 0, and with the empty string
+   as the info. The first 32 bytes are used as the AES key, the next 32 bytes
+   are used as the MAC key, and the last 16 bytes are used as the AES
+   initialization vector.
+4. Stringify the JSON object, and encrypt it using AES-CBC-256 with PKCS#7
+   padding. This encrypted data, encoded using unpadded base64, becomes the
+   `ciphertext` property of the `session_data`.
+5. Pass the raw encrypted data (prior to base64 encoding) through HMAC-SHA-256
+   using the MAC key generated above. The first 8 bytes of the resulting MAC
+   are base64-encoded, and become the `mac` property of the `session_data`.
```
sumnerevans, line 154: This is not how the current algorithm works. See https://github.com/matrix-org/matrix-spec/pull/1712

It passes an empty byte string instead of the raw encrypted data.

proposals/4048-signed-key-backup.md
sumnerevans: Do any such clients or bots exist currently?

I feel like a bot or client writing but not reading from backup really should be an implementation requirement of this MSC to prove that it is actually useful to keep the ability to write but not read from the backup.

sumnerevans (commented):
I think that this MSC is a good opportunity to fix the issues with the existing backups (namely that the MAC is useless due to a libolm bug https://github.com/matrix-org/matrix-spec/pull/1712). There doesn't seem to be any reason to maintain backwards compatibility, and instead we should just hard-fork the backups (the old backups can be for old keys and new keys can go into the new backup).

I'm also still not entirely convinced that an asymmetric backup is the way to go, but I'm open to being shown examples of that being a useful feature.

proposals/4048-signed-key-backup.md
dkasak: I'm not sure whether they exist at the moment. But retaining the ability to have clients of reduced power (rather than requiring that they are all of equal, maximal power) seems like an obviously useful property.

Can you expand on how an example implementation would prove this further than a thought experiment? Or why the benefits are not obvious from the thought experiment?

sumnerevans: It's not obvious to me that a non-maximally-powerful client is useful. Maybe I have not been privy to some discussions about such potential use-cases?

Even the bot case that @richvdh doesn't really make sense to me. Why would I provide the bot access to my account?

I also don't know of any other chat network which has the concept of less-privileged clients (so as far as I know we are not copying prior art). It feels like such a client would be very confusing to end users.

I'm just a bit wary of doing things because it's a "useful property" when the property has not actually been shown to be useful (think: the overcomplicated megolm ratchet which has a mechanism for skipping by multiples of $$2^n$$ but is not used for >100 messages in most cases rendering the complexity pointless).

dkasak: As far as I know, the power asymmetry was deliberately factored into the current key backup mechanism, hence its asymmetric design. So while I fully sympathise with overengineering concerns, the switch to a symmetric design would lead to a loss of functionality. Which makes me think arguments to the contrary would be helpful: why an asymmetric design *is not* useful.

In this particular case, I feel the utility of the asymmetry is easy to demonstrate. Privacy oriented applications such as Signal deliberately provide limited history support, which makes them naturally more suitable for highly sensitive environments and provides additional assurance against your entire conversation history falling into the wrong hands. On the other hand, Matrix goes all in and history is first class replicated to all devices.

A way to disable history access on a particular device dramatically reduces this gap by giving you the ability to, for example, provision a phone from which you can talk securely with your contacts without the risk of your entire conversation history being disclosed if it's lost. However, given the asymmetric design, you still retain the ability to view those conversations from a more secure device such as a workstation with good physical security.

***

There is one large plot hole in the above: currently typical clients provide each other verified client with its own copy of the cryptographic user identity, negating the benefits described above (if you have the MSK, you can just ask for the write part of the backup key). So to actually benefit from this, this would need to be coupled with a way to withhold the identity from some devices.

sumnerevans: With [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153) and the general push for Invisible Crypto, there will basically never be a case where you don't perform cross-signing and share the master key.

I still think that having a client in the wild demonstrating an actual use-case for asymmetry would be ideal here, however I'm not willing to die on this hill.

chrkaatz:
May I ask for the progress and timeline for this topic? Is there a plan to have this implemented / finalized within the next couple of months?

We are facing an issue, that we shouldn't use libolm anymore in favor of vodozemac but need key backup as a hard requirement.

richvdh:
> May I ask for the progress and timeline for this topic? Is there a plan to have this implemented / finalized within the next couple of months?

We do hope to implement this in the medium term future, but it's not currently a focus for the Element team. "Next couple of months" is unlikely, I'm afraid.
 
> We are facing an issue, that we shouldn't use libolm anymore in favor of vodozemac but need key backup as a hard requirement.

I'm afraid I don't see the connection.

chrkaatz:
Of course, that connection is not easily visible. The Matrix protocol is used for one of Germany's healthcare digitalization efforts to enable communication between patients, healthcare professionals and even insurances. For that, there are specifications which defines the use of the matrix protocol and unfortunately, the agency responsible told us to switch to vodozemac in favor of megolm, but left open the issue regarding key backups even though the issue is known to them.

PR #4049, opened 2023-08-24
MSC4049: Sending events as a server or room

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/send-as-server-or-room/proposals/4049-send-as-server-or-room.md)

Dependencies:
* [MSC4047](https://github.com/matrix-org/matrix-spec-proposals/pull/4047): Send Keys
  * Depends on [MSC4046](https://github.com/matrix-org/matrix-spec-proposals/pull/4046)

### DIFF
--- a/proposals/4049-send-as-server-or-room.md
+++ b/proposals/4049-send-as-server-or-room.md
@@ -1,0 +1,180 @@
+# MSC4049: Sending events as a server or room
+
+Matrix rooms operate off a principle where only users can send events, and thus only users can change
+settings (state events) or send messages as themselves. When a server wants to make changes in a room
+on behalf of users, that server needs to "impersonate" or otherwise puppet a user account it controls
+in the room. While technically feasible, the approach is often disagreeable from (at minimum) a privacy
+perspective.
+
+A server might want to send an event without associated user when a user is not semantically responsible
+for the event. For example:
+
+* Keeping the `via` list for an [`m.space.child`](https://spec.matrix.org/v1.7/client-server-api/#mspacechild)
+  event updated.
+* Replacing the functionality of a [system alerts](https://spec.matrix.org/v1.7/client-server-api/#server-notices)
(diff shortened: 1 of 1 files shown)


proposals/4049-send-as-server-or-room.md
```diff
@@ -0,0 +1,180 @@
...
+### Full diff: Redaction algorithm
+
+**TODO**: This section. Ensure new power level properties are not redacted. `m.space.parent` might also need
+protecting in some way as it becomes an auth event.
+
+### Examples and test vectors
+
+**TODO**: This section.
+
+## Potential issues
```
turt2live, line 155: seemingly I forgot to mention `m.room.create`: it needs to be sent by a user ID, not a server (or room, not that it can...) because it implies that *servers* create rooms when it's actually users who do. I don't think anything overly breaks if we were to allow it, but the semantic reasons justify banning (imo).

andrewzhurov:
Sending as a "Room" sounds confusing to me. "Room" is [a rather unfortunate name](https://github.com/matrix-org/matrix-spec/issues/416#issuecomment-1055836830) for a Matrix Event Graph to begin with.

What the described use-case wants, it seems, is to publish events from some sort of role _identifier_ (e.g., Acme corp) that may be collectively hold. E.g., top-level officers are authorized to publish announcements on behalf of Acme corp id.

A way to manage such id in Self-Sovereign Identifier space is to e.g. use a MultiSig Autonomous Identifier (AID) of KERI.
Basically, it states keys that need to sign in order for event to be deemed authentic.
Given we can have Profile Rooms #1769, such room is naturally a MultiSig AID, that could be used to manage authorized keys and add authenticity to actions on behalf of this ID. E.g., create an announcement event with one of the authorized keys and link to it from Profile Room to add authenticity. Or simply have a dedicated Announcements Room (with MultiSig requirement on it's events), a kind of Announcer ID of Acme corp.

PR #4050, opened 2023-09-02
MSC ????: MUID verification

jucktnich:


proposals/4050-muid-verification.md
```diff
@@ -0,0 +1,75 @@
+# MSC4050: Matrix user id verification for third parties
+
+If you want to sign up for a service that requires a guaranteed contact method, the current standard process is to send a verification code or link to the user. A similar method is used for so-called magic links. But this method is heavily flawed since it requires you to change the app, and for verification codes to copy the code. Finally, you need to delete the message. This proposal should simplify this process with minimal effort for the websites and services, from now on called third parties, using this method.
```
turt2live, line 3: Please limit lines to ~100 characters.

turt2live:
@jucktnich please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md) on your changes to have this MSC be eligible for FCP later on in the process.

proposals/4050-muid-verification.md
```diff
@@ -0,0 +1,75 @@
...
+  - ```m.link``` ```Object``` ```optional```<br>This verification method lets the user open a link to verify his identity. This could be done without the user seeing it, just opening in the background after the user accepts the request.
+    - ```m.url``` ```String``` - Link that should be opened if the user accepts
+    - ```m.hide``` ```Bool``` ```optional``` - Should the user NOT see the webpage behind the link (Defaults to ```false```)
+    - ```m.expect_response``` ```Bool``` ```optional``` - If the matrix-client opens the link, should the client expect [json](#verify)  as the response and therefor display an error if this doesn't happen? (Defaults to ```false```)
+- ```m.disapprove_methods``` ```Object``` ```optional```<br>This object contains all possible methods to tell the webservice, that it was NOT you, who tried to sign up/log in. The methods are the same as for the ```m.verification_methods```.
+
+### Verify
+If the user takes an action, the server should send a new event to let the client know the current state.
+  - ```type``` ```String``` - ```m.muid_verification.result```
```
tulir, line 39: The conventional short form of "Matrix user ID" is "mxid", not "muid" (https://spec.matrix.org/v1.8/appendices/#user-identifiers). For JSON field names, `user_id` is also an option.

InezMc:
Private sign off has been recieved.

proposals/4050-mxid-verification.md
```diff
@@ -0,0 +1,92 @@
...
+| example.org wants to verify its you who signed up.|
+|  +------------------+       +------------------+  |
+|  |    Disapprove    |       |     Approve      |  |
+|  +------------------+       +------------------+  |
++---------------------------------------------------+
+```--->
+
+## Proposal
+### Requesting verification
+The back-end of the requesring server sends a matrix event to the user. The content object has the following keys:
```
uhoreg, line 25: The backend cannot send events directly to the user, it must be done in a room.  You should probably specify how the room is created and managed.

Also, you should specify what event type is used here.

PR #4051, opened 2023-09-03
[WIP] MSC4051: Using the create event as the room ID

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/room-id-is-create-event/proposals/4051-create-event-as-room-id.md)

### DIFF
--- a/proposals/4051-create-event-as-room-id.md
+++ b/proposals/4051-create-event-as-room-id.md
@@ -1,0 +1,48 @@
+# MSC4051: Using the create event as the room ID
+
+Matrix has a dedicated [room ID grammar](https://spec.matrix.org/v1.8/appendices/#room-ids) which
+aims to ensure there is a single identifier representing the room. This is done by using the origin
+server's name in the ID, which often leads to confusion among users and admins that the listed server
+"owns" the room.
+
+**TODO(TR): Insert security context.**
+
+This proposal uses the create event's event ID as the room ID instead, ensuring there is a fully
+unique value that can be used.
+
+## Proposal
+
+A new room version is established to accomodate the event format, redaction, and identifier changes
+from this proposal.
+
+The `room_id` field is removed from events, and no longer protected from redaction. The create event's
+reference hash is used and prefixed with `!` when a room ID is needed (such as in API endpoints).
+
+The create event is already selected as an auth event for all events in the room, allowing servers to
+easily determine which room ID the sent event belongs.
+
+Events served over the client-server API MUST still include a `room_id`. Clients should already be
+treating the value as opaque, and therefore should not break when encountering the serverless value.
+
+**TODO(TR): Maybe go into *a bit* more detail...**
+
+## Potential issues
+
+**TODO(TR): This**
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


andrewzhurov:
Seems like a good idea. Event id, being a hash, would uniquely identify a room.
Room is a communication context. Oftentimes communication within a Room leads to new conversations that is losely related to the original context, we model it as Threads.
Perhaps we could go a step further and make Threads their own Room (a separate Matrix Event Graph), that forked from a particular event of a parent Thread. Would improve efficiency on scroll, as one scrolling the parent Thread would not need to load child's Thread events (that are not displayed anyways).

proposals/4051-create-event-as-room-id.md
```diff
@@ -0,0 +1,48 @@
...
+
+This proposal uses the create event's event ID as the room ID instead, ensuring there is a fully
+unique value that can be used.
+
+## Proposal
+
+A new room version is established to accommodate the event format, redaction, and identifier changes
+from this proposal.
+
+The `room_id` field is removed from events, and no longer protected from redaction. The create event's
```
tranquillity-codes, line 18: `room_id` is still to be kept in the encrypted payload of E2EE messages, right ? I think it should probably be mentioned in the MSC, perhaps as a security consideration.

networkException: What do you mean? Clients don't encrypt the room id when sending events

poljar: They do: https://spec.matrix.org/v1.11/client-server-api/#mmegolmv1aes-sha2

networkException: Ok, then thats probably hidden by the various view source implementations in Element X Android / Android and Web

turt2live: Clarification on what the concern is here would be appreciated. I'm having trouble parsing both the original comment and replies.

Links to existing spec are helpful.

tranquillity-codes: If I understand the proposal right, nowhere in it does it state to keep room_id for events *outgoing* from the client inside m.room.encrypted, to address:
> We include the room ID in the payload, because otherwise the homeserver would be able to change the room a message was sent in.
Link https://spec.matrix.org/v1.11/client-server-api/#mmegolmv1aes-sha2

turt2live: This MSC doesn't really operate at that layer, so would not affect encryption. It only affects the federation and room behaviour characteristics. 

I can try to make this clearer, but clients should see no material difference before and after this MSC.

tranquillity-codes: Yea, I would suggest to make it clearer in the MSC exactly which usages of room_id it affects

uhoreg: With my crypto hat on, since has security implications, I think it would be worth clarifying somewhere that this MSC doesn't affect the encrypted payload.

tulir:
This has been replaced by MSC4291

PR #4056, opened 2023-09-15
MSC: Role-Based Access Control (mk II)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/rbac-pl/proposals/4056-role-based-access-control.md)

### DIFF
--- a/proposals/4056-role-based-access-control.md
+++ b/proposals/4056-role-based-access-control.md
@@ -1,0 +1,171 @@
+# MSC4056: Role-Based Access Control (mk II)
+
+A common request in Matrix is the ability to assign a set of permissions to a set of users without
+having to give inherited permissions. Power levels currently provide an ability to define roles,
+though each higher power level inherits the permissions from under it. A more formal Role-Based
+Access Control (RBAC) mechanism would allow a user to inherit permissions from another role while
+having specifics be negated, unlike power levels.
+
+This proposal aims to improve upon the learnings of [MSC2812](https://github.com/matrix-org/matrix-spec-proposals/pull/2812),
+particularly around the decentralization and state resolution components. The specific permissions
+supported by roles are left out of scope in this version of the proposal.
+
+MSC2812 fails to operate when state resolution is considered due to how it is structured. In isolation,
+the MSC offers a sturdy framework for RBAC, but when two changes utilizing a role are in conflict,
+state resolution will fail to pick a "winning" change. This proposal addresses the problem by applying
+a power level structure over top of roles, providing a deliberate hierarchy of roles.
+
+**TODO: Insert text about why RBAC is cool/wanted, and why we've avoided it in the past (NTFS hell,
+flight console UX, etc)**
+
+## Potential issues
+
+*Author's note: This section is up here to be clear about which problems are considered TODO items.*
+
+This MSC does *not* have:
+
+* Backwards compatibility with existing clients/tooling.
+* Explicit permissions a role could have. It is assumed for now that the permissions implied by power
+  levels are maintained at a minimum. For example, ability to send an event of a given type.
+* A complete solution. It outlines a framework at the moment.
+
+## Proposal
+
+In a future room version, due to dramatic changes relating to how rooms work, we do the following.
+
+`m.room.power_levels` is stripped of its usefulness. Clients MAY send it, but it will be treated as
+any other random event. Servers MAY optionally decide to prevent clients from sending it by accident
+in applicable room versions.
+
+A new `m.role` state event is introduced, with `state_key` being an arbitrary opaque string. The
+`state_key` becomes the "Role ID". The `content` for the role is as follows:
+
+```jsonc
+{
+  "profile": {
+    // display name, avatar, colour, etc? Use extensible events content blocks where reasonable.
+  },
+  "permissions": {
+    // TODO: Replace example with actual permission (m.invite or whatever)
+    "org.example.my_permission": true,
+    "org.example.another": {
+      "is_complicated_type": true
+    }
+  }
+}
+```
+
+`m.profile` is inspired by [MSC3949](https://github.com/matrix-org/matrix-spec-proposals/pull/3949),
+where allowing a user to define such characteristics is clearly desirable. `m.permissions` are the
+permissions themselves (which may be non-boolean values).
+
+The users "affected" by this role are described in a second new state event: `m.role_map` with an
+empty string `state_key`. The `content` being:
+
+```jsonc
+{
+  "role_id": {
+    "users": ["@alice:example.org"],
+    "order": 1
+  },
+  "another_role_id": {
+    "users": ["@bob:example.org"],
+    "order": 5
+  }
+}
+```
+
+Each role MUST have a distinct `order` number. Roles with higher `order` override conflicting permissions
+from lower `order` roles. For example, given the following:
+
+* "Role A" with `order: 1` and permission `first: true, second: true`.
+* "Role B" with `order: 2` and permissions `first: false, third: true`.
+
+... a user with both roles would have permissions `first: false, second: true, third: true`. `first` is
+overridden by the `order: 2` role, but `second` and `third` are not conflicting.
+
+When an event is sent by a user, it MUST reference the `m.role_map` and `m.role` events applicable
+to the `sender` from that map as `auth_events`.
+
+MSC2812 has a fundamental flaw in how it authorizes events, preventing it from working reliably in
+a decentralized environment. When a user with Role A is trying to ban a user, and the target user
+is removing that permission from Role A at the same time, the conflict becomes impossible to predict
+because state resolution will ultimately tiebreak on event ID. This unexpected behaviour can be (at
+best) confusing for users.
+
+This MSC fixes that flaw by retaining a concept of "power level", but only applying it to the roles
+themselves through `order`. When state resolution is attempting to resolve a power struggle, it would
+first determine which role has higher `order`, then let that event "win" the conflict. In the cases
+where the two involved users have the *same* role, timestamps and event IDs are used as tiebreaks,
+like with any other event. In other words, a user's power level is effectively the highest `order`
+role they possess with the required permission(s).
+
+When the users involved in a conflict are at the same effective power level, it is less surprising
+to those users that a conflict would be resolved by timestamp or "flip of a coin" (event ID ordering).
+For other cases however, users will typically have an implied hierarchy in mind even if `order` didn't
+exist, leading to an expectation that users with "Admin" have "more power" than those with "Moderator".
+The `order` field formalizes the user preference (and therefore expectation).
+
+For the above reasons relating to state resolution, we intentionally mix user assignment and ordering
+into the same event type. Ideally, the `order` (and possibly even assignment) would be done on the
+`m.role` events themselves, however in a conflicted set of state it becomes difficult or impossible
+to enforce "one role per order number" during event authorization.
+
+### Detailed authorization rule changes
+
+**TODO**
+
+### Detailed state resolution changes
+
+**TODO**
+
+### Detailed redaction algorithm changes
+
+**TODO**
+
+### Test vectors
+
+**TODO**
+
+## Potential issues
+
+**TODO: Complete more of this section.**
+
+* A limited number of roles can be used and assigned, as `m.role_map` will be subject to size restrictions.
+* This MSC is intentionally does not support per-user role ordering. This is a complicated feature to
+  represent (when considering state resolution), and is empirically not used by Discord. A future room
+  version may introduce such a feature.
+
+## Alternatives
+
+[MSC2812](https://github.com/matrix-org/matrix-spec-proposals/pull/2812) is the primary alternative to
+this MSC. [MSC3073](https://github.com/matrix-org/matrix-spec-proposals/pull/3073) inherits from MSC2812
+and is subject to the same issues described by this proposal. MSC3073 does introduce per-user role
+ordering, though not in a way compatible with state resolution/conflicts.
+
+[MSC3949](https://github.com/matrix-org/matrix-spec-proposals/pull/3949) aims to introduce the "profile
+information" for a role without introducing RBAC itself. This proposal inherits the ideas of MSC3949.
+
+No other significant alternatives are considered. Some minor representation details are possible, however
+clients like Discord are generally seen as the "gold standard" for user-facing RBAC implementation. Those
+clients don't make use of many theorized alternatives.
+
+## Security considerations
+
+**TODO**
+
+## Unstable prefix
+
+While this proposal is not incorporated into a stable room version, implementations should use `org.matrix.msc4056`
+as an unstable room version, using [room version 11](https://spec.matrix.org/v1.8/rooms/v11/) as a base. The
+event types are also prefixed in this room version.
+
+| Stable | Unstable |
+|--------|----------|
+| `m.role` | `org.matrix.msc4056.role` |
+| `m.role_map` | `org.matrix.msc4056.role_map` |
+
+## Dependencies
+
+As of writing, this MSC is being evaluated as a potential feature for use in the MIMI working group at
+the IETF through the Spec Core Team's efforts.


gabrc52:
Thanks for this MSC! I really like the role feature (but I am biased because I'm also a Discord user)

* Another alternative to consider with ordering is to put it in the `m.role` itself, in a similar way to [`m.space.child`](https://spec.matrix.org/v1.8/client-server-api/#ordering-of-children-within-a-space). However, it might make it harder for clients to implement reordering. What are the arguments for and against having a separate `m.role_map` (other than consistency)?
* I can confirm that in my experience, Discord's per-user role ordering is barely used if at all.
* How would looking up someone's roles in a specific room look like? (for example, if clients want to implement displaying the color for the highest role, or showing someone's roles when clicking on their profile) Would it make sense to have an endpoint for this, or should clients be responsible for keeping track of the state events (as is done for names and avatars)?

Off-topic(?) suggestions for future MSCs which may inform design of this MSC:

* (Discord feature) It would be nice to be able to mention roles
* (Discord feature) It would be nice to have the ability for roles to be space-wide as opposed to/in addition to room-wide. How could space-wide display names, roles, etc look like?

proposals/4056-role-based-access-control.md
turt2live, line 1: @gabrc52 says:

> (I'm not sure what's the process for regular users to leave comments, so let me know if I'm doing it wrong)
> 
> Thanks for this MSC! I really like the role feature (but I am biased because I'm also a Discord user)
> 
> * Another alternative to consider with ordering is to put it in the `m.role` itself, in a similar way to [`m.space.child`](https://spec.matrix.org/v1.8/client-server-api/#ordering-of-children-within-a-space). However, it might make it harder for clients to implement reordering. What are the arguments for and against having a separate `m.role_map` (other than consistency)?
> * I can confirm that in my experience, Discord's per-user role ordering is barely used if at all.
> * How would looking up someone's roles in a specific room look like? (for example, if clients want to implement displaying the color for the highest role, or showing someone's roles when clicking on their profile) Would it make sense to have an endpoint for this, or should clients be responsible for keeping track of the state events (as is done for names and avatars)?
> 
> Off-topic(?) suggestions for future MSCs which may inform design of this MSC:
> 
> * (Discord feature) It would be nice to be able to mention roles
> * (Discord feature) It would be nice to have the ability for roles to be space-wide as opposed to/in addition to room-wide. How could space-wide display names, roles, etc look like?

(please post comments on the Files Changed tab)

turt2live: > * Another alternative to consider with ordering is to put it in the `m.role` itself, in a similar way to [`m.space.child`](https://spec.matrix.org/v1.8/client-server-api/#ordering-of-children-within-a-space). However, it might make it harder for clients to implement reordering. What are the arguments for and against having a separate `m.role_map` (other than consistency)?

This isn't done due to the state resolution concerns raised in the MSC. It opens the door to two roles having the same order, creating problems (aside from the problems associated with non-canonical events).

> * I can confirm that in my experience, Discord's per-user role ordering is barely used if at all.

I couldn't even find it as an option.

> * How would looking up someone's roles in a specific room look like? (for example, if clients want to implement displaying the color for the highest role, or showing someone's roles when clicking on their profile) Would it make sense to have an endpoint for this, or should clients be responsible for keeping track of the state events (as is done for names and avatars)?

This MSC assumes clients have ways of managing state events in rooms on their own. If a dedicated endpoint is required, it would be in another MSC.

> Off-topic(?) suggestions for future MSCs which may inform design of this MSC:
> 
> * (Discord feature) It would be nice to be able to mention roles
> * (Discord feature) It would be nice to have the ability for roles to be space-wide as opposed to/in addition to room-wide. How could space-wide display names, roles, etc look like?

These suggestions sound very reasonable and desirable, though they'd be handled by a future MSC (or two).

proposals/4056-role-based-access-control.md
```diff
@@ -0,0 +1,175 @@
...
+
+MSC2812 has a fundamental flaw in how it authorizes events, preventing it from working reliably in
+a decentralized environment. When a user with Role A is trying to ban a user, and the target user
+is removing that permission from Role A at the same time, the conflict becomes impossible to predict
+because state resolution will ultimately tiebreak on event ID. This unexpected behaviour can be (at
+best) confusing for users.
+
+This MSC fixes that flaw by retaining a concept of "power level", but only applying it to the roles
+themselves through `order`. When state resolution is attempting to resolve a power struggle, it would
+first determine which role has higher `order`, then let that event "win" the conflict. In the cases
```
cvwright, line 98: Can users have multiple roles?  And if so, how do you decide which role to use for each user in the power struggle?

For a given sender you could take their role with the highest `order`.  But then does this require that the powers given to higher-order roles must be a superset of lower-order roles?  Because what if the highest-order role does not itself include the power to send the given event?

You could decide whether the event itself is authorized by looking at the user's total powers as the set union of all their roles' powers.  And to resolve conflicts, you could take the user's highest-order role as the user's order in the power struggle.  Does that work?

proposals/4056-role-based-access-control.md
```diff
@@ -0,0 +1,175 @@
...
+When the users involved in a conflict are at the same effective power level, it is less surprising
+to those users that a conflict would be resolved by timestamp or "flip of a coin" (event ID ordering).
+For other cases however, users will typically have an implied hierarchy in mind even if `order` didn't
+exist, leading to an expectation that users with "Admin" have "more power" than those with "Moderator".
+The `order` field formalizes the user preference (and therefore expectation).
+
+For the above reasons relating to state resolution, we intentionally mix user assignment and ordering
+into the same event type. Ideally, the `order` (and possibly even assignment) would be done on the
+`m.role` events themselves, however in a conflicted set of state it becomes difficult or impossible
+to enforce "one role per order number" during event authorization.
```
cvwright, line 112: Yes, it seems important to have all of the `order`s for all the roles kept in one place so you can guarantee it's always internally consistent.

Would it be possible to move the assignment of users to roles out of the role map?  If you had one "authorization" event per user, then that could save you from some of the challenges with the current `m.power_levels` event:
1. You're no longer limited by the max size of an event.  For a very large room, you could have tens of thousands of muted users.
2. Less chance of conflicts in state res when you have mods / admins on different servers all taking action at the same time.  You could have mod A muting user X, mod B muting user Y, and mod C promoting user Z all at the same time.

proposals/4056-role-based-access-control.md
Mathias-g: @cvwright I think going with how it works in Discord whenever possible is preferable, and in Discord it is the cumulative powers that roles assigned to you give you. So for instance, if you have the ability to join a sub-room in a space with one role, but not the ability to post messages in it, but you have the ability both to join the room and post from another role, you can both join the room and post in it, no matter the order of the roles.

Cloudwalk9:
I just want to throw my idea in. I already mentioned it to @turt2live but I want to share it more publicly.

What if we implement something like this in two stages that don't necessarily depend on each other, and maintain backwards compatibility with the existing power level system, essentially retaining it and making migration to newer room versions very seamless?

## Per-user permission overrides
Allow users to be granted permissions on an individual basis regardless of their power level, and they can perform an action/send an event with the authority of their power level. Say, kick requires PL50, User is PL25 but is granted kick. User can only kick users PL24 and below.

## Groups
Implement "roles" as groups. Users can be in multiple groups. Groups are assigned power levels just like users, and inherit permissions in the same way as users.

## Putting these together

Per-group permission overrides and you get RBAC.

proposals/4056-role-based-access-control.md
florianjacob: I also was wondering whether more separation is possible, brought the point up in #matrix-spec:matrix.org - [@turt2live said](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$5iv52NcjNQVA7yUkhUewL0NfxP-rtlwijJz37qUbom4?via=matrix.org&via=envs.net&via=element.io) that separating role assignment and role ordering into two separate big events would be a point of conflict already, but was not sure about the exact argument anymore. [@Gnuxie remembered a big discussion](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$D78SThAibqPSOfMKYHX-levWvqift6HiT4pkXqG6Z3M?via=matrix.org&via=envs.net&via=element.io) on the topic, thinking that the basic issue is that changing role assignments and role ordering might very well require atomicity of changes so that there is no time in between the then-separate events where the users have something else than the intended final permissions.

Gnuxie: The discussion started [here](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$Ap5RtPYuY8McjjDkJyO27yZE5f7TH-8bQFlG3VTk_ZI?via=matrix.org&via=envs.net&via=element.io), picks up on the point [here](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$75ljaekukZzTzut9wSNLKNKj1GoUvTPc3g7Ct6l4KLY?via=matrix.org&via=envs.net&via=element.io) and ended [here](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$Ap5RtPYuY8McjjDkJyO27yZE5f7TH-8bQFlG3VTk_ZI?via=matrix.org&via=envs.net&via=element.io).

From reading this over, I misremembered. There are two problems:

1. When membership and permissions are described in the same event, and the subject sends this event, there is nothing in the auth rules that can constrain that user from declaring themselves to have whatever permissions they like. [discussion](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$0pv9JVVKzuRE6mVBUGQMq44vNTZ1-l19yFcKgqt8Zl8?via=matrix.org&via=envs.net&via=element.io).  
2. When membership and permissions are described in the same event, room admins also have the power to change the permissions of other users. If an admin changes the permissions while another user leaves, that user can either be forced back into the room against their will, or the admin's permission changes are ignored. [discussion](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$xDKbNNgi_Qi8nwQS6DIbbcmwgk_TS86hSTPW9PirtHo?via=matrix.org&via=envs.net&via=element.io). 

If it helps, the way i have avoided this issue in other proposals ([MSC4124](https://github.com/matrix-org/matrix-spec-proposals/pull/4124)) is to separate membership into two events, one that room admins use to describe whether the subject (in this case servers) is allowed to participate, and one that the subject controls that describes their intent to participate.

proposals/4056-role-based-access-control.md
Nursehlatshwayo: ❤️

proposals/4056-role-based-access-control.md
```diff
@@ -0,0 +1,175 @@
...
+from lower `order` roles. For example, given the following:
+
+* "Role A" with `order: 1` and permission `first: true, second: true`.
+* "Role B" with `order: 2` and permissions `first: false, third: true`.
+
+... a user with both roles would have permissions `first: false, second: true, third: true`. `first` is
+overridden by the `order: 2` role, but `second` and `third` are not conflicting.
+
+When an event is sent by a user, it MUST reference the `m.role_map` and `m.role` events applicable
+to the `sender` from that map as `auth_events`.
```
Foxtrek64, line 88: I've used several systems with this approach and I've never really been a fan. In many cases, I have several roles that should be considered "equal" in power. For example, I might have separate donor roles for individual subscription-style donation services (think patreon) - I don't want patreon donors to have more sway than donors of other platforms, which is to me what this order sets up. Many systems with this order property also explicitly disallow multiple roles having the same order, because then the conflict resolution system falls apart.

Additionally, this sort of weighting creates situations where you have users able to do things that you thought they weren't, or unable to do things that you think they should be able to do.

The order can work for some places - Discord uses this for determining what roles a user with the Manage Roles permission is allowed to manage, the color of a user's name, and a few other cosmetic items. However, conflict resolution is handled differently.

The two most prevalent systems appears to be `Explicit Allow Wins` (Discord) and `Explicit Deny Wins` (Microsoft AD). In order to implement this however, we would need to effectively have a tri-state. I believe this can be implemented in json by omitting the value, such as in your example (role  2 does not specify a value for `second` and therefore it doesn't apply any rules to it). In code, this could be represented as a `null` value.

The "undefined" value is effectively a pass-through. It is never considered in the permission flow. By default, all permissions are "undefined", which effectively functions as a soft deny which can be then overridden by an explicit allow. This leads to three possible conditions:

1. No permission - the user is not a member of any roles that allow this permission. Effective: deny.
2. Denied - the user is a member of a role that denies this permission.
3. Allowed - the user a member of a role that allows this permission.

On the UI side, I like the phrase "unset" or "not set" to describe the functionality of this value.

We would still need to choose between allow/deny winning, but here's what those might look like:

**Explicit Allow Wins**
* "Role A" with and permission `first: true, second: true`.
* "Role B" with and permissions `first: false, third: true`.

`first` is set true. Override for false is ignored. Result: allow.
`second` is set true. Value is not overridden. Result: deny.
`third` is set true. Value is not overridden. Result: deny.

**Explicit Deny Wins**
* "Role A" and permission `first: true, second: true`.
* "Role B" and permissions `first: false, third: true`.

`first` is set `true` but overridden by `false`. Result is: deny
`second` is set `true` and not overridden. Result is: allow.
`third` is set `true` and is not overridden. Result is: allow

Also to consider - we should have a built-in everyone role. Just about every system has this, and it is a base set of permissions applied to all users. This allows for simplified configuration scenarios, allowing you to set common values for everyone (e.g. everyone can read chat history and speak). It also allows for more advanced scenarios, such as marking speak as undefined and granting it via another role.

Cloudwalk9: > Also to consider - we should have a built-in everyone role. Just about every system has this, and it is a base set of permissions applied to all users. This allows for simplified configuration scenarios, allowing you to set common values for everyone (e.g. everyone can read chat history and speak). It also allows for more advanced scenarios, such as marking speak as undefined and granting it via another role.

Power level 0 is the everyone role. That is, I think we should keep PLs around to enforce rank hierarchy on groups/roles. I don't like the idea of wholesale getting rid of them. It would keep backwards compatibility and make RBAC opt in on old rooms after upgrade, and opt out on new rooms if for some reason you decide you want PL based permissions. The idea is the "everyone" role would be implicit by comprising any person not assigned a role and having permissions no higher than PL0.

mpeter50: @Foxtrek64 I think you made a mistake in the Explicit Allow Wins example. Shouldnt all of them result in an allow?

This approach looks more user friendly to me than what is in the current document. But, maybe it would worth it to be able to still override an allow, either with a third (or actually forth) value for a permission, or a property that applies to the whole role. One use of it could be to temporarily revoke permissions even for a moderator.
How is it normally handled in discord communities when a moderator does something questionable, and their permissions are suspended until the circumstances are clarified?

But, as Cloudwalk9 also said, maybe we could reinterpret what a power level means: a role with a higher power level has more power, and so it can revoke permissions granted by roles with a lower level.
This way users wouldnt have a power level anymore, but a default role instead.
Roles with the same power levels would work according to the Explicit Allow Wins principle, like on Discord. If a community does not want to deal with power levels and the complications of overrides, they could just keep using the same power levels for all roles, maybe a separate one for room staff.
But to avoid confusion, it should be explicitely noted in the spec then that power levels have an entirely new meaning.

Cloudwalk9: > But, as Cloudwalk9 also said, maybe we could redefine what a power level means: a role with a higher power level has more power, and so it can revoke permissions granted by roles with a lower level.
> This way users wouldnt have a power level anymore, but roles instead.
> Roles with the same power levels would work according to the Explicit Allow Wins principle, like on Discord. If a community does not want to deal with power levels and the complications of overrides, they could just keep using the same power levels for all roles, maybe a separate one for room staff.
> But to avoid confusion, it should be explicitely noted in the spec then that power levels have an entirely new meaning.

Not necessarily a new meaning but an extension of their purpose. "Permission overrides" are key here. I think RBAC should be emergent from two separate MSCs.

1. Should implement roles and figure out their ranking and coloring semantics just for vanity, so users can be assigned named roles.

2. Crucially, the ability to allow or deny a user or group specific permissions regardless of their power level. So we have a concept of "effective power level" vs actual power level. Someone is granted kick at PL25 when kick requires 50. They can only kick 24 and below. PL25 is their effective power level for kick. A group/role is granted ban at PL20 and they're added to it, they can only ban PL19 and below because that's the effective power level inherited from their group. But now they're added to another group with PL40 that has kick granted explicitly. Logically this would create a weird inversion where someone can kick people who are allowed to ban, but I can foresee some weird niche use cases. Ultimately power level should have the final say just as it does now.

They can be implemented independently and RBAC can emerge from them either naturally or with a third, simpler MSC that makes groups subject to the auth rules, or the MSC implementing groups/roles could make them subject to the auth rules from the get-go and permission overrides are figured out in another MSC. I personally think this is the most dynamic, backwards compatible way forward and it keeps the old system around nonintrusively.

mpeter50: > when kick requires 50

Do we want to keep using power levels for plain permissions too? I think that would complicate things unnecessarily.

> personally think this is the most dynamic, backwards compatible way forward

Is it useful to have backwards compatibility here? This would need a new room version anyway.

Cloudwalk9: > > when kick requires 50
> 
> Do we want to keep using power levels for plain permissions too? I think that would complicate things unnecessarily.
> 
> > personally think this is the most dynamic, backwards compatible way forward
> 
> Is it useful to have backwards compatibility here? This would need a new room version anyway.

Nah, the defaults for *new* rooms under this model could be every current mod permission set to PL100 (owner) by default with an optional preset "Moderator" role at PL50, but it's granted the permissions using new allow/deny list overrides. Room upgrades would just be ridiculously easy, perhaps clients could show a prompt indicating the new feature so admins can transition at their leisure, since the old system is not only still around but forms the new system's rank enforcement.

Foxtrek64: > But, maybe it would worth it to be able to still override an allow, either with a third (or actually forth) value for a permission, or a property that applies to the whole role. One use of it could be to temporarily revoke permissions even for a moderator.

I come from the Microsoft world so I may be a bit biased here. While I described both systems, I prefer and am more familair with Explicit Deny Wins.

In your example, if I had a moderator that was abusing their power I would simply remove them from the moderator role. I can always restore them later once the situation is resolved. However, if I wanted to follow your example I could create a restricted moderator role that prevents them from exercising certain moderator powers just by setting those permissions to deny. Because of the explicit deny, the effective permission is deny.

Or another example - let's say I have a community where people are unable to speak unless they have the Member role. How they get this role is not important but maybe it's provided via an integration, e.g. having an account on a website and linking your Matrix account. But then for moderation I also want to have a Muted role that prevents a particular member from speaking. This is how I would set that up:

Everyone:
 - Read Messages: Allow
 - Send Messages: Not Set
 Member:
 - Read Messages: Not Set
 - Send Messages: Allow
 Muted:
 - Read Messages: Not Set
 - Send Messages: Deny
 
 In practice, what this means is that if the user does not have the member role, they are allowed to read messages and not speak.
 When the member role is added, Read Messages is not set so it doesn't apply, meaning the Allow value from Everyone is what is used. The Send Messages value is now set to Allow, enabling speaking perms.
 When the member is added to the Muted role, their ability to read messages is not changed. But their ability to send messages is explicitly denied. This prevents them once again from sending the message but allows them to keep their member role. This may be desirable depending on the meaning of the role - someone could be a patron for instance, but have their ability to speak be temporarily rescinded. Just because they are in trouble does not mean they are no longer a member or a patron.

>  I think you made a mistake in the Explicit Allow Wins example. Shouldnt all of them result in an allow?
Yes, thanks for catching that.

Foxtrek64: > Nah, the defaults for new rooms under this model could be every current mod permission set to PL100 (owner) by default with an optional preset "Moderator" role at PL50, but it's granted the permissions using new allow/deny list overrides. Room upgrades would just be ridiculously easy, perhaps clients could show a prompt indicating the new feature so admins can transition at their leisure, since the old system is not only still around but forms the new system's rank enforcement.

I do like the "let's do both" idea. These systems could be (and should be) entirely separate from each other, and I think the cleanest option is to just prompt when creating the room what system you want. This also allows older rooms to be upgraded to a version that supports RBAC and to transition at the admin's leisure, as you suggest.

I would use an enum property to denote whether the room uses RBAC or Power Levels, and which configuration is used and which is ignored depends on the value of this property. A boolean value could be used but I tend to reserve boolean values for things that can be answered with yes or no (e.g. whether to overwrite a file). When it's a "which one" question, I prefer enums.

If Power Levels is selected, the system continues to work exactly as it does today. But if RBAC is selected, then the power level configuration is ignored (but still allowed to be present), and instead the role permission check is used.

This bright line separation means that we don't have to figure out how to get roles and power levels to behave together. I personally believe that RBAC and Power Levels cannot be made to work together - you could use roles and power levels, and anyone with the role has the associated power level, but then that's not RBAC.

The only consideration we would need if we move to the system proposed in this thread is to somehow recognize the room owner. There may be a facility for this already, or it may need to be added. The idea is that the room owner should, even with only the Everyone role, be able to create roles, delegate permissions, and manage other aspects of the room.

mpeter50: I agree that its probably better to have it either traditional power levels or RBAC, and not both at the same time.

But I think it would be useful to be able to assign roles to power levels (maybe a different name), for setting up an explicit order in which roles' contradicting directives override each other.
In this setup users wouldnt have power levels, only roles would. Roles with the same power level would be processed according to Explicit Allow Wins, they would "add up", otherwise the role with the higher power level wins for permissions that are specified in multiple roles.

> The only consideration we would need if we move to the system proposed in this thread is to somehow recognize the room owner. There may be a facility for this already, or it may need to be added.

I dont think there is a system for that now. For rooms that work with power levels, or are upgraded from a previous room version, perhaps the owner could be designated as the user with the higher power level. If there are multiple such, then the one that's latest join time was the earliest.
For rooms that work with RBAC, assuming that the specification ends up assigning power levels to roles, the owner could be designated as the oldest member in any of the highest power level roles.

Perhaps it could be further limited so that a bot account cannot be designated as the owner, to avoid trouble.
This way, basically permissions cant be messed up in such a way that the room is broken.

> Edit: I may create a separate proposal for this one, since this proposal is so fundamentally different.

Yeah, room owner designation is better to have in a separate MSC, it is a different concept that this one would make use of.

Cloudwalk9: What I'm proposing basically is that they can coexist. Power level always take precedence but ideally in a new room version, it takes a backseat to RBAC as the new default, and is just used to enforce ranking in a new default configuration. Not that they're "separate" as in you choose one or the other, but they both mesh. Clients should be discouraged from exposing the knobs for both simultaneously of course.

I could see like a "diplomatic immunity" use case where someone is granted a power level that can't kick or ban, but someone in a role with the same power level who CAN otherwise kick or ban by an override can't do it to this person, because the "effective power level" for the kick matches the user's power level, so it's not allowed.

PR #4057, opened 2023-09-22
MSC0000: Static Room Directory

networkException:


tcpipuk:
Currently it links to this, which lists the homeservers in a room and requires authentication:
https://spec.matrix.org/v1.8/server-server-api/#get_matrixfederationv1querydirectory

I think you meant to link to this for the current non-static room directory:
https://spec.matrix.org/v1.8/server-server-api/#get_matrixfederationv1publicrooms

proposals/4057-static-room-directory.md
```diff
@@ -0,0 +1,114 @@
...
+
+As it is expected that a web server administrator has higher privilege than a user of a potential homeserver
+on the same domain name, the mapping defined at this path takes precedence over a potential room directory
+served over federation.
+
+The following code block is an exemplary static room directory served at `https://example.com/.well-known/matrix/rooms`:
+
+```json
+{
+    "#hq:example.com": "!OGEhHVWSdvArJzumhm:matrix.org",
```
uhoreg, line 28: Just having the room ID isn't enough.  To join a room, we need to know what server(s) can be used to join (like the `via` parameter in [`matrix` URIs](https://spec.matrix.org/unstable/appendices/#matrix-uri-scheme)

proposals/4057-static-room-directory.md
```diff
@@ -0,0 +1,114 @@
...
+limits.
+
+As resolving inherently exposes metadata to a third party, caching might also be desireable for privacy reasons.
+
+Users on homeservers that do not support this feature will not be able to resolve room aliases served statically.
+A workaround for this could be limiting the ability for setting a statically served room alias as the canonical alias
+of a room to a new room version, this could discourage users of using the feature in a state in which few homeservers
+support it.
+
+## Alternatives
```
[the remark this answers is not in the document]
networkException: Perhaps the naming is a little unclear here. Essentially the goal of this MSC is to allow the functionality of mauliasproxy by just hosting a static file.

From my understanding that functionality would generally be regarded as *room aliases* (using `/_matrix/federation/v1/query/directory`), whereas the endpoint that whats called *(public) room directory* is what clients can query via their server over federation to search for rooms (using `/_matrix/federation/v1/publicRooms`). I'm not sure how these interact currently, but perhaps the MSC would be better off to be named "Static Room Aliases" then?

proposals/4057-static-room-directory.md
networkException: Included `servers` to match https://spec.matrix.org/v1.8/server-server-api/#get_matrixfederationv1querydirectory

proposals/4057-static-room-directory.md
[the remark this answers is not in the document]
networkException: Renamed the proposal and clarified this

proposals/4057-static-room-directory.md
```diff
@@ -0,0 +1,135 @@
...
+statically. A workaround for this could be limiting the ability for setting a statically served room alias
+as the canonical alias of a room to a new room version, this could discourage users of using the feature in
+a state in which few homeservers support it.
+
+## Alternatives
+
+Instead of defining room alias mappings in a well-known path, the current `.well-known/matrix/server` content
+could be extended to include a new field, delegating the room aliases for an entire domain to the federation
+api of a different homeserver. This would also require some sort of permission system to support the use case
+described in the abstract however, as otherwise any user on a homeserver would be able to create a room alias.
```
networkException, line 102: Instead of delegating this, the `/.well-known/matrix/server` file could also be extended to include a key like `m.static_room_aliases`

proposals/4057-static-room-directory.md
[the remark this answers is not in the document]
dasha-uwu: @networkException

>is to allow the functionality of mauliasproxy by just hosting a static file.

this shifts the complexity (extra http request, caching, etc) from a static alias provider (using the proxy or the proposed mechanism) to the alias resolver. i don't see why we would want to do this given that there are much more alias resolvers than static alias providers in the public matrix network

proposals/4057-static-room-directory.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+A static room alias mapping is a JSON string of an object mapping from room aliases as keys to an object
+with the corresponding room id in the `room_id` field and a list `servers` containing at least one server name.
+
+### Resolving steps
+
+To resolve an room alias `#hq:example.com` with the first example above,
+
+1. first a server MUST now send an https `GET` request to alias domain's `.well-known/matrix/roms` path (given `example.com` that is `https://example.com/.well-known/matrix/rooms`) and decode the served JSON.
```
Henry-Hiles: s/roms/rooms, probably :)

proposals/4057-static-room-directory.md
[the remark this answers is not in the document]
networkException: Right, I'd argue that there aren't more static aliases because it's hard to set up.

Even when people looking to provide vanity aliases know about the federation API technically supporting such behavior, the user experience is quite bad. You need to think about maybe managing a signing key and at the very least you need a Web Server that can switch on query parameters (NGINX doesn't by default!)

proposals/4057-static-room-directory.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+and mistakenly override an internal mapping of `#matrix:matrix.org` to `!evilRoomId:matrix.org`. Homeservers
+should take care to verify that mappings actually match the domain they were requested on and discard any
+others.
+
+## Unstable prefix
+
+[MSC2324](https://github.com/matrix-org/matrix-spec-proposals/pull/2324) describes well-known apis as not being
```
networkException, line 126: > #### .well-known and other APIs that can't be versioned
>
> Best effort is appreciated. Typically these endpoints will be receiving minor behavioural changes or new fields. New fields should be appropriately prefixed, and behaviour changes should be rolled out cautiously by implementations (waiting until after FCP has concluded is probably best to ensure there's no major problems with the new behaviour).

Given we currently directly referenced a versioned federation API response, perhaps there should be a top level JSON key referencing that version?

PR #4059, opened 2023-10-14
[WIP] MSC4059: Mutable event content

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/mutable-events/proposals/4059-mutable-event-content.md)

### DIFF
--- a/proposals/4059-mutable-event-content.md
+++ b/proposals/4059-mutable-event-content.md
@@ -11,6 +11,8 @@ history for clients to use - this can make the feature unhelpful in many cases.
 
 ## Proposal
 
+In a future room version...
+
 **TODO**: Detail.
 
 Events must self-declare as mutable by *not* providing a content hash over
--- a/proposals/4059-mutable-event-content.md
+++ b/proposals/4059-mutable-event-content.md
@@ -1,0 +1,132 @@
+# MSC4059: Mutable event content
+
+[Edits](https://spec.matrix.org/v1.8/client-server-api/#event-replacements) are
+used to "replace" an event's contents, and are structured as events themselves.
+This can, in some circumstances, cause excessive event traffic which slows down
+scrollback.
+
+This proposal introduces a complimentary feature to edits which allows events to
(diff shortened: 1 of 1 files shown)


proposals/4059-mutable-event-content.md
```diff
@@ -0,0 +1,134 @@
...
+This makes the feature useful only when loss is tolerable. A server can always
+try to ask the origin for the latest copy of mutable events, but that server
+might not be online anymore. Other participating servers might not have the
+latest copy either.
+
+## Alternatives
+
+**TODO**: Detail.
+
+Just send edits, or regular state updates!?
```
andrewzhurov, line 121: Sending edits (deltas) seems like a good idea.
It may lower event's size substantially (given the difference between event's content and made changes is great, such as type corrections, which I'd expect to be the most used case). 
E.g., original text: "The wolrd is round" , changes `[del 6(l), del 7(r), add 5 r, add 6 r]` or `[cursor at 8 (r), del, del, r, l]`.
new text: "The world is round".
I.e., we could use a CRDT data model to represent edits.

What I like about this approach is how we have persistency. If you reply to one's event you _want_ it to point to exact content you referred to. Later on this content may evolve, but the pointer is at that particular version in history.
Makes us safe from subversion of semantics.
E.g., one posted "I love petting cats!" and you reply "Me too, would do it all day long!". Later original poster could change it to "I love torturing cats!", but your semantics are not subverted, as it points to the exact snapshot that's about petting.

andrewzhurov: We would not need a full-fledged CRDT if a message is edited only from one device, having it as a simple log of edits. But given you have multiple devices there may inevitably be a case where you make edits from both. E.g., it's a darn long-form message one works on for a couple of days from both your laptop and mobile (while enjoing comfort of whatever one's sitting on).

turt2live:
I don't even know where I was going with this...

PR #4060, opened 2023-10-14
MSC4060: Accept room rules before speaking

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/rules-prompt/proposals/4060-room-rules.md)

### DIFF
--- a/proposals/4060-room-rules.md
+++ b/proposals/4060-room-rules.md
@@ -1,0 +1,101 @@
+# MSC4060: Accept room rules before speaking
+
+A common feature among other chat platforms is a requirement for users to
+acknowledge that they've read the room's rules before they're allowed to send
+a message. It's typically a challenge to implement such a feature reliably as
+there's rarely anything preventing an automated client from accepting the rules
+and promptly spamming, however for the typical case a sufficient barrier is
+simply one which appears *most* of the time in front of users.
+
+This proposal introduces a feature where rooms can define rules that users must
+accept before they're allowed to speak. Perhaps controversially, this MSC does
+*not* require a new room version as it does not tie acceptance of the rules into
+the event authorization rules. This is a deliberate choice as a change in rules
+would consitute a requirement that all past agreements are rendered invalid,
+which may be a desired trait in some communities but not all. This is explored
+in a little bit more detail in the Alternatives section.
+
+## Proposal
+
+A new `m.room.rules` state event with empty string state key is defined to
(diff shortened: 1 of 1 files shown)


proposals/4060-room-rules.md
```diff
@@ -0,0 +1,101 @@
...
+
+1. [MSC4058](https://github.com/matrix-org/matrix-spec-proposals/pull/4058) -
+   the accepting user adds some structured metadata to the rules event.
+2. [MSC4059](https://github.com/matrix-org/matrix-spec-proposals/pull/4059) -
+   if changed to support mutations from unoriginal senders, mutating the event
+   content in some way. This is probably dangerous though.
+3. [Event relationships](https://spec.matrix.org/v1.8/client-server-api/#forming-relationships-between-events) -
+   like we've done for pretty much everything else.
+
+When `m.room.rules` is present in a room, and the user hasn't accepted them, the
```
uhoreg, line 50: Presumably, if a new `m.room.rules` event is sent, everyone will need to accept the new rules, even if they had previously accepted the old rules.  (Which means, among other things, that if a client sees a new `m.room.rules` state event, it should re-prompt the user to accept it.)

proposals/4060-room-rules.md
```diff
@@ -0,0 +1,101 @@
...
+
+**TODO**: Detail.
+
+## Alternatives
+
+This MSC deliberately does not tie acceptance of the rules into the event
+authorization algorithm. Doing so with the current power level structure in
+Matrix rooms could potentially lead to over-imposed opinions on what it means
+to "speak" in a room, as well as requiring users to frequently accept rule
+changes.
```
HarHarLinks, line 80: In my understanding this MSC is limited in this way to keep it small and simple?

As far as I see, there is a trade-off of initially not offering the customizability of
- "what it means to speak"
- how to deal with rule updates
- maybe something I am not seeing or perhaps I am missing some specific context this is trying to address

and the worse UX and community admin effort of (creating,) setting up, and maintaining a bot that handles these.

While I think the additional effort is clear, the "worse UX" could mean e.g.
- a bot redacts messages and posts a "please accept the rules" message -> spammy and disruptive to the rest of the community, especially if in a big community this happens constantly
- a bot manages the PL to disallow sending at all, while no user feedback is possible (composer will only say "you can't send messages" and not "please accept the rules")

Most of this will then later become obsolete when a follow-up MSCs expands the possibilities.

Is this trade-off worth it to not make a bigger throw from the start? I would like to see further explanation why.

PR #4069, opened 2023-10-23
MSC4069: Inhibit profile propagation

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/inhibit-profile/proposals/4069-inhibit-profile-propagation.md)

Implementations:
* Synapse (homeserver): https://github.com/matrix-org/synapse/pull/16636
* *Usage of API not yet implemented*

### DIFF
--- a/proposals/4069-inhibit-profile-propagation.md
+++ b/proposals/4069-inhibit-profile-propagation.md
@@ -1,4 +1,4 @@
-# MSC0000: Inhibit profile propagation
+# MSC4069: Inhibit profile propagation
 
 The current [`PUT /avatar_url`](https://spec.matrix.org/v1.8/client-server-api/#put_matrixclientv3profileuseridavatar_url)
 and [`PUT /displayname`](https://spec.matrix.org/v1.8/client-server-api/#put_matrixclientv3profileuseriddisplayname)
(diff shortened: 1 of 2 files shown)


proposals/4069-inhibit-profile-propagation.md
```diff
@@ -0,0 +1,85 @@
...
+intelligent rate limits to state event changes. For example, detecting that the user is updating their
+profile information in a subset of rooms shortly after updating their global profile.
+
+## Potential issues
+
+This adds a bit of API bloat for a relatively uncommon feature, though may enable other features to
+be more readily possible. For example, a client *introducing* per-space memberships might use this
+new query parameter to make it easier to manually apply `m.room.member` event changes.
+
+## Alternatives
```
tezlm, line 56: Instead of propagating at `/avatar_url` and `/displayname`, it could be part of `m.room.member`. A `"locked": true` (or similar) in the content would cause the server to not update that member event.

turt2live: so, I've been thinking about this a lot. There's some potential privacy concerns with revealing the information. Specifically, it tells other members of the room that the user is *definitely* using a customized, non-default, profile in the room. On the other hand, any malicious party wanting to make use of that information could just GET the user's default profile to find their "real" profile anyways. A separate MSC would be good to add protections/guarantees around the user's [default] profile (who can access it, etc).

Eventually it'd also be good to move the profile *out* of the `m.room.member` event, to give at minimum separation of concerns. Currently the membership event's primary purpose is to denote the user's, well, membership. Profile information is added on as a convenience. While scalable, it feels a bit overloaded to use the event for two things.

The third concern is mostly one of backwards compatibility. Clients may already have UX which says the user's profile changes *everywhere*. If we add `locked: true` to the membership event, that client UX is now a lie. This is relatively minor, but worth noting as an incompatibility. By preventing propagation at the endpoint level, existing clients can maintain their UX and new clients can more obviously add a toggle or similar to "overwrite" the profile in all rooms, or otherwise tailor their UI for their specific use case.

Overall, I think I'm still in favour of using the endpoint-level control, though a counter MSC to discuss the `m.room.member` flagging would be interesting.

olivia-fl: A possibly third option: `m.profile_locked` as a per-room account data event. This state would not be visible to other users. I would prefer this to a `locked` property in the membership event, but would prefer endpoint-level control over both.

cloudrac3r (changes requested):
I think this feature should work on the DELETE endpoints as well as PUT, but right now the MSC only says PUT. Want to change it?

turt2live:
https://github.com/matrix-org/matrix-spec-proposals/pull/4466 has gotten more traction - closing in favour of that.

PR #4071, opened 2023-10-29
[WIP] MSCXXXX: Pagination Token Headers

Fizzadar:


proposals/4071-pagination-token-headers.md
```diff
@@ -0,0 +1,56 @@
...
+If a client makes a request specifying both header and query string servers should reject the request
+with a 400 response and `M_INVALID_PARAM` error code.
+
+## Potential issues
+
+As noted above, this is a hack to work around CORS restrictions. Unfortunately CORS isn't going to
+change any time soon (if ever). Whether the advantages outweigh the "ugliness" of this hack is up 
+for discussion here.
+
+## Alternatives
```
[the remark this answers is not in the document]
Fizzadar: Ha! The one term I didn't search for, `preflight` :) Linked in alternatives and expanded a bit on things. Essentially I see this MSC as a temporary workaround because it's dead simple to implement, and long run one of the alternatives are the real fix.

proposals/4071-pagination-token-headers.md
```diff
@@ -0,0 +1,66 @@
...
+
+|Endpoint (link to current spec)|Query Parameter|Header Alternative|
+|-|-|-|
+|[`sync`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3sync)|`since`|`X-Matrix-Pagination-Since`|
+|[`messages`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomidmessages)|`from`|`X-Matrix-Pagination-From`|
+|[`messages`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomidmessages)|`to`|`X-Matrix-Pagination-To`|
+
+Servers must treat the values exactly the same as the query string versions.
+
+If a client makes a request specifying both header and query string servers should reject the request
```
MTRNord, line 30: Hm I wonder about backwards compatiblity here. If the server doesn't support these headers yet how can a client know about this? Imho it makes sense to allow both to be sent as long as they are identical.

I know this kinda circumvents the point of the proposal however a client could still choose to only send one.

MTRNord: An alternative could be to just have it somewhere in I guess the versions endpoint denoted that a server supports this. Clients request that endpoint usually anyway.

MTRNord (approved):
Apart from my one comment I believe this is a much needed change :)

proposals/4071-pagination-token-headers.md
Fizzadar: Yeah I was thinking this would be indicated to clients via the versions endpoint (but then only included the unstable flag).

Good point about making the switchover, I think specifying both as long as they're the same is reasonable and avoids having to figure out detection.

proposals/4071-pagination-token-headers.md
```diff
@@ -0,0 +1,66 @@
...
+is absolutely a hack, and if accepted must only be used to work around this specific problem.
+
+Note that there are a bunch of less hacky [alternatives](#alternatives) but none of them are simple
+and quick to implement such as this. This is very much a patch to alleviate the immediate issue while
+better solutions can be worked out.
+
+## Proposal
+
+Super simple - clients may pass pagination tokens as a header instead of query parameters using the
+format `X-Matrix-Pagination-$PARAM`, the following table lists the available headers by endpoint:
```
uhoreg, line 20: The `X-` prefix is deprecated: https://www.rfc-editor.org/rfc/rfc6648  The `Matrix-` prefix should be enough to ensure that it doesn't collide with any other HTTP headers.

PR #4072, opened 2023-10-30
MSC4072: Handling devices with no one-time keys in `/keys/claim`

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/dead_device_key_claims/proposals/4072-exhausted_devices_in_key_claim_response.md)


richvdh:
I eventually decided this was easier to do on the client (https://github.com/matrix-org/matrix-rust-sdk/pull/2805)

PR #2666, opened 2020-07-05
MSC2666: Get rooms in common with another user

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/hs/shared-rooms/proposals/2666-get-rooms-in-common.md)

Author: @Half-Shot, @ShadowJonathan 

Implementations:
- Synapse: https://github.com/matrix-org/synapse/pull/15621 + https://github.com/element-hq/synapse/pull/19279
- Ruma: https://github.com/ruma/ruma/pull/1547
- JS SDK: https://github.com/matrix-org/matrix-js-sdk/pull/3381

<details><summary>Click to reveal Legacy Implementations</summary>

- matrix-react-sdk PR: matrix-org/matrix-react-sdk#4897
- matrix-js-sdk PR: matrix-org/matrix-js-sdk#1417
- Synapse (Went through a few iterations)
  - https://github.com/matrix-org/synapse/pull/7785
  - https://github.com/matrix-org/synapse/pull/12036
  - https://github.com/matrix-org/synapse/pull/12038
  - https://github.com/matrix-org/synapse/pull/12836
</details>

---

# Spec PR note

The endpoint path was clarified in https://github.com/matrix-org/matrix-spec-proposals/commit/9a49a4779d2f467ce853df296c16b65b4ddc0b8a

---

SCT stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/2666#issuecomment-3614930041)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/2666#issuecomment-3614713505)

PR #4073, opened 2023-11-07
MSC4073: Shepherd teams

joepie91:
An attempt at addressing the problems with SCT throughput.

### DIFF
--- a/proposals/4073-shepherd-teams.md
+++ b/proposals/4073-shepherd-teams.md
@@ -1,0 +1,64 @@
+# MSC4073: Shepherd teams
+
+Matrix currently has significant problems with its specification change process. There are several common complaints from contributors:
+
+1. "Nobody responds to my MSC"
+2. "It takes forever to get an MSC passed"
+3. "It's not clear what the status of my MSC is, or what I am supposed to do next"
+
+From conversations with various Spec Core Team (SCT) members, all of these have (at least) one identical root cause: the SCT does not have enough bandwidth to manage, review, and approve all of the submitted MSCs. This causes communication to become minimal, and for MSCs considered "less critical" to languish and eventually become abandoned by their authors, who often give up on Matrix protocol development entirely due to their bad experiences. The SCT cannot be bypassed, because its final approval is required for any MSC to be considered mergeable.
+
+## The current process
+
+In the current MSC process, it is the responsibility of the SCT to manage the discussion and direction of every MSC, and ultimately give its final approval before a merge is possible. A shepherd may be assigned to guide the MSC process independently of the SCT - but they ultimately only serve the role of a *guide*, and do not hold any authority over the final decision to merge or reject. A shepherd is also not required, and frequently is not assigned. Altogether, this often leaves an impression with contributors of "appealing to a disinterested SCT".
+
+## Proposal
+
+This MSC aims to solve these problems, or at least make a first significant step towards doing so, by changing the scope of responsibility of the SCT. It introduces the concept of a "shepherd *team*", and makes them primarily responsible for the MSC's progress. This team is assigned on a *per-MSC* basis, and it is modelled after [the "shepherd teams"](https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md) in [the NixOS RFC process](https://github.com/NixOS/rfcs#readme).
+
+The new process will work as follows:
+
+1. Community members submit nominations for shepherds, in the MSC's discussion thread. The nominees should be people who are closely familiar with the subject matter of the MSC, but not be the author. Community members may also nominate themselves.
+2. The SCT selects multiple nominees to become the "shepherd team"; ideally 3 or 4 people.
+	- The SCT may likewise nominate candidates (eg. if there are too few nominees), but should leave the opportunity for other community members to object to the SCT's nominations.
+	- No more than half of the shepherds may be SCT members.
+	- The shepherd selection should avoid conflicts of interest, and be a diverse representation of perspectives on the topic (within the boundaries of the Code of Conduct).
+3. The shepherd team will be responsible for guiding the discussion on the proposal. This discussion should incrementally improve the proposal until outstanding concerns have been resolved. They may request administrative assistance from the SCT if eg. moderator intervention is required.
+4. Once the shepherd team feels that all concerns have been resolved satisfactorily, and that there is no meaningul refinement left to be done, they can carry out a final technical review, and call FCP (Final Comment Period).
+	- This requires unanimous agreement from all shepherds.
+	- The SCT must be informed immediately by the shepherds when the FCP starts, so that the SCT can make a public 'last call' for feedback.
+	- The FCP process itself remains as it was before.
+5. If FCP passes without objection, the shepherds declare the MSC accepted, and notify the SCT. The SCT will then merge the proposal.
+
+Some of the most important takeaways from this:
+
+- The SCT is no longer the final reviewer and approver; this responsibility is delegated to the shepherd team for that MSC.
+- The shepherd team is trusted to ensure that:
+	1. the proposal represents diverse interests, and
+	2. the proposal does not conflict with other proposals, coordinating with shepherds for other MSCs if necessary.
+- The SCT no longer concerns itself with the *content* of MSCs, but only with supervising the process.
+- The SCT may still intervene where necessary to resolve administrative disputes, or replace shepherds if they become unavailable.
+
+This is a significant step towards a more community-driven protocol, by putting more of the specification process in the hands of community members, and more crucially, it frees up significant capacity from the SCT.
+
+This proposal does not currently encompass the question of "who writes the final specification text?" - that is left unchanged from the current situation, and may be clarified or changed in a separate MSC.
+
+## Anticipated outcomes
+
+Primarily, this should unblock the many MSCs that some community members feel strongly about, but that the SCT has not had the capacity to advance, eg. because it accommodates a relatively specialized need. This should overall significantly improve on the throughput and responsiveness of the MSC process, much like it has already done for NixOS.
+
+Indirectly, this would improve community trust and engagement regarding protocol development, because contributors would no longer feel ignored, and they can continue building additional work on top of previous work - where that previously would have been blocked on a 'stuck' MSC, and the uncertainty of it ever getting merged.
+
+Another anticipated indirect effect is increased transparency in the specification design process; it is currently an issue that a lot of specification design happens 'internally', behind closed doors, and this can currently be gotten away with due to every MSC going through a central point of review (the SCT) before merging, where any incompatibilities can be ironed out.
+
+In the 'shepherd team' model, this is no longer possible, as one cannot expect 'internal' and undocumented decisions to be taken into account by other shepherd teams in their final review, and so all such specification development *must* happen out in the open (as is ultimately desirable for an open protocol).
+
+## Security concerns
+
+It is *technically* possible for an organized effort to disrupt the specification by passing something without SCT review. However, this has not been a problem for NixOS at all, and would require compromising the *entire* shepherd team. Additionally, administrative intervention from the SCT remains possible in exceptional cases (like when a shepherd has become unavailable), and this could likewise be applied to such cases.
+
+## Alternative approaches
+
+- __Scaling up the SCT:__ Organizations become more unwieldy as they try to scale up monolithically. Additionally, this requires more full-time commitments than anyone currently seems to be willing and able to fund, and would bring conflict-of-interest concerns with it. This does not seem viable.
+- __Doing nothing:__ Increasing amounts of potential contributors are bouncing off Matrix due to the problematic MSC process, and it is well-documented that there are currently very few external contributions. This would likely lead to a slow death of Matrix as an open protocol.
+- __Bigger MSC scopes:__ This would reduce the amount of units for the SCT to review and manage, but make changesets so large that the discussion of the implementation details would likely exponentially explode, making the problem worse on another axis.


turt2live:
@joepie91 when you get a chance, please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes for this MSC to be eligible for FCP acceptance. Also, lines should be limited to approximately 120 characters.

proposals/4073-shepherd-teams.md
turt2live, line 1: Process-wise, I think it's actually best if we sit on this and think about it for a while (ironically). The MSC is largely written in response to [recent events](https://matrix.org/blog/2023/11/06/future-of-synapse-dendrite/), and we should be cognisant that those events may be clouding judgement. Similarly, we should not rush into decisions when it comes to process: they need careful thought and consideration, even if appearing simple on the surface.

joepie91:
> please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes for this MSC to be eligible for FCP acceptance

Completed privately.

> Also, lines should be limited to approximately 120 characters.

Fixed.

proposals/4073-shepherd-teams.md
joepie91: > The MSC is largely written in response to [recent events](https://matrix.org/blog/2023/11/06/future-of-synapse-dendrite/)

That is not quite accurate; none of these issues are new, and all of them have been brought up in a more informal manner many times, including suggestions to have the SCT do less. What *has* changed recently, are two things:

1. The Foundation is now showing itself more receptive towards governance changes than before, and
2. The situation has worsened to a degree that the Foundation itself expresses public concern about a lack of third-party contributions, and there is a lot of unrest in the broader Matrix community about what this means for the future of Matrix, for which there need to be concrete answers.

So rather than this being a sudden novel idea, it's an idea that has been floating around in various form for quite some time already, and has already proven its worth in another community (which has actually partly modeled its spec process *on* that of Matrix, to my knowledge), and it is now at a point where concrete solutions need to start appearing, so as to not to lose the remaining trust of the community.

> Similarly, we should not rush into decisions when it comes to process: they need careful thought and consideration, even if appearing simple on the surface.

Of course, careful consideration is important, that is part of why I submitted this as an MSC rather than trying to backchannel it. But this does not need to translate into a long approval process in terms of *time*, and it's not clear to me in what way this proposal would be expected improve by letting it sit, in and of itself.

In this consideration, given the circumstances and the worsening community trust issues, I also think it is important to ask: what possible worst-case scenarios *are* there for the outcome of this proposal, that are worse than the outcome of *not* changing the process?

Because governance change proposals do not need to be perfect to be useful, they just need to be an improvement over what we currently have, so if it turns out that something was overlooked, this can always be rectified down the line (especially since the SCT still retains the power to intervene if really necessary).

InezMc:
Private signoff has been recieved.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+	- The SCT may likewise nominate candidates (eg. if there are too few nominees), but should leave the opportunity
+	  for other community members to object to the SCT's nominations.
+	- No more than half of the shepherds may be SCT members.
+	- The shepherd selection should avoid conflicts of interest, and be a diverse representation of perspectives on
+	  the topic (within the boundaries of the Code of Conduct).
+3. The shepherd team will be responsible for guiding the discussion on the proposal. This discussion should
+   incrementally improve the proposal until outstanding concerns have been resolved. They may request administrative
+   assistance from the SCT if eg. moderator intervention is required.
+4. Once the shepherd team feels that all concerns have been resolved satisfactorily, and that there is no meaningful
+   refinement left to be done, they can carry out a final technical review, and call FCP (Final Comment Period).
```
clokep, line 47: It is unclear to me from this text what the role of the original author is at this point; I see two options:

1. The shepherd team makes suggestions and collaborates with the original author to improve the MSC until it is ready for FCP.
2. The author relinquishes all control and the shepherd team essentially becomes the authors.

I'm not sure which is meant by the above text, both have pros and cons. 1 is closer to what we do today, while 2 is closer to some other processes (IETF, maybe?)

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+   nominate themselves.
+2. The SCT selects multiple nominees to become the "shepherd team"; ideally 3 or 4 people.
+	- The SCT may likewise nominate candidates (eg. if there are too few nominees), but should leave the opportunity
+	  for other community members to object to the SCT's nominations.
+	- No more than half of the shepherds may be SCT members.
+	- The shepherd selection should avoid conflicts of interest, and be a diverse representation of perspectives on
+	  the topic (within the boundaries of the Code of Conduct).
+3. The shepherd team will be responsible for guiding the discussion on the proposal. This discussion should
+   incrementally improve the proposal until outstanding concerns have been resolved. They may request administrative
+   assistance from the SCT if eg. moderator intervention is required.
```
clokep, line 45: Due to the use of GitHub the shepherds probably would need to depend on the SCT to resolve comment threads and such. This is a bit annoying, but hopefully not a huge deal.

They might also need to rely on them to modify the text (via merging PRs?) or getting access to the original source repository to modify the pull request.

proposals/4073-shepherd-teams.md
joepie91: Something closer to 1, but not exactly the same; the author remains primarily responsible for improving the proposal in response to review comments or questions (from anyone, not just the shepherd team!), with the shepherd team overseeing this process, and essentially supervising this improvement process.

So a (non-exhaustive) example of the task division would be:
- __Any community member:__ Leave review comments, raise concerns, make suggests for improvements
- __The author(s):__ Update the MSC text in response to such comments, answer clarification questions about the proposal
- __The shepherd team:__ Keep an eye on proceedings, encourage people to raise concerns that they have, potentially seek out people who may be interested in doing review, periodically poll the author for the status of the MSC if it is unclear, call FCP, other per-MSC administrative tasks
- __The SCT:__ Oversee the entirety of the spec process, the work that the various shepherd teams are doing, assign and/or replace shepherd team members, intervene administratively in exceptional cases, notify shepherd teams if they think that collaboration between the teams of different MSCs is necessary

Optionally, the author(s) and the shepherd team may agree to transfer editing responsibilities to the shepherd team, if eg. the original author(s) can no longer commit to working on the proposal, but it is still desirable for it to move ahead.

In practice, I don't believe this changes anything for the author compared to the current process; it only affects (reduces) the responsibilities that the SCT has, and as a result, much of the communication and coordination that would have previously been between the author and the SCT, would now be between the author and the shepherd team for their MSC.

(I can make all of this more explicit in the MSC itself, if desired; so far I've kept the proposal limited to describing only the things that *change* from the current process.)

proposals/4073-shepherd-teams.md
uhoreg: Regarding ability to edit the MSCs, it might be possible to just add shepherds as maintainers to this repository, and make sure that MSCs have the "Allow edits and access to secrets by maintainers " setting enabled.  Presumably, shepherds should be able to be trusted to behave properly and not abuse their privileges (e.g. by interfering with other MSCs).

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+
+This MSC aims to solve these problems, or at least make a first significant step towards doing so, by changing the
+scope of responsibility of the SCT. It introduces the concept of a "shepherd *team*", and makes them primarily
+responsible for the MSC's progress. This team is assigned on a *per-MSC* basis, and it is modelled after
+[the "shepherd teams"](https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md) in
+[the NixOS RFC process](https://github.com/NixOS/rfcs#readme).
+
+The new process will work as follows:
+
+1. Community members submit nominations for shepherds, in the MSC's discussion thread. The nominees should be people
```
uhoreg, line 34: Obviously, this will depend on peoples' willingness to be shepherds, so we'll need to see how much involvement we can get.  I suspect that the community is at a point where we should be able to get shepherds in at least most subject areas.  Some other areas, I'm not so sure about.  We may need to find out how many people there are who would be willing to be shepherds, in what areas, and how much time they can commit to.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Proposal
+
+This MSC aims to solve these problems, or at least make a first significant step towards doing so, by changing the
+scope of responsibility of the SCT. It introduces the concept of a "shepherd *team*", and makes them primarily
+responsible for the MSC's progress. This team is assigned on a *per-MSC* basis, and it is modelled after
+[the "shepherd teams"](https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md) in
+[the NixOS RFC process](https://github.com/NixOS/rfcs#readme).
+
+The new process will work as follows:
```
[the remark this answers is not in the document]
joepie91: That is a fair point; I'm not opposed to adding such a shortcut, but I do think that it should be a well-defined one, such that that path is only taken when there's reasonable certainty that it *can* be passed quickly. My main concern is that if there isn't such a defined threshold, it can become appealing for SCT members to take on more work than they can realistically handle, out of excitement/optimism.

I'm thinking something along the lines of:

> If the SCT believes that the proposal can be passed within 3 weeks, then the SCT itself may function as the shepherd team for that MSC, skipping the shepherd assignment process. If, against expectations, after 3 weeks the proposal has not yet been merged, the MSC is no longer eligible for this accelerated treatment, and the normal shepherd assignment process must be followed belatedly.

... but I'm open to other suggestions as well.

proposals/4073-shepherd-teams.md
joepie91: That's a point I hadn't considered yet. I think that's a viable option, but it's also going to depend on whether all the current SCT members feel comfortable with that approach.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+
+1. Community members submit nominations for shepherds, in the MSC's discussion thread. The nominees should be people
+   who are closely familiar with the subject matter of the MSC, but not be the author. Community members may also
+   nominate themselves.
+2. The SCT selects multiple nominees to become the "shepherd team"; ideally 3 or 4 people.
+	- The SCT may likewise nominate candidates (eg. if there are too few nominees), but should leave the opportunity
+	  for other community members to object to the SCT's nominations.
+	- No more than half of the shepherds may be SCT members.
+	- The shepherd selection should avoid conflicts of interest, and be a diverse representation of perspectives on
```
uhoreg, line 41: This is kind of vague, which isn't necessarily a bad thing.  But I wanted to note that different people may have different opinions on what is acceptable for a "diverse representation of perspectives".  For example, would it be acceptable if everyone on the shepherd team is an Element employee, if they represent different subject areas (e.g. server, client, crypto)?  Obviously, I would recommend avoiding such a situation, but it could happen if, e.g. nobody else has the time to be a shepherd.

I don't want to try to set out a list of hard-and-fast rules for what is and isn't acceptable.  But maybe the SCT could come up with, and publish, a list of guidelines that it tries to follow when choosing the shepherd team.

uhoreg (commented):
Thanks for writing this up.  In general, I think this is an idea worth exploring.

proposals/4073-shepherd-teams.md
clokep: > Something closer to 1, but not exactly the same; the author remains primarily responsible for improving the proposal in response to review comments or questions (from anyone, not just the shepherd team!), with the shepherd team overseeing this process, and essentially supervising this improvement process.

I was pretty sure this is what was meant, but wanted to clarify. Thanks! 👍 I would summarize this almost as the SCT delegating responsibility of a particular MSC (or group of MSCs) to another set of people. Sounds reasonable to me.

proposals/4073-shepherd-teams.md
clokep: > I assume this comment is under the assumption of the second option in https://github.com/matrix-org/matrix-spec-proposals/pull/4073#discussion_r1384887804?

Only somewhat -- if the author is unresponsive or if the shepherds are deciding what to do with an MSC, they seem like they would need *some* editorial power to resolve threads, etc.

Just to be upfront -- I don't think the technology not being perfect here is something that should block this proposal, just wanted to point out a particular sticking point.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+	  the topic (within the boundaries of the Code of Conduct).
+3. The shepherd team will be responsible for guiding the discussion on the proposal. This discussion should
+   incrementally improve the proposal until outstanding concerns have been resolved. They may request administrative
+   assistance from the SCT if eg. moderator intervention is required.
+4. Once the shepherd team feels that all concerns have been resolved satisfactorily, and that there is no meaningful
```
clokep, line 46: This seems to assume that the shepherd team will do an FCP-merge call, what if they feel it is unsalvageable and should be rejected? I suspect it is fine, but not really called out.

proposals/4073-shepherd-teams.md
joepie91: > For example, would it be acceptable if everyone on the shepherd team is an Element employee, if they represent different subject areas (e.g. server, client, crypto)? Obviously, I would recommend avoiding such a situation, but it could happen if, e.g. nobody else has the time to be a shepherd.

Right. I think all of these are important points; organizational diversity and variation in expertise are both important for this. But I think it should be broader than that; for example, if there is a group of people who want to push enthusiastically for a specific feature or spec change, then even if they are not formally one organization, you probably wouldn't want the shepherd team to consist *just* of those people, but rather also involve someone with a different view on the issue.

So a few (non-exhaustive) examples of undesirable situations would be:
- All shepherds are Element employees
- All shepherds are client-only developers
- All shepherds *really* want the proposal to be approved, but others in the community may not want that

At the same time, I recognize that this can be a difficult bar to meet, especially early on when we don't have a lot of potential candidates yet (as regaining trust in the spec process will take time for people). So this is defined somewhat loosely, to give the SCT room to make a judgment call based on the available nominees and the specific situation.

This means that there is an implicit assumption that the SCT itself acts impartially, but given that the SCT currently has essentially total authority over the spec process, that is already the case and this wouldn't be a *new* assumption, so it is one that I feel is acceptable to retain in this proposal. If that causes problems in practice, I think it can be revisited later in a follow-up MSC.

> But maybe the SCT could come up with, and publish, a list of guidelines that it tries to follow when choosing the shepherd team.

I think that would be a good approach. I do think that would probably be out of scope for this MSC, and rather "just" be an SCT policy matter, in the same way as the checklist currently is.

proposals/4073-shepherd-teams.md
joepie91: > Only somewhat -- if the author is unresponsive or if the shepherds are deciding what to do with an MSC, they seem like they would need _some_ editorial power to resolve threads, etc.

Right. I'd be inclined to classify this as an exceptional case requiring administrative intervention; if the author is unresponsive, and no explicit transfer of editing responsibility has taken place, then (assuming shepherds do not straight-up get access to the spec repository) it would be up to the shepherd team to request administrative intervention from the SCT.

As long as this doesn't happen frequently, it sounds sufficiently workable to me as a solution. And if it *does* happen frequently, and authors frequently abandon MSCs, then that is IMO a bug in the specification process that needs to be addressed - because we shouldn't be seeing that sort of abandonment rate in a healthy spec process.

proposals/4073-shepherd-teams.md
joepie91: Right, in that case it is also the shepherd team's responsibility to reject the MSC, and notify the TSC of such. I will make that more explicit.

proposals/4073-shepherd-teams.md
joepie91: I suspect that initially, it will be difficult to find enough shepherds, because a lot of people have already (halfway) given up on the spec process due to past experiences, and it will take time to 'rebuild trust' on that point.

Eventually, I expect this problem to resolve itself; it is already common for implementers to involve small communities of people for review, and for people outside of that community to have critical feedback in more informal ways. I think that a lot of that feedback and motivation can be moved into shepherd roles in the long term.

Given that I expect a somewhat rocky start, it may be a good idea to deal with the *existing* MSCs by slowly and gradually moving them over to the shepherd model, to get them out of the weeds. There are a number of MSCs I can think of that a lot of people are passionate about, and that have also raised criticisms, and which would probably be good candidates to try out the shepherd process on to start with. If this goes well, that can be expanded to other existing MSCs.

How to deal with existing MSCs should probably be part of the proposal too... I will add that soon as well.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,109 @@
...
+This MSC aims to solve these problems, or at least make a first significant step towards doing so, by changing the
+scope of responsibility of the SCT. It introduces the concept of a "shepherd *team*", and makes them primarily
+responsible for the MSC's progress. This team is assigned on a *per-MSC* basis, and it is modelled after
+[the "shepherd teams"](https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md) in
+[the NixOS RFC process](https://github.com/NixOS/rfcs#readme).
+
+The new process will work as follows:
+
+1. Community members submit nominations for shepherds, in the MSC's discussion thread. The nominees should be people
+   who are closely familiar with the subject matter of the MSC, but not be the author. Community members may also
```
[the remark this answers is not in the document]
joepie91: Thinking through it a bit, I'm not certain that that approach would be ideal. The final state of an MSC does not necessarily look much like its initial draft, depending on how much refinement and feedback it has gone through, and especially for more complex design problems, it is very possible that what initially seemed to be the worse solution, actually ends up being the optimal one after a few rounds of refinement.

Given that, I think it would actually be beneficial for conflicting proposals to proceed independently, with the understanding that coordination between the shepherd teams is needed before any FCP is called; so that each proposal can independently develop, and the comparison ends up being between the *refined* proposals, rather than between drafts. In the current model, this would place a significant burden on the SCT, but in a shepherd-driven model, that would not be a concern; they would be separate teams, like with any other two unrelated MSCs.

Eventually, the author may decide to withdraw their proposal because *they* feel that the alternative is better (in which case the relevant shepherd team is disbanded), or when both proposals have reached a near-FCP stage, the shepherd teams might in their coordination decide that only one of them is worth moving ahead with, or even decide that a merge between the two should take place (in which case the SCT could reorganize the shepherd team with an exemption from the usual no-authors rule). That would be dependent on the specific case.

I feel like this is more a matter of "providing good guidance to shepherd teams, for coordinating this stuff", than a matter of having a strict written no-conflict policy.

proposals/4073-shepherd-teams.md
joepie91: Relatedly: clear guidance will be necessary for shepherds, and how to conduct their part of the job. This is more of a note-to-self, in that I should add this to the proposal.

proposals/4073-shepherd-teams.md
joepie91: > I would summarize this almost as the SCT delegating responsibility of a particular MSC (or group of MSCs) to another set of people.

That sounds about right to me :)

proposals/4073-shepherd-teams.md
joepie91: This has been clarified now.

proposals/4073-shepherd-teams.md
joepie91: This has been clarified now - I've added an "SCT tasks" section that includes this.

proposals/4073-shepherd-teams.md
joepie91: (Both of those points have now been added to the proposal.)

proposals/4073-shepherd-teams.md
clokep: LGTM, thanks.

proposals/4073-shepherd-teams.md
```diff
@@ -0,0 +1,141 @@
...
+2. The SCT selects multiple nominees to become the "shepherd team"; ideally 3 or 4 people.
+	- The SCT may likewise nominate candidates (eg. if there are too few nominees), but should leave the opportunity
+	  for other community members to object to the SCT's nominations.
+	- No more than half of the shepherds may be SCT members.
+	- The shepherd selection should avoid conflicts of interest, and be a diverse representation of perspectives on
+	  the topic (within the boundaries of the Code of Conduct).
+3. The shepherd team will be responsible for guiding the discussion on the proposal. This discussion should
+   incrementally improve the proposal until outstanding concerns have been resolved. They may request administrative
+   assistance from the SCT if eg. moderator intervention is required.
+4. If the MSC proposes to standardize functionality that is already informally supported in one or more protocol
```
joepie91, line 46: Note: I have belatedly added this section to explicitly call out that shepherds are responsible for seeking review from existing implementers; I believe that this is currently a task that the SCT does, but it requires insight into the details of pending proposals that the SCT may no longer (always) have under this proposal.

proposals/4073-shepherd-teams.md
uhoreg: (Note to self: hitting the "resolve conversation" button doesn't submit the comment.)

Yes, I think that the "tasks" section sufficiently covers this concern.

proposals/4073-shepherd-teams.md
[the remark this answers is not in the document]
uhoreg: Yeah, it may be better for conflicting MSCs to proceed independently.  I think we should just leave this open as something to consider.

proposals/4073-shepherd-teams.md
ShadowJonathan: Speaking from my experience of being a shepherd (on https://github.com/matrix-org/matrix-spec-proposals/pull/2666), and from a past idea of mine; This problem can be fixed by introducing a bot that can answer to commands like `/res`, `/edit`, or other such functions.

This may increase noise, but it would at least allow the shepherd some control over otherwise locked functions of a shepherded PR.

One issue I also had was the inability to push commits to the PR branch; this can be solved by instructing the bot to push commits onto the PR branch, targeting some branch the user is currently working on. (Force pushes would be disallowed, and only fast-forward merges/tracking would work)

PR #4074, opened 2023-11-07
Server side annotation aggregation

dihydrogenmonoxide16:
Proposal describes the way how clients and servers could interact in a more efficient and scalable way working with annotation (reaction) - heavy rooms.

rendered (TBD)

### DIFF
--- a/proposals/4042-server-side-annotations-aggregation.md
+++ b/proposals/4042-server-side-annotations-aggregation.md
@@ -1,0 +1,149 @@
+# MSC4042: Server side annotations aggregation
+Currently, the specification for [`m.annotation aggregation`](https://spec.matrix.org/v1.8/client-server-api/#server-side-aggregation-of-mannotation-relationships) says that such relations should not be aggregated
+on the server side.
+
+This requires servers to deliver clients all annotations they receive. In most cases clients need just a number of annotations of every type. Therefore, delivering aggregated annotations instead of single events could dramatically reduce server-client traffic in
+rooms with many reactions.
+
+## Background
+
+Aggregation is a process when client and/or server "summarises" together many events of the same type. 
+
+In early specification versions there was only client-side aggregation and client was solely responsible for receiving and
+summarizing all events to make a compact "view" of them.
+
+Later server side aggregation was introduced with idea to:
+- decrease client-server traffic for cases when clients do not need
+all events of the same type, but can know only their count
+- deliver summary of associated children events together with their
+parent event given that they can be significantly separated in timeline
+
+Given 1000 users reacted with "👍" annotation to event A,
+it is normally sufficient for clients to know that there were 1000
+reactions of that type "👍" to event A.
+
+If clients need to get the exact list of users who reacted and additional information about these 1000 reactions, clients can use
+existing [`relations (relationships) API`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api)
+
+Servers tried to provide clients with server side generated annotation aggregates before, but these attempts were not successful mainly because both sides (server and client) tried to
+aggregate and the same time. That used to lead to the situation when
+given the potentially outdated aggregate, clients could not understand which events had already been included into the server provided aggregate and which - not.
+
+## Proposal
+
+Clients will not be responsible for aggregation of relations of "m.annotation" relation type anymore. Such relations will always be aggregated by the server.
+
+Client API will always deliver annotations aggregates, including both
+local and known federation events. Homeserver will be responsible for counting annotations correctly.
+
+Whenever requested, server will always provide up-to-date (or nearly up-to-date) aggregates of "m.annotation" via [`messages`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomidmessages),
+[`relations`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api), [`get event by id`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomideventeventid) and similar client api endpoints.
+
+For server-server API, homeserver should provide aggregates, including only local events (events created and signed by self).
+
+Servers are allowed to limit maximum number of different "m.annotation" keys they aggregate to a single event by a reasonable value (that corresponds to a limited number of different reaction kinds, e.g. "👍", "👎"). This is needed to avoid the situation when malicious users may attack server creating jumbo-events (events with arbitrary high count of different reaction kinds). Servers may configure this threshold on their discretion, but
+this number should not be lower than 16. When annotation key limit
+is reached, no new keys should be added to the aggregate, but older
+keys would continue to be aggregated (change their count). Annotations beyond that threshold should still be available via [`relations`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api) endpoints.
+
+## Aggregate formats
+
+Full aggregate format (with parent ClientEvent/PDU):
+```
+{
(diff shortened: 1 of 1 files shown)


turt2live:
@dihydrogenmonoxide16 when you get a chance, please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes for this MSC to be eligible for FCP acceptance. Also, lines should be limited to approximately 120 characters.

dihydrogenmonoxide16:
> @dihydrogenmonoxide16 when you get a chance, please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on the changes for this MSC to be eligible for FCP acceptance. Also, lines should be limited to approximately 120 characters.

addressed

dihydrogenmonoxide16:
@turt2live , I added changes to make the behavior backwards-compatible with older clients as we discussed before

proposals/4074-server-side-annotations-aggregation.md
```diff
@@ -0,0 +1,270 @@
...
+reactions of that type "👍" to event A.
+
+If clients need to get the exact list of users who reacted and
+additional information about these 1000 reactions, clients can use
+existing [`relations (relationships) API`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api)
+
+Servers tried to provide clients with server side generated
+annotation aggregates before, but these attempts were not successful
+mainly because both sides (server and client) tried to
+aggregate and the same time. That used to lead to the situation when
```
turt2live, line 38: and *what* at the same time?

proposals/4074-server-side-annotations-aggregation.md
```diff
@@ -0,0 +1,270 @@
...
+given the potentially outdated aggregate, clients could not
+understand which events had already been included into the server
+provided aggregate and which - not.
+
+## Proposal
+
+Clients will not be responsible for aggregation of relations of
+"m.annotation" relation type anymore. Such relations will always be
+aggregated by the server. The only exception to this rule is E2E
+encrypted events, which should be solely aggregated by the client.
```
turt2live, line 48: Does the encryption condition only apply once [reactions are encrypted](https://github.com/matrix-org/matrix-spec/issues/660)?

proposals/4074-server-side-annotations-aggregation.md
```diff
@@ -0,0 +1,270 @@
...
+local and known federation events. Homeserver will be responsible
+for counting annotations correctly.
+
+Whenever requested, server will always provide up-to-date (or nearly
+up-to-date) aggregates of "m.annotation" via [`messages`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomidmessages),
+[`relations`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api), [`get event by id`](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3roomsroomideventeventid) and
+similar client api endpoints.
+
+For server-server API, homeserver should provide aggregates,
+including only local events (events created and signed by self).
```
turt2live, line 65: I'm not following this bit. Servers distribute the events as-is without consideration for the underlying feature, meaning each server should be able to determine how many of each event there is. Reactions can also be used as a form of "DAG healing" when rooms get complex. This is because they are normally higher traffic than regular messages, giving more opportunity for the DAG to heal extremities. The delay associated with needing to process these complicated event trees is also (typically) less critical than message traffic.

proposals/4074-server-side-annotations-aggregation.md
```diff
@@ -0,0 +1,270 @@
...
+keys they aggregate to a single event by a reasonable value (that
+corresponds to a limited number of different reaction kinds, e.g. "👍",
+"👎"). This is needed to avoid the situation when malicious users may
+attack server creating jumbo-events (events with arbitrary high count of
+different reaction kinds). Servers may configure this threshold on their
+discretion, but this number should not be lower than 16. When annotation
+key limit is reached, no new keys should be added to the aggregate, but
+older keys would continue to be aggregated (change their count).
+Annotations beyond that threshold should still be available via
+[`relations`](https://spec.matrix.org/v1.8/client-server-api/#relationships-api) endpoints.
```
turt2live, line 77: How would a server communicate that it hit the limit?

proposals/4074-server-side-annotations-aggregation.md
```diff
@@ -0,0 +1,270 @@
...
+apply their aggregate on top (in addition) of the aggregate server
+may already provide to event. Such aggregates would never overlap
+as server never aggregates encrypted events and simple deterministic
+logic to merge server and client aggregates exists in this case.
+
+## Unstable prefix
+
+No new identifiers are proposed; it is proposed that servers implementing
+this
+proposal simply do so on the existing endpoints.
```
turt2live, line 266: `filter_server_aggregated_relation_types` and the `m.reaction` EDU appear new, as do a few other identifiers. They will need to be namespaced.

turt2live (approved):
overall this seems like a great concept. I'm not sure on the client synchronization stuff myself, but others should be. A listed client implementation would help quite a bit in this understanding, I suspect.

proposals/4074-server-side-annotations-aggregation.md
turt2live, line 1: Implementation requirements:

* [ ] Server supporting filter opt-in behaviour
* [ ] Client using the filter behaviour

dihydrogenmonoxide16:
> overall this seems like a great concept. I'm not sure on the client synchronization stuff myself, but others should be. A listed client implementation would help quite a bit in this understanding, I suspect.

@turt2live , I'm doing a follow-up work to answer your comments and extend/adjust the proposal. Will post updates in ~1 week due to high work load.

turt2live:
No worries! Take the time you need :)

proposals/4074-server-side-annotations-aggregation.md
dihydrogenmonoxide16: I completely removed this clause. Potentially this problem already exists now, but only on the client side. I think it would better get addressed separately in order not to overcomplicate this change proposal (already big enough and technically complex)

proposals/4074-server-side-annotations-aggregation.md
dihydrogenmonoxide16: prefixed with `msc4074.`, also renamed `filter_server_aggregated_relation_types` to `msc4072.not_aggregated_relations` to better match other filter params naming

proposals/4074-server-side-annotations-aggregation.md
dihydrogenmonoxide16: What is the expectation here?

- to add this comment to the MXC inself
- some pseudocode
- working client/server implementation somewhere
- arch doc
?

Or is it just a ref comment for other contributors?

dihydrogenmonoxide16:
> No worries! Take the time you need :)

@turt2live, thank you for your feedback and collaboration! I answered your comments and addressed the points you mentioned. Added few improvements that turned to be necessary for a few niche corner cases (gap resolution, older event updates). Asked few questions there also.

proposals/4074-server-side-annotations-aggregation.md
tulir: The third option, a working client/server implementation somewhere. MSCs must be proven with implementations before they can be accepted: https://spec.matrix.org/proposals/#implementing-a-proposal

It generally doesn't matter who implements it or where, so implementation requirements are commented as a reference for everyone

PR #4075, opened 2023-11-08
MSC4075: MatrixRTC invites and notifications

toger5:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixrtc-call-ringing/proposals/4075-rtc-notification-event.md)

### DIFF
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -1,4 +1,4 @@
-# MSC0000: MatrixRTC Call Ringing
+# MSC4075: MatrixRTC Call Ringing
 It is important that a call initiated on clientA can make targeted clients ring.
 This is of interest in 1:1 Rooms/Calls but also in bigger rooms ringing can be desired.
 
@@ -29,7 +29,7 @@ This event contains the following fields by leveraging intentional mentions.
 }
 ```
 
-In the following we define **call** as any matrix RTC session with the same `"application"` and the same application specific data. In the calse of `"m.call"`, the same `"call_id"`.
+In the following we define **call** as any matrix RTC session with the same `"application"` and the same application specific data. In the case of `"m.call"`, the same `"call_id"`.
 
 On retrieval, the event should not be rendered in the timeline. But if the notify conditions (listed below) apply, the client has to inform the user about the **call** with an appropriate user flow. 
 For `notify_by == "ring"` some kind of sound is required (except if overwritten by another client specific setting), 
@@ -79,7 +79,7 @@ This is another timeline event where any room participant can send a push notifi
 
 ## Unstable prefix
 While this MSC is not present in the spec, clients and widgets should:
-   - Use org.matrix.mscXXXX. in place of m. in all new identifiers of this MSC. (`m.call.notify`)
+   - Use org.matrix.msc4075. in place of m. in all new identifiers of this MSC. (`m.call.notify`)
 
 ## Dependencies
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -26,6 +26,7 @@ This event contains the following fields by leveraging intentional mentions.
 {
   "content": {
     "m.mentions": {"user_ids": [], "room": true | false},
+    "m.relates_to": {"rel_type":"m.rtc.notification", "eventId":"$rtc_member_event_id"},
     "notification_type": "ring | notification",
   }
 }
@@ -38,8 +39,13 @@ The fields are defined as follows:
 - `notification_type` required string:\
   The type of notification to send.\
   `ring`: The client should ring.\
-  `notification`: The client should show a notification.
-- `session` required object: the `session` content from the corresponding `m.rtc.member` event.
+  `notification`: The client should show a notification.\
+- `m.relates_to`: optional:\
+  A relation (with type: `m.rtc.session_parent`) to the session participation event (`m.rtc.member`) of the notifying member.
+  If available this can be used to gather session data for this notify event.
+  It should always be used if the notify is related to an rtc session. (Eventually this event might be used to send
+  RTC session independent ringing notifications with a custom message. A bot controlled tea timer for instance.)
+The `session` information for this notify event can be obtained by the relation to the `m.rtc.member` event.
 
 In the following we define **call** as any MatrixRTC session with the
 same `"session"` contents.
@@ -99,6 +105,8 @@ Sending a `m.rtc.notification` should happen only if all of these conditions app
 - If the user has not yet received a `m.rtc.notification` for the **call** they want to
   participate but the other condition applies. (So the obvious case is, that this
   is the first user in a new call session).
+- If possible the user should first send their `m.rtc.member` event first to allow setting up the relation
+  for the notify event.
 
 ### Limitations and recommendations
 
@@ -192,6 +200,7 @@ level to ring the whole room.
 While this MSC is not present in the spec, clients and widgets should:
 
 - Use `org.matrix.msc4075.rtc.notification` in place of `m.rtc.notification` as the event type
+- Use `org.matrix.msc4075.rtc.session_parent` in place of `m.rtc.session_parent` relation type.
 
 ## Dependencies
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -120,29 +120,27 @@ Sending a `m.rtc.notification` should happen only if all of these conditions app
 - The sending client can compute the "exact" (or at least "a good approximation" if the local clocks are not configured correctly)
   at which the `m.rtc.notification` ring will end using the `lifeitime` + `start_ts`. This allows the sending client to
   show a local dialing/ringing animation/indicator/sound.
-
-### Limitations and recommendations
-
-- Encrypted rooms configured as `mentions only` are currently not sending push
-  notifications for encrypted events. Hence the client would not ring even though
-  the ring event contains `m.mentions`.
-  - As a stop gat, it is recommended, that the client sends unencrypted `m.rtc.notification`
-    events in such rooms.
-  - As soon as [MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996)
-    is supported `m.has_mentions` should be used instead of unencrypted call
-    notify events.
-- Wanting to ring a user who you do not have a shared room with is not possible.
-  It might be an undesired capability that your device can be started to ring
-  by users you have not yet interacted on matrix.
-  On the other hand this might be desired to mimic what ppl expect from using
-  the telephone network.
-  Entering a matrix userId allows to call someone (Ring their phone).
-  (It would be possible to disable/configure this on the receiving
+ the receiving
   device)
   - The location to put this information would be the invite event.
     This would be an edge case and only required for the specific usecase
     of being able to ring without a shared DM/Room.
     It should be discussed in an additional MSC and is not part of this proposal.
+
+### Extensible Events
+
+The concept of [extensible events](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+is used to offer a fallback f clients that do not support noticiation events and benefit from a fallback text reperentation.
+
+The body should contain a `m.text` field with the following content:
+
+```json5
+// notification
+ m.text: [{"body": "Call started by {sender}"}]
+
+// ring event
+ m.text: [{"body": "Call started by {sender} 🎶"}]
+```
 
 ## Alternatives
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -19,7 +19,7 @@ mentions make sure no unnecessary push notification is sent.
 
 This event contains the following fields including intentional mentions and extensible events.
 
-```json
+```json5
 {
   "content": {
     "application": "m.call | m.other_matrix_session_type | ...",
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -51,7 +51,8 @@ The fields are defined as follows:
   `ring`: The client should ring.\
   `notification`: The client should show a notification.\
 - `m.relates_to`: optional:\
-  A relation (with type: `m.rtc.session_parent`) to the session participation event (`m.rtc.member`) of the notifying member.
+  A relation (with type: `m.rtc.notification.parent`) to the session participation event (`m.rtc.member`) of the
+  notifying member.
   If available this can be used to gather session data for this notify event.
   It should always be used if the notify is related to an rtc session. (Eventually this event might be used to send
   RTC session independent ringing notifications with a custom message. A bot controlled tea timer for instance.)
@@ -215,7 +216,7 @@ level to ring the whole room.
 While this MSC is not present in the spec, clients and widgets should:
 
 - Use `org.matrix.msc4075.rtc.notification` in place of `m.rtc.notification` as the event type
-- Use `org.matrix.msc4075.rtc.session_parent` in place of `m.rtc.session_parent` relation type.
+- Use `org.matrix.msc4075.rtc.notification.parent` in place of `m.rtc.notification.parent` relation type.
 
 ## Dependencies
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -19,7 +19,7 @@ mentions make sure no unnecessary push notification is sent.
 
 This event contains the following fields by leveraging intentional mentions.
 
-```
+```json
 {
   "content": {
     "application": "m.call | m.other_matrix_session_type | ...",
@@ -94,8 +94,7 @@ Sending a `m.call.notify` should happen only if all of these conditions apply:
 
 It would be possible to use the call member room state events to determine a call start.
 The logic would be as following:
-
-_If we receive an event we check if there is are already other members
+_If we receive an event we check if  are already other members
 (call.member events) for the call. In case there is not we make the phone ring._
 
 Pros:
@@ -105,10 +104,7 @@ Pros:
   start a new call. Because they would automatically send an event by joining.
 - There would be less traffic. With the proposed solution the first one who joins
   needs to send a `m.call.notify` event and a `m.call.member` state event.
-- Very flexible, since the `m.call.notify` event is separate from a call
-  participation. This allows an external instance (a meeting organizer bot) to
-  just ring all the users which are invited to a meeting without needing to be
-  part of the call.
+
 
 Cons
 
@@ -117,6 +113,11 @@ Cons
   (Consider a very large room where I want to start a call only for the interested
   ones who want to discuss a side project. It would be very annoying if the
   initiator could not control how and who is going to be informed about that call.)
+
+- Additionally, it is not as flexible as the proposed separate event.
+  Which allows an external instance (a meeting organizer bot) to
+  just ring all the users which are invited to a meeting without needing to
+  participate in the call with a `m.call.member` event of the call.
 - Push notifications would need to be sent for EVERY `m.call.member` state event
   update. For each joining and leaving user and for each membership update during
   a call (due to a SFU (single forwarding unit) change, changing devices
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -1,80 +1,89 @@
-# MSC4075: MatrixRTC Call Ringing
+# MSC4075: MatrixRTC Ringing
 
-It is important that a call initiated on clientA can make targeted clients ring.
+It is important that an RTC session initiated on clientA can make targeted clients ring.
 This is of interest in 1:1 Rooms/Calls but also in bigger rooms ringing can be desired.
 
-Legacy calls are using room events to negotiate the call.
-A client could use the initial steps in the negotiation to also make the phone ring.
+The existing "legacy" [1:1 VoIP calling in Matrix](https://spec.matrix.org/v1.11/client-server-api/#voice-over-ip)
+uses room events to negotiate the call.
+A client can use the initial steps in the negotiation to also make the phone ring.
 
-With MatrixRTC based calls this signalling is done over state events.
-Also MatrixRTC enables large group calls which makes it very
+With [MSC4143: MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+this signalling is done over state events. Furthermore, [MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+allows for large group calls (not just 1:1) which makes it very
 desirable to have more configurations over the ringing process.
 
 ## Proposal
 
-A new event `m.call.notify` is proposed which can be sent by a client that
-wants to start a call. This event is added to the push rules for clients which
-support calling so they get push notifications. The push rules for intentional
+A new event `m.rtc.notify` is proposed which can be sent by a client that
+wants to notify others about the existence of an session for an MatrixRTC application.
+This event is added to the push rules for clients which
+support the application type so they receive push notifications. The push rules for intentional
 mentions make sure no unnecessary push notification is sent.
 
-This event contains the following fields including intentional mentions and extensible events.
+This event contains the following fields by leveraging intentional mentions.
 
 ```json5
 {
-  "type": "m.call.notify",
   "content": {
-    "application": "m.call | m.other_matrix_session_type | ...",
     "m.mentions": {"user_ids": [], "room": true | false},
     "notify_type": "ring | notification",
+    "session": {
+      // session content from the corresponding m.rtc.member event
+      "application": "m.call | m.other_matrix_session_type | ...",
+    },
     // Application specific data,
     // optional fields to disambiguate which session
     // this notify event belongs to:
-
-    // for application = "m.call":
-    "call_id": "some_id",
-    // Extensible events fallback
-    "m.text": "<@room|@user1, @user2 and @user3>
-      <notify_type == 'ring' ? 'RingRing!!' : ''> Join the call in this room with a supported client."
-    }
+  }
 }
 ```
 
+The fields are defined as follows:
+
+- `m.mentions` optional:\
+  Has the structure as defined for `m.mentions` in the [Client-Server API](https://spec.matrix.org/v1.11/client-server-api/#definition-mmentions).
+- `notify_type` required string:\
+  The type of notification to send.\
+  `ring`: The client should ring.\
+  `notification`: The client should show a notification.
+- `session` required object: the `session` content from the corresponding `m.rtc.member` event.
+
 In the following we define **call** as any MatrixRTC session with the
-same `"application"` and the same application specific data.
-In the case of `"m.call"`, the same `"call_id"`.
+same `"session"` contents.
 
-If it does support the event it should not be rendered in the timeline.
-(Call timeline elements are done using the `m.call.member` state events.)
+### Client behaviour on receiving a `m.rtc.notify` event
 
-But, if the notify conditions (listed below) apply,
-the client has to inform the user about the **call** with an appropriate user experience.
-For `notify_by == "ring"` some kind of sound is required
+On retrieval, the client should not render the event in the timeline.
+If the notify conditions (listed below) apply,
+the client has to inform the user about the **call** with an appropriate user flow.
+For `notify_type == "ring"` some kind of sound is required
 (except if overwritten by another client specific setting),
-for `notify_by == "notification"` a visual indication is enough.
+for `notify_type == "notification"` a visual indication is enough.
 This visual indication should be more than an unread indicator
 and similar to a notification banner.
 This is not enforced by the spec however and ultimately a client choice.
 
-Ringing (or notifying) should happen only if all these conditions apply:
+The client should only inform the user if all of the following conditions apply:
 
-- `m.call.notify` content:\
-  If the user is not part of the `m.mentions` section as defined in
-  [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952)
-  this event can be ignored. (Push notifications are automatically filtered
+- `m.rtc.notify` content:\
+  If the user is *not* listed in the `m.mentions` section as defined in the\
+  [Client-Server API](https://spec.matrix.org/v1.11/client-server-api/#definition-mmentions),\
+  the event should be ignored. (Push notifications are automatically filtered
   so this only is important for events received via a sync)
 - Local notification settings:\
-  If the room is set to silent, it will never play a ring sound. A `m.call.notify`
-  event will at most be used to mark the room as unread or update the rooms
+  If the room is set to silent, the client should never play a ring sound.
+  In this scenario, a `m.rtc.notify`
+  event should at most be used to mark the room as unread or update the rooms
   "has active **call** icon". (the exact behavior is up to the client)
 - Currently playing a ring sound (room timeline):\
   If the user already received a ring event for this **call** and is playing
-  the ring sound any incoming `m.call.notify` for the same **call**
-  should be ignored. If the user failed to pick up and a new `m.call.notify`
+  the ring sound any incoming `m.rtc.notify` for the same **call**
+  should be ignored. If the user failed to pick up and a new `m.rtc.notify`
   arrives for the same room the device should ring again.
 - Current user is a member of the the **call** (room state):\
-  None of the devices should ring if they receive a `m.call.notify` if the
-  rooms state `m.call.member` event of the user contains a membership for
-  the **call** in the `m.call.notify` event.
+  None of the devices should ring if they receive a `m.rtc.notify` if the
+  rooms state `m.rtc.member` event of the user contains a membership for
+  the **call** in the `m.rtc.notify` event.
   This includes stopping the current ring sound if the room state updates so
   this condition is true.
 - If a notify event is received in "real time":\
@@ -85,77 +94,72 @@ Ringing (or notifying) should happen only if all these conditions apply:
   is allowed. Any client which is not able to receive the event in this period should
   not ring to prohibit (annoying/misleading/irrelevant) outdated rings.
 
-Sending a `m.call.notify` should happen only if all of these conditions apply:
+### Client behaviour when sending a `m.rtc.notify` event
+
+Sending a `m.rtc.notify` should happen only if all of these conditions apply:
 
 - If the user deliberately wants to send a new notify event
-  (It is possible to send a `m.call.notify` for an ongoing call if that
+  (It is possible to send a `m.rtc.notify` for an ongoing call if that
   makes sense. Starting a call ahead of time, planning in a small group,
   ringing another set of users at a specific time so they don't forget to join.
   Ringing one specific user again who missed joining during the first ring.)
-- If the user has not yet received a `m.call.notify` for the **call** they want to
+- If the user has not yet received a `m.rtc.notify` for the **call** they want to
   participate but the other condition applies. (So the obvious case is, that this
   is the first user in a new call session).
-
-### Fallback with extensible events
-
-If clients do not support `m.call.notify` events, they will however draw the extensible event fallback.
-The exact content is up the the sending client and can contain more detailed instructions on how to get
-access to a supporting client. It has to include the `@room` or the mentioned user list and has
-to communicate the message that those users are invited to participate in a call.
 
 ### Limitations and recommendations
 
 - Encrypted rooms configured as `mentions only` are currently not sending push
   notifications for encrypted events. Hence the client would not ring even though
   the ring event contains `m.mentions`.
-  - As a stop gap, it is recommended, that the client sends unencrypted `m.call.notify`
+  - As a stop gat, it is recommended, that the client sends unencrypted `m.rtc.notify`
     events in such rooms.
   - As soon as [MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996)
     is supported `m.has_mentions` should be used instead of unencrypted call
     notify events.
-- Ringing a user who you do not have a shared room with is not supported
-  by this MSC.
+- Wanting to ring a user who you do not have a shared room with is not possible.
   It might be an undesired capability that your device can be started to ring
   by users you have not yet interacted on matrix.
-  On the other hand this might be desired to mimic what people expect from using
+  On the other hand this might be desired to mimic what ppl expect from using
   the telephone network.
-  (It should be possible to disable/configure this on the receiving
+  Entering a matrix userId allows to call someone (Ring their phone).
+  (It would be possible to disable/configure this on the receiving
   device)
   - The location to put this information would be the invite event.
-    This would be an edge case and only required for the specific use case
+    This would be an edge case and only required for the specific usecase
     of being able to ring without a shared DM/Room.
     It should be discussed in an additional MSC and is not part of this proposal.
 
 ## Alternatives
 
-It could be possible to use the call member room state events to determine a call
+### Use call member room state events
+
+It would be possible to use the call member room state events to determine a call
 start.
 The logic would be as following:
-_If we receive an event we check if there are already other members
+_If we receive an event we check if  are already other members
 (call.member events) for the call. In case there is not we make the phone ring._
 
 Pros:
 
 - This would not require any new event.
-- The clients can not "forget" to ring the others when they
-  start a new call, because they would automatically send an event by joining.
-- There would be less traffic. With the proposed solution in this MSC, the first
-  client who joins needs to send a `m.call.notify` event and a `m.call.member`
-  state event.
+- The clients can not "forget" to ring the others about the when they
+  start a new call. Because they would automatically send an event by joining.
+- There would be less traffic. With the proposed solution the first one who joins
+  needs to send a `m.rtc.notify` event and a `m.rtc.member` state event.
 
 Cons
 
 - All the ringing conditions run on the receiving user. There is no way for the
-  user who starts the call to decide if it should ring the other participants.
+  user who start the call to decide if it should ring the other participants.
   (Consider a very large room where I want to start a call only for the interested
   ones who want to discuss a side project. It would be very annoying if the
   initiator could not control how and who is going to be informed about that call.)
-
 - Additionally, it is not as flexible as the proposed separate event.
   Which allows an external instance (a meeting organizer bot) to
   just ring all the users which are invited to a meeting without needing to
-  participate in the call with a `m.call.member` event of the call.
-- Push notifications would need to be sent for EVERY `m.call.member` state event
+  participate in the call with a `m.rtc.member` event of the call.
+- Push notifications would need to be sent for EVERY `m.rtc.member` state event
   update. For each joining and leaving user and for each membership update during
   a call (due to a SFU (single forwarding unit) change, changing devices
   (could even happen for screen shares if the screen share is implemented as a
@@ -178,13 +182,13 @@ Cons
 ## Security considerations
 
 This is another timeline event where any room participant can send a push
-notification to others. Since this will make clients ring this has a high
-impact on the receiver. However, since ringing has to obey the mute settings, it is
-very easy for the targeted users to mitigate unwanted ringing. It can be very
+notification to others. Since this will make clients ring this has a higher
+effect on the receiver. Since ringing has to obey the mute settings, it is
+very easy for the targeted users to mitigate the "attack". It can be very
 much compared to spamming a room with "@room" messages.
 
-The default power level for `m.call.notify` is `50` and equivalent to the default
-power level required for `m.call.member` state events.
+The default power level for `m.rtc.notify` is `50` and equivalent to the default
+power level required for `m.rtc.member` state events.
 
 Additional control is provided indirectly with the use of intentional mentions.
 Setting `"notifications":{"room":X}` allows to choose `X` for the power required
@@ -194,9 +198,8 @@ level to ring the whole room.
 
 While this MSC is not present in the spec, clients and widgets should:
 
-- Use `org.matrix.msc4075.` in place of `m.` in all new identifiers of this MSC.
-(`m.call.notify`)
+- Use `org.matrix.msc4075.call.notify` in place of `m.rtc.notify` as the event type
 
 ## Dependencies
 
-This MSC builds on Intentional Mentions [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952).
+This MSC builds on [MSC4143: MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143).
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -35,7 +35,7 @@ This event contains the following fields by leveraging intentional mentions.
 }
 ```
 
-In the following we define **call** as any matrix RTC session with the
+In the following we define **call** as any MatrixRTC session with the
 same `"application"` and the same application specific data.
 In the case of `"m.call"`, the same `"call_id"`.
 
@@ -74,7 +74,7 @@ Ringing (or notifying) should happen only if all these conditions apply:\
 - If a notify event is received in "real time":\
   Notify events that are older then **10 seconds** are ignored (using the local
   timestamp computed via `unsigned.age`).\
-  Otherwise a clients syncing for the first time would ring for outdated call events.
+  Otherwise a client syncing for the first time would ring for outdated call events.
   In general ringing only makes sense in "real time". A 10 second syncing latency
   is allowed. Any client which is not able to receive the event in this period should
   not ring to prohibit (annoying/misleading/irrelevant) outdated rings.
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -25,6 +25,8 @@ This event contains the following fields by leveraging intentional mentions.
 ```json5
 {
   "content": {
+    "sender_ts": 1752583130365,
+    "lifetime": 30000,
     "m.mentions": {"user_ids": [], "room": true | false},
     "m.relates_to": {"rel_type":"m.rtc.notification", "eventId":"$rtc_member_event_id"},
     "notification_type": "ring | notification",
@@ -34,6 +36,14 @@ This event contains the following fields by leveraging intentional mentions.
 
 The fields are defined as follows:
 
+- `sender_ts` the local timestamp observed by the sender device. Is used in combination with lifetime to evaluate if
+  the notification event is valid. To mitigate clients lying about the `sender_ts` this value has to be checked with `origin_server_ts`.
+  The diff between `sender_ts` and `origin_server_ts` should be at most 20s.
+  If not the `origin_server_ts` should be used for the lifetime
+  computation instead.
+- `lifetime` the relative time to the `sender_ts` for which the `ring` is active or to define the window in which the
+  `notification` is not ignored. The recommended value is **30 seconds**.
+  The receiving client **should** cap the lifetime to an upper bound. (recommended: **2 minutes**).
 - `m.mentions` optional:\
   Has the structure as defined for `m.mentions` in the [Client-Server API](https://spec.matrix.org/v1.11/client-server-api/#definition-mmentions).
 - `notification_type` required string:\
@@ -86,12 +96,12 @@ The client should only inform the user if all of the following conditions apply:
   This includes stopping the current ring sound if the room state updates so
   this condition is true.
 - If a notify event is received in "real time":\
-  Notify events that are older then **20 seconds** are ignored (using the local
-  timestamp computed via `unsigned.age`).\
+  Notify events that are older then **`lifetime`** are ignored (using the
+  `sender_ts` timestamp).\
   Otherwise a client syncing for the first time would ring for outdated call events.
-  In general ringing only makes sense in "real time". A 20 second syncing latency
-  is allowed. Any client which is not able to receive the event in this period should
-  not ring to prohibit (annoying/misleading/irrelevant) outdated rings.
+  In general ringing only makes sense in "real time".
+  Any client which is not able to receive the event in the `lifetime` period should
+  not ring to prohibit annoying/misleading/irrelevant/outdated rings.
 
 ### Client behaviour when sending a `m.rtc.notification` event
 
@@ -107,6 +117,9 @@ Sending a `m.rtc.notification` should happen only if all of these conditions app
   is the first user in a new call session).
 - If possible the user should first send their `m.rtc.member` event first to allow setting up the relation
   for the notify event.
+- The sending client can compute the "exact" (or at least "a good approximation" if the local clocks are not configured correctly)
+  at which the `m.rtc.notification` ring will end using the `lifeitime` + `start_ts`. This allows the sending client to
+  show a local dialing/ringing animation/indicator/sound.
 
 ### Limitations and recommendations
 
@@ -138,8 +151,8 @@ Sending a `m.rtc.notification` should happen only if all of these conditions app
 It would be possible to use the call member room state events to determine a call
 start.
 The logic would be as following:
-_If we receive an event we check if  are already other members
-(call.member events) for the call. In case there is not we make the phone ring._
+*If we receive an event we check if  are already other members
+(call.member events) for the call. In case there is not we make the phone ring.*
 
 Pros:
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -17,7 +17,7 @@ wants to start a call. This event is added to the push rules for clients which
 support calling so they get push notifications. The push rules for intentional
 mentions make sure no unnecessary push notification is sent.
 
-This event contains the following fields by leveraging intentional mentions.
+This event contains the following fields including intentional mentions and extensible events.
 
 ```json
 {
@@ -31,7 +31,10 @@ This event contains the following fields by leveraging intentional mentions.
 
     // for application = "m.call":
     "call_id": "some_id",
-  }
+    // Extensible events fallback
+    "m.text": "<@room|@user1, @user2 and @user3>
+      <notify_type == 'ring' ? 'RingRing!!' : ''> Join the call in this room with a supported client."
+    }
 }
 ```
 
@@ -39,9 +42,11 @@ In the following we define **call** as any MatrixRTC session with the
 same `"application"` and the same application specific data.
 In the case of `"m.call"`, the same `"call_id"`.
 
-On retrieval, the event should not be rendered in the timeline.
-But if the notify conditions (listed below) apply,
-the client has to inform the user about the **call** with an appropriate user flow.
+If it does support the event it should not be rendered in the timeline.
+(Call timeline elements are done using the `m.call.member` state events.)
+
+But, if the notify conditions (listed below) apply,
+the client has to inform the user about the **call** with an appropriate user experience.
 For `notify_by == "ring"` some kind of sound is required
 (except if overwritten by another client specific setting),
 for `notify_by == "notification"` a visual indication is enough.
@@ -89,6 +94,13 @@ Sending a `m.call.notify` should happen only if all of these conditions apply:
 - If the user has not yet received a `m.call.notify` for the **call** they want to
   participate but the other condition applies. (So the obvious case is, that this
   is the first user in a new call session).
+
+### Fallback with extensible events
+
+If clients do not support `m.call.notify` events, they will however draw the extensible event fallback.
+The exact content is up the the sending client and can contain more detailed instructions on how to get
+access to a supporting client. It has to include the `@room` or the mentioned user list and has
+to communicate the message that those users are invited to participate in a call.
 
 ### Limitations and recommendations
 
--- a/proposals/4075-call-notify-event.md
+++ b/proposals/4075-call-notify-event.md
@@ -22,9 +22,9 @@ This event contains the following fields by leveraging intentional mentions.
 ```
 {
   "content": {
-    "application": "m.call" | "m.other_matrix_session_type" | "...",
+    "application": "m.call | m.other_matrix_session_type | ...",
     "m.mentions": {"user_ids": [], "room": true | false},
-    "notify_type": "ring" | "notification",
+    "notify_type": "ring | notification",
     // Application specific data,
     // optional fields to disambiguate which session
     // this notify event belongs to:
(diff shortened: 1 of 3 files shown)


proposals/4075-call-notify-event.md
turt2live, line 1: Please limit lines to approximately 120 characters.

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,142 @@
...
+  specific list of users to notify want to be specified.
+- It would not make it possible to ring without participating.
+
+## Security considerations
+
+This is another timeline event where any room participant can send a push
+notification to others. Since this will make clients ring this has a higher
+effect on the receiver. Since ringing has to obey the mute settings, it is
+very easy for the targeted users to mitigate the "attack". It can be very
+much compared to spamming a room with "@room" messages.
```
davidegirardi, line 131: Can we add some words about default power levels here or somewhere else?

We could alternatively model a default behaviour in the spec itself so we avoid issue by client applications forgetting to implement default power levels.

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,142 @@
...
+But if the notify conditions (listed below) apply,
+the client has to inform the user about the **call** with an appropriate user flow.
+For `notify_by == "ring"` some kind of sound is required
+(except if overwritten by another client specific setting),
+for `notify_by == "notification"` a visual indication is enough.
+This visual indication should be more than an unread indicator
+and similar to a notification banner.
+This is not enforced by the spec however and ultimately a client choice.
+
+Notify conditions depend on:\
```
davidegirardi, line 52: I don't see anything about closed/completed calls. What should a client do if it was offline and it receives the state event when backfilling?

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,142 @@
...
+  This includes stopping the current ring sound if the room state updates so
+  this condition is true.
+
+Sending a `m.call.notify` should happen if all of these conditions apply:
+
+- If the user deliberately wants to send a new notify event
+  (It is possible to send a `m.call.notify` for an ongoing call if that
+  makes sense. Starting a call ahead of time, planning in a small group,
+  ringing another set of users at a specific time so they don't forget to join.
+  Ringing one specific user again who missed joining during the first ring.)
```
davidegirardi, line 82: Sending an empty `"user_ids": []` array is also OK right?
If someone wants to start an optional "drop-in" call.

toger5: Starting a call is transparent via the call.mebers event anyways. So a notify event without any content (room= false and empty user_ids) is equivalent to not sending a notify event.

The client will display ongoing calls based on memberships anyways

proposals/4075-call-notify-event.md
toger5: I updated the msc. (thanks for the feedback)

uhoreg:
@toger5 You've added some implementations in the PR description.  Do these implement the MSC in full, or are there parts of the MSC that are still lacking implementation?

toger5:
Yes it's implemented in full. At least on the platforms shown in the description (EW).

There are optional configurations that are still missing. (Allowing to send a notification to a specific subset of users) but the API for this is already part of the SDK. There is no ui for this however and we are not sure we want that in EW.

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,164 @@
...
+
+- This would not require any new event.
+- The clients can not "forget" to ring the others about the when they
+  start a new call. Because they would automatically send an event by joining.
+- There would be less traffic. With the proposed solution the first one who joins
+  needs to send a `m.call.notify` event and a `m.call.member` state event.
+- Very flexible, since the `m.call.notify` event is separate from a call
+  participation. This allows an external instance (a meeting organizer bot) to
+  just ring all the users which are invited to a meeting without needing to be
+  part of the call.
```
TobiasFella, line 111: This sounds like an advantage of *separate* events, not of the *combined* approach

toger5: This is a mistake. Thanks for finding it.

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,165 @@
+# MSC4075: MatrixRTC Call Ringing
+
+It is important that a call initiated on clientA can make targeted clients ring.
+This is of interest in 1:1 Rooms/Calls but also in bigger rooms ringing can be desired.
+
+Legacy calls are using room events to negotiate the call.
+A client could use the initial steps in the negotiation to also make the phone ring.
+
+With matrix RTC based calls this signalling is done over state events.
```
giomfo, line 9: @toger5 what are these state events? Are these all `m.call.member` events?

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,165 @@
...
+and similar to a notification banner.
+This is not enforced by the spec however and ultimately a client choice.
+
+Ringing (or notifying) should happen only if all these conditions apply:\
+
+- `m.call.notify` content:\
+  If the user is not part of the `m.mentions` section as defined in
+  [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952)
+  this event can be ignored. (Push notifications are automatically filtered
+  so this only is important for events received via a sync)
```
giomfo, line 58: @toger5 is this `m.call.notify` event encrypted when it is sent in an encrypted room? 
I think it is, so the Push notification could not be filtered server side. The notification will be filtered locally by the client. 
Note: If this event is encrypted, the call notification may have some trouble in encrypted mentions-only rooms, like it is described in
[MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996). Indeed the encrypted event will not be pushed by the server in that case.

toger5: That is good to know. I am wondering if a @room notify event should be unencrypted in that case. This would sacrifice privacy but make it more reliable.
If this is transparent to the user that could be acceptable maybe.

proposals/4075-call-notify-event.md
toger5: Yes exactly

proposals/4075-call-notify-event.md
```diff
@@ -0,0 +1,165 @@
...
+  separate participating device) or matrix RTC business logic of the call.)
+  This could result in a lot of push notification with no obvious/simple way to
+  filter them.
+- It would require bloating the `call.member` event if the `notify_type` or a
+  specific list of users to notify want to be specified.
+- It would not make it possible to ring without participating.
+
+### Explicitly unspecified conditions
+
+- The duration of the ring sound is a deliberately chosen
```
td-famedly, line 134: Instead of hardcoding 10 seconds, I am assuming the clients could probably just look at the `lifetime` field for the invite event mapped to `call_id` from the notify event.

proposals/4075-call-notify-event.md
toger5: Thanks for also linking MSC3996.

toger5: I added a `Limitations and recommendations` [section ](https://github.com/matrix-org/matrix-spec-proposals/blob/a57720b3b887214e84c94359aa75bdbb49db8eb5/proposals/4075-call-notify-event.md#L93-L115) so this is explicitly mentioned.

proposals/4075-call-notify-event.md
toger5: Which invite event are you referring to? The general idea is to allow the receiver to choose a ring duration and not the sender.
So reading from any event would result in someone else choosing how long you device should ring.

td-famedly: > The general idea is to allow the receiver to choose a ring duration and not the sender.

To me that sounds a bit counter intuitive, f.ex I would not want to stop ringing the device after 10 seconds if the m.call.invite lifetime was 60 seconds and I can still pick up the call. Or does the "stop ringing after 10 seconds" also automatically send a reject then? (might be out of scope for this PR)

toger5: This is independent of the legacy call invite event. It is just what we use for matrix RTC sessions where there is no enforced call buildup procedure. (You just post your state event)
So most likely there is no call.invite event which lifetime could be used. The call notify event is the one that could include a lifetime. But we decided against that as described in the msc and the comment above.
It's also a slightly different semantic since the calling participant can stay in the call for however long they desire. Where a call invite has a Lifetime.

toger5: >  So reading from any event would result in someone else choosing how long you device should ring.

This topic was adressed by a recent update to the MSC. The sender knowing how long the receiver should ring is helpful to indicate if we are still ringing on the sender device.

So a sender defined lifetime makes sense. The MSC is explicit about the fact, that the final word on the ring duration is the receiving client.

toger5: This is now referenced in the MCS: https://github.com/matrix-org/matrix-spec-proposals/commit/aee19b084886bf28b94aff93841565b41a3cb2c7

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+
+```json5
+{
+  "type":"m.rtc.notification", // org.matrix.msc4075.rtc.notification
+  "content": {
+    "sender_ts": 1752583130365,
+    "lifetime": 30000,
+    "m.mentions": {"user_ids": [], "room": true | false},
+    "m.relates_to": {"rel_type":"m.reference", "event_id":"$rtc_member_event_id"},
+    "notification_type": "ring | notification",
```
Half-Shot: Noting that I'm playing around with the idea of including a `media_hint` field that gives a soft indication of whether the call is a "audio" or "video" (+audio) call. This is just to indicate between trusted callers that they can start with audio or video off, but the actual call semantics remain the same and they may switch at will. The intention is so that we can indicate this properly in notifications.

```suggestion
    "notification_type": "ring | notification",
    "m.call.intent": "audio | video | something-else",
```

Half-Shot: In practice this is now in our implementation and has certainly been proven a useful addition

BillCarsonFr: Looks like the current implementation in js-sdk uses `m.call.intent`https://github.com/matrix-org/matrix-js-sdk/blob/4d0d32307eb4f1ce1fb65080fcca704f5bdedc31/src/matrixrtc/types.ts#L150

BillCarsonFr: Is it required?

BillCarsonFr: Also in this MSC it is never talking about calls, it is talking about rtc sessions. So maybe the naming is bad?
Maybe it should just be a `rtc_intent` that depends on the actual session beeing notified? like `voice_call` | `video_call` | `white_board_invite` ...

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+  (Eventually this event might be used to send
+  RTC session independent ringing notifications with a custom message. A bot-controlled tea timer for instance.)
+The `session` information for this notification event can be obtained by the relation to the `m.rtc.member` event.
+
+In the following we define **call** as any MatrixRTC session with the
+same `"session"` contents.
+
+### Client behaviour on receiving a `m.rtc.notification` event
+
+On retrieval, the client should not render the event in the timeline.
```
Susurrus: I'm working on an implementation for Fractal (https://gitlab.gnome.org/World/fractal/-/issues/1720) and I believe this should be the event to anchor a rendering of the call in the timeline. If we're in a MatrixRTC world, what else would we use?

toger5: The long term plan on no matrixRTC timeline rendering is to use the `m.rtc.member` events they mark join and leaves.
A time based overlap computation of join/leave intervals (member sessions) will then be used to compute a historic call.

Similar to how we currently group other timeline events in clients (e.g. UserA and 10 other users join the room)
a new item would combine all the session information (UserA and 8 others participated in a call from X.XX until Y.YY)

toger5: But using the notification event is an acceptable stop gap. There are discussions if we want a specific notification as part of the m.call MSC instead of a general purpose m.notification event.

(mostly because that would allow a notification acknowlege response, which is needed for a reliable ring feedback UX)

So pls dont put too much time into the `m.rtc.notification` based approach. Make it simple.
Feel free to explore the "time based overlap computation of join/leave intervals" approach. It probably needs work in the rust sdk since this would be where the code should life.

If you look into this I would be very happy to be kept in the loop (ping me once a plan is laid out in an issue). (I already did some consideration on this topic and it would be unfortunate to waste resources due to a lack of communication)

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+With [MSC4143: MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+this signalling is done over state events. Furthermore, [MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+allows for large group calls (not just 1:1) which makes it very
+desirable to have more configurations over the ringing process.
+
+## Proposal
+
+A new event `m.rtc.notification` is proposed which can be sent by a client that
+wants to notify others about the existence of a session for a MatrixRTC application.
+This event is added to the push rules for clients which
```
BillCarsonFr: Just to add that, regardless of push rules, having a separate event for ringing can also allow to easily configure Power Level to decide who can ring or not in a room

proposals/4075-rtc-notification-event.md
Susurrus: I think it should be added to the MSC then, since it mentions this shouldn't be rendered, that it points to an MSC that contains the part that would be rendered. Otherwise this is confusing looking at this from the application side.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+For `notification_type == "ring"` some kind of sound is required
+(except if overwritten by another client specific setting),
+for `notification_type == "notification"` a visual indication is enough.
+This visual indication should be more than an unread indicator
+and similar to a notification banner.
+This is not enforced by the spec however and ultimately a client choice.
+
+The client should only inform the user if all of the following conditions apply:
+
+- `m.rtc.notification` content:\
```
Susurrus: I think also if it's not a room mention then it should be ignored, right?

Tho I'm a bit confused by the mentions how they're used here, since they're an optional field on the message. So a client could send one with no mentions according to this spec, but every other client should ignore it then; is there a reason to allow that? I'd think `m.mentions` should be required with either `room` or have at least one user listed.

proposals/4075-rtc-notification-event.md
Susurrus: Ruma currently implements this with `call_intent`. It's useful to be able to devide if you notify someone that there's a "call" or a "video call" incoming.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+This is of interest in 1:1 Rooms/Calls but also in bigger rooms ringing can be desired.
+
+The existing "legacy" [1:1 VoIP calling in Matrix](https://spec.matrix.org/v1.11/client-server-api/#voice-over-ip)
+uses room events to negotiate the call.
+A client can use the initial steps in the negotiation to also make the phone ring.
+
+With [MSC4143: MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+this signalling is done over state events. Furthermore, [MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
+allows for large group calls (not just 1:1) which makes it very
+desirable to have more configurations over the ringing process.
```
Johennes: I think this may still refer to the initial version of MatrixRTC from before sticky events.

```suggestion
With [MSC4143: MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
this signalling is done over a mixture of state events and sticky
timeline events. Furthermore, [MatrixRTC](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)
allows for large group calls (not just 1:1) which makes it very
desirable to have more configurations over the ringing process.
```

proposals/4075-rtc-notification-event.md
Johennes: > Also in this MSC it is never talking about calls, it is talking about rtc sessions. So maybe the naming is bad? Maybe it should just be a `rtc_intent` that depends on the actual session beeing notified? like `voice_call` | `video_call` | `white_board_invite` ...

Yeah, `m.call.intent` doesn't fit well with the generic MatrixRTC design of this proposal. Maybe we could keep this information in the `application` objects in slots or member events?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+
+The fields are defined as follows:
+
+- `sender_ts` the local timestamp observed by the sender device. Is used in combination with lifetime to evaluate if
+  the notification event is valid. To mitigate clients lying about the `sender_ts` this value has to be checked with `origin_server_ts`.
+  Receivers **SHOULD** use `origin_server_ts` if `|sender_ts - origin_server_ts| > 20000 ms`.
+- `lifetime` the relative time to the `sender_ts` for which the `ring` is active or to define the window in which the
+  `notification` is not ignored. The recommended value is **30 seconds**.
+  The receiving client **SHOULD** cap the lifetime to an upper bound. (recommended: **2 minutes**).
+  `lifetime` **MUST** be non-negative and SHOULD NOT exceed 120000 ms.
```
Johennes: To avoid ambiguity, we could make the unit of time explicit here.

```suggestion
- `lifetime` the time in milliseconds relative to `sender_ts` for which the `ring` is active / the
  `notification` is not ignored. The recommended value is **30 seconds**.
  The receiving client **SHOULD** cap the lifetime to an upper bound. (recommended: **2 minutes**).
  `lifetime` **MUST** be non-negative and SHOULD NOT exceed 120000 ms.
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+  The type of notification to send.\
+  `ring`: The client should ring.\
+  `notification`: The client should show a notification.\
+- `m.relates_to`: optional:\
+  A relation (with type: `m.reference`) to the session participation event
+  (`m.rtc.member`) of the notifying member.
+  If available this can be used to gather session data for this notification event.
+  It should always be used if the notification is related to an RTC session.
+  (Eventually this event might be used to send
+  RTC session independent ringing notifications with a custom message. A bot-controlled tea timer for instance.)
```
Johennes: For such a session-independent event, how would a receiving client know which slot to join when there is more than one in the room? I think this use case might need further thought and we should consider making the relation mandatory for the time being. That way receiving clients can route into the correct slot reliably.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+The `session` information for this notification event can be obtained by the relation to the `m.rtc.member` event.
+
+In the following we define **call** as any MatrixRTC session with the
+same `"session"` contents.
```
Johennes: What is `"session"` here? It doesn't seem to be defined anywhere.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+If the notification conditions (listed below) apply,
+the client has to inform the user about the **call** with an appropriate user flow.
+For `notification_type == "ring"` some kind of sound is required
+(except if overwritten by another client specific setting),
+for `notification_type == "notification"` a visual indication is enough.
+This visual indication should be more than an unread indicator
+and similar to a notification banner.
+This is not enforced by the spec however and ultimately a client choice.
+
+The client should only inform the user if all of the following conditions apply:
```
Johennes: I think this formulation is slightly misleading because most of the conditions below are in the format "when X then DON'T notify".

```suggestion
The following rules apply to the receiving client when notifying the user:
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+
+- `m.rtc.notification` content:\
+  If the user is *not* listed in the `m.mentions` section as defined in the\
+  [Client-Server API](https://spec.matrix.org/v1.11/client-server-api/#definition-mmentions),\
+  the event should be ignored. (Push notifications are automatically filtered
+  so this only is important for events received via a sync)
+- Local notification settings:\
```
Johennes: Is silencing a room really a local setting? Wouldn't you use push rules for this as well?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+  If the user already received a ring event for this **call** and is playing
+  the ring sound any incoming `m.rtc.notification` for the same **call**
+  should be ignored. If the user failed to pick up and a new `m.rtc.notification`
+  arrives for the same room the device should ring again.
+- Current user is a member of the **call** (room state):\
+  None of the devices should ring if they receive a `m.rtc.notification` if the
+  room's state `m.rtc.member` event of the user contains a membership for
+  the **call** in the `m.rtc.notification` event.
+  This includes stopping the current ring sound if the room state updates so
+  this condition is true.
```
Johennes: I think this needs updating to the new sticky events design.

```suggestion
  None of the devices should ring if they receive an `m.rtc.notification` event
  if there is any connected `m.rtc.member` event in the **call** associated
  with the `m.rtc.notification` event.
  This includes stopping the current ring sound if such a membership event
  appears.
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+  Which allows an external instance (a meeting organizer bot) to
+  just ring all the users which are invited to a meeting without needing to
+  participate in the call with a `m.rtc.member` event of the call.
+- Push notifications would need to be sent for EVERY `m.rtc.member` state event
+  update. For each joining and leaving user and for each membership update during
+  a call (due to a SFU (single forwarding unit) change, changing devices
+  (could even happen for screen shares if the screen share is implemented as a
+  separate participating device) or MatrixRTC business logic of the call.)
+  This could result in a lot of push notifications with no obvious/simple way to
+  filter them.
```
Johennes: I think in encrypted rooms this is true regardless, isn't it? Push Rules need to be re-evaluated after decryption on the client and that only works if the client gets push notifications for each encrypted event.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+  `notification` is not ignored. The recommended value is **30 seconds**.
+  The receiving client **SHOULD** cap the lifetime to an upper bound. (recommended: **2 minutes**).
+  `lifetime` **MUST** be non-negative and SHOULD NOT exceed 120000 ms.
+
+- `m.mentions` optional:\
+  Has the structure as defined for `m.mentions` in the [Client-Server API](https://spec.matrix.org/v1.11/client-server-api/#definition-mmentions).
+- `notification_type` required string:\
+  The type of notification to send.\
+  `ring`: The client should ring.\
+  `notification`: The client should show a notification.\
```
Johennes: The `notification` use case is slightly unclear to me – especially why it is combined with a lifetime.

For ringing, the lifetime makes a lot of sense. You ring because you're waiting in the call.

But if you don't ring, is the use case not more of a singular notification in the sense of a duration-less "call started" ping? Why would you care about the duration the banner is shown to the recipient in this case?

proposals/4075-rtc-notification-event.md
Johennes: Actually, it is part of #4196 already.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,227 @@
...
+    "sender_ts": 1752583130365,
+    "lifetime": 30000,
+    "m.mentions": {"user_ids": [], "room": true | false},
+    "m.relates_to": {"rel_type":"m.reference", "event_id":"$rtc_member_event_id"},
+    "notification_type": "ring | notification",
+  }
+}
+```
+
+The fields are defined as follows:
```
Johennes: #4196 adds further application-specific fields. We should maybe explicitly state here that the fields can be extended based on the MatrixRTC app?

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: This is now part of the notification event after ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: Resolved with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,280 @@
...
+}
+```
+
+- `notification_type` required:\
+  The type of notification to trigger. One of `ring`, `notification`.
+- `sender_ts`:\
+  The local timestamp on the sending device when the event was emitted. This is used in combination
+  with `lifetime` to evaluate if the notification event is valid. To mitigate clients lying about
+  the `sender_ts`, this value has to be checked against `origin_server_ts`. Receivers SHOULD use
+  `origin_server_ts` if `|sender_ts - origin_server_ts| > 20000 ms`.
```
Johennes: Should we clarify why we trust the sender's local time to be more accurate than the sending server's local time?

proposals/4075-rtc-notification-event.md
Johennes: I think this has been made slightly more clear with ffaad72f8a5aa3377a8f77176f498f8f296ee86d.

proposals/4075-rtc-notification-event.md
Johennes: I have made the mentions required with ed9e4c3b01ec6c33b60c92bd37f84e666784c672.

proposals/4075-rtc-notification-event.md
Johennes, line 1: @ara4n [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4143#discussion_r1856299594):

> Something which is very unclear right now is the lifecycle of a call (and the draft of MSC4196 doesn't really help).
>
> Specifically, what has happened to `m.call.invite`? `m.call.answer`? `m.call.select_answer`? `m.call.hangup`? The intention of MSC3401 was to keep using these (as per https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/group-voip/proposals/3401-group-voip.md#basic-call) even when an SFU is involved, in order to preserve signalling for:
>
> * Call was placed but isn't ringing yet (not that we have in legacy VoIP)
> * Call is ringing
> * Caller gave up and stopped ringing
> * Call has been answered on a device
> * Call was rejected locally
> * Call was rejected on all devices
> * Call was answered
> * Call was hung up
>
> i.e. all the state machine transitions that go into the lifecycle of a call.
>
> I'm very worried that this seems to have been collapsed to a single `m.call.notify` event from MSC4075 + `m.rtc.member` events here, which seems to be missing most of the above. For instance, I can't see how the caller tells the callee that it's no longer ringing them - not to mention stuff like early media (MSC3635) or ringback tones.
>
> Legacy VoIP was already a dumbed down version of SIP, and I'm worried that the lack of signalling semantics here is going to make mimicing SIP or PSTN semantics really hard (let alone bridging to it), and would be a backwards step from legacy voip.

Johennes: @fkwp and @ara4n had a chat about this out of band and agreed to consider this resolved. The RTC member event lifetime and the RTC notification event should cover the required lifecycle.

mmaudet:
While researching push behaviour for a deployment where every room is encrypted, I ran into what looks like an accidental deletion in this MSC.

### What happened

- `a57720b3b8` (2024-05-13) added a `### Limitations and recommendations` section, in response to @giomfo's [question above](https://github.com/matrix-org/matrix-spec-proposals/pull/4075#discussion_r1562541302) about whether this event is encrypted.
- `93f317760d` (2025-07-15), whose commit message is *"add extensible events section"*, removed it.

The removal does not look deliberate. The commit is about extensible events, and the same hunk also mangled the neighbouring "ring a user you share no room with" bullet, leaving an orphaned fragment that has since been cleaned up.

The text that was lost:

> - Encrypted rooms configured as `mentions only` are currently not sending push notifications for encrypted events. Hence the client would not ring even though the ring event contains `m.mentions`.
>   - As a stop gat, it is recommended, that the client sends unencrypted `m.rtc.notification` events in such rooms.
>   - As soon as [MSC3996: Encrypted mentions-only rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/3996) is supported `m.has_mentions` should be used instead of unencrypted call notify events.

The proposal now contains no occurrence of the word "encrypt" anywhere.

### Why it still matters

@Johennes reopened essentially the same question on 2026-05-27 and it has not been answered.

The delivery story in the current text is one bullet: *"The notification event is not filtered out by the current push rules."* In a room where every timeline event is `m.room.encrypted`, that bullet cannot distinguish a ring from a message:

- `.m.rule.call` has a single condition, `event_match` on `type` against `m.call.invite`. On the wire the type is `m.room.encrypted`, so the rule can never fire, and its `"sound": "ring"` tweak never reaches the gateway.
- What fires instead is `.m.rule.encrypted`, whose actions are `["notify"]` and nothing else, or `.m.rule.encrypted_room_one_to_one`, which adds sound `default`. The spec describes these as "all or nothing" matches.
- Since [only the highest-priority rule's actions are sent to the push gateway](https://spec.matrix.org/v1.19/client-server-api/#push-rules), a ring and an "ok" hand the gateway byte-identical actions.

A device that is asleep therefore receives the same push for both, and can only tell them apart after fetching and decrypting the event. That is workable, and it is roughly what Element X does on both platforms, but it is a real constraint on ring latency and reliability that implementers should not have to rediscover.

### Suggestion

Restore the point, either as the old Limitations section or as a paragraph under "Security considerations", which currently covers abuse vectors only. Even one sentence, saying that in encrypted rooms ringing depends on the client decrypting the event and that push rules cannot distinguish a ring from a message, would set the right expectation.

Happy to open a PR against the branch if that would be useful.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+  ...
+}
+```
+
+The schema of `m.rtc.decline` is as follows:
+
+- `m.relates_to` (required, object): An `m.reference` relation to the `m.rtc.invite` event which is
+  being declined.
+- `sticky_key` (required, string): The events sticky key as per [MSC4354]. MUST be equal to the
+  event ID of the event that is being declined.
```
robintown:
```suggestion
- `sticky_key` (required, string): The event's sticky key as per [MSC4354]. MUST be equal to the
  event ID of the `m.rtc.invite` event that is being declined.
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+[MSC4354]: https://github.com/matrix-org/matrix-spec-proposals/pull/4354
+
+### Sending invites
+
+Both `m.rtc.invite` and `m.rtc.decline` MUST be sent encrypted when the room is encrypted.
+
+A sending client SHOULD only consider an extended invite valid as long as all of the following
+conditions apply:
+
+- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the room
```
robintown:
```suggestion
- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the state of the room
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+- The invite is the current invite entry in the ephemeral sticky events map for the sender
+  and slot and not a withdrawal (that is, an invite event whose `content` is empty except
+  for `sticky_key`).
+- The client's current [push rules] produce an action of `notify` for the event.
+- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the room
+  where the invite was received.
+- The `lifetime`, as measured from `sender_ts` and capped to 2 minutes, has not elapsed. If
+  `sender_ts` is more than 20 seconds ahead of `origin_server_ts`, the `lifetime` SHOULD be
+  measured from `origin_server_ts` instead. This limits the impact of a malicious user faking
+  `sender_ts` to trigger long-lived notifications.
```
robintown: But then the remote server could choose to put `origin_server_ts` in the future as well, so I think this ultimately needs to be limited by the receipt time on the client. (Ideally it would be limited by the receipt time on the local server, so there's no risk of a notification that "comes back from the dead" every time you log in a new device… but an MSC fixing `age` / introducing a new `received_ts` thing is of course a whole nother project)

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+- `m.mentions` is either empty, missing or contains the client's user ID (either directly or
+  through a room mention).
```
robintown: Is this redundant to the "push rules produce an action of `notify`" condition?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+
+If the invite is valid, the receiving client has three options:
+
+1. It can accept the invite by joining the slot with an appropriate `m.rtc.member` event as
+   per [MSC4143]. Once the event is observed by other devices of the user, it invalidates the
+   invite.
+1. It can decline the invite by sending an `m.rtc.decline` event. Again, once the event is
+   observed by other devices of the user, it invalidates the invite.
+1. It can ignore the event by doing nothing. The invite will remain valid until either
+   the user accepts or declines the invite on another device or its `lifetime` has elapsed.
```
robintown, line 146: With legacy VoIP, Element clients had a variety of issues with stale invites coming back from the dead on client startup or in cases of poor push/sync performance. I see a potential for client devs to fall into very similar traps with `m.rtc.invite` if they implement the above bullet list naively, but I think we could significantly improve the odds of a decent implementation if we essentially fold the "three options" specification at the end (join/decline/timeout) into the bullet list, and then make the "three options" explanation non-normative.

Specifically the problem I see is that clients can have gappy syncs, syncs that contain not only an invite but also a join event followed by a leave event, etc., and it's not clear what it means to "observe" a join or decline event in all these cases. Therefore it might be better to phrase the conditions in terms of "what's currently sticky" and "what's in the ephemeral map", which are more well-defined.

I had written out more of my thinking here, but it's probably easiest if I send a PR to show what I mean?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+
+Again, how exactly receiving clients render invites in their UI is left as an implementation
+detail. A reasonable choice could, for instance, be to use a ringing UI in [direct chats] and
+a banner notification in group chats.
+
+Clients may also tweak their notification UI based on the referenced `m.rtc.slot` event and the
+`m.rtc.member` events currently joined to that slot. As an example, a client could choose to only
+ring for invites to slots hosting an `m.call` application. Similarly, the client could evaluate
+the `intent` property from [MSC4196] to display invites to audio calls differently than invites to
+video calls.
```
robintown, line 159: I guess if you invite someone to a call with no joined members, it's not clear which intent the call is meant to have. That could be an argument for allowing an `intent` in the notification again. Could also be a non-issue, depending on how you look at it.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+  }],
+  "actions": ["notify", {
+    "set_tweak": "sound",
+    "value": "ring"
+  }]
+}
+```
+
+`.m.rule.rtc.invite_for_room` matches `m.rtc.invite` events with the `room` property of `m.mentions`
+set to `true` (provided that the sender has the proper power level to trigger room notifications).
```
robintown, line 250: A downside of reusing `m.mentions` for invites is that the distinction between a `room: true` invite and an invite without `m.mentions` is pretty subtle, I don't know how you'd explain that in a user interface :confused:

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,382 @@
...
+
+Finally, in rooms that are neither set to muted nor mentions-only, `m.rtc.invite` events with `m.mentions`
+notify if the user is validly targeted via the event's `m.mentions` (via one of the two override rules)
+and also if the event has no or empty `m.mentions` (via the underride rule).
+
+| Push rule configuration | Invite with room mention | Invite with user mention | Invite without mention |
+| ----------------------- | ------------------------ | ------------------------ | ---------------------- |
+| Default | ✅ Notifies | ✅ Notifies | ✅ Notifies |
+| Mentions-only | ✅ Notifies | ✅ Notifies | ❌ Silent |
+| Muted | ❌ Silent | ❌ Silent | ❌ Silent |
```
robintown, line 300: How would a client implement a "mentions and calls only" setting for a room?

proposals/4075-rtc-notification-event.md
Johennes: > I had written out more of my thinking here, but it's probably easiest if I send a PR to show what I mean?

Yes, that would be awesome. I think I haven't yet fully understood the change you propose but it'll probably get clearer then.

proposals/4075-rtc-notification-event.md
Johennes: Hm, yeah. I think we'd have the same problem if we introduced a separate `invitees` property (similar to legacy VoIP's `invitee`)?

I would naively say that without mentions, a call shouldn't ring whereas with mentions it should. Maybe we could achieve that by removing the `ring` tweak from the underride rule? 🤔

proposals/4075-rtc-notification-event.md
Johennes: I think you'd just need a single `room`-kind rule that mutes anything that isn't caught by `override` rules. Then the two default override rules for `m.rtc.invite` and the two default override rules for `m.mentions` catch both calls and normal mentions.

Johennes: You could also achieve calls but no mentions otherwise by using the same `room`-kind rule and disabling the two default override rules for `m.mentions`.

Or, thirdly, no calls but using the `room`-kind rule and setting the actions on the two override rules for `m.rtc.invite` to `[]` (which means stop processing).

Do you think it would help to spell these combinations out in the proposal?

proposals/4075-rtc-notification-event.md
Johennes: Good point. I have integrated this into the potential issues section with 127034965f3ed2f0f460706067ea66f1b1e6953f.

proposals/4075-rtc-notification-event.md
Johennes: Hm, good point. The 2 minute cap would still apply there, too, though. So I think both malicious users and malicious servers could only cause 2-minute rings – which I think is fine?

proposals/4075-rtc-notification-event.md
Johennes: Yeah, something is wrong here. 🤔

I think we should separate validating invites from invite notifications. The latter only determine whether you get a ping / ring. But you should be able to also act on invites in a muted room when you observe them. So the validity rules here should only determine whether an invite is actionable for you.

For that, I think, we'd need the `m.mentions` condition because an invite directed at another user should not be actionable to you. But we should drop the `notify` condition from above because that only determines whether you were notified.

Does that make sense?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,387 @@
...
+
+- `slot_id` (required, string): The `state_key` of the `m.rtc.slot` event for which the invite is
+  handed out.
+- `sender_ts` (required, integer): The timestamp (in milliseconds since the epoch) when the sending
+  client created the event.
+- `lifetime` (required, integer): The time in milliseconds that the invite is valid for. MUST be
+  non-negative and SHOULD NOT be larger than 2 minutes. The RECOMMENDED value is 90 seconds.
+- `m.mentions`: (optional, object): A [mentions] object to optionally direct the invite at a set
+  of users in the room. If omitted or empty, the event is not targeted at specific users and may
+  be acted upon by any room member.
```
Johennes: @jmartinesp raised the question of how clients are supposed to decide if and which mentions to include. 

Naively, I think in DMs, you would probably always include the other user in the mention because you want to ping/ring them. In rooms, you'd use either a room-mention (if you want to ping/ring the room) or no mention at all (if you want to achieve a non-ringing / banner notification at most).

Allowing no mentions at all also helps still allow calls and invites when users lack the power level for room mentions.

All of this should probably be added to the proposal text.

proposals/4075-rtc-notification-event.md
robintown: If we switched to an `invitees` property, then we could rule out the existence of something like an `m.mentions: { room: true }` invite though, no? (Or are you imagining `invitees` would exist in addition to `m.mentions`?)

In my mind it would be ideal to have just two variants: invites with specific invitees, and invites directed at the general room. We shouldn't need a third variant that's more 'passive' than a room invite, because you can always just start a session without sending any invites. I'm not experienced enough with push rules to know whether that's doable…

proposals/4075-rtc-notification-event.md
robintown: That makes sense!

proposals/4075-rtc-notification-event.md
robintown: > The 2 minute cap would still apply there, too, though.

Ah, I guess I can read "The `lifetime`, as measured from `sender_ts` and capped to 2 minutes, has not elapsed" that way, yes. My initial reading led me to think this was instructing to use `stop_ringing_ts = origin_server_ts + min(lifetime, 2 minutes)` rather than `ring_duration = min(origin_server_ts + lifetime - now, 2 minutes)`.

proposals/4075-rtc-notification-event.md
robintown: Thanks, sounds plausible, I just wanted to check that it was possible with this push rule design, as I haven't personally studied up on push rules. Unless someone else has the same question, don't think there's a particular need to include this.

proposals/4075-rtc-notification-event.md
Johennes: Have made it so in 0a51a04434198d8aa2d16526e410eb246e88af40.

proposals/4075-rtc-notification-event.md
Johennes: Oh, now I see. Yes, that is easy to misread. Have rephrased it in 56b9c2888ba1a2af7ea40f2aa27fa3a01348428d to make it clearer.

proposals/4075-rtc-notification-event.md
Johennes: > In my mind it would be ideal to have just two variants: invites with specific invitees, and invites directed at the general room. We shouldn't need a third variant that's more 'passive' than a room invite, because you can always just start a session without sending any invites.

Oh. Yes, brilliant, I think you're exactly right! I have removed the `underride` rule in c10ee930852b391714011bd849ee814ada9eb101 and made `m.mentions` required. I think `invitees` would also work but I'd expect it to mostly just duplicate what `m.mentions` already supports. There's still the point that `invitees` would avoid overloading `m.mentions` though. I have included that as an alternative.

proposals/4075-rtc-notification-event.md
Johennes: I think this is now resolved by removing the underride rule and making mentions required.

proposals/4075-rtc-notification-event.md
Johennes: I think this is resolved as this MSC leaves rendering of invites (previously notifications) as an implementation detail and RTC sessions are explained in #4143.

proposals/4075-rtc-notification-event.md
Johennes: This has since been clarified in the text.

proposals/4075-rtc-notification-event.md
robintown: https://github.com/matrix-org/matrix-spec-proposals/pull/4539

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,389 @@
...
+clients to prompt other users to join a MatrixRTC session and to reject a received invitation, respectively.
+
+The schema of `m.rtc.invite` is as follows:
+
+- `slot_id` (required, string): The `state_key` of the `m.rtc.slot` event for which the invite is
+  handed out.
+- `sender_ts` (required, integer): The timestamp (in milliseconds since the epoch) when the sending
+  client created the event.
+- `lifetime` (required, integer): The time in milliseconds that the invite is valid for. MUST be
+  non-negative and SHOULD NOT be larger than 2 minutes. The RECOMMENDED value is 90 seconds.
```
robintown, line 31: Now that I think about it, is there a particular reason to have a `lifetime` field in addition to the sticky duration? (Assuming senders literally set the sticky duration to the desired lifetime.)

Johennes: Oh. Excellent question! Hm, I guess this probably stems from `lifetime` on the legacy VoIP invite event. 🤔 

I suppose dropping `lifetime` would make all the conditions a little easier. The lifetime being exceeded basically becomes the event stopping to be sticky. I also cannot think of a reason why you would ever want to set a different lifetime and stickyness duration. If it's sticky for too long, the invite expires anyway. If it's sticky for too short, some clients won't see it. So yeah, I think this is a good idea and probably helps simply things a bit. 👍

Johennes: In starting to make this change, I realised that it would not allow us to measure the lifetime from `sender_ts` anymore because the event stops being sticky based on `origin_server_ts`. I am not sure if we really need the `sender_ts` logic or not. For now, I have added an alternative describing how to use the sticky duration as the lifetime in 78a3f75f42d58b3b357d1b8571c91d45fc1bd5ac. We can still figure out which variant we prefer in SCT review.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+## Proposal
+
+Two new room events `m.rtc.invite` and `m.rtc.decline` are introduced. These events can be sent by
+clients to prompt other users to join a MatrixRTC session and to reject a received invitation, respectively.
+
+The schema of `m.rtc.invite` is as follows:
+
+- `slot_id` (required, string): The `state_key` of the `m.rtc.slot` event for which the invite is
+  handed out.
+- `sender_ts` (required, integer): The timestamp (in milliseconds since the epoch) when the sending
```
gewitternacht: I'm no timestamp expert, but I suppose it should be specified which epoch? :D

```suggestion
- `sender_ts` (required, integer): The timestamp (in milliseconds since the Unix epoch) when the sending
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+  client created the event.
+- `lifetime` (required, integer): The time in milliseconds that the invite is valid for. MUST be
+  non-negative and SHOULD NOT be larger than 2 minutes. The RECOMMENDED value is 90 seconds.
+- `m.mentions`: (required, object): A [mentions] object to direct the invite at either a set
+  of users in the room or the entire room. One of the two options MUST be used. An `m.mentions`
+  object with both `room` missing or set to false and `user_ids` missing or empty is invalid.
```
gewitternacht: Generally, such empty `m.mentions` objects are not invalid – the linked part of the spec even mentions "If there are no mentions to include it can be an empty object."

So is this supposed to say that
> An `m.mentions` object with both `room` missing or set to false and `user_ids` missing or empty _makes the `m.rtc.invite` event invalid._

?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+- `sticky_key` (required, string): The event's sticky key as per [MSC4354]. MUST be equal to `slot_id`.
+  This ensures that receivers only maintain one active invite per slot and sender.
```
gewitternacht, line 37: I was very confused for a while as to which event's sticky key this is referring to, but I guess the `m.rtc.invite` event _itself_ is a sticky event, and it gets assigned a `sticky_key` of `slot_id`? If yes, it would be good to mention that these events (and `m.rtc.decline` events) are sticky events at some earlier point.

[after having read further along: oh, it's mentioned in line 72 below – a little late for my tastes :D]

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+    "sticky_key": "$1"
+  }
+}
+```
+
+Clients MUST send both `m.rtc.invite` and `m.rtc.decline` as sticky events as per [MSC4354] for the
+associated delivery guarantee. The sticky durations of `m.rtc.invite` events, `m.rtc.member` events
+which accept an invite, or `m.rtc.decline` events which decline an invite SHOULD NOT be smaller than
+the invite's `lifetime`. Additionally, clients MUST implement the ephemeral map algorithm as per
+[MSC4354] to construct a state-like store of invite events. Tracking decline events in a map isn't
```
gewitternacht:
```suggestion
[MSC4354] to construct a state-like store of `m.rtc.invite` events. Tracking `m.rtc.decline` events in a map isn't
```

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+necessary because there is no need to update them after being sent.
+
+Additionally, both `m.rtc.invite` and `m.rtc.decline` MUST be sent encrypted when the room is encrypted.
+
+[mentions]: https://spec.matrix.org/v1.19/client-server-api/#user-and-room-mentions
+[MSC4354]: https://github.com/matrix-org/matrix-spec-proposals/pull/4354
+
+### Sending invites
+
+A sending client SHOULD only consider an extended invite valid as long as all of the following
```
gewitternacht: It might be me, but I'm not used to "extended" being used like this: I suppose this concerns invites that were "issued/handed out/sent" by the sending client, and not "prolonged/made to be valid for a longer time" (which would be my first interpretation of "extended" here)?

(the same use occurs several times below)

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+conditions apply:
+
+- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the state of the
+  room where the invite was sent.
+- The invite's `lifetime`, as measured from `sender_ts` and capped at 2 minutes, has not elapsed.
```
gewitternacht, line 93: Does "capped at 2 minutes" mean something like: for the invite event to be valid, it should hold that 
- `current time < sender_ts + min(lifetime, 2 minutes)` ?

Is this meant to protect against maliciously high `lifetime`s? If yes, should there also be a mechanism for a maliciously set `sender_ts`, like making some kind of plausibility check, or adding the requirement `sender_ts < current time` in the above invitation?

Having read further along: This is actually mentioned for _Receiving invites_ below and in the Security Considerations, but I think it should be mentioned here too. Otherwise, combined with "senders SHOULD NOT emit invites when another valid invite exists", I think this would enable malicious room members to "block" anyone else from sending out legitimate invites by specifying a `sender_ts` far in the future.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+and an otherwise empty content and the same or a larger sticky duration. Alternatively, the event
+MAY also be redacted which will remove it from the ephemeral map of sticky events.
+
+To prevent stale invitations, clients SHOULD withdraw previously extended invitations when the
+inviter themselves leaves the slot with no other members currently joined to the slot.
+
+How exactly sending clients present extended invitations in their UI is left as an implementation
+detail. For instance, a sending client could use a ringing UI in [direct chats] while it is waiting
+for the invite to be acted on and stop ringing when the invite is accepted or declined (see the next
+section).
```
gewitternacht: > stop ringing when the invite is accepted or declined

... or becomes invalid after `lifetime` or 2 minutes elapsed?
(not sure if this makes sense)

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+the semantics of `m.mentions`. Additionally, a power level flag could be introduced that
+allows to configure the threshold for MatrixRTC room invites separately from the threshold
+for room mentions on other messages. It's unclear what use cases would require this, however.
+Additionally, an `invitees` property would largely duplicate what `m.mentions` already supports.
+
+### Treating sticky duration as lifetime
+
+The lifetime of `m.rtc.invite` could also be inferred from their sticky duration which would
+eliminate the need for an explicit `lifetime` property. This would make it impossible to measure
+the lifetime from `sender_ts`, however, to account for clock-drift between clients and servers.
```
gewitternacht, line 383: I don't get this. Maybe because it's too vague how the sticky duration would be used, or I don't understand what it means to "measure the lifetime from `sender_ts`"? In any case, I think I'd need an explicit "compare timesptamp x with timestamp y" for both approaches to really understand this.

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+
+## Security considerations
+
+### Inviting without a valid `m.rtc.member` event
+
+Under this proposal, the sender of `m.rtc.invite` events does not need to be joined to the slot
+themselves in order to make the invite valid. This may seem like a potential abuse vector. However,
+room members who are able to send `m.rtc.invite` events will commonly also be able to send
+`m.rtc.member` events. Thus requiring a member event to send invites doesn't provide additional
```
gewitternacht: What is "commonly"? Should this be added as a recommendation in this MSC, like "the PL to send `m.rtc.invite` events is recommended to be at least as high as the PL for `m.rtc.member` events"? But I'm not sure how useful this is and who this recommendation would even address... Perhaps it could be included in the next sentence addressing room administrators?

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+protection. The actual access control for sending invites is the power level for sending
+`m.rtc.invite` events which room administrators can raise as needed. When power levels allow
+sending invitations, users can still mitigate abusive invites by configuring push rules accordingly.
+
+Besides this, requiring the sender to have an `m.rtc.member` event also complicates invite processing.
+Particularly on mobile, it would mean that after receiving a push notification, the client would
+have to fetch a state event and two different room events, including the information required to
```
gewitternacht: Am I correct in thinking that even now, it has to fetch a state event and _one_ room event? (The wording makes it sound a bit like currently, no fetching of events at all is needed.)

proposals/4075-rtc-notification-event.md
```diff
@@ -0,0 +1,461 @@
...
+`lifetime` of 2 minutes.
+
+Similarly, a malicious server could fake `origin_server_ts` to be in the future to cause long-lived
+notifications. Again, the client-side 2 minute cap for the remaining lifetime acts as a mitigation.
+
+### Room-level invites
+
+Using room mentions on `m.rtc.invite` events will notify all users in a room. This has the same
+potential for abuse as room mentions on normal messages and can be mitigated by increasing the
+power level required to send room notifications.
```
gewitternacht, line 425: This is mentioned several times in this MSC, but (having no prior experience with mentions) how do power levels work here, i.e., who checks them at which point? Maybe I'm not looking properly, but I don't see this mentioned in https://spec.matrix.org/v1.19/client-server-api/#user-and-room-mentions; and there's a [`notifications.room`](https://spec.matrix.org/v1.19/client-server-api/#mroompower_levels_notifications) property in `m.room.power_levels`, but how does it take effect?

proposals/4075-rtc-notification-event.md
Johennes: Fair, yes. 😅

proposals/4075-rtc-notification-event.md
Johennes: Yes, indeed. I can see how that could be misread.

```suggestion
  of users in the room or the entire room. One of the two options MUST be used. An `m.mentions`
  object with both `room` missing or set to false and `user_ids` missing or empty makes the
  `m.rtc.invite` event invalid.
```

proposals/4075-rtc-notification-event.md
Johennes: Oh, yeah it's mentioned afterwards only. I have made a small note of it before this part in c5af560fed4e39ff8bdcb74b6601856f51fe4b46.

proposals/4075-rtc-notification-event.md
Johennes: Yeah, you probably have a point. Have replaced the occurrences in 5c790cefcb356db420515951f7e24f2fc025675e.

proposals/4075-rtc-notification-event.md
Johennes: Yes, good point.

```suggestion
detail. For instance, a sending client could use a ringing UI in [direct chats] while it is waiting
for the invite to be acted on and stop ringing when the invite is accepted, declined (see the next
section) or expires.
```

proposals/4075-rtc-notification-event.md
Johennes: Yes, indeed. Have tried to make that clearer in 22de2a6b0cc24c4ae32ec4bc0767a9dbbd9952c9.

proposals/4075-rtc-notification-event.md
Johennes: The "commonly" was supposed to mean that often you would _not_ set dedicated power level for specific event types and then all events would be governed by the `events_default` level. But also _if_ you set power levels for these events, it seems natural to set both to the same value unless you really want to prevent people from inviting / ringing others.

Johennes: Have tried to make it a little clearer in c0945d2ca3dd036bd560e7852a04314e3c3621f9.

proposals/4075-rtc-notification-event.md
Johennes: Have tried to make it clearer in d823ff0d7d0d3bae278afc53666196a6aa850a71.

proposals/4075-rtc-notification-event.md
Johennes: Looks like this isn't spelled out very well in the spec apart from this warning:

> If an encrypted event contains an `m.mentions` in its payload, it should be encrypted as normal. To properly process mentions in encrypted rooms, events must be decrypted first. See [receiving notifications](https://spec.matrix.org/v1.19/client-server-api/#receiving-notifications).

Due to encryption, the server doesn't check mentions when it checks power levels in step in step 8 of the [authorization rules](https://spec.matrix.org/v1.19/rooms/v12/#authorisation-rules). That means the `room` power level is only checked by receiving clients via their push rules. In other words, events with room mentions are allowed by servers even if the sender doesn't have the room power level but receivers don't allow it to trigger notifications.

I think this is a spec clarity problem rather than a problem of this proposal.

Johennes: Have opened https://github.com/matrix-org/matrix-spec/issues/2443.

proposals/4075-rtc-notification-event.md
Johennes: Good point, yes. Have applied the same logic already used for receivers in 78686c55cb63d2ab7186517e45e05bf8b5227a45.

proposals/4075-rtc-notification-event.md
robintown: I see. My take is that whatever system for coordinating timings that works for sticky events in general ought to be good enough for invites as well. They added a `sticky_duration_ttl_ms` field to the sticky events spec, which I rather like as it allows the client to not deal with clock timestamps at all, unlike the `sender_ts` solution which just trades one clock skew for another.

(Fine with leaving this until SCT review, just wanted to share my thoughts.)

Johennes: Yeah, I'm not terribly fond of `sender_ts` either. I'd expect it won't survive SCT review but wanted to keep as close as reasonably possible to the current implementation so that we can meet the implementation requirements more easily.

proposals/4075-rtc-notification-event.md
Johennes, line 1: Implementation requirements:

- [x] Client
- [x] Server

Johennes: > [!NOTE]
> For historical reasons, the implementations use the unstable identifiers as listed in the proposal and contain relations and properties such as `intent` that have been removed from the proposal.

### synpase

- https://github.com/element-hq/synapse/pull/20227

### matrix-js-sdk

- Initial implementation (outdated format and validation rules): https://github.com/matrix-org/matrix-js-sdk/pull/4826
- Invite validation functions & sending invites and declines sticky: https://github.com/matrix-org/matrix-js-sdk/pull/5544

### element-web

- Ringing: https://github.com/matrix-org/matrix-react-sdk/pull/11870
- Applying invite validation & synchronising push rules: https://github.com/element-hq/element-web/pull/35089

PR #4076, opened 2023-11-12
MSC4076: Let E2EE clients calculate app & E2EE-room badge counts themselves

ara4n:


proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,127 @@
+# MSC4076: Let E2EE clients calculate app badge counts themselves
```
TheArcaneBrony, line 1: I fail to see how you couldn't already calculate them clientside?

richvdh: Agreed. TBH I've only skimmed this, but I don't entirely follow why this MSC is necessary for clients to calculate their own badge counts.

(You might justifiably argue that the counts should be removed on the grounds of redundancy, but that doesn't appear to be what this MSC is saying?)

ara4n: You can and should calculate them clientside. However, the server currently overrides them and there is no way to stop the server from doing that, prior to this MSC.

MTRNord: Clearly some crucial information is missing here? Since to my understanding you can just make either fully ignore the server response and do it without it or have an if case for only e2ee rooms. I don't see how that currently impossible? The server isn't demanding a particular execution path on the client obviously?

clokep: > However, the server currently overrides them and there is no way to stop the server from doing that, prior to this MSC.

Is this done at the OS-level and the server needs to not include that information to not fight the client doing it?

ara4n: > Is this done at the OS-level and the server needs to not include that information to not fight the client doing it?

yes.

proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Solution
+
+When a pusher is registered for a given device via `POST /_matrix/client/v3/pushers/set`, we add a new boolean field
+to the request called `e2ee_badge_aware` which defaults to `false` if absent.
+
+If `e2ee_badge_aware` is set to true, then the pusher will not specify `unread` or `missed_calls` in the
+`POST /_matrix/push/v1/notify` request to the target push gateway, and so not override the client's app badge.
+
+We also add a new optional field to read receipts called `cleared_notifs` which defaults to `false` if absent.  This is
```
manuroe, line 53: How a matrix client can know that the user is using a `e2ee_badge_aware` client and, thus, it should add `cleared_notifs` to read receipts it sends?

Why a client should care about the "e2ee_badge_awareness" of other clients used by the user?
Instead, should the server do that automatically for those e2ee_badge_aware clients and only for them?

ara4n: > How a matrix client can know that the user is using a e2ee_badge_aware client and, thus, it should add cleared_notifs to read receipts it sends?

The intention here was that the optional field would only be set on RRs sent by e2ee-badge-aware clients, which obviously know that they are e2ee-badge-aware already. 

I think the problem here is that the MSC misses the rationale as to why this is a good idea:

Previously, the badge count would reduce on other clients because the server would calculate it and push it to them.  However, by switching to having the server no longer calculate the count for given clients, it needs another way to reduce their count, to avoid stuck notifs.  So instead it sends them a 'blank' push to prod them to recalculate their counts.

manuroe: thanks for the clarification. 
(this is now anyway a topic of https://github.com/matrix-org/matrix-spec-proposals/pull/4234).

proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,106 @@
...
+gateways in the Push API is currently wrong[^3]: it says it's "The number of unread messages a user has across all of
+the rooms they are a member of." but in practice it's "The number of rooms the user has with unread push
+notifications" (although synapse has a config option to change that, confusingly). So we should fix that too.
+
+## Solution
+
+When a pusher is registered for a given device via `POST /_matrix/client/v3/pushers/set`, we add a new boolean field
+to the request called `disable_badge_count` which defaults to `false` if absent.
+
+If `disable_badge_count` is set to true, then the pusher will not specify `unread` or `missed_calls` in the
```
manuroe, line 51: In this case, can we directly remove `counts`, the parent of `unread` and `missed_calls` from [`POST /_matrix/push/v1/notify`](https://spec.matrix.org/v1.12/push-gateway-api/#post_matrixpushv1notify)?

manuroe:
Synapse PR: https://github.com/element-hq/synapse/pull/17975

It removes the parent `counts` field as proposed in https://github.com/matrix-org/matrix-spec-proposals/pull/4076#discussion_r1863495327. I will undraft the PR once we agree on the spec.

proposals/4076-e2ee-aware-badges.md
ara4n: fine by me.

ara4n:
>  I will undraft the PR once we agree on the spec.

I am not seeing any disagreements on the spec?  There is no reason for the PR to be draft if you think it's good for review.  The only thing stopping FCP on this MSC right now is the lack of implementation, so please don't block the implementations on the MSC getting FCP'd ;)

manuroe:
> > I will undraft the PR once we agree on the spec.
> 
> I am not seeing any disagreements on the spec? There is no reason for the PR to be draft if you think it's good for review. The only thing stopping FCP on this MSC right now is the lack of implementation, so please don't block the implementations on the MSC getting FCP'd ;)

I wanted to have a Synapse PR that matches the MSC. With your last commit on this MSC, the Synapse PR can be reviewed. Thanks!

proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,106 @@
...
+track their own accurate `unread_notification` count for E2EE rooms (while still using the `unread_notification` count
+in the /sync response for non-E2EE rooms if they want), and from that calculate an accurate app badge count
+clientside.
+
+The problem is that the server still tries to override the app badge count whenever it sends a push to the client though,
+which at best will be wrong (either understeering due to missing E2EE mentions, or oversteering due to alerting about
+rooms with invisible events) - and at worst will race and fight[^2] with the app badge count maintained locally by the
+app, if any.
+
+Therefore we need a way for sophisticated E2EE clients to tell the server to stop overriding their app badge count.
```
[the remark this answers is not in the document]
ara4n: yes, it's done at the OS level, so the app doesn't get the choice.

I'm not aware of a standard way to configure push gateways for clients in general to say "please don't try to override the app badge count" - wouldn't doing it in a generic way end up being precisely the same as this MSC?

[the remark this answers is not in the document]
reivilibre: The difference is that the application developer is the one deploying Sygnal, therefore they can do whatever they want in the 'last mile' between Sygnal and their application. If that means Element configures their Sygnal so that it doesn't ship badge counts to Element iOS clients, then there's no particular reason anyone else needs to be involved.
To me, it feels like a better fit for a platform-specific quirk.

jmartinesp:
I may have missed some discussion in the PR, but when the MSC says 'we add a new boolean field to the request called `disable_badge_count`', should that field be added to the root object, like this?

```json
{
  "pushkey": "-",
  "app_id": "foo.bar",
  "kind": "http",
  "data": {
    "url": "https://matrix.org/_matrix/push/v1/notify",
    "format": "event_id_only",
    "default_payload": {
      "cs": "-"
    }
  },
  "app_display_name": "MyApp",
  "device_display_name": "MyDevice",
  "profile_tag": "mobile_",
  "lang": "en",
  // Here
  "org.matrix.msc4076.disable_badge_count": true
}
```

Or is it supposed to be inside the `data` payload, like this?

```json
{
  "pushkey": "-",
  "app_id": "foo.bar",
  "kind": "http",
  "data": {
    "url": "https://matrix.org/_matrix/push/v1/notify",
    "format": "event_id_only",
    // Here
    "org.matrix.msc4076.disable_badge_count": true,
    "default_payload": {
      "cs": "-"
    }
  },
  "app_display_name": "MyApp",
  "device_display_name": "MyDevice",
  "profile_tag": "mobile_",
  "lang": "en"
}
```

proposals/4076-e2ee-aware-badges.md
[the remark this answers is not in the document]
richvdh: > This document doesn't answer what feels like a fair question to have: Why do the sophisticated E2EE clients not just ignore the count?

(comment moved to https://github.com/matrix-org/matrix-spec-proposals/pull/4076#discussion_r1390329111 to make way for discussion about whether this field is necessary)

proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,106 @@
...
+[MSC4234](https://github.com/matrix-org/matrix-spec-proposals/pull/4234)
+ 
+Finally, a separate but related problem is that historically the description of the `unread` count passed to push
+gateways in the Push API is currently wrong[^3]: it says it's "The number of unread messages a user has across all of
+the rooms they are a member of." but in practice it's "The number of rooms the user has with unread push
+notifications" (although synapse has a config option to change that, confusingly). So we should fix that too.
+
+## Solution
+
+When a pusher is registered for a given device via `POST /_matrix/client/v3/pushers/set`, we add a new boolean field
```
richvdh, line 48: Please link to the existing spec, to help readers who don't have this context at their fingertips:

```suggestion
When a pusher is registered for a given device via [`POST /_matrix/client/v3/pushers/set`](https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3pushersset), we add a new boolean field
```

proposals/4076-e2ee-aware-badges.md
richvdh: This has come up [again](https://github.com/matrix-org/matrix-spec-proposals/pull/4076#discussion_r1866071915), so even though the answer is kinda there in the first sentence and footnote, I don't think hit has been satisfactorily resolved.

I'd suggest rephrasing the introduction to give a bit of background for those who can't remember whether the A in APNS means "Apple" or "Android", far less how APNS actually works.

> Apple push notification service (APNS) allows the badge count on applications to be updated based on a field in push notifications, without the application running any code (see [APNS documentation](https://developer.apple.com/documentation/usernotifications/generating-a-remote-notification#2943362)). Push notifications in Matrix currently use this property; however, its behaviour in encrypted clients is unsatisfactory due to the following reasons: 
> * ... etc

proposals/4076-e2ee-aware-badges.md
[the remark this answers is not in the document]
richvdh: > The difference is that the application developer is the one deploying Sygnal, therefore they can do whatever they want in the 'last mile' between Sygnal and their application. If that means Element configures their Sygnal so that it doesn't ship badge counts to Element iOS clients, then there's no particular reason anyone else needs to be involved.
> To me, it feels like a better fit for a platform-specific quirk.

Emphatic +1 to this. Currently, Sygnal is actively copying the `counts.unread` field into the right place in the push notification payload. If we now feel that this is inappropriate for Element apps then... just don't do that? Sygnal already has all the information it needs to make this decision; we don't need to complicate the C-S API, homeservers, and Push Gateway API by bouncing it through Synapse.

proposals/4076-e2ee-aware-badges.md
```diff
@@ -0,0 +1,106 @@
...
+to the request called `disable_badge_count` which defaults to `false` if absent.
```
richvdh, line 49: As noted by @jmartinesp [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4076#issuecomment-3335011626), this is unclear: is `disable_badge_count` meant to be at the top level or under `data`?

richvdh: The question is really: does this only affect the `http` push kind, or would it also affect other push kinds (specifically, email).

It looks like it would be somewhat specific to the http `kind`. It also feels somewhat analagous to `format` (see https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3pushersset_request_pusherdata) (and, for that matter, `uri`), and so should be in `data`.

[Honestly though, this API is *horrible*, in that `data` conflates "parameters to the homeserver" (`uri`) and "opaque data to pass through to the push gateway" (`default_payload`, for example).]

richvdh:
> I may have missed some discussion in the PR, but when the MSC says 'we add a new boolean field to the request called `disable_badge_count`', should that field be added to the root object, like this?

I opened a [thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4076#discussion_r2407115996) about this.

PR #4078, opened 2023-11-14
Transparent pusher creation

stefanceriu:


proposals/4078-transparent-pusher-creation.md
```diff
@@ -0,0 +1,29 @@
+# MSC4078: Registering pushers against push notificatiom services should forward back failures
```
clokep, line 1:
```suggestion
# MSC4078: Registering pushers against push notification services should forward back failures
```

proposals/4078-transparent-pusher-creation.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Background
+
+It's quite common for push notification services to throw back various errors when trying to register a pusher. [The spec](https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3pushersset) doesn't currently specify that those errors should be sent back to the original caller, letting it believe everything is fine. It handles missing parameters in the request body and rate-limiting but nothing coming back from the push notification service.
```
clokep, line 5: Linking to a version of the spec is preferred so that it is a permalink:

```suggestion
It's quite common for push notification services to throw back various errors when trying to register a pusher. [The spec](https://spec.matrix.org/v1.8/client-server-api/#post_matrixclientv3pushersset) doesn't currently specify that those errors should be sent back to the client, letting it believe everything is fine. It handles missing parameters in the request body and rate-limiting but nothing coming back from the push notification service.
```

proposals/4078-transparent-pusher-creation.md
```diff
@@ -0,0 +1,29 @@
...
+
+The push notification service can decide to deny requests for a variety of reasons ranging from bad request and internal errors to certificate and device token expirations e.g. [APNs errors](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns)
+
+## Problem
+
+We are getting reports that users are missing push notifications entirely which we tracked down to APNs returning 410s behind the scenes. As the application is never informed about this there is no straight forward way for it to try to re-register a pusher with a new device token. The application does try to update its pusher every time it becomes active but that's no good if the token it's using is always the same.
+
+## Proposal
+
+The most straight forward solution would be to directly expose the underlying errors on POST `/pushers`. We cannot know beforehand what the final service used is so we shouldn't attempt to wrap errors and just rely on HTTP semantics.
```
clokep, line 15: I think I'm not *quite* following what's happening; is it the following:

1. The client configures a pusher.
2. The homeserver happily uses that pusher for a bit.
3. For some reason, the push service starts erroring.
  1. The error gets receive by the push gateway which passes it on.
  2. The homeserver receives the error and deletes the pusher.
4. The user receives no more pushes.
5. On next app launch (?) we go back to 1.

Is that accurate or is something else happening?

I'm unsure why `POST /pushers/set` is the proper place to return an error? How would the homeserver even track that?

stefanceriu: > On next app launch (?) we go back to 1.

yes, but this time round the the token will be reported as invalid by the APNs and the error will be forwarded back from synapse to the client. The client can then requests a new token from Apple and start again from 1.

> I'm unsure why POST /pushers/set is the proper place to return an error? How would the homeserver even track that?

talked about this offline: we believe the best solution is to check that the pusher config is actually valid as soon the creation request comes in as opposed to deferring it until trying to send the first push and then failing silently.

proposals/4078-transparent-pusher-creation.md
```diff
@@ -0,0 +1,44 @@
...
+## Problem
+
+We are getting reports that users are missing push notifications entirely which we tracked down to APNs returning 410s
+behind the scenes. As the application is never informed about this there is no straight forward way for it to try to
+re-register a pusher with a new device token. The application does try to update its pusher every time it becomes active
+but that's no good if the token it's using is always the same.
+
+## Proposal
+
+The most straight forward solution would be to directly expose the underlying errors on POST `/pushers`. We cannot know
```
MTRNord, line 23: I am not sure if the actual error of the backend is actually useful for the client. Not only is it not helpful for the enduser if it seems them but also for the app it might be a variety of used backends so error handling probably ends up being a catch all regardless. I think a generic "error happened" is enough here unless I miss some usecase here.

stefanceriu: I'm not sure either to be honest. For APNs for example there's quite a big difference between a recoverable 410 and a completely unrecoverable 403, we wouldn't want the app to keep trying over and over again for no reason..

MTRNord: I guess replicating the temp fail vs permanent fail could make sense yeah. That allows for a banner like "notification is currently unavailable" or a automatic preregister attempt.

clokep: There's also quite a few layers to pass the error back through, e.g. push server -> push gateway -> homeserver -> client. This will need care to not mangle it.

proposals/4078-transparent-pusher-creation.md
```diff
@@ -0,0 +1,43 @@
...
+beforehand what the final service used is so we shouldn't attempt to wrap errors and just rely on HTTP semantics.
```
turt2live, line 23: The error shouldn't be exposed directly, and should be wrapped. Something like this might be fine enough:

```json5
{
  "errcode": "M_UNKNOWN",
  "error": "Unable to register pusher",
  "upstream_errcode": "410", // intentionally a string, to allow for `GOOG_FAIL` or whatever else goes on in the wild
  "upstream_error": "Words from the upstream about what a 410 actually means"
}
```

...with a 400 status code. This gives clients (hopefully) enough information to do something about it, and can still be sanitized by the server so it's consistent with the remainder of the protocol. This also allows a server to change the error, though it should only be doing this to remove accidentally disclosed credentials. 

Would be worth saying that `upstream_error` may not be a human readable string, such as HTML, XML, or encoded binary depending on the upstream system.

turt2live (changes requested):
the idea overall seems sensible to me, though I think we should be producing standardized Matrix errors

proposals/4078-transparent-pusher-creation.md
stefanceriu: Oh, I like that, I'll make the change.

Thanks for looking at it!

turt2live (approved):
looks good to me for implementation purposes

PR #4079, opened 2023-11-14
MSC4079: Server-Defined Client Landing Pages

williamkray:
[Rendered](https://github.com/williamkray/matrix-spec-proposals/blob/williamkray/client-server-landing-page-html/proposals/4079-client-server-landing-page-html.md)

Signed-off-by: William Kray <msc@williamkray.com>

### DIFF
--- a/proposals/4079-client-server-landing-page-html.md
+++ b/proposals/4079-client-server-landing-page-html.md
@@ -51,6 +51,9 @@ MSC](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/176
   clients may choose to load and reload this content more frequently.
 - Clients may choose to implement this feature as a "home" button or as default content in the main
   view when no conversation is active.
+- Clients that support this functionality must at minimum support a plaintext formatting for minimum
+  functionality. Additional formats and mimetypes are optional but may provide a superior user
+  experience.
 
 ## Examples
 


proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,69 @@
...
+Example formatting using in-line HTML:
+
+```
+{
+  "m.homeserver": {
+    "user_home_page_html": "<h1>Welcome to our Matrix Homeserver!</h1><p>Visit our website to make a donation.</p>"
+  }
+}
+```
+
```
MTRNord, line 52: How about mxc URLs like used elsewhere?

proposals/4079-client-server-landing-page-html.md
turt2live, line 1: @williamkray please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes for this MSC to be eligible for approving FCP, and limit lines to approximately 100 characters.

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,69 @@
...
+  - Changelogs for server updates can be displayed, helping users to understand recent changes without external navigation.
+  - Links to other supported clients may be supplied in case users are looking for alternative interfaces.
+
+## Specification
+
+  - A new optional field will be defined within the well-known Matrix configuration for clients: user_home_page.
+  - The user_home_page field can be defined by either in-line HTML content or a fully-qualified domain name and path to an HTML document.
+  - The HTML content will be sanitized by the client and restricted to the subset of HTML currently allowed for messages.
+  - This field can be queried by clients during the login process or refreshed periodically.
+  - Clients may choose to implement this feature as a "home" button or as default content in the main view when no conversation is active.
```
joepie91: This should be more clearly specified; *must* a client implement this functionality, with a choice between these two options? Or are these merely two suggested representations, and may a client also choose not to implement it at all? Are they allowed to implement it in a manner not specified here?

The field is specified as 'optional' further up in the MSC, but that seems to only describe whether the server is required to *provide* such content, not whether the client is required to *consume* it.

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,69 @@
...
+  - Announcement of new features or user guides following an update can be nicely formatted and made readily available upon client login.
+  - Changelogs for server updates can be displayed, helping users to understand recent changes without external navigation.
+  - Links to other supported clients may be supplied in case users are looking for alternative interfaces.
+
+## Specification
+
+  - A new optional field will be defined within the well-known Matrix configuration for clients: user_home_page.
+  - The user_home_page field can be defined by either in-line HTML content or a fully-qualified domain name and path to an HTML document.
+  - The HTML content will be sanitized by the client and restricted to the subset of HTML currently allowed for messages.
+  - This field can be queried by clients during the login process or refreshed periodically.
```
joepie91: This should be more specific; leaving the refresh policy undefined is likely to result in situations where some people inexplicably see outdated (and now-inaccurate) information, and it's not clear how to resolve it because it would depend on the client.

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,69 @@
...
+## Abstract
+This proposal extends the Matrix protocol to allow servers to define custom HTML-formatted documents that can be rendered by clients as a user home page. This feature introduces a non-intrusive yet flexible manner to communicate server-specific information such as donation requests, scheduled maintenance, updates, new features, and changelogs.
+
+## Background
+Currently, there is no standardized method in Matrix for server admins to communicate information directly within user clients. While rooms can serve this purpose, they lack the persistence and prime visibility that many server administrators desire.
+
+## Proposal
+Introduce an optional configuration for matrix servers that allows them to specify an HTML-formatted document, adhering to the current subset of HTML utilized for message formatting. Clients can render this document in a dedicated "home" view or as a placeholder page when no active conversations are selected.
+
+## Use Cases
```
joepie91, line 17: I think an additional benefit of something like this, would be that it can improve on the relation between a homeserver operator and their users; right now, homeservers are sort of "homogeneous", in that people use them like a neutral service that just magically exists and they don't know their admin, which can cause problems for sustainability (something that eg. Mastodon does a lot better at).

Having server-specific homepages to introduce people to the community on that homeserver, might help to create a bit more of a social connection to that specific homeserver, making it more appealing for people to eg. run small semi-public homeservers.

proposals/4079-client-server-landing-page-html.md
williamkray: i honestly don't know, i feel like clients may want to choose whether to support this feature. it's not critical to the functionality, and server operators may opt to only support or deploy clients that do support this feature. i don't think it should be a showstopper for a client to be considered fully-baked. but i'm open to feedback.

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,76 @@
...
+## Background Currently, there is no standardized method in Matrix for server admins to communicate
+information directly within user clients. While rooms can serve this purpose, they lack the
+persistence and prime visibility that many server administrators desire.
+
+## Proposal Introduce an optional configuration for matrix servers that allows them to specify an
+HTML-formatted document, adhering to the current subset of HTML utilized for message formatting.
+Clients can render this document in a dedicated "home" view or as a placeholder page when no active
+conversations are selected.
+
+## Use Cases
```
uhoreg, line 23: Some of these could be done using [Server Notices](https://spec.matrix.org/v1.8/client-server-api/#server-notices)

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,76 @@
...
+Example formatting using in-line HTML:
+
+``` { "m.homeserver": { "user_home_page_html": "<h1>Welcome to our Matrix Homeserver!</h1><p>Visit
+our website to make a donation.</p>" } } ```
+
+Only one value is needed. If both in-line and URI definitions are defined, clients will prioritize
+the in-line HTML.
+
+
+## Security Considerations To mitigate any potential security risks from malicious content:
```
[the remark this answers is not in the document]
williamkray: fixed, i'm trying to do too many things at once lol

williamkray:
i think html is already a pretty widely accepted expectation and is included throughout the spec already.

deepadmax:
> External resources (e.g., images, stylesheets) MUST NOT be fetched by default to avoid privacy leaks and must be explicitly allowed by the client.

The standard Matrix media repository would be a good fit for this. The client would only ever have to communicate with the homeserver. The homeserver would serve the contents of the homepage to the user directly, as it does with everything else.

proposals/4079-client-server-landing-page-html.md
williamkray: mxc URI support has been added below to serve media and files, while excluding attempts to serve scripts or CSS for security purposes (at this point in time)

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,124 @@
...
+  readily available upon client login.
+- Changelogs for server updates can be displayed, helping users to understand recent changes without
+  external navigation.
+- Links to other supported clients may be supplied in case users are looking for alternative
+  interfaces.
+
+## Specification
+
+- A new optional field will be defined within the well-known Matrix configuration for clients:
+  user_home_page.
```
Tremoneck, line 41: Wouldn't 'server_home_page' be better as user home page could suggest that each user could get their own homepage instead of it being for the server instance?

williamkray: Naming feels wrong for sure... Since it's already in the `m.homeserver` object within the client endpoint, the terms server and client both feel redundant... Maybe just `landing_page`?

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,125 @@
...
+{ 
+  "m.homeserver": { 
+    "landing_page": [
+        {
+            "mimetype": "text/plain",
+            "body": "# Visit https://your.domain/user-home.html for the latest server announcements."
+        },
+        {
+            "mimetype": "text/html",
+            "body": "https://your.domain/user-home.html"
```
MichaelKohler, line 72: Kind of newish to the spec, so I'm wondering if it's common to have a URL and load that, or have inline content and render that, all on the same field?

williamkray: dunno, this is my first MSC. format here is following previous suggestion and is based on the extensible event MSC, to allow for plaintext messaging as a fallback render if clients don't support the html format, while continuing to support the idea of simple, in-line content OR separating the content out to another hosted file somewhere (which presumably can be managed separately from however this json is managed). clients would either be presented with the in-line content which they would render, OR a url they follow to fetch the content to render.

this is wide open for improvement.

deepadmax: > Kind of newish to the spec, so I'm wondering if it's common to have a URL and load that, or have inline content and render that, all on the same field?

From what I know, it's not something that's supported or even common. `text/html` means the `body` *is* HTML. Putting the URL in there would simply render text spelling out that URL.

deepadmax: >  clients would either be presented with the in-line content which they would render, OR a url they follow to fetch the content to render.

I really don't think the client should fetch *any* external data. A server owner could, if they'd want a specific website to be reflected, run a service that routinely fetches the website, cleans the HTML, and updates the landing page accordingly.

williamkray: from a technical perspective, seems like maybe having a mimetype of `application/http` would make more sense for this purpose...

i think this is a reasonable use case for fetching external data... this endpoint is controlled by the homeserver administrators and is already used for server delegation. they should be able to point another endpoint for the client to use for other purposes, and the client is responsible for supporting that data endpoint in a sane and safe way. this same concept is currently leveraged for jitsi endpoints, openmap tiling services, etc. which are arguably more critical attack vectors compared to a limited subset of html formatting.

deepadmax: That is a good point. Though it's a bit different with Jitsi, as that's a *widget* and fundamentally operates differently. Whereas this sort of feature *could* be implemented strictly through Matrix. Either way, I do like your idea of an `application/http` mimetype; I think that could be the way to go.

williamkray: that's actually a perfect parallel: this IS effectively a widget, but rather than having the widget tied to a room, it's rendered by the client elsewhere (as determined by the client). instead of room state events defining the widget content (which becomes impossible if the widget is not defined by a room), we are moving that definition to the well-known endpoint. i imagine concepts defined here may eventually be extended to proper widgets at a later point in time. my hope is that this barrier to entry is a little easier for client developers, as currently very few clients actually support widgets at the time of this writing.

deepadmax: That's actually a really good reason to serve the content directly, in a standard format. Not all clients (are going to) want to support widgets.[^1] Keeping it simple and internal will likely lead to more clients implementing the landing page. It would lend itself to a *much lower* barrier to entry. I feel it would be mistaken to require widget support in order to support landing pages.

[^1]: Widgets are written in the web languages (HTML, JavaScript, CSS) and so by definition require the inclusion of a web view in your software. That means one extra dependency. And if it's not utilised anywhere else in the app, it could also be considered bloat.

williamkray: Right but that is solved by supporting the extensible event format proposed above... If remote content isn't supported by the client it can fall back to something else it can render. This is the whole reason for the changes made to align with that msc.

proposals/4079-client-server-landing-page-html.md
```diff
@@ -0,0 +1,125 @@
...
+  landing_page. This is nested within the `m.homeserver` object.
+- The landing_page field can be defined as a list of formats following the [Extensible Event
+MSC](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md),
+  allowing clients which do not render a specific format to fall back to other message types. HTML
+  is supported as either in-line, or by pointing to an html file at an accessible http(s) endpoint
+  reachable by the client.
+- HTML content will be sanitized by the client and restricted to the subset of HTML currently
+  allowed for messages.
+- This field can be queried by clients during the login or initial loading process, and refreshed at
+  least once every 12 hours if the client has been open the entire time. These are minimums, and
```
Johennes, line 50: Just a thought but maybe we could let servers define the refresh period via cache control headers?

proposals/4079-client-server-landing-page-html.md
williamkray: i've updated the verbiage to clarify that minimum support (if a client supports this feature at all, which is itself optional) is plaintext. anything beyond that goes, with html/http examples demonstrated as part of this MSC.

PR #4080, opened 2023-11-15
MSC XXXX: Cryptographic Identities (Client-Owned Identities)

devonh:
[Rendered](https://github.com/devonh/matrix-spec-proposals/blob/cryptoIDs/proposals/XXXX-cryptographic-identities.md)

### DIFF
--- a/proposals/XXXX-cryptographic-identities.md
+++ b/proposals/XXXX-cryptographic-identities.md
@@ -69,13 +69,20 @@ A `txn_id` is added to the request parameters. Clients should generate an ID uni
 access token; it will be used by the server to ensure idempotency of requests.
 
 Request:
-```json
+```
 {
     pdus:  [
-        room_version:  string,
-        via_server:  string,  //  optional
-        pdu: PDU  //  signed  PDU
+        PDUInfo    
     ]
+}
+```
+
+PDUInfo:
+```
+{
+    room_version:  string,
+    via_server:  string,  //  optional
+    pdu: PDU  //  signed  PDU
 }
 ```
 
@@ -96,7 +103,7 @@ needs to be told which pseudoID to correlate to this room for this user.
 The response includes the new fields: `room_version` and `pdus`.
 
 Request:
-```json
+```
 {
     ...,
     sender_id: string
@@ -104,7 +111,7 @@ Request:
 ```
 
 200 OK Response:
-```json
+```
 {
     room_id: string,
     room_version: string,
@@ -140,7 +147,7 @@ the event to the invited user’s homeserver.
 **TODO**: document /make_invite & /send_invite endpoints
 
 200 OK Response:
-```json
+```
 {
     pdu:  PDU
 }
@@ -153,7 +160,7 @@ These are added to help the client when sending the join event to the `/send_pdu
 server chosen by the homeserver to perform the join via.
 
 200 OK Response:
-```json
+```
 {
     room_id: string,
     room_version: string,
@@ -167,7 +174,7 @@ server chosen by the homeserver to perform the join via.
 The leave endpoint is extended to return a `pdu` for the client to sign.
 
 200 OK Response:
-```json
+```
 {
     pdu:  PDU 
 }
@@ -178,7 +185,7 @@ The leave endpoint is extended to return a `pdu` for the client to sign.
 The `/send` & `/state` endpoints are extended to return the `pdu` in the response for the client to sign.
 
 200 OK Response:
-```json
+```
 {
     event_id: string,
     pdu:  PDU
@@ -191,7 +198,7 @@ A `one_time_pseudoids` field is added to the `/keys/upload` endpoint in order to
 the purposes of inviting the user to new rooms.
 
 Request:
-```json
+```
 {
     ...,
     one_time_pseudoids: map[string]OneTimePseudoID
@@ -199,7 +206,7 @@ Request:
 ```
 
 200 OK Response:
-```json
+```
 {
     ...,
     one_time_pseudoid_counts: map[string]int
@@ -207,7 +214,7 @@ Request:
 ```
 
 OneTimePseudoID: 
-```json
+```
 “algorithm:KeyID”: {
     “key”: ”base64_bytes”
 }
@@ -220,7 +227,7 @@ The `/sync` endpoint will need to be extended to report the one-time pseudoID co
 `one_time_pseudoids` for that algorithm.
 
 200 OK Response:
-```json
+```
 {
     ...,
     one_time_pseudoids_count: map[string]int
@@ -234,7 +241,7 @@ all pseudoIDs to regular mxids when sending events to the client. Then the clien
 internally in order to correctly sign future events sent to the room.
 
 200 OK Response (InvitedRoom JSON Object):
-```json
+```
 {
     invite_state:  InviteState,
     one_time_pseudoid: string
--- a/proposals/XXXX-cryptographic-identities.md
+++ b/proposals/XXXX-cryptographic-identities.md
@@ -1,0 +1,409 @@
+# Cryptographic Identities (Client-Owned Identities)
+**THIS MSC IS IN PROGRESS AND WILL CHANGE AS IMPLEMENTATIONS LAND**
+
+Today’s Matrix does not allow users to move their account between homeservers. It would be beneficial to be able 
+to move a user account from one homeserver to another while allowing that user to maintain their existing room 
+memberships, power levels in those rooms, message history, account data and end-to-end encrypted sessions.
+
+With Pseudonymous Identities (pseudoIDs) we have decoupled a user’s mxid from the identity used to track room 
+membership. The new pseudoIDs (also known as user_room_keys/senderids) are both created and managed 
+entirely by the homeserver. PseudoIDs was the first step towards having portable accounts in Matrix.
+
+With Cryptographic Identities we aim to take portable accounts one step further by moving the Pseudonymous 
+Identities off the server and onto the client in order for the client to have full ownership over their identity.
+Clients are then responsible for performing full event signing on a per-event basis. This step brings us closer to
+portable accounts primarily in two ways.
+
+1.  Users now fully own their identity in the Matrix ecosystem and have the control to move their identity.
+2.  It improves the security of the Matrix ecosystem by making it more difficult for a homeserver to act maliciously
+on a client’s behalf. This is particularly important once accounts are portable in order to prevent a homeserver from
+being able to continue operating on a user’s behalf after that user has moved their account off the homeserver.
+
+Cryptographic Identities does not get us all the way to having account portability. Further MSC/s will be required to
+create the appropriate endpoints and other changes in order to successfully port a user’s account from one homeserver
+to another.
+
+## Proposal
+
+PseudoIDs are generated and stored by the client. When joining a room for the first time, a pseudoID should be
+generated for that room. All events are signed by the client using their pseudoID and are no longer signed by the
+user’s homeserver with the exception of the mxid_mapping in the m.room.member event.
+
+### Event Signing
+
+Events are required to be signed by the pseudoID. In order for this to work with client-owned keys, clients need to
+obtain the full version of events before they can be signed. This proposal introduces a few changes to the C-S API
+endpoints used to send events between the client and the server. Any C-S API endpoint which previously was used to
+send events, now returns the fully formed version of those event/s to the client (minus the signatures block). The
+event/s are no longer processed by the server while handling these endpoints. The client then signs the event/s and
+forwards them to the server via a new `/send_pdus` endpoint. When handling events sent to this new endpoint the server
+should process the event/s like normal by adding them to their respective rooms.
+
+A homeserver should avoid processing room events from the client until they have been sent via the `/send_pdus`
+endpoint to ensure the client actually signs the event so it can be successfully sent into the room.
+
+### Endpoint Additions
+
+##### POST /_matrix/client/v1/send_pdus/{txnId}
+
+**Rate-limited**: Yes
+**Requires authentication**: Yes
+
+Fully formed PDUs are sent to this endpoint to be committed to a room DAG. Clients are expected to have signed the
+events sent to this endpoint. Homeservers should reject any event which isn’t properly signed by the client.
+
+Events sent to this endpoint are processed in the order they are received. A homeserver should check the validity of
+each event before sending it to the room. This includes verifying the signature of the event matches the pseudoID
+found in the `sender` field of the event. If the event is for a `remote` invite or join, the relevant `/send_invite`
+or `/send_join` over federation should be performed prior to adding the event to the room.
+
+If any event is invalid all events are rejected by the homeserver. Invalid events include those that are not correctly
+signed, whose event fields are invalid (such as a state event missing a `state_key` field), or the homeserver deems
+the event invalid for some other reason. This approach is taken because this failure mode is most likely due to a
+programming error. Failures of this nature result in a HTTP status 400. A [standard error response](https://spec.matrix.org/v1.8/client-server-api/#standard-error-response)
+will be returned. As well as the normal common error codes, other reasons for rejection include:
+
+-   M_DUPLICATE_ANNOTATION: The request is an attempt to send a [duplicate annotation](https://spec.matrix.org/v1.8/client-server-api/#avoiding-duplicate-annotations).
+
+A `txn_id` is added to the request parameters. Clients should generate an ID unique across requests with the same
+access token; it will be used by the server to ensure idempotency of requests.
+
+Request:
+```json
+{
+    pdus:  [
+        room_version:  string,
+        via_server:  string,  //  optional
+        pdu: PDU  //  signed  PDU
+    ]
+}
+```
+
+### Endpoint Changes
+
+Effected endpoint versions all need to be bumped since the underlying behaviour is changed with this proposal. When
+hitting any of these endpoints the resulting events are no longer immediately added to the room. Instead the client
+is required to send the returned event/s to the `/send_pdus` endpoint after signing them in order for the event/s to
+be added to the room DAG.
+
+##### POST /_matrix/client/v4/createRoom
+
+Room creation adds a new `sender_id` field to the request body. The `sender_id` must be valid [Unpadded Base64](https://spec.matrix.org/v1.8/appendices/#unpadded-base64)
+and 32 bytes in size in order to be a valid ed25519 public key. This field is used for the homeserver to be able to
+fully create all the necessary room creation events on behalf of the client. Since this is a new room the homeserver
+needs to be told which pseudoID to correlate to this room for this user.
+
+The response includes the new fields: `room_version` and `pdus`.
+
+Request:
+```json
+{
+    ...,
+    sender_id: string
+}
+```
+
+200 OK Response:
+```json
+{
+    room_id: string,
+    room_version: string,
+    pdus:  [  PDU ]
+}
+```
+
+##### POST /_matrix/client/v4/rooms/{roomId}/invite
+
+Inviting users to a room has a number of changes in order to make it work. First, since the pseudoID for a given user
+and room needs to be created by the client, we cannot rely on the existing invite sequence which relies on the invited
+user’s homeserver to fully populate the invite event. Instead we need a way for the invited user to be part of the
+loop and provide a pseudoID in order to finalize the event. It would not be acceptable to require the invited client
+to be available at all times in order to respond to an invite request in real time. Matrix does not currently have a
+requirement that client communications be synchronous and this proposal seeks to preserve asynchronous communications
+when participants are unreachable. Instead, this proposal introduces the concept of one-time pseudoIDs.
+
+One-time pseudoIDs are uploaded to the user’s homeserver so that they can be claimed and used whenever that user
+receives a room invite. In order for a user to be available for invite, one-time pseudoIDs should be created and
+uploaded to a user’s current homeserver. This should take the same shape as one-time keys for encryption do today.
+The one-time pseudoIDs should be signed by the device’s ed25519 key to verify they were created by that device.
+
+When a client wants to invite a new user to a room for the first time, they need to query the invited user’s
+homeserver for one of the invited user’s one-time pseudoIDs. They can then use that pseudoID to create an invite
+event for the user.
+
+The invite response includes a new `pdu` field.
+
+Two new endpoints are also added to the S-S API: `/make_invite` & `/send_invite`. These endpoints are required in
+order to split out generating an invite event, and having the inviting client sign that event, from actually sending
+the event to the invited user’s homeserver.
+
+**TODO**: document /make_invite & /send_invite endpoints
+
+200 OK Response:
+```json
+{
+    pdu:  PDU
+}
+```
+
+##### POST /_matrix/client/v4/join/{roomIdOrAlias} && POST /_matrix/client/v4/rooms/{roomId}/join
+
+A number of fields are added to the response of the `/join` endpoints: `room_version`, `via_server`, and `pdu`.
+These are added to help the client when sending the join event to the `/send_pdus` endpoint. The `via_server` is the
+server chosen by the homeserver to perform the join via.
+
+200 OK Response:
+```json
+{
+    room_id: string,
+    room_version: string,
+    via_server: string,
+    pdu:  PDU
+}
+```
+
+##### POST /_matrix/client/v4/rooms/{roomId}/leave
+
+The leave endpoint is extended to return a `pdu` for the client to sign.
+
+200 OK Response:
+```json
+{
+    pdu:  PDU 
+}
+```
+
+##### PUT /_matrix/client/v4/rooms/{roomId}/send/{eventType}/{txnId} && PUT /_matrix/client/v4/rooms/{roomId}/state/{eventType}/{stateKey}
+
+The `/send` & `/state` endpoints are extended to return the `pdu` in the response for the client to sign.
+
+200 OK Response:
+```json
+{
+    event_id: string,
+    pdu:  PDU
+}
+```
+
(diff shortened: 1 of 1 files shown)


proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,411 @@
...
+
+Requiring the `master_signing_key` to sign a join event in order to attest a user is who they claim to be would
+typically require the user to enter their recovery passphrase every time they join a room. This is because clients
+do not usually store this key. This would lead to a large burden on users and would be best to avoid if at all 
+possible.
+
+### Additional Attack Vectors
+
+Clients can modify events prior to signing them and sending them to the server for processing. This can lead to
+issues if the client were to change something such as the `prev_events` which could lead to further problems.
```
neilalexander, line 339: This is more than just an attack vector — it's a fundamental part of how events are created and it cannot be an afterthought.

As it stands today, clients generally know nothing of `prev_events` or forward extremities. When the client wants to send an event into a room, it sends a "client event" containing only client-controlled data to the server, which the server then pads out and populates with additional fields and signatures. The `prev_events` form the DAG relationships between events and populating the `prev_events` with as many real forward extremities as possible is what keeps the graph moving in a forward direction and prevents long-lived forks in the room graph (ideally).

It's also worth noting that only servers are tracking forward extremities at this stage as clients generally can't be expected to — clients often only know a subset of what the server knows about the room history and `/sync` doesn't make any guarantees that the responses won't contain gaps. Different servers in the room might have different ideas about what the forward extremities are depending on history, but eventually if all servers are participating actively in the room by sending events and populating `"prev_events"` correctly, this should reconcile eventually.

As I see it there are only really two workable options:

1. The client signature should only cover certain fields of the event, such as the `"content"` and `"type"` etc, and a server-added signature would need to cover additional things such as `"prev_events"` fields — anyone receiving the event then needs to verify both the client and the server signatures
2. The client needs to be informed of the forward extremities before every single event is created, at the risk that multiple clients can and will race in creating events — this would increase the room graph complexity which in turn increases the processing/state resolution costs on the servers (whereas right now this race can only happen between *servers* sending at the same time)

This is largely why the room join dance over federation has the additional `/make_join` step, as a server already participating in the room is required to populate things like `"prev_events"` into the template event before the joining server signs it and sends it to `/send_join`.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,411 @@
...
+	-   Typing
+	-   Locations? EDU/State?
+    
+
+### Auth Rules
+
+A new room version will be required to account for the modifications to the auth rules.
+
+Invite events no longer require a signature from the invited user’s homeserver. This signature requirement does not
+appear to have an obvious benefit and would make invite events overly onerous with the new room invite process.
```
neilalexander, line 291: It's not entirely without benefit. The invited server signature is what allows participating servers to prove that the invited server was online at the time of the invite and that the invited server didn't have good reason to reject the invite, i.e. because the user didn't exist or similar. Otherwise it becomes considerably easier to just fill the room state with masses of obviously fake invite events for obviously fake users which servers then need to resolve forever more without ever knowing if they have a genuine use or not.

devonh: Ah, thanks for enlightening me!

If it is a matter of keeping a signature specifically from the invited user's homeserver, then that should be easy enough to keep in place. It only becomes a problem if the signature needed to come from the client. (And since myself and others weren't aware of this benefit, we figured we could just remove the "unnecessary" homeserver signature)

We are trying to avoid homeserver signatures as much as possible on events to remove the burden of fetching server keys. The only other server signature location with pseudoIDs/cryptoIDs is in the `mxid_mapping` field of the `m.room.member` event to verify the mxid for the user belonging to that homeserver.

devonh: I have added a diagram of the proposed federated invite flow to help clarify things.

The proposed addition of one-time cryptoIDs involved when inviting users to a room provides an opportunity for the invited user's homeserver to ensure that user is resident on that homeserver at the time. Otherwise the homeserver should fail the invite during the call to `/make_invite`.

Invites still use the mxid in the proposal so you don't (and can't) know a user's cryptoID until you have sent the invite and the user's homeserver has allocated one of the one-time cryptoIDs as the user's cryptoID for that room. Until that time the only thing that can be known is the user's mxid.

You could run into the case where the user moves to a new homeserver in the middle of the invite flow. This only becomes an issue if the user never receives the invite event from any homeserver. As long as they receive the event at some point, they should also have the matching private key necessary in order to accept the invite, regardless of which homeserver they are currently resident on. There might need to be something in place when designing account portability to help minimize how often this happens. (such as a grace period to forward such incoming events from the old homeserver to the new). 

I may not be understanding your fourth point fully. If someone lies about where an invited user lives, this is no different than trying to invite a user with the wrong mxid in today's matrix.

proposals/4080-cryptographic-identities.md
devonh: Everything you said is correct. And we came to the same conclusion of those being the 2 options.
We chose to go with option 2 because option 1 leads to issues where you need to delegate which server currently has signing authority for a user (ie. if a user moves to a new homeserver, then the old homeserver should no longer be able to sign and create events on behalf of the user. This becomes an issue since the old homeserver can replay old events from the client with different `auth_events` & `prev_events`, making it possible for a malicious homeserver to be able to do bad things)

The bulk of this proposal outlines the endpoint changes necessary to make option 2 work such that the client is able to sign full versions of every PDU.

I left out some words around how to prevent a client from changing event fields (such as `prev_events`) since I didn't think it was relevant to the spec and should only be an implementation detail. But the gist of it is that a server should store the hash (eventID) sent to the client and only accept signed events from a client that have a matching hash. This doesn't prevent a malicious homeserver from colluding with a maclicious client. But that's also true of today's matrix network.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+sending flow fast since non-state events wouldn’t have the additional client-server round trip.
+
+### Clients sign full events via room extremity tracking
+
+In this model the client would be responsible for creating a full event (including `prev_events` and `auth_events`) 
+by tracking and resolving the room’s state.
+
+**Advantages**: This has the advantage of events being fully signed by the cryptoID and avoiding a second round trip.
+
+**Disadvantages**: This has the disadvantage of requiring clients to do state resolution.
```
neilalexander, line 512: Clients cannot do state state resolution — they don't have the necessary information about the state before/after each event in order to do so.

In any case, the problem here is not state resolution, but that it requires the clients to track forward extremities, which they also cannot really do given the fact that `/sync` and friends do not return breadth-first scans of the graph but rather a linearisation of it. It's also likely that new forward extremities can build up further back in history from newly arriving events over federation and clients may not have any reason to find out about these until a new event is sent that refers to one of them.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+
+Clients are now responsible for signing the `content` field of events but they don’t sign the full event. This means 
+that a malicious homeserver could take the contents of an existing room event and replace everything else in the 
+event without anyone knowing. This could include fields such as `type`, `state_key`, `prev_events`, and `room_id`. 
+In order to minimize the effects of a replay attack, the client should sign the combination of `type`, `state_key`, 
+and `content`. Signing `type` prevents reusing the contents in an event of another event `type`. Signing `state_key` 
+prevents attacks such as changing the `membership` state of another user. Signing `content` prevents a malicious 
+homeserver from generating arbitrary `content` on behalf of a client. 
+
+Even with the above mitigation, a malicious homeserver could still replay an event in the same room, with the same
```
neilalexander, line 494: The natural solution here is to include the `"prev_events"` in the signature, which anchors that event to a specific position in the graph and effectively prevents replays elsewhere, but this would require rethinking the signature scheme. This still also assumes that the client knows what the `"prev_events"` even are.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+- `signature`
+
+##### Replay Attacks
+
+Clients are now responsible for signing the `content` field of events but they don’t sign the full event. This means 
+that a malicious homeserver could take the contents of an existing room event and replace everything else in the 
+event without anyone knowing. This could include fields such as `type`, `state_key`, `prev_events`, and `room_id`. 
+In order to minimize the effects of a replay attack, the client should sign the combination of `type`, `state_key`, 
+and `content`. Signing `type` prevents reusing the contents in an event of another event `type`. Signing `state_key` 
+prevents attacks such as changing the `membership` state of another user. Signing `content` prevents a malicious
```
neilalexander, line 491: Some consideration needs to be made for power events, i.e. kicks and bans, where another user or the server itself needs to effectively take action on behalf of another user.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+values have been modified. If a homeserver receives an event with a `nonce` that is identical to another event with 
+the same `type` and `state_key`, that event should be rejected. Clients have the option to also perform `nonce` 
+validation in this way if the possibility of colluding homeservers is suspected. When sending an event via the C-S 
+API, a homeserver should verify that the `nonce` of the new event is unique or reject the event from the client. 
+If the event is rejected in this way, the homeserver should return a response status of 400 with an errcode of 
+`M_DUPLICATE_NONCE`.
+
+**Problem**: How can a client generate a usable nonce? 
+
+**Problem**: How could a homeserver validate a nonce as being unique without requiring them to know the entire room DAG?
```
neilalexander, line 474: In short, this isn't feasible, and it may well open up a possible attack vector whereby multiple events for the same `"type"` and `"state_key"` can be sent with the same `"nonce"` and then it creates a race where different servers might accept one or the other depending on which one they learn about first. If this happened with power events then that would have knock-on effects with event auth and you'd fracture the room.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+a `hash` of the full contents that can be used to verify the full event. Signing the redacted `content` instead 
+of just the `hash` allows for further field validation (ie. fields from `m.room.member`, `m.room.join_rules`, and 
+`m.room.power_levels` to name a few). If a server detects the `signature` is wrong it should reject the event. If a 
+client detects the `signature` is wrong it should alert the user who can then decide what further action to take, if 
+any. A client detecting an invalid `signature` means their homeserver either didn’t check the `signature` or did 
+check the `signature` and didn’t reject the event.
+
+Servers should also check that the full event was signed by one of the keys present in the `allowed_signing_keys` 
+field, or by the cryptoID itself. If the event was not signed by one of these keys, the server should reject the 
+event. Allowing events to be signed by the cryptoID keeps the possibility of clients to perform state resolution
```
neilalexander, line 458: I already made mention of this in another comment but clients shouldn't and can't be expected to do state resolution. Not least because in order for them to do so, they need to track the state before/after each event and to track forward extremities, and they need the ability to contact other servers over federation to fill in those knowledge gaps, by which point you've effectively built a homeserver. That is why P2P was predicated on portable homeservers. There's no way forward where clients running state res is the right answer.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
...
+fields are used to prevent malicious homeservers from spoofing events on a client’s behalf. The signature is created 
+by signing the combination of `type`, `state_key` and redacted `content`. Signing the `room_id` is not required since 
+cryptoIDs are already unique per room but could be added if necessary. The `nonce` value should be generated as a 
+random 128-bit integer (or UUIDv4?), encoded as unpadded base64 (or hex for UUID?). It is especially important for 
+the `nonce` to be unique for each `type` and `state_key` pairing in order to ensure events cannot be replayed by a 
+malicious homeserver. The hash is the sha-256 hash, encoded as unpadded base64, of the `content` fields before 
+`signatures` have been added (including the new `nonce`, and `allowed_signing_keys` fields). The `hash` is required in 
+order to be able to verify the event `signature` if the event `content` is ever redacted.
+
+All events are now required to possess the above mentioned fields inside of `content`.
```
neilalexander, line 424: Duplicating many of the top-level event keys and putting them under client control inside `"content"` instead of properly rethinking the top-level event format feels like it will create technical debt that will never be repaid.

proposals/4080-cryptographic-identities.md
neilalexander: I'm not really sure I'm following — in the "one-time" crypto ID case, who owns the private key for that identity? If the client isn't the specific owner of that key and isn't online at the time to sign the proposal, then how can the client trust an invite that it learns about later, i.e. after a migration?

proposals/4080-cryptographic-identities.md
MTRNord: Also some clients fundamentally cant talk to other HS over fed since they have no way to properly do DNS which is a fundamental requirement we have for S-S Api endpoints.

proposals/4080-cryptographic-identities.md
neilalexander: Preventing a client from changing the `"prev_events"` is a pretty important detail and needs some explanation in the MSC in that case, as that is a critical security component.

I also think that serious consideration needs to be made for the fact that multiple clients could end up requesting proto-events with the same `"prev_events"` while other clients are signing/uploading their signed events, which increases the number of forward extremities and increases the state resolution frequency & cost on the server. What will the server do to protect itself from lots of users in the same room creating a mass amount of forward extremities at the same time, either by dumb chance or by collusion?

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,553 @@
+# MSC4080: Cryptographic Identities (Client-Owned Identities)
```
MTRNord, line 1: While @neilalexander already mentioned lots of technical flaws I think in the current for even from a purely logical standpoint its illogical.

A) It feels like you want a HS in the client. I strongly doubt a Client dev would implement this even if it became merged as is
B) It feels like this would be what you expect from P2P clients but at that point since you anyway write a HS for that the whole MSC would be obsolete at its core since the HS is already the client.

Also on phones something like this feels like a burden if not even impossible due to the common battery optimizations by the OS.

MTRNord (changes requested):
I strongly believe in its current form its not a good direction this MSC proposes

proposals/4080-cryptographic-identities.md
devonh: Thank you for further detailing why this alternative is not viable.
I'll update this section to better explain why we don't recommend it.

proposals/4080-cryptographic-identities.md
devonh: I agree. That would be the natural solution if a client could somehow know what the `prev_events` are. 
This is the main reason we stopped investigating the client-delegated signing alternative and have instead proposed the changes outlined in this MSC.

proposals/4080-cryptographic-identities.md
devonh: I'm glad you agree :smile: 
If someone can come up with a feasible approach to protecting against such attacks then this alternative may be a better approach to what is being proposed in this MSC.

proposals/4080-cryptographic-identities.md
devonh: This alternative proposal of client-delegated event signing isn't predicated on being able to run state resolution. The comment was added only as an aside to outline that it wasn't being completely designed out of the picture if this solution were to be accepted.

proposals/4080-cryptographic-identities.md
devonh: If the replay attack issue present in the client-delegated event signing alternative proposal can be solved then this could be looked at further. This approach of modifying `content` was documented to make it clear where the problem arises and why this alternative isn't being proposed in this MSC.

proposals/4080-cryptographic-identities.md
devonh: The client owns the private keys for all their cryptoIDs (one-time or otherwise).

proposals/4080-cryptographic-identities.md
devonh: Thanks for the comment :smile: 

It seems as though these opinions are based around the alternative proposals of either having clients delegate event signing to a homeserver or generating `prev_events` themselves somehow.

I agree with you that we should avoid spec changes that require clients to behave like homeservers. The alternative proposals have been listed to provide context as to how we arrived at the actual proposal.

The proposal presented in this MSC is only asking that clients now sign events. No other homeserver behaviour is expected of clients.

proposals/4080-cryptographic-identities.md
devonh: I can add words describing how a homeserver should protect against modified events.

RE: increased forward extremities - this issue is already present today with federated homeservers.

devonh:
> Hi. I wrote a small article about this: https://benstokman.me/blog/my-thoughts-on-msc-4080/

Thank you for taking the time to review the MSC.

I agree that things would be simpler if we could just use a user's Master Signing Key.

The problem with this is solely around deniability. Currently, Matrix's encryption protocol provides users with plausable deniability. If all events are now signed client-side with a user's Master Signing Key (or PGP key), this deniability aspect of encryption will be greatly diminished. The system proposed in this MSC was designed to try and preserve the existing features of the protocol.
CryptoIDs are not cryptographically connected to a user's Master Signing Key or device keys. A cryptoID could be forged on a user's behalf, and in combination with a colluding or compromised homeserver, could generate all of the same events.

If there is some mechanism to cut down on the number of keys while maintaining plausable deniability, then that would be the preferred approach. But this MSC is exploring what it would look like without that.

Also, a system that allows changing signing keys may be possible, but it is also full of potential issues due to federation and state resolution. The old keys could still be used to send events with prev_events that refer to a point in the room DAG before the signing keys were changed. In the current matrix, these events would be valid because they could have just been received late.
This gets especially dangerous when considering state events.

benjistokman:
I found a middle ground:

* Devon's design *is* better. Any kind of self-signed events introduces the risk of communication failure after key loss. However, a way to back up/restore key data would have to be made anyway to support portable accounts. Periodic reminders to to a user to back up their keys would be useful.
  * Perhaps a way to procedurally generate the cryptoIDs could be implemented so an older backup can restore newer keys. [Single use] encryption keys aren't as important. I am not sure if this is feasible or possible within the limitation of plausible denyability.
* Only events that deal with invites, power level changes, etc. should ***require*** a user-signed event. If a user looses their key setup but still has their account password, having the ability to send out a message saying something like, "Hey I lost my keys we'll need to re-verify when we meet next" would be *very* useful—even if unencrypted. The other users' clients should still show a warning the same way as any unsigned and/or unencrypted event.
  * Valid server-signed events that are properly encrypted/signed should be treated the same way as a user-signed message of the same event type.
  * There should be a client-side option to "hide server-signed events" should a homeserver start sending spam.
* There still needs to be a way to change singing keys. Being entirely unable to expire/replace cryptographic keys is plain and simple bad security practice.
  * In order to get around the race condition, there should be seprate events to add keys and another to remove keys. To do a key migration, a client first adds the new key, then prompts the user to open all their devices/applications. It then removes the old key some reasonable time in the future, say, 4 hours later.
* There should be a setting per room that requires users to acknowledge account migrations before they are shown new messages. Many users *will* be confused by account migrations. Imagine if you're in an email chain with someone and you're suddenly composing a reply to an unrecognized address.
  * Putting a "this user has migrated their account" in the chat feed is not sufficient. There's no guarantee whatsoever that *every* user would *always* notice this. Requiring an action is better.
  * This would be a per-room setting similar to the "never send encrypted messages to unverified sessions" setting.
  * This should be the default in encrypted rooms only. Public+unencrypted rooms don't really need it.
  * Keys should not be sent to sessions before the user acknowledges the migration.
  * This may also help limit consequences from key leaks.

erlend-sh:
https://matrix.org/blog/2024/03/why-matrix-org/

> There are alternatives to a prominent onboarding homeserver like the Matrix.org homeserver. We don't find any of them satisfactory:
> 
> * Picking a server randomly when a user wants to sign up in the app is confusing and has a lot of security implications. This is not acceptable.
> * Forcing the user to choose a server when they sign up is a highly cumbersome process that does not respect the user's time and resources.
> * **Portable identities are very desirable but will take time to materialise.**
> * **P2P Matrix depends on a portable identity mechanism and is further away in the future.**

It’s painfully obvious that portable identities are the best solution to a whole range of problems in the federated paradigm.

Since lack of funding is the main issue, I’d like to point @devonh & co. to https://summerofprotocols.com/research/sop2024

Specifically the [Protocol Improvement Grant](https://summerofprotocols.com/research/sop2024#pigs):

> Protocol Improvement Grants: This track is aimed at discovering the general principles of improving protocols. The track will award grants of $90,000 each to 5 two-person teams to prototype and field-test improvements to a real-world protocol.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,584 @@
...
+
+The `/send` & `/state` endpoints are extended to return the `pdu` in the response for the client to sign.
+
+200 OK Response:
+```
+{
+    event_id: string,
+    pdu:  PDU
+}
+```
```
AndrewFerr, line 235: To be clear, is the `"event_id"` field included in the PDU that needs signing?  I assume it is, given that events currently have `"event_id"` outside the `"unsigned"` field, making it subject to being included in event signatures.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,584 @@
...
+## Proposal
+
+CryptoIDs are generated and stored by the client. When joining a room for the first time, a cryptoID should be
+generated for that room. All events are signed by the client using their cryptoID and are no longer signed by the
+user’s homeserver with the exception of the mxid_mapping in the m.room.member event.
+
+### Event Signing
+
+Events are required to be signed by the cryptoID. In order for this to work with client-owned keys, clients need to
+obtain the full version of events before they can be signed. This proposal introduces a few changes to the C-S API
```
andrewzhurov, line 35: Could client craft the full version of an event on its own?
afaik, the only field that server fills up is `prev_events`.
Client does have Room events locally (they may be stale, but I'd argue we want to bake that "stale" causality - that is indeed what client sees, baking server's causality is wrong, as it's not the one observed when creating an event. Prob intsead of calling it "stale" a better word is "true":).

Further, would allow clients to issue events while being offline.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,584 @@
...
+the attestation of the cryptoID. A client could choose whether to use different room signing keys per room (the
+benefit of doing this would be to ensure that knowing a user’s identity in one room did not lead to knowing that
+same user’s identity in another room), or use the same room signing key for all rooms. Then at a later time clients
+could use some out of band attestation mechanism to “cross-sign” in order to verify the user/s are who they say they
+are. This has the additional benefit of not needing to enter the user’s recovery passphrase to provide the attestation
+as clients could store these room signing keys.
+
+### Identity Sharing Between Devices
+
+The cryptoIDs of a user are shared between devices using secret storage similar to the way encryption keys are shared.
```
andrewzhurov, line 386: Seems we could make use of [Profile Room](https://github.com/matrix-org/matrix-spec-proposals/pull/1769/files) to manage authorized devices & their currently used keys, as they are rotated over time.
This way keys are non-extractable, giving increased security.
And devices help each other to maintain user's identity. One device gets lost - unlink it.
Also, Profile Room can be used to anchor "interaction" (e.g., message) events, proving it's authenticity even across key rotations.

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,584 @@
...
+
+In this model the client would add a `allowed_signing_keys` field to their `m.room.member event` in order to delegate 
+event signing to another party. Homeservers still have full authority over a client’s events in this scenario since 
+the client doesn’t sign any part of each event to verify they are the sender.
+
+**Advantages**: This has the advantage of not adding additional size to each event.
+
+**Disadvantages**: This has the disadvantage of giving over full event control to the delegated homeserver. It also has 
+the disadvantage of trying to resolve `allowed_signing_keys` if a client wants to remove authority from a homeserver 
+or there are conflicts in the room DAG. Revocation of a delegated key is known to be extremely problematic.
```
andrewzhurov, line 555: Good remark. Bummer we don't have `prev_events` with sync. (Synapse does)
Serving clients full events would make it possible. This approach [is mentioned in MSC3871](https://github.com/matrix-org/matrix-spec-proposals/pull/3871/files#diff-569fcc6732d621597bc1371e7cb299effb13888d9f2a3ef23760a00de401f723R187).

proposals/4080-cryptographic-identities.md
```diff
@@ -0,0 +1,584 @@
...
+
+### Endpoint Additions
+
+##### POST /_matrix/client/v1/send_pdus/{txnId}
+
+**Rate-limited**: Yes
+**Requires authentication**: Yes
+
+Fully formed PDUs are sent to this endpoint to be committed to a room DAG. Clients are expected to have signed the
+events sent to this endpoint. Homeservers should reject any event which isn’t properly signed by the client.
```
AndrewFerr, line 53: Since among a [PDU's required fields](https://spec.matrix.org/v1.10/rooms/v11/#event-format-1) is the timestamp of when the event was created (`origin_server_ts`), would it be an issue if there were a large delay between the time an event is created & when it's actually inserted into the DAG (i.e. if a client waits a long time between calling `/send` and `/send_pdus`)?

AIUI that would at least affect time-based syncing, such that events would be synced in timestamp order rather than DAG order (which would be different only if other events got inserted between `/send` and `/send_pdus`).

But given that timestamp/DAG discrepancy is allowed for [timestamp massaging by appservices](https://spec.matrix.org/latest/application-service-api/#timestamp-massaging), perhaps this is nothing out of the ordinary.

PR #4081, opened 2023-11-21
MSC4081: Claim fallback key on network failure

kegsay:


proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,75 @@
+# MSC4081: Claim fallback key on network failures
+
+In order for clients to establish secure communication channels between devices, they need to "claim" one-time keys (OTKs) that were previously uploaded by the device they wish to talk to. One-time keys, as the name suggests, must only be used once. However, this presents several problems:
```
uhoreg, line 3: Please wrap your lines at around 80 to 120 characters.

(Go home, spell check.  You're drunk.  Well, you did catch one typo...)

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Proposal
+
+Currently, fallback keys are _only_ claimed on key exhaustion, not due to transient network failures. This MSC proposes to change the semantics to allow fallback keys to be returned by the `/keys/claim` endpoint if the server the target device is on is unreachable. In order for servers to return fallback keys during the network failure, the fallback keys must be cached _in advance_ on the claiming user's homeserver. This MSC proposes adding a new key `fallback_keys` to the `m.device_list_update` EDU. This MSC proposes changing the spec wording (bold is new):
```
uhoreg, line 11: The problem with pre-sending the fallback key is that the server keeps track of whether anyone has seen the fallback key, and once it has been sent to anyone, the server will note in the sync that it has been used (by removing the algorithm from `device_unused_fallback_keys`), at which point, the client will generate a new fallback key (modulo rate-limiting so that it doesn't evict private keys too quickly).  The client can't wait until it receives a message using that key.  It has to create a new key once it gets sent to anyone because otherwise someone could fetch the key and start trying to crack it, and they would have extra time to do so.

kegsay: I'm not sure this is necessarily something we wish to optimise for? It's a _public key_. If it cannot be public "for too long" then it kind of defeats the purpose of having it be public at all, and it makes me wonder how systems entirely dependent on unchanging public keys (bitcoin, ethereum, libp2p) manage the risk.

kegsay: I've added words in defence of this approach in the security considerations section. Please take a look, and if I'm missing something please yell.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,129 @@
...
+| keys             | {string: string}        | Required:  Public identity keys. The names of the properties should be in the format  <algorithm>:<device_id>. The keys themselves should be encoded as specified by the key algorithm.                   |
+| signatures       | Signatures              | Required:  Signatures for the device key object. A map from user ID, to a map from  <algorithm>:<device_id> to the signature.   The signature is calculated using the process described at  Signing JSON. |
+| user_id          | string                  | Required: The ID of the user the device belongs to. Must match the user ID used when logging in.                                                                                                          |
+| **fallback_key** | **{string: KeyObject}** | **The fallback key for this device, if set. The format of this object is identical to the /keys/claim response for a single device. This replaces any previously sent fallback key.**                                                                         |
```
richvdh, line 42: Currently, clients are told that their fallback key has been "used" via the `/sync` API (specifically, via its absence in the `unused_fallback_keys` property, IIRC).

Does issuing the fallback key in an `m.device_list_update` count as "used" for this purpose? It seems like you're assuming not, but please make that explicit.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,129 @@
...
+                "@alice:example.com": {
+                    "ed25519:JLAFKJWSCS": "FLWxXqGbwrb8SM3Y795eB6OA8bwBcoMZFXBqnTn58AYWZSqiD45tlBVcDa2L7RwdKXebW/VzDlnfVJ+9jok1Bw"
+                }
+            }
+        }
+    }
+}
+```
+
+As a reminder, clients SHOULD rotate their fallback key when they realise it has been used, with some lag time
```
richvdh, line 61: Again, I think you need to clarify what "used" here means, and that it is *different* to what clients do today.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,129 @@
...
+to account for federation. As per MSC2732, 1 hour is recommended. When clients change their fallback key, a new
+`m.device_list_update` EDU MUST be sent.
+
+This proposal has no client-side changes.
```
richvdh, line 65: Yes it does.

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: I'm not sure if @uhoreg agrees, but I *think* the reason the key needs to rotate is for forward secrecy (ie, to guard against *future compromise* of the fallback key), rather than cracking. Once you throw away the private part of the FK, any sessions set up using that FK can no longer be compromised via compromise of the FK.  (This is also the reason we prefer a OTK where available: there is only one conversation that can be compromised via a given OTK.)

 It's also worth noting that the Signal [spec](https://signal.org/docs/specifications/x3dh/#publishing-keys) says:

> Bob will also upload a new signed prekey and prekey signature at some interval (e.g. once a week, or once a month). 

Note the reference to rotating it periodically, rather than "as soon as it is used".

A final point: in E2EE, there is no real reason to trust my own homeserver any more than anyone else's. If it's ok for my public FB key to sit for a long time on my own HS, it should also be ok for it to go out to other HSes.

---

So I think this aspect is *probably* fine? Though we do need to make sure that existing clients know they need to rotate the key.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+As a reminder, clients SHOULD rotate their fallback key when they realise it has been used, with some lag time
+to account for federation. As per MSC2732, 1 hour is recommended. When clients change their fallback key, a new
+`m.device_list_update` EDU MUST be sent.
```
richvdh, line 63: During rotation, clients will need to hang onto the private half of their old key for a while, to make sure that everyone else has got the memo (possibly for days, to account for federation delays). Given that, it seems like clients should create a new key immediately but keep the old one working for a while.

Also, per discussion elsewhere: the signal protocol recommends a periodic rotation of the prekey, whether or not it has been used. It seems like that might be a good defence in depth here?

proposals/4081-claim-fallback-keys-on-network-failure.md
kegsay: Added:
```diff
-This proposal has no client-side changes.
+The definition of when a fallback key is "used" also needs to change. Previously, a key is "used"
+_if it is claimed by another device_. When this happens, the client is told this via `/sync`, either by reducing
+the one-time key count by 1, or by removing the algorithm from the `device_unused_fallback_key_types`. This proposal
+makes it impossible to know if the fallback key has been claimed by another device, as it is sent eagerly over
+federation. Therefore, this changes the fallback key semantics to be "when the device receives and successfully
+decrypts an initial pre-key to-device event which uses that key". As per the specification, this is identified as
+`type: 0` messages. This will require client-side changes to change when new fallback keys get uploaded.
+
+Due to this change, it is recommended that the fallback key is _also cycled periodically_, e.g once per week.
+This ensures that any sessions the client is unaware of will eventually expire.
```
which I think addresses all your comments.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+DH operation is performed.
+
+This optionality makes the protocol robust to OTK exhaustion and transient network failures (e.g to a database to
+claim OTKs as Signal is not federated).
+
+## Security Considerations
+
+Ultra secure clients may be unhappy that fallback keys are being returned and not one-time keys, because they
+dislike the slightly weaker security properties fallback keys provide. This could be resolved by adding a flag to
+the `/keys/claim` endpoint to state whether returning a fallback key is acceptable to the client or not. If this
```
uhoreg, line 103: I don't think this is necessary.  Fallback keys are already supposed to have a `fallback: true` property (though that got missed in the initial spec PR), so a client can do a normal `/keys/claim` and if it sees the `fallback: true`, then it can discard the key.  Discarding a fallback key is the same as not receiving a key because you say you don't want a fallback key.

Since the `fallback` property is covered by the signature on the key, a malicious server can't remove it.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+to account for federation. As per MSC2732, 1 hour is recommended. When clients change their fallback key, a new
+`m.device_list_update` EDU MUST be sent.
+
+The definition of when a fallback key is "used" also needs to change. Previously, a key is "used"
+_if it is claimed by another device_. When this happens, the client is told this via `/sync`, either by reducing
+the one-time key count by 1, or by removing the algorithm from the `device_unused_fallback_key_types` array. This proposal
+makes it impossible to know if the fallback key has been claimed by another device, as it is sent eagerly over
+federation. Therefore, this changes the definition of "used" to be "when the device receives and successfully
+decrypts an initial pre-key to-device event which uses that key". As per the specification, this is identified as
+`type: 0` messages. This will require client-side changes to change when new fallback keys get uploaded.
```
uhoreg, line 71: I think we need some sort of transition plan for clients to be able to use the new behaviour.  Otherwise, strange things may happen, depending on how the `device_unused_fallback_key_types` property gets redefined.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+Currently, fallback keys are _only_ claimed on key exhaustion, not due to transient network failures. This MSC
+proposes to change the semantics to allow fallback keys to be returned by the `/keys/claim` endpoint if the server
+the target device is on is unreachable. In order for servers to return fallback keys during the network failure,
+the fallback keys must be cached _in advance_ on the claiming user's homeserver. This MSC proposes adding a new
+key `fallback_keys` to the `m.device_list_update` EDU. This MSC proposes changing the spec wording (bold is new):
+
+> Servers must send `m.device_list_update` EDUs to all the servers who share a room with a given local user, and
+> must be sent whenever that user’s device list changes (i.e. for new or deleted devices, when that user joins a
+> room which contains servers which are not already receiving updates for that user’s device list, or changes in
+> device information such as the device’s human-readable name **or fallback key**).
```
uhoreg, line 31: One potential issue that I can see with pre-sending fallback keys is something like this scenario:

- Alice uploads fallback key A, which gets sent to Bob's server
- Bob's server goes down for a while
- Alice receives some olm messages that use the fallback keys, so rotates her keys, first to fallback B and then to fallback C.  At this point, she has evicted private key A.  Since Bob's server is down, it doesn't receive the new fallback keys.
- Alice's server goes down for a while, and then Bob's server comes back up
- Bob tries to establish an Olm session with Alice, receives fallback A, and sends an encrypted message to Alice
- Alice's server comes back up, and Alice receives the message from Bob, but can't decrypt since she doesn't have private key A any more

This is worse than the situation where Bob doesn't get any OTK for Alice, since if he doesn't receive any key, he knows about the failure and can retry later.  On the other hand, this is likely a very rare scenario, so may not be worth worrying about.

proposals/4081-claim-fallback-keys-on-network-failure.md
uhoreg: Yeah, I think that if we rotate periodically, it should be OK.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
+# MSC4081: Claim fallback key on network failures
```
uhoreg, line 1: @ara4n's [comment](https://github.com/matrix-org/matrix-spec-proposals/commit/ef28b251981f6185c73f956302771a9114749dd0#r133195177) somehow ended up on the commit rather than on the PR, so copying here:

> I'm a bit worried about this: it's (nominally) weakening security in order to work around network reliability issues. it reminds me of our misadventures in key gossiping, where we similarly weaken security to mainly work around bad retry mechanisms and network unreliability.

> If our server can't talk to the other server, i wonder if we should warn the sender (e.g. a "can't contact bob.com!" warnings on the message) and then retry? the sender will know to keep the app open while it tries to retry (just as they would if they were stuck sending the message too)? This feels better than to give up and send the message with (nominally) lower security, and could also make the app feel more responsive with appropriate UX (i.e. rather than being stuck in 'sending' state for ages while a /key/claims times out, it could declare itself sent to 10 out of 11 servers, or similar).

proposals/4081-claim-fallback-keys-on-network-failure.md
kegsay: > Discarding a fallback key is the same as not receiving a key because you say you don't want a fallback key.

Not sure I agree with this statement. Over federation, I don't think there is a backchannel to say "I claimed this key but actually I'm not using it", so the act of calling [this endpoint over federation](https://spec.matrix.org/unstable/server-server-api/#post_matrixfederationv1userkeysclaim) would cause the claimed server to tell the client to rotate their fallback key. In other words, we can't distinguish the difference between receiving and discarding.

But either way, it sounds like my concern around knowing if a key is fallback or not and choosing to use it or not has already been thought of, which is great. I shall update the MSC.

proposals/4081-claim-fallback-keys-on-network-failure.md
kegsay: We can add in a query parameter to the `/keys/claim` endpoint if we need to. This would:
 - indicate that the client is happy to use a fallback key in the case of interrupted connectivity.
 - tell clients when to use the new "used" behaviour.

proposals/4081-claim-fallback-keys-on-network-failure.md
kegsay: This could be mitigated by either:
 - signalling to the sender that the session was undecryptable. Whilst this can be prompt, I would be worried about DoS and oracle-like attacks though. DoS because now attackers have a way to cause clients to send traffic automatically by sending keys the client doesn't have. Oracle because an attacker can send various keys and know if the client has the private key on-disk still, i.e it exposes whether the key has been evicted or not which is a critical part of forward secrecy.
 - treating fallback-initiated sessions as unreliable, and hence if you do not get an established session after time N, try to claim another OTK and try again? Unsure of the security implications here, as an attacker may be able to make use of the fact that there may be >1 established Olm session?

proposals/4081-claim-fallback-keys-on-network-failure.md
BillCarsonFr: Can we clarify to what extent it weakens security? The use of fallback keys is I think mitigated by the fact that the double ratchet will do a DH step on the next message and that will restore security.
Fallback key exists so that communication do not break when all OTKs are exausted (as of convenience and it's mitigated), why can't also they be used for transiant federation connectivity problems?
Maybe the could be modified to have a ttl? 
And in case of replay attack (same prekey message sent other and other), couldn't the client add some additional mitigiations (as it could already BTW)? Like detecting abusive use of fallback?

kegsay: I'd like to re-emphasise the *nominal* reduction in security: in reality there is negliable impact, further reinforced by other secure protocols (Signal in this case) allowing OTKs to be optional in the setup phase. I think this MSC is overall net positive, as it makes the protocol more robust, and fixes concrete bugs we've seen in the wild.

proposals/4081-claim-fallback-keys-on-network-failure.md
uhoreg: Oh, that's true.  From the point of view of the owner of the fallback key, there is a difference.  I was just looking at it from the point of view of the person requesting the key.

On the other hand, if we're changing the definition of "used" to having received a message using that key, then the owner of the fallback key won't be using the server's flag to determine whether they should rotate the fallback.

proposals/4081-claim-fallback-keys-on-network-failure.md
uhoreg: I think the issue is more about when you upload a key, you need to indicate to the server whether it can pre-send the key to others, and so you'll wait until receiving an olm message using the key, rather than the server's flag, to indicate whether you should rotate.

proposals/4081-claim-fallback-keys-on-network-failure.md
ara4n: I guess the important thing is to emphasise in the threat model is that OTKs are security theatre whatever once you introduce fallback keys - given an attacker can force use of the fallback key by both exhausting the OTK pool (which leaves an audit trail), as well as simply deny the network (which doesn't leave an audit trail).

So, it feels like the only reason end up we keep OTKs is:
 a) To enjoy their (nominal) security properties for paranoid deployments which disable fallback keys
 b) To keep exercising the OTK code path, even when fallback keys are around, to help stop it regressing for deployments where fallback keys are disabled.

In which case, yes, perhaps this MSC isn't as bad as it felt at first.

proposals/4081-claim-fallback-keys-on-network-failure.md
ara4n: I think this scenario is: "what if Bob isn't told that Alice has rotated her fallback key, and tries to use a stale cached fallback key".  This is very similar to the "what if Alice restores her server from backup, and starts handing out stale OTKs" failure mode.

I think we have to consider these as wedged sessions, and keep trying to retry setup from the client (with your server nudging you to retry by waking you up with a push, or similar, when the server sees that the remote server has updated its device list).

kegsay: Agreed, this is basically the same as OTK reuse where the session is encrypted for a key the client has dropped. The frequency of these is quite different though.

Server backups will reliably cause OTK reuse currently, because we have no mechanism to tell clients that they need to reupload all their OTKs again (and even if we did, there would be a race condition where some user has claimed it during the reupload process). As a result, if during the bad deployment a user claimed 5 keys, upon rollback, the next 5 OTKs will be bad due to key reuse, guaranteed.

> what if Bob isn't told that Alice has rotated her fallback key

For this to happen the two servers need to be partitioned for time N where N is the time between uploading a new fallback key and deleting the old fallback key on the client. N is configurable, and [X3DH](https://signal.org/docs/specifications/x3dh/#publishing-keys) provides an example interval of "once a week, or once a month". Looking at the [Android source](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/java/org/thoughtcrime/securesms/jobs/PreKeysSyncJob.kt#L49) it seems they try to cycle it every 2 days, with a max time of 14 days. This feels like a long enough time for most ephemeral network partitions to resolve themselves. Therefore, the likelihood of actually seeing this is much more remote.

kegsay:
Synapse will also need https://github.com/element-hq/synapse/pull/16875 in addition to MSC4081 to allow clients to start Olm sessions when the server is down.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+sessions. This can be mitigated by cycling the fallback key (and hence deleting the private key) once it has been
+"used", with some lag time to account for slow networks.
+
+## Proposal
+
+Currently, fallback keys are _only_ claimed on key exhaustion, not due to transient network failures. This MSC
+proposes to change the semantics to allow fallback keys to be returned by the `/keys/claim` endpoint if the server
+the target device is on is unreachable. In order for servers to return fallback keys during the network failure,
+the fallback keys must be cached _in advance_ on the claiming user's homeserver. This MSC proposes adding a new
+key `fallback_keys` to the `m.device_list_update` EDU. This MSC proposes changing the spec wording (bold is new):
```
richvdh, line 26:
```suggestion
key `fallback_keys` to the [`m.device_list_update` EDU](https://spec.matrix.org/v1.9/server-server-api/#definition-mdevice_list_update). This MSC proposes changing the spec wording (bold is new):
```

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+| keys             | {string: string}        | Required:  Public identity keys. The names of the properties should be in the format  <algorithm>:<device_id>. The keys themselves should be encoded as specified by the key algorithm.                   |
+| signatures       | Signatures              | Required:  Signatures for the device key object. A map from user ID, to a map from  <algorithm>:<device_id> to the signature.   The signature is calculated using the process described at  Signing JSON. |
+| user_id          | string                  | Required: The ID of the user the device belongs to. Must match the user ID used when logging in.                                                                                                          |
+| **fallback_key** | **{string: KeyObject}** | **The fallback key for this device, if set. The format of this object is identical to the /keys/claim response for a single device. This replaces any previously sent fallback key.**                                                                         |
```
richvdh, line 42: I don't think this is practical. The `DeviceKeys` object is signed the device itself (it comes from the [`/keys/upload`](https://spec.matrix.org/v1.9/client-server-api/#post_matrixclientv3keysupload) request), which means that the server can't randomly add extra fields to it.

Instead, I think we can add a `fallback_keys` property to the top level of the content of the `m.device_list_update` EDU, which contains the `fallback_keys` property from the `/keys/upload` request.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+As a reminder, clients SHOULD rotate their fallback key when they realise it has been "used", with some lag time
+to account for federation. As per MSC2732, 1 hour is recommended. When clients change their fallback key, a new
+`m.device_list_update` EDU MUST be sent.
```
richvdh, line 63: why does delaying rotation by 1 hour help anything?

First of all: if you throw away the private part of an old key at the same time as publishing a new key, there is a race, no matter when you do it. What you need to do is publish the new key (which you might as well do immediately the key has been used) and then *later* throw away the private part of the old key.

Secondly: one hour seems awfully short, if our goal is to survive federation outages, which can go on for hours or even days.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+
+Ultra secure clients may be unhappy that fallback keys are being returned and not one-time keys, because they
+dislike the slightly weaker security properties fallback keys provide. This could be resolved by adding a flag to
+the `/keys/claim` endpoint to state whether returning a fallback key is acceptable to the client or not. If this
+flag is not set/missing, fallback keys would not be returned in place of OTKs, meaning this MSC would be entirely
+opt-in, and hence require client-side changes. However, a malicious server can trivially ignore this flag and
+return the fallback key anyway, and the client would not be able to detect this. For this reason, it feels like
+security theater to add this flag.
+
+A malicious actor who can control network conditions can force a client to use a fallback key by temporarily
```
richvdh, line 109:
```suggestion
A malicious actor who can control network conditions (but not the servers themselves) can force a client to use a fallback key by temporarily
```

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+preventing two homeservers from communicating. Previously, the only way a malicious actor could force a client to
```
richvdh, line 110:
```suggestion
preventing two homeservers from communicating. Previously, the only way such an actor could force a client to
```

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh, line 109: it would be good to make it clearer that each paragraph is a separate concern -- eg with a bullet list.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,139 @@
...
+access to the key for an extended time is a security risk, and the user does not trust their own homeserver (not
+unreasonable given this is for E2EE) then any concerns _are already present today_, just not over federation.
+
+## Alternatives
+
+Do nothing. In this scenario, if the remote server is unreachable when the client calls `/keys/claim`, the message
+will not be encrypted for that device, and the end user will be unable to decrypt the message. What's worse, this
+will persist until the client decides to retry the `/keys/claim` endpoint, which could be seconds or much longer.
+As a data point, Matrix Rust SDK currently uses [15 seconds](https://github.com/matrix-org/matrix-rust-sdk/issues/2804)
+and this is seen as very low.
```
richvdh, line 137: Another alternative is for the sending client to remember that it was unable to claim keys for a given device, and retry periodically.

richvdh (changes requested):

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: > N is configurable, and [X3DH](https://signal.org/docs/specifications/x3dh/#publishing-keys) provides an example interval of "once a week, or once a month". Looking at the [Android source](https://github.com/signalapp/Signal-Android/blob/main/app/src/main/java/org/thoughtcrime/securesms/jobs/PreKeysSyncJob.kt#L49) it seems they try to cycle it every 2 days, with a max time of 14 days.

As discussed in [the other thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4081/files#r1504079888), there is confusion here between "how often do we create a new key" and "how long do we keep old keys, having created a new one". The X3DH spec just says "eventually" for the latter:

> After uploading a new signed prekey, Bob may keep the private key corresponding to the previous signed prekey around for some period of time, to handle messages using it that have been delayed in transit. **Eventually**, Bob should delete this private key for forward secrecy.

I *think* the Signal Android app is using 30 days for this period (https://github.com/signalapp/Signal-Android/blob/940cee0f30d6a2873ae08c65bb821c34302ccf5d/app/src/main/java/org/thoughtcrime/securesms/crypto/PreKeyUtil.java#L210-L239).

Nevertheless, I agree with the principles in this discussion: If Alice and Bob's servers manage to overlap their downness for long enough, then yes Bob will be unable to message Alice. But that's an extreme case, and I don't think it should prevent us making this incremental improvement in reliability even if we still end up bolting on retries later on.

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: I've clarified the MSC to say that claiming clients can refuse to use fallback keys if they like.

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: I've added a section of the MSC to add a request parameter to `/keys/upload`.

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: Not quite sure what to do with this thread. Maybe I should add a section to the MSC to call out the issue?

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: > I guess the important thing is to emphasise in the threat model is that OTKs are security theatre whatever once you introduce fallback keys - given an attacker can force use of the fallback key by both exhausting the OTK pool (which leaves an audit trail), as well as simply deny the network (which doesn't leave an audit trail).
> 
> So, it feels like the only reason end up we keep OTKs is: a) To enjoy their (nominal) security properties for paranoid deployments which disable fallback keys b) To keep exercising the OTK code path, even when fallback keys are around, to help stop it regressing for deployments where fallback keys are disabled.
> 
That's not quite right. As I see it, OTKs guard against a *passive* attacker, who has nevertheless managed to snarf the network data, and then later gets access to [the data on] Bob's device. You don't have to be paranoid and disable fallback keys to benefit from them. I've linked to https://crypto.stackexchange.com/a/52825 in the doc, as I think it really helps explain this.

So yes, an attacker with access to the network between homeservers can now force use of a fallback key where previously no communication would happen at all. But it's far easier to claim all the OTKs than it is to get access to the network to block that /claim request, so I'm not sure it's really moving the needle?

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: updated

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: Have attempted to clarify all this.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,215 @@
...
+networking problem).
+
+Once a new key has been uploaded, the private part of the old key should be scheduled for deletion. This cannot
+happen immediately, since there may be other messages in flight which rely on the old key. This was also true of
+the original fallback keys implementation
+([MSC2732](https://github.com/matrix-org/matrix-spec-proposals/pull/2732)), however there could now be a much more
+significant delay between the old key being used to encrypt a message and that message being received at the
+recipient, and MSC2732's recommendation (the lesser of "as soon as the new key is used" and 1 hour) is inadequate
+We therefore recommend significantly increasing the period for which an old fallback key is kept on the client, to
+30 days after the key was replaced, but making sure that at least one old fallback key is kept at all
```
ara4n, line 144: I think this means that after 30d of netsplit, a user on a server which has cached an old fallback key will no longer be able to establish an Olm session to you?

This feels like a pretty major limitation which should be called out - and communicated to the sending user when it happens?

I still like the idea of warning the user in general what users they can’t communicate with (due to no OTKs, or due to expired fallback keys), so the user can go and complain and get the problem solved.

proposals/4081-claim-fallback-keys-on-network-failure.md
uhoreg: I suspect that the issue should be rare enough that it's not worth trying to solve it right now.  So at most, add something in the MSC that says something about it.

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: I think it's worth saying that, after 30 days of netsplit, it's a good bet that any to-device messages you send aren't going to be delivered for a while anyway. (In other words: what good does obtaining a fallback key do if you then can't actually send anything you encrypt with it?)

Still, you're not wrong. It's also worth saying that this 30 days is entirely under the client's control; so if you are working in an environment where you expect your homeserver to go incommunicado for 3 months, perhaps you can configure your client to keep old fallback keys that long.

As for detecting and reporting the situation to the user: yes, that might be nice, if only it can end up in a rageshake (I can hear @pmaier1's voice in my head saying "users don't want to be bothered with this sort of technical detail!"). The problem is, how can we detect it? The problem is that Alice (who is using the fallback key) has no way of knowing that Bob has expired that key, *because they are netsplit*. Timestamping the keys doesn't help, because an old key could also mean that Bob hasn't used his client for 30 days. We could maybe detect the situation once Bob comes back online, but that doesn't help the user get the problem solved in the first place.

IMHO this starts to get into questions about "how long is it reasonable for clients/servers to be offline/unreachable, and still expect messages to get delivered". FB Messenger, for example, actually logs out any devices that are unused after 30 days. We don't necessarily need to go that far, but "forever" seems an unreasonable expectation; if we actually set some expectations then we could work towards having sensible UX when that time expires.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -10,69 +10,141 @@ only be used once. However, this presents several problems:
...
+      "keys": {
+        "curve25519:JLAFKJWSCS": "3C5BFWi2Y8MaVvjM8M22DBmh24PmgR0nPvJOIArzgyI",
+        "ed25519:JLAFKJWSCS": "lEuiRJBit0IG6nUf5pUzWTUEsRVVe/HJkoKuEww9ULI"
+      },
+      "signatures": {
+        "@alice:example.com": {
+          "ed25519:JLAFKJWSCS": "dSO80A01XiigH3uBiDVx/EjzaoycHcjq9lfQX0uWsqxl2giMIiSPR8a4d291W1ihKJL/a+myXS367WT6NAIcBA"
+        }
+      },
+      "user_id": "@alice:example.com"
```
kegsay, line 92:
```suggestion
      "user_id": "@john:example.com"
```

proposals/4081-claim-fallback-keys-on-network-failure.md
kegsay: > IMHO this starts to get into questions about "how long is it reasonable for clients/servers to be offline/unreachable, and still expect messages to get delivered". FB Messenger, for example, actually logs out any devices that are unused after 30 days. We don't necessarily need to go that far, but "forever" seems an unreasonable expectation; if we actually set some expectations then we could work towards having sensible UX when that time expires.

💯 - if we actually bothered to do this then maybe we could clear out the to_device_inbox at _some_ point...

I disagree that "I still like the idea of warning the user in general what users they can’t communicate with (due to no OTKs, or due to expired fallback keys), so the user can go and complain and get the problem solved." is a useful property to be trying to preserve here. If I'm on a HS and it is down, then you cannot talk to me no matter what you try to do. In some cases there may be an alternative way of contacting me, to which I can either A) thank you for being a real-life PagerDuty and restart the server or B) shrug and say I don't actually control it, as it's on $homeserver I don't control.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -98,42 +170,46 @@ claim OTKs as Signal is not federated).
...
-
+1. Do nothing. In this scenario, if the remote server is unreachable when the client calls `/keys/claim`, the
+   message will not be encrypted for that device, and the end user will be unable to decrypt the message. What's
+   worse, this will persist until the client decides to retry the `/keys/claim` endpoint, which could be seconds or
+   much longer.  As a data point, Matrix Rust SDK currently uses [15
+   seconds](https://github.com/matrix-org/matrix-rust-sdk/issues/2804) and this is seen as very low.
+
+2. Clients could remember that they were unable to claim keys for a given device, and retry periodically. The main
+   problem with this approach (other than increased complexity in the client) is that it requires the sending
+   client to still be online when the remote server comes online, and to notice that has happened. There may be
```
kegsay, line 213: It basically rules out asynchronous communication when initially talking to someone, as both HSes need to be online at exactly the same time. This feels very suboptimal and rules out the ability to run E2EE Matrix under certain network conditions.

proposals/4081-claim-fallback-keys-on-network-failure.md
```diff
@@ -0,0 +1,215 @@
...
+|-----------------------|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `device_keys`         | `DeviceKeys`              | Identity keys for the device. May be absent if no new identity keys are required.
+| `fallback_keys`       | `OneTimeKeys`	            | The public key which should be used if the device’s one-time keys are exhausted, **or if the user's homeserver is unreachable**. [etc]
+| `one_time_keys`       | `OneTimeKeys`	            | One-time public keys for “pre-key” messages. The names of the properties should be in the format <algorithm>:<key_id>. The format of the key is determined by the key algorithm. May be absent if no new one-time keys are required.
+| **`eager_share_fallback_keys`** | **`boolean`**   | **Whether the `fallback_keys` should immediately be sent to other homeservers which have a user which share a room with this user. Omitting this property is the same as setting it to `false`.
```
richvdh, line 43:
```suggestion
| **`eager_share_fallback_keys`** | **`boolean`**   | **Whether the `fallback_keys` should immediately be sent to other homeservers which have a user which share a room with this user. Omitting this property is the same as setting it to `false`.**
```

proposals/4081-claim-fallback-keys-on-network-failure.md
richvdh: > both HSes need to be online at exactly the same time.

Whatever we do, both HSes need to be online at the same time because one has to make an HTTP request to the other. I guess you mean that the sending client, *and* both HSes, have to be online. In which case, yes I agree with you and this is a succinct summary of why this approach is insufficient.

kegsay: Yes that's what I meant, sorry.

kegsay:
Even with this MSC, we require https://github.com/element-hq/synapse/issues/11374 for offline support (so device list updates are eagerly shared too).

kegsay:
When I wrote this MSC originally, I did not think this would be a sufficiently frequent cause of UTDs. However, as we have fixed other UTD causes, this is now a significant chunk of the remaining decryption failures. As such, we should probably try to prototype this and land a solution at some point. See https://github.com/element-hq/synapse/issues/17267 for real world causes which this MSC would fix.

kegsay:
The big matrix.org outage this week caused more UTDs in the wild. This MSC would have helped some of them.

PR #4083, opened 2023-12-16
MSC4083: Delta-compressed E2EE file transfers

ara4n:
A rough proposal for delta-compressing file transfers, originally written for Third Room, but apparently i never committed it at the time - so submitting it as a MSC for posterity, in the hope that it saves some time in future next time someone wants to do incremental binary updates against a file in Matrix.  (@hughns: should we ever get back to the Matrix Files SDK, this might be of interest)

### DIFF
--- a/proposals/4083-delta-compressed-file-transfers.md
+++ b/proposals/4083-delta-compressed-file-transfers.md
@@ -1,0 +1,113 @@
+# MSC4083: Delta-compressed E2EE file transfers
+
+## Problem
+
+When collaborating on a large file of some kind, it's common to store that file in Matrix, and then need a way to
+express incremental changes to it.  For instance, in Third Room, you might store a large glTF scene graph as a GLB
+file, and then want to express a small change to it (e.g. using the editor to transform part of the scene graph).  Or
+you might want to store a change to a markdown or HTML file.
+
+Currently, your only option is to save a whole new copy of the file - or invent your own delta-compression scheme at
+the application layer. Instead, we could make Matrix itself aware of delta-compression, letting the content repository
+help users efficiently collaborate around updates to binary files, regardless of what the file is.
+
+## Solution
+
+When uploading a file, specify that it's a delta against a previous piece of content, using a given algorithm.
+
+ * `delta_base` is the mxc URL of the content the delta applies to
+ * `delta_format` is the file format of the binary diff
+   * This MSC defines `m.vcdiff.v1.gzip` to describe gzipped RFC3284 compatible binary VCDIFF payloads, picked for
+     computation efficiency rather than patch size (whereas bsdiff + bzip might provide better patch size at worse
+     computation complexity; other MSCs are welcome to propose different diff formats).
+
+Clients should upload a new snapshot of a piece of content if the sum of the deltas relative to the last snapshot
+is larger than 50% of the original piece of content
+
+For instance:
+
+`POST /_matrix/media/v3/upload?delta_base=mxc://matrix.org/b4s3v3rs10n&delta_format=m.vcdiff.v1.gzip`
+
+returning:
+```json
+{
+  "content_uri": "mxc://matrix.org/n3wv3rs10n"
+}
+```
+
+(or with the same parameters for MSC2246-style `POST /_matrix/media/v3/create`).
+
+The server tracks the graph of which deltas apply to which files, so it can only hand the relevant deltas to clients
+when they download them.
+
+For instance, when downloading a delta-compressed piece of content, a client might ask to pull in any delta dependencies
+it doesn't already have stored locally, relative to the last version that it has a full copy of:
+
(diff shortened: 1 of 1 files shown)


proposals/4083-delta-compressed-file-transfers.md
```diff
@@ -0,0 +1,113 @@
...
+
+The server tracks the graph of which deltas apply to which files, so it can only hand the relevant deltas to clients
+when they download them.
+
+For instance, when downloading a delta-compressed piece of content, a client might ask to pull in any delta dependencies
+it doesn't already have stored locally, relative to the last version that it has a full copy of:
+
+`GET /_matrix/media/v3/download/matrix/org/n3wv3rs10n?delta_base=mxc://matrix.org/b4s3v3rs10n`
+
+This would return an ordered multipart download of the deltas (once unencrypted, if needed) to apply to the base-version
```
uhoreg, line 48: For encrypted files, how do clients discover the encryption keys for each delta and the base file?

proposals/4083-delta-compressed-file-transfers.md
```diff
@@ -0,0 +1,113 @@
...
+the application layer. Instead, we could make Matrix itself aware of delta-compression, letting the content repository
+help users efficiently collaborate around updates to binary files, regardless of what the file is.
+
+## Solution
+
+When uploading a file, specify that it's a delta against a previous piece of content, using a given algorithm.
+
+ * `delta_base` is the mxc URL of the content the delta applies to
+ * `delta_format` is the file format of the binary diff
+   * This MSC defines `m.vcdiff.v1.gzip` to describe gzipped RFC3284 compatible binary VCDIFF payloads, picked for
```
uhoreg, line 20:
```suggestion
   * This MSC defines `m.vcdiff.v1.gzip` to describe gzipped [RFC3284](https://datatracker.ietf.org/doc/html/rfc3284) compatible binary VCDIFF payloads, picked for
```

proposals/4083-delta-compressed-file-transfers.md
```diff
@@ -0,0 +1,113 @@
...
+     computation efficiency rather than patch size (whereas bsdiff + bzip might provide better patch size at worse
+     computation complexity; other MSCs are welcome to propose different diff formats).
+
+Clients should upload a new snapshot of a piece of content if the sum of the deltas relative to the last snapshot
```
uhoreg, line 24: Clients must also upload a new snapshot when needed to ensure that secrecy is preserved in encrypted rooms.  e.g. if a new user joins, a new snapshot must be uploaded, otherwise the new user would need to be able to decrypt the file state from before they joined the room.

proposals/4083-delta-compressed-file-transfers.md
ara4n: i just realised the same thing :) i guess this pushes it back towards putting the delta links on the m.file events rather than the content repository, and using aggregations perhaps as a way to grab all the events needed to download a given file.

proposals/4083-delta-compressed-file-transfers.md
```diff
@@ -0,0 +1,113 @@
+# MSC4083: Delta-compressed E2EE file transfers
+
+## Problem
+
+When collaborating on a large file of some kind, it's common to store that file in Matrix, and then need a way to
+express incremental changes to it.  For instance, in Third Room, you might store a large glTF scene graph as a GLB
+file, and then want to express a small change to it (e.g. using the editor to transform part of the scene graph).  Or
+you might want to store a change to a markdown or HTML file.
+
+Currently, your only option is to save a whole new copy of the file - or invent your own delta-compression scheme at
```
MTRNord, line 10: Couldn't you also use an existing delta format like the one used by OTA updates on android and encrypt that separately here? Or is the concern that due to e2ee shenanigans, intermediate deltas are lost here? (I am not saying that this a good approach. Just an alternative that also came to mind for me)

proposals/4083-delta-compressed-file-transfers.md
ara4n: Alternatively, could be evil and reuse the same IV & Key for every diff on a given file - but calculate the actual IV as `IV' = H(IV, $content_id)`. This would mean that diffs have to be created as async uploads so you know their content_id before they can be encrypted by the client though; and the multipart download would have to include content IDs. 

I'm not convinced this is better than using an aggregation API to say "give me all the events for the diffs needed to construct this $event_id", and then firing off a tonne of parallel reqs to the media repo to grab the required media files (which is arguably only 2 roundtrips too).  But it avoids having to fiddle around with events at all.

proposals/4083-delta-compressed-file-transfers.md
ara4n: but this proposal does propose using an existing delta format (vcdiff - rfc3284) and encrypting the diffs?

MTRNord: I must have been asleep while writing this comment I guess 😱 sorry.

PR #4084, opened 2023-12-18
MSC: Improving security of MSC2244

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/2244-ext-uia/proposals/4084-mass-redactions-alterations.md)

Implementation requirements: 
* [ ] Unstable room version containing both MSC2244 and this MSC's changes.
* [ ] Example of a server processing redactions safely.
* [ ] Example of a client engaging in these rooms.

### DIFF
--- a/proposals/4084-mass-redactions-alterations.md
+++ b/proposals/4084-mass-redactions-alterations.md
@@ -1,0 +1,232 @@
+# MSC4084: Improving security of MSC2244
+
+[MSC2244](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2244-mass-redactions.md)
+introduces an idea of "mass redactions", where a single redaction event can target more than one event.
+This provides signficantly improved performance when dealing with large volumes of spam or redacting
+one's own messages, but is also irreversible.
+
+This proposal amends MSC2244 to accommodate previously-undisclosed security considerations.
+
+## Background context
+
+MSC2244 was accepted through FCP under an older process where implementation was not required before
+FCP. Some time later, the SCT noted significant process overhead when implementation happens second,
+so updated the process to require implementation before FCP can begin. MSC2244 however had already
+finished FCP at the time - it's now stuck in a state where FCP is completed, but the MSC is not
+accepted because it is not implemented.
+
+Implementation on MSC2244 was blocked on internal security context from the SCT as of March 2022,
+despite the MSC successfully completing FCP in November 2019. The concerns, now captured by this MSC,
+were discovered through a brief attempt to implement the MSC during development of another feature.
+
+After some unintended process delays, this MSC exists to cover the internal context captured throughout
+2022.
+
+With the new process in place, issues like the ones covered by this MSC are easier to catch before an
+MSC is accepted then merged. Unfortunately, because of the older process in play, the MSC needs to
+be modified by another MSC first - namely, the MSC described by this document.
+
+### Unstable implementation
+
+Under the older process MSC2244 falls under, implementation can happen *after* FCP is completed. To
+qualify as "accepted", the MSC will need such an implementation and review from the SCT before it
+can continue. It's not until the MSC is assigned a stable room version that it becomes merged to the
+specification, and unstable identifiers (including the unstable room version) can be dropped.
+
+Acceptance on MSC2244 is additionally qualified on resolution of the concerns described in this MSC,
+though the exact solutions may vary. Once both implementation and security considerations are made,
+the SCT will mark MSC2244 as accepted. The MSC would then qualify for inclusion in a (future) stable
+room version.
+
+## Problem
+
+MSC2244 provides room moderators a way to redact multiple messages in a single event, reducing overall
+traffic and therefore speed of removal. This is an incredibly useful tool when dealing with volumetric
+spam - the sender is not required to generate the same volume of redactions (an amplification attack).
+
+Mass redactions can also be used to remove an entire user from the conversation history, either
+electively or by choice of a room moderator. When performed electively, the user is essentially trying
+to erase their presence from the network/room. Though some metadata is retained, the conversation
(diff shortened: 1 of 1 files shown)


proposals/4084-mass-redactions-alterations.md
```diff
@@ -0,0 +1,232 @@
...
+This MSC also forces servers to withhold redactions until *all* target events are received and validated.
+In the case of volumetric spam, this means a server will be required to first download that spam then
+redact it. With each event being maximum 65kb in size, this could be significant network traffic. The
+alternative is a malicious sender could cause future problems or confuse clients due to trivially
+possible bugs in servers, thus exploiting individual implementation bugs to alter the conversation
+history depending on viewer.
+
+## Alternatives
+
+No significant alternatives not already discussed identified.
```
ara4n, line 210: One thing that could mitigate abuse of this would be if it was reversible within a given timeframe - ie a cooling off or antiabuse or remorse timeframe. Same for account deactivation. Both could give the user a one-time code they could use to countermand the previous destructive action. If combined with a grace period for redactions in general (which most servers implement), this could be best of both worlds? (And stop floods of reactivation requests to support@...)

turt2live: The DAG level gets complicated for all kinds of reasons, but a possible approach here might be to split redactions into two events: one which *applies* the redaction, and another which actually commits to it. https://github.com/matrix-org/matrix-spec-proposals/pull/3531 kinda does this, though a more direct approach would be to have `m.room.soft_redaction` go out to apply a redaction pending formal redaction. The operation would apply to clients (and maybe servers over the CS API but not storage?) so the effect can be quickly realized. Then, if the moderator or sender regrets that decision they can redact the soft redaction to "cancel" it. If not cancelled, either the client or the server could commit the redaction with today's `m.room.redaction` event.

Something to consider is who has the ability to undo a redaction, as it'd be a bit weird to think you've redacted your password from the public chat but then a moderator reveals it again.

wj25czxj47bu6q (changes requested):

proposals/4084-mass-redactions-alterations.md
```diff
@@ -0,0 +1,232 @@
...
+to erase their presence from the network/room. Though some metadata is retained, the conversation
+history is completely destroyed in this scenario. This causes two problems of varying severity:
+
+1. A choice has been imposed upon other members of the room, where their copy of the conversation has
+   been altered. Matrix is structured with a belief that everyone has an equal right to the conversation,
+   and a built-in tool which *easily* allows that conversation to be altered is not ideal. However,
+   servers wishing to perform such a function on behalf of their users are already able to remove/disable
+   rate limits on sending individual redaction events. Many servers over the years have elected to do
+   exactly that to erase a user from the network as best as possible. Therefore, for the purposes of
+   this MSC, it is considered a non-issue to support mass redactions in Matrix.
```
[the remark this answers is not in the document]
MTRNord: Hm but wouldnt this revoke the right to be forgotten by a user under GDPR and friends? Like mass redactions also help on that front.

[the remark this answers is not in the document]
turt2live: Relative time is difficult in Matrix, but this sounds like something for a future MSC.

proposals/4084-mass-redactions-alterations.md
KitsuneRal: I guess [MSC4117](https://github.com/matrix-org/matrix-spec-proposals/pull/4117) is a potential answer to this?

proposals/4084-mass-redactions-alterations.md
```diff
@@ -0,0 +1,232 @@
...
+not have itself (or which are illegally included in the array), though it's theoretically possible for
+a client to gain access to the unmodified `m.room.redaction` event and redact the illegal targets
+anyways. This MSC proposes changes to how MSC2244 handles redactions to eliminate potential points of
+abuse/trickery from malicious actors.
+
+## Proposal
+
+To help ensure that malicious actors are not able to effectively deactivate a user's account, the
+existing [`PUT /_matrix/client/v3/rooms/:roomId/send/:eventType/:txnId`](https://spec.matrix.org/v1.9/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
+endpoint MUST NOT accept `m.room.redaction` events in room versions supporting mass redactions. The
```
turt2live, line 81: this can probably be relaxed to "`m.room.redaction` events targeting more than one event" rather than "thou shall not redact"

proposals/4084-mass-redactions-alterations.md
```diff
@@ -0,0 +1,232 @@
...
+server MUST respond with a 400 HTTP status code and `M_BAD_STATE` error code for this condition.
+Clients are still able to redact individual events at a time using the `/redact` endpoint, if desired.
+
+To ensure that clients are able to actually use mass redactions, a copy of `/send` is introduced with
+[User-Interactive Authentication](https://spec.matrix.org/v1.9/client-server-api/#user-interactive-authentication-api)
+(UIA) optionally used by the server. The new `PUT /_matrix/client/v4/rooms/:roomId/send/:eventType/:txnId`
```
turt2live, line 87: UIA is now "legacy" and replaced in part by OAuth2: https://spec.matrix.org/v1.15/client-server-api/#client-authentication

It's not clear how to use OAuth2 for UIA-like features like this.

proposals/4084-mass-redactions-alterations.md
KitsuneRal: `m.room.soft_redaction` is additive to `m.room.redaction`, I don't think it would even require a new room version. Changing the semantics of `m.room.redaction` would probably require a new room version to make sure users can tell between reversible and irreversible redactions.

PR #4085, opened 2023-12-20
MSC4085: Improved CS-API Errors

MTRNord:
[Rendered](https://github.com/MTRNord/matrix-doc/blob/patch-4/proposals/4085-improved-cs-api-errors.md)

This is a living document at this time and is expected to take some discussion and time to figure out where there are things to improve and then finding a good error code. (Additionally checking dendrite and conduit aka the younger servers is planned for this)

For those coming by over Christmas/new year time: I likely get around to filling this in around mid January

### DIFF
--- a/proposals/4085-improved-cs-api-errors.md
+++ b/proposals/4085-improved-cs-api-errors.md
@@ -1,0 +1,40 @@
+# MSC4085: Improved C-S-Api Errors
+
+In many places of the specification there currently are no error codes defined for obvious error cases.
+This means that servers have to either decide it on their own, use `M_UNKNOWN` too often or cant handle it
+properly. Additionally this also means that clients often end up preferring to state errors as verbatism
+rather than having ways to show rich errors to users of their application.
+This MSC is trying to improve this by proposing additional errors which can be used by both servers and
+clients to improve the above stated scenarios.
+
+## Proposal
+
+The Proposal focuses on adding error codes to the C-S API primarily. The primary focus is formally defining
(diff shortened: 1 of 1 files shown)


proposals/4085-improved-cs-api-errors.md
```diff
@@ -0,0 +1,40 @@
+# MSC4085: Improved C-S-Api Errors
```
TheArcaneBrony, line 1: I wonder if message *objects* could be standardised? Eg. say you want to attach a stacktrace or code reference (eg. `/src/MyEventHandler.cs:91`) and and other info to the error.
Or should these just be added under namespaced keys?

MTRNord: The spec already does say the messages in errors are meant to be human readable. Imho having stacktraces in them is an anti goal.

TheArcaneBrony: i mean, it could have a message field? though i guess it'd make more sense if stack traces are a separate top level field?

MTRNord: The plan I have is to use the existing format. Stacktraces tbh should be a dedicated MSC as the idea of this one is mostly documenting existing errors used or errors people agreed upon using otherwise.

MTRNord:
I sadly moved on from s2s development due to the complexity of it. Its not reasonable to pursue as a sole maintainer currently. Hence I dont think I am qualified for this MSC

PR #4089, opened 2024-01-12
MSC4089: Delivery Receipts

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/delivery-receipts/proposals/4089-delivery-receipts.md)

Author is a Matrix Spec Core Team (SCT) member, employed by Element, and operates the t2bot.io service. However, this proposal is written and published as a SCT member working in the area of the Digital Markets Act (DMA) and similar initiatives. 

----

Implementation requirements:
* [ ] Client sending delivery receipts in a privacy-respecting manner (such as a user preference option).
* [ ] Server routing delivery receipts as required by the proposal.

### DIFF
--- a/proposals/4089-delivery-receipts.md
+++ b/proposals/4089-delivery-receipts.md
@@ -1,4 +1,4 @@
-# MSC4088: Delivery Receipts
+# MSC4089: Delivery Receipts
 
 A common feature among messaging clients is "delivery receipts". Similar to [read receipts](https://spec.matrix.org/v1.9/client-server-api/#receipts),
 delivery receipts denote when a client has received *and* decrypted a given message. They do not
@@ -99,12 +99,12 @@ public rooms.
 ## Unstable prefix
 
 Before FCP, clients use [`/versions`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientversions)
-to see if the server supports `org.matrix.msc4088` as an unstable feature. If the server *does* support
-the feature, clients can use `org.matrix.msc4088.delivery` in place of `m.delivery` throughout this MSC.
+to see if the server supports `org.matrix.msc4089` as an unstable feature. If the server *does* support
+the feature, clients can use `org.matrix.msc4089.delivery` in place of `m.delivery` throughout this MSC.
 
 After FCP, but before being released in the spec, clients can look for the same `/versions` feature
 flag and try to send `m.delivery` receipts to the server. If the server responds with an error, the
-client can fall back to `org.matrix.msc4088.delivery`.
+client can fall back to `org.matrix.msc4089.delivery`.
 
 After being released in the spec, clients can use that specification version in the `/versions` response
 to determine if the server supports `m.delivery` receipts. If the server does, the client can send
--- a/proposals/4089-delivery-receipts.md
+++ b/proposals/4089-delivery-receipts.md
@@ -58,8 +58,9 @@ Content-Type: application/json
 The server then implies `ts` and sends the resulting EDU to the target event's origin server. The
 delivery receipt must only be routed to the target event sender.
 
-An `m.read[.private]` receipt does *not* imply delivery, but rather that a client either does not
-support delivery receipts or was not able to decrypt the message.
+Receiving a `m.read[.private]` receipt without a delivery receipt does *not* imply
+delivery, but rather that a client either does not support delivery receipts or was not
+able to decrypt the message.
 
 Clients are welcome to implement additional requirements before sending a delivery receipt. For example,
 a user setting to disable delivery receipts in public rooms specifically, or all rooms the user is in.
--- a/proposals/4089-delivery-receipts.md
+++ b/proposals/4089-delivery-receipts.md
@@ -112,4 +112,4 @@ delivery receipts as such. If not, the previous advice still applies regarding u
 
 ## Dependencies
 
-There are no depedencies.
+There are no dependencies.
--- a/proposals/4088-delivery-receipts.md
+++ b/proposals/4088-delivery-receipts.md
@@ -1,0 +1,115 @@
+# MSC4088: Delivery Receipts
+
+A common feature among messaging clients is "delivery receipts". Similar to [read receipts](https://spec.matrix.org/v1.9/client-server-api/#receipts),
+delivery receipts denote when a client has received *and* decrypted a given message. They do not
+denote whether a user has "seen" that message, such as in a push notification.
+
+In a Digital Markets Act (DMA) world, the need for a protocol to support popular messaging features
+rises. This proposal introduces a concept of delivery receipts to Matrix, aiming to be compatible
+with potential bridges and protocol converters which need such a feature to operate.
+
+## Proposal
+
+The specification already has a general concept of [receipts](https://spec.matrix.org/v1.9/client-server-api/#receipts),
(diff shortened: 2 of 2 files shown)


proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,115 @@
...
+    "$eventid": {
+      "m.read": {
+        // per spec
+      },
+      "m.read.private": {
+        // per spec
+      },
+      "m.delivery": {
+        "@user:example.org": {
+          "ts": 1661384801651, // same as m.read[.private]
```
uhoreg, line 41: Since each client sends a delivery receipt, shouldn't we have one more level in here, giving device IDs?

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,115 @@
...
+ensure that if a client fails to receive a response it doesn't retry forever.
+
+## Potential issues
+
+If a user sends an event in Matrix HQ, they could potentially receive 50 thousand delivery receipts
+multiplied by the number of devices each of those users has. The mobile device could potentially exceed
+data limits just receiving these receipts. Servers should bundle delivery receipts to deliver them to
+the client rather than sending each one individually.
+
+**TODO**: Maybe we also limit size or require a client to opt-in with a filter param of some sort?
```
uhoreg, line 81: Random ideas:

- state event indicating whether to send delivery receipts
- only send delivery receipts in rooms up to a certain size
- sender includes a property in the event indicating whether they want to receive delivery receipts

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,115 @@
...
+POST /_matrix/client/v3/rooms/!room:example.org/receipt/m.delivery/$event
+Content-Type: application/json
+{}
+```
+
+The server then implies `ts` and sends the resulting EDU to the target event's origin server. The
+delivery receipt must only be routed to the target event sender.
+
+An `m.read[.private]` receipt does *not* imply delivery, but rather that a client either does not
+support delivery receipts or was not able to decrypt the message.
```
dkasak, line 62:
```suggestion
Receiving a `m.read[.private]` receipt without a delivery receipt does *not* imply
delivery, but rather that a client either does not support delivery receipts or was not
able to decrypt the message.
```

Was this the intended meaning?

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,115 @@
...
+rises. This proposal introduces a concept of delivery receipts to Matrix, aiming to be compatible
+with potential bridges and protocol converters which need such a feature to operate.
+
+## Proposal
+
+The specification already has a general concept of [receipts](https://spec.matrix.org/v1.9/client-server-api/#receipts),
+which is currently in use only by read (or "seen") receipts. We extend this module with a new `m.delivery`
+receipt to track individual events being received by the client.
+
+Delivery receipts are sent by all of a user's clients to report back to the event's sender that an
```
uhoreg, line 17: Do delivery receipts get sent to all of the sender's devices, or just the device that sent the event?

proposals/4089-delivery-receipts.md
turt2live: er, yes. Good catch!

proposals/4089-delivery-receipts.md
turt2live: I think I intended this to be all of the sender's devices, but very open to opinions either way.

proposals/4089-delivery-receipts.md
TheArcaneBrony, line 81: Alternatively: Perhaps we could require servers to suppress delivery receipts for public and/or large rooms? This would serve as a fallback for clients that may errorneously send these events, and would also allow servers to specify an upper bound (also see: presence).

turt2live: I think we probably want to investigate this from a privacy angle rather than technical angle for now, as I'd like privacy to drive technology rather than technology drive privacy.

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,116 @@
...
+event was decrypted successfully. They are not "up to" markers - they are for singular events. Other
+users in the room do *not* see delivery receipts for events they didn't send.
```
clokep, line 19: This is a *lot* of data: (messages in a room) * (devices in a room) might be difficult to implement efficiently.

If it *is* doable though, might make more sense for `m.read` also. 🤷

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,116 @@
...
+a user setting to disable delivery receipts in public rooms specifically, or all rooms the user is in.
+Defaults may be applied which prevent delivery receipts from being sent. It is recommended that delivery
+receipts be sent by default for encrypted private rooms (where join rules are not public).
+
+Delivery receipts should not be sent for events prior to a user's join.
+
+If a server receives a duplicate request to send a delivery receipt, it should 200 OK it. This is to
+ensure that if a client fails to receive a response it doesn't retry forever.
+
+## Potential issues
```
clokep, line 75: Is there any interaction with message edits? Do you still consider it delivered if the initial message was?

proposals/4089-delivery-receipts.md
turt2live: Yea, there's vague ideas about only using delivery receipts in "trusted" rooms, where the number of devices and such is likely to be fewer.

andrewzhurov: Having causal delivery would help us to know what others know, as their event would implicitly ack everything below it (`prev_events`). This does come with cost of having no `gaps` - when you join a room you need to download everything. That's not Client's concern however, but Home Servers'. They are "public inboxes" and they could indeed require causal delivery. Could "delivery receipts" be solved this way?

proposals/4089-delivery-receipts.md
Johennes: > Feels like there could be a thing where the sending client whether they want to receive delivery receipts by sticking a property in the event, and that would be determined by settings or room size?

https://github.com/matrix-org/matrix-spec-proposals/pull/4300 (written without knowledge of this proposal) does this.

proposals/4089-delivery-receipts.md
```diff
@@ -0,0 +1,116 @@
...
+
+Clients are welcome to implement additional requirements before sending a delivery receipt. For example,
+a user setting to disable delivery receipts in public rooms specifically, or all rooms the user is in.
+Defaults may be applied which prevent delivery receipts from being sent. It is recommended that delivery
+receipts be sent by default for encrypted private rooms (where join rules are not public).
+
+Delivery receipts should not be sent for events prior to a user's join.
+
+If a server receives a duplicate request to send a delivery receipt, it should 200 OK it. This is to
+ensure that if a client fails to receive a response it doesn't retry forever.
```
HarHarLinks, line 73: That sounds like a PUT rather than POST?

proposals/4089-delivery-receipts.md
HarHarLinks: I would consider the original delivered and the edit delivered when I get an extra delivery receipt for the replacement event. same as read receipts? am i missing something?

PR #2403, opened 2020-01-09
MSC2403: Add "knock" feature

Sorunome:
[Rendered](https://github.com/Sorunome/matrix-doc/blob/soru/knock/proposals/2403-knock.md)

Soru is also interested in providing implementations in both synapse and riot-web for this, however, as her workload is currently rather high so the development progress would likely be slowly.

Signed-off-by: Sorunome <mail@sorunome.de>

[Synapse PR](https://github.com/matrix-org/synapse/pull/6739)

Shepherd: @anoadragon453

:arrow_right: **Link to the [FCP comment](https://github.com/matrix-org/matrix-doc/pull/2403#issuecomment-775361303).**

PR #4091, opened 2024-01-23
Retrospectively re-update MSC2403 to clarify that invite->knock is *not* legal.

kegsay:
Reverts https://github.com/matrix-org/matrix-spec-proposals/pull/3850

See https://github.com/matrix-org/matrix-spec/pull/1717

Rationale: https://github.com/matrix-org/matrix-spec/issues/1710

### DIFF
--- a/proposals/2403-knock.md
+++ b/proposals/2403-knock.md
@@ -438,8 +438,8 @@ first we summarise the semantics of the proposed changes.
 
 ### Current membership
 Only users without a current membership or with their current membership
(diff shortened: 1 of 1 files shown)


richvdh:
For links: this updates #2403.

richvdh (approved):
This LGTM, but in the spirit of us having royally fscked this up in the past, maybe others would like to weigh in and double-check?

dbkr (approved):
Seems good, thanks for wading through this and getting it sorted!

PR #4061, opened 2023-10-18
Add an "MSC Checklist" document and link to it from various places

anoadragon453:
This PR adds an "MSC Checklist" to the repo. This is essentially just a list of checks to perform on an MSC before considering it accepted (or checking a box on a `@mscbot fcp merge` comment).

The document is linked from various places so MSC authors are aware of it, hopefully leading to self-checks being performed.

PR #3730, opened 2022-02-18
Fix membership state table and diagram

turt2live:
There were 2 missing cases which are legal under the [auth rules](https://spec.matrix.org/v1.2/rooms/v9/#authorization-rules):
* `invite->knock` (a fairly silly thing to do, but legal)
* `external->leave (via /kick)` (another somewhat silly thing to do, but no different than `external->ban (via /ban)`)

The state table considered the first as illegal, which is untrue.

Preview: https://pr3730--matrix-org-previews.netlify.app

Issue #1710, opened 2018-11-07
Server ACL event format follows the same m.room.aliases blunder.


jevolk:
The Server Access Control List room state event and functionality announced by #1383 utilizes large JSON arrays within the content of a single event type with `state_key=""`. This design is inferior to a granular approach.

The shortcomings of formatting events in such a manner are well detailed by #1130. These large single state events with `state_key=""` add complications to the specification and implementations:
- Client authors must properly respect the existing list contents to update the event intuitively (through the event interface). This is subject to the ABA problem.
- Additional endpoints have to be designed and specified to manipulate the list content if the former point is to be mitigated similarly to `m.room.aliases`.
- This is still subject to the ABA problem between servers even if the former point mitigates ABA between clients.
- The authorship of entries in the ACL list is hard to track. A single state event only has one sender and one timestamp indicating the last update. It is slow and difficult to conduct a larger trace of all ACL event overwrites in room history to determine the time and author of each entry in the latest list. Clients in practice simply provide an inferior UX which gives the impression the latest sender is the whole list's author.
- The maximum size of a single event limits the list. While the limit has been reasonable in practice, it is only quite barely. Consider an ideal 65536 bytes available; consider some domain/entry length of 64 bytes: `65536 / 64 = 1024`. Consider 256 byte entries: `65536 / 256 = 256`. Finally consider #1709 halving those limits when two entries are required to match every one. These numbers are not quite astronomical to reach.

#### Solution

The solution is to simply take a granular approach and design events that respect the key-value-store nature of the matrix room state system as intended. In this case, the `m.room.server_acl` event with `state_key=""` should be deprecated and `m.room.server_acl` events with `state_key="target.domain"` should emerge.

PR #4092, opened 2024-01-23
MSC4092: Enforce tests around sensitive parts of the specification

kegsay:


proposals/4092-enforce-tests.md
turt2live, line 1: Implementation requirements (in my opinion):

* [ ] Multiple MSCs touching the affected areas electing to use the requirements of this MSC, demonstrating that the solution works.
* [ ] Multiple clarifications, particularly those listed, electing to use the requirements of this MSC for the same reason.

proposals/4092-enforce-tests.md
```diff
@@ -0,0 +1,86 @@
...
+This MSC proposes enforcing extra requirements on proposals _and clarifications_ that modify these rules
+in order to ensure that any changes remain consistent and correct.
+
+## Proposal
+
+Recently, it has become apparent that there are insufficient guard rails around changes to the specification,
+such that [clarifications](https://github.com/matrix-org/matrix-spec/issues/1710) have been made to the
+specification incorrectly. These changes can have security implications on the entire Matrix ecosystem. This
+proposal proposes that any changes to the following areas of the specification MUST have end-to-end testing in
+at least one homeserver implementation:
```
MTRNord, line 17: How will this work with sytest/complement? Does that count? Since its not really one test per HS but one per practically all mainline HSs. Afaik all major ones currently test against complement and sytest currently. Does that count or is this about tests within the HS itself?

Additionally: Will there from foundation be sufficient support to get these PRs landed in those tools? In the past it felt like there wasnt sufficient manpower to takle issues/bugs/PRs on that front.

proposals/4092-enforce-tests.md
```diff
@@ -0,0 +1,86 @@
...
+
+The specification contains the rules that all clients and servers must follow in order to interoperate.
+It is crucial that these rules are correct and consistent for the health of the Matrix ecosystem.
+This MSC proposes enforcing extra requirements on proposals _and clarifications_ that modify these rules
+in order to ensure that any changes remain consistent and correct.
+
+## Proposal
+
+Recently, it has become apparent that there are insufficient guard rails around changes to the specification,
+such that [clarifications](https://github.com/matrix-org/matrix-spec/issues/1710) have been made to the
```
Gnuxie, line 14: I'm not confident at all that tests in the original MSC could have prevented this specific incident. As I understand it the chain of events started with [adding a diagram](https://github.com/matrix-org/matrix-spec-proposals/pull/3730) without carefully reading the authorization rules that prohibited the transition. That's to say, the real problem was not understanding the MSC during the process of incorporating the MSC into all relevant parts of the specification. In the description to this PR, the authorization rules that prohibited `invite->knock` were cited directly where the opposite is claimed `invite->knock (a fairly silly thing to do, but legal)`. This was only something that could have been caught if all the work around interoperating an MSC into the specification is reviewed very thoroughly. It should be almost as important as reviewing the MSC itself, and this is probably not the case. 

What is missing from @kegsay 's analysis is the PR's to the specification itself that updated the room version with the new rules and anything else relevant to this MSC. Since this change to the diagram could probably have been caused by other inconsistencies introduced after the MSC has been merged.

proposals/4092-enforce-tests.md
MTRNord: Or in other words, it feels like sytest/complement currently is touched if the dendrite or synapse devs have to touch it anyway. It seems nearly impossible to reach support as a community member working on a non "the big 3" (synapse, dendrite, conduit) implementation or to report bugs. :/ So I realistically don't see how MSC requirements will work without any way to reach devs for it whatsoever currently. That will end up simply blocking MSCs for years again. SCT made great progress to not do infinite blocking anymore so imho we need reliable complement/sytest or other alternatives before this MSC can move forward.

An alternative being that people will write random scripts as tests for their MSC. That doesn't seem to be the intention of anyone here, I think.

And yes, the above is frustration speaking. Sorry to dump that in here. But it's the topic of the MSC somewhat.

MTRNord: And to clarify: I really want this MSC. Its long missing and a major improvement to the process thats fundamentally necessary. I just dont see it possible unless there exist the necessary resources to actually do this.

proposals/4092-enforce-tests.md
Gnuxie: What's worse, is that if the diagram was added during the process of incorporating the MSC into the specification (I don't know what the timeline is on this, this could have been months after, so for the following you will have to imagine that this diagram was added during incorporation). And there are clear issues about the MSC, ie being ambiguous in the case of `invite->knock`. Then why is there not a process to highlight these ambiguities, suspend incorporation and work on clarifications?

Gnuxie: We could continue: why is the culture around the specification to not only ignore  a blatantly "silly thing to do" which could equally served as a surface for vulnerability in the future in implementations in other cases. Ignoring it would have been bad, but the SCT went further. They went onward to actually reinforce this behavior because it was assumed that an implementation had permitted it?

Gnuxie: (footnote) And if we are going to point to https://github.com/matrix-org/matrix-spec/issues/1142 as context for https://github.com/matrix-org/matrix-spec/pull/1175, it should be noted that this discussion and clarification happened within the space of a month, and bypassed the MSC and associated review process entirely.

Gnuxie: (endnote) I hadn't actually realised that the act of reinforcing `knock->invite` was not done as an MSC until writing this analysis, and if I had known that from that start, this would be written very differently. Because now it looks like you haven't even followed your own processes.

So now there's another thing to address. In a situation where we image that during incorporation of an MSC into the specification there are several errors produced that are inconsistent with the original MSC, and these passages of the specification have been lying there for a significant amount of time and have been there through several official releases. Why does the SCT believe it can unilaterally alter the text of the specification, and make room versions in one spec release incompatible with the same versions in another spec release?

ara4n: This thread is incredibly hard to follow, and seems mainly to be complaining about the problems specific to https://github.com/matrix-org/matrix-spec/issues/1710, and doing so in a particularly combative manner.

You could try to distill it down to a single coherent comment rather than a stream of consciousness and take it there. That said, if your point boils down to “the SCT messed up a MSC and a spec clarification, they suck!!1!” (which is how it comes across currently) then it may not be worth doing even that, given Kegan has already done a much better job of documenting the misadventure in that bug.

MTRNord: I dont think its right to dismiss/mark as resolved this thread.

Clearly evidence showed that _someone_, whether it is SCT or someone along the path, messed up in https://github.com/matrix-org/matrix-spec/issues/1710 (which is CLEARLY mentioned in this MSC as a core example here for the MSC reasoning!). Imho this can be improved with this MSC but not PROTECT against this. The question gnuxie here is after I think is to find a way to ensure https://github.com/matrix-org/matrix-spec/issues/1710 doesn't happen again in the future. Either by requiring deeper/closer work with the MSC author on the spec text (for example, by requiring spec text to be written by the msc author) or to have a mandatory review by the MSC author might be solutions to what gnuxie is after imho. Possibly this should have been in another place but this MSC reasons that it solves what happened in https://github.com/matrix-org/matrix-spec/issues/1710 so imho its reasonable to mention what happened here.

deepbluev7, line 14: (Summarizing @Gnuxie's comment as requested)

This issue currently summarizes chronologically, where the failure happened in that case: https://github.com/matrix-org/matrix-spec/issues/1710

Specifically the original MSC and the original spec did match up correctly, but there have been several process failures to then make the current specification incorrect. First a diagramm was modified incorrectly, then the spec was updated to match that diagramm, while insufficiently verifying the original MSCs intent and currently implemented behaviour in servers. As such having tests to cover this behaviour would probably not be sufficient. Both in the case that the tests were modified to match the new spec text or they were not modified, because their logic was deemed to already match the new spec text, they might not have caught this issue.

Instead we may need to rethink our processes for specification clarifications and modifying the text of the specification in general. In the end the whole Matrix ecosystem relies on the specification to be correct, but we have a stricter review process for proposals than we have for the specification. Possibly we need to have some variation of "SCT ticks" for any change made to the specifications text as well, even if clarifications seem small and obvious, in this case they were the source of a major mistake in the specification. We need to ensure that the specification in itself is consistent to be able to rely on tests matching what the specification says.

Additionally we may need to be stricter on modifications to older room versions. We have often accepted those to be problematic and decided to not modify them, but in this case we did modify them, believing we just fixed a minor issue in the specification text, while other times such changes did go through an MSC.

proposals/4092-enforce-tests.md
ara4n: Sytest has 11 open PRs now and 1213 closed ones - and complement has 11 open PRs now and 575 closed ones (all of which are in some state of review).  So while we don't have unlimited bandwidth available, in practice it feels like it makes reasonable forwards progress - certainly better than many other projects on github.com/matrix-org

proposals/4092-enforce-tests.md
ara4n: thanks for summarizing; agreed on all of this. I'm still not sure this is the right MSC to fix this (and it's debugging the specific failure mode of #1710) - perhaps one solution would be double-review on spec PRs. (Checkboxes feels overkill to me, though).  We should also spell out the policy on modifying old room versions explicitly (if we don't already, but I don't think we do).

proposals/4092-enforce-tests.md
```diff
@@ -0,0 +1,86 @@
...
+ - Room Versions
+ - Server-Server API:
+    * Authentication
+    * PDUs
+    * Room State Resolution
+    * Signing Events
+    * Joining Rooms
+
+All of these areas are exclusive to server implementations. This means end-to-end testing MUST test with a
+real homeserver implementation.
```
clokep, line 27: Room versions aren't exclusively server implementations. Clients must know about redaction rules, for example.

proposals/4092-enforce-tests.md
```diff
@@ -0,0 +1,86 @@
...
+
+*This MSC proposes a change in procedure to how proposals and clarifications land in the specification.
+It does not add any new endpoints or alter the Matrix protocol in any way.*
+
+The specification contains the rules that all clients and servers must follow in order to interoperate.
+It is crucial that these rules are correct and consistent for the health of the Matrix ecosystem.
+This MSC proposes enforcing extra requirements on proposals _and clarifications_ that modify these rules
+in order to ensure that any changes remain consistent and correct.
+
+## Proposal
```
clokep, line 11: If you needed a secondary example, the issues with restricted join rules (MSC3083) originally missing the needed changes to the redaction algorithm (and hence why room version 9 exists, see MSC3289).

See also some of the discussion in #4061.

PR #2530, opened 2020-05-07
MSC2530: Body field as media caption

tulir:
[Rendered](https://github.com/tulir/matrix-doc/blob/body-as-caption/proposals/2530-body-as-caption.md)

This is an alternative proposal to #2529

PR #4095, opened 2024-01-31
MSC4095: Bundled URL previews

tulir:
[Rendered](https://github.com/beeper/matrix-spec-proposals/blob/bundled-url-previews/proposals/4095-bundled-url-previews.md)

Implementations:
* Legacy Beeper Desktop, Android and iOS render incoming bundled preview data and bundle previews in outgoing messages using data from the `/preview_url` endpoint.
* The next-gen Beeper clients also generate preview data locally instead of asking the homeserver to do it.
* gomuks web: rendering: https://github.com/tulir/gomuks/pull/560, sending: https://github.com/tulir/gomuks/pull/563
* mautrix-{[whatsapp](https://github.com/mautrix/whatsapp/blob/master/urlpreview.go),[signal](https://github.com/mautrix/signal/blob/main/msgconv/urlpreview.go),[imessage](https://github.com/mautrix/imessage/blob/master/urlpreview.go)} and [beeper-imessage](https://github.com/beeper/imessage/blob/main/msgconv/urlpreview.go) bridge preview data in both directions.
* mautrix-{[discord](https://github.com/mautrix/discord/blob/v0.6.5/portal_convert.go#L543-L585),[telegram](https://github.com/mautrix/telegram/blob/v0.15.1/mautrix_telegram/portal_util/message_convert.py#L357-L408),[twitter](https://github.com/mautrix/twitter/blob/v0.1.7/mautrix_twitter/portal.py#L555-L624),[instagram](https://github.com/mautrix/instagram/blob/v0.3.1/mautrix_instagram/portal.py#L1481-L1508)} bridge preview data from the remote network to Matrix.

Signed-off-by: Tulir Asokan <tulir@beeper.com>

---

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4095#issuecomment-4014486174)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4095#issuecomment-4014511124)


proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,97 @@
...
+  object for encrypted thumbnail images. Similar to encrypted image messages,
+  the URL is inside this object, and not in `og:image`.
+* `matrix:image:size` - The byte size of the image, like in `/preview_url`.
+* `og:image` - An `mxc://` URI for unencrypted images, like in `/preview_url`.
+* `og:url` - Standard OpenGraph tag for the canonical URL of the previewed page.
+* Any other standard OpenGraph tags.
+
+At least one of `matrix:matched_url` and `og:url` MUST be present. All other
+fields are optional.
+
```
uhoreg, line 29: I think that it might help to have an example of an event that has an `m.url_previews` field.

Also maybe say something about how this fits with extensible events.  At first glance, it looks like it should fit with extensible events with no changes?  (an example for this would also be helpful.)

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,97 @@
...
+links to the user's homeserver, and prevents fake previews. However, this would
+leak the user's IP address to all links they receive, so it is not an
+acceptable solution.
+
+## Unstable prefix
+Until this MSC is accepted, implementations should apply the following renames:
+
+* `com.beeper.linkpreviews` instead of `m.url_previews`
+* `beeper:image:encryption` instead of `matrix:image:encryption`
+* `matched_url` instead of `matrix:matched_url`
```
uhoreg, line 97:
```suggestion
* `beeper:matched_url` instead of `matrix:matched_url`
```

?

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,97 @@
...
+a client using the [`/_matrix/media/v3/preview_url`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixmediav3preview_url)
+endpoint. This is a relatively good approach, but a major downside is that the
+user's homeserver gets all links the user's client wants to show a preview for,
+which means using it in encrypted rooms will effectively leak parts of messages.
+
+## Proposal
+The proposed solution is allowing clients to bundle URL preview metadata inside
+events.
+
+A new field called `m.url_previews` is added. The field is an array of objects,
```
uhoreg, line 12: Is this intended only for textual events, or can it be used on non-text events too? (images? polls?)

proposals/4095-bundled-url-previews.md
tulir: The implementations have that field without a prefix, will add a note to clarify that

proposals/4095-bundled-url-previews.md
tulir: Hmm, there's technically no reason it couldn't be used in any event, but it might be weird if an image has URL previews. The current spec also doesn't seem to define which msgtypes should get previews.

proposals/4095-bundled-url-previews.md
tulir: Added examples and note that it's compatible with extensible events

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Web clients
+Web clients likely can't generate previews themselves due to CORS and other
+such protections.
+
+Clients could use the existing URL preview endpoint to generate a preview and
+bundle that data in events, which has the benefit of only leaking the link to
+one homeserver (the sender's) instead of all servers. When doing this, clients
+would have to download the preview image and reupload it to get a persistent
+`mxc://` URI, and possibly encrypt it before uploading.
```
clokep, line 221: It would increase scope, but an alternative would be to add a flag asking the homeserver to persist the images automatically. (I.e. not clean them up.)

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,244 @@
...
+
+Alternatively, clients could simply not include preview data at all and have
+receiving clients fall back to the old behavior (meaning no previews in
+encrypted rooms unless the receiver opts in).
+
+## Security considerations
+Fake preview data as covered in potential issues.
+
+## Alternatives
```
clokep, line 230: Check out https://github.com/matrix-org/matrix-spec/blob/main/attic/drafts/url_previews.md for some other pros / cons & alternatives. I think this is essentially proposal 4 in there? The current `/preview_url` is proposal 2 in there, and it says:

> Best solution is probably a combination of both 2 and 4.

> As a first cut, let's do #2 and have the receiver hit the API to calculate its own previews (as it does currently for image thumbnails). We can then extend/optimise this to option 4 as a special extra if needed.

So sounds like overall this was the plan all along! 🎉

tulir: Oh yeah it is. I'll mention that in the alternatives section, also to cover why the other options mentioned in the doc are bad.

proposals/4095-bundled-url-previews.md
tulir: I think it'd be better to make a separate MSC if extending `/preview_url` was wanted, so I updated that paragraph to mention the possibility of such a MSC.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,261 @@
...
+
+</details>
+
+## Potential issues
+### Fake preview data
+The message sender can fake previews quite trivially. This is considered an
+acceptable compromise to achieve non-leaking URL previews in encrypted rooms.
+
+Clients may choose to ignore embedded preview data in unencrypted rooms and
+always use the `/preview_url` endpoint.
```
TheArcaneBrony, line 207: On the other hand, using it in order to disable previews would be a nice alternative to having to wrap URLs in a redirect page.
Perhaps a field to explicitly disable URL previews may be considered? (`matrix:disable_preview`?)

tulir: If you want to disable previews, simply include an empty previews array (`"m.url_previews": []`) as mentioned in the MSC under the "Receiving messages with m.url_previews" header.

When the field is present, clients shouldn't search the body text for URLs, which means only the listed URLs are previewed. If the list is empty, nothing is previewed.

tulir: Oh I guess the quoted part can be misinterpreted. Maybe I should add a clarification that "ignore embedded preview data" means "only look at `matrix:matched_url` and ignore other fields", it does not mean "ignore the `m.url_previews` list"

TheArcaneBrony: Oh that's good to know, finally an official way to disable embeds! I support this :)

proposals/4095-bundled-url-previews.md
sumnerevans: It probably doesn't make sense for a media-only message to have a URL preview, but if the media has a caption (for example using #2530), then we would want to allow for want to allow for URL previews on the caption.

tulir: Have clarified that it's meant for text messages, but isn't limited to them for future-proofing

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,278 @@
...
+one homeserver (the sender's) instead of all servers. When doing this, clients
+would have to download the preview image and reupload it to get a persistent
+`mxc://` URI, and possibly encrypt it before uploading.
+
+Alternatively, clients could simply not include preview data at all and have
+receiving clients fall back to the old behavior (meaning no previews in
+encrypted rooms unless the receiver opts in).
+
+## Security considerations
+Fake preview data as covered in potential issues.
```
ara4n, line 246: I think it's worth calling out some more explicit security concerns here:
 * This will leak the sending client's IP to the URL they are previewing.
 * The client will need to be careful not to let itself get pwned by malicious content at that URL (e.g. XML parsing exploits in the HTTP library; billion lol attacks...)
 * The client should be very careful not to preview URLs provided by other users - e.g. when replying to a message or quoting it, to stop an attacker sending a malicious URL to a user in order to discover their IP or otherwise pwn them.
    * Concretely, we don't want a world where you receive spam saying "Click reply to this message to win $20M!!! https://evil.com", where the act of replying generates a preview to https://evil.com which then harvests your IP and serves you a malformed image in its URL preview thumbnail which then pwns your app
    * Another concrete attack could be sending a user a malicious URL (hidden in a hyperlink, perhaps? or hidden by mangled UTF sequences) which hits an RFC1918 address on their network to attack them - https://192.168.0.1/<shellcode> or whatever, and encouraging the user to reply to or quote the msg

One might also want to require an allowlist of IPs the sender's spider is allowed to hit anyway, to try to avoid disasters where users are social-engineered into sending malicious URLs in general, which they never click on, but still get 'clicked on' by the URL previewer, causing chaos.

I'm sure there are a bunch more attack vectors here too...

proposals/4095-bundled-url-previews.md
ara4n: i had *totally* forgotten i wrote that (pre-MSC) doc

proposals/4095-bundled-url-previews.md
tulir: Have added more security considerations

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,305 @@
...
+as the billion laughs attack.
+
+### Local IPs should not be previewed by default
+Clients should prevent previewing non-public IP addresses by default. To do
+this, clients must check the DNS records of a domain before connecting to the
+resolved IP, as public domains may point to private IPs. For web clients, these
+limits are generally handled by the browser (see the [Private Network Access
+spec](https://wicg.github.io/private-network-access/)).
+
+## Alternatives
```
Johennes, line 275: As another alternative, Signal has a link preview feature where the sending client generates the preview by querying the external site through a privacy-preserving TCP relay operated by Signal: https://signal.org/blog/i-link-therefore-i-am/

tulir: That's already compatible with this MSC as the MSC doesn't care how the sending client generates preview data.

Probably should be mentioned as an option in security considerations though 🤔

Johennes: Ah, yes you're right. It's more like an extension / addition rather than an alternative.

tulir: Have added a mention to the security considerations section that talks about clients leaking IPs

Johennes: Looks great, thank you. 👍

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,311 @@
...
+  present in the body.
+* `matrix:image:encryption` - An [EncryptedFile](https://spec.matrix.org/v1.9/client-server-api/#extensions-to-mroommessage-msgtypes)
+  object for encrypted thumbnail images. Similar to encrypted image messages,
+  the URL is inside this object, and not in `og:image`.
+* `matrix:image:size` - The byte size of the image, like in `/preview_url`.
+* `og:image` - An `mxc://` URI for unencrypted images, like in `/preview_url`.
+* `og:url` - Standard OpenGraph tag for the canonical URL of the previewed page.
+* Any other standard OpenGraph tags.
+
+At least one of `matrix:matched_url` and `og:url` MUST be present. All other
```
zecakeh, line 27: I believe that the intent of having only the `og:url` should be clarified. What is the purpose of sending the preview of a link that is not in the `body`?

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,311 @@
...
+receiving clients should fall back to the old behavior of requesting a preview
+using `/preview_url`.
+
+Clients may choose to ignore bundled data and ask the homeserver for a preview
+even if bundled data is present, as a security measure against faking preview
+data.
+
+Clients may also choose to verify that the matched_url is present in the
+`body` field before displaying a full preview. However, in order to avoid losing
+data, clients SHOULD still display ignored entries somehow, e.g. just rendering
```
zecakeh, line 62: This looks like a security issue. It means that some links will only be displayed by clients that support this feature, which could complicate the work of moderators.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,311 @@
...
+`matrix:image:size` field can still be used for image metadata, but the
+`og:image` field should be omitted for encrypted thumbnails.
+
+If clients use the `/preview_url` endpoint as a helper for generating preview
+data, they should reupload the thumbnail image (if there is one) to create a
+persistent `mxc://` URI, as well as encrypt it if applicable. A future MSC
+could also extend `/preview_url` with a parameter to request a persistent URI.
+
+#### Receiving messages with `m.url_previews`
+If an object in the list contains only `matrix:matched_url` and no other fields,
```
zecakeh, line 52: What about objects that only contain an `og:url`?

proposals/4095-bundled-url-previews.md
tulir: There's no technical reason to require links to be in `body` to be previewed. Some other platforms allow it, so bridges may end up sending previews with no match in the message text.

proposals/4095-bundled-url-previews.md
tulir: Hmm, it'd probably be fine to not fetch a preview in those cases and just render the provided data (the url), at least I don't have any use cases for previews that both have no bundled data and are not present in `body`

proposals/4095-bundled-url-previews.md
tulir: I guess that is a potential issue (but a T&S one, not security). Might be a bigger topic though, because we won't be able to add any features to messages if we refuse everything that isn't fully represented in old clients.

proposals/4095-bundled-url-previews.md
zecakeh: In that case I believe that it should be clarified that this is a new feature introduced by this MSC, the possibility to add URL previews to the message from URLs that are not present in the text.

tulir: Oh yes, that's true

proposals/4095-bundled-url-previews.md
zecakeh: Given that we clarified that having `og:url` without `matrix:matched_url` is a new feature for URL previews, it could make sense to have the same behavior for both fields providing the URL.

But the reason I am raising this point is mostly to avoid having undefined behavior.

proposals/4095-bundled-url-previews.md
zecakeh: Sorry I should have clarified, I am not raising this as a blocking concern, but I believe it should be added to the list of potential issues.

proposals/4095-bundled-url-previews.md
```diff
@@ -31,6 +31,9 @@ URL previews are primarily meant for text-based message types (`m.text`,
 `m.notice`, `m.emote`), but they may be used with any message type, as even
 media messages may have captions in the future.
 
+Allowing the omission of `matched_url` is effectively a new feature to send URL
+previews without a link in the message text.
```
gnunicorn, line 35: maybe that shouldn't be the case though? the field is called `matched_urls`, which would indicate they need to match something. I'd add that a client might only show previews it also finds a match in the text/markup for? This would also fix the moderation-can't-see-it-issue listed below.

tulir: Spammers can put anything they want in the bundled data and say it's for an innocent link that's present in `body`, so removing that feature wouldn't help that much. It can be removed though if nobody wants that feature

tulir: Actually decided to just drop the feature for now, but with a note that it could be added back in another MSC

tulir:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

tulir:
This has been in production at Beeper for years, so

@mscbot fcp merge

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,318 @@
...
+receiving clients SHOULD fall back to the old behavior of requesting a preview
+using `/preview_url`.
+
+Clients SHOULD ensure that `matrix:matched_url` is set and that the value is
+present in `body`, and ignore the entry if not. A future MSC may change this
+recommendation to allow URL previews without a corresponding link in the body.
+
+Clients MAY choose to ignore bundled data and ask the homeserver for a preview
+even if bundled data is present, as a security measure against faking preview
+data.
```
clokep: Is there any guidance for when you would want to trust the bundled URL preview vs. ignoring it? I'm not sure if it is the spec's place to include this, but I'm wondering if you had rough guidance you were thinking of when writing this.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,318 @@
...
+are using old clients.
+
+### Generating previews will leak IPs
+The sender's client will leak its IP when it fetches previews for URLs typed by
+the user. This is generally an acceptable tradeoff, as long as clients take
+care never to generate previews for links the user did not type.
+
+For example, if a client generates reply fallbacks, it MUST NOT generate
+previews for links in the fallback. Clients SHOULD also be careful with links
+when starting to edit a message, possibly by not generating new previews at
```
clokep: I think the reply fallback bit can be removed since we don't do that anymore?

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Visibility in old clients (T&S)
+Clients that don't support this MSC will not display any of the data in the
+preview field, which could be abused by spammers if all moderators in a room
+are using old clients.
+
+### Generating previews will leak IPs
+The sender's client will leak its IP when it fetches previews for URLs typed by
+the user. This is generally an acceptable tradeoff, as long as clients take
+care never to generate previews for links the user did not type.
```
clokep: It is probably fine since the user likely visited that URL already in order to share it. Not sure if that's worth including or is obvious.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,318 @@
+# Bundled URL previews
+Currently, URL previews in Matrix are generated on the server when requested by
+a client using the [`/_matrix/media/v3/preview_url`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixmediav3preview_url)
+endpoint. This is a relatively good approach, but a major downside is that the
+user's homeserver gets all links the user's client wants to show a preview for,
+which means using it in encrypted rooms will effectively leak parts of messages.
```
clokep, line 6: It also means that users on different homeservers may get different previews, this is often seen with geographically distributed stuff, e.g. YouTube working different in Europe and the US. (Although a malicious server could also modify data for different homeservers as well.)

proposals/4095-bundled-url-previews.md
tulir: One possibility would be ignoring bundled data in unencrypted rooms (often public with untrusted users) and only using it in encrypted rooms (private, more trusted users). I guess that could be added as an example at least

proposals/4095-bundled-url-previews.md
tulir: Removed and reworded the section to address the comment below as well

proposals/4095-bundled-url-previews.md
tulir: Added a note about that

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,319 @@
...
+would have to download the preview image and reupload it to get a persistent
+`mxc://` URI, and possibly encrypt it before uploading.
+
+Alternatively, clients could simply not include preview data at all and have
+receiving clients fall back to the old behavior (meaning no previews in
+encrypted rooms unless the receiver opts in).
+
+## Security considerations
+Fake preview data as covered in potential issues.
+
```
uhoreg, line 261: It may be worth noting that allowing the sender to define the preview image gives a malicious user more opportunity to introduce abusive images into a room.

tulir: Added in https://github.com/matrix-org/matrix-spec-proposals/pull/4095/commits/5d609977f1b8de27845fade1ca24ac643d36749c

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+endpoint, effectively only using `m.url_previews` as a whitelist of URLs to
+preview.
+
+### More image uploads
+Currently previews are generated by the server, which lets the server apply
+caching and delete thumbnail images quickly. If the data was embedded in events
+instead, the server would not be able to clean up images the same way.
+
+Linked media (e.g. [MSC3911]) would alleviate this by removing the media when
+the message is removed.
```
erikjohnston, line 241: Ideally we'd have a better mechanism than this, as we events are often not deleted for a very long time (if at all). Given how much we struggle with disk usage today, I'd be keen to avoid making it worse.

I'm wondering if it's worth considering adding a flag to the media API to flag that this data is ancillary, and therefore be deleted after a period of time? We can't *force* clients to correctly mark media upload as ancillary, but may be good enough.

Additionally, we need to consider the impact of uploaded images against a user's media quota. I guess we should recommend that the client's thumbnail any image to an appropriate size? Plus if we did add the ancillary flag on metadata we could exclude such files from the user's quota (or have different quotas)

What are other people's thoughts?

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+cases where previews could be triggered by something the user did not input
+manually, such as when starting to edit a message.
+
+Clients MAY also provide extra safeguards, such as only offering a button to
+generate previews, rather than generating them immediately after the user types
+a URL. However, this is a UX decision and is therefore ultimately up to the
+client to decide.
+
+Clients could also use a privacy-preserving TCP relay to proxy all URL preview
+requests [like Signal does](https://signal.org/blog/i-link-therefore-i-am/).
```
hughns: Signal doesn't do this any more: https://community.signalusers.org/t/beta-feedback-for-the-upcoming-android-4-69-release/16219/4


```suggestion
requests like Signal [used to](https://community.signalusers.org/t/beta-feedback-for-the-upcoming-android-4-69-release/16219/4) [do](https://signal.org/blog/i-link-therefore-i-am/).
```

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
+# Bundled URL previews
+Currently, URL previews in Matrix are generated on the server when requested by
+a client using the [`/_matrix/media/v3/preview_url`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixmediav3preview_url)
+endpoint. This is a relatively good approach, but a major downside is that the
+user's homeserver gets all links the user's client wants to show a preview for,
+which means using it in encrypted rooms will effectively leak parts of messages.
+Homeserver-generated previews also mean different users could see different
+previews for the same URL, e.g. due to regional differences.
```
dbkr, line 8: Honestly this could be a good thing if it means the preview a user sees is more likely to be what they'll actually see if they click on the link.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+#### Sending preview data
+When sending previews to encrypted rooms, clients SHOULD encrypt preview images
+and put them in the `matrix:image:encryption` field. Other `og:image:*` and the
+`matrix:image:size` field can still be used for image metadata, but the
+`og:image` field is omitted for encrypted thumbnails.
+
+If clients use the `/preview_url` endpoint as a helper for generating preview
+data, they SHOULD reupload the thumbnail image (if there is one) to create a
+persistent `mxc://` URI, as well as encrypt it if applicable. A future MSC
+could also extend `/preview_url` with a parameter to request a persistent URI.
```
dbkr, line 46: I'm not sure I see how a persistent URI would be particularly helpful in this situation.

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+
+#### Receiving messages with `m.url_previews`
+If an object in the list contains only `matrix:matched_url` and no other fields,
+receiving clients SHOULD fall back to the old behavior of requesting a preview
+using `/preview_url`.
```
dbkr: Is this a thing that is expected to happen?

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+
+Clients SHOULD ensure that `matrix:matched_url` is set and that the value is
+present in `body`, and ignore the entry if not. A future MSC may change this
+recommendation to allow URL previews without a corresponding link in the body.
```
dbkr: I'm not sure what future MSCs may or may not do is relevant here tbh. Future MSCs could do absolutely anything so this is kind of a non-statement.

proposals/4095-bundled-url-previews.md
dbkr: Line 62 has answered my question so it might be clearer to lead with the proposal that URLs that aren;t mentioned in `m.url_previews` don't get previews (ie. line 67-68).

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+where each object contains OpenGraph data representing a single URL to preview,
+similar to what the `/preview_url` endpoint currently returns:
+
+* `matrix:matched_url` - The URL that is present in `body` and triggered this preview
+  to be generated.
+* `matrix:image:encryption` - An [EncryptedFile](https://spec.matrix.org/v1.9/client-server-api/#extensions-to-mroommessage-msgtypes)
+  object for encrypted thumbnail images. Similar to encrypted image messages,
+  the URL is inside this object, and not in `og:image`.
+* `matrix:image:size` - The byte size of the image, like in `/preview_url`.
+* `og:image` - An `mxc://` URI for unencrypted images, like in `/preview_url`.
```
dbkr: The fact that this and `matrix:image:encryption` refer to images in the plural confused me : I realise you're probably referring to this field over plural array entries but I think it would be clearer to make them singular (`matrix:image:size` refers to a singular image and `matrix:matched_url` refers to a singular URL).

proposals/4095-bundled-url-previews.md
```diff
@@ -0,0 +1,327 @@
...
+The proposed solution is allowing clients to bundle URL preview metadata inside
+events.
+
+A new field called `m.url_previews` is added. The field is an array of objects,
+where each object contains OpenGraph data representing a single URL to preview,
+similar to what the `/preview_url` endpoint currently returns:
+
+* `matrix:matched_url` - The URL that is present in `body` and triggered this preview
+  to be generated.
+* `matrix:image:encryption` - An [EncryptedFile](https://spec.matrix.org/v1.9/client-server-api/#extensions-to-mroommessage-msgtypes)
```
dbkr: ooi, why `matrix:image:encryption` vs `matrix:image:encrypted`

proposals/4095-bundled-url-previews.md
tulir: The bundled preview is persistent, so an mxc URI that expires in a day is no good. This paragraph is talking about clients using `/preview_url` to generate the bundled preview. When doing that, the client has to download the temporary uri and reupload it to generate a persistent uri. The unnecessary roundtrip could be removed by a future MSC to ask the server to generate a persistent uri in the first place.

proposals/4095-bundled-url-previews.md
tulir: Hmm, encrypted does make more sense, will change the stable version

proposals/4095-bundled-url-previews.md
tulir: Changed to "An `mxc://` URI if the thumbnail image is unencrypted," and similar for encrypted

proposals/4095-bundled-url-previews.md
tulir: Added a sentence in that paragraph

proposals/4095-bundled-url-previews.md
tulir: The sentence was added because it was previously a feature in this MSC. I updated it to explicitly say that now.

proposals/4095-bundled-url-previews.md
tulir: True, though the homeserver's region may have nothing to do with the user's region. Everyone in the room seeing the same content in messages is also nice.

Does this need further explanation in the MSC text or was this just a general comment?

proposals/4095-bundled-url-previews.md
richvdh: Honestly, I find all this acceptable. Given the intended usecase (displaying a small preview in the timeline), clients should not be uploading massive images as part of URL previews, so I'd rather hope that the practical overhead is minimal. If clients *do* upload huge images, that should count as an abuse, and homeservers should enforce it by counting it against the user's media quota.

However, it seems that there is a corollary here which should be mentioned in the MSC: if the url preview process ends up downloading a multi-megabyte JPEG from the target site, then clients need to watch out for that and either drop the image from the preview altogether, or resize it.

proposals/4095-bundled-url-previews.md
dbkr: Not blocking, just thought it was worth noting.

proposals/4095-bundled-url-previews.md
dbkr: Oh right, I think I see.

nvnvnvnvnvnvnv:
Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/0810dce3a6ccd8eebcf6530234f2feadf14bf384

PR #4097, opened 2024-02-06
[WIP] MSC4097: Interactions between media redirection and authentication

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/media-redirect-auth-cdn/proposals/4097-media-redirect-with-auth.md)

I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published under a generic Matrix.org Foundation role following discussions around FOSDEM 2024 regarding CDN usage in Matrix.

----

Implementation requirements:
* [ ] Write more of the MSC
* TBD once written


proposals/4097-media-redirect-with-auth.md
```diff
@@ -0,0 +1,60 @@
...
+## Proposal
+
+**TODO**: The words! Somehow the client needs to get the key (`/config`?), and needs to know how to use
+it. Also, pick an encryption algorithm.
+
+## Potential issues
+
+**TODO**: This section.
+
+## Alternatives
```
tulir, line 46: Wouldn't having the auth/expiration data in a header instead of the URL fulfill the same purpose (preventing copying links) without the added complexity of an additional layer of encryption?

turt2live: CDNs find it difficult to use headers, unfortunately.

tulir: CDNs find it easier to receive a temporary encryption key via a side channel and encrypt media on the fly than just read a header in the request and do some simple hmac check? 🤔 Fastly seems to have a header reading method at least: https://developer.fastly.com/reference/vcl/functions/headers/header-get/

A static per-media key would be easier for the CDN as it wouldn't need to know about encryption at all, and would also have the extra benefit of at-rest encryption, although that wouldn't be opt-in for clients.

turt2live: The concern was that headers are slow and might not be compatible with all providers. Will take a look though.

turt2live: ftr, Cloudflare and headers don't get along unless you have an Enterprise plan, which affects caching pretty badly.

turt2live:
This MSC is largely replaced by spec-independent approaches. It's currently the plan to publish guides rather than rely on spec features for how to implement this sort of stuff.

PR #4098, opened 2024-02-09
Proposal to use the SCIM protocol for provisioning

azmeuk:
[Rendered](https://github.com/yaal-coop/matrix-spec-proposals/blob/scim/proposals/4098-provisioning-with-scim.md)

[Implementation draft in synapse](https://github.com/yaal-coop/synapse/tree/scim)


azmeuk:
I am wondering how far to push the implementation I have started, since discussions about the MSC may change the scope of the implementation. So far it covers the nominal cases, but no error cases nor any advanced features. Basically, I added a bunch of endpoints for the basic SCIM thing, but I am not sure what is needed for this MSC to go forward.

dubiousgit:
I'm a big fan of SCIM provisioning because I think it solves a well-understood problem in the Enterprise space, and reduces the management overhead greatly. It also opens the door for default room provisioning with specific group membership I believe, which I would also approve. Waiting on this one with interest.

proposals/4098-provisioning-with-scim.md
```diff
@@ -0,0 +1,202 @@
...
+to implement the Matrix provisioning protocol.
+
+## Security considerations
+
+SCIM security considerations are related in [a dedicated paragraph](https://www.rfc-editor.org/rfc/rfc7644#section-7) in
+RFC7644.
+
+## Unstable prefix
+
+The unstable prefix to use for the root SCIM endpoint is `/_matrix/client/unstable/coop.yaal/scim/`.
```
clokep, line 183: The proposal is missing what the stable location of this would be.

proposals/4098-provisioning-with-scim.md
```diff
@@ -0,0 +1,202 @@
+# MSC4098: User account provisioning using SCIM
+
+This proposal suggests to use the SCIM protocol as a standard way to achieve
+user account [provisioning](https://en.wikipedia.org/wiki/Account_provisioning), fix several use cases uncovered
+by the specification, and in the end help reduce friction for system administrators managing Matrix servers.
+
+## Proposal
```
clokep, line 7:
```suggestion
```

This whole section seems like background, not part of the proposal. I'd rename the "Detailed implementation proposal" to "Proposal" as that gets into the actual proposed changes.

proposals/4098-provisioning-with-scim.md
```diff
@@ -0,0 +1,202 @@
...
+While this is useful from the point of view of the synapse ecosystem, this has limitations from an IAM server - or system
+administrator - perspective.
+Indeed, in cases an IAM software is connected to a multiple services,
+it implies that it implements the custom provisioning protocol for each one of those services, multiplicating the
+maintenance burden and the room for bugs.
+
+To paraphrase
+[MSC1779](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1779-open-governance.md),
+*"interoperability is better than fragmentation*". This is why this proposal advises to use SCIM as
+a standard provisioning protocol for Matrix.
```
clokep, line 72: I agree that using a standard (SCIM) over the custom implementation is better... but I think this MSC fails to mention why this is important to be in the spec as opposed to an implementation specific set of endpoints.

proposals/4098-provisioning-with-scim.md
azmeuk: Thank you for your review. I reorganized the sections and tried to describe a better chain of ideas :
-  some use-cases are not covered by the spec
- →  a provisioning protocol would cover them
- → having provisioning in the spec would help interoperability
- → adopting an existing provisioning standard would help interoperability even better, and save some design time
- → SCIM is the only relevant provisioning standard

proposals/4098-provisioning-with-scim.md
azmeuk: I am not sure about this. Would `/_scim` or `/_matrix/scim/` be good candidates?

proposals/4098-provisioning-with-scim.md
azmeuk: I attempted to answer to this in the [A common protocol for the Matrix ecosystem](https://github.com/yaal-coop/matrix-spec-proposals/blob/scim/proposals/4098-provisioning-with-scim.md#a-common-protocol-for-the-matrix-ecosystem) paragraph.

sandhose:
I'm sceptical of adding SCIM specifically to the Matrix C-S API.
Fundamentally, Matrix isn't a user and identity management API, it's not supposed to be.
It doesn't mean however that multiple homeserver implementation can't agree implementing SCIM with a specific schema, just that I don't think it makes sense within the Matrix spec.

What does make sense would be to define Matrix-specific SCIM schemas (mostly the MXID, as everything else has a 1-1 mapping with existing SCIM attributes?), but I wouldn't tie it to Matrix-specific auth, nor to a specific C-S endpoint

pierreozoux:
@sandhose @pmaier1 currently, from my understanding, [Matrix spec has a user and identity management API](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3register), is it correct? (sorry, I'm no specialist of matrix spec, discovering this world)

The idea of this proposal is to first add a new standard API for user provisioning.

azmeuk:
Closing this, as the synapse implementation has been rejected, and the discussions conclude that the MSC might be out of scope.

PR #4102, opened 2024-02-15
MSC4102: Clarifying precedence in threaded and unthreaded read receipts in EDUs

kegsay:


proposals/4102-unthreaded-read-receipts.md
```diff
@@ -0,0 +1,160 @@
...
+
+This becomes a major problem when the threaded receipt "wins" and is returned _instead of_ the unthreaded receipt.
+Some clients, such as Element X, do not handle threads and will ignore receipts with a non-main `thread_id`, assuming that
+clients will also be sending unthreaded receipts. When the threaded receipt "wins" however, these clients will simply
+not see the receipt, despite the sender sending it.
+
+## Proposal
+
+When a server is combining receipts into an EDU, if there are multiple receipts for the same (user, event, receipt type), always
+choose the receipt which is unthreaded (has no `thread_id`) when aggregating into an EDU.
```
turt2live, line 60: Combining leads to data loss, which is possibly why we still have stuck notifications/badge counts. If a threaded *and* unthreaded receipt are queued for sending, those should be independent EDUs because maps can't support both.

kegsay: Losing the threaded receipt isn't really data loss though, which is what the "Potential Issues" section talks about.

turt2live: hmm, fair point. I'm not sure this needs to be an MSC then - it feels like a clarification that can be made directly to the spec, given it's effectively talking about a data optimization and not functional specified behaviour.

andybalaam (approved):
Looks good to me. I'd prefer to change the format but that would be a lot of pain and this proposal will fix the issue.

Related: https://github.com/matrix-org/matrix-spec/issues/1684

proposals/4102-unthreaded-read-receipts.md
```diff
@@ -0,0 +1,160 @@
+# MSC4102: Clarifying precedence in threaded and unthreaded read receipts in EDUs
+
+*This MSC assumes some knowledge around threaded receipts. Please read
+[MSC3771: Read receipts for threads](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3771-read-receipts-for-threads.md)
+as for brevity some concepts won't be re-explained.*
+
+This proposal clarifies previously undefined behaviour around how read receipts are expressed over
+the CSAPI and SSAPI. This clarification reliably provides clients with sufficient information to
+determine read receipts, without placing additional burdens on client implementations.
```
clokep, line 9: https://github.com/element-hq/synapse/pull/16927 doesn't seem to make any changes to the server-server API and Synapse [currently will send multiple EDUs over federation](https://github.com/matrix-org/synapse/blob/be65a8ec0195955c15fdb179c9158b187638e39a/synapse/federation/sender/per_destination_queue.py#L219-L248) if there's both an unthreaded and threaded receipt (or just multiple threads).

This is since the EDU format is a bit different there. It maps room ID -> receipt type -> user ID -> {event IDs, data: {ts, thread ID (if applicable)}. (The client-server API maps event ID -> receipt type -> user ID -> {ts, thread ID (if applicable)}.)

kegsay: Indeed, which is fine. Federation can send multiple EDUs if it wants to, so long as at the end of the day the receiving server has the right receipts to combine them in `/sync`. The Complement test does check that this MSC works over federation.

clokep: I guess my point is that there isn't undefined behavior of the SSAPI.

proposals/4102-unthreaded-read-receipts.md
```diff
@@ -0,0 +1,160 @@
+# MSC4102: Clarifying precedence in threaded and unthreaded read receipts in EDUs
```
MadLittleMods, line 1: I assume this fixes https://github.com/matrix-org/matrix-spec/issues/1727 ?

(update PR description if so)

PR #4103, opened 2024-02-15
MSC4103: Make threaded read receipts opt-in in /sync

ara4n:


proposals/4103-threaded-read-receipt-sync-filter.md
turt2live, line 1: Implementation requirements:

* [ ] Client using the opt-in flag to enable the behaviour.
* [ ] Client *not* using the opt-in flag which clearly fixes the bug described.

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
...
+would automatically ignore them.  However, this would require threaded clients to send double the read receipts whenever
+the user reads the main thread - both an `m.read` and an `m.read_thread`, which feels inefficient.  (It would however
+mean that [MSC4102](https://github.com/matrix-org/matrix-spec-proposals/pull/4102) would not be necessary, as the RR types
+would be clearly distinct).
+
+Instead, we propose that they continue to follow MSC3771 (i.e. putting a `thread_id` on the `m.read` RR).  However, we
+give threaded clients the ability to explicitly opt-in to threaded read receipts in the form of a new sync filter
+param. If clients explicitly include `threaded_read_receipts: true` on their sync filter, then the server will send
+them `m.read` events as received by the server.  If this filter is missing or false, then the server must only send the
+client `m.read` events whose `thread_id` is missing or `main`.
```
bnjbvr, line 54: I think this is an API breaking change, because one would expect to receive threaded receipts from /sync by default. So, to preserve backwards compatibility with clients, how about making it opt-out? That is, thread receipts would be sent if the filter is `true` or missing, and not sent if it's set to `false`.

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
...
+## Problem
+
+When we added threaded RRs in [MSC3771](https://github.com/matrix-org/matrix-spec-proposals/pull/3771) we broke
+existing (i.e. non-thread-aware) clients from reliably being able to calculate whether a room is unread, given threaded
+read receipts are interpreted as unthreaded ones.
+
+For instance, let's assume Alice is logged in on two clients: a desktop client which is thread aware, and a mobile
+client which is not.
+
+ * Alice has a bunch of unread msgs on her main timeline and a bunch of unread threads.
```
bnjbvr, line 12:
```suggestion
 * Alice has a bunch of unread messages on her main timeline and a bunch of unread threads.
```

(and a few times below too)

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
...
+ * The threads are newer than the main timeline msgs.
+ * She reads a the most recently active thread on desktop, which sends a threaded RR.
```
bnjbvr, line 14:
```suggestion
 * She reads the most recently active thread on desktop, which sends a threaded RR.
```

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
...
+
+ * If Alice reads a thread on desktop, then her mobile doesn't care.
+ * If Alice reads a main timeline msg on desktop, then her mobile will want to show the flattened timeline as read up to
+   that msg.
+ * If Alice reads a given msg on mobile, then she sends an unthreaded RR, which can mark all threads prior to that point
+   as read (for threaded clients). Non-thread-aware clients will be synced to the same RR point, as you'd expect.
+
+## Proposal
+
+Non-threade-aware clients should only act on unthreaded RRs.
```
bnjbvr, line 35:
```suggestion
Non-thread-aware clients should only act on unthreaded RRs.
```

proposals/4103-threaded-read-receipt-sync-filter.md
kegsay: ..if we're having to add a filter to filter them out for thread unaware clients then why not just ask clients to check for thread_id: main? I think Matthew's point is that it was a breaking change to include threaded RRs by default, and this is trying to fix it. This puts the onus on threaded clients to add this filter flag to ask for threaded receipts. I agree it is a breaking change from the current status quo though.

bnjbvr: I *think* this MSC also implies a server-side change: now a server can't merge a threaded receipt with an unthreaded receipt on the same event id anymore, because two clients might request the two different flavors of the filter. Am I understanding it right?

kegsay: I don't see this anywhere in the MSC. The proposal is just:
 - add `threaded_read_receipts` as a sync filter param.
 - default to ??? if missing (default=true doesn't break threaded clients today, default=false breaks threaded clients but _unbreaks thread unaware clients_ which were broken when MSC3771 landed.
 - if set, send ALL THE RECEIPTS. If unset, only send unthreaded/thread_id:main receipts.

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
+# MSC4103: Make threaded read receipts opt-in in /sync
```
kegsay, line 1: This should have been done when MSC3771 landed really, because now it is going to break clients which implement MSC3771 in a minor way (they need to set this filter to true). This will however _unbreak_ unthreaded clients which will now be able to process receipts correctly again.

proposals/4103-threaded-read-receipt-sync-filter.md
```diff
@@ -0,0 +1,77 @@
...
+
+As a result, to get sensible room read state semantics for non-thread-aware clients, they would all need to become aware of
+the new `thread_id` field, breaking existing clients, even if they don't care about threads.  This feels contradictory and
+suboptimal.
+
+We could solve this by making threaded RRs a different event type (e.g. `m.read_thread`), so non-thread-aware clients
+would automatically ignore them.  However, this would require threaded clients to send double the read receipts whenever
+the user reads the main thread - both an `m.read` and an `m.read_thread`, which feels inefficient.  (It would however
+mean that [MSC4102](https://github.com/matrix-org/matrix-spec-proposals/pull/4102) would not be necessary, as the RR types
+would be clearly distinct).
```
clokep, line 48: This was rejected by MSC3771 FWIW (see https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3771-read-receipts-for-threads.md#receipt-type). Not saying it shouldn't be revisited, just noting this.

PR #4106, opened 2024-02-19
MSCXXXX Join as Muted

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-spec-proposals/blob/FSG-Cat-Join-As-Muted/proposals/xxxx-join-as-muted.md)

This MSC enhances membership screening capabilities. 

### DIFF
--- a/proposals/xxxx-join-as-muted.md
+++ b/proposals/xxxx-join-as-muted.md
@@ -1,0 +1,76 @@
+# MSCXXXX: Join as Muted
+
+## Introduction 
+
+Currently in matrix we lack the ability to implement proper membership screening mechanisms. This is
+due to that we cant use PL based screening as it has all the problems that MSC3909 and MSC3907 covers.
+
+This proposal aims to by extending MSC3909 solve this problem as membership screening is a function that
+is desired by some end users. Stakeholders like The Draupnir Project that the Author does participate
+in can not proceed to solve these end user concerns in a satisfactory manner without mechanisms like
+those that this proposal enable.
+
+This Proposal resolves this problem by introducing a mechanism where all join rules flip from transitioning
+membership to `join` from a previous state like `leave`,`invite`,`knock` or not having a membership at all to `mute`.
+
+The idea with this being that if you join a room as muted you cant interact in a race condition way where
+you get to fire of spam before automatic tooling mutes you pending membership screening.
+
+
+## Proposal
+
+A room that uses Join as muted operating mode will put a `default_membership` key in its `m.room.join_rules`
+this key can only for now contain either `join` or `mute` as its state to indicate normal or altered operating
+mode. If this is missing from the event normal behaviour is assumed. 
+
+When `default_membership` equals `mute` trying to will instead of setting your membership to `join` will set it
+to `mute` with a special consideration being that if you do not have a previous membership event, you are allowed
+to set profile information. This is to make it so you have valid profile information for your initial join as muted
+members are not allowed to update profile information usually to avoid an attack vector.
+
+To be written. Proper list of all the transitions in detail.
+
+### Authorisation rules.
+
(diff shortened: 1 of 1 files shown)


proposals/4106-join-as-muted.md
```diff
@@ -0,0 +1,79 @@
...
+access controls can do. And as for the why not just mute them on join issue that is described earlier.
+
+Also the alternative of making `default_membership` be a different but Boolean key was not picked as to
+enable this mechanism to be extended in the future without causing client side stakeholders to be confused.
+This design choice encourages being adaptable to that this can change in the future and therefore avoiding
+a v11 situation. (Room version 11 caused Element iOS to crash when trying to open a v11 room.)
+
+A separate state event was dismissed due to that it makes the most sense to put this in the join rules to the author.
+This is a weak alternatives case the author does recognise and is therefore left open as a possible alternative for now.
+
```
tezlm, line 58: Other things that currently exist:

- Setting the default power level. Currently the closest thing to join as muted, but it causes issues with the power level state event when lots of users are unmuted.
- `knock` join rule: This exists for the use case of screening people before they join.

Between setting default power levels for "most users will stay muted" and knock for "most users will be unmuted", I'm not sure what the purpose of this msc is? It does combine the two use cases together but would end up adding a third way to implement default mutes a la xkcd: standards. My two best guesses are that `knock` implies that most people will become unmuted while `mute` makes no such implication, and `mute` could allow previewing the room (...though, having a history_visibility rule to allow invite/knock users to join would be nice). What else could mute do that knocking doesn't?

FSG-Cat: Default powerlevels are adressed as completely and utterly unsuited for this usecase the prior art of MSC3909 can explain it a bit more. But lets summarise that Powelevels are ridicoulously expensive state res wise compared to memberships in my understanding and also have the even bigger achiles heel of limited user count that can be affected as a fundemental and unfixable flaw under current models. (Multiple events have been dismissed wholesale for now)

`Knock` is intresting but it has the flaw of that you cant join the room in the sense that you have to solve Peeking over federation to make Knocking a viable alternative in my opinion as you might want to see if the room is for you for example before trying to pass membership screening. And also it requires you to rotate join rules event when you go under attack something that has historically been stupidly prone to just creating a state reset pain point.

So Knock will have to be dismissed as an alternative as its a true flaw of the MSC that its not addressed. Powerlevels are tho dismissed wholesale properly i thought but i might need to do a second pass.

tezlm: This begs the question why have a separate default-mute join rule type rather than improve peeking over federation - it looks like it'll be implemented anyways and will obsolete this msc?

FSG-Cat: Tbh i would be happy to have this be redundant if that materialised and was found to be viable.

proposals/4106-join-as-muted.md
```diff
@@ -0,0 +1,86 @@
...
+### Authorisation rules.
+
+To be written. Contributions welcome. (Please note that this MSC builds on MSC3909)
+
+## Potential issues
+
+This MSC comes with the disadvantage of even further complicating membership transitions than they are already.
+The author dismisses this concern due to that this is the only viable to the author way of implementing this feature.
+
+## Alternatives
```
turt2live, line 45: Implementing this at the DAG level will be quite difficult, I suspect. The auth rules are not time-relative, so we'd need another event to say "this user is no longer silenced". (also, we prefer "silence" to "mute"). 

Instead, it may be easier to implement this in a time-relative system like policy servers. The policy server could be configured to silence people for the first 10 minutes or until a moderator rescues them from the condition. This would not require spec changes, either - it could be implemented with relative ease.

PR #4107, opened 2024-02-19
MSC4107: Feature-focused versioning

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/feature-versioning/proposals/4107-feature-focused-versioning.md)

I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published as a member of the SCT.

----

Implementation requirements:
* [ ] Should verify the design fixes the problems before implementing 😇

### DIFF
--- a/proposals/4107-feature-focused-versioning.md
+++ b/proposals/4107-feature-focused-versioning.md
@@ -9,7 +9,7 @@ Reading material:
 
 Implementation of a feature in Matrix can span multiple different systems and require multiple different
 identifiers. Typically, a feature begins its implementation as an *unstable* feature flag attached to
-a particular MSC. As that MSC progresses, the feature flag may change to accomodate sub-versioning of
+a particular MSC. As that MSC progresses, the feature flag may change to accommodate sub-versioning of
 the proposed feature. When the MSC finally finishes a `merge` FCP, implementations can then use *stable*
 identifiers - the ones described in the actual normative sections of the MSC (`m.*`). Typically, clients
 require further feature negotiation to determine if they can actually use that feature though, because
@@ -23,7 +23,7 @@ There are several issues with this approach. Most obviously, clients want to kno
 mechanics outside the versioning scheme can ensure MSCs are written up as formal spec.
 
 In a related concern, server implementations *often* lag behind the spec release cycle due to the size
-and complexity of the spec. The spec itself is very monolothic, and that causes friction when a server
+and complexity of the spec. The spec itself is very monolithic, and that causes friction when a server
 wishes to support a given subset or limited scope, or when a server implementation is trying to get
 started and immediately falls behind on client support. Unfortunately, reducing the release cadence
 from quartly to biannually does not have an effect of easier releases to implement - the releases just
--- a/proposals/4107-feature-focused-versioning.md
+++ b/proposals/4107-feature-focused-versioning.md
@@ -44,7 +44,7 @@ is modified to add a new `features` key. The existing `unstable_features` and `v
     "msc3440": ["org.matrix.msc3440", "fcp-stable", "v1.4", "v1.5", "v1.6"],
     "msc1772": ["org.matrix.msc1772"],
     "encryption": ["r0.0.0", /* ... */, "v1.9"],
-    "msc2659": ["fcp-stable:msc4107"],
+    "msc2659": ["fcp-stable:msc2659"],
     "core": ["r0.0.0", /* ... */, "v1.9"]
   },
 
--- a/proposals/4107-feature-focused-versioning.md
+++ b/proposals/4107-feature-focused-versioning.md
@@ -41,7 +41,7 @@ is modified to add a new `features` key. The existing `unstable_features` and `v
 {
   // New field
   "features": {
-    "msc3440": ["org.matrix.msc3440", "fcp-stable", "v1.4", "v1.5", "v1.6"],
+    "msc3440": ["org.matrix.msc3440", "fcp-stable:msc3440", "v1.4", "v1.5", "v1.6"],
     "msc1772": ["org.matrix.msc1772"],
     "encryption": ["r0.0.0", /* ... */, "v1.9"],
     "msc2659": ["fcp-stable:msc2659"],
--- a/proposals/4107-feature-focused-versioning.md
+++ b/proposals/4107-feature-focused-versioning.md
@@ -1,0 +1,241 @@
+# MSC4107: Feature-focused versioning
+
+**Disclaimer**: This MSC is an independent and initial idea which requires significant review and iteration.
+
+Reading material:
+* https://spec.matrix.org/v1.9/#specification-versions
+* https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientversions
+* https://spec.matrix.org/v1.9/client-server-api/#modules
+
+Implementation of a feature in Matrix can span multiple different systems and require multiple different
+identifiers. Typically, a feature begins its implementation as an *unstable* feature flag attached to
+a particular MSC. As that MSC progresses, the feature flag may change to accomodate sub-versioning of
+the proposed feature. When the MSC finally finishes a `merge` FCP, implementations can then use *stable*
+identifiers - the ones described in the actual normative sections of the MSC (`m.*`). Typically, clients
+require further feature negotiation to determine if they can actually use that feature though, because
+it hasn't yet been assigned to a specification version. This leads to *unstable* feature flags to denote
+the *stable* state, such as `org.matrix.msc4107.stable`. Some time after that, the stable MSC gets
+converted to formal spec, and that spec gets released as an assigned number. Clients are then able to
+detect feature support using that specification number.
+
+There are several issues with this approach. Most obviously, clients want to know if they can use
+*features*, but the versioning prioritizes specification. Specification is important, however other
+mechanics outside the versioning scheme can ensure MSCs are written up as formal spec.
+
+In a related concern, server implementations *often* lag behind the spec release cycle due to the size
+and complexity of the spec. The spec itself is very monolothic, and that causes friction when a server
+wishes to support a given subset or limited scope, or when a server implementation is trying to get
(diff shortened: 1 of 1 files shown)


proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+the spec, and clients would in practice be checking `core + msc4107` support of the server. This still
+encourages the server to keep up to date on the core functionality, and push for complete features as
+well.
+
+Another possible way to ensure full compliance with the specification is The Matrix.org Foundation C.I.C.
+running a "certification" program to verify a given implementation's level of spec support.
+
+**TODO**: There's probably other issues. They should be documented.
+
+## Alternatives
```
turt2live, line 226: Raised by @ara4n : Another option is to promote the use of `unstable_features` more, saying that if implementations haven't yet filled the delta of `$current_version` and `$target_version`, they continue to use (un)stable prefixed feature flags that clients should expect. 

I haven't really thought about consequences here, but we may need to adjust the process to encourage *servers* to drop unstable flags after 2 months from spec release, but inform clients of a different timeline (>2 months, potentially significantly so).

proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+  }
+}
+```
+
+... and that's it. If the second MSC were to change the feature dramatically, it may be best to use
+a new feature ID instead. For example, if "threading v2" is completely incompatible with "threading v1",
+then a new feature ID would be recommended. The SCT is responsible for determining when/if a new feature
+ID should be used.
+
+## Potential issues
```
turt2live, line 210: A possible concern is when features overlap and depend on each other. For example, threads requiring sync changes. Representing this in `unstable_features` isn't impossible (see https://github.com/matrix-org/matrix-spec-proposals/pull/4107/files#r1495129299 ), though I think this MSC makes it clearer/different. 

Threads being assigned `msc3440` as an example, and `/sync` being considered `core`, a client would look for a compatible version of each. If something like [MSC4103](https://github.com/matrix-org/matrix-spec-proposals/pull/4103) comes along, the server would be required to implement `msc3440: ["org.matrix.msc4103"], core: ["org.matrix.msc4103"]` before clients will use it in earnest.

turt2live, line 210: Mentioned by @tulir : A potential benefit or issue, depending on how you look at it, would be that a server could support `v1.3` of everything except `appservice`, where it deliberately downgrades to `v1.1` due to default config or something. This would match Synapse's current behaviour.

tulir: To clarify, Synapse defaults to only sending the `Authorization` header in appservice transactions, which means it's not compatible with v1.1, v1.2 and v1.3 of the appservice spec (which required the `access_token` query param). Synapse still advertises support for all spec versions even though it's not compatible with the older versions with the default configuration. With this MSC, it could accurately advertise what it supports (e.g. `"appservice": ["v1.4", "v1.5", ...]` and `"core": ["v1.1", ...]`)

clokep: > Synapse still advertises support for all spec versions even though it's not compatible with the older AS spec versions with the default configuration.

Synapse still advertises support for all **client** spec versions. It isn't really clear if `/_matrix/client/versions` applies to other aspects of the spec or not (e.g. appservices, federation, etc.).

proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+    "org.matrix.msc4107": true
+  }
+}
+```
+
+The new `features` key maps a feature ID to array of supported versions by the server. The supported
+versions are a combination of unstable feature flags (like `org.matrix.msc4107`), the string `fcp-stable`,
+and specification releases. The feature ID is typically the MSC number, though for historical modules
+in the Client-Server API some are named features. If a given feature spans multiple MSCs, it is expected
+that an MSC to allocate a feature ID is created.
```
clokep, line 63: Using MSC numbers feels really awkward from a client developer point of view. You really need to understand the whole way the spec changes over time, the MSC process, etc. when (as a client developer) you might wish to just check "are threads supported?". It feels like process leaking into the protocol.

proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+## Potential issues
+
+This approach could fragment implementation slightly, allowing servers to selectively implement a given
+MSC up to the "latest" spec release, but exactly zero other features receive the same treatment. This
+situation is how XMPP largely works: implementations are a mix of core functionality and whatever
+desirable XEPs are needed. We at Matrix want to avoid this. Though the MSC doesn't go into detail about
+it currently, the author's theory is that a `core` feature ID could be assigned to most functions in
+the spec, and clients would in practice be checking `core + msc4107` support of the server. This still
+encourages the server to keep up to date on the core functionality, and push for complete features as
+well.
```
clokep, line 219: I think this will cause a lot of fragmentation. Maybe that's OK, but it definitely needs to be acknowledged.

Would it make sense to consider the versions in light of the feature profiles instead? Does that help at all?

proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+
+Another possible way to ensure full compliance with the specification is The Matrix.org Foundation C.I.C.
+running a "certification" program to verify a given implementation's level of spec support.
```
clokep, line 222: This seems like it would just be *another* hurdle, i.e. making it harder to solve the problem this MSC outlines.

proposals/4107-feature-focused-versioning.md
```diff
@@ -0,0 +1,241 @@
...
+There are several issues with this approach. Most obviously, clients want to know if they can use
+*features*, but the versioning prioritizes specification. Specification is important, however other
+mechanics outside the versioning scheme can ensure MSCs are written up as formal spec.
+
+In a related concern, server implementations *often* lag behind the spec release cycle due to the size
+and complexity of the spec. The spec itself is very monolithic, and that causes friction when a server
+wishes to support a given subset or limited scope, or when a server implementation is trying to get
+started and immediately falls behind on client support. Unfortunately, reducing the release cadence
+from quartly to biannually does not have an effect of easier releases to implement - the releases just
+become chunkier (or don't happen at all because it doesn't feel "worth it" to release).
```
clokep, line 30: I would personally be very interested in understanding why this is -- are there particular features that are extremely hard to implement? Is there just a lot of historical features to catch-up on?

proposals/4107-feature-focused-versioning.md
turt2live: The appservice API extends the client-server API, and so it should be included in `/versions` (appservices are clients by definition). Federation would be out of scope here.

proposals/4107-feature-focused-versioning.md
turt2live: Yea, I'm not sure what the best balance is here. In the draft that wasn't published the feature ID was actually named "threading" and such, but then we either require an MSC to assign the feature ID or an unstable prefix, at which point we've fragmented the purpose of this system back into unstable/stable/different-stable.

proposals/4107-feature-focused-versioning.md
turt2live: Consider how?

proposals/4107-feature-focused-versioning.md
turt2live: We've always kinda wanted to do this in some capacity, but yea: the alternatives are not fully considered here.

proposals/4107-feature-focused-versioning.md
turt2live: From my perspective it's complexity in the spec. Clients naturally work at a cadence which is significantly faster than the server, but servers which don't have full time engineering teams behind them can't advertise feature support until they've implemented the *entire* specification at that version. Clients then move forward and start dropping old versions, which breaks compatibility and affects user experience of the protocol.

This MSC aims to solve the last part of that without introducing a ton of fragmentation: servers already have support for plenty of features, though not necessarily enough to advertise full compatibility with a given spec version. Clients can then use this to utilize the features the server supports without arbitrarily having to say that the whole server is incompatible because an unrelated section of the spec hasn't been addressed yet in the server.

Overall, the goal is to get features in front of users sooner, which is part of the reason why unstable implementation exists in the process. This is an evolution of that to allow those features to *stay* in front of users without having to rely on unstable prefixes for an eternity (as that's the current workaround).

clokep: > This MSC aims to solve the last part of that without introducing a ton of fragmentation

I fail to see how it aims to protect against fragmentation. It seems each individual feature could be decided to be supported?

turt2live: The specific mechanic isn't yet figured out. The MSC presents making `core` a thing as the likely mechanism, but other options could include a compliance program run by the Foundation. Personally, I prefer the `core` library being larger and therefore an active dependency in most features, thus requiring servers to implement a modern version of the spec before clients turn on the feature, but still being careful to avoid making the requirement constantly be `$latest`.

richvdh: Largely reiterating clokep's question here. It seems, from discussion elsewhere, that we aren't sure whether the desired situation is:

 * Servers implement a given spec version (say, v1.1) with a few extra features on top (say, private read receipts), or:
 * Servers implement 90% of a given spec version (say, v1.8), but find a few features hard to implement (say, edits), which means they can't advertise support for v1.8.

This MSC appears to assume the former, but @ara4n has assumed the latter in conversation. I am therefore worried we are solving the wrong problem, and I would like to see input from maintainers of the affected servers (Dendrite and Conduit primarily, I assume?) before discussing further.

turt2live:
This doesn't feel like it's the correct solution to the problem raised, so self-rejecting it.

PR #4110, opened 2024-02-23
MSC4110: Fewer Features

cloudrac3r:
[Rendered](https://github.com/cloudrac3r/matrix-spec-proposals/blob/4110-fewer-features/proposals/4110-fewer-features.md)

This is hopefully a progression from [Poorer Features](https://github.com/matrix-org/matrix-spec-proposals/pull/3968).

Signed-off-by: Cadence Ember <cadence@disroot.org>

### DIFF
--- a/proposals/4110-fewer-features.md
+++ b/proposals/4110-fewer-features.md
@@ -1,0 +1,383 @@
+# MSC4110: Fewer features
+
+The Fewer Features proposal aims to make Matrix bridges work better with other platforms. Currently, it is hard for clients and Matrix users to know if what they send in a bridged room will be supported or not. The goal of Fewer Features is to let bridges communicate which Matrix features are supported on the external platform. Then, clients and bots can change how they act based on what the external platform supports.
+
+In a nutshell, bridges can set a state event that says which Matrix features are discouraged or disallowed in this room. Clients and bots can use this information to disable features, change UI elements, or change the way events are sent. This will give the best possible experience to people on the other side of the bridge. It will also reassure Matrix users that their actions are working properly, because incompatible actions simply cannot be sent.
+
+
+## Proposal
+
+
+### When a bridge should set Fewer Features
+
+Fewer Features is a request from the bridge for the client to not use a certain feature. This means Fewer Features is ideal when:
+
+* the external platform does not support the feature
+* or the bridge does not handle the feature in a satisfactory way
+
+However, Fewer Features may not be ideal when:
+
+* the external platform supports the feature and the bridge is able to bridge it
+* the external platform doesn't support the feature, but the bridge does special converting to make it work
+
+### State event
+
+A bridge that wants to set fewer features can send a state event like so:
+
+```json
+{
+    "content": {
+        "send": {
+            "FEATURE_DECLARATION_KEY": DESIRABILITY_LEVEL,
+        },
+        "send_default": DESIRABILITY_LEVEL,
+        "receive": {
+            "FEATURE_DECLARATION_KEY": DESIRABILITY_LEVEL,
+        },
+        "receive_from_users": [
+            "@_theplatform_*:example.net"
+        ],
+        "send_reaction": {
+            "reaction_list": [
+                EMOJI,
+                ANOTHER_EMOJI
+            ],
+            "reactions_per_user_per_event": NUMBER,
+            "reactions_per_event": NUMBER
+        }
+    },
+    "type": "m.room.event_features",
+    "state_key": "@my_cool_bridge:example.net"
+}
+```
+
+The type shall be `m.room.event_features` and the state key shall be the user ID that sends the event (the bridge bot's user ID).
+
+Everything in `content` is optional. An empty dictionary is considered the same as an unset dictionary. An empty array is considered different to an unset array.
+
+### Desirability levels
+
+* `0`: Default value, same as it being unset. The bridge understands and will send/convert/substitute the event for the destination platform.
+* `-1`: This feature will be worse over the bridge. The client can choose to *Discourage, Substitute, or Disallow* the feature.
+* `-2`: This feature will not work over the bridge, and the bridge will fall back to the `body` or `formatted_body` instead. The client can choose to *Substitute or Disallow* the feature.
+* `-3`: This feature will not work over the bridge, and fallbacks aren't a good enough substitute. The client must *Disallow* the feature.
+
+### Discourage, Substitute, Disallow
+
+* *Discourage* means: Warn the user that the feature may not work well over the bridge, but allow them to use it anyway (e.g. by displaying a notice while the user is using the feature).
+* *Substitute* means: The bridge isn't able to provide a suitable substitute for this event, but the sending client could. The client should leave the feature as-is AND also provide a fallback `body` or `formatted_body` that is suitably close to parity with the original intention. The bridge will use the substitution provided by the client.
+* *Disallow* means: Disable the feature entirely (e.g. by removing UI elements that would activate it, or by using different preferences while in the room).
+
+It is the client's choice whether their Substitution is "suitably close to parity" to be able to use it. This should be considered on a feature-by-feature and client-by-client basis.
+
+### Feature declarations - sending
+
+When preparing or sending events, clients must keep these feature declarations in mind. Clients receiving events that use these features do not have to change their behaviour.
+
+#### `reply`
+
+This affects the [rich replies](https://spec.matrix.org/v1.9/client-server-api/#rich-replies) section of the spec.
+
+This means the external platform doesn't support replies. If the feature is Substituted, the bridge will send through the `formatted_body` reply fallback instead. If the feature is Disallowed, the reply fallbacks aren't good enough, and clients shouldn't send replies at all.
+
+#### `replace`
+
+This affects the [event replacements](https://spec.matrix.org/v1.9/client-server-api/#event-replacements) section of the spec.
+
+This means the external platform doesn't support edits. It is unlikely that clients can generate a suitable fallback for edits, as the original event will remain unchanged in chat, and it will be difficult for external users to see the difference unless the client's `formatted_body` fallback uses a diffing algorithm between old and new (the spec does not recommend this). If the feature is Substituted, clients might choose to disable it unless they have really good `formatted_body`.
+
+#### `redact`
+
+This affects sending a redaction according to [section 7.9.2 of the spec](https://spec.matrix.org/v1.9/client-server-api/#redactions).
+
+This means the external platform doesn't support redacting events.
+
+To comply with personal data laws, clients MUST still allow the user to redact the event on Matrix-side, but must inform the user it will still be visible on the external platform.
+
+#### `matrix.to_room`
+
+This affects the [matrix.to navigation](https://spec.matrix.org/v1.9/appendices/#matrixto-navigation) section of the spec.
+
+This means the external platform doesn't support links or references to rooms, so the bridge won't be able to convert the `matrix.to` links to a suitable representation.
+
+If the feature is Substituted, clients might be able to display a room name instead of a link. If the feature is Disallowed, clients shouldn't send `matrix.to` room links.
+
+#### `matrix.to_user`
+
+This means the external platform doesn't support links or references to users, so the bridge won't be able to convert it.
+
+If the feature is Substituted, clients might be able to display a user name instead of a link. If the feature is Disallowed, clients shouldn't send `matrix.to` user links.
+
+#### `matrix.to_event`
+
+This means the external platform doesn't support links or references to specific events, so the bridge won't be able to convert it.
+
+It is unlikely that clients can substitute this feature. If the feature is Disallowed, clients shouldn't send `matrix.to` event links.
+
+#### `rich_text`
+
+This affects the [event content HTML](https://spec.matrix.org/v1.9/client-server-api/#mroommessage-msgtypes) section of the spec.
+
+This means the external platform doesn't support any kind of rich text, and all messages will display as plain text. This will also affect the `<img>` HTML element.
+
+If the feature is Substituted, the bridge will send through the `body`. If the feature is Disallowed, clients should disable rich text controls and should not convert typed Markdown or HTML to rich text.
+
+#### `line_break`
+
+This means messages on the external platform cannot contain line breaks.
+
+#### `file`
+
+This means the external platform doesn't support files. The bridge may be able to send through a hyperlink to the file. If the feature is Disallowed, clients should not upload files. This will also affect the `<img>` HTML element.
+
+#### `voip`
+
+This affects the [Voice over IP](https://spec.matrix.org/v1.9/client-server-api/#voice-over-ip) section of the spec.
+
+This means the external platform doesn't support calls or the bridge is not capable of negotiating a call.
+
+Specific codecs of VoIP (e.g. if a platform supports voice but not video) are managed by the VoIP module. They are not managed by this module.
+
+#### `reaction`
+
+This affects the [event annotations and reactions](https://spec.matrix.org/v1.9/client-server-api/#event-annotations-and-reactions) section of the spec.
+
+This means the external platform doesn't support reactions. It is unlikely that clients can Substitute this feature. If the event is Disallowed, clients shouldn't send reactions.
+
+#### `sticker`
+
+This affects the [sticker messages](https://spec.matrix.org/v1.9/client-server-api/#sticker-messages) section of the spec.
+
+This means the external platform doesn't support stickers. Clients might choose to Substitute with an image instead, if `file` is available.
+
+#### `thread`
+
+This affects the [threading](https://spec.matrix.org/v1.9/client-server-api/#threading) section of the spec.
+
+This means the external platform doesn't support threads. Note that as per the spec, thread fallbacks look like rich replies. Discourage will suggest Matrix users not to create threads (if they do, they'll appear as rich replies). Substitute will always appear as rich replies. Disallow should remove UI elements that start new threads.
+
+### Feature declarations - sending reactions
+
+Unfortunately, other platforms often have special behaviour around reactions, which needs to be declared separately. Reaction behaviour is defined under the `send_reaction` key.
+
+Fallbacks are not available for these keys. Once the defined limit of a key is hit, the feature should be disallowed.
+
+#### `reaction_list`
+
+This is a list of all reactions that can be added to a message. This will probably be a list of emojis, but can also include text. (If the key is not specified, Matrix users can react with any text.) The special value `mxc://*` means any MXC URL can be used as a reaction.
+
+#### `reactions_per_event`
+
+This is the number of reactions that each Matrix user can add to each event. (If the key is not specified, there is no limit.)
+
+Once the limit is reached for a user on an event, clients must either:
+
+* disallow further reactions
+* OR automatically remove the existing reaction, then add the newly chosen reaction
+
+### Feature declarations - receiving
+
+The received events declaration only supports desirability levels 0 and -3.
+
+Clients receiving events that use a Disallowed feature should hide the feature.
+
+This only affects events received from the users in `receive_from_users`. `receive_from_users` should reference the simulated users managed by the bridge. Each string in `receive_from_users` may contain at most one asterisk `*` which acts as a wildcard.
+
+#### `read_receipt`
+
+This affects the [read receipts](https://spec.matrix.org/v1.9/client-server-api/#receipts) section of the spec.
+
+This means the external platform doesn't support read receipts. As such, the read receipts from external users will always match the messages they send (since sending a message automatically updates the read marker). Clients should not display these read receipts because they are misleading.
+
+### Extending to future features
+
+Features added to Matrix in the future will need their own feature declaration keys. There are two ways to add new keys to Fewer Features:
+
+1. In the MSC or spec for a new feature, the proposal can state what key to use.
+2. If any feature in the spec does not have a key, a new MSC can be created to simply propose adding a key for it.
+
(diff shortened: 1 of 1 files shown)


proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+The type shall be `m.room.event_features` and the state key shall be the user ID that sends the event (the bridge bot's user ID).
+
+Everything in `content` is optional. An empty dictionary is considered the same as an unset dictionary. An empty array is considered different to an unset array.
+
+### Desirability levels
+
+* `0`: Default value, same as it being unset. The bridge understands and will send/convert/substitute the event for the destination platform.
+* `-1`: This feature will be worse over the bridge. The client can choose to *Discourage, Substitute, or Disallow* the feature.
+* `-2`: This feature will not work over the bridge, and the bridge will fall back to the `body` or `formatted_body` instead. The client can choose to *Substitute or Disallow* the feature.
+* `-3`: This feature will not work over the bridge, and fallbacks aren't a good enough substitute. The client must *Disallow* the feature.
```
cloudrac3r, line 63: Looking for feedback: Do these make sense? Do they cover all use cases? Does it make sense for them to be ordered like this?

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+This means the external platform doesn't support read receipts. As such, the read receipts from external users will always match the messages they send (since sending a message automatically updates the read marker). Clients should not display these read receipts because they are misleading.
+
+### Extending to future features
+
+Features added to Matrix in the future will need their own feature declaration keys. There are two ways to add new keys to Fewer Features:
+
+1. In the MSC or spec for a new feature, the proposal can state what key to use.
+2. If any feature in the spec does not have a key, a new MSC can be created to simply propose adding a key for it.
+
+In addition, any MSCs that are not merged can use their `org.matrix.mscxxxx` number as the feature key. For example, the feature declaration key for [bot indicators][MSC4015] can be `org.matrix.msc4015`. This could be used to inform bots that this bridge will not bridge their bot markers to the external platform.
```
cloudrac3r, line 199: Is this helpful, or is it not necessary?

jooooscha:
I would like to see something lile this. But while reading this, I was wondering if this could also be useful for client to client communication. For example, a client could communicate that it does not support threads, or location sharing, or any other client feature.

proposals/4110-fewer-features.md
cloudrac3r, line 1: @jooooscha To keep discussions organised and to know when they're each resolved, we tend to keep them on the diff. I'm moving your comment to the diff so I can respond to it:

> I would like to see something lile this. But while reading this, I was wondering if this could also be useful for client to client communication. For example, a client could communicate that it does not support threads, or location sharing, or any other client feature.

That's an interesting idea. It might be good for less advanced clients to declare Fewer Features in a 1-1 chat, so the other person can avoid using incompatible features. And the trusted_private_chat preset for 1-1 chats allows both people to set any state event, so anybody could add Fewer Features to the room. But I don't think this would scale to group chats. If a single client could disable rich messaging for everyone in the room, that would be a pretty bad experience for everyone else, so we probably shouldn't allow that.

Thoughts?

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+* the external platform supports the feature and the bridge is able to bridge it
+* the external platform doesn't support the feature, but the bridge does special converting to make it work
+
+### State event
+
+A bridge that wants to set fewer features can send a state event like so:
+
+```json
+{
+    "content": {
```
tezlm, line 29: maybe instead of having the features be an object, it could be a flat hierarchical key~value map, ie.

```jsonc
{
	"reply": -1,
	"replace": -1,
	"formatting.bold": -1,
	"formatting.link": -1,
	"formatting.link.user": -1,
	"formatting.link.event": -1,
	"file": -1,
	"file.image": -1,
    // etc...
}
```

Then, `formatting.link` can be set to discourage any links and `formatting.link.event` could be set to discourage event links specifically. This would also be good for backwards compatibility, since new link type would still be disallowed if `formatting.link` was set.

cloudrac3r: @tezlm As mentioned in "Differences to Poorer Features", I am deliberately limiting entire features rather than individual types, keys, and elements. So I am deliberately not including a way to disable bold formatting.

There is no use case for this because there are no platforms that allow/disallow bold formatting specifically. Even if there was one, the bridge should be able to deal with it, rather than offloading the work onto the client.

If there is a use case then I'm happy to reconsider making it a fine-grained features list like you suggested, but in this proposal I'm currently going with a coarse-grained design because it's the simple option that would be easier for clients to implement.

proposals/4110-fewer-features.md
cloudrac3r: Yep, I think this is helpful. One good use for it is things in the flagship client that are very visible to many users (= high likelihood of being sent by users who expect it to work) but which have an unstable prefix and are still going through the spec. Such as the polls feature, which was in Element for a while but only recently MSC merged.

proposals/4110-fewer-features.md
jooooscha: I agree in general. But maybe, the less advanced client could also send Fewer Features to group chats and the more advanced clients decide what to do with it. I am thinking about a more advanced client trying to create a poll, but only one or perhaps none of the other ten group participants can handle polls. It could still be helpful to know that for the creator (client) of the poll.

Not sure if this is in the scope of Fewer Features.

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Discourage, Substitute, Disallow
+
+* *Discourage* means: Warn the user that the feature may not work well over the bridge, but allow them to use it anyway (e.g. by displaying a notice while the user is using the feature).
+* *Substitute* means: The bridge isn't able to provide a suitable substitute for this event, but the sending client could. The client should leave the feature as-is AND also provide a fallback `body` or `formatted_body` that is suitably close to parity with the original intention. The bridge will use the substitution provided by the client.
+* *Disallow* means: Disable the feature entirely (e.g. by removing UI elements that would activate it, or by using different preferences while in the room).
+
+It is the client's choice whether their Substitution is "suitably close to parity" to be able to use it. This should be considered on a feature-by-feature and client-by-client basis.
+
+### Feature declarations - sending
```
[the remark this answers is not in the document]
cloudrac3r: Good idea, we should add this. Though we'd have to consider that messages might become longer after being converted by the bridge. So even if the client complies with the stated length, the bridge might convert it to something longer and then still have to do some extra work to make the transformed message fit within the limit. (For example, this might happen if the bridged platform's format for user mentions is longer than Matrix's format.)

There's also a question of whether we count the length from `body` or `formatted_body`, and if `formatted_body`, whether we count HTML elements as length. This gets even more complicated for extensible events.

Do you have any ideas how to make this work?

All I can think of is saying the maximum length is a "best guess" and bridges should still be able to handle the situation where it goes over the limit? If this stops Matrix users sending textwalls to IRC 90% of the time, that would at least be an improvement...

proposals/4110-fewer-features.md
cloudrac3r: That's true, it would be cute if when starting a poll in a group chat it could show a notice like "5 of the 10 people in this room can respond to polls". It could be like a larger scale version of a Discouraged feature.

However, I have no idea how to implement this. The Fewer Features proposal relies on state events but regular users can't set state arbitrary state events.

sumnerevans: Another complicating factor here is that users may use multiple clients with different capabilities. I normally run Element Desktop, Gomuks, and SchildiChat, all of which have different sets of supported features. It's unclear how I would indicate which of these features I as a user support.

proposals/4110-fewer-features.md
progval: Most IRC servers allow channel operators to disallow (or strip) formatting.

proposals/4110-fewer-features.md
progval: In MSC3968 I supported more levels in order to allow expressing "X will be bad over the bridge, but less bad than Y"

proposals/4110-fewer-features.md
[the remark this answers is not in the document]
progval: I gave this a shot in MSC3969.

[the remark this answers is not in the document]
clokep: > So even if the client complies with the stated length, the bridge might convert it to something longer and then still have to do some extra work to make the transformed message fit within the limit.

That's a good point, I don't have a solution for that unfortunately. Maybe it isn't so useful and bridges just need to handle it.

proposals/4110-fewer-features.md
tulir: > There is no use case for this because there are no platforms that allow/disallow bold formatting specifically. Even if there was one, the bridge should be able to deal with it, rather than offloading the work onto the client.

There are lots of platforms with less formatting features than Matrix even if bold is usually supported. For example:

* Most networks don't support inline images, collapsible blocks (`<details>`), tables and font colors
* WhatsApp doesn't support spoilers
* Signal doesn't support blockquotes and lists
* Twitter supports user mentions but no other formatting

Bridges can deal with them to some extent, but obviously won't be able to perfectly bridge features that aren't supported. I think there should be a way to tell clients about unsupported features. Some clients may want that info even if other clients won't customize input format based on bridge. I could very well imagine using the [goldmark](https://github.com/yuin/goldmark) markdown parser in Go and disabling specific features per room.

proposals/4110-fewer-features.md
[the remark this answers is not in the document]
progval: What bridges would convert it to something longer? The IRC bridge always makes it shorter, and most other bridges probably just keep the HTML.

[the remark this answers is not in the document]
cloudrac3r: Sure, let's give it a try. @progval do you like the approach you took in MSC3969 (and do you want to continue it?) or do you think we should do it differently in Fewer Features?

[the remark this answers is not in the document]
progval: I haven't tried implementing it, so your guess is as good as mine

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Feature declarations - sending reactions
+
+Unfortunately, other platforms often have special behaviour around reactions, which needs to be declared separately. Reaction behaviour is defined under the `send_reaction` key.
+
+Fallbacks are not available for these keys. Once the defined limit of a key is hit, the feature should be disallowed.
+
+#### `reaction_list`
+
+This is a list of all reactions that can be added to a message. This will probably be a list of emojis, but can also include text. (If the key is not specified, Matrix users can react with any text.) The special value `mxc://*` means any MXC URL can be used as a reaction.
```
tulir, line 167: I wonder if it'd make more sense to have `send` values be objects so extra info like reaction limits, file size limits and possibly text length limits could all be in there

```json
{
  "send": {
    "reaction": {
      "desirability": 0,
      "list": ["👍️", "👎️"]
    },
    "file": {
      "desirability": 0,
      "mimetypes": ["image/png", "image/jpeg"],
      "max_size": 102400
    },
    "text": {
      "desirability": 0,
      "max_length": 100
    },
    "sticker": {
      "desirability": 0
    }
  }
}
```

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+If the feature is Substituted, the bridge will send through the `body`. If the feature is Disallowed, clients should disable rich text controls and should not convert typed Markdown or HTML to rich text.
+
+#### `line_break`
+
+This means messages on the external platform cannot contain line breaks.
+
+#### `file`
+
+This means the external platform doesn't support files. The bridge may be able to send through a hyperlink to the file. If the feature is Disallowed, clients should not upload files. This will also affect the `<img>` HTML element.
```
tulir, line 131: This should probably also be more granular. For example, Instagram only allows images, videos and voice messages. WhatsApp allows anything when sending as document, but only specific mimetypes when sending as image/video/audio (https://developers.facebook.com/docs/whatsapp/on-premises/reference/media/). I think being able to set mimetype restrictions per msgtype would allow mapping everything properly.

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+This means the external platform doesn't support links or references to users, so the bridge won't be able to convert it.
+
+If the feature is Substituted, clients might be able to display a user name instead of a link. If the feature is Disallowed, clients shouldn't send `matrix.to` user links.
+
+#### `matrix.to_event`
+
+This means the external platform doesn't support links or references to specific events, so the bridge won't be able to convert it.
+
+It is unlikely that clients can substitute this feature. If the feature is Disallowed, clients shouldn't send `matrix.to` event links.
```
tulir, line 115: These should have more generic names rather than referencing matrix.to specifically because [`matrix:` URIs](https://spec.matrix.org/v1.9/appendices/#matrix-uri-scheme) also exist (and bridges supporting recent versions of the spec should support both). The spec calls them ["User and room mentions"](https://spec.matrix.org/v1.9/client-server-api/#user-and-room-mentions), although that could be confused with `@room` mentions 🤔

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
+# MSC4110: Fewer features
+
+The Fewer Features proposal aims to make Matrix bridges work better with other platforms. Currently, it is hard for clients and Matrix users to know if what they send in a bridged room will be supported or not. The goal of Fewer Features is to let bridges communicate which Matrix features are supported on the external platform. Then, clients and bots can change how they act based on what the external platform supports.
+
+In a nutshell, bridges can set a state event that says which Matrix features are discouraged or disallowed in this room. Clients and bots can use this information to disable features, change UI elements, or change the way events are sent. This will give the best possible experience to people on the other side of the bridge. It will also reassure Matrix users that their actions are working properly, because incompatible actions simply cannot be sent.
```
tulir, line 5: Lines should be wrapped to 120 characters at most

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+This is a list of all reactions that can be added to a message. This will probably be a list of emojis, but can also include text. (If the key is not specified, Matrix users can react with any text.) The special value `mxc://*` means any MXC URL can be used as a reaction.
+
+#### `reactions_per_event`
+
+This is the number of reactions that each Matrix user can add to each event. (If the key is not specified, there is no limit.)
+
+Once the limit is reached for a user on an event, clients must either:
+
+* disallow further reactions
+* OR automatically remove the existing reaction, then add the newly chosen reaction
```
tulir, line 176: I'm not completely sure if this is a good idea or not. There are several networks that only allow one reaction per message, but my bridges already handle that by automatically redacting the previous reaction. If the client were to do the redaction, then the bridge would bridge that removal over to the remote network first, instead of just making a single API call to replace the reaction.

tulir, line 176: Another thing this reminded me of: Telegram allows 3 reactions for premium users and 1 reaction for normal users. Some platforms also have per-user limits for things like file size and message length. I think it's fine not to solve such cases here, but should probably be noted in potential issues.

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+This means the external platform doesn't support edits. It is unlikely that clients can generate a suitable fallback for edits, as the original event will remain unchanged in chat, and it will be difficult for external users to see the difference unless the client's `formatted_body` fallback uses a diffing algorithm between old and new (the spec does not recommend this). If the feature is Substituted, clients might choose to disable it unless they have really good `formatted_body`.
+
+#### `redact`
+
+This affects sending a redaction according to [section 7.9.2 of the spec](https://spec.matrix.org/v1.9/client-server-api/#redactions).
+
+This means the external platform doesn't support redacting events.
+
+To comply with personal data laws, clients MUST still allow the user to redact the event on Matrix-side, but must inform the user it will still be visible on the external platform.
```
tulir, line 95: > clients MUST still allow the user to redact the event on Matrix-side

I'm not sure if the spec is allowed to have such strong wording about client UI, `SHOULD` would probably be more appropriate. Also, rooms can already prevent users from redacting messages by setting the `m.room.redaction` event power level to non-default value.

proposals/4110-fewer-features.md
tulir: It is helpful (and even necessary) to note that non-spec extensions can be added, but instead of talking about MSCs specifically, it should just define that all feature identifiers must follow the [Common Namespaced Identifier Grammar](https://spec.matrix.org/v1.9/appendices/#common-namespaced-identifier-grammar). That spec already defines that identifiers starting with `m.` are reserved for specced features, and that custom identifiers following java package naming style are allowed.

There's also no requirement for unstable prefixes in MSCs to use the `org.matrix.mscxxxx` format even though it's fairly common, any namespaced identifier is a valid unstable prefix.

proposals/4110-fewer-features.md
cloudrac3r: Fair enough, I don't like the name either. I originally picked it because [the spec appendix](https://spec.matrix.org/v1.9/appendices/#matrixto-navigation) called it out by name.

If I'm going with the hierarchical values from the suggestion above (`formatting.bold`, `formatting.link` etc) I'll put them under `formatting.link.room` and `formatting.link.user` which I think works pretty well. After all, the matrix.to room links really behave like room _links_ and not mentions.

proposals/4110-fewer-features.md
cloudrac3r: > Also, rooms can already prevent users from redacting messages by setting the `m.room.redaction` event power level to non-default value.

Rooms can prevent users redacting their _own_ events?? I can't believe Matrix allows that. I'll update my proposal to not have that assumption.

proposals/4110-fewer-features.md
cloudrac3r: Do you mean I should change the feature keys to all start with `m.`?

proposals/4110-fewer-features.md
cloudrac3r: You're right, it's not realistic to ask human users to manually configure which events they would like to receive. It also (probably?) can't be done automatically by the clients because we can't distinguish between actively used clients and inactively used clients, unless we did some extreme to-device-messaging, which seems out of scope for Fewer Features.

proposals/4110-fewer-features.md
cloudrac3r: > There are several networks that only allow one reaction per message, but my bridges already handle that by automatically redacting the previous reaction.

That's a smart idea for handling it, nice!

> I'm not completely sure if this is a good idea or not.

As far as I'm aware, our choices are:

1. Assume all bridge developers thought of this idea and implemented it in their bridge, and so remove `reactions_per_event` from the proposal because it isn't needed
2. Assume some bridge developers didn't think of this idea or weren't able to implement it, and keep `reactions_per_event` to be used by bridges that can't handle this situation themselves
3. Keep `reactions_per_event` anyway so that some clients can provide graphical feedback

Do you have a favourite choice?

tulir: If there was a way to say "limited to X reactions, but the bridge will redact the previous one automatically" it would accurately represent what happens now, but I'm not sure if that's actually useful to clients. I guess clients could have a "local echo" type thing where they hide the previous reaction locally immediately rather than waiting for the redaction? That would provide the best UX in the happy path, but would cause weirdness if the bridge fails to do the redaction for whatever reason.

Other than that, might be fine to keep it around even if it's not used

proposals/4110-fewer-features.md
tulir: Oh right, they currently don't start with that 🤔

Yeah, I think they should be changed, having `m.` on standard keys makes it clearer that custom keys are allowed too.

proposals/4110-fewer-features.md
```diff
@@ -0,0 +1,383 @@
...
+
+This affects the [rich replies](https://spec.matrix.org/v1.9/client-server-api/#rich-replies) section of the spec.
+
+This means the external platform doesn't support replies. If the feature is Substituted, the bridge will send through the `formatted_body` reply fallback instead. If the feature is Disallowed, the reply fallbacks aren't good enough, and clients shouldn't send replies at all.
+
+#### `replace`
+
+This affects the [event replacements](https://spec.matrix.org/v1.9/client-server-api/#event-replacements) section of the spec.
+
+This means the external platform doesn't support edits. It is unlikely that clients can generate a suitable fallback for edits, as the original event will remain unchanged in chat, and it will be difficult for external users to see the difference unless the client's `formatted_body` fallback uses a diffing algorithm between old and new (the spec does not recommend this). If the feature is Substituted, clients might choose to disable it unless they have really good `formatted_body`.
```
tulir, line 87: Just remembered another type of limit that many other networks have: maximum number of edits and maximum time after sending that message can be edited. Similarly, redactions also have a time limit on some networks.

proposals/4110-fewer-features.md
tulir, line 1: Beeper is planning on using a different format, which is documented at https://github.com/mautrix/go/blob/main/event/capabilities.d.ts

Notable differences:
* Each file message type is handled separately, with allowed mime types listed
* Formatted text is split into features to allow signaling exactly what's supported
* Levels range from -2 to 2 instead of -3 to 0

Johennes: This is outside the scope of this proposal but maybe adhoc capability checks via normal room events could be good enough? If you want to know if it makes sense to send a poll in a room, your client could send some `m.features.query` request. Active devices of other users would then respond with `m.features.answer` to indicate their level of support. You won't get answers from everyone in the room but at least you get some level of assurance about how many room members run a client that properly supports polls.

Kladki, line 1: Can I request the "application services" label be added to this MSC? I struggled to find it because it was missing this label.

PR #4113, opened 2024-02-24
MSC4113: Image hashes in Policy Lists

MTRNord:
[Rendered](https://github.com/MTRNord/matrix-doc/blob/patch-5/proposals/4113-media-hashes-in-policy-lists.md)

Potential COIs:

- I do volunteer work for Draupnir
- I volunteer at the Matrix foundation for the website
- I do work for Nordeck

Signed-off-by: Marcel Radzio <support@nordgedanken.dev>

### DIFF
--- a/proposals/4113-media-hashes-in-policy-lists.md
+++ b/proposals/4113-media-hashes-in-policy-lists.md
@@ -13,19 +13,24 @@ The MSC recommends that the hash itself is used for the state_key as it
 should be sufficiently unique.
 
 In the content of the event, there are expected to be hash implementation-specific 
-values. Therefore the event has a mandatory `type` key which differentiates the 
+values. Therefore the event has a mandatory typekey which differentiates the 
 various types of hashes. Each of the types MUST follow the in matrix common java 
 style namespace format. For example, a pdqhash type would look like `m.pdqhash` 
 where `m` is your namespace and `pdqhash` is the hash type.
+
+An object approach is choosen to make sure that in the future the events can be 
+easily upgraded while staying backwards compatible with old implementations.
+For example when there are issues with hash algos discovered in the future.
 
 Such an event in full would look like this:
 
 ```json
 {
   "content": {
-    "type": "m.pdqhash",
-    "hash": "d8f8f0cce0f4a84f0e370a22028f67f0b36e2ed596623e1d33e6b39c4e9c9b22",
-    "quality": "100"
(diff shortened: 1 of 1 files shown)


proposals/4113-media-hashes-in-policy-lists.md
```diff
@@ -0,0 +1,83 @@
...
+
+This means that for the hash type `m.pdqhash` the content MUST include at 
+least these 2 fields: `hash` and `quality`.
+
+## Potential issues
+Since there might be multiple implementations there might be multiple hash types 
+being used. There is currently no easy way to have a list of used type strings 
+across matrix.
+
+An action is hard to define and needs to be discussed as part of the MSC process.
```
MTRNord, line 58: As mentioned in the MSC I am not sure how to best define actions in the MSC. Suggestions are welcome

TheArcaneBrony:
Personal observations and opinions (based on original version):

1) use a dictionary format for types, so multiple can be specified in a single event (eg. i want to support md4, md5, sha128, sha256, sha384, sha1, sha3, ... file hashes)
 - this also allows you to implement newer formats while keeping backwards compatibility, *without* having to send 2 million events per policy (/h)
2) the proposed format only works for images, so calling it "media" is misleading (is an audio file not media? how about a word document or a zip file?)
3) Perhaps a policy tags list may be useful? This would allow bots to decide an action based on that, eg. csam would be a ban but lewd imagery may not be, depending on the room it's posted in

proposals/4113-media-hashes-in-policy-lists.md
```diff
@@ -0,0 +1,99 @@
+# MSC4113: Image hashes in Policy Lists
+Currently Policy lists are mainly focused on users, rooms and servers and not content. 
+This proposal adds a fourth kind: Media. Especially this focuses on image media 
+which can be hashed and compared using 
+[Perceptual hashes](https://en.wikipedia.org/wiki/Perceptual_hashing).
+
+The use for these are crowd-gathered hashlists for potentially bad media which 
+allow people to then scan their media with and also block dangerous media.
```
turt2live, line 8: This also, unfortunately, creates a list of media that those bad actors will *want* on their servers. This is what makes this sort of feature difficult, and illegal in many jurisdictions: listing media which *may* contain CSAM, terrorism, etc is often as illegal as distributing the content itself.

PR #4114, opened 2024-02-24
MSCXXXX: Matrix as a password manager

Johennes:


proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,174 @@
...
+space to securely and conveniently store and share secrets. This proposal
+outlines a scheme for storing a hierarchy of secrets in Matrix by borrowing
+from standard concepts such as rooms and spaces.
+
+## Proposal
+
+### Secret hierarchy
+
+Two new room types `m.vault` and `m.vault.secret` are introduced. Vault-rooms
+work similar to [spaces] and can group other vault-rooms or secret-rooms.
```
FSG-Cat, line 13: So why isnt this just spaces? I mean the secrets room ok that makes some sense but why m.vault isnt just a space i do not think the proposal properly addresses. I mean its not even an addressed alternative.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,174 @@
...
+discouraged - clients are not generally expected to have a way to render the
+timeline of these rooms. As such, vault- and secret-rooms should be created
+with `m.room.power_levels` which prohibit normal events by setting
+`events_default` to a suitably high number. In the default power level
+structure, this would be 100.
+
+Additionally, vault- and secret-rooms should be created with a join rule of
+`invite` to prevent unintended access without explicit sharing.
+
+To include a secret (or another vault) in a vault, an `m.vault.child` state
```
MTRNord, line 26: Why does the MSC introduce another type of spaces here instead of using spaces? That seems like a weird thing to do. We would end up with major duplication of existing logic around spaces.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,174 @@
...
+
+No rooms other than those of type `m.vault` and `m.vault.secret` are allowed to
+be stored in `m.vault.child` events.
+
+Unlike with spaces, there is no corresponding `m.vault.parent` event, meaning a
+vault or secret does not know which parent vaults contain it. While in spaces
+this backlink exists to aid discoverability, this feature appears unessential
+for secret hierarchies.
+
+### Storing secrets
```
MTRNord, line 53: I think in modern days, it should be MANDATORY for these to be encrypted. Storing passwords unencrypted seems like an issue that's intentionally chosen. Afaik all major security advisories require as part of their suggestions for passwords to be securely stored by either hashing or encrypting them. I don't see why we are not doing that here.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,174 @@
...
+
+The standard `m.room.name` and `m.room.avatar` state events can be used to label
+and decorate vaults and secrets. These are not currently encryptable but will be
+once [MSC3414] lands. While exposing vault and secret names is not considered a
+security concern by other password managers such as [pass], it can still be a
+privacy concern. Therefore, clients should warn users appropriately in the meantime.
+
+## Potential issues
+
+When sharing _and_ federating, secrets can end up being stored on different home
```
MTRNord, line 131: In companies, this is very much an usecase. Bitwarden for example provides this as a feature. Imho we should specify this if we add this.

MTRNord (commented):
I see some very major flaws with this around the security of the secrets. This feels like a excel spreadsheet with metadata added :/ Thats against any modern security recommendation.

proposals/4114-matrix-as-a-password-manager.md
Johennes: You're right that it's currently missing from the alternatives, thanks.

As for the reasoning, I wanted to prevent mixing secret-rooms with other rooms in the same vault / space. Though maybe that's not actually required and we could just reuse spaces. 🤔

proposals/4114-matrix-as-a-password-manager.md
Johennes: It already says below:

> `m.secret` events are not meant to be used in rooms other than those of type
> `m.vault.secret` and should always be encrypted.

Is that not sufficient?

We can't store the secrets as hashes because in a password manager you want to recover the actual secret, not just verify it against user input.

proposals/4114-matrix-as-a-password-manager.md
Johennes: Interesting, I wasn't aware. Could you link to where they describe their federation support? I'm curious how it works but couldn't really find anything around multi-instance in their docs.

MTRNord: Will take a look. It might be not true fed as its multiple orgs within the bitwarden cloud Service.

But even apart from that In matrix at least in my company people prefer using their private servers for internal usage or some dedicated Account. So it might be that employees are not on the same homeserver.

proposals/4114-matrix-as-a-password-manager.md
Johennes: No problem. It's maybe also a bit too hidden down there. I tried to clarify it further by adding a remark that both room types should be created with encryption enabled in the section above.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,180 @@
...
+holds an ordered array of section definitions with the following properties:
+
+- `title` – A textual description (optional)
+- `fields` – An ordered array of field definitions (required)
+
+Fields in turn have the following properties:
+
+- `title` – A textual description (optional)
+- `type` – An identifier for the type of value stored (optional). One of:
+  - `text` – Any text. If `type` is omitted, this is the default.
```
HarHarLinks, line 67: I think it's essential to have a custom field which you add here as `text`. What I am missing is a custom field that also indicates that its content is a secret, such as... a matrix SSSS recovery key, that should not be displayed by default.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,180 @@
...
+    }]
+  }
+}
+```
+
+Clients can choose to use the order of sections and fields in the event for
+sorting data in the UI but are not required to do so.
+
+`m.secret` events are not meant to be used in rooms other than those of type
+`m.vault.secret` and should always be encrypted.
```
HarHarLinks, line 109: Maybe I'm just scarred from the last 7 or so years that we had regular UTDs, but I'm still scared to lose my passwords this way, even with SSSS. MSC4114-based password managers should really offer an offline encrypted backup feature, but I'm not certain the spec may see it in it's area of responsibility.

proposals/4114-matrix-as-a-password-manager.md
Johennes: Oh, good call! This makes me wonder if there should just be a `conceal: true|false` property available on every field type? It could default to `true` for `type: password` and `false` otherwise.

proposals/4114-matrix-as-a-password-manager.md
Johennes: Yeah, agreed. This sounds like a great point to add to the "Potential issues" section.

HarHarLinks: I like this proposition for the concept of sharing for passwords this might enable, which seems much nicer than what my experience has been with bitwarden and keepass2. However it is ironic that matrix with it's required secondary password which will even be needed to access the encrypted passwords here, basically requires another password manager. It's a bit of a hen-and-egg problem.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,180 @@
...
+should be set so that being invited into a vault-room enables users to also
+join all of its child-rooms.
+
+The standard `m.room.name` and `m.room.avatar` state events can be used to label
+and decorate vaults and secrets. These are not currently encryptable but will be
+once [MSC3414] lands. While exposing vault and secret names is not considered a
+security concern by other password managers such as [pass], it can still be a
+privacy concern. Therefore, clients should warn users appropriately in the meantime.
+
+## Potential issues
```
HarHarLinks, line 130: Usually, password managers require some (softer) re-authentication to access passwords after some timeout, even though they remain signed into e.g. some cloud service, as a precaution. Implementing the same would... introduce a third password (after Login, SSSS) to matrix. I'm not sure how to handle this. This might actually be somewhat smart to have as a requirement so you block other clients from accessing your vault-secrets, as these clients might not have been implemented with the same security precautions in mind as your matrix-passwordmanager.

Johennes: Could the re-auth be forced via [soft logout](https://spec.matrix.org/latest/client-server-api/#soft-logout)? When there is no refresh token, it already requires going through `/login` again.

> If it does not have a refresh token available, or refreshing fails with `soft_logout: true`, the client can acquire a new access token by specifying the device ID it is already using to the login API.

It might also be handy to be able to trigger a soft logout from the client to lock the device.

proposals/4114-matrix-as-a-password-manager.md
Johennes: Have made it so.

proposals/4114-matrix-as-a-password-manager.md
HarHarLinks: Sounds technically feasible, but it does nothing to address the second half of my comment about securing vault secrets even better than regular messages and potentially rather distrusting "regular" clients that don't specifically address the vault use case. Do I understand correctly the context of your concept is a "one matrix account for all your needs" as usual, adding the password manager use case?

Johennes: For a start, I was actually thinking about separating the two use cases and using a dedicated home server for passwords. This is the approach I'm trying to prototype in https://github.com/Johennes/vaultmeister with a local dendrite instance.

My general thinking was that if it's secure enough to handle all your private communication, it's secure enough to also store your passwords. Therefore, I hadn't considered any additional authentication layers to guard access to vaults and passwords so far. However, I think I see your point now. If we were to combine messaging and passwords in a single client, you wouldn't want your passwords to be unlocked and readable just because you're chatting.

I'm not sure how to best address this right now. 🤔

If we wanted the server to enforce authentication, we'd need to exclude vault- and secret-rooms from `/sync` & friends and create separate APIs. That feels like a quite invasive endeavor.

Alternatively, maybe we could change the encryption scheme for events in vault- and password rooms? Most trivially, for instance, we could apply an additional symmetric encryption using some master key. Clients that are not password-capable wouldn't know how to perform the decryption and couldn't accidentally expose your passwords.

HarHarLinks: I see your idea and hate having had to bust the bubble.

The latter alternative is what I had in mind at first. At that point, could we not lose the regular e2ee, including it's drawbacks?

However:
- What properties are required from this symmetric encryption? Is there another standard this should be based on rather than reinventing the wheel?
- How does sharing pws work with the introduction of a symmetric key?

Johennes: > The latter alternative is what I had in mind at first. At that point, could we not lose the regular e2ee, including it's drawbacks?

Yeah, I guess not much point in double-encrypting.

> * What properties are required from this symmetric encryption? Is there another standard this should be based on rather than reinventing the wheel?
> * How does sharing pws work with the introduction of a symmetric key?

I think the symmetric key would have to be shared among users using asymmetric encryption. This could probably be done via to-device messages. That would be akin to what Bitwarden has detailed [here](https://bitwarden.com/help/bitwarden-security-white-paper/#sharing-data-between-users). They don't mention what happens when somebody leaves an organization. I guess that would require rotating the symmetric key and re-encrypting all vaults.

What's really unfortunate about this is that we'd be rebuilding the access and sharing mechanisms that already exist in Matrix. :/

andybalaam: > What's really unfortunate about this is that we'd be rebuilding the access and sharing mechanisms that already exist in Matrix. :/

Yeah, I'm not sure there's much point in this proposal unless it re-uses the auth, access, and encryption of Matrix.

So a separate Matrix user for passwords seems like the most practical way to use this.

proposals/4114-matrix-as-a-password-manager.md
```diff
@@ -0,0 +1,186 @@
...
+    }]
+  }
+}
+```
+
+Clients can choose to use the order of sections and fields in the event for
+sorting data in the UI but are not required to do so.
+
+`m.secret` events are not meant to be used in rooms other than those of type
+`m.vault.secret` and should always be encrypted.
```
HarHarLinks, line 111: This will basically require clients to implement key sharing with other users to be implemented so you can invite them after having created a secret.

proposals/4114-matrix-as-a-password-manager.md
HarHarLinks: I'm not the greatest expert on the low levels of encryption (in Matrix or otherwise), but I wonder, does Matrix e2ee messaging encryption meet the standards required of a state of the art password manager? I.e. how hard is it to crack a password? What considerations that are applicable to consider the messaging use case "secure", e.g. rotating keys, are (not) applicable to a password manager that is closer to write-once-read-often?

Johennes: I'm quite out of my depth here, too, but trying to compare with the [Bitwarden Whitepaper](https://bitwarden.com/help/bitwarden-security-white-paper/):

### Login

Bitwarden's client uses PBKDF2 with 600,000 iterations to derive a 256-bit master key from your account password and email. The account password and master key are then turned into a master password hash using PBKDF-SHA256. That hash is sent to the server where it is hashed again using PBKDF2-SHA256 with 600,000 iterations before verifying the login.

In Matrix, when using the `m.login.password` flow, the client simply sends the user name and password to the server. The spec doesn't appear to mandate the use of key derivation functions on either side. Synapse uses [bcrypt](https://github.com/element-hq/synapse/blob/develop/synapse/handlers/auth.py#L1642) with a default of ... wait for it ... [12 iterations](https://github.com/element-hq/synapse/blob/develop/synapse/config/registration.py#L79). However, the value can be configured by the home server admin.

With the other Matrix login flows or OIDC, I think, it would depend on what your SSO or identity provider uses. Keycloak, for instance, uses [PBKDF2-SHA512 with 210,000 iterations](https://www.keycloak.org/docs/latest/server_admin/#hashing-iterations) by default.

In summary, the notable difference here is that Matrix doesn't mandate any key derivation for the account password on the client side. I suppose a specialized Matrix Password Manager client could just send the derived hash instead of the password in the `m.login.password` flow to achieve the same level of protection. That would make the account unusable with other Matrix clients which may actually be desirable.

### Other aspects

To be researched... 😅

Johennes: ### Encryption of vault items / messages

Bitwarden uses cryptographically secure random data to generate a single 512-bit symmetric key (consisting of a 256-bit encryption key and a 256-bit MAC key). This key is used to en- and decrypt _all_ vault items using [AES-256 CBC](https://bitwarden.com/help/what-encryption-is-used).

To protect the key, it is encrypted with the stretched master key using AES-256 and a random 128-bit initialization vector. The encrypted symmetric key is synced across clients via the server.

The stretched master key is 512 bits and derived from the 256-bit master key through HKDF. Recall from my previous message that the master key itself is obtained from the account email and password via PBKDF2 with 600,000 iterations. If I understand this correctly, this process is designed so that brute-forcing the account password is hard because once you have the password, you can derive the stretched master key which would then allow you to decrypt the symmetric key giving access to all vault items.

In Matrix, on the other hand, we have [Megolm](https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md) which also uses AES-256 CBC to symmetrically encrypt cleartext but obtains the key differently. Megolm is session-based where each session uses a ratchet that is initialized with 1024-bit cryptographically secure random data. The ratchet is wound forward through hashing on each encrypted message and the symmetric encryption key is derived by hashing the ratchet value. Additionally Megaolm uses Ed25519 to provide message authenticity through signatures. 

Megolm keys are synced among clients via either key requests in to-device messages or the server-side key backup. The latter is encrypted using AES-256 CBC where the [256-bit curve25519 private key](https://spec.matrix.org/v1.8/client-server-api/#recovery-key) is commonly stored in 4S. The key that unlocks 4S itself can be derived from a passphrase using PBKDF2 where the number of iterations is [configurable](https://spec.matrix.org/v1.8/client-server-api/#deriving-keys-from-passphrases). This also explains why Matrix doesn't enforce any key stretching during login as the login password doesn't unlock the keys needed to decrypt messages.

In summary, the key difference here is that Bitwarden uses a single symmetric key to encrypt everything whereas Matrix uses per-message keys. Both upload encrypted keys to the server to sync them across clients and use PBKDF2 key stretching to make cracking the password that protects the key hard.

Johennes: I need to condense all of this back into the MSC but in total, I'm not currently seeing why passwords would be any less secure in Matrix than in Bitwarden.

Johennes: The points from this thread have now been included in the proposal.

proposals/4114-matrix-as-a-password-manager.md
Johennes: Resolving as I have now explicitly descoped federation from this proposal to limit its scope.

I'm not sure if maybe you'd like to rework your comment in the main thread, too? It seems quite far off. The proposal already doesn't mimic an excel spreadsheet because secrets are rooms. Plus, encryption is mandatory in both secret and vault rooms. I've included a comparison of cryptographic primitives between Matrix and Bitwarden in the proposal.

PR #4115, opened 2024-02-26
MSC4115: membership information on events

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/membership_on_events/proposals/4115-membership-on-events.md)

CoI declaration: I am employed by Element, and a member of the Matrix Spec Core Team (SCT). This proposal is written as part of my work within the Crypto team within Element.

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4115#issuecomment-2122194253)

### DIFF
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -125,6 +125,8 @@ None foreseen.
 
 ## Unstable prefix
 
+While this proposal is in development, the name `io.element.msc4115.membership`
+should be used in place of `membership`.
 
 ## Dependencies
 
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -104,7 +104,7 @@ Example event including the new property:
   "type": "m.room.member",
   "unsigned": {
     "age": 1567437,
-    "membership": "leave",
+    "membership": "join",
     "redacted_because": {
       "content": {
         "reason": "spam"
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -126,8 +126,7 @@ Example event including the new property:
 
 ## Potential issues
 
- * Depending on server implementation, it may be expensive or difficult to
-   implement? Feedback welcome from HS authors on this point.
+None foreseen.
 
 ## Alternatives
 
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -54,13 +54,13 @@ according to the state of the room at the time of the event being returned. If
 the user had no membership at that point (ie, they had yet to join or be
 invited), `membership` is set to `leave`.  Any changes caused by the event
 itself (ie, if the event itself is a `m.room.member` event for the requesting
-user) are *excluded*.
+user) are *included*.
 
 In other words: servers should follow the following algorithm when populating
 the `unsigned.membership` property on an event E and serving it to a user Alice:
 
-1. Consider the room state just *before* event E landed (accounting for state
-   resolution across E's `prev_events`, but not E itself).
+1. Consider the room state just *after* event E landed (accounting for E
+   itself, but not any other events in the DAG which are not ancestors of E).
 2. Within the state, find the event M with type `m.room.member` and `state_key`
    set to Alice's user ID.
 3. * If no such event exists, set `membership` to `leave`.
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -1,0 +1,131 @@
+# MSC4115: membership metadata on events
+
+## Background
+
+Consider the following Event DAG:
+
+```mermaid
+graph BT;
+    B[Bob joins];
+    B-->A;
+    C-->A;
+    D-->B;
+    D-->C;
+```
+
+Bob has joined a room, but at the same time, another user has sent a message
+`C`.
+
+Depending on the configuration of the room, Bob's server may serve the event
+`C` to Bob's client. However, if the room is encrypted, Bob will not be on the
+recipient list for `C` and the sender will not share the message key with Bob,
+even though, in an absolute time reference, `C` may have been sent at a later
+timestamp than Bob's join.
+
+Unfortunately, there is no way for Bob's client to reliably distinguish events
+such as `A` and `C` that were sent "before" he joined (and he should therefore
+not expect to decrypt) from those such as `D` that were sent later.
+
+This issue is discussed in more detail at
+https://github.com/element-hq/element-meta/issues/2268.
+
+As a partial solution to this problem, we propose a mechanism for servers to
+inform clients of room membership at each event.
+
+## Proposal
+
+The `unsigned` structure contains data added to an event by a homeserver when
+serving an event over the client-server API.  (See
+[specification](https://spec.matrix.org/v1.9/client-server-api/#definition-clientevent)).
+
+We propose adding a new property, `membership`, which should contain the
+membership of the user making the request, according to the state of the room
+at the time of the event being returned. If the user had no membership at that
+point (ie, they had yet to join or be invited), `membership` is set to `leave`.
+Any changes caused by the event itself (ie, if the event itself is a
+`m.room.member` event for the requesting user) are *excluded*.
+
+In other words: servers should follow the following algorithm when serving an
+event E to a user Alice:
+
+1. Consider the room state just *before* event E landed (accounting for state
+   resolution accross E's `prev_events`, but not E itself).
+2. Within the state, find the event M with type `m.room.member` and `state_key`
+   set to Alice's user ID.
+3. * If no such event exists, set `membership` to `leave`.
+   * Otherwise, set `membership` to the value of the `membership` property of
+     the content of M.
+
+The new property should be *required* for all servers implementing a version of
+the spec that includes this MSC. However, clients needing to maintain
+compatibility with earlier versions of the spec will need to consider it as
+optional.
+
+For the avoidance of doubt, the new `membership` property is added to all
+Client-Server API endpoints that return events, including
+[`/sync`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3sync),
+[`/messages`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3roomsroomidmessages),
+[`/state`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3roomsroomidstate),
+and deprecated endpoints such as
+[`/events`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3events)
+and
+[`/initialSync`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3events).
+
+
+Example event including the new property:
+
+```json
+{
+  "content": {
+    "membership": "join"
+  },
+  "event_id": "$26RqwJMLw-yds1GAH_QxjHRC1Da9oasK0e5VLnck_45",
+  "origin_server_ts": 1632489532305,
+  "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
+  "sender": "@example:example.org",
+  "state_key": "@user:example.org",
+  "type": "m.room.member",
+  "unsigned": {
+    "age": 1567437,
+    "membership": "leave",
+    "redacted_because": {
+      "content": {
+        "reason": "spam"
+      },
+      "event_id": "$Nhl3rsgHMjk-DjMJANawr9HHAhLg4GcoTYrSiYYGqEE",
+      "origin_server_ts": 1632491098485,
+      "redacts": "$26RqwJMLw-yds1GAH_QxjHRC1Da9oasK0e5VLnck_45",
+      "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
+      "sender": "@moderator:example.org",
+      "type": "m.room.redaction",
+      "unsigned": {
+        "membership": "leave",
+        "age": 1257
+      }
+    }
+  }
+}
+```
+
+## Potential issues
+
+ * Depending on server implementation, it may be expensive or difficult to
+   implement? Feedback welcome from HS authors on this point.
+
+## Alternatives
+
+https://github.com/element-hq/element-meta/issues/2268#issuecomment-1904069895
+proposes use of a Bloom filter — or possibly several Bloom filters — to
+mitigate this problem in a more general way. It is the opinion of the author of
+this MSC that there is room for both approaches.
+
+## Security considerations
+
+None foreseen.
+
+## Unstable prefix
+
+
+## Dependencies
+
+None.
--- a/proposals/4115-membership-on-events.md
+++ b/proposals/4115-membership-on-events.md
@@ -89,9 +89,9 @@ and
 [`/initialSync`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3events).
 
 
-Example event including the new property:
+Example event including the new property, as seen in the response to a request made by `@user:example.org`:
 
-```json
+```json5
 {
   "content": {
     "membership": "join"
@@ -100,11 +100,12 @@ Example event including the new property:
   "origin_server_ts": 1632489532305,
   "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
   "sender": "@example:example.org",
-  "state_key": "@user:example.org",
+  "state_key": "@example:example.org",
   "type": "m.room.member",
   "unsigned": {
     "age": 1567437,
-    "membership": "join",
+    // @user:example.org's membership at the time this event was sent 
+    "membership": "leave",
     "redacted_because": {
       "content": {
         "reason": "spam"
@@ -116,13 +117,13 @@ Example event including the new property:
       "sender": "@moderator:example.org",
       "type": "m.room.redaction",
       "unsigned": {
-        "membership": "leave",
+        // @user:example.org's membership at the time the redaction was sent
+        "membership": "join",
         "age": 1257
       }
     }
   }
 }
-```
 
 ## Potential issues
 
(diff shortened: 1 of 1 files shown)


proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,131 @@
...
+As a partial solution to this problem, we propose a mechanism for servers to
+inform clients of room membership at each event.
+
+## Proposal
+
+The `unsigned` structure contains data added to an event by a homeserver when
+serving an event over the client-server API.  (See
+[specification](https://spec.matrix.org/v1.9/client-server-api/#definition-clientevent)).
+
+We propose adding a new property, `membership`, which should contain the
```
richvdh, line 41: Thoroughly unconvinced that `membership` is the best name here. "Whose membership?"

Suggestions?

turt2live: `my_membership` is the top of my head answer

proposals/4115-membership-on-events.md
turt2live, line 1: Implementation requirements:
* [ ] Server populating field
* [ ] Client using field (demonstrating the solution solves the presented problem)

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,133 @@
...
+
+Depending on the configuration of the room, Bob's server may serve the event
+`C` to Bob's client. However, if the room is encrypted, Bob will not be on the
+recipient list for `C` and the sender will not share the message key with Bob,
+even though, in an absolute time reference, `C` may have been sent at a later
+timestamp than Bob's join.
+
+Unfortunately, there is no way for Bob's client to reliably distinguish events
+such as `A` and `C` that were sent "before" he joined (and he should therefore
+not expect to decrypt) from those such as `D` that were sent later.
```
andybalaam, line 27: Why does the server send an event (other than the user's join event) to the user if they were not a member of this room at the time?

I can understand that if history visibility is turned on, or the room is unencrypted, but I would naively expect such historical events to be sent when requested via another API.

richvdh: I don't entirely follow the question. `/messages` will return events from before you joined the room, assuming `history_visibility` is set to something other than `joined`.

andybalaam: OK, so this is to cover situations when `history_visibility != joined`, and if an event from "before" they join happens when `history_visibility == joined` then this indicates a server bug?

richvdh: yes.

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,133 @@
...
+
+This issue is discussed in more detail at
+https://github.com/element-hq/element-meta/issues/2268.
+
+As a partial solution to this problem, we propose a mechanism for servers to
+inform clients of room membership at each event.
```
ara4n, line 33:
```suggestion
inform clients of their room membership at each event.
```

I had to read this all the way through before I realised it's talking about the server telling Alice whether Alice was in the room at the point a given event was received by the server. I think this would have stopped my confusion.

proposals/4115-membership-on-events.md
ara4n: i think `membership` is good enough here, tbh - as long as the MSC itself is clearer that it's talking about the syncing user's membership rather than the sender's or similar.

proposals/4115-membership-on-events.md
richvdh: sorry!

proposals/4115-membership-on-events.md
uhoreg: I think it's clear enough.  The sender must be in the room at the time they sent the event, so it wouldn't make sense to send the sender's membership.  `membership` might not be a great name, but I can't think of anything that's reasonably short, that makes it much clearer, and doesn't sound awkward.  So I think `membership` is good enough.

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+Bob has joined a room, but at the same time, another user has sent a message
+`C`.
+
+Depending on the configuration of the room, Bob's server may serve the event
+`C` to Bob's client. However, if the room is encrypted, Bob will not be on the
+recipient list for `C` and the sender will not share the message key with Bob,
+even though, in an absolute time reference, `C` may have been sent at a later
+timestamp than Bob's join.
```
bradtgmurray, line 23: To test my own understanding, and assuming this graph is Alice's view of the world, this MSC does not solve the problem of how Alice's client should respond to a room key request from Bob with the index used for message C? Alice can't use this MSC to determine Bob's membership at the time of event C, correct?

richvdh: > assuming this graph is Alice's view of the world

This doesn't really make much sense as an assumption. Both Alice's server and Bob's server should have the *same* view of the world. And *neither* Alice's client *nor* Bob's client can have this view of the world, because clients see a "flattened" version of the DAG graph.

>  Alice can't use this MSC to determine Bob's membership at the time of event C, correct?

Correct.

As discussed at https://github.com/element-hq/element-meta/issues/2268#issuecomment-1889756997: there are two parts to dealing with this send/join race. Part 1 is making sure that the *sender* has a clear idea of who was a member at the point of the event; part 2 is making sure that the *recipient* knows whether or not they were a member at the point  of the event. This MSC deals only with part 2 (which we actually observe to be a more common problem than part 1).

richvdh:
This seems to be working, so

@mscbot fcp merge

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+
+For the avoidance of doubt, the new `membership` property is added to all
+Client-Server API endpoints that return events, including
+[`/sync`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3sync),
+[`/messages`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3roomsroomidmessages),
+[`/state`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3roomsroomidstate),
+and deprecated endpoints such as
+[`/events`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3events)
+and
+[`/initialSync`](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3events).
```
clokep, line 83: Is this a full list? `/context`, `/relations`, `/search` might be other ones.

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+2. Within the state, find the event M with type `m.room.member` and `state_key`
+   set to Alice's user ID.
+3. * If no such event exists, set `membership` to `leave`.
+   * Otherwise, set `membership` to the value of the `membership` property of
+     the content of M.
+
+It is recommended that homeservers populate the new property wherever
+practical, but they may omit it if necessary (for example, if calculating the
+value is expensive, servers might choose to only implement it in encrypted
+rooms). Clients must in any case treat the new property as optional.
```
clokep, line 73: After implementation -- was this necessary or should homeservers always populate it?

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+  }
+}
+```
+
+## Potential issues
+
+ * Depending on server implementation, it may be expensive or difficult to
+   implement? Feedback welcome from HS authors on this point.
```
clokep, line 124: Any idea if this was true now that it has been implemented?

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+
+A similar scenario can arise even in the absence of a forked DAG: clients
+see events sent when the user was not in the room if the room has [History
+Visibility](https://spec.matrix.org/v1.10/client-server-api/#room-history-visibility)
+set to `shared`. (This is fairly common even in encrypted rooms, partly because
+that is the default state for new rooms even using the `private_chat` preset
+for the [`/createRoom`](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3createroom)
+request, and also because history-sharing solutions such as
+[MSC3061](https://github.com/matrix-org/matrix-spec-proposals/pull/3061) rely
+on it.
```
TheArcaneBrony, line 39: Is there any reason a client by itself couldn't distinguish membership based on it's own understanding of event ordering? The only exception here I can think of is events that are received out of order (eg. federation latency)
During pagination with /messages, a client would be aware of whether it is before or after it's own join position, no?

Am I missing something?

proposals/4115-membership-on-events.md
richvdh: It's absolutely *not* a full list.

proposals/4115-membership-on-events.md
richvdh: It was fine in Synapse.

proposals/4115-membership-on-events.md
richvdh: There's an example in the introduction to the MSC. Bob was not a member at event C, but Bob's client has no way to tell that. The fact that C arrived at Bob's client after Bob's join event is absolutely not an indication that Bob was a member at that point.

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,148 @@
...
+inform clients of their room membership at each event.
+
+## Proposal
+
+The `unsigned` structure contains data added to an event by a homeserver when
+serving an event over the client-server API.  (See
+[specification](https://spec.matrix.org/v1.9/client-server-api/#definition-clientevent)).
+
+We propose adding a new optional property, `membership`. If returned by the
+server, it should contain the membership of the user making the request,
```
turt2live, line 52: applies throughout: should means SHOULD or MUST?

turt2live (approved):
overall seems fine. I would like to hear from other client authors if they think this will help their E2EE reliability as well, though don't think we need to block FCP on that feedback.

proposals/4115-membership-on-events.md
turt2live: (could the above comment, particularly the last paragraph, be written into the MSC?)

proposals/4115-membership-on-events.md
clokep: I questioned it since it says all endpoints returning events, but then gives a decently long list. Would probably be clearer to either reduce the list or list all endpoints.

proposals/4115-membership-on-events.md
TheArcaneBrony: That makes sense, and seems to line up with the exception I mentioned earlier.
I'm not super familiar with the server-server spec, but wouldn't this imply that servers *are* aware of event order besides stream order? 
I wonder if an alternative solution might be to signal where in history an event should be inserted in the timeline to clients (eg. an `after_event: $abcdefg`), which sounds like it would also achieve the same effect of clients being aware that these events are before their join event.

proposals/4115-membership-on-events.md
richvdh: I particularly wanted to call out the deprecated endpoints, since they might otherwise get forgotten.

I'll emphasise that the list is not complete.

proposals/4115-membership-on-events.md
richvdh: I'll remove this, given I was able to implement it in synapse without any material impact.

proposals/4115-membership-on-events.md
richvdh: I've expanded the part under "Alternatives" which discusses this.

proposals/4115-membership-on-events.md
richvdh: have attempted to MUST my shoulds

proposals/4115-membership-on-events.md
richvdh: I still think it's reasonable to leave it open.

proposals/4115-membership-on-events.md
richvdh: done, I hope

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,155 @@
...
+## Alternatives
+
+1. https://github.com/element-hq/element-meta/issues/2268#issuecomment-1904069895
+   proposes use of a Bloom filter — or possibly several Bloom filters — to
+   mitigate this problem in a more general way. It is the opinion of the author of
+   this MSC that there is room for both approaches.
+
+2. We could attempt to calculate the membership state on the client side. This
+   might help in a majority of cases, but it will be unreliable in the presence
+   of forked DAGs. It would require clients ti implement the [state resolution
```
clokep, line 140:
```suggestion
   of forked DAGs. It would require clients to implement the [state resolution
```

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,155 @@
...
+  },
+  "event_id": "$26RqwJMLw-yds1GAH_QxjHRC1Da9oasK0e5VLnck_45",
+  "origin_server_ts": 1632489532305,
+  "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
+  "sender": "@example:example.org",
+  "state_key": "@user:example.org",
+  "type": "m.room.member",
+  "unsigned": {
+    "age": 1567437,
+    "membership": "leave",
```
KitsuneRal, line 107: I spent good 10 minutes trying to understand how `"membership": "leave"` can be the user's state right after the event with `"content": { "membership": "join" }` before I realised it is _another_ user's state. That small detail doesn't seem to be mentioned around the example, would be great to have it in the eventual spec PR?

KitsuneRal (approved):

proposals/4115-membership-on-events.md
richvdh: nope, it's just a bad example!

At one point the MSC said that the `unsigned` membership should give the state *before* the event in question. (Hence, "leave".) That later got updated, but I forgot to change the example. 

```suggestion
    "membership": "join",
```

richvdh: wait, no, I've confused myself now :facepalm:

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,155 @@
...
+      "sender": "@moderator:example.org",
+      "type": "m.room.redaction",
+      "unsigned": {
+        "membership": "leave",
+        "age": 1257
+      }
+    }
+  }
+}
+```
```
richvdh, line 125:
```suggestion
Example event including the new property, as seen in the response to a request made by `@user:example.org`:

```json5
{
  "content": {
    "membership": "join"
  },
  "event_id": "$26RqwJMLw-yds1GAH_QxjHRC1Da9oasK0e5VLnck_45",
  "origin_server_ts": 1632489532305,
  "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
  "sender": "@example:example.org",
  "state_key": "@example:example.org",
  "type": "m.room.member",
  "unsigned": {
    "age": 1567437,
    // @user:example.org's membership at the time this event was sent 
    "membership": "leave",
    "redacted_because": {
      "content": {
        "reason": "spam"
      },
      "event_id": "$Nhl3rsgHMjk-DjMJANawr9HHAhLg4GcoTYrSiYYGqEE",
      "origin_server_ts": 1632491098485,
      "redacts": "$26RqwJMLw-yds1GAH_QxjHRC1Da9oasK0e5VLnck_45",
      "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
      "sender": "@moderator:example.org",
      "type": "m.room.redaction",
      "unsigned": {
        // @user:example.org's membership at the time the redaction was sent
        "membership": "join",
        "age": 1257
      }
    }
  }
}
```
```

File: proposals/4115-membership-on-events.md (hunk 11 above)
Comment (richvdh) [reply proposals/4115-membership-on-events.md:108]: hopefully it's now slightly clearer.

File: proposals/4115-membership-on-events.md [hunk 13]
```diff
@@ -0,0 +1,156 @@
...
+      "sender": "@moderator:example.org",
+      "type": "m.room.redaction",
+      "unsigned": {
+        // @user:example.org's membership at the time the redaction was sent
+        "membership": "join",
+        "age": 1257
+      }
+    }
+  }
+}
```
dbkr, line 126:
```suggestion
}```
```

proposals/4115-membership-on-events.md
```diff
@@ -0,0 +1,156 @@
...
+serving an event over the client-server API.  (See
+[specification](https://spec.matrix.org/v1.9/client-server-api/#definition-clientevent)).
+
+We propose adding a new optional property, `membership`. If returned by the
+server, it MUST contain the membership of the user making the request,
+according to the state of the room at the time of the event being returned. If
+the user had no membership at that point (ie, they had yet to join or be
+invited), `membership` is set to `leave`.  Any changes caused by the event
+itself (ie, if the event itself is a `m.room.member` event for the requesting
+user) are *included*.
```
dbkr, line 63: This seems generally well thought through, although it makes it apparent that a client has no way of knowing the actual room state at any given point, and this is a special-case workaround for the user's own membership. Is this is temporary simple fix that will later be extended into some way to know the exact state at a given event (eg. state hash)?

proposals/4115-membership-on-events.md
richvdh: thanks, applied manually

proposals/4115-membership-on-events.md
richvdh: Thanks for the context @turt2live. I think that further confirms my thinking here: for now at least, we should make a simple change that solves the problem at hand rather than trying to be too clever and solve the problem more generally.

proposals/4115-membership-on-events.md
KitsuneRal: 😅 It certainly is.

proposals/4115-membership-on-events.md
dbkr: Fair enough, solving the problem at hand feels like the right way to go.

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1847

turt2live:
Merged 🎉

PR #4119, opened 2024-03-13
MSC: Voluntary content flagging

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/voluntary-flagging/proposals/4119-voluntary-flagging.md)

I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published under my Element hat as a member of the Trust & Safety team, allocated to the Foundation.

### DIFF
--- a/proposals/4119-voluntary-flagging.md
+++ b/proposals/4119-voluntary-flagging.md
@@ -1,0 +1,114 @@
+# MSC4119: Voluntary content flagging
+
+When a user on Matrix sends content which violates a room's code of conduct or a server's terms of
+service, other users often [report](https://spec.matrix.org/v1.9/client-server-api/#reporting-content)
+the event to expediate its removal. Clients typically offer to [ignore](https://spec.matrix.org/v1.9/client-server-api/#ignoring-users)
+the sender as well, preventing further events from being shown to the reporter.
+
+Most server implementations of the existing reporting system simply send the report to the server
+administrators rather than the accused sender's server or applicable room moderators. This can lead
+to some delay in the content being taken down, particularly when it is federated from a non-local
+server. This proposal leaves improving the reporting communication line to another MSC, but does aim
+to address expediated content removal where it's obvious that the content is against the rules.
+
+With this MSC, when a user reports content they *should* be presented with an option to disclose to
+the room that they submitted a report. Ideally, these disclosures would protect the sender's identity
+to avoid targeted abuse, but for now this is left to a future MSC. Contributors for such a feature
+should consider revealing the sender's identity to moderators to reduce abuse of the system.
+
+This MSC's proposed system can additionally be expanded to social media use cases, similar to Twitter's
+[Community Notes](https://en.wikipedia.org/wiki/Community_Notes) feature. For now, the added context
+from this MSC is rather mechanical to suggest that other clients hide or blur the content.
+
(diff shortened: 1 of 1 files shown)


proposals/4119-voluntary-flagging.md
```diff
@@ -0,0 +1,114 @@
...
+  be blurred, and display name/user ID hidden if possible. "Several" may be defined as a percentage
+  of room members in small-medium sized rooms, or 10+ senders in larger rooms.
+* If a "trusted" user flags content, the client applies the above actions immediately without waiting
+  for further confirmation. If the sender is partially trusted, the client may require a couple more
+  confirmations before actually minimizing the content. "Trusted" has a range of values:
+
+  * Untrusted - default value.
+  * Partially trusted - sender has a DM with the user.
+  * Trusted - user has explicitly listed the sender as a trusted person, or has cross-sign verified
+    the user.
```
Gnuxie, line 71: Can we make it clear that this is an example of how to define a trusted user? Since that needs to be completely subjective. And it shouldn't be based on having a DM or cross signing with them, since you have those already with people you don't trust.

proposals/4119-voluntary-flagging.md
```diff
@@ -0,0 +1,114 @@
+# MSC4119: Voluntary content flagging
```
TheArcaneBrony, line 1: I wonder if this could be worded better? "Voluntary content flagging" sounds more like content warnings/content notices and spoilers.

turt2live: Suggestions are welcome. There is an amount of that in here as well.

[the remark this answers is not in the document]
turt2live: This proposal is part of the design process: all MSCs are evaluated for usefulness as part of regular review, and can be accepted or rejected accordingly later in the process. 

The history here is someone brought up a concern in the spec room about being able to hide reported content in their clients, but the spec obviously doesn't have that feature right now. It was realized that my half-written community notes MSC could be tweaked a bit to serve the purpose, and so I did that and opened it.

As this proposal is early in the review process, more research is certainly required. It starts with community notes because that was most familiar, but an evaluation of Discourse's community flagging feature would be good as well. I suspect that it'd still be kicked to a future MSC though.

proposals/4119-voluntary-flagging.md
```diff
@@ -0,0 +1,114 @@
...
+these data points is left as an implementation detail, though a suggested approach is:
+
+* If several people append the same flag to an event, the event should be minimized or have a click
+  barrier to prevent the content from being immediately accessible. The sender's avatar should additionally
+  be blurred, and display name/user ID hidden if possible. "Several" may be defined as a percentage
+  of room members in small-medium sized rooms, or 10+ senders in larger rooms.
+* If a "trusted" user flags content, the client applies the above actions immediately without waiting
+  for further confirmation. If the sender is partially trusted, the client may require a couple more
+  confirmations before actually minimizing the content. "Trusted" has a range of values:
+
```
TobiasFella, line 67: IMO it would make sense to recommend that clients hide *all* messages sent by the respective person.

PR #4123, opened 2024-04-06
MSC4123: Allow `knock` -> `join` transition

Kladki:
[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/knock-join-transition/proposals/4123-knock-join-transition.md)

### DIFF
--- a/proposals/4123-knock-join-transition.md
+++ b/proposals/4123-knock-join-transition.md
@@ -27,7 +27,7 @@ In a new room version, we make `knock` -> `join` a valid transition, as it would
 easier to handle, as an unnecessary `knock` -> `leave` transition would no longer need to be sent.
 
 When it comes to authorizing a `knock` -> `join` transition, the same logic can be used as a `leave` -> `join`
-transition, as the same could be achived with a `knock` -> `leave` -> `join` transition.
+transition, as the same could be achieved with a `knock` -> `leave` -> `join` transition.
 
 ## Potential issues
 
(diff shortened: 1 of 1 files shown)


proposals/4123-knock-join-transition.md
tulir, line 1: @zecakeh pointed out that these transitions are already allowed according to the [auth rules](https://spec.matrix.org/v1.10/rooms/v11/#authorization-rules)

As long as the user isn't banned, rule 4.3.6 allows joining if `join_rule` is `public` regardless of current membership, and rule 4.3.5.3 does the same for restricted rooms if the authorised via user key is valid.

image

So it seems like [the table](https://spec.matrix.org/v1.10/client-server-api/#mroommember) and [the graph](https://spec.matrix.org/v1.10/client-server-api/#room-membership) in the client-server API are just wrong and can be fixed without an MSC

tulir: And apparently ¼ of the pull request already exists: matrix-org/matrix-spec#1709 (missing parts: knock -> join for public rooms in the graph, and updating the table for both cases)

Kladki:
As per https://github.com/matrix-org/matrix-spec-proposals/pull/4123#discussion_r1554948586

proposals/4123-knock-join-transition.md
FSG-Cat: i have to ask if this has been confirmed to be correct behavior so we arent walking straight into another Knock Invite mess. Like im not taking a stance on if this needs a MSC or not just asking that we make sure we arent walking into another mess of that magnitude.

Kladki:
ruma (and hence conduit) already allowed the `knock` -> `join` transition, not sure about synapse and dendrite.

proposals/4123-knock-join-transition.md
Kladki: Ruma (and hence Conduit) already allowed the `knock` -> `join` transition, not sure about Synapse and Dendrite.

PR #4124, opened 2024-04-09
MSC4122: Simple Server Authorization

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/server-auth-simple/proposals/4122-server-auth-simple.md)

### DIFF
--- a/proposals/0000-server-auth-simple.md
+++ b/proposals/0000-server-auth-simple.md
@@ -1,0 +1,130 @@
+# MSC0000: Simple server authorization
+
+This MSC proposes simple authorization rules that consider the origin
+server of a given event, with the aim of replacing `m.room.server_acl`.
+
+This is a compromising MSC based on [MSC4099](https://github.com/matrix-org/matrix-spec-proposals/pull/4099)
+that tries to use concepts even more inline current authorization events,
+and without changing the `/make_join` handshake.
+This MSC was also created in reaction to [MSC2870](https://github.com/matrix-org/matrix-spec-proposals/pull/2870),
+that describes itself as stop gap to cover what the MSC has described
+short comings of `m.room.server_acl`. We also agree that MSC2870 is
+a stop gap and that the `m.room.server_acl` has severe shortcomings,
+but we take the view that after 4 years of proposed stop-gapping,
+there is enough time to introduce a more complete solution.
+
+## Proposal
+
+### The subscription authorization rule
+
+This rule is be inserted after rule 3 in version 11, the check
+for `m.room.create`'s content field `m.federate`.
+
+1. If the type is `m.server.subscription`:
+   1. If the `state_key` does not contain the server name for the
+      origin server, reject.
+   2. If the origin servers's current participation is `permitted`, allow.
+   2. If the `m.server.subscription_rule` is `deny`, reject.
+   3. If the origin server's current participation is `deny`, reject.
+   4. Otherwise allow.
+
+The purpose of this rule is to allow a server to send a subscription
+event, even if the `sender` has no membership event.
+
+### The participation authroziation rule
+
+This rule is to be inserted before rule 4 in version 11,
+the check for `m.room.member`, and after the subscription rule
+described in this proposal.
+
+1. If the origin server's current `participation` state is not `permitted`:
+   1. If the `participation` state is `deny`, reject.
+   2. If the `m.server.subscription_rule` is `deny`, reject.
+   3. If the `m.server.subscription_rule` is anything other than `passive`, reject.
+
+### The `m.server.participation` authorization event, `state_key: ${origin_server_name}`
+
+This is an authorization event that is used to authrorize events
+originating from the server named in the `state_key`.
+
+`participation` can be one of `permitted` or `deny`.
+`participation` is protected from redaction.
+
+A denied server must not be sent a `m.server.participation` event unless
+the targeted server is already present within the room, or it has
+an existing `m.server.subscription` event.
+This is to prevent malicious servers being made aware of rooms that
+they have not yet discovered.
+
+A `reason` field can be present alongside `participation` in order to
+explain the reason why a server has been `denied`.
+This reason is to be shown to a joining, or previously present
+server, so that the server's users can understand why they are not
+being allowed to participate.
+
+### The `m.server.subscription_rule` event, `state_key: ''`
+
+This event has one field, `rule` which can be one of the following:
+
+- `deny`: Users are unable to send the `m.server.subscription` event
+  unless there is an existing `m.server.participation` event for the
+  server.
+- `passive`: Users can send the `m.server.subscription` event without
+  corresponding membership or server participation.
+
+`rule` is protected from redaction.
+
+### The `m.server.subscription` event, `state_key: ${origin_server_name}`
+
+This event has one field, `reason` which is optional. This field
+is redactable. The intent is for the reason field to be redactable
+by room admins.
+
+Servers MUST check the `membership` of the `sender` before sending
+to confirm that the `membership` of the sender is not `ban`.
+This is to prevent users changing the reason to be malicious without
+being joined to the room, without forcing room admins to `deny` the
+entire server from participating. Room admins should `deny`
+the server if they continue to modify the reason maliciously or
+fail to perform the aforementioned check. This will prevent the server
+from sending furhter `m.server.subscription` events.
+
+## Potential issues
+
+### Racing with `m.server.subcription_rule`?
+
+We will embed `m.server.subscription_rule` in `m.room.create` if it
+someone raises concerns about a potential race condition or other issue
+about this conflicting with `m.server.participation`. However, stating
+that there might be without elaboration is not helpful, I'd need to
+know how the race works. If there is insistence, then we will embed
+within the `m.room.create` event.
+
+### `subscription` might not be an ideal term
+
+`subscription` might not be an ideal term.
+
+### Soft failure of backfilled messages
+
+Servers that had `participation` of `permitted` that are later
+denied via `deny`, will have their historical messages soft failed by
+servers which later join.
+
+This should be addressed with [MSC4104](https://github.com/matrix-org/matrix-spec-proposals/pull/4104).
+
+## Alternatives
+
+- [MSC4099](https://github.com/matrix-org/matrix-spec-proposals/pull/4099) Participation based authorization for servers in the Matrix DAG
+- [MSC3953](https://github.com/matrix-org/matrix-spec-proposals/pull/3953) Server capability DAG
+
+## Security considerations
+
+None considered.
+
+## Unstable prefix
+
+`me.marewolf.msc0000.*`
+
+## Dependencies
+
+No dependencies
--- a/proposals/0000-server-auth-simple.md
+++ b/proposals/0000-server-auth-simple.md
@@ -74,8 +74,24 @@ This event has one field, `rule` which can be one of the following:
   server.
 - `passive`: Users can send the `m.server.subscription` event without
   corresponding membership or server participation.
+- `active`: Users can send the `m.server.subscription` event but
+  cannot send any other event without a corresponding
+  participation of `permitted`.
 
 `rule` is protected from redaction.
+
+The `passive` state allows for rooms to operate as they do today,
+new servers can freely join a room and start sending events without
+prior approval from the administrators
+
+The `active` state allows for a much safer way to run public Matrix rooms,
+new servers can join a room, send the `m.srever.subscription` event
+but cannot do more until a room administrator permits the new joiner with
+an `m.server.participation` event. We expect that in practice automated
+tooling will perform a simple reputation check and immediately permit
+a new server to participate. This is an essential part of the proposal
+as the `active` mechanism eliminates a current shortfall that
+`m.room.server_acl` is a purely reactive tool in a join wave attack.
 
 ### The `m.server.subscription` event, `state_key: ${origin_server_name}`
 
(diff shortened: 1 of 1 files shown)


proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,149 @@
...
+
+This rule is be inserted after rule 3 in version 11, the check
+for `m.room.create`'s content field `m.federate`.
+
+1. If the type is `m.server.subscription`:
+   1. If the `state_key` does not contain the server name for the
+      origin server, reject.
+   2. If the origin server's current participation is `permitted`, allow.
+   2. If the `m.server.subscription_rule` is `deny`, reject.
+   3. If the origin server's current participation is `deny`, reject.
```
Gnuxie, line 31: This rule that consider the `current participation` seem weak, maybe impossible? I'm asking for some help here:

The authorization rules for `membership` use similar woreding `current membership`, and it's also not clear to me whether this current membership needs to be referred to by the considered event as an authorization event for it to be considered "current". 

The reality is that it does, especially around the rule for invite, "current membership" does seem to mean that the considered event needs to reference this membership event as an authorization event.

If this is true, then this rule will not work.

Gnuxie: I'm going to rewrite the rules not to appeal to the phrasing `current membership` because this does not seem right at all.

Gnuxie: This was clarified in the spec room https://matrix.to/#/%23matrix-spec%3Amatrix.org/%24RUKP8nFl7uHyL6MTk9DejDEveoviTrsRm4ko00uyomc?via=matrix.org&via=envs.net&via=element.io

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,164 @@
...
+### The `m.server.knock` event, `state_key: ${origin_server_name}`
+
+This event has no fields, because it can only be sent once, and
+therefore cannot be edited if the wrong or malicious information
+is provided.
+
+The intent of the event is to only let the room administrators
+explicitly aware of the server's existence.
+
+### The `make_server_knock` handshake
```
Gnuxie, line 115: If someone can find a way to avoid the handshake, that would be great.

The only reason why it is included is because we use the signing of the `m.server.knock` template to verify the origin of the event. And it stops other servers creating `m.server.knock` events for servers that do not exist or servers that are not wanting to send a knock event to the room.

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+This could be addressed with [MSC4104](https://github.com/matrix-org/matrix-spec-proposals/pull/4104).
+
+### Mismatch with `m.room.power_levels`
+
+There is an argument to be made that the ability to manage
+`m.server.participation` should not be flat in the way that
+`m.room.server_acl` is. Consider Alice being the room creator
+and Bob being an admin. Bob could create an `m.server.participation`
+event that denies Alice's server from participating, even if Alice
+is the same or a higher power level.
```
Gnuxie, line 155: It's unclear to me how to deal with this, considering there is no conception of server power or authority, unlike in previous proposals. It initially seems very complicated to map user power level back to servers, but thinking about it, it could be done.

You could take the power level event, and map each user to a server and work out what the server's highest ambient power level is.

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
+# MSC4124: Simple server authorization
+
+This MSC proposes simple authorization rules that consider the origin
+server of a given event, with the aim of replacing `m.room.server_acl`.
```
turt2live, line 4: This could do with a paragraph or two before this one which describes how server ACLs work today, and why they're insufficient. 

For example:
> [Server ACLs] in Matrix operate as a network firewall described by the room, requiring participation from all joined servers to be fully enforced. As they are not applied to the room model though, leaks can occur where a server not enforcing the ACLs (typically due to simply not implementing them) "pulls" events from banned servers into the room via `prev_events`. When an event referencing the banner server's events is sent over federation, the remote servers in turn fetch the banned events and present them to users.
> 
> Workarounds currently include banning/kicking all users from a banned server, and monitoring for leaks to potentially ban servers not enforcing the ACLs. These solutions require a room moderator to be highly knowledgeable on how ACLs actually work, which is rarely the case.

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+
+This is a compromising MSC based on [MSC4099](https://github.com/matrix-org/matrix-spec-proposals/pull/4099)
+that tries to use concepts even more inline current authorization events
+This MSC was also created in reaction to [MSC2870](https://github.com/matrix-org/matrix-spec-proposals/pull/2870),
+that describes itself as stop gap to cover what the MSC has described
+short comings of `m.room.server_acl`. We also agree that MSC2870 is
+a stop gap and that the `m.room.server_acl` has severe shortcomings,
+but we take the view that after 4 years of proposed stop-gaping,
+there is enough time to introduce a more complete solution.
```
turt2live, line 13: The language around MSC2870 here feels a bit strong. I'd suggest something along the lines of this:

```suggestion

[MSC2870](https://github.com/matrix-org/matrix-spec-proposals/pull/2870)
describes a mechanism to help prevent rooms from breaking when a
room moderator redacts a server ACL event, and notes that such a
mechanism should be considered temporary until another MSC can replace
the server ACLs system with something more built into the room model.
This MSC is one variation on that future proposal.
```

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+
+Related issues:
+- https://github.com/matrix-org/matrix-spec/issues/928
+
+## Proposal
+
+### The `m.server.knock` authorization rule
+
+This rule is be inserted after rule 3 in version 11, the check
```
turt2live, line 22: link to the spec please

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+for `m.room.create`'s content field `m.federate`.
+
+1. If the type is `m.server.knock`:
+   1. If the `state_key` does not contain the server name for the
+      origin server, reject.
+   2. If there is existing state for the origin server's `m.server.knock`, reject.
```
turt2live, line 28: What does "existing state" mean here?

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+This is because an `m.server.participation` event set to `deny` will
+usually be topologically older than an `m.server.knock` due to the
+`m.server.participation` usually referencing a recent
+`m.room.power_levels` event. And so `m.server.knock` events could
+be crafted by malicious servers without restriction without
+rule 1.2.
+
+### The `m.server.participation` authorization rule
+
+This rule is to be inserted before rule 4 in version 11,
```
turt2live, line 49: link to the spec please

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+described in this proposal.
+
+1. If the origin server's current `participation` state is not `permitted`:
+   1. If the `participation` state is `deny`, reject.
+   2. If the `m.server.knock_rule` is `deny`, reject.
+   3. If the `m.server.knock_rule` is anything other than `passive`, reject.
+
+### The `m.server.participation` authorization event, `state_key: ${origin_server_name}`
+
+This is an authorization event that is used to authorize events
```
turt2live, line 60: "authorization event" likely needs a link to the spec, describing what that means.

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+This rule is to be inserted before rule 4 in version 11,
+the check for `m.room.member`, and after the `m.server.knock` rule
+described in this proposal.
+
+1. If the origin server's current `participation` state is not `permitted`:
+   1. If the `participation` state is `deny`, reject.
+   2. If the `m.server.knock_rule` is `deny`, reject.
+   3. If the `m.server.knock_rule` is anything other than `passive`, reject.
+
+### The `m.server.participation` authorization event, `state_key: ${origin_server_name}`
```
turt2live, line 58: Can users from the affected server send this event? ie: can `@alice:example.org` ban `example.org` (or permit it).

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,173 @@
...
+that describes itself as stop gap to cover what the MSC has described
+short comings of `m.room.server_acl`. We also agree that MSC2870 is
+a stop gap and that the `m.room.server_acl` has severe shortcomings,
+but we take the view that after 4 years of proposed stop-gaping,
+there is enough time to introduce a more complete solution.
+
+Related issues:
+- https://github.com/matrix-org/matrix-spec/issues/928
+
+## Proposal
```
turt2live, line 18: Overall I think some sequence diagrams and example events would be helpful here. It's not clear to me what causes a server to try knocking (is it based on user action? which actions?), or how the various events really interact with each other.

turt2live (commented):
Thanks for opening the MSC! I'm still thinking about how this might work long-term in a room, and have left some clarifying questions/comments in the meantime.

proposals/4124-server-auth-simple.md
Gnuxie: is there anywhere in the specification the describes what it means?

I have linked to https://spec.matrix.org/v1.10/server-server-api/#auth-events-selection for now.

proposals/4124-server-auth-simple.md
Gnuxie: In the way the proposal is currently written, if they have the power level to send `m.server.participation` then yes. I'm not sure whether that makes sense though, as the only time it needs to be allowed is when `m.server.knock_rule` is `active` around room creation, but maybe the room creation flow needs describing in the MSC.

Gnuxie: I probably need to add an equivalent of rule 4.3.1 https://spec.matrix.org/unstable/rooms/v11/, `[Changed in this version] If the only previous event is an m.room.create and the state_key is the sender, allow.` for sending `m.server.participation` after `m.room.create`

proposals/4124-server-auth-simple.md
Gnuxie: Yes, that's a really good idea, I'll try find the way other MSCs embed diagrams

Gnuxie: I've added diagrams for room creation and joining a room, is there any other scenarios you think will be helpful to show on a graph?

proposals/4124-server-auth-simple.md
turt2live: That's probably the best place for now.

proposals/4124-server-auth-simple.md
turt2live: The diagrams look like a great start, thank you! Could we get some examples of what the flows look like when the knock rule isn't `passive`? 

Separately, some words on what Bob's user experience looks like would be good. This won't be a normative section, but helps ensure the proposal has considered user impact in its decisions.

proposals/4124-server-auth-simple.md
Gnuxie: I have now added a giant context section, but maybe this section should be moved to https://github.com/matrix-org/matrix-spec/issues/928? what do you think?

Gnuxie: I think that the section should be moved to the issue because it's independent to the proposal, but the proposal should be modified to keep referring back to it. But I want to hear from you before I do that.

turt2live: Proposals should have enough context to be understood in isolation, as issues may be edited or mutated after the MSC is accepted. https://github.com/matrix-org/matrix-spec-proposals/pull/4126 is an example for incorporating issue context into the proposal.

Gnuxie: Ok, in that case I will add each point to the issue and make a shorter reference to them.

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,427 @@
...
+It is theorized by me completely from thin air that the server ACL
+event can contain roughly `512` entries before a limit to the
+event size is reached. A survey of #matrix-org-coc-bl:matrix.org
+suggests that most ACL events contain `150`~ deny entries.
+However, there have been unstable periods in Matrix's history
+due to vulnerability of servers where the number of deny entries
+has been much greater. For example, Synapse previously had weak
+registration requirements by default that was exploited in a major
+incident. There could be a need to protect rooms from vulnerable
+servers again in the future.
```
Gnuxie, line 94: @tulir said

> 64kb / 20 bytes per server = 3k
> or if someone attacked the system with max length domains (63 bytes + tld = 75 or so in total) the limit would be 873

proposals/4124-server-auth-simple.md
```diff
@@ -0,0 +1,427 @@
+# MSC4124: Simple server authorization
```
kegsay, line 1: I think we do need something like this (DAG aware server ACLs) in Matrix, in order to address the transitive delivery of abusive events ("leaky servers" in the proposal).

I think this MSC is trying to do too much though. Specifically, the "ambient room access" problem and associated explicit allow list adds a footgun (locking out good servers) and makes the federation more closed in general (the bootstrapping problem) to an otherwise good proposal.

Gnuxie: > I think this MSC is trying to do too much though. Specifically, the "ambient room access" problem and associated explicit allow list adds a footgun (locking out good servers) and makes the federation more closed in general (the bootstrapping problem) to an otherwise good proposal.

From my perspective solving the "ambient room access" problem is the only way to make the system safe by default and I would not be satisfied by a solution that doesn't address this. We don't go far enough in this proposal in my opinion, because once a server has been allowed to participate just once, we get into the soft failure problem and we're never able to "stop" the server using its initial access to create events.

Going further to address the issue of "irrevocability of access" would also allow the proposal to be less skewed to the risk of managing access in a "first encounter" with a new server. Which would help with bootstrapping.  

Given that policy servers have changed the precedent on authorities in rooms (in the time since this proposal was written), giving more control to policy servers might be a new way to help solve the issues with this proposal.

Gnuxie:

PR #4125, opened 2024-04-09
Proposal to specify servers to join via for federated invites

Kladki:
[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/federation-invite-via/proposals/4124-federation-invite-via.md)

### DIFF
--- a/proposals/4124-federation-invite-via.md
+++ b/proposals/4124-federation-invite-via.md
@@ -1,0 +1,86 @@
+# MSC4124: Specify servers to join via when inviting over federation
+
+## Introduction
+
+Currently when a user is invited over federation, the receiver of the invite does not know what servers
+are in the room, so when attempting to join over federation, servers typically have to guess what server
+to join via, typically with the `sender`'s server.
+
+However, the server of the user who sent the invite may not be online, meaning that the user is unable to
+join the room they are invited to until the `sender`'s server comes online, if it ever does. If they do not
+know anyone else in the room, they cannot even ask out-of-bounds for someone else to invite them to the room.
+
+## Proposal
+
+The proposed fix is to create a new version of the `PUT /_matrix/federation/*/invite` endpoint based on
+`PUT /_matrix/federation/v2/invite`, with the only change being an additional unsigned `via` field in the request
+body to inform the server it is inviting of possible servers it can join via. This endpoint would be denoted as
+`PUT /_matrix/federation/v3/invite`. Here is an example of this new body:
+
+```json
+{
+  "content": {
+    "membership": "invite"
+  },
+  "origin": "matrix.org",
+  "origin_server_ts": 1234567890,
+  "sender": "@alice:matrix.org",
+  "state_key": "@bob:example.org",
+  "type": "m.room.member",
+  "unsigned": {
+    "invite_room_state": [
+      {
+        "content": {
+          "name": "Example Room"
+        },
+        "sender": "@alice:matrix.org",
+        "state_key": "",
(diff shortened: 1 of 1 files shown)


proposals/4125-federation-invite-via.md
```diff
@@ -0,0 +1,94 @@
...
+join the room they are invited to until the `sender`'s server comes online, if it ever does. If they do not
+know anyone else in the room, they cannot even ask out-of-bounds for someone else to invite them to the room.
+
+In the case of rejecting invites, it can lead to users of the room believing that the invited user has neither
+accepted nor rejected the invite, while it might have been locally rejected by the invited user's homeserver due
+to it being unable to reach the sender's homeserver.
+
+## Proposal
+
+The proposed fix is to create a new version of the `PUT /_matrix/federation/*/invite` endpoint based on
```
zecakeh, line 19: If the change is only to add a new field, why create a new version of just adding the field to the old version? This seems like a backwards-compatible change to me.

zecakeh: I believe that new versions are only for breaking changes, but don't take my word for it.

Kladki: Seems like you are right: https://spec.matrix.org/proposals/#technical-notes
> Backwards incompatible changes to endpoints are allowed when no other alternative exists, and must be versioned under a new major release of the API.

Kladki: Would it make sense to add this field to both `v1` and `v2` of this endpoint, or just `v2`?

zecakeh: I don't know the server-server API very well but it seems to me that `v1` is only kept around for compatibility with old implementations. If a server doesn't implement `v2` it's unlikely to implement a new field in `v1`.

Kladki: [Done](https://github.com/matrix-org/matrix-spec-proposals/pull/4125/commits/b0fbcd3934b19c2ea94a4b52228263171cb86df2)

proposals/4125-federation-invite-via.md
```diff
@@ -0,0 +1,100 @@
...
+
+In the case of rejecting invites, it can lead to users of the room believing that the invited user has neither
+accepted nor rejected the invite, while it might have been locally rejected by the invited user's homeserver due
+to it being unable to reach the sender's homeserver.
+
+## Proposal
+
+The proposed fix is to add an additional `via` field in the request body of `PUT /_matrix/federation/v2/invite`
+to inform the server it is inviting of possible servers it can join via. The same is **not** done for `v1` of this
+endpoint as `v2` should be used by default, and a sever that only impelemnts `v1` is unlikely to update to support
```
zecakeh, line 21:
```suggestion
endpoint as `v2` should be used by default, and a server that only implements `v1` is unlikely to update to support
```

Kladki: I would have thought the spell check CI would have run by now. :thinking:

proposals/4125-federation-invite-via.md
Kladki, line 1: Now that I think about it, it would make sense that for a given room, if an invite happens over federation for one user with servers `a` `b` and `c` specified in the via field, but then later another user on the same server also gets invited with servers `d` `e` and `f`, the server should try to join via the servers specified in the most recent invite, so in this case `d` `e` and `f`. 

Not sure if it would be a good idea to try servers `a` `b` and `c` as a fallback though.

PR #4126, opened 2024-04-10
MSC4126: Deprecation of query string auth

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/deprecate-qs-auth/proposals/4126-deprecate-query-string-auth.md)

Fixes https://github.com/matrix-org/matrix-spec/issues/1780

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4126#issuecomment-2048506009)

### DIFF
--- a/proposals/4126-deprecate-query-string-auth.md
+++ b/proposals/4126-deprecate-query-string-auth.md
@@ -1,0 +1,74 @@
+# MSC4126: Deprecation of query string auth
+
+Presently, the Client-Server API allows clients to provide their access token via the `Authorization`
+request header or via an `access_token` query string parameter, described [here](https://spec.matrix.org/v1.10/client-server-api/#using-access-tokens).
+Clients are already encouraged to use the header approach, though the query string option exists for
+largely backwards compatibility reasons.
+
+The query string approach is subject a number of security, usability, and practical concerns, discussed
+on [matrix-spec#1780](https://github.com/matrix-org/matrix-spec/issues/1780):
+
+* The query string of an HTTP request is often logged by the client itself, middleware reverse proxy,
+  and application/homeserver as well. Though some of these layers may be aware of this issue, they
+  can trivially accidentally break to log sensitive credentials again. By contrast, headers are not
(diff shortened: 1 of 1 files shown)


proposals/4126-deprecate-query-string-auth.md
turt2live, line 1: Implementation requirements:

* None, in my opinion. The MSC is not feasible to implement beyond what has already been described in the proposal itself.

tulir:
Looks good, I see no reason to delay this

@mscbot fcp merge

proposals/4126-deprecate-query-string-auth.md
```diff
@@ -0,0 +1,74 @@
...
+* Having two ways of doing things could lead to compatibility issues, where a client using the query
+  string approach is tried against a server which only supports the header. The client ends up not
+  working, leading to subpar user experience.
+
+* Most clients have already adopted the header approach, largely forgetting that the query string
+  even exists. Continuing to support the query string option leaves some maintenance burden for what
+  is effectively unused code.
+
+* Matrix has [decided](https://matrix.org/blog/2023/09/matrix-2-0/) to adopt OIDC for authentication,
+  which [does not allow](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.3.2)
```
richvdh, line 32: This isn't quite right. https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics is a "Best practices" document for OAuth 2.0, so this should maybe be:

```suggestion
  which is based on OAuth 2.0, which [advises against](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-4.3.2)
```

Really, this is just evidence that this is a sensible change to make (ie, others are thinking the same way), rather than a concern in its own right.

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1808

richvdh:
Merged to the spec! :tada:

PR #2832, opened 2020-10-24
MSC2832: HS -> AS authorization header

tulir:
[Rendered](https://github.com/tulir/matrix-doc/blob/appservice-auth-fix/proposals/2832-appservice-auth-fix.md)

Fixes matrix-org/matrix-spec#679 

This has been implemented by mautrix-python for a long time already, as I had mistakenly thought the spec was fixed long ago.

Signed-off-by: Tulir Asokan &lt;tulir@maunium.net&gt;

----

[FCP Checkboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/2832#issuecomment-1151403491)

PR #4127, opened 2024-04-10
MSC4127: Removal of query string auth

### LINKED ISSUE #1780: Passing access tokens as query parameters should be deprecated
Some history:
 * Originally, access tokens were passed from clients to servers as an `access_token` query parameter.
 * https://github.com/matrix-org/matrix-doc/pull/1044 and https://github.com/matrix-org/matrix-doc/pull/1517 changed the spec to encourage clients to pass the access token via an `Authorization: Bearer` header, while continuing to permit the use of the query parameter for clients that needed it.

IMHO it is past time to deprecate support for the query parameter with a view to removing it in a future spec version, for the following reasons:

 * Use of the query parameter is highly unsatisfactory from the security point-of-view (in particular: it is very easy to leak an access token via request logs or an accidental cut-and-paste), and its presence is an unnecessary footgun for client implementors.
 * Having multiple ways to do something is bad for compatibility and best avoided where practical. (Client X gets developed against Synapse which supports the query parameter. Server Y is developed against the Element clients which do not use the query parameter. All is well until someone tries to use Client X with Server Y.)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/remove-qs-auth/proposals/4127-remove-query-string-auth.md)

Requires https://github.com/matrix-org/matrix-spec-proposals/pull/4126
Fixes https://github.com/matrix-org/matrix-spec/issues/1780

----

SCT Stuff:

No MSC checklist

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4127#issuecomment-4453448418)

### DIFF
--- a/proposals/4127-remove-query-string-auth.md
+++ b/proposals/4127-remove-query-string-auth.md
@@ -1,0 +1,56 @@
+# MSC4127: Removal of query string auth
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126) for details on query
+string authentication and the prerequisite deprecation.
+
+## Proposal
+
+Query string authentication becomes *removed* from the [Client-Server API](https://spec.matrix.org/v1.10/client-server-api/#using-access-tokens)
+and [Identity Service API](https://spec.matrix.org/v1.10/identity-service-api/#authentication).
+
+The prior deprecation required by the [deprecation policy](https://spec.matrix.org/v1.10/#deprecation-policy)
+is described by [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126).
+
+As a process note, it was considered to bundle the deprecation and removal into MSC4126 instead of
+having two separate MSCs, breaking the deprecation policy slightly, though this introduces consequences
+which are, in the author's opinion, undesirable. Specifically, the spec release changelog for the
+deprecation would say "as per MSC4126" and imply the MSC being moved to the [merged](https://spec.matrix.org/proposals/#process)
+state, however the second half of the MSC (the feature removal) would not have landed yet. This would
+leave the MSC stuck in an awkward `spec-pr-missing` state because half of it had not been written up
+or landed yet. To maintain a clearer changelog, ease process considerations, and give clients another
+opportunity to object to the removal, the two MSC approach required by the deprecation policy is best.
+
+## Potential issues
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126).
+
+## Alternatives
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126).
+
+More of a process note, during drafting of this proposal the author considered combining the deprecation
+and removal into a single MSC, against the advice of the deprecation policy. The aim was to reduce
+process burden and more quickly get the security risk removed from the specification, though merging
+both operations into a single MSC potentially makes the spec release changelog unclear and leaves the
+MSC in an awkward released-but-not-merged state. To avoid figuring out how to navigate the process
+states, increasing the process burden rather than decreasing it, the deprecation policy's two MSC
+requirement has been followed here. However, there are still other process efficiencies which can
+take place - see the "Dependencies" section of this MSC for details.
+
+## Security considerations
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126).
+
+## Unstable prefix
+
+This proposal cannot feasibly have an unstable prefix. Clients are already discouraged from using
+query string authentication and should switch to `Authorization` as soon as possible, regardless of
+this MSC.
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126) for related details.
+
+## Dependencies
+
+[MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126) must be *released* in the
+specification before this proposal can be merged, though this MSC *may* progress through Final Comment
+Period concurrent to MSC4126 if desirable.


proposals/4127-remove-query-string-auth.md
turt2live, line 1: Implementation requirements:

* None, in my opinion. The MSC is not feasible to implement beyond what has already been described in the proposal itself.

proposals/4127-remove-query-string-auth.md
```diff
@@ -0,0 +1,56 @@
+# MSC4127: Removal of query string auth
+
+See [MSC4126](https://github.com/matrix-org/matrix-spec-proposals/pull/4126) for details on query
+string authentication and the prerequisite deprecation.
+
+## Proposal
+
+Query string authentication becomes *removed* from the [Client-Server API](https://spec.matrix.org/v1.10/client-server-api/#using-access-tokens)
+and [Identity Service API](https://spec.matrix.org/v1.10/identity-service-api/#authentication).
```
clokep, line 9: An issue that Synapse ran into with [MSC2832](https://github.com/matrix-org/matrix-spec-proposals/pull/2832) which removed query parameter auth for app services is that we did not want to drop support for old spec versions, but *did* want to follow this guidance. (https://github.com/matrix-org/synapse/issues/15379 has some of the discussion.)

We ended up saying Synapse was still compatible with 1.4, but had a configuration flag for re-enabling the old behavior (disabling it by default in the name of secure defaults).

----

I'm not sure if I have a recommendation, but wanting to be both backwards compatible and secure by default was difficult in this situation. I guess one option would have been to only declare support for the older versions if the config flag was flipped, but I think various clients would have stopped working if we did that.

proposals/4127-remove-query-string-auth.md
tulir: I looked through https://matrix.org/ecosystem/ and found the following projects that seemed to still be using the query parameter:

* Client SDKs
  * libcmatrix: https://source.puri.sm/Librem5/libcmatrix/-/blob/c4e9a8552f6870829a987820cc0f66323dd5e9ff/src/cm-net.c#L271 (cc @agx)
  * Matrix-ClientServer-API-java: https://github.com/JojiiOfficial/Matrix-ClientServer-API-java/blob/cb733066b8cc0c1ea51047efba8212de67965331/src/main/java/de/jojii/matrixclientserver/Bot/Syncee.java#L83 (cc @JojiiOfficial)
  * I couldn't find how https://github.com/Dominaezzz/matrix-kt does auth, but it's archived
* Bot SDKs
  * MatrixBot .NET Core SDK: https://github.com/enimatek-nl/matrixbot-dotnet-sdk/blob/adebefc612de3f002a1f1cc3183923745876709a/MatrixBot.Sdk/MatrixBot.cs#L202
  * simplematrixbotlib: https://codeberg.org/imbev/simplematrixbotlib/src/commit/17ead9cb2cc86b09e39c21acc015ad1500d4ed4b/simplematrixbotlib/api.py#L112 (I think it uses matrix-nio for most requests, so probably just that one outdated request - cc @imbev)
  * mxbt: https://codeberg.org/librehub/mxbt/src/commit/aee47b2a47c2e997277c727dac61d301850785de/mxbt/bot.py#L127 (seems to be copied from simplematrixbotlib)
* Bots
  * Time-to-Matrix: https://github.com/etkecc/ttm/blob/b9e80e9977eb257b677e25346ea720ae336d5e6b/internal/matrix/matrix.go#L185 (cc @aine-etke)
  * hemppa: https://github.com/vranki/hemppa/blob/907a73eb36e5e2b38dbd3ec35503baf91ec7404a/bot.py#L543 (cc @vranki)
* Bridges: none
* Clients: none (but Chatty uses libcmatrix mentioned above)

There are also a bunch of matches on GitHub search, but I'd guess most of them are ad-hoc scripts, unmaintained projects and other such things: https://github.com/search?q=%2F_matrix+access_token%3D++NOT+is%3Aarchived+NOT+is%3Afork&type=code

tulir: Not sure how GitHub sorts those search results, but when I opened it again, the order had changed and the top result was GitLab. Should probably open an issue or PR for them 🤔 https://gitlab.com/gitlab-org/gitlab/-/blob/v17.8.0-ee/app/models/integrations/matrix.rb#L7

Looked through some of the search results, others that are probably worth notifying:

* https://github.com/sitespeedio/sitespeed.io/blob/f8a6408b7c0b6734e68976dbd9db93070f8d21ea/lib/plugins/matrix/send.js#L20
* https://github.com/badges/shields/blob/152b8e9a64623ce5677c1ef3cd802f7291855611/services/matrix/matrix.service.js#L200 (apparently shields.io still defaults to guest access for non-matrix.org servers, and the guest access uses the query param)
* https://github.com/netdata/netdata/blob/fa91f42a15470c6ed1d478cbe40f89de81d2d6b8/src/health/notifications/alarm-notify.sh.in#L2099
* https://github.com/TwiN/gatus/blob/fa3e5dcc6e0138953042d0ac68c576f13e6c10ad/alerting/provider/matrix/matrix.go#L106

turt2live: libcmatrix, simplematrixbotlib, mxbt, ttm, and hemppa appear to fit a definition of 'maintained', so may be quickly fixed by the pings (hopefully).

I've tried starting a fix for gitlab, but forking the project is taking too long, so will have to be another time if no one else does it :D

tulir: GitLab seems to have some kind of system where you don't have to make your own fork https://docs.gitlab.com/ee/development/contributing/first_contribution/index.html

It automatically sent an access request without further prompting when I clicked the link in the second bullet point, so I guess I'll try making a patch too :D

agx:
> libcmatrix, simplematrixbotlib, mxbt, ttm, and hemppa appear to fit a definition of 'maintained', so may be quickly fixed by the pings (hopefully).

I'm more of a drive by contributor to libcmatrix but put out a patch here: https://source.puri.sm/Librem5/libcmatrix/-/merge_requests/104

proposals/4127-remove-query-string-auth.md
tulir: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/178475

tulir:
No need for fallbacks, the Authorization header has been supported since r0.3.0 (released in 2017) https://spec.matrix.org/legacy/client_server/r0.3.0.html#client-authentication

imbev:
Addressed in simplematrixbotlib v2.12.2 https://codeberg.org/imbev/simplematrixbotlib/commit/6522377ebcd1da58f6912cd17696a2dac6dd791e

proposals/4127-remove-query-string-auth.md
turt2live: We've run into a number of cases over the last few years where the spec tries to remove a feature, but implementations can't meaningfully do that without dropping support for old spec versions (which they don't want to do). 

I do feel it's important we figure out how we can remove things without dropping old spec version support, but I also feel like it's important to at least clean up the spec for newer implementations to start from. A future, different, proposal can address the removal concerns more broadly, in my opinion.

turt2live:
This got stuck a couple of years ago due to concerns about not actually being able to "remove" something from the spec. Mentioned in https://github.com/matrix-org/matrix-spec-proposals/pull/4127#discussion_r3243358789 though, I'd like to see us clean up the spec and (continue to) punt the removal problem down the road to a future MSC. 

I'll commit to thinking about the removal problem, but can't promise any particular timeline or MSCs will result from it.

@mscbot fcp merge

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2408

turt2live:
Merged 🎉

PR #4129, opened 2024-04-11
Tweaks to the QR code login crypto

dkasak:
- Derive separate encryption keys and nonces for the two sides of the secure channel
- <strike>Mandate that the new device commits to a particular identity key (by setting the device ID to the public part) and proves ownership of it (via a combination of ECDH and HMAC-SHA256)</strike> these have been moved to https://github.com/matrix-org/matrix-spec-proposals/pull/4130
- Spell out HKDF parameters in text too
- Misc style fixes

PR #4108, opened 2024-02-22
MSC4108: Mechanism to allow OAuth 2.0 API sign in and E2EE set up via QR code

hughns:
**SCT Stuff**

[MSC Checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4108#issuecomment-4500516017)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4108#issuecomment-4500521390)

Designated reviewers:

* [ ] @erikjohnston 
* [x] @richvdh 

----

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/element-hq/oidc-qr-login/proposals/4108-oidc-qr-login.md)

Dependencies:

- [ ] https://github.com/matrix-org/matrix-spec-proposals/pull/4388
- [x] https://github.com/matrix-org/matrix-spec-proposals/pull/4341

Video demos:

**New EX scans existing EW**
MSC4108 v2025 New EX scans existing EW

**Existing EX scans new EW**
MSC4108 v2025 Existing EX scans new EW

**New EX scans existing EX**
MSC4108 v2025 New EX scans existing EX

---

The authors are employed by Element to write this MSC.

---

## Major revisions

The is an older version of this proposal (referred to as "2024") that has some significant differences from the current proposal.

To help avoid confusion on the status, the following is hopefully helpful:

Version name|Status|Proposal revision|Supported by|Encryption scheme|Discovery of support|Proof-of-MSC implementation status
-|-|-|-|-|-|-
2025|MSC is feature complete and being reviewed [Diff to 2024 version](https://github.com/matrix-org/matrix-spec-proposals/compare/87f8317a902cd7bc5c2d2d225f71021b3a509e2d...element-hq/oidc-qr-login)|[latest](https://github.com/matrix-org/matrix-spec-proposals/blob/element-hq/oidc-qr-login/proposals/4108-oidc-qr-login.md)|No production implementations. Experimental feature in Synapse.|HPKE|Discovery using `GET /_matrix/client/v1/rendezvous`|Fully implemented (see below)
[...]

PR #4130, opened 2024-04-16
Mandate that the new device commits to a particular identity key

hughns:
Originally from https://github.com/matrix-org/matrix-spec-proposals/pull/4129, these are suggested changes to https://github.com/matrix-org/matrix-spec-proposals/pull/4108:

- Mandate that the new device commits to a particular identity key (by setting the device ID to the public part) and proves ownership of it (via a combination of ECDH and HMAC-SHA256)

Co-authored-by: Damir Jelić <poljar@termina.org.uk>

### DIFF
--- a/proposals/4108-oidc-qr-login.md
+++ b/proposals/4108-oidc-qr-login.md
@@ -375,9 +375,8 @@ Participants:
 Regardless of which device generates the QR code, either device can be the existing (already signed in) device. The
 other device is then the new device (one seeking to be signed in).
 
-Symmetric encryption uses a separate encryption key for each sender, both derived from a shared secret using HKDF. A
-separate deterministic, monotonically-incrementing nonce is used for each sender. Devices initially set both nonces to
-`0` and increment the corresponding nonce by `1` for each message sent and received.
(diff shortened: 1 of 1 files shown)


hughns:
Apologies, I meant to have this as draft as it is destined to be merged into MSC4108

hughns:
Having discussed this today with @dkasak and @poljar, the conclusion is that these changes should not be part of MSC4108.

Whilst "device ID = identity key" feels like a sensible direction to propose for the [reasons stated](https://github.com/matrix-org/matrix-spec-proposals/pull/4130/files#diff-4e390b7eed93d4f7bebdde02bf6c876ff7ad9fb3470a0a6cbf4d651c1825a85dR809-R813), the implications of such a change are beyond the scope of the present (Sign in with QR) MSC.

We haven't been able to identify a benefit to the QR mechanism by having a proof of possession prior to sharing the secrets.

There may be an overall benefit in having the homeserver/OP enforce proof of possession during auth/API calls. This may mean that the check that the existing device does to ensure the new device is online before sharing secrets would be after the homeserver/OP has also approved the device.

PR #4132, opened 2024-04-17
MSC4132: Deprecate Linking to an Event Against a Room Alias

pixlwave:
[Rendered](https://github.com/pixlwave/matrix-spec-proposals/blob/doug/event-uris/proposals/4132-deprecate-event-on-room-alias-uris.md)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4132#issuecomment-2061584975)

### DIFF
--- a/proposals/4132-deprecate-event-on-room-alias-uris.md
+++ b/proposals/4132-deprecate-event-on-room-alias-uris.md
@@ -18,16 +18,20 @@ the homeserver. In contrast, event URIs built against a room alias can break und
   ancestor.
 - The alias is removed/changed resulting in a dead link that can't be resolved.
 
-The proposal would deprecate the 2 following URIs:
+The proposal would [deprecate](https://spec.matrix.org/v1.10/#deprecation-policy) the 2 following URIs:
 - Link to `$event` in `#somewhere:example.org`: `matrix:r/somewhere:example.org/e/event`
 - Link to `$event` in `#somewhere:example.org`: `https://matrix.to/#/%23somewhere:example.org/%24event%3Aexample.org`
+
+Specifically this means it would be marked in the spec as not to be used in future and clients should:
+1. Stop making new event URIs with room aliases.
+2. Continue to accept them as it will take time for other clients to update and legacy events will continue to exist.
 
 
 ## Potential issues
 
 Whilst most clients do take the sensible route and generate event URIs against a room ID (even when it has an alias), as
-it is part of the spec these kinds of links likely exist somewhere in room history. This would mean that after
-deprecation clients may still want to handle these URIs when tapped.
+it is part of the spec these kinds of links likely exist somewhere in room history and on the World Wide Web. This would
+mean that after deprecation clients may still want to handle these URIs when received.
 
 
 ## Alternatives
@@ -37,9 +41,7 @@ An alternative would be to leave things as they are, although this wouldn't be t
 
 ## Security considerations
 
-It is unlikely that accepting this change would introduce any security considerations. If anything, it may improve the
-hypothetical situation where a deep link could be being redirected to modified content by moving an existing room alias
-to a malicious room.
+It is unlikely that accepting this change would introduce any security considerations.
 
 
 ## Dependencies
--- a/proposals/4132-deprecate-event-on-room-alias-uris.md
+++ b/proposals/4132-deprecate-event-on-room-alias-uris.md
@@ -40,7 +40,7 @@ mean that after deprecation clients may still want to handle these URIs when rec
 
 Additionally, whilst not a new problem some clients don't always update the `via` fields, which poses a problem for
 rooms that have lots of single user homeservers: A single user leaving can result in the `via` to become non functional
-(which is essentially the equivalent of the alias being removed) and over time this can result in unreachable room IDs.
+which over time can result in unreachable room IDs (this is essentially the equivalent of the alias being removed).
 
 
 ## Alternatives
--- a/proposals/4132-deprecate-event-on-room-alias-uris.md
+++ b/proposals/4132-deprecate-event-on-room-alias-uris.md
@@ -2,8 +2,8 @@
 
 ## Introduction
 
-The spec supports 2 types of URI, the [Matrix scheme](https://spec.matrix.org/unstable/appendices/#matrix-uri-scheme)
-and [matrix.to](https://spec.matrix.org/unstable/appendices/#matrixto-navigation) links which allow deep linking to
+The spec supports 2 types of URI, the [Matrix scheme](https://spec.matrix.org/v1.10/appendices/#matrix-uri-scheme)
+and [matrix.to](https://spec.matrix.org/v1.10/appendices/#matrixto-navigation) links which allow deep linking to
 Matrix users, rooms and events. Event URIs can be constructed against either a room ID or a room alias the latter of
 which is fragile issue as a room's alias is mutable and so event links may be broken by changes to the alias. Users
 expect deep links to continue working so that older links continue to show the correct content. Therefore it is proposed
--- a/proposals/4132-deprecate-event-on-room-alias-uris.md
+++ b/proposals/4132-deprecate-event-on-room-alias-uris.md
@@ -23,8 +23,13 @@ The proposal would [deprecate](https://spec.matrix.org/v1.10/#deprecation-policy
 - Link to `$event` in `#somewhere:example.org`: `https://matrix.to/#/%23somewhere:example.org/%24event%3Aexample.org`
 
 Specifically this means it would be marked in the spec as not to be used in future and clients should:
-1. Stop making new event URIs with room aliases.
+1. Stop making new event URIs with room aliases, generating the examples above as the following respectively:
+    - `matrix:roomid/opaqueid:example.org/e/event?via=example.com`
+    - `https://matrix.to/#/!opaqueid%3Aexample.org/%24event%3Aexample.org?via=example.com`
 2. Continue to accept them as it will take time for other clients to update and legacy events will continue to exist.
+
+It is worth nothing that this change would require clients to always add the `via=` fields when building event URIs as
+these are expected for URIs that reference a room ID.
 
 
 ## Potential issues
@@ -32,6 +37,10 @@ Specifically this means it would be marked in the spec as not to be used in futu
 Whilst most clients do take the sensible route and generate event URIs against a room ID (even when it has an alias), as
 it is part of the spec these kinds of links likely exist somewhere in room history and on the World Wide Web. This would
 mean that after deprecation clients may still want to handle these URIs when received.
+
+Additionally, whilst not a new problem some clients don't always update the `via` fields, which poses a problem for
+rooms that have lots of single user homeservers: A single user leaving can result in the `via` to become non functional
+(which is essentially the equivalent of the alias being removed) and over time this can result in unreachable room IDs.
 
 
 ## Alternatives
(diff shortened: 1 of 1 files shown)


proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,47 @@
...
+The proposal would deprecate the 2 following URIs:
+- Link to `$event` in `#somewhere:example.org`: `matrix:r/somewhere:example.org/e/event`
+- Link to `$event` in `#somewhere:example.org`: `https://matrix.to/#/%23somewhere:example.org/%24event%3Aexample.org`
+
+
+## Potential issues
+
+Whilst most clients do take the sensible route and generate event URIs against a room ID (even when it has an alias), as
+it is part of the spec these kinds of links likely exist somewhere in room history. This would mean that after
+deprecation clients may still want to handle these URIs when tapped.
```
clokep, line 30: I'm not really sure what "tapped" means here.

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Alternatives
+
+An alternative would be to leave things as they are, although this wouldn't be the best choice for the user.
+
+
+## Security considerations
+
+It is unlikely that accepting this change would introduce any security considerations. If anything, it may improve the
+hypothetical situation where a deep link could be being redirected to modified content by moving an existing room alias
```
clokep, line 41: Event IDs are globally unique so I don't *think* there's a real security hole here, but 100% agree on the UX win.

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Proposal
+
+As room IDs are immutable, event URIs built against these will continue to work for as long as the room is reachable by
+the homeserver. In contrast, event URIs built against a room alias can break under the following circumstances:
+- The room is upgraded, resulting in the alias pointing to the new room, which won't contain any events from the
+  ancestor.
+- The alias is removed/changed resulting in a dead link that can't be resolved.
+
+The proposal would deprecate the 2 following URIs:
```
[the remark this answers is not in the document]
ara4n: ...and mark it in the spec as not to be used in future.

In terms of whether we can yank it out entirely: has anyone *ever* seen one of these in the wild? I'm searching the matrix.org DB for them, but anecdotally i don't think i've ever seen a client which created them

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,47 @@
...
+- Link to `$event` in `#somewhere:example.org`: `matrix:r/somewhere:example.org/e/event`
+- Link to `$event` in `#somewhere:example.org`: `https://matrix.to/#/%23somewhere:example.org/%24event%3Aexample.org`
+
+
+## Potential issues
+
+Whilst most clients do take the sensible route and generate event URIs against a room ID (even when it has an alias), as
+it is part of the spec these kinds of links likely exist somewhere in room history. This would mean that after
```
bmarty, line 29: `room history` yes but maybe also `World Wide Web` as a larger place to find matrix.to links :)

proposals/4132-deprecate-event-on-room-alias-uris.md
clokep: Ah-ha! I suspect "received" might be more general!

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,47 @@
+# MSC4132: Deprecate Linking to an Event Against a Room Alias.
+
+## Introduction
+
+The spec supports 2 types of URI, the [Matrix scheme](https://spec.matrix.org/unstable/appendices/#matrix-uri-scheme)
+and [matrix.to](https://spec.matrix.org/unstable/appendices/#matrixto-navigation) links which allow deep linking to
```
turt2live, line 6:
```suggestion
The spec supports 2 types of URI, the [Matrix scheme](https://spec.matrix.org/v1.10/appendices/#matrix-uri-scheme)
and [matrix.to](https://spec.matrix.org/v1.10/appendices/#matrixto-navigation) links which allow deep linking to
```

(the unstable links may break without notice, and after the MSC is merged would point to new info instead of old)

proposals/4132-deprecate-event-on-room-alias-uris.md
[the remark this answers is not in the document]
turt2live: I've certainly seen them, but they were manually generated rather than by a client. Best practice appears to already be to not generate such links.

I don't think we can safely jump to removal, though. https://github.com/matrix-org/matrix-spec-proposals/pull/4127 is an example of another trivial thing which ended up wanting a second MSC anyways.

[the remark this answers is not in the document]
turt2live: re: clarifying what deprecation means - it'd be good to also link to the deprecation policy: https://spec.matrix.org/v1.10/#deprecation-policy

proposals/4132-deprecate-event-on-room-alias-uris.md
turt2live, line 1: Implementation requirements:

None, in my opinion. This MSC deprecates a feature and already appears to be unimplemented (for the most part).

turt2live:
... and because I haven't sent enough comments in the last few minutes: the deprecation requirements can be clarified concurrent to checkbox gathering, imo. I see no reason to delay this MSC.

@mscbot fcp merge

pixlwave:
> ... and because I haven't sent enough comments in the last few minutes: the deprecation requirements can be clarified concurrent to checkbox gathering, imo. I see no reason to delay this MSC.

Just for clarification as I haven't done this before, I think this is saying its fine to update the proposal based on the comments whilst SCT members are looking at it?

proposals/4132-deprecate-event-on-room-alias-uris.md
pixlwave: Are you telling me that messaging doesn't always revolve around a touch screen?! 😅

proposals/4132-deprecate-event-on-room-alias-uris.md
pixlwave: Oh yeah! 😅

proposals/4132-deprecate-event-on-room-alias-uris.md
pixlwave: TIL they're globally unique.

pixlwave:
Updated taking all of the comments above into consideration.

turt2live:
> > ... and because I haven't sent enough comments in the last few minutes: the deprecation requirements can be clarified concurrent to checkbox gathering, imo. I see no reason to delay this MSC.
> 
> Just for clarification as I haven't done this before, I think this is saying its fine to update the proposal based on the comments whilst SCT members are looking at it?

Yes, sorry. The proposal can be updated in all kinds of ways up until Final Comment Period (FCP) actually starts - when 8/10 people check their boxes and no concerns are formally raised. Things which SCT members believe should prevent acceptance of the MSC are raised via `@mscbot concern`. Everything else is non-blocking, though may still feel important to address.

turt2live:
(though please avoid force pushing, as it makes review harder. We squash all the commits down to a single commit upon acceptance)

proposals/4132-deprecate-event-on-room-alias-uris.md
KitsuneRal: There was a somewhat contentious discussion in the community whether we should assume them to remain globally unique and not scoped to rooms, for technical reasons ("sharding" the event id space, effectively). That is not relevant to the issue discussed here though: an event id is essentially a hash of certain parts of event JSON, therefore unique by design, barring hash collisions.

pixlwave:
My bad, good to know you squash 👍

proposals/4132-deprecate-event-on-room-alias-uris.md
clokep: Yeah -- I don't think we need any change here.

proposals/4132-deprecate-event-on-room-alias-uris.md
Kladki: Just pointing out that it seems that Nheko does this as of 0.11.3, which is currently the latest release.

dbkr:
@mscbot concern outstanding comments

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,49 @@
...
+- Link to `$event` in `#somewhere:example.org`: `https://matrix.to/#/%23somewhere:example.org/%24event%3Aexample.org`
+
+Specifically this means it would be marked in the spec as not to be used in future and clients should:
+1. Stop making new event URIs with room aliases.
+2. Continue to accept them as it will take time for other clients to update and legacy events will continue to exist.
+
+
+## Potential issues
+
+Whilst most clients do take the sensible route and generate event URIs against a room ID (even when it has an alias), as
```
MTRNord, line 32: While this isnt a new Problem it might be worth mentioning again: clients already for spaces dont always Update the vias leading to unstable or unreachable ids after a while as some rooms have lots of single User homeservers. In which case a user leaving can result in the via to become non functional. This is the equivalent of the alias being removed basically.

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,49 @@
...
+
+The spec supports 2 types of URI, the [Matrix scheme](https://spec.matrix.org/v1.10/appendices/#matrix-uri-scheme)
+and [matrix.to](https://spec.matrix.org/v1.10/appendices/#matrixto-navigation) links which allow deep linking to
+Matrix users, rooms and events. Event URIs can be constructed against either a room ID or a room alias the latter of
+which is fragile issue as a room's alias is mutable and so event links may be broken by changes to the alias. Users
+expect deep links to continue working so that older links continue to show the correct content. Therefore it is proposed
+to deprecate event links against a room alias.
+
+
+## Proposal
```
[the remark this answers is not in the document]
pixlwave: Updated.

proposals/4132-deprecate-event-on-room-alias-uris.md
```diff
@@ -0,0 +1,49 @@
...
+
+As room IDs are immutable, event URIs built against these will continue to work for as long as the room is reachable by
+the homeserver. In contrast, event URIs built against a room alias can break under the following circumstances:
+- The room is upgraded, resulting in the alias pointing to the new room, which won't contain any events from the
+  ancestor.
+- The alias is removed/changed resulting in a dead link that can't be resolved.
+
+The proposal would [deprecate](https://spec.matrix.org/v1.10/#deprecation-policy) the 2 following URIs:
+- Link to `$event` in `#somewhere:example.org`: `matrix:r/somewhere:example.org/e/event`
```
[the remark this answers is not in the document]
pixlwave: Oh yeah, good point, I've added these.

One small thing I noticed: The examples I have given are from the spec, which interestingly don't the `:` in the room alias of the matrix.to event URI (but does for everything else). Probably not worth worrying about but mentioning it in case I'm missing something here.

proposals/4132-deprecate-event-on-room-alias-uris.md
pixlwave: Have added something to cover this, I think it captures everything you've mentioned.

ara4n:
ironically, i just found a permalink indexed by room alias in the wild: https://github.com/element-hq/element-web/issues/2150#issue-176068929

richvdh:
merged! :tada:

PR #4136, opened 2024-04-26
MSC4136: Shared retry hints

ara4n:


proposals/4136-shared-retry-hints.md
```diff
@@ -0,0 +1,104 @@
...
+it.  Similarly the recommendation is to wait 1 week before retrying element.io if it's down.
+
+The spec currently does not specify anything about how federation retry schedules: as part of this change, we propose
+explicitly adding that:
+
+ * Servers SHOULD follow exponential or geometric backoff schedules (and MUST NOT retry linearly)
+ * Servers SHOULD reset their retry schedule for a given domain if they receive traffic from that domain, and immediately retry.
+ * On joining a room, servers SHOULD attempt to connect to all newfound servers (sending a no-op transaction if there are
+   no other events to be sent), but queued in such a way to prioritise servers with lower `retry_after` value first,
+   ensuring alive servers are prioritised over likely dead servers.
```
richvdh, line 59: I fear this "prioritise servers with lower retry_after" bit is non-trivial to implement, and, if you're saying we have to attempt connect to all servers anyway, probably doesn't bring that much value? We're still going to max out our CPU for 5 minutes attempting to connect to a bunch of dead servers; whether that happens before or after sending our first message to the "recently good" servers doesn't seem likely to make much difference to me.

proposals/4136-shared-retry-hints.md
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+```
+
+This means that the joined room has three participating servers, but `example.com` is not reliably responding over
+federation, and the resident server recommends that the joining server should wait 1 hour before retrying to connect to
+it.  Similarly the recommendation is to wait 1 week before retrying element.io if it's down.
+
+The spec currently does not specify anything about how federation retry schedules: as part of this change, we propose
+explicitly adding that:
```
Kladki, line 53: Would whether servers persist backoffs between restarts be up to the implementation, or would you like to specify that as well?

proposals/4136-shared-retry-hints.md
```diff
@@ -0,0 +1,104 @@
...
+## Problem
+
+New Matrix servers currently have no idea which other servers are online or not, and so start a retry schedule from
+first principles.
+
+This is far from ideal, as joining a room with many participating servers (e.g. 10K in #matrix:matrix.org right now) is
+incredibly heavy on the connecting server as the newly joined server will have to make 10K connection attempts as
+rapidly as possible (both SRV and .well-known lookups, and then /send attempts) to identify alive servers.  Meanwhile,
+dead servers (e.g. domains which no longer run Matrix servers) will be
+[hammered](https://mastodon.matrix.org/@mnot@techpolicy.social/112319234007365786) by connection attempts.
```
clokep, line 12: This sounds like the real problem, not the load on new servers. Is an alternative to have a way to automatically remove people that are on dead servers? Maybe using MSC4049?

proposals/4136-shared-retry-hints.md
turt2live, line 53: it's unclear to me that this is a required measure given the problem statement. Why do we need to strongly specify retry behaviour instead of adding implementation guidance to the existing spec?

proposals/4136-shared-retry-hints.md
turt2live, line 1: Implementation requirements:

* [ ] Server producing values
* [ ] Server consuming values, and making decisions based on those values
* [ ] Evidence of the rate limit strategy being applied as described

proposals/4136-shared-retry-hints.md
TheArcaneBrony: I suppose it would lower the mean (noticeable) federation latency as "dead servers" are only tried last. For synapse, wouldn't this be a matter of `SELECT * FROM destinations ORDER BY retry_after ASC;`? (or using a subquery with the same effect)

PR #4139, opened 2024-05-06
MSC: Bot buttons & conversations

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/t2bot/bot-buttons/proposals/4139-bot-buttons.md)

I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published as operator/developer at t2bot.io.

----

Fixes https://github.com/matrix-org/matrix-spec/issues/232
See also https://github.com/matrix-org/matrix-spec/issues/1651


proposals/4139-bot-buttons.md
turt2live, line 1: Implementation requirements:

* [ ] Client rendering prompts
* [ ] Bot sending prompts

Note that there is a prerequisite on Extensible Events which needs to be unpicked.

proposals/4139-bot-buttons.md
```diff
@@ -0,0 +1,201 @@
...
+        "@alice:example.org",
+        "@bob:example.org",
+      ],
+      // These are the options a user has. Note the 2 distinct types and 3 label approaches.
+      "prompts": [
+        {
+          // `type` is the prompt type: "preset" (show a button) or "input" (shown below)
+          "type": "preset",
+          // `id` is used by the bot to figure out what prompt the user picked. It is an opaque ID.
+          "id": "1d6",
```
TheArcaneBrony, line 69: This seems to indicate that interactions are stateful?

I'd been thinking about a similar-ish system, but making the client just send a `m.interaction` event with a bot-defined content (possibly as a sub-key?)

eg. a bot would define something like 
```json
// other content stuffs ...
"interactions": [
  {
    "type": "command",
    "metadata": {
      "action": "roll",
      "dice_type": "1d6"
    }
]
```

When a user clicks this command button, the client sends an event:
```json
{
  "type": "m.interaction",
  "content": {
     "metadata": {
        "action": "roll",
        "dice_type": "1d6"
     }, 
     // reply stuff
  }
}
```

I think, personally, I'd love to see a system that sticks more to existing interaction systems in order to avoid extra maintenance burden, and to make it easier to onboard clients.

proposals/4139-bot-buttons.md
```diff
@@ -0,0 +1,201 @@
...
+SHOULD be used for `preset` prompts, using the provided `label`, and text inputs with `label` as a
+prefix or placeholder, and validation per `validator`, SHOULD be used for `input` prompts. For
+example:
+
+![](./images/4139-01-dice-bot-welcome.png)
+
+[Codepen](https://codepen.io/turt2live/pen/gOyVvaY) (note: doesn't do validation)
+
+The user is then able to click on one of the buttons or submit text through the `input` option. That
+reply looks as follows:
```
Gnuxie, line 129: What about if there was just a generic button type that if you clicked, would send an event?

Even better, why can't *other* things be clicked as a conversation reply (or an argument to a slash command), such as another user,, or another event in the room timeline, or another button from another event in the timeline.

Being able to select another button in the timeline as a conversation reply or command argument would unlock a huge door, bots would be able to represent output as different button types. For example, Draupnir would be able to render each individual policy in the response to the equivalent of `!rules matching @spam:example.com` as a button, that could be selected as an argument in future commands or interactions.

turt2live: All of this sounds wonderful and very much needed indeed, though the scope for the MSC needs to end somewhere. Expansion is possible through other MSCs currently, and as this one progresses maybe it brings concepts into it.

proposals/4139-bot-buttons.md
turt2live: Bots can GET the previous event if they don't keep state themselves.

proposals/4139-bot-buttons.md
Gnuxie: Well, it's my opinion that this would reduce the scope of this MSC and future ones by being able to express them using the "button type primitive" (alternatively presentation type). But I do understand how that can be seen differently.

turt2live: I'm not sure I'm following the suggestion in that case, sorry. I understood your comment to mean supporting different types of input, not reusing buttons for everything (which I'm not sure how that'd even work).

proposals/4139-bot-buttons.md
turt2live, line 1: https://github.com/matrix-org/matrix-spec-proposals/pull/4332 is some more modern thinking in the area. It may be interesting to leverage the ideas from 4332 in this proposal, if not wholesale adopt it.

turt2live: ie: have buttons be commands

PR #3767, opened 2022-04-07
MSC3677: Time based notification filtering

kerryarchibald:

PR #3768, opened 2022-04-08
Add in-app notification action proposal

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/notify-in-app/proposals/3768-notify-in-app.md)

PR #4141, opened 2024-05-08
Time based notification filtering

hanadi92:
Took on this old MSC https://github.com/matrix-org/matrix-spec-proposals/pull/3767 on time based notification filtering (do not disturb) and only added a detailed spec on the day-of-week numbering schema.

p.s. I kept the commits from the original author and just added mine on top.

Signed-off-by: @hanadi92

### DIFF
--- a/proposals/xxxx-time-based-notification-filtering.md
+++ b/proposals/xxxx-time-based-notification-filtering.md
@@ -1,4 +1,4 @@
-# MSC3767: Time based notification filtering
+# MSC3XXXX: Time based notification filtering
 Do not disturb / focus features are becoming standard across operating systems and networking products. Users expect to
 be able to manage the level of noisiness from an application based on day and time.
 
@@ -21,10 +21,10 @@ to filter based on time of day and day of week.
 
 **Time matching interval definition**
 
-| key | type | value | description | Required |
-| ---- | ----| ----- | ----------- | -------- |
+| key | type | value                 | description                                                                                                                                                                                                                                             | Required |
+| ---- | ----|-----------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| -------- |
 | `time_of_day` | string[] | tuple of `hh:mm` time | Tuple representing start and end of a time interval in which the rule should match. Times are [ISO 8601 formatted times](https://en.wikipedia.org/wiki/ISO_8601#:~:text=As%20of%20ISO%208601%2D1,minute%20between%2000%20and%2059.). Times are inclusive | Optional. When omitted all times are matched.  |
-| `day_of_week` | number[] | array of integers 0-7 | An array of integers representing days of the week on which the rule should match, where 0 = Sunday, 1 = Monday, 7 = Sunday | **Required** |
+| `day_of_week` | number[] | array of integers 0-6 | An array of integers representing days of the week on which the rule should match, where 0 = Sunday, 1 = Monday, .., 6 = Saturday                                                                                                                       | **Required** |
 
 
 - `time_of_day` condition is met when the server's timezone-adjusted time is between the values of the tuple, or when no
@@ -49,19 +49,20 @@ to true.
         },
         {
             // no time_of_day, all day is matched
-            "day_of_week": [6, 7] // Weekend
+            "day_of_week": [0, 6] // Weekend
         }
-},
+    ]
+}
 ```
 
-A primary usecase for this condition is creating 'do not disturb' behaviour.
+A primary use case for this condition is creating 'do not disturb' behaviour.
 For example, Wednesday morning focus time rule:
 ```json5
 {
     "rule_id": ".m.rule.master",
     "default": true,
     "enabled": true,
-    "conditions": [
+    "conditions": {
         "kind": "time_and_day",
         "timezone": "Europe/Berlin",
         "intervals": [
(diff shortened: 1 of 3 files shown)


proposals/xxxx-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
+# MSC3XXXX: Time based notification filtering
```
turt2live, line 1:
```suggestion
# MSC4141: Time based notification filtering
```

proposals/xxxx-time-based-notification-filtering.md
turt2live, line 1: Implementation requirements:

* [ ] Client exposing UI for these options
* [ ] Server supporting these options
* [ ] Users demonstrating that this feature works for them (lack of bug reports, significant complaints, etc)

proposals/xxxx-time-based-notification-filtering.md
turt2live: please also rename the file accordingly

proposals/xxxx-time-based-notification-filtering.md
hanadi92: server implementation on Synapse is already open and ready for review here https://github.com/element-hq/synapse/pull/16858

proposals/4141-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+##### `dont_notify` and Do Not Disturb behaviour
+`dont_notify` will stop badges from being
+updated in the client during 'do not disturb' hours, so the user will not be
+able to locate messages that were silenced when they are back online.
+To silence push notifications but allow discovery of missed messages in app, `notify_in_app` as proposed in
+[MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768) should
+be used.
```
sumnerevans, line 87: I rely on seeing the red dots on chats that I have been mentioned in overnight in order to figure out what I need to address every morning, so I think that this is a significant enough UX problem to require blocking this MSC on [MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768).

proposals/4141-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Potential issues
+- If a user changes timezone their push rules will not automatically update.
```
sumnerevans, line 90: As someone who travels often, I think that this is a major concern. At the very least, we need to add a recommendation that clients either automatically change timezone when they enter a new timezone, or prompt the user to manually change it.

But, as mentioned I think that this functionality should be implemented by the operating system the client is running on, as it has full timezone and DND state information.

proposals/4141-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Potential issues
+- If a user changes timezone their push rules will not automatically update. 
+
+## Alternatives
+
+#### System
+Some systems (e.g. iOS) have their own DND / focus mode but this is only an option if all of your devices are within
+that vendor ecosystem (here Apple) and doesn't help when you have e.g. an iPad and an Android phone. This also needs to
+be configured per device.
```
sumnerevans, line 97: Personally, I think that this entire functionality should be handled by the operating system of the client. Fundamentally, the client is where the necessary information (namely DND status) resides.

I can think of many edge cases that can cause a server-driven approach as suggested in this MSC to fail:

* **Travel:** as already mentioned, users will be confused when the notification rules are off by whatever timezone offset they've traveled to.
* **Overridden DND off:** if you are up late (maybe pulling an all-nighter to study for a test) and want to have your notifications on all night (maybe classmates texting study tips), you would want to be able to turn of the  DND and have notifications show up.

Clients should just spit notifications out and allow the operating system to hide/show them depending on whatever rules are configured on the OS level. This will ensure that we can take advantage of whatever notification management features are provided by operating systems both now and in the future.

Obviously, it is annoying to have to configure it on multiple ecosystems, but I think this MSC actually makes it worse as you would have to configure your Matrix client DND settings in addition to your OS DND settings.

sumnerevans, line 97: It appears that per-device rules are going away: https://github.com/matrix-org/matrix-spec/pull/1842

proposals/4141-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
...
+This rule would be assessed only on the device. This is not easily achieved on every platform. 
+
+#### ms offsets for time intervals
+Previous proposals used ms offsets to represent time of day instead of `hh:mm`. Ms offsets may behave incorrectly on
+days with a DST jump and are less intuitive.
+
+## Security considerations
+- Stores user's timezone on the server. DND periods saved to the server without timezone information would reveal
+  information about a user's approximate timezone anyway. Users who do not wish to store their timezone can set DND
+  periods in UTC (this option should be available in clients implementing time based notification filtering).
```
sumnerevans, line 129: I think that forcing all times to be expressed in UTC time should be listed as an alternative. Since users will not be interacting directly with the state event, I think clients should always translate the user's setting into UTC time.

proposals/4141-time-based-notification-filtering.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Unstable prefix
+
+- While this MSC is not considered stable `time_and_day` should be referred to as `org.matrix.msc4141.time_and_day`
+
+## Dependencies
+None.
```
sumnerevans, line 136: As mentioned in another comment, I think [MSC3768](https://github.com/matrix-org/matrix-spec-proposals/pull/3768) should be a dependency.

sumnerevans (commented):
I'm generally opposed to this proposal. I think that the functionality proposed here should be left up to the operating system the client is running on as the OS is the source of truth for timezone as well as DND status.

PR #2781, opened 2020-09-18
MSC2781: Remove the reply fallbacks from the specification

deepbluev7:
Since I hit another fallback bug today, I thought I should finally propose this. Let's see, how this goes.

[Rendered](https://github.com/deepbluev7/matrix-doc/blob/drop-the-fallbacks/proposals/2781-down-with-the-fallbacks.md)

Implementations:
* [react-sdk](https://github.com/matrix-org/matrix-react-sdk/pull/6964)

See also:
* matrix-org/matrix-spec-proposals#3644 

fixes matrix-org/matrix-spec#368

fixes matrix-org/matrix-spec#350 ?

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>

---

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/2781#issuecomment-2108904800)

PR #4144, opened 2024-05-10
MSC4144: Per-message profiles

tulir:
[Rendered](https://github.com/beeper/matrix-spec-proposals/blob/per-message-profile/proposals/4144-per-message-profile.md)

Implementations:

* Clients rendering profiles:
  * gomuks web: https://github.com/tulir/gomuks/pull/566
  * Sable: https://github.com/SableClient/Sable/pull/309
  * Cinny: https://github.com/nexy7574/cinny/commit/38454be6a0bec99328329e96ede9de4708dcba8a
* Bots sending profiles:
  * Linear maubot: https://github.com/beeper/linear-maubot/commit/88dfdd4ffa9c58ec4961e6351761997e5a2a930d
  * GitHub maubot: https://github.com/maubot/github/commit/c6d768b2ec3657779e927646ad1f872d624f34df
  * mautrix-discord: https://github.com/mautrix/discord/blob/v0.7.3/portal_convert.go#L360-L373
  * mautrix-slack: https://github.com/mautrix/slack/blame/v0.2.1/pkg/msgconv/from-slack.go#L92-L100
  * mautrix-irc: https://github.com/mautrix/irc/commit/a506e9b1459e3dbe4031ce4338204d0d7b664f9c
  * any bridge written using mautrix-go/bridgev2: https://github.com/mautrix/go/pull/484

Signed-off-by: Tulir Asokan <tulir@beeper.com>

### DIFF
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -1,0 +1,123 @@
+# MSC4144: Per-message profiles
+Currently profiles in Matrix are defined by `m.room.member` state events, and
+there is no easy way to have different profiles per message.
+
+## Proposal
+The proposed solution is a new field called `m.per_message_profile`, which
+contains a displayname and/or avatar URL to override the default profile,
+plus an ID to identify different profiles within the same Matrix user.
+
+```json
+{
+  "msgtype": "m.text",
+  "body": "Hello, World!",
+  "m.per_message_profile": {
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo"
+  }
+}
+```
+
+The `id` field is required and is an opaque string. Clients may use it to group
+messages with the same ID like they would group messages from the same sender.
+For example, bridges would likely set it to the immutable remote user ID.
+
+### Encrypted avatars
+Because the profile is inside the ciphertext in encrypted events, the entire
+profile can be hidden the server, as long as the avatar is also encrypted.
+Encrypted avatars are placed under `avatar_file` instead of `avatar_url`.
+The `avatar_file` field has the same schema as the `file` field in
+`m.room.message` events.
+
+<details>
+<summary>Encrypted avatar example</summary>
+
+```json
+{
+  "msgtype": "m.text",
+  "body": "Hello, World!",
+  "m.per_message_profile": {
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_file": {
+      "v": "v2",
+      "key": {
+        "alg": "A256CTR",
+        "ext": true,
+        "k": "8dXeBMBMthuXGY5zmUh9Mi0aqC1kndMZ4NCa-0RhELc",
+        "key_ops": [
+          "encrypt",
+          "decrypt"
+        ],
+        "kty": "oct"
+      },
+      "iv": "L6zup2cR570AAAAAAAAAAA",
+      "hashes": {
+        "sha256": "/cTs+PajUcznbV3h1w5gh1AHnLjrKQVl2jU3xLCqoBI"
+      },
+      "url": "mxc://maunium.net/eKLhozQduElYSgBkWjtwSXoi"
+    }
+  }
+}
+```
+
+</details>
+
+### Behavior of omitted and empty fields
+If the `displayname` field is omitted, null, or an empty string, the
+displayname from the member event should be used instead. Setting an empty
+displayname using a per-message profile is not supported, as there aren't any
+clear use cases for it.
+
+However, there are use cases for setting an empty avatar, so `avatar_url` being
+an empty string should be treated as clearing the avatar and falling back to
+the client's default blank avatar behavior (e.g. generating one based on the
+displayname). If both `avatar_url` and `avatar_file` are omitted or null, the
+avatar from the member event should be used instead.
+
+### Extensible profiles
+This MSC is not related to extensible profiles and does not attempt to
+implement them. However, in case extensible profiles are implemented as
+something that can be referenced (e.g. room IDs), the MSC adding them could
+allow per-message profiles to specify which extensible profile is used.
+
+## Use cases
+
+### Bridging
+Per-message profiles will allow making "lighter-weight" bridges that don't need
+appservice access. Currently the only option for such bridges is to prepend the
+displayname to the message, which is extremely ugly.
+
+Such bridges would obviously have downsides, like not being able to start chats
+via standard mechanisms, and not being able to see the member list on Matrix.
+However, those may be acceptable compromises for non-puppeting bridges that
+only operate in specific predetermined rooms.
+
+### Feature-parity with other platforms
+Other chat applications such as Slack and Discord have "webhooks" which allow
+per-message profile overrides. This MSC effectively enables the same on Matrix.
+
+### Roleplaying, plural users, etc
+Some users want to be able to switch between profiles quickly, which would be
+much easier using this MSC. Currently easiest way is to have multiple accounts,
+which has other benefits, but is much more cumbersome to manage.
+
+## Potential issues
+Implementing encrypted avatars could cause difficulty for clients that assume
+that avatars are always unencrypted mxc URIs.
+
+## Alternatives
+None.
+
+## Security considerations
+
+### Preventing impersonation
+To prevent impersonation using per-message profiles, clients should somehow
+indicate to the user that the message has a per-message profile with an easy
+way to see the user's MXID or default profile. For example, a client could have
+a small `via @user:example.com` text next to the per-message displayname.
+
+## Unstable prefix
+`com.beeper.per_message_profile` should be used instead of `m.per_message_profile`
+until this MSC is accepted.
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -138,10 +138,17 @@ unnecessarily.
 ## Security considerations
 
 ### Preventing impersonation
-To prevent impersonation using per-message profiles, clients should somehow
+To prevent impersonation using per-message profiles, clients MUST somehow
 indicate to the user that the message has a per-message profile with an easy
 way to see the user's MXID or default profile. For example, a client could have
 a small `via @user:example.com` text next to the per-message displayname.
+
+To improve user experience, clients MAY omit the indicator when the sender
+account has sufficiently high power level, and the displayname is unique among
+members of the room (i.e. it does not require disambiguation in the
+["Calculating the display name for a user" spec](https://spec.matrix.org/v1.12/client-server-api/#calculating-the-display-name-for-a-user)).
+
+TODO: define the power level
 
 ## Unstable prefix
 `com.beeper.per_message_profile` should be used instead of `m.per_message_profile`
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -158,7 +158,10 @@ account has sufficiently high power level, and the displayname is unique among
 members of the room (i.e. it does not require disambiguation in the
 ["Calculating the display name for a user" spec](https://spec.matrix.org/v1.12/client-server-api/#calculating-the-display-name-for-a-user)).
 
-TODO: define the power level
+TODO: define the power level  
+TODO2: what if the same name is used by multiple different bridges using
+per-message profiles? Maybe it's enough to recommend configuring bridges to
+append a network identifier to names?
 
 ## Unstable prefix
 `com.beeper.per_message_profile` should be used instead of `m.per_message_profile`
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -23,6 +23,11 @@ The `id` field is required and is an opaque string. Clients may use it to group
 messages with the same ID like they would group messages from the same sender.
 For example, bridges would likely set it to the immutable remote user ID.
 
+This ID is scoped to the real MXID used to send the event, but is otherwise
+global. In other words, the same ID in different rooms from the same account
+can be considered to be the same user, but the same ID in the same room from
+different accounts are considered to be different users.
+
 ### Encrypted avatars
 Because the profile is inside the ciphertext in encrypted events, the entire
 profile can be hidden from the server, as long as the avatar is also encrypted.
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -28,6 +28,10 @@ global. In other words, the same ID in different rooms from the same account
 can be considered to be the same user, but the same ID in the same room from
 different accounts are considered to be different users.
 
+The field is allowed for all message events, which currently means
+`m.room.message` and `m.sticker`. A future MSC may expand the field to other
+events, such as reactions.
+
 ### Encrypted avatars
 Because the profile is inside the ciphertext in encrypted events, the entire
 profile can be hidden from the server, as long as the avatar is also encrypted.
@@ -102,7 +106,8 @@ there's demand for bridging without requiring server admin access.
 Such bridges would obviously have downsides, like not being able to start chats
 via standard mechanisms, and not being able to see the member list on Matrix.
 However, those may be acceptable compromises for non-puppeting bridges that
-only operate in specific predetermined rooms.
+only operate in specific predetermined rooms. Non-message events like reactions
+are also not supported by this MSC, but they could be allowed in the future.
 
 This method also allows encrypting profile info, which reduces metadata leaked
 by bridging.
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -94,6 +94,58 @@ implement them. However, in case extensible profiles are implemented as
 something that can be referenced (e.g. room IDs), the MSC adding them could
 allow per-message profiles to specify which extensible profile is used.
 
+### Fallbacks for old clients
+In order for users on old clients to see the per-message profile data, sending
+clients SHOULD include a fallback containing the per-message displayname.
+When a fallback is used, the per-message profile object MUST include
+`"has_fallback": true`.
+
+#### Plaintext fallback
+The fallback MUST be at the beginning of the plaintext `body` and consist of
+the exact value of the `displayname` property, followed by a colon (`:`) and
+a space (`\x20`), e.g. `cat: original message`.
+
+To remove the plaintext fallback, trim the prefix `{displayname}: ` from the body.
+
+#### HTML fallback
+If a `formatted_body` of type `org.matrix.custom.html` is present, it SHOULD
+include the same fallback text inside a `strong` tag with the `data-mx-profile-fallback`
+attribute, e.g. `<strong data-mx-profile-fallback>cat: </strong>original message`.
+
+The displayname MUST be HTML-escaped and there MUST NOT be any HTML tags inside
+the fallback. The tag MUST NOT have any other attributes. The attribute MAY
+have an empty string as its value (`=""`) for compatibility with all HTML
+generators, but MUST NOT have any other value. The HTML fallback is not mandated
+to be at the beginning of the string, as there may be valid reasons to put it
+inside another tag, such as a `<p>` tag.
+
+To remove the HTML fallback, either use a HTML parser to drop the entire `strong`
+tag with a `data-mx-profile-fallback` attribute, or replace matches of the
+following regex with an empty string: `<strong\s+data-mx-profile-fallback(?:="")?\s*>([^<]+): </strong\s*>`
+
+#### Other details
+If a fallback is used in a media message, the `filename` property MUST be set
+to ensure the `body` is treated as a caption rather than a file name.
+
+Fallbacks MUST only be used when the `displayname` property is a non-empty
+string, i.e. they are not used when falling back to the member event name.
+
+#### Example with fallback
+```json
+{
+  "msgtype": "m.text",
+  "body": "cat: Hello, World!",
+  "format": "org.matrix.custom.html",
+  "formatted_body": "<p><strong data-mx-profile-fallback>cat: </strong>Hello, World!</p>",
+  "m.per_message_profile": {
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo",
+    "has_fallback": true
+  }
+}
+```
+
 ## Use cases
 
 ### Bridging
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -25,7 +25,7 @@ For example, bridges would likely set it to the immutable remote user ID.
 
 ### Encrypted avatars
 Because the profile is inside the ciphertext in encrypted events, the entire
-profile can be hidden the server, as long as the avatar is also encrypted.
+profile can be hidden from the server, as long as the avatar is also encrypted.
 Encrypted avatars are placed under `avatar_file` instead of `avatar_url`.
 The `avatar_file` field has the same schema as the `file` field in
 `m.room.message` events.
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -158,10 +158,9 @@ account has sufficiently high power level, and the displayname is unique among
 members of the room (i.e. it does not require disambiguation in the
 ["Calculating the display name for a user" spec](https://spec.matrix.org/v1.12/client-server-api/#calculating-the-display-name-for-a-user)).
 
-TODO: define the power level  
-TODO2: what if the same name is used by multiple different bridges using
-per-message profiles? Maybe it's enough to recommend configuring bridges to
-append a network identifier to names?
+The recommended power level to check is the level for a `m.per_message_profile`
+state event (i.e. check `events` -> `m.per_message_profile` and fall back to
+`state_default` if not set).
 
 ## Unstable prefix
 `com.beeper.per_message_profile` should be used instead of `m.per_message_profile`
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -108,7 +108,9 @@ Implementing encrypted avatars could cause difficulty for clients that assume
 that avatars are always unencrypted mxc URIs.
 
 ## Alternatives
-None.
+Per-message profiles could be transmitted more compactly by defining the profile
+in a new state event and only referencing the state key in the message event.
+However, that approach wouldn't enable encrypting per-message profiles.
 
 ## Security considerations
 
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -87,16 +87,24 @@ allow per-message profiles to specify which extensible profile is used.
 ### Bridging
 Per-message profiles will allow making "lighter-weight" bridges that don't need
 appservice access. Currently the only option for such bridges is to prepend the
-displayname to the message, which is extremely ugly.
+displayname to the message, which is extremely ugly. Even though they're ugly,
+there are still rooms that use bot-based bridges like matterbridge, which shows
+there's demand for bridging without requiring server admin access.
 
 Such bridges would obviously have downsides, like not being able to start chats
 via standard mechanisms, and not being able to see the member list on Matrix.
 However, those may be acceptable compromises for non-puppeting bridges that
 only operate in specific predetermined rooms.
 
+This method also allows encrypting profile info, which reduces metadata leaked
+by bridging.
+
 ### Feature-parity with other platforms
 Other chat applications such as Slack and Discord have "webhooks" which allow
 per-message profile overrides. This MSC effectively enables the same on Matrix.
+
+For example, Discord's [execute webhook](https://discord.com/developers/docs/resources/webhook#execute-webhook)
+API takes `username` and `avatar_url` as optional parameters.
 
 ### Roleplaying, plural users, etc
 Some users want to be able to switch between profiles quickly, which would be
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -74,7 +74,10 @@ However, there are use cases for setting an empty avatar, so `avatar_url` being
 an empty string should be treated as clearing the avatar and falling back to
 the client's default blank avatar behavior (e.g. generating one based on the
 displayname). If both `avatar_url` and `avatar_file` are omitted or null, the
-avatar from the member event should be used instead.
+avatar from the member event should be used instead. If the member event does
+not have an avatar defined either, and the client uses the displayname to
+generate fallback avatars, it should use the per-message displayname for the
+fallback avatar rather than the global one.
 
 ### Extensible profiles
 This MSC is not related to extensible profiles and does not attempt to
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -116,9 +116,21 @@ Implementing encrypted avatars could cause difficulty for clients that assume
 that avatars are always unencrypted mxc URIs.
 
 ## Alternatives
+### New state events
 Per-message profiles could be transmitted more compactly by defining the profile
 in a new state event and only referencing the state key in the message event.
-However, that approach wouldn't enable encrypting per-message profiles.
+However, that approach wouldn't enable encrypting per-message profiles without
+inventing encrypted state events. Additionally, even with encrypted state
+events, some kind of sender identifiers would be leaked via state keys.
+
+### Appservices
+Appservices work perfectly fine for bridging already now, but they require
+admin access to a server, which is not available for everyone. Additionally,
+they have similar metadata issues as the "New state events" alternative above.
+
+For use cases involving a single human user, having multiple mxids (regardless
+of whether they're registered manually or via an appservice) complicates things
+unnecessarily.
 
 ## Security considerations
 
--- a/proposals/4144-per-message-profile.md
+++ b/proposals/4144-per-message-profile.md
@@ -32,6 +32,14 @@ The field is allowed for all message events, which currently means
 `m.room.message` and `m.sticker`. A future MSC may expand the field to other
 events, such as reactions.
 
+### Field limits
+After JSON decoding, `id` and `displayname` MUST be at most 255 bytes of UTF-8
+and every code point MUST be a Unicode scalar. Null bytes (`\u0000`) are not
+allowed.
+
+Like with other images and avatars, clients should ignore any avatar URL that
+is not a `mxc://` URI.
+
 ### Encrypted avatars
 Because the profile is inside the ciphertext in encrypted events, the entire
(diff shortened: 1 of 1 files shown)


onestacked:
Alternative: Setup profiles (Probably as state messages) beforehand and then only refer to the profiles in the actual message.
This would reduce message size, especially if later MSC's add more things to the Profile.

Not sure if this MSC should be changed, but it I think should at least be mentioned as an alternative.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+### Roleplaying, plural users, etc
+Some users want to be able to switch between profiles quickly, which would be
+much easier using this MSC. Currently easiest way is to have multiple accounts,
+which has other benefits, but is much more cumbersome to manage.
+
+## Potential issues
+Implementing encrypted avatars could cause difficulty for clients that assume
+that avatars are always unencrypted mxc URIs.
+
+## Alternatives
```
tulir, line 110: > Alternative: Setup profiles (Probably as state messages) beforehand and then only refer to the profiles in the actual message.
This would reduce message size, especially if later MSC's add more things to the Profile.
> 
> Not sure if this MSC should be changed, but it I think should at least be mentioned as an alternative.

@onestacked MSC comments should always be in threads (to allow replying, marking as resolved)

Definitely a good idea to list under alternatives though

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,125 @@
...
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo"
+  }
+}
+```
+
+The `id` field is required and is an opaque string. Clients may use it to group
+messages with the same ID like they would group messages from the same sender.
+For example, bridges would likely set it to the immutable remote user ID.
```
clokep, line 24: So why not use separate MXIDs that each have their own profile? Is this purely about keeping more information encrypted?

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Bridging
+Per-message profiles will allow making "lighter-weight" bridges that don't need
+appservice access. Currently the only option for such bridges is to prepend the
+displayname to the message, which is extremely ugly.
+
+Such bridges would obviously have downsides, like not being able to start chats
+via standard mechanisms, and not being able to see the member list on Matrix.
+However, those may be acceptable compromises for non-puppeting bridges that
+only operate in specific predetermined rooms.
```
clokep, line 95: What's the upside of doing it this way? Not needing admin access on the server?

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Feature-parity with other platforms
+Other chat applications such as Slack and Discord have "webhooks" which allow
+per-message profile overrides. This MSC effectively enables the same on Matrix.
```
clokep, line 99: Not required, but for those of us not doing bridging often, links to their docs would be great.

proposals/4144-per-message-profile.md
tulir: I think this is basically the same question as your first comment. The assumptions in both are correct: per-message profiles would allow bridging without admin access and with more encryption.

Beeper is going to switch to local bridges with an encrypted cloud "backup". The backup isn't allowed to leak any metadata, which means the remote user profile info must be encrypted and even user IDs can't be in plaintext. I want to try to keep the room data as Matrix-compatible as possible, hence this MSC.

There are rooms that use extremely ugly relaybot bridges even now (matterbridge), so there's clearly some demand for bridging without requiring admin access. This MSC would make those bridges look much nicer.

proposals/4144-per-message-profile.md
tulir: Added link to https://discord.com/developers/docs/resources/webhook#execute-webhook

proposals/4144-per-message-profile.md
turt2live, line 1: Implementation requirements:
* [ ] Client sending avatars
* [ ] Client using avatars

proposals/4144-per-message-profile.md
clokep: Thanks for explaining! Does this mean that encrypted state (MSC3414?) is another alternative?

tulir: To some extent yes, although that'd still leak some kinds of user IDs: even if the state keys were hashes or something, it'd still leak that a different user sent a message, which is more metadata than encrypted per-message profiles. It also wouldn't help with the non-admin-bridges

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,145 @@
...
+### Roleplaying, plural users, etc
+Some users want to be able to switch between profiles quickly, which would be
+much easier using this MSC. Currently easiest way is to have multiple accounts,
+which has other benefits, but is much more cumbersome to manage.
+
+## Potential issues
+Implementing encrypted avatars could cause difficulty for clients that assume
+that avatars are always unencrypted mxc URIs.
+
+## Alternatives
```
sumnerevans, line 118: Reminds me of MSC3464. This is more general than that MSC, though.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,145 @@
...
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo"
+  }
+}
+```
+
+The `id` field is required and is an opaque string. Clients may use it to group
+messages with the same ID like they would group messages from the same sender.
+For example, bridges would likely set it to the immutable remote user ID.
```
kevincox, line 24: What is the scope of this `id`? I assume it needs to be scoped at least to the sender to avoid spoofing. But is it also scoped to the room, or are messages from the same sender in different rooms assumed to be the same?

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
+# MSC4144: Per-message profiles
+Currently profiles in Matrix are defined by `m.room.member` state events, and
+there is no easy way to have different profiles per message.
+
+## Proposal
+The proposed solution is a new field called `m.per_message_profile`, which
+contains a displayname and/or avatar URL to override the default profile,
+plus an ID to identify different profiles within the same Matrix user.
+
```
bkil, line 9: I would like to propose a new field to forward the remote username or nickname in.
* The `id` can be either randomly generated or a decimal sequence number, but it is definitely not something that I would expect to be prominently displayed in a client. Incidentally, the mxid serves this purpose on Matrix internally.
* The nickname is a short word that is quick to type in when we want to @-mention another member. Usually the localpart of the mxid acts as a poor substitute on matrix right now.
* Not all platforms support a display name, but where they do, it may contain the full, long personal name including spaces, pronouns, affiliations, position, emoji, etc. Many platforms also autocomplete for this when @-mentioning another user.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
+# MSC4144: Per-message profiles
+Currently profiles in Matrix are defined by `m.room.member` state events, and
+there is no easy way to have different profiles per message.
+
+## Proposal
+The proposed solution is a new field called `m.per_message_profile`, which
+contains a displayname and/or avatar URL to override the default profile,
```
bkil, line 7: We should provide guidance about the maximal length of `displayname`

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+plus an ID to identify different profiles within the same Matrix user.
```
bkil, line 8: We may want to share guidance about what information clients should use for name coloring. This would probably be the `id` (but the remote `nickname` may also be a second choice)

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+If the `displayname` field is omitted, null, or an empty string, the
+displayname from the member event should be used instead. Setting an empty
+displayname using a per-message profile is not supported, as there aren't any
+clear use cases for it.
+
+However, there are use cases for setting an empty avatar, so `avatar_url` being
+an empty string should be treated as clearing the avatar and falling back to
+the client's default blank avatar behavior (e.g. generating one based on the
+displayname). If both `avatar_url` and `avatar_file` are omitted or null, the
+avatar from the member event should be used instead.
```
bkil, line 77: What should happen if `avatar_url` and `avatar_file` is missing and the member event is also missing an avatar? I suggest generating an avatar from the remote displayname (or nickname) in such a case, not from the member. With such logic, we could save on having to include the empty string for `avatar_url` in each event.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+None.
+
+## Security considerations
+
+### Preventing impersonation
+To prevent impersonation using per-message profiles, clients should somehow
+indicate to the user that the message has a per-message profile with an easy
+way to see the user's MXID or default profile. For example, a client could have
+a small `via @user:example.com` text next to the per-message displayname.
+
```
bkil, line 142: User experience could be greatly improved in a room with mostly remote messages if this source disclaimer could be hidden (or only shown when clicking on the avatar) if all of the following conditions are met:

* The event sender is a moderator (e.g., has permission level for redaction or a new level)
* The impersonated user is not a member in the room

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+### Roleplaying, plural users, etc
+Some users want to be able to switch between profiles quickly, which would be
+much easier using this MSC. Currently easiest way is to have multiple accounts,
+which has other benefits, but is much more cumbersome to manage.
+
+## Potential issues
+Implementing encrypted avatars could cause difficulty for clients that assume
+that avatars are always unencrypted mxc URIs.
+
```
bkil, line 117: We should mention how this will interact with `<mx-reply>`

proposals/4144-per-message-profile.md
tulir: The spec doesn't have any concept of name coloring currently, so probably no point mentioning it here either

proposals/4144-per-message-profile.md
tulir: `<mx-reply>` no longer exists (#2781)

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+The `id` field is required and is an opaque string. Clients may use it to group
+messages with the same ID like they would group messages from the same sender.
+For example, bridges would likely set it to the immutable remote user ID.
+
+### Encrypted avatars
+Because the profile is inside the ciphertext in encrypted events, the entire
+profile can be hidden the server, as long as the avatar is also encrypted.
```
bkil, line 28: > hidden **from** the server

proposals/4144-per-message-profile.md
tulir: There's currently no mention support here, but that might make sense if mention support is added. Not entirely sure if it should be in this MSC or a separate one (the feature can be neatly separated from the rest of the MSC and it requires modifying the `matrix.to`/`matrix:` URI spec)

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+    "id": "meow",
+    "displayname": "cat",
+    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo"
+  }
+}
+```
+
+The `id` field is required and is an opaque string. Clients may use it to group
+messages with the same ID like they would group messages from the same sender.
+For example, bridges would likely set it to the immutable remote user ID.
```
bkil, line 24: It can happen that multiple platforms are relayed to the same Matrix room and clashes in names may happen. We should either mention that this `id` should contain a sufficient discriminant along with the remote user ID (e.g., prefixing it with the protocol or bridge gateway ID and a delimiter) or that we could introduce another field for this.

This is also important if we want to enable clients and bots to transparently ignore or ban relayed remote users by such a GUID.

tulir: This sounds like the same question as https://github.com/matrix-org/matrix-spec-proposals/pull/4144#discussion_r1695534155

proposals/4144-per-message-profile.md
bkil, line 117: We should mention that relaying reactji will be limited as a single Matrix user can only place the same reaction once on the same event. A workaround would be to perturbate the second and further application of the same reaction by appending a discriminator to the end (for example, invisible characters, numbers, the nick or user `id` in worst case). As maintaining or querying the reaction multiplicity can be difficult by a bridge, always appending the user `id` might be the easiest default workaround. All such counters will still show `1`, but at least it will be visible who has or has not reacted on an actionable message or legacy manual poll.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,123 @@
...
+### Bridging
+Per-message profiles will allow making "lighter-weight" bridges that don't need
+appservice access. Currently the only option for such bridges is to prepend the
+displayname to the message, which is extremely ugly.
+
+Such bridges would obviously have downsides, like not being able to start chats
+via standard mechanisms, and not being able to see the member list on Matrix.
+However, those may be acceptable compromises for non-puppeting bridges that
+only operate in specific predetermined rooms.
+
```
bkil, line 98: We could suggest that a client should try its best to autocomplete @-mentions based on the `displayname` (and nicknames) found in the recently buffered chat log as well as from the member list at present.

It would be ideal to rewrite such user pills before sending to a form that notifies the remote user such as by inserting their remote nickname (or lacking that, the `displayname`), possibly suffixed by whitespace and the `id` in brackets (or just the protocol if we introduced a separate field for that) to mitigate any ambiguity.

proposals/4144-per-message-profile.md
tulir: Probably makes sense to be user-specific (different sender mxid = different per-message profiles) but otherwise global (not room-specific)

proposals/4144-per-message-profile.md
tulir: Done

proposals/4144-per-message-profile.md
tulir: I think that's the same as https://github.com/matrix-org/matrix-spec-proposals/pull/4144#discussion_r1855526892

proposals/4144-per-message-profile.md
tulir: Scope has been defined to be only user-specific now

proposals/4144-per-message-profile.md
tulir: I've added that now, but there's also 2 new todos to figure out the power level to use and whether anything needs to be done about multiple bridges where the remote users have conflicting displaynames

proposals/4144-per-message-profile.md
tulir: Leaning towards defining mentions to be out of scope here

bkil: It can still be useful without support for mentions. For example, extant Matrix clients fall back to also showing the mxid (that can not clash) in case of a display name clash in a given room. Similarly they may want to fallback to the remote account name (remote nickname) when the remote display name clashes.

proposals/4144-per-message-profile.md
bkil: It would align with the present spec to state that
* the size of `id` MUST NOT exceed 255 bytes
* the size of `displayname` MUST NOT exceed 255 bytes

proposals/4144-per-message-profile.md
bkil: That's a pity. It's standardized on XMPP: https://xmpp.org/extensions/xep-0392.html

proposals/4144-per-message-profile.md
bkil: Both of them could be any arbitrary UTF-8 string. Why would you want to restrict it if you already state that `id` is opaque? Clients are already expected to properly sanitize and wrap it for their own rendering.

Including a 64kB `displayname` in each reply sounds excessive. I'd rather we had a recommendation for how to shorten it (i.e., cut it properly on a UTF-8 boundary before reaching 256 bytes and ideally trimming whitespace at its edges).

bkil: I would mention a suggestion that if the protocol and router discriminant, delimiter and the remote user ID together would exceed 255 bytes so it can't fit in the 
`id`, hashing should be used (and the discriminant should be adjusted to specify this in this case). You can always find a protocol that allows just a little longer IDs than you have anticipated, or there's the simple case of cascading such relays by independent parties, increasing the length of each ID by just a little bit on every hop. 256 bits of entropy is usually considered good enough for a UUID and that fits within 43 characters of unpadded base64 or 64 characters of hex, so it should still leave plenty for further metadata. E.g.,
* For the user `@nick42:matrix.example`, the `id` could be `mx1-mxid:@nick42:matrix.example`
* For a user with an excessive remote ID (`@too_long_name@matrix.example`), this could be `mx1-hash:04b9becd1c9306453d9aa80033a8a70fe9bd4c64989a06b470bc46ef6b8154c7`

proposals/4144-per-message-profile.md
tulir: gomuks web renders per-message profiles as of https://github.com/tulir/gomuks/pull/566

Beeper's linear bot will now set per-message profiles for webhook notices: https://github.com/beeper/linear-maubot/commit/88dfdd4ffa9c58ec4961e6351761997e5a2a930d

image

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,168 @@
...
+## Proposal
+The proposed solution is a new field called `m.per_message_profile`, which
+contains a displayname and/or avatar URL to override the default profile,
+plus an ID to identify different profiles within the same Matrix user.
+
+```json
+{
+  "msgtype": "m.text",
+  "body": "Hello, World!",
+  "m.per_message_profile": {
```
Half-Shot, line 14: I wonder if with extensible messages, it would make sense to support the (horrible) format of `{name}: {body}`. E.g.

```json
{
  "msgtype": "m.text",
  "body": "Hello, World!",
  "m.text": [
      { "body": "<strong>cat</strong>: Hello, world", "mimetype": "text/html" },
      { "body": "cat: Hello, world" }
  ],
  "m.per_message_profile": {
    "id": "meow",
    "displayname": "cat",
    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo"
  }
}
```
as currently the fallback for unsupported clients isn't *great*.

tulir: That might cause ptsd from reply fallbacks 😹 Maybe it is the best option though. Needs very strict rules to ensure the fallback can always be safely stripped out. An indicator for whether the fallback is present at all is probably also a good idea.

I don't think extensible events would help at all unless you duplicated the entire message (one copy with a fallback and another without), but duplication is also bad

Perhaps something along the lines of 

```json
{
  "msgtype": "m.text",
  "body": "cat: Hello, World!",
  "format": "org.matrix.custom.html",
  "formatted_body": "<strong data-mx-profile-fallback>cat: </strong>Hello, World!",
  "m.per_message_profile": {
    "id": "meow",
    "displayname": "cat",
    "avatar_url": "mxc://maunium.net/hgXsKqlmRfpKvCZdUoWDkFQo",
    "has_fallback": true
  }
}
```

where if `has_fallback` is true, `body` MUST be prefixed with the exact value of displayname, followed by a colon and a space, and `formatted_body` (if present) MUST have the same prefix inside a `strong` tag with the `data-mx-profile-fallback` attribute

tulir: Having the has_fallback flag would make the change fully compatible with all existing per-message profile messages and it would allow removing it in the future like reply fallbacks

The HTML version might have to allow not being the prefix, because otherwise it wouldn't really work with paragraph tags (you'd get a forced linebreak if you did something like `<strong data-mx-profile-fallback>cat: </strong><p>message text</p>`). A HTML parser should still be able to just drop anything with the `data-mx-profile-fallback` attribute and it might be reasonably safe to regex too because there won't be any html inside like there were in reply fallbacks

tulir: I have added fallbacks as per my example above to the MSC

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,234 @@
...
+
+When per-message profiles are used for bridging, there's no way to mention
+specific users, as mentions will just target the single bridge bot. A future
+MSC could define a way to specify which profile to mention.
+
+## Alternatives
+### New state events
+Per-message profiles could be transmitted more compactly by defining the profile
+in a new state event and only referencing the state key in the message event.
+However, that approach wouldn't enable encrypting per-message profiles without
```
olivia-fl, line 208: Per-room profile data (in `m.room.member`) is not currently encrypted. Is there a reason that per-message profile data is more sensitive than per-room profile data, so that using state events for per-message profiles should be rejected while using them for per-room profiles is okay?

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,234 @@
...
+Implementing encrypted avatars could cause difficulty for clients that assume
+that avatars are always unencrypted mxc URIs.
+
+When per-message profiles are used for bridging, there's no way to mention
+specific users, as mentions will just target the single bridge bot. A future
+MSC could define a way to specify which profile to mention.
+
+## Alternatives
+### New state events
+Per-message profiles could be transmitted more compactly by defining the profile
```
olivia-fl, line 206: This design has a second, more significant advantage, beyond just efficiency. It allows retroactively updating the profile information in old messages sent with a given profile id in a straightforward way. With the proposed design where profile fields are embedded in the messages, changing profile information in old messages requires sending an update to every message, generating a significant number of `m.replacement` events. This also requires the client searching through the full backscroll for the room, which may not even always be possible.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,234 @@
...
+have an empty string as its value (`=""`) for compatibility with all HTML
+generators, but MUST NOT have any other value. The HTML fallback is not mandated
+to be at the beginning of the string, as there may be valid reasons to put it
+inside another tag, such as a `<p>` tag.
+
+To remove the HTML fallback, either use a HTML parser to drop the entire `strong`
+tag with a `data-mx-profile-fallback` attribute, or replace matches of the
+following regex with an empty string: `<strong\s+data-mx-profile-fallback(?:="")?\s*>([^<]+): </strong\s*>`
+
+#### Other details
```
olivia-fl, line 143: My understanding is that, with the current spec text, `m.replacement` events are able to edit per-message profile fields, and clients should display the post-edit profile information in the UI. This is a bit non-obvious, and at least one current client implementation (sable) does *not* implement this behavior, so I think it's worth calling out explicitly if it is intended.

(if it is not intended, then there is a gap in this design for changing the profile information associated with existing messages)

proposals/4144-per-message-profile.md
tulir: As covered in the use cases section, bridges and other things that want to protect profile metadata can opt to use only per-message profiles and not store any of that in room state.

proposals/4144-per-message-profile.md
tulir: Probably makes sense to allow edits as there's no particular reason to disallow them, have specified that explicitly in https://github.com/matrix-org/matrix-spec-proposals/pull/4144/commits/a399fdac76b0aa9f1afc9691267f474168e4948b

proposals/4144-per-message-profile.md
tulir: I've expanded the description of that alternative in https://github.com/matrix-org/matrix-spec-proposals/pull/4144/commits/d4161fd4cfa25becbad356e48a62305635b4061c (this MSC won't be pursuing that option as the benefits of just having it in the message still outweigh the negatives)

disastercurl:
To address one of the main criticisms I have with this MSC: the suggestion of matching specifically this tag: `<strong\s+data-mx-profile-fallback(?:="")?\s*>([^<]+): </strong\s*>`

If we are recommending regular expressions, I would suggest the following, which is safer, handling multi-line input, other tags, matching the same closing tag, etc.
```js
input = input.replace(/<(\w+)[^>]*\bdata-mx-profile-fallback\b[^>]*>[\s\S]*?<\/\1>/g, "")
```

This however is not 100% safe, no regular expression is. Personally I would like to argue for non-fallback content to be explicitly stated in that MSC in a separate property field, because regex should be used for matching, not content manipulation in matrix client or bridge implementations.

Similarly, suggesting trimming the plaintext component by "PMP name length + 2" also assumes that clients implement the spec accurately as it stands today, and do not deviate. This also cannot be guaranteed.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,249 @@
...
+The displayname MUST be HTML-escaped and there MUST NOT be any HTML tags inside
+the fallback. The tag MUST NOT have any other attributes. The attribute MAY
+have an empty string as its value (`=""`) for compatibility with all HTML
+generators, but MUST NOT have any other value. The HTML fallback is not mandated
+to be at the beginning of the string, as there may be valid reasons to put it
+inside another tag, such as a `<p>` tag.
+
+To remove the HTML fallback, either use a HTML parser to drop the entire `strong`
+tag with a `data-mx-profile-fallback` attribute, or replace matches of the
+following regex with an empty string: `<strong\s+data-mx-profile-fallback(?:="")?\s*>([^<]+): </strong\s*>`
```
tulir, line 141: @Stealthii said outside of a thread:

> To address one of the main criticisms I have with this MSC: the suggestion of matching specifically this tag: `<strong\s+data-mx-profile-fallback(?:="")?\s*>([^<]+): </strong\s*>`
>
> If we are recommending regular expressions, I would suggest the following, which is safer, handling multi-line input, other tags, matching the same closing tag, etc.
> ```js
> input = input.replace(/<(\w+)[^>]*\bdata-mx-profile-fallback\b[^>]*>[\s\S]*?<\/\1>/g, "")
> ```
>
> This however is not 100% safe, no regular expression is. Personally I would like to argue for non-fallback content to be explicitly stated in that MSC in a separate property field, because regex should be used for matching, not content manipulation in matrix client or bridge implementations.
>
> Similarly, suggesting trimming the plaintext component by "PMP name length + 2" also assumes that clients implement the spec accurately as it stands today, and do not deviate. This also cannot be guaranteed.

tulir: > Personally I would like to argue for non-fallback content to be explicitly stated in that MSC in a separate property field

Duplicating the entire content is a really bad idea. Each fallback doubles the content, so if there are 3 different fallbacks (edits, formatting, per-message profiles), that'd result in 2\*2\*2 = 8 duplicates of the content. It might be "only" 6 duplicates if the non-fallback per-message profile is only inside `m.new_content`, but if anyone decided to invent a new thing needing a fallback copy, it'd jump to 12 or more

> which is safer, handling multi-line input, other tags, matching the same closing tag, etc.

The regex is intentionally fairly strict, other tags are not meant to be allowed. It could maybe slightly be relaxed in terms of the trailing space inside the fallback to allow `\n` there (`<br>` is obviously not allowed), perhaps `:\s*` instead of `: `

Alternatively, I guess the fallback itself could be put in another field, such that clients can simply do `formatted_body.replace(fallback, "")` rather than using a regex 🤔

> Similarly, suggesting trimming the plaintext component by "PMP name length + 2" also assumes that clients implement the spec accurately as it stands today, and do not deviate. This also cannot be guaranteed.

The only consequence of deviating from the spec is that the message will render incorrectly. If clients don't implement the spec correctly and their messages render differently than they intended, that's their problem, not the spec's.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,249 @@
...
+#### Plaintext fallback
+The fallback MUST be at the beginning of the plaintext `body` and consist of
+the exact value of the `displayname` property, followed by a colon (`:`) and
+a space (`\x20`), e.g. `cat: original message`.
+
+To remove the plaintext fallback, trim the prefix `{displayname}: ` from the body.
+
+#### HTML fallback
+If a `formatted_body` of type `org.matrix.custom.html` is present, it SHOULD
+include the same fallback text inside a `strong` tag with the `data-mx-profile-fallback`
```
cloudrac3r, line 129: I think it would be good if the HTML fallback could include the per-message avatar as well as the username. This could be done with an img tag per the custom emojis specification.

Example: `<strong data-mx-profile-fallback> username: </strong>`

Avatars are important for building mental associations of individual users. I think this would greatly improve people's understanding of the messages if they are using a client that doesn't support this MSC.

tulir: I don't think the MSC is going to suggest that, users who want to understand messages better should use a client that supports per-message profiles. However, if the fallback removal was done using a string replacement instead of regex (as mentioned in the thread above this one), then the sender could technically add images as an implementation detail.

proposals/4144-per-message-profile.md
Thaodan: It would be really useful to have the fallback without making it confusing for users which might see those.

Similarly if the original username and the fallback is the same it doesn't make much sense to have both. Not sure if that's possible but it would cause less frictions with clients which don't support per message profiles.

tulir: Not sure how that's related to this thread, but fallbacks are already completely optional. Clients can freely omit the fallback when sending messages if they don't think it provides any value.

Thaodan: clients who don't handle per message user names, display usernames twice even if the senders username outside of that message would be the same, when  receiving them from clients who do send fallbacks.

tulir: As I just said, clients don't have to include a fallback. If the sending client notices that it's going to send a per-message profile with the same name as the user's global profile (e.g. because only the avatar is changed), it can simply skip adding the fallback. The fallback is completely optional, so the choice of whether to include it is up to the sender.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,249 @@
...
+## Proposal
+The proposed solution is a new field called `m.per_message_profile`, which
+contains a displayname and/or avatar URL to override the default profile,
+plus an ID to identify different profiles within the same Matrix user.
+
+```json
+{
+  "msgtype": "m.text",
+  "body": "Hello, World!",
+  "m.per_message_profile": {
```
ajbura, line 14: Could it also include origin of original profile?

something like: `"origin": "discord.com"`

it can help client render where this user actually is from.

tulir: Generally the `sender` of the event is a sufficient hint for where the message is from. I don't think there's any benefit to a new field, especially since it wouldn't be able to say the real origin anyway (like what if it's a discord webhook that comes from github or another chat network?)

Blackilykat:
I'm not active in matrix spec proposals but am developing a bot which benefits from this feature and I'd like to give my 2 cents.

I think the sender should have more freedom on how to format the HTML fallback. Currently, it is required to be in exactly a `<strong>` tag with the `data-mx-profile-fallback` attribute containing exactly the text `<profile>: `. I think it would be beneficial to allow any tag (i.e. span) with the same attribute and any contents, including HTML formatted contents. This would include the original fallback but allow bots to customize the fallback if needed.

My use case: I am bridging a Minecraft chat over matrix. Different players have different chat formats which I would like to include in the fallback so outdated clients will see chat exactly as shown in-game. Clients which support this feature would instead show the message as if it was sent by that user, ignoring the HTML content inside the fallback.

If a client supports HTML-formatted content it should already have the logic to correctly parse HTML tags and ignoring a first tag with the `data-mx-profile-fallback` attribute, regardless of which tag it is or its contents, should not be difficult to implement.

proposals/4144-per-message-profile.md
tulir: @Blackilykat [posted](https://github.com/matrix-org/matrix-spec-proposals/pull/4144#issuecomment-5216616714) outside of a thread:

> I'm not active in matrix spec proposals but am developing a bot which benefits from this feature and I'd like to give my 2 cents.
>
> I think the sender should have more freedom on how to format the HTML fallback. Currently, it is required to be in exactly a `<strong>` tag with the `data-mx-profile-fallback` attribute containing exactly the text `<profile>: `. I think it would be beneficial to allow any tag (i.e. span) with the same attribute and any contents, including HTML formatted contents. This would include the original fallback but allow bots to customize the fallback if needed.
>
> My use case: I am bridging a Minecraft chat over matrix. Different players have different chat formats which I would like to include in the fallback so outdated clients will see chat exactly as shown in-game. Clients which support this feature would instead show the message as if it was sent by that user, ignoring the HTML content inside the fallback.
>
> If a client supports HTML-formatted content it should already have the logic to correctly parse HTML tags and ignoring a first tag with the `data-mx-profile-fallback` attribute, regardless of which tag it is or its contents, should not be difficult to implement.

tulir: Clients may want to remove fallbacks at a different stage than HTML parsing, so requiring the use of a parser is probably a bad idea. Several clients had trouble with reply fallbacks too.

A separate field containing the exact fallback string to remove should be easy and safe to remove, but it also feels slightly weird and has never been done before. Maybe I should just do that anyway

Blackilykat: Maybe you could include the length of the formatted fallback string instead of the whole thing to avoid sending/storing duplicate data.

tulir: That would require it to strictly be a prefix, which may not be the case if the message uses paragraph tags or other such things

proposals/4144-per-message-profile.md
KitsuneRal: Bikeshedding but `per_message` part looks a bit extraneous here, given that this is a subobject in a message event.

proposals/4144-per-message-profile.md
```diff
@@ -0,0 +1,249 @@
...
+
+Modifying per-message profiles using edits is allowed, i.e. clients SHOULD
+include the per-message profile object in `m.new_content` when sending edits
+and read it when receiving edits. Including the per-message profile object in
+the edit fallback at the top level is not necessary, as all clients that support
+per-message profiles can be assumed to support edits.
+
+### Field limits
+After JSON decoding, `id` and `displayname` MUST be at most 255 bytes of UTF-8
+and every code point MUST be a Unicode scalar. Null bytes (`\u0000`) are not
```
KitsuneRal: Bytes and UTF-8 are orthogonal. Maybe this is better?
```suggestion
After JSON decoding, `id` and `displayname` MUST be at most 255 bytes and consist
of valid Unicode scalars encoded as UTF-8. Null bytes (`\u00`) are not
```

q-TheDog:
Client Implementation ([Haven Desktop](https://github.com/Haven-Organization/haven-desktop)): https://github.com/Haven-Organization/haven-desktop/commit/15fd359b2c6e61a2357c9239dad0628aaa854b13

PR #4145, opened 2024-05-14
MSC4145: Simple verified accounts

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/verified-accounts-part1/proposals/4145-simple-verified-accounts.md)

### DIFF
--- a/proposals/4145-simple-verified-accounts.md
+++ b/proposals/4145-simple-verified-accounts.md
@@ -1,0 +1,105 @@
+# MSC4145: Simple verified accounts
+
+[Server notices](https://spec.matrix.org/v1.10/client-server-api/#server-notices) and similar alerts
+sent by servers are typically sent by a general purpose user such as `@support:example.org`. These
+messages can [look like a phishing attempt](https://github.com/element-hq/element-meta/issues/1759)
+when they lack any "official" markings.
+
+This proposal introduces a simple server-local verification mechanism to identify "verified"
+accounts. This proposal does *not* introduce a mechanism to request/assign verification, nor manage
+that verification - this is left as an implementation detail under this proposal, and may be
+improved by a future MSC.
+
+## Proposal
+
+The user's [global profile](https://spec.matrix.org/v1.10/client-server-api/#profiles) has a new
+*optional* `m.verified` property added, as shown by the below endpoints.
+
+`m.verified` takes on a slightly weird shape to allow for future expansion, such as the potential
+for a signed object, attestation, etc. The `m.verified` object can only contain a single boolean
+currently: `verified`. When `m.verified` is supplied, `verified` can only be `true` (otherwise the
+user is considered "unverified" or "default").
+
+[`GET /_matrix/client/v3/profile/{userId}`](https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientv3profileuserid)
+response:
+```jsonc
+{
+  "displayname": "Support",
+  "avatar_url": "mxc://example.org/abc123",
+  "m.verified": { // new!
+    "verified": true
+  }
+}
+```
+
+If the user is not verified, `m.verified` will not be present.
+
+**New Endpoint** - `GET /_matrix/client/v3/profile/{userId}/m.verified`
+
+*No request body.*
+
+Response:
+```jsonc
+{
+  "m.verified": {
+    "verified": true
+  }
+}
+```
+
+The endpoint behaves similar to [`GET /_matrix/client/v3/profile/{userId}/displayname`](https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientv3profileuseriddisplayname):
+if the user doesn't exist or is not verified locally on the server, the endpoint returns 404 `M_NOT_FOUND`.
+Note that this new endpoint does *not* go over federation to determine verified status - for this
+proposal, verification is very intentionally a server-local state.
+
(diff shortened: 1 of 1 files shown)


andrewzhurov:
Will this benefit from having verification as Verifiable Credential (VC)?

Would open up possibility to accumulate all sorts of VCs (e.g., your real name, age from a goverment, phone number from a phone provider).

Would be a broadly used standard.

VCs could be stored in a Profile Room, that is used to manage all related to ones self-sovereign identity. #1769
Makes it easier to discover everything related to a person.

proposals/4145-simple-verified-accounts.md
```diff
@@ -0,0 +1,105 @@
...
+
+Verification status is not expected to change often, so longer cache values are preferred.
+
+## Potential issues
+
+This proposal doesn't work over federation. This is a feature, not a bug. Communicating verification
+over federation would ideally mean having a method for end clients to confirm the verification
+without the use of the attached homeserver. Such a problem is challenging to solve quickly.
+
+## Alternatives
```
turt2live, line 86: @andrewzhurov says:

> Will this benefit from having verification as Verifiable Credential (VC)?
> 
> Would open up possibility to accumulate all sorts of VCs (e.g., your real name, age from a goverment, phone number from a phone provider).
> 
> Would be a broadly used standard.
> 
> VCs could be stored in a Profile Room, that is used to manage all related to ones self-sovereign identity. #1769 Makes it easier to discover everything related to a person.

turt2live: [Verifiable Credentials](https://www.w3.org/TR/vc-data-model-2.0/) look very interesting, and likely simple enough where we could skip straight to them - thanks for raising it! A concern this MSC aims to avoid is designing a complete system too early, as such a decision could end up adding significant time before the benefit is realized. I'll have to give the W3C draft a more serious read.

proposals/4145-simple-verified-accounts.md
```diff
@@ -0,0 +1,105 @@
...
+
+[Server notices](https://spec.matrix.org/v1.10/client-server-api/#server-notices) and similar alerts
+sent by servers are typically sent by a general purpose user such as `@support:example.org`. These
+messages can [look like a phishing attempt](https://github.com/element-hq/element-meta/issues/1759)
+when they lack any "official" markings.
+
+This proposal introduces a simple server-local verification mechanism to identify "verified"
+accounts. This proposal does *not* introduce a mechanism to request/assign verification, nor manage
+that verification - this is left as an implementation detail under this proposal, and may be
+improved by a future MSC.
```
bxdxnn, line 11: I think the bare minimum could be providing an admin API to change the verification status?

proposals/4145-simple-verified-accounts.md
```diff
@@ -0,0 +1,105 @@
...
+  "m.verified": {
+    "verified": true
+  }
+}
+```
+
+The endpoint behaves similar to [`GET /_matrix/client/v3/profile/{userId}/displayname`](https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientv3profileuseriddisplayname):
+if the user doesn't exist or is not verified locally on the server, the endpoint returns 404 `M_NOT_FOUND`.
+Note that this new endpoint does *not* go over federation to determine verified status - for this
+proposal, verification is very intentionally a server-local state.
```
bxdxnn, line 53: 1. Technically, we can make this feature federated and leave it up to clients whether they want (or how do they want) to show federated verifications. It can be done in a way like showing a different color for the badge.
It's useful to have the verification badge federated in cases like server admins interacting over federation, e.g. matrix.org's moderation writes to an admin of another server and vice-versa, or for official government servers.

2. Of course someone would make a scam/bot server with all the users verified, but that's why the clients should have the badge visualy separated between federation/local.

3. As a follow-up, we might also have a "verified servers" feature that allows admins to set a list of verified servers similar to this feature, so that clients can be even more confident that the account isn't phishing, and therefore e.g. show the badges the same way as "local verified" or have another "super-verified" badge.

proposals/4145-simple-verified-accounts.md
```diff
@@ -0,0 +1,105 @@
...
+header, or for a minimum of 24 hours. Similarly, `m.verified` on the general `/profile/{userId}`
+endpoint must be cached for at least 24 hours.
+
+Verification status is not expected to change often, so longer cache values are preferred.
+
+## Potential issues
+
+This proposal doesn't work over federation. This is a feature, not a bug. Communicating verification
+over federation would ideally mean having a method for end clients to confirm the verification
+without the use of the attached homeserver. Such a problem is challenging to solve quickly.
```
bxdxnn, line 84: See my comment above

proposals/4145-simple-verified-accounts.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Alternatives
+
+Significant alternatives are implied throughout. Namely, a proper verification system could be
+introduced instead, though such a system requires significant engineering effort.
+
+## Security considerations
```
bxdxnn, line 91: If this will be federated, a point about the "visual separation" from my comment above can be added there.

bxdxnn (commented):
@turt2live is there anything that can/should be done/helped with for this MSC to proceed further?

PR #4146, opened 2024-05-23
MSC4146: Shared Message Drafts

cyrneko:
[Rendered](https://github.com/cyrneko/matrix-spec-proposals/blob/shared-message-drafts/proposals/4146-Shared-Message-Drafts.md)

This MSC proposes a way to share message drafts between clients, to edit and/or send them on clients they weren't initially written in.

Signed off by: Simon Müller (simeonlps@proton.me)

### DIFF
--- a/proposals/4146-Shared-Message-Drafts.md
+++ b/proposals/4146-Shared-Message-Drafts.md
@@ -40,6 +40,8 @@ This could be done something like this, with a field to indicate what room it re
 }
 ```
 
+The format of the event should be somewhat compatible with clients that do not implement the MSC, so that in those clients the draft gets displayed as a message that the user could still manually copy or edit to also edit the draft. Theoretically there could be an extra field, say `m.draft_full_event` that contains a full event with intentional mentions, media and more, but this is an idea that I am still exploring.
+
 When a user is in multiple `m.drafts` rooms, the client should take whatever draft is chronologically the newest for a given room from any `m.drafts` room, unless there is two or more drafts with the same timestamp. When there is two or more drafts with the same timestamp, the client should give the user the option of which draft to keep, redacting the other drafts in whatever room the drafts are in.
 
 The Client should also only import it's own events when checking `m.drafts` rooms, unless explicitly requested otherwise by the user.
--- a/proposals/4146-Shared-Message-Drafts.md
+++ b/proposals/4146-Shared-Message-Drafts.md
@@ -35,8 +35,8 @@ DMs, Spaces or Groups. A creation event could look something like this:
 
 In said room, the client would then send normal `m.text` events, optionally with
 a formatted body, and would edit the message every ´n´ seconds or once the user
-stops typing. <span style="color: grey">(This is more of a UI/UX implementation
-detail though)</span>
+stops typing. (This is more of a UI/UX implementation
+detail though)
 
 Another thing that would be introduced would be a field that defines what room
 this is supposed to be for / In which room this draft was written
@@ -57,12 +57,66 @@ relates to:
 }
 ```
 
+This format also works with media:
+
+```json
+{
+  "content": {
+    "body": "image.png",
+    "info": {
+      "h": ...,
+      "mimetype": "image/png",
+      "size": ...,
+      "w": ...,
+      "xyz.amorgan.blurhash": "..."
+    },
+    "msgtype": "m.image",
+    "url": "mxc://example.net/...",
+    "m.draft_in": "!ROOMID:example.com"
+  }
+}
+```
+
+In the case of Media, a client would take the `body`, `info` and `url` fields
+into account when importing it into the editor. In the case of text, a client
+would import the `body`, `formatted_body` and `format` fields
+
 The format of the event should be somewhat compatible with clients that do not
 implement the MSC, so that in those clients the draft gets displayed as a
 message that the user could still manually copy or edit to also edit the draft.
-Theoretically there could be an extra field, say `m.draft_full_event` that
-contains a full event with intentional mentions, media and more, but this is an
-idea that I am still exploring.
+
+Additionally, a client can include the `m.draft_full_event` field in the draft,
+which allows the client to upload a full event including media, intentional
+mentions and other metadata that are not contained in the `body` or
+`formatted_body` of an event. Such an event would look something like this:
+
+```json
+{
+  "type": "m.room.message",
+  "content": {
+    "msgtype": "m.text",
+    "body": "**test!!!**",
+    "format": "org.matrix.custom.html",
+    "formatted_body": "<strong>test!!!</strong>",
+    "m.draft_in": "!ROOMID:example.com",
+    "m.draft_full_event": {
+      ...
+      "m.mentions": {
+        "user_ids": [
+          "@1:example.com",
+          "@2:example.com"
+        ]
+      }
+    }
+  }
+}
+
+```
+
+This would be useful in events where a client may want to have more information
+in the draft than what fits into the existing body fields, like drafting a poll,
+drafting with intentional mentions (like in the example above) or drafting a
+custom event. This field is **not** required.
 
 When a user is in multiple `m.drafts` rooms, the client should take whatever
 draft is chronologically the newest for a given room from any `m.drafts` room,
@@ -123,6 +177,11 @@ By implementing this by repurposing an **encrypted** room you can avoid the
 drafts - and as such messages that might go into an encrypted room - being
 leaked to the homeserver. Encryption for a feature like this is a goal that
 should not be neglected.
+
+Clients should consider only uploading drafted media to the room if explicitly
+requested by the user, as this could otherwise lead to media that never gets
+deleted from the server; This could pose a threat to privacy if the `m.drafts`
+room is unencrypted.
 
 ## Unstable prefix
 
--- a/proposals/4146-Shared-Message-Drafts.md
+++ b/proposals/4146-Shared-Message-Drafts.md
@@ -92,7 +92,7 @@ Proposed unstable prefixes for `m.drafts` and `m.draft_in` would be:
 | prefix                        | description                       |
 |-------------------------------|-----------------------------------|
 | `org.matrix.msc4146.drafts`   | For the room type                 |
-| `org.matrix.msc4146.draft_in` | For the actual events in the room |
+| `org.matrix.msc4146.draft_in` | For the room ID information on the events |
 
 ## Dependencies
 
--- a/proposals/4146-Shared-Message-Drafts.md
+++ b/proposals/4146-Shared-Message-Drafts.md
@@ -1,14 +1,39 @@
-# MSC4146: Shared Message Drafts
+# MSCXXXX: Shared Message Drafts
 
-In Matrix, most clients support saving drafts, yet there is no way to share this data. This proposal outlines how such data could be shared.
+In Matrix, most clients support saving drafts, either in LocalStorage or some 
+other platform-specific storage so that you can come back to writing a message 
+later if you have not finished it, however this information does not carry over 
+between clients, meaning you can only continue your draft or send it on the 
+device that you initially started writing it in.
+
+This proposal aims to address that problem by sharing ideas on a mechanism to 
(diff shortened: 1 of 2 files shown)


proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+
+An alternative way to implement this would be to instead add this data to account data, adding a field similar to what Element does with `io.element.recent_emoji`
+
+The field could be named something like `m.room_drafts` or `m.drafts` (just like the room type from idea #1) and could contain something like this:
+
+```json
+{
+  "type": "m.room_drafts",
+  "content": {
+    "!ROOMID:example.com": {
```
clokep, line 71: Using [room specific account data](https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientv3useruseridroomsroomidaccount_datatype) could remove a layer of mapping in this.

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+By implementing this by repurposing an **encrypted** room you can avoid the drafts - and as such messages that might go into an encrypted room - being leaked to the homeserver. Encryption for a feature like this is a goal that should not be neglected.
+
+## Unstable prefix
+
+Proposed unstable prefixes for `m.drafts` and `m.draft_in` would be:
+
+| prefix                        | description                       |
+|-------------------------------|-----------------------------------|
+| `org.matrix.msc4146.drafts`   | For the room type                 |
+| `org.matrix.msc4146.draft_in` | For the actual events in the room |
```
clokep, line 95: I don't think this matches the proposal?

```suggestion
| `org.matrix.msc4146.draft_in` | For the room ID information on the events |
```

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
+# MSC4146: Shared Message Drafts
+
+In Matrix, most clients support saving drafts, either in LocalStorage or some other platform-specific storage so that you can come back to writing a message later if you have not finished it, however this information does not carry over between clients, meaning you can only continue your draft or send it on the device that you initially started writing it in.
+
+This proposal aims to address that problem by sharing ideas on a mechanism to share message drafts between devices, and serves as a place of discussion for such a feature.
+
+This being my first MSC, without having read the entire spec there will be mistakes and oversights I've made as to how Matrix operates and how a feature like this may be implemented, so I'd very much appreciate some pointers or help if I got anything wrong :)
```
clokep, line 7: Thanks for providing this context! But I think it can be removed from the actual MSC text which is meant to be mostly technical. 😄

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Proposal
+
+In Matrix, especially when used in work environments, there is often the need to draft out a message before sending it out, like wanting someone else to proofread it first, writing a message about an ongoing event, or a plethora of other scenarios where a message should just remain unsent until the time comes. However the experience is frustrating because you are limited only to one device, or even just one client on the same device. 
+
+If you're ever in a situation where you want to send the message, edit it, or maybe even delete the draft, you cannot do so without first going back to the device you initially wrote it on. This means you cannot carry over a draft from your Phone to your Computer to have a more comfortable typing experience, or to see the text on a bigger screen, and is generally just an inconvenience.
```
clokep, line 14: This feels like it is mostly duplicating the intro, maybe combine it?

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+
+My first idea would be to repurpose rooms, similar to how it's proposed for Profiles-As-Rooms or how Spaces are implemented. A new room type, `m.drafts` or `m.draft` would be added, allowing the client to differentiate this room from DMs, Spaces or Groups. A creation event could look something like this:
```
clokep, line 16: Write this as if it is the change:

```suggestion
Rooms will be repurposed to store drafts, similar to how it's proposed for Profiles-As-Rooms or how Spaces are implemented. A new room type, `m.drafts` would be added to allow clients to differentiate it from a normal room. A creation event could look something like this:
```

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+    "msgtype": "m.text",
+    "body": "**test!!!**",
+    "format": "org.matrix.custom.html",
+    "formatted_body": "<strong>test!!!</strong>",
+    "m.draft_in": "!ROOMID:example.com"
+  }
+}
+```
+
+## Potential issues
```
clokep, line 48: Is it valid for other members to be in that room? What happens if a user is in multiple `m.drafts` rooms?

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+
+*Not all proposals are perfect. Sometimes there's a known disadvantage to implementing the proposal,
+and they should be documented here. There should be some explanation for why the disadvantage is
+acceptable, however - just like in this example.*
```
clokep, line 52: I think this is from the template?

```suggestion
```

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+
+There are some issues with both approaches, notably I thought of:
```
clokep, line 54: What do you mean by "both" here? What's the other approach?

proposals/4146-Shared-Message-Drafts.md
TGRush: fair change, I see no issues

proposals/4146-Shared-Message-Drafts.md
TGRush: Got it! Thanks for the heads-up

proposals/4146-Shared-Message-Drafts.md
TGRush: I did try a bit *too* hard on following the template maybe, but yeah that is (again) a fair point

proposals/4146-Shared-Message-Drafts.md
TGRush: Thank you!

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,99 @@
...
+  "content": {
+    "creator": "@example:example.com",
+    "room_version": "9",
+    "type": "m.drafts"
+  },
+  ...
+}
+```
+
+In said room, the client would then send normal `m.text` events, optionally with a formatted body, and would edit the message every ´n´ seconds or once the user stops typing. <span style="color: grey">(This is more of a UI/UX implementation detail though)</span>
```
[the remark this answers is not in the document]
TGRush: I figured it would be easiest to either only save the plaintext or formatted body of an event because from what I've seen, the client would be parsing the input anyways before sending it off as an event

So if you are mentioning someone in a draft, and it gets synced to another device, the client would add intentional mentions anyways. I made this decision in an effort to minimize the amount of data that needs to be synced, but I think having a copy of the *whole* event as it would be sent out may actually be beneficial if you're drafting something with Media for example :thinking: 

I'll reconsider this for sure, thank you!

proposals/4146-Shared-Message-Drafts.md
TGRush: I think multiple members apart from yourself to be in the room would be fine, but I will have to clarify in the proposal that the client should only read it's own events (e.g don't import drafts from other users), unless explicitly requested otherwise by the user

As for being in multiple `m.drafts` rooms, the client should take whatever is chronologically the newest draft for any given room, unless there is two drafts for one room with the same timestamp. In that case the user should be given the option to pick between which draft to keep, and have the other draft redacted.

Thanks!

proposals/4146-Shared-Message-Drafts.md
TGRush: oh boy yeah, I forgot to remove that one :upside_down_face:

proposals/4146-Shared-Message-Drafts.md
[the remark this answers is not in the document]
clokep: Yeah, I think that makes sense. Just clarifying the idea here.

proposals/4146-Shared-Message-Drafts.md
TGRush: That is a little leftover from when I was writing this locally, I initially had the structure of the document a bit differently so it made sense to write it that way, but after I thought of encryption I moved the 2nd idea into the *Alternatives* section instead.

I'll remove/fix this in the next commit

proposals/4146-Shared-Message-Drafts.md
TGRush: I'll look into that, thank you! :)

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,95 @@
+# MSC4146: Shared Message Drafts
+
+In Matrix, most clients support saving drafts, yet there is no way to share this data. This proposal outlines how such data could be shared.
+
+## Proposal
+
+There is often the need to draft out a message before sending it out, like wanting someone else to proofread it first, writing a message about an ongoing event, or a plethora of other scenarios where a message should just remain unsent until the time comes. However the experience is frustrating because you are limited only to one device, or even just one client on the same device.
```
turt2live, line 7: (please wrap lines to ~100 characters)

proposals/4146-Shared-Message-Drafts.md
turt2live, line 1: Implementation requirements:

* [ ] Client recognizing and using the feature

proposals/4146-Shared-Message-Drafts.md
cyrneko: oh sorry, I completely missed that! I'll get to it in a moment :)

cyrneko: had something to do, it should be wordwrapped at ~80 characters now :)

proposals/4146-Shared-Message-Drafts.md
```diff
@@ -0,0 +1,198 @@
...
+would be unavailable to a client that has just authenticated but has not been
+verified yet, and as such cannot get the keys for the messages in the room and
+decrypt any of the drafts.
+- Clients that do not support this MSC might display
+the drafts room normally, allowing the user to send arbitrary messages into the
+room, which might become messy.
+- If a client removes the `m.draft_in` as part
+of an edit the draft would no longer be associated with a room, causing
+confusion on the users' end as to where their draft went.
+
```
uhoreg, line 145: Another issue is that converting to and from a user's input to a Matrix message can be lossy.  For example, if a user starts writing a message on a client that uses Markdown, and resumes editing later on a different client that uses a different input format, some formatting may get lost.

cyrneko: I think that shouldn't be *too* much of a problem considering the plaintext should (in theory) be stored in the `body`, alongside a pre-formatted version in the `formatted_body`

The client is expected to import the `body`, although maybe I should make that more explicit in the document

(or, alternatively to the `body`, if available the `m.draft_full_event` if it's something like a poll or other more complex event, but I have not gotten much feedback on that yet)

proposals/4146-Shared-Message-Drafts.md
uhoreg: Account data would have been my first guess for implementing this feature, but you'd need to figure out some encryption scheme to avoid leaking data to the server.

cyrneko:
I think that, at least for the time being, I'll drop this idea. A few reasons for that:
- as surfaced to me by clokep, it's probably more sensible to use per-room account data, even if that's "less backwards-compatible".
- The project I was working on that would have implemented this no longer exists, and I don't think anyone else is interested in implementing this MSC.

So, until next time, thank you all for your time! it is much appreciated <3

PR #4147, opened 2024-05-29
MSC4147: Including device keys with Olm-encrypted events

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/include_device_keys/proposals/4147-including-device-keys-with-olm-encrypted-events.md)

disclosure: I am on the Spec Core Team and employed by Element.  This MSC is written as part of my work in the Crypto team at Element

---

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4147#issuecomment-2453093009)

### DIFF
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -15,7 +15,7 @@ recipient does not have to try to query the sender's keys.
 
 ## Proposal
 
-The plaintext payload of the [`m.olm.v1.curve25519-aes-sha2` encryption
+The plaintext payload of `m.room.encrypted` events encrypted with the [`m.olm.v1.curve25519-aes-sha2` encryption
 algorithm](https://spec.matrix.org/unstable/client-server-api/#molmv1curve25519-aes-sha2)
 is currently of the form:
 
@@ -57,8 +57,8 @@ plaintext payload will now look something like:
     "user_id": "<user_id>",
     "device_id": "<device_id>",
     "keys": {
-      "ed25519:<device_id>": "<our_ed25519_key>",
-      "curve25519:<device_id>": "<our_curve25519_key>"
+      "ed25519:<device_id>": "<sender_ed25519_key>",
+      "curve25519:<device_id>": "<sender_curve25519_key>"
     },
     "signatures": {
       "<user_id>": {
@@ -75,7 +75,7 @@ payload must be the same as the `device_keys`.`keys`.`ed25519:<DEVICEID>`
 property.  If they differ, the recipient should discard the event.
 
 As the `keys` property is now redundant, it may be removed in a future version
-of Olm.
+of the Matrix specification.
 
 ## Potential issues
 
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -1,4 +1,4 @@
-# MSC4147: Including device keys with Olm-encrypted events
+# MSC4147: Including device keys with Olm-encrypted to-device messages
 
 Summary: a proposal to ensure that messages sent from a short-lived (but
 genuine) device can be securely distinguished from those sent from a spoofed
@@ -38,7 +38,7 @@ recipient does not have to try to query the sender's keys.
 
 ## Proposal
 
-The plaintext payload of `m.room.encrypted` events encrypted with the [`m.olm.v1.curve25519-aes-sha2` encryption
+The plaintext payload of to-device messages encrypted with the [`m.olm.v1.curve25519-aes-sha2` encryption
 algorithm](https://spec.matrix.org/v1.12/client-server-api/#molmv1curve25519-aes-sha2)
 is currently of the form:
 
@@ -102,8 +102,8 @@ of the Matrix specification.
 
 ## Potential issues
 
-Adding this property will increase the size of the event.  We found it
-increased the length of a typical `m.room_key` message from about 1400 to 2400
+Adding this property will increase the size of the to-device message.  We found it
+increased the length of a typical encrypted `m.room_key` message from about 1400 to 2400
 bytes (a 70% increase). This will require increased storage on the recipient
 homeserver, and increase bandwidth for both senders and recipients. See
 [Alternatives](#alternatives) for discussion of mitigation strategies.
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -2,7 +2,7 @@
 
 When a Megolm session is sent from one device to another via Olm, the recipient
 can
-[query](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3keysquery)
+[query](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3keysquery)
 the sender's device keys and check whether the device has been cross-signed in
 order to determine whether the sending device can be trusted.  However, this
 does not work if the sending device has since logged out as the recipient will
@@ -16,7 +16,7 @@ recipient does not have to try to query the sender's keys.
 ## Proposal
 
 The plaintext payload of `m.room.encrypted` events encrypted with the [`m.olm.v1.curve25519-aes-sha2` encryption
-algorithm](https://spec.matrix.org/unstable/client-server-api/#molmv1curve25519-aes-sha2)
+algorithm](https://spec.matrix.org/v1.12/client-server-api/#molmv1curve25519-aes-sha2)
 is currently of the form:
 
 ```json
@@ -36,7 +36,7 @@ is currently of the form:
 
 We propose to add a new property: `device_keys`, which is a copy of what the
 server would return in response to a
-[`/keys/query`](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3keysquery)
+[`/keys/query`](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3keysquery)
 request, as the device keys for the sender's device.  In other words, the
 plaintext payload will now look something like:
 
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -95,7 +95,7 @@ in the `device_keys` is meaningless.  The recipient will need to re-query the
 device keys, and will need to treat the sender as untrusted if it fails to do
 so.  The sender could include the self-signing key, signed by the
 master-signing key, in the plaintext event, so that if the user only resets
-their self-signing key but retains their masster-signing key, the recipient can
+their self-signing key but retains their master-signing key, the recipient can
 still check the sender's device keys.  However, this will further increase the
 size of the event, and it is not common for clients to reset the self-signing
 key without also resetting the master-signing key, so this is unlikely to give
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -57,8 +57,8 @@ is currently of the form:
 }
 ```
 
-We propose to add a new property: `device_keys`, which is a copy of what the
-server would return in response to a
+We propose to add a new property: `sender_device_keys`, which is a copy of what
+the server would return in response to a
 [`/keys/query`](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3keysquery)
 request, as the device keys for the sender's device.  In other words, the
 plaintext payload will now look something like:
@@ -75,7 +75,7 @@ plaintext payload will now look something like:
   "keys": {
     "ed25519": "<sender_ed25519_key>"
   },
-  "device_keys": {
+  "sender_device_keys": {
     "algorithms": ["<supported>", "<algorithms>"],
     "user_id": "<user_id>",
     "device_id": "<device_id>",
@@ -94,7 +94,7 @@ plaintext payload will now look something like:
 ```
 
 If this property is present, the `keys`.`ed25519` property of the plaintext
-payload must be the same as the `device_keys`.`keys`.`ed25519:<DEVICEID>`
+payload must be the same as the `sender_device_keys`.`keys`.`ed25519:<DEVICEID>`
 property.  If they differ, the recipient should discard the event.
 
 As the `keys` property is now redundant, it may be removed in a future version
@@ -119,7 +119,7 @@ information to the user; that is left for the future.
 
 ### Minor variations
 
-The `device_keys` property could be added to the cleartext.  That is, it could
+The `sender_device_keys` property could be added to the cleartext.  That is, it could
 be added as a property to the `m.room.encrypted` event.  This information is
 already public, as it is accessible from `/keys/query` (while the device is
 logged in), and does not need to be authenticated as it is protected by the
@@ -129,14 +129,14 @@ including it in the encrypted payload, it leaves open the possibility of
 it replacing the `keys` property, which must be part of the encrypted payload
 to prevent an [unknown key-share attack](https://github.com/element-hq/element-web/issues/2215).
 
-The `device_keys` property could be added to the cleartext by the sender's
+The `sender_device_keys` property could be added to the cleartext by the sender's
 homeserver, rather than by the sending client. Possibly within an `unsigned`
 property, as that is where properties added by homeservers are customarily
 added.  It is not clear what advantage there would be to having this
 information being added by the client.
 
 To mitigate the increased size of to-device events under this proposal, the
-`device_keys` could be sent only in pre-key messages (Olm messages
+`sender_device_keys` could be sent only in pre-key messages (Olm messages
 with `type: 0` in the `m.room.encrypted` event) — with the rationale that if
 the Olm message is a normal (non-pre-key) message, this means that the
 recipient has already decrypted a pre-key message that contains the
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -75,7 +75,7 @@ plaintext payload will now look something like:
     "ed25519": "<sender_ed25519_key>"
   },
   "device_keys": {
-    "algorithms": ["<supported>", "<algoriithms>"],
+    "algorithms": ["<supported>", "<algorithms>"],
     "user_id": "<user_id>",
     "device_id": "<device_id>",
     "keys": {
@@ -124,9 +124,12 @@ information to the user; that is left for the future.
 The `device_keys` property could be added to the cleartext.  That is, it could
 be added as a property to the `m.room.encrypted` event.  This information is
 already public, as it is accessible from `/keys/query` (while the device is
-logged in), and does not need to be authenticated as it is protected by .the
-self-signing signature, so it does not seem to need to be encrypted.  However,
-there seems to be little reason not to encrypt the information.
+logged in), and does not need to be authenticated as it is protected by the
+self-signing signature, so it does not seem to need to be encrypted. However,
+there seems to be little reason not to encrypt the information. In addition, by
+including it in the encrypted payload, it leaves open the possibility of
+it replacing the `keys` property, which must be part of the encrypted payload
+to prevent an [unknown key-share attack](https://github.com/element-hq/element-web/issues/2215).
 
 The `device_keys` property could be added to the cleartext by the sender's
 homeserver, rather than by the sending client.  Possibly within an `unsigned`
--- a/proposals/4147-including-device-keys-with-olm-encrypted-events.md
+++ b/proposals/4147-including-device-keys-with-olm-encrypted-events.md
@@ -1,15 +1,16 @@
 # MSC4147: Including device keys with Olm-encrypted events
 
-Summary: a proposal to ensure that messages sent from short-lived devices can
-be securely distinguished from a spoofed device.
+Summary: a proposal to ensure that messages sent from a short-lived (but
+genuine) device can be securely distinguished from those sent from a spoofed
+device.
 
 ## Background
 
 When a Matrix client receives an encrypted message, it is necessary to
 establish whether that message was sent from a device genuinely belonging to
 the apparent sender, or from a spoofed device (for example, a device created by
-an attacker with access to the sender's account such as the server admin, or a
-man-in-the-middle).
+an attacker with access to the sender's account such as a malicious server
+admin, or a man-in-the-middle).
 
 In short, this is done by requiring a signature on the sending device's device
 keys from the sending user's [self-signing cross-signing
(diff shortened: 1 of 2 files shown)


proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
+# MSC4147: Including device keys with Olm-encrypted events
+
+When a Megolm session is sent from one device to another via Olm, the recipient
+can
+[query](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3keysquery)
```
turt2live, line 5: versioned links throughout please :)

(unstable and latest do not guarantee that links will work over time)

proposals/4147-including-device-keys-with-olm-encrypted-events.md
turt2live, line 1: Implementation requirements:

* [ ] Client sending this information
* [ ] Client using this information
* [ ] Demonstration of the core problem being solved (likely covered by the above tasks)

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+        "ed25519:<ssk_id>": "<ssk_signature>",
+      }
+    }
+  }
+}
+```
+
+If this property is present, the `keys`.`ed25519` property of the plaintext
+payload must be the same as the `device_keys`.`keys`.`ed25519:<DEVICEID>`
+property.
```
clokep, line 75: What should the client do if they're different? Reject the event?

What if the `device_keys` property doesn't match what the server would return? What sorts of issues could this cause?

uhoreg: Yes, if they're different, the client should reject.  I'll add that to the text.

If the `device_keys` property doesn't match what the server returns, I think it depends on how they differ, as some changes could be expected. e.g. if a user rotates their cross-signing keys, then the `signatures` will be different.  But I think that the short answer is that the client can more-or-less do the same thing it would do if it got two different responses to `/keys/query` for the same device.

clokep: That makes sense. I guess I was wondering if a rogue device could do something odd, but a device can already set their server keys so that doesn't make much sense.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+  "keys": {
+    "ed25519": "<sender_ed25519_key>"
+  },
+  "device_keys": {
+    "algorithms": ["<supported>", "<algoriithms>"],
+    "user_id": "<user_id>",
+    "device_id": "<device_id>",
+    "keys": {
+      "ed25519:<device_id>": "<our_ed25519_key>",
+      "curve25519:<device_id>": "<our_curve25519_key>"
```
richvdh, line 61: note that `our_ed25519_key` is used to refer to the *recipient's* ed25519 key further up. I think you mean:

```suggestion
      "ed25519:<device_id>": "<sender_ed25519_key>",
      "curve25519:<device_id>": "<sender_curve25519_key>"
```

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+```
+
+If this property is present, the `keys`.`ed25519` property of the plaintext
+payload must be the same as the `device_keys`.`keys`.`ed25519:<DEVICEID>`
+property.  If they differ, the recipient should discard the event.
+
+As the `keys` property is now redundant, it may be removed in a future version
+of Olm.
```
richvdh, line 78:
```suggestion
of the Matrix specification.
```

This is nothing to do with Olm, afaict?

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+not be able to query the sender's device keys.  For example, this can happen if
+the recipient is offline for a long time.
+
+One way to solve this is to include a copy of the device keys in the
+Olm-encrypted message, along with the cross-signing signatures, so that the
+recipient does not have to try to query the sender's keys.
+
+## Proposal
+
+The plaintext payload of the [`m.olm.v1.curve25519-aes-sha2` encryption
```
richvdh, line 18:
```suggestion
The plaintext payload of `m.room.encrypted` events encrypted with the [`m.olm.v1.curve25519-aes-sha2` encryption
```

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+much benefit.
+
+## Alternatives
+
+The `device_keys` property could be added to the cleartext.  That is, it could
+be added as a property to the `m.room.encrypted` event.  This information is
+already public, as it is accessible from `/keys/query` (while the device is
+logged in), and does not need to be authenticated as it is protected by .the
+self-signing signature, so it does not seem to need to be encrypted.  However,
+there seems to be little reason not to encrypt the information.
```
richvdh, line 111: I feel there was some cryptographic reason why `"keys": { "ed25519": "<sender_ed25519_key>" }` had to go inside the encrypted part of the events, and, particularly if we hope to get rid of `keys` at some point, that presumably also applies here. It might be good to fire up the time machine and see if we can remember what that was all about. (Let me know if you draw blanks/think it was before your time)

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
+# MSC4147: Including device keys with Olm-encrypted events
```
ara4n, line 1: Are we sure we want to do this? It feels like a very blunt solution to the problem, and duplicating per-device metadata on every event ever sent feels very inefficient.

What about letting the origin server keep publishing deleted devices in /keys/query instead? After all, if the server or account has entirely gone away (or reset its cross-signing keys), then we won't be able to check the signatures meaningfully anyway. So we might as well rely on the origin server to keep publishing the keys for us, rather than cluttering every single event with them?

richvdh: > After all, if the server or account has entirely gone away (or reset its cross-signing keys), then we won't be able to check the signatures meaningfully anyway. 

I'm not sure this is quite right. We might well have the user's cross-signing key even if the server is offline?

Also: to-device messages are necessarily ephemeral. Once they are delivered, we can throw away the data associated with them. You're saying that servers would have to keep records of all devices ever created by all users, which doesn't sound like a long-term win.

ara4n: > I'm not sure this is quite right. We might well have the user's cross-signing key even if the server is offline?

true, i guess we might have a cache.

> You're saying that servers would have to keep records of all devices ever created by all users, which doesn't sound like a long-term win.

yup, precisely - at least available for querying by an id (no need to publish them all in /keys/query).

This feels much more preferable than duplicating the same data on every event ever sent by a user on the server, which would then also effectively be keeping records of all devices ever created by all users.

Practically, this is going to roughly double the size of all events we have to sync, which is going to slow down EX sync in network-bounded environments by 2x - which is unfortunate if it's avoidable.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+
+When a Megolm session is sent from one device to another via Olm, the recipient
+can
+[query](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3keysquery)
+the sender's device keys and check whether the device has been cross-signed in
+order to determine whether the sending device can be trusted.  However, this
+does not work if the sending device has since logged out as the recipient will
+not be able to query the sender's device keys.  For example, this can happen if
+the recipient is offline for a long time.
```
BillCarsonFr, line 10: Maybe we can be more specific that this is an attempt to fix the `sent by a deleted or unknown device`.
Currently when a message is sent from short lived device (log in, send one message then log out), there is no way for the recipient to establish authenticity of this message. It could have been a MITM attack with the server creating a one-time device to send a message.
As part of invisible crypto, messages which authenticity cannot be guaranteed will be ignored instead of shown with a red warning. So we need a way to make the difference between a short-lived legit device and a one shot MITM.
Notice that this fixes the problem of deleted devices if they were properly cross-signed, but it would still be possible to have messages sent from an unknown or deleted identity, but identities have a longer lifespan than devices and this would happen less oftent

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,135 @@
...
+already public, as it is accessible from `/keys/query` (while the device is
+logged in), and does not need to be authenticated as it is protected by .the
+self-signing signature, so it does not seem to need to be encrypted.  However,
+there seems to be little reason not to encrypt the information.
+
+The `device_keys` property could be added to the cleartext by the sender's
+homeserver, rather than by the sending client.  Possibly within an `unsigned`
+property, as that is where properties added by homeservers are customarily
+added.  It is not clear what advantage there would be to having this
+information being added by the client.
```
BillCarsonFr, line 117: The is an other alternative: have homeservers remember deleted devices (and deleted identities?)

proposals/4147-including-device-keys-with-olm-encrypted-events.md
uhoreg: The Element crypto team has decided to implement this MSC as it currently is, in order to get real-world data on how much it affects performance.  If it affects performance too much, we will look at other options such as only sending the device keys in pre-key messages.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
uhoreg: This has been implemented in the Rust `matrix-sdk-crypto` crate, both sending and using.  Demonstration of the problem being solved: https://youtu.be/b1jJlT2ENT8?feature=shared&t=345

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: 04acb6b4

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: hopefully clarified in 407e1d91

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: "keys" was added to solve https://github.com/element-hq/element-web/issues/2215, to stop Eve simply forwarding an Olm message created by Alice, and passing it off as her own. Put in that light, it's clear that `keys` must go in the encrypted body, and, if we hope `device_keys` will replace it someday, so must it.

richvdh: documented in 5bb2093a

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: Documented this alternative in 0deb58f0.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: A few links to the implementation.

On the sender side, the field is added [here](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-crypto/src/olm/session.rs#L159). 

On the receiver side, it is deserialized after decryption into a [DecryptedOlmV1Event](https://github.com/matrix-org/matrix-rust-sdk/blob/70bcddfba5e19f2d0fd11577edb55b9868f89678/crates/matrix-sdk-crypto/src/types/events/olm_v1.rs#L178-L180). It is then used in [SenderDataFinder](https://github.com/matrix-org/matrix-rust-sdk/blob/70bcddfba5e19f2d0fd11577edb55b9868f89678/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data_finder.rs#L174) whose job it is to construct a [SenderData](https://github.com/matrix-org/matrix-rust-sdk/blob/70bcddfba5e19f2d0fd11577edb55b9868f89678/crates/matrix-sdk-crypto/src/olm/group_sessions/sender_data.rs#L49) which records our knowledge about the sending device for a given session.

richvdh:
I think this is ready for FCP, so:

@mscbot fcp merge

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+  "recipient_keys": {
+    "ed25519": "<our_ed25519_key>"
+  },
+  "keys": {
+    "ed25519": "<sender_ed25519_key>"
+  }
+}
+```
+
+We propose to add a new property: `device_keys`, which is a copy of what the
```
richvdh, line 60: I did wonder if we could improve the name here; in particular, to reflect the fact that these are the keys for the *sender* device.

```suggestion
We propose to add a new property: `sender_device_keys`, which is a copy of what the
```

turt2live:
@mscbot concern implementation-needs-checking label

(the needs-implementation label was not removed when the FCP was proposed or implementations listed, so I am assuming they have not been checked out of an abundance of caution)

richvdh:
I've looked at the implementation, and checked that it matches the proposal to the best of my ability.

turt2live:
@mscbot resolve implementation-needs-checking label

proposals/4147-including-device-keys-with-olm-encrypted-events.md
uhoreg: In the payload, we have also have `recipient_keys` and `keys` properties.  Given that we've raised the possibility that `keys` could be removed in a future version, it seems like `sender_device_keys` would provide some symmetry with `recipient_keys`.

dbkr:
Thanks for writing up the alternative of keeping deleted devices in the device list: this otherwise would have been my question.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: Let's commit to `sender_device_keys`, then

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+there seems to be little reason not to encrypt the information. In addition, by
+including it in the encrypted payload, it leaves open the possibility of
+it replacing the `keys` property, which must be part of the encrypted payload
+to prevent an [unknown key-share attack](https://github.com/element-hq/element-web/issues/2215).
+
+The `sender_device_keys` property could be added to the cleartext by the sender's
+homeserver, rather than by the sending client. Possibly within an `unsigned`
+property, as that is where properties added by homeservers are customarily
+added.  It is not clear what advantage there would be to having this
+information being added by the client.
```
clokep, line 136: Wouldn't this reduce the storage size significantly?

ara4n:
For the record, I am still not happy with the extreme duplication of the denormalised device data per-event, rather than keeping deleted devices around device lists until their to-device msgs have been exhausted.

On the other hand, the current behaviour amounts to a security issue, and in practice this has been already implemented and shown working in the Element clients.  So I'm going to approve FCP in order that we ship the fix rapidly for everyone, and am going to pray that someday we remove the denormalisation again (perhaps around the same time as senders-as-keys).

Feels really wrong though, and it's a bad smell this didn't surface before implementation began.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+property.  If they differ, the recipient should discard the event.
+
+As the `keys` property is now redundant, it may be removed in a future version
+of the Matrix specification.
+
+## Potential issues
+
+Adding this property will increase the size of the event.  We found it
+increased the length of a typical `m.room_key` message from about 1400 to 2400
+bytes (a 70% increase). This will require increased storage on the recipient
```
clokep, line 107: I'm confused, but maybe due to my lack of remembering the flow of encryption -- isn't this suggesting modification of `m.room.encrypted` events? Can you clarify the relation to `m.room_key`?

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+it replacing the `keys` property, which must be part of the encrypted payload
+to prevent an [unknown key-share attack](https://github.com/element-hq/element-web/issues/2215).
+
+The `sender_device_keys` property could be added to the cleartext by the sender's
+homeserver, rather than by the sending client. Possibly within an `unsigned`
+property, as that is where properties added by homeservers are customarily
+added.  It is not clear what advantage there would be to having this
+information being added by the client.
+
+To mitigate the increased size of to-device events under this proposal, the
```
clokep, line 138: Again I'm a bit confused -- this proposal doesn't talk about modifying to-device messages?

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+in the to-device messages is found to be significant, it would be worth
+revisiting this.
+
+## Security considerations
+
+If a device is logged out, there is no indication why it was logged out.  For
+example, an attacker could steal a device and use it send a message.  The user,
+upon realizing that the device has been stolen, could log out the device, but
+the message may still be sent, if the user does not notice the message and
+redact it.
```
clokep, line 174: Can you spell out the impact of this? I think it is that by now including the keys the messages will be "trusted" and logging out the device won't make the messages automatically untrusted?

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+with `type: 0` in the `m.room.encrypted` event) — with the rationale that if
+the Olm message is a normal (non-pre-key) message, this means that the
+recipient has already decrypted a pre-key message that contains the
+information, and so does not need to be re-sent the information), or if the
+signatures change (for example, if the sender resets their cross-signing keys),
+or if the sender has not yet sent their `device_keys`.  However, this requires
+additional bookkeeping, and it is not clear whether this extra complexity is
+worth the reduction in bandwidth.
+
+### Alternative approach
```
clokep, line 149: Is persisting info into rooms as a permanent record another alternative? Does that even make sense?

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: Ah yeah, we're using "event" when really we mean "to-device message". I'll go through and change that, for clarity.

Anyway: `m.room.encrypted` to-device messages are used to carry an encrypted `m.room_key` payload. So this should probably be:

```suggestion
Adding this property will increase the size of the to-device message.  We found it
increased the length of a typical encrypted `m.room_key` message from about 1400 to 2400
bytes (a 70% increase). This will require increased storage on the recipient
```

proposals/4147-including-device-keys-with-olm-encrypted-events.md
clokep: Fair enough about having an explicit ability for that, just trying to understand impact!

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: Hopefully this is clearer now.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
richvdh: Yeah, you could make room members be devices, rather than users (MLS-style); as part of that, you could embed the device keys/signatures in the state events. It's something we've considered in the past (it helps with account portability, among other things); however there are a lot of moving parts to get it right, and having to send a membership event to every room you're in every time you log in is a bit of a killer.

It would really be a radically different approach, whilst this MSC is more of an incremental improvement.

clokep:
Thanks for answering all my questions @richvdh -- I think this is OK as an incremental improvement, but sounds like we might need to do a bigger improvement at some point in the future.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
clokep: Thanks! Not sure if this should be included or not as an alternative.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
uhoreg: > Can you spell out the impact of this? I think it is that by now including the keys the messages will be "trusted" and logging out the device won't make the messages automatically untrusted?

Yes, that's the issue.

I think that at some point, we want to have a mechanism for users to indicate that a device was compromised, and that messages from it shouldn't be trusted.  But that would be a separate MSC and for now, users can probably do so manually by sending messages to rooms.

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+the time the recipient queries for it. This makes the received message
+indistinguishable from one sent from a spoofed device.
+
+Current implementations work around this by displaying a warning such as "sent
+by a deleted or unknown device" against the received message, but such
+messaging is unsatisfactory: a message should be either trusted or not.
+
+We propose to solve this by including a copy of the device keys in the
+Olm-encrypted message, along with the cross-signing signatures, so that the
+recipient does not have to try to query the sender's keys.
```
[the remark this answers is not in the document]
richvdh: That was very much *not* part of this MSC. It's something we could do in future, but IMHO it would need another MSC, and careful thought, since it would mean cutting off the significant part of the ecosystem which doesn't yet implement *this* MSC.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2122

proposals/4147-including-device-keys-with-olm-encrypted-events.md
```diff
@@ -0,0 +1,183 @@
...
+        "ed25519:<ssk_id>": "<ssk_signature>",
+      }
+    }
+  }
+}
+```
+
+If this property is present, the `keys`.`ed25519` property of the plaintext
+payload must be the same as the `sender_device_keys`.`keys`.`ed25519:<DEVICEID>`
+property.  If they differ, the recipient should discard the event.
```
richvdh, line 98: We don't seem to have specified anything here about `sender_device_keys`.`keys`.`curve25519:<DEVICEID>`, which seems like an oversight: if nothing else, it allows inconsistency within the event which is generally a big red flag in encryption stuff.

The sample implementation of this (matrix-rust-sdk) does check that `curve25519:<DEVICEID>` matches the olm session used to encrypt the message (https://github.com/matrix-org/matrix-rust-sdk/blob/4d027ec405a20b54c6f9d2e31a20bf2dd1c3854d/crates/matrix-sdk-crypto/src/olm/account.rs#L1539-L1548); I'm inclined to consider it an oversight in the MSC.

cc @uhoreg

richvdh:
Spec PR is now merged :tada:

PR #4149, opened 2024-06-04
Update CSP Directives for Media Repository

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/csp/proposals/9999-media-repo-csp.md)

Signed-off-by: Tom Foster [tom@tcpip.uk](mailto:tom@tcpip.uk)


proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `object-src` directive pertains to legacy web plugins, which are deprecated. The use of
+`<object>` elements is being reconsidered for removal, rendering this directive obsolete.
+
+### Proposed CSP Directive
+
+After the proposed changes, the updated CSP directive for the media repository would be:
+
+```plaintext
+Content-Security-Policy: default-src 'self'; style-src 'self';
```
girlbossceo, line 55: The most strictest CSP should actually be what conduwuit is using: https://github.com/girlbossceo/conduwuit/commit/55ebc311356ddb5a0cab9197c602871c44a3eba9#diff-f1f2f943ea9e58c1e23398d14b71ad683d6818d70db8f1d8251eab491d49303fR24

`sandbox; default-src 'none'; font-src 'none'; script-src 'none'; frame-ancestors 'none'; form-action 'none'; base-uri 'none';`

The spec recommended one is `sandbox; default-src 'none'; script-src 'none'; plugin-types application/pdf; style-src 'unsafe-inline'; object-src 'self';`, but I found some additional directives that do not use `default-src` and require explicit set to none.

- `frame-ancestors 'none'` sets the modern CSP-equivalent of `X-Frame-Options: Deny`
- `base-uri 'none'` sets the `<base>` HTML tag to nothing, if somehow inline HTML was rendered and tried to use `<base>`.
- `form-action 'none'` disallows form POST requests anywhere

The list of CSP directives can be found at https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#browser_compatibility

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,88 @@
...
+
+The current Content Security Policy (CSP) directives recommended for the media repository in the
+Matrix specification contain outdated and potentially insecure directives. This proposal aims to
+update these directives to enhance security and align with modern web standards.
+
+The issues with the existing directives are as follows:
+
+1. `plugin-types application/pdf;` is a legacy directive that modern browsers do not use, and PDFs
+   are not allowed as per MSC2702.
+2. `style-src 'unsafe-inline';` allows inline CSS, which poses security risks.
```
girlbossceo, line 13: `unsafe-inline` is also only relevant for HTML, which MSC2702 disallows rendering inline.

proposals/4149-media-repo-csp-directives.md
girlbossceo: This would also mean, updating the proposal section to not only remove the legacy directives, but also add additional ones that are more strict.

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,88 @@
...
+After the proposed changes, the updated CSP directive for the media repository would be:
+
+```plaintext
+Content-Security-Policy: default-src 'self'; style-src 'self';
+```
+
+## Potential issues
+
+Updating CSP directives could potentially cause issues for implementations that rely on the
+outdated directives. However, these changes should not adversely impact existing implementations as
```
girlbossceo, line 61: It's also worth noting that the CSP only affects the *content being loaded*, not *what* is loading the content. Clients can still inline render media and such as normal, the CSP will just forbid (theoretically) an image making network requests.

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,88 @@
...
+One alternative is to maintain the current CSP directives, accepting the associated security risks
+and reliance on outdated web standards. However, this approach would not align with the goal of
+improving the security posture of the Matrix media repository.
+
+Another alternative could be to adopt a more permissive CSP, but this would compromise security and
+increase the risk of XSS attacks and other vulnerabilities.
+
+## Security considerations
+
+Removing outdated directives and disallowing inline styles reduces the attack surface and mitigates
```
girlbossceo, line 76: While it does reduce browser attack surface and encourages browsers to drop these legacy directives (based on global web usage), I think I personally would remove the "mitigates XSS" stuff. The `sandbox` directive, the existing `default-src 'none'` directive, and homeservers implementing MSC2702 practically eliminate the concerns for XSS. `sandbox` directive on its own disallows same-origin requests, and `default-src 'none'` will remove a ton of attack surface.

proposals/4149-media-repo-csp-directives.md
tcpipuk: Sorry, you're absolutely right! I made some silly assumptions. I've just uploaded a copy that not only removes the unsafe fields, but adds the extra ones you're currently using with a summary to explain why. Thank you!

proposals/4149-media-repo-csp-directives.md
tcpipuk: Ah, do you think it's worthwhile I update to clarify [MSC2702](https://github.com/matrix-org/matrix-doc/pull/2702) is also what makes the `style-src` redundant?

proposals/4149-media-repo-csp-directives.md
tcpipuk: Fair point, I wasn't really making an argument that there is currently a huge security vulnerability, but more that the suggestions reduce the risk of any potential (current or future) bugs in clients. Does the wording in the new version sound better?

proposals/4149-media-repo-csp-directives.md
girlbossceo: Yep that sounds good. As far as I can tell, the original reasoning for these CSP directives were chosen when inline HTML was technically allowed, and these directives were a way of encouraging homeservers to reduce the attack surface from that by sending a strict CSP, but it is only a "should", and previously would be "okay" for a server to allow arbitrary user-uploaded inline HTML rendering without a CSP.

proposals/4149-media-repo-csp-directives.md
girlbossceo: I think this sounds better to me, yes.

proposals/4149-media-repo-csp-directives.md
girlbossceo: Yes this should be good to me

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,101 @@
+# MSC4149: Update CSP Directives for Media Repository
```
girlbossceo, line 1: @clokep said as a recommendation to provide some CSP resources that aren't just vendored here.

https://infosec.mozilla.org/guidelines/web_security#content-security-policy - CSP recommendations by Mozilla Security
https://internet.nl/faqs/appsecpriv/ - internet.nl's website security tester on CSP
https://csp-evaluator.withgoogle.com/ - Google's CSP validator
https://xmpp.org/extensions/xep-0363.html#server - XMPP's XEP-0363 on CSP

tcpipuk: Good shout. Want me to quote them now, or wait for @clokep to confirm if these are the sort of references he was looking for?

turt2live, line 1: Implementation requirements:

* [ ] Clients demonstrating near-zero impact.
* [ ] 3D Worlds clients (like Third Room) continue to function.
* [ ] Review from the Matrix Security Team (will be requested closer to FCP)

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,101 @@
...
+### Details of the Proposal
+
+#### Remove `plugin-types application/pdf;`
+
+Modern browsers no longer use the `plugin-types` directive. It was originally intended for use with
+legacy plugins such as those for PDF viewing, which are no longer common practice. Furthermore,
+[MSC2702](https://github.com/matrix-org/matrix-doc/pull/2702) explicitly disallows PDFs, making
+this directive unnecessary and potentially misleading.
+
+#### Remove `style-src 'unsafe-inline';`
```
ara4n, line 34: I have a nagging feeling that this was included for a good reason - perhaps the ability to share all-in-one HTML files and have the style render correctly when viewed in browser (rather than having to mess around uploading the CSS for the HTML to a separate repo location, and then link to it, when sharing HTML)? But this is a wild guess; i can't remember the specifics.

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+### Details of the Proposal
+
+#### Remove `plugin-types application/pdf;`
+
+Modern browsers no longer use the `plugin-types` directive. It was originally intended for use with
+legacy plugins such as those for PDF viewing, which are no longer common practice. Furthermore,
+[MSC2702](https://github.com/matrix-org/matrix-doc/pull/2702) explicitly disallows PDFs, making
+this directive unnecessary and potentially misleading.
```
turt2live, line 32: MSC2702 doesn't require a server to follow this guidance, and some may still permit PDFs to be embedded. Clients may further iframe or otherwise embed the PDF inline to the chat, ignoring some of the safety control headers (at least where they can - web browser clients will find this difficult). 

The proposal to remove the directive may still be valid, though the language around MSC2702 is questionable.

proposals/4149-media-repo-csp-directives.md
girlbossceo: I believe it was for when inline HTML was allowed before MSC2702, in general: https://github.com/matrix-org/matrix-spec-proposals/pull/4149#discussion_r1626529386

proposals/4149-media-repo-csp-directives.md
```diff
@@ -0,0 +1,101 @@
...
+```plaintext
+sandbox; default-src 'none'; script-src 'none'; font-src 'none';
+frame-ancestors 'none'; form-action 'none'; base-uri 'none';
+```
+
+### Details of the Proposal
+
+#### Remove `plugin-types application/pdf;`
+
+Modern browsers no longer use the `plugin-types` directive. It was originally intended for use with
```
turt2live, line 29: "Modern browsers" feels like it's answering a question we haven't asked in the spec yet: at what point do we consider a client "unsupported"? Currently, the CSP is intended to protect old Internet Explorer users from various attack vectors, though even with that there's still holes. Should we be dropping support for IE?

proposals/4149-media-repo-csp-directives.md
tcpipuk: I've added these as references for now, thanks!

proposals/4149-media-repo-csp-directives.md
ara4n: sounds plausible. i think the motivation here was to ensure that HTML and PDFs loaded straight into browser tabs from web clients (rather than ending up annoying saved to disk, and having to be manually viewed from there). However, the shift away from web clients to Electron etc makes that much more moot.

proposals/4149-media-repo-csp-directives.md
tcpipuk: What sort of client testing would you like to see? I can override the CSP in my reverse proxy to prove certain things work correctly in a number of clients if needed?

I'm not sure how the change from the current recommended CSP would affect Third Room, as it's already set to `default-src 'none';` and it's unclear whether it requires CSS? Is there someone with knowledge of Third Room's requirements that could weigh-in? I don't know if I have the necessary hardware/software to test this myself.

proposals/4149-media-repo-csp-directives.md
tcpipuk: As the recommended CSP in the spec, I'd assume we want to _recommend_ people can't load AIO HTML files directly, but that server administrators might want to choose to ignore the recommended CSP if they have specific requirements?

proposals/4149-media-repo-csp-directives.md
tcpipuk: As this is a _recommended_ CSP, should I continue to use this, but simply recommend the spec clarify some circumstances you may want to extend the recommended value?

proposals/4149-media-repo-csp-directives.md
girlbossceo: I feel like if Internet Explorer support was a target demographic, there would be recommendations of some more ancient headers like `X-Download-Options`,`X-Permitted-Cross-Domain-Policies`, and `X-XSS-Protection: 1; mode=block` (causes vulnerabilities in modern browsers).

Though, my honest answer to this:

> Should we be dropping support for IE?

Is yes. [Microsoft has dropped support for it in 2022](https://learn.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge) and perhaps when a vendor/OEM drops general public support for it, that should be a hard limit on when a browser or client is officially considered "unsupported". Other factors could be argued, but if the developers themselves say it's unsupported then it's unsupported full stop.

tcpipuk: Given that [Microsoft ended support for IE in June 2022](https://support.microsoft.com/en-us/windows/internet-explorer-help-23360e49-9cd3-4dda-ba52-705336cc0de2) I'd say it's reasonable to deprecate support for a browser that hasn't had security updates in two years.

I'm not sure it's a question that needs to be answered during this MSC though - the current CSP already blocks scripts, for example. We're simply saying if scripts are considered unsafe in uploaded media, then it'd be safer to include plugins, CSS, and fonts in that too

proposals/4149-media-repo-csp-directives.md
tcpipuk: In the current case, someone can upload a HTML file that uses formatting tags, but should we be allowing the full gamut of CSS just in case someone wants to view full HTML pages directly from the media repo?

proposals/4149-media-repo-csp-directives.md
turt2live: The question is more about what the protocol's position is on supporting things like IE. That conversation doesn't need to happen here on the MSC, but I think it does block the MSC.

proposals/4149-media-repo-csp-directives.md
turt2live: At a quick glance the MSC's language for the proposed change is fine, it's just text like "MSC2702 explicitly disallows PDFs" which has me uncomfortable. It currently reads as though MSC2702 forces a change when it doesn't.

Other references to MSC2702 throughout the rest of the MSC appear to have the same concern.

proposals/4149-media-repo-csp-directives.md
turt2live: somewhat related: previews like that are going to break with MSC3916 (unless clients invent local proxies using service workers)

proposals/4149-media-repo-csp-directives.md
turt2live: This sort of client testing would ideally come from the developers themselves as they will be more aware of the nuances for their users. For Third Room, considering the development team is largely disbanded, the scene/world should load without issue.

proposals/4149-media-repo-csp-directives.md
tcpipuk: Given that [MSC3916](https://github.com/matrix-org/matrix-doc/pull/3916) is in FCP, does that mean this is considered a non-issue, or do I need to plan for the potential workarounds?

proposals/4149-media-repo-csp-directives.md
tcpipuk: Got it, thanks. Is the new wording suitable?

proposals/4149-media-repo-csp-directives.md
tcpipuk: Understood. How should this be raised? Is this something Josh and/or the governing board should discuss, or something that would be agreed within the SCT? What can I do to help that conversation happen?

turt2live: This thread is currently hosting the conversation - leaving it here is best for now in my opinion, though opening an issue on matrix-spec and linking it in this thread would also be fine. 

The SCT will provide the bulk of the opinion, though will be shaped by the community's response as well. This should happen as part of normal MSC review.

proposals/4149-media-repo-csp-directives.md
tcpipuk: Understood, I suppose, given the (fairly subtle) nature of the change, any clients affected would probably be more unusual ones... is the idea that I'd be seeking confirmation from more common client developers that this update to responses to the media repository doesn't reveal any unexpected behaviour in their clients?

tcpipuk: Having discussed with Nico (of [Nheko](https://github.com/Nheko-Reborn/nheko)) it appears he doesn't embed a web browser, and the client downloads other media to play locally, so doesn't use/respect CSPs for media requests.

I assume it will be a similar story on Android/iOS apps, so testing will primarily need to focus on web-based/Electron clients.

proposals/4149-media-repo-csp-directives.md
girlbossceo: Just for reference when the discussions happen here:

The spec recommended ("should") CORS header `Cross-Origin-Resource-Policy` on the media content repo is not an IE-supported header.

Additionally, CSP on IE is only supported by the legacy `X-Content-Security-Policy` header. So the existing CSP would not do anything on Internet Explorer as they would need to send the `X-` prefixed header.

https://en.wikipedia.org/wiki/Content_Security_Policy#Status

> Internet Explorer 10 and Internet Explorer 11 also support CSP, but only sandbox directive, using the experimental X-Content-Security-Policy header.

tcpipuk:
Is this MSC considered redundant now [MSC3916](https://github.com/matrix-org/matrix-doc/pull/3916) has been merged, or should we consider progressing as servers still support the deprecated unauthenticated media endpoints?

proposals/4149-media-repo-csp-directives.md
tcpipuk: Just added a comment to this effect at the bottom, but is this MSC considered redundant now? I'm not aware of a timeline for the deprecated endpoints to be retired yet, so should we look to get this in place to secure them in the meantime?

turt2live:
The authenticated endpoints could still end up needing a CSP depending on how they're used - updating it doesn't seem like a bad idea.

turt2live:
(please also avoid force-pushing after review has been given. It makes it harder to see what's changed, and we squash commits anyways)

PR #4150, opened 2024-06-04
MSCXXXX: m.allow recommendation for moderation policy lists

Johennes:
Implementations:

* https://github.com/matrix-org/mjolnir/blob/5e35efd1dbc0097a7c19bed2831a1308a29d7be7/src/models/ListRule.ts#L63 & https://github.com/matrix-org/mjolnir/blob/5e35efd1dbc0097a7c19bed2831a1308a29d7be7/src/models/AccessControlUnit.ts#L266


proposals/4150-m-allow-recommendation.md
turt2live, line 1: Implementation requirements:

* [ ] Publisher of new recommendation
* [ ] Consumer of new recommendation

Johennes: Doesn't mjolnir satisfy both of these already?

turt2live: Very possibly, but the implementations would need to be reviewed. Can the original PRs rather than commits be linked on this thread please?

proposals/4150-m-allow-recommendation.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Security considerations
+
+None on top of what's already [listed] in the spec.
+
+
+## Unstable prefix
+
+Until this proposal is accepted into the spec, implementations should refer to `m.allow` as
+`org.matrix.msc4150.allow`.
```
FSG-Cat, line 51: As long as this proposal does not deviate from the behavior that is shared by Mjolnir and Draupnir i would recomed that the prefix remains as the one Draupnir and Mjolnir uses as its what exists in the wild already. (Unless there is a process reason against this?)

While i cant speak to Mjolnir for all as i have never run a true mjolnir for all i can speak as to how this is used on the admin side of Draupnir for all. Its used to create the allow list for who can ask for a bot and said policy list is against better judgement stored in the admin room for now.

proposals/4150-m-allow-recommendation.md
Gnuxie: Draupnir publishes the `org.matrix.mjolnir.allow` policies in the appservice admin room to designate the users that have access [here](https://github.com/the-draupnir-project/Draupnir/blob/27e058711ebf9414b9db258b180ec1337b7e6296/src/appservice/AccessControl.ts#L76) via [this command](https://github.com/the-draupnir-project/Draupnir/blob/27e058711ebf9414b9db258b180ec1337b7e6296/src/appservice/bot/AccessCommands.tsx#L13-L27).

Draupnir consumes them [here](https://github.com/the-draupnir-project/Draupnir/blob/27e058711ebf9414b9db258b180ec1337b7e6296/src/appservice/AppServiceDraupnirManager.ts#L169-L172) to enforce access upon invitation of the appservice bot via the same [access code as mjolnir](https://github.com/Gnuxie/matrix-protection-suite/blob/7fbf691f87056e01de7175b5322b25a901311409/src/Protection/AccessControl.ts#L94-L124).

However, this implementation exists explicitly with the intent that `org.matrix.mjolnir.allow` is only used to describe an entity that is allowed to participate to begin with. That's to say the absence of a policy rule with an allow recommendation for any entity type then implies that there is a policy rule for that entity type with the `entity` field that has the glob value `"*"`. Which is a long winded way of saying this is explicitly not an an implementation of a recommendation that performs anti-ban and is not the opposite to `m.ban`. This is implemented anonymously with `allow` from the `m.room.server_acl` algorithm.

proposals/4150-m-allow-recommendation.md
```diff
@@ -0,0 +1,70 @@
...
+## Potential issues
+
+When both `m.ban` and `m.allow` recommendations are used in the same policy list, confusion could arise if
+two rules with opposing recommendations match against an entity. Given that the spec doesn't currently
+impose _any_ logic for how policy lists are to be interpreted, it could well leave this question as an
+implementation detail, too.
+
+However, if a resolution is indeed desired, the spec could mandate that `m.ban` rules override `m.allow`
+rules but not the other way around. This matches both [Mjolnir's implementation] and the way
+[server access control lists] are evaluated.
```
Gnuxie, line 35: I think that we would need to explicitly choose one or the other here to prevent future clients from implementing them both, and then failing spectacularly when someone mixes them.

Separately, anti-ban is risky and there are other ways of "undoing" a ban (by redacting the policy or replacing the state event with a blank one). In the future, I would like consumers of policy lists to also be able to annotate policies with other policies[^reinforcement] [^agree-disagree], and this would be a way of ignoring/undoing a policy from a list that you are consuming but do not have write access to, which is safer than marking an entity as immune. 

[^reinforcement]: https://github.com/matrix-org/matrix-spec-proposals/pull/3849
[^agree-disagree]: https://github.com/the-draupnir-project/planning/issues/3 - MSC pending.

proposals/4150-m-allow-recommendation.md
Gnuxie: > Can the original PRs be linked

The original PR won't be useful, the work was done as part of the mjolnir for all in this PR https://github.com/matrix-org/mjolnir/pull/364. Mjolnir for all works by having a set of users who have access to a hosted Mjolnir bot, and this set of users is managed using a policy list with policies with the `org.matrix.mjolnir.allow` recommendation. The specific [commit](https://github.com/matrix-org/mjolnir/pull/364/commits/074a16bfaf3cd254070f32b73cf76188739e2fa9) is here but that won't be useful. Mjolnir for all didn't have the commands to manage this list but Draupnir does and the PR to do that is https://github.com/the-draupnir-project/Draupnir/pull/47 (which probably is reviewable), however this PR is old.

proposals/4150-m-allow-recommendation.md
clokep: This was my initial thought too! 👍 If the proposal morphs to change from that behavior a new prefix could be used.

(Also this gets you an implementation for free, I think.)

TheArcaneBrony: i think it might be worthwhile to also mention `org.matrix.mjolnir.allow` as unstable prefix already used in shadowspec realm?

Johennes: Good point. 👍

Have changed the vendor prefix to `org.matrix.mjolnir`.

proposals/4150-m-allow-recommendation.md
Johennes: I think I agree but I wonder if then we'd need to turn the spec's current _suggesting_ language on how to interpret policy lists into a prescription? It'd feel slightly off if we were to mandate how `m.ban` and `m.allow` interact when the concrete semantics of both are deliberately left as implementation details.

proposals/4150-m-allow-recommendation.md
turt2live: I haven't reviewed the implementations, sorry.

PR #4151, opened 2024-06-05
MSC4151: Reporting rooms (Client-Server API)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/report-room/proposals/4151-report-room.md)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4151#issuecomment-2150623218)

### DIFF
--- a/proposals/4151-report-room.md
+++ b/proposals/4151-report-room.md
@@ -53,8 +53,8 @@ Like `/report/:eventId`, handling of the report is left as a deliberate implemen
   spoiler tags on report reasons/content.
 
 * Clients should hide rooms the user reports by default to both discourage duplicate reports and to
-  remove the harmful content from the user's view. This may require filtering room directory responses
-  and room lists for the user.
+  remove the harmful content from the user's view, where possible. This may require filtering room
+  directory responses and room lists for the user, or an "ignore room API" like [MSC3840](https://github.com/matrix-org/matrix-doc/pull/3840).
 
   If the user is joined to a room, the client may wish to offer the user an option to leave the room.
 
--- a/proposals/4151-report-room.md
+++ b/proposals/4151-report-room.md
@@ -1,0 +1,104 @@
+# MSC4151: Reporting rooms (Client-Server API)
+
+The specification [already contains](https://spec.matrix.org/v1.10/client-server-api/#reporting-content)
+a module for being able to report events, though this functionality does not extend to rooms. Being
+able to report rooms is important for user safety: clients have room directories, invite lists,
+links to rooms, etc which all don't have an event ID to reference. If the client has *any* event ID
+for the room, it can use the existing 'report event' API to report the room instead. However, this
+only works if the user has visibility on the event ID being reported too.
+
+This proposal introduces an endpoint for reporting rooms, expanding the capabilities of the reporting
+module. The scope of this proposal is intentionally narrow to ensure quick traversal of the MSC process.
+Other, future, MSCs may further expand the suite of endpoints available to clients (like reporting
+users, media, etc).
+
+## Proposal
+
+Taking inspiration from [`POST /rooms/:roomId/report/:eventId`](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3roomsroomidreporteventid),
+a new endpoint is introduced:
+
+```
+POST /_matrix/client/v3/rooms/:roomId/report
+{
+  "reason": "<user-supplied, optional>"
+}
+```
+
+`reason` is a human-readable string describing the reason for the report. The string may be blank,
+but *must* be provided (to align with `/report/:eventId`).
+
+**Note**: `score` is not carried over from `/report/:eventId` because it has not proven useful. A
+future MSC may introduce it.
+
+There are no restictions on who can report a room: knowing the room ID is sufficient. This is to
+ensure that results from the room directory, invites, links, etc can all be reported. If the room
+does not exist on the server, the endpoint returns `404 M_NOT_FOUND`. Otherwise, `200` with `{}` as
+a response body.
+
+Like `/report/:eventId`, handling of the report is left as a deliberate implementation detail.
+
+## Safety considerations
+
+* Server admins may be exposed to harmful content through `reason`. This is an existing issue with
+  the reporting module, and difficult to fix. Applications which expose report reasons of any kind
+  are encouraged to place disclosures in the user experience path. For example, a page explaining
+  that the tool may contain harmful content before allowing the user temporary access, or the use of
+  spoiler tags on report reasons/content.
+
+* Clients should hide rooms the user reports by default to both discourage duplicate reports and to
+  remove the harmful content from the user's view. This may require filtering room directory responses
+  and room lists for the user.
+
+  If the user is joined to a room, the client may wish to offer the user an option to leave the room.
+
+* Users may report whole rooms instead of events in that room, particularly during a harmful content
+  spam wave. Administrators and safety teams should be cautious to avoid shutting down or banning
+  whole rooms, as the room may be legitimate otherwise. Automated decision making is not suggested
+  for a similar reason.
+
+* 'Report flooding' is more easily possible with this new endpoint, where many users report a room
+  with the hope of getting it shut down/banned. Mentioned a few times in this proposal, automated
+  decision making is not recommended for this endpoint to prevent consequences like this from
+  happening.
+
+## Potential issues
+
+* Within the Trust & Safety environment, it is well known that `reason` alone is insufficient for an
+  informed report. Self-triage categories and mandatory `reason` for some of those categories help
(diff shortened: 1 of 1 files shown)


proposals/4151-report-room.md
turt2live, line 1: Implementation requirements:

* [ ] Server supporting endpoint

turt2live: Typically it would be. Client implementation is meant to show two things:

1. Clients want the feature being described.
2. Clients can incorporate the feature into their project.

The first is satisfied outside of any client's control: legal requirements for some app stores and geographic regions require such an endpoint. This isn't really described in the MSC, but probably should be. Essentially, users *must* be given a way to report all kinds of content, including rooms. 

The second is not a perceived risk for this MSC, so is not a strong implementation requirement.

Gnuxie: There might not be a perceived risk that a client cannot implement the MSC, but there is a risk that implementation leads to the discovery of issues. This is kind of exacerbated for this MSC though, because receiving a report is not specified and that is UX that critically would need to be considered alongside this MSC (and already has been in part by the discussion on how to handle `reason`). I do not think this particular MSC should be held back though as it already is late, but something should be done to follow up.

turt2live: Yea, for clarity: everything surrounding handling reports, including informing the reporter of action taken and estimates for resolution, are being handled in a different MSC series. It's expected that those MSCs will be opened soon, though are still being drafted as we speak.

turt2live:
Changing hats from T&S to SCT, I see no reason to delay this MSC.

@mscbot fcp merge

proposals/4151-report-room.md
```diff
@@ -0,0 +1,111 @@
...
+
+`reason` is a human-readable string describing the reason for the report. The string may be blank,
+but *must* be provided (to align with `/report/:eventId`).
+
+**Note**: `score` is not carried over from `/report/:eventId` because it has not proven useful. A
+future MSC may introduce it.
+
+There are no restictions on who can report a room: knowing the room ID is sufficient. This is to
+ensure that results from the room directory, invites, links, etc can all be reported. If the room
+does not exist on the server, the endpoint returns `404 M_NOT_FOUND`. Otherwise, `200` with `{}` as
```
clokep, line 42: Is this a potential leak of whether a server is currently involved in a room or not?

proposals/4151-report-room.md
```diff
@@ -0,0 +1,111 @@
...
+Other, future, MSCs may further expand the suite of endpoints available to clients (like reporting
+users, media, etc).
+
+## Proposal
+
+Taking inspiration from [`POST /rooms/:roomId/report/:eventId`](https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3roomsroomidreporteventid),
+a new endpoint is introduced:
+
+```
+POST /_matrix/client/v3/rooms/:roomId/report
```
clokep, line 28: I don't want to bikeshed too much, but should this be v1? Could be a bit hard since it is a subpath of a previous path.

proposals/4151-report-room.md
turt2live: (the above is noted by https://github.com/matrix-org/matrix-spec-proposals/pull/4151/commits/7aafad3916f3fa1f7c49f89bf2891b701ae8bfb1 ftr)

proposals/4151-report-room.md
turt2live: I debated this while writing the MSC, and could be easily convinced either way. The only rationale I have for using v3 is that the endpoint it is modeled on is v3, so it'd indeed be confusing to have this be v1.

proposals/4151-report-room.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Safety considerations
+
+* Server admins may be exposed to harmful content through `reason`. This is an existing issue with
+  the reporting module, and difficult to fix. Applications which expose report reasons of any kind
+  are encouraged to place disclosures in the user experience path. For example, a page explaining
+  that the tool may contain harmful content before allowing the user temporary access, or the use of
+  spoiler tags on report reasons/content.
+
+* Clients should hide rooms the user reports by default to both discourage duplicate reports and to
```
ara4n, line 55: this is easier said than done, and wouldn’t be synchronised  across clients. perhaps it would be worth explicitly calling out that it requires an “ignore room” API that doesn’t currently exist?l (eg MSC3840)?

proposals/4151-report-room.md
turt2live: It would reveal the information, though I'm not sure the implications of that rise to the level of a leak. I believe the spec already allows clients to discover this information in other ways, but haven't tried to cite my sources on that :p

proposals/4151-report-room.md
turt2live: The media `/create` API is probably the closest example for a new endpoint within an existing module - it used v1 while the existing endpoints were v3. 

I'm fine either way - if folks +1 (or +0.1) this thread, I'll move it to v1.

proposals/4151-report-room.md
clokep: This is a local user only API so, yeah you likely can figure that info out already. (Which doesn't seem great, but no reason to block this MSC on it.)

proposals/4151-report-room.md
turt2live: addressed in b3fa180a850b47ac82867e9ffc6223fc974b5c91, I hope

proposals/4151-report-room.md
turt2live: I've mentioned it in dd85e178678fcf97bf42bfe7d40bd91d1cb96920

proposals/4151-report-room.md
```diff
@@ -0,0 +1,117 @@
...
+* The event report API could be relaxed to support an empty string for the event ID, though this feels
+  effectively like a new endpoint anyways. This MSC introduces such an endpoint.
+
+* The event report API's history visibility check could also be removed, though, as per
+  [MSC2249](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2249-report-require-joined.md),
+  this is intentional behaviour.
+
+## Security considerations
+
+* Rate limiting is strongly recommended for this new endpoint.
```
kegsay, line 105: I'd also suggest a size limit to the reason key, e.g 512KB.

turt2live: I'm not sure we need to do that in the spec. Clients should generally expect `413 Payload Too Large` errors when submitting things to an API.

kegsay: This adds needless variation and configuration, especially since when server authors write code they often just follow the spec as written, so are unlikely to include any size limits unless it is mentioned.

proposals/4151-report-room.md
clokep: I think I'd lean towards v3 as an extension of an existing endpoint. If others feel strongly, speak up!

proposals/4151-report-room.md
clokep: Putting limits on everything should really be the norm so we don't need to back into it afterward. I'm unsure I've seen that error code be considered for any endpoint before.

KitsuneRal: That looks more like a systematic issue across the spec, worth addressing also systematically in another MSC?

proposals/4151-report-room.md
dbkr: I agree the endpoint's similarity to event ID reporting means requiring a client impl is probably overkill here, but for future reference, I definitely think we need to call out explicitly where and why we're skipping it. We can't just arbitrarily decide to ignore our own rules.

proposals/4151-report-room.md
turt2live: I've moved this to a dedicated MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4154

proposals/4151-report-room.md
```diff
@@ -0,0 +1,117 @@
...
+
+* Authentication is required for this new endpoint.
+
+## Unstable prefix
+
+While this proposal is not considered stable, implementations should use `/_matrix/client/unstable/org.matrix.msc4151/report/:roomId`
```
sandhose, line 111: Shouldn't that be `/rooms/:roomId/report` at the end?

```suggestion
While this proposal is not considered stable, implementations should use `/_matrix/client/unstable/org.matrix.msc4151/rooms/:roomId/report`
```

proposals/4151-report-room.md
turt2live: For logging purposes:

* Client (app): https://github.com/element-hq/element-ios/pull/7805
* Client (sdk): https://github.com/matrix-org/matrix-ios-sdk/pull/1859
* Server: https://github.com/element-hq/synapse/pull/17270

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1938

PR #4153, opened 2024-06-10
MSCxxxx: Invisible Cryptography

uhoreg:


proposals/4153-invisible-crypto.md
turt2live, line 1: Implementation requirements:

* [x] Client embodying these values (preferably cross platform).

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,122 @@
+# MSC4153: Invisible Cryptography
```
deepbluev7, line 1: I don't see how that name has anything to do with the proposal?

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,122 @@
...
+cross-signing keys.
+
+Users should have Secret Storage with a default key that encrypts the private
+cross-signing keys and key backup key (if available)
+
+The spec currently does not give recommendations for what information is stored
+in Secret Storage, or even whether Secret Storage is available to users.  A
+user’s Secret Storage should contain the user’s cross-signing secret keys and
+the key backup decryption key (if the user is using key backup).  This ensures
+that users have a more consistent experience.
```
richvdh, line 35: don't these two paragraphs say the same thing? Or am I missing a subtlety?

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Messages from non-cross-signed devices should be untrusted
+
+Similarly, clients have no assurance that encrypted messages sent from
+non-cross-signed devices were sent by the user, rather than an
+impersonator. Therefore messages sent from non-cross-signed devices cannot be
+trusted and should be displayed differently to the user. For example, the
+message could be displayed with a warning, or may be hidden completely from the
+user. Again, clients may be allow the user to override this behaviour for
+specific devices for development or testing purposes.
```
dkasak, line 90: I think we are missing the mark here by not requiring the receiving client to ignore all encrypted packets sent by an unsigned device.

Since unsigned devices provide no authentication at all, they frankly make very little sense from a cryptographic perspective and should simply be considered as improperly initialised devices. Any kind of message sent from an improperly configured device should simply be ignored, or at the very least staged somewhere without processing it until the device becomes properly initialised.

Additionally, the paragraph talking about room keys should at the very least also include secrets.

proposals/4153-invisible-crypto.md
uhoreg: The first is referring to this paragraph in https://spec.matrix.org/unstable/client-server-api/#key-storage:

> A key can be marked as the “default” key by setting the user’s account data with event type `m.secret_storage.default_key` to an object that has the ID of the key as its key property. The default key will be used to encrypt all secrets that the user would expect to be available on all their clients. Unless the user specifies otherwise, clients will try to use the default key to decrypt secrets.

and is saying that the user's Secret Storage should have such a key.

Whereas the second paragraph is describing what secrets should be stored.

Perhaps it would make more sense if I switched the order of the paragraphs, since the first paragraph also touches on what's stored.

proposals/4153-invisible-crypto.md
uhoreg: > I think we are missing the mark here by not requiring the receiving client to ignore all encrypted packets sent by an unsigned device.

I was mixing some things up in my head when I wrote that bit (confusing unsigned devices with changing identities).  I've fixed it to say that such messages should be ignored.



> Additionally, the paragraph talking about room keys should at the very least also include secrets.

Added

proposals/4153-invisible-crypto.md
uhoreg: After discussion with the rest of the Element crypto team, we came to the consensus that "Invisible crypto" referred more to specific UI changes that this MSC would enable (such as getting rid of some shields), rather than the recommendations in this MSC.  So I've given this MSC a more boring name.  The filename will remain the same, so that we don't lose comments.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+### Bots and application services
+
+This is a special case to the issue above, but seems to be a large enough class
+that it deserves its own mention: support for cross-signing in bots and
+application services may be less common than in interactive clients.  When a
+client fully implements this proposal, users will be unable to interact with
+bots and application services in encrypted rooms if they do not support
+cross-signing.
+
+TODO: status of cross-signing in bots/application services
```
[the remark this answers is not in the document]
uhoreg: When you say "most encryption-capable bridges (i.e. mautrix-*) only generate keys ...", are you talking about device keys? Can you explain how megolm session distribution works with mautrix bridges? Is it something like: bridge generates megolm session, bridge bot sends megolm session to recipient, ghost sends encrypted message to room?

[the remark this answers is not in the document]
tulir: Yes, exactly

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Devices should be cross-signed
+
+Clients should encourage users to cross-sign their devices.  This includes both
+when logging in a new device, and for existing devices.  Clients may even go so
+far as to require cross-signing of devices by preventing the user from using
+the client until the device is cross-signed.  If the user cannot cross-sign
+their device (for example, if they have forgotten their Secret Storage key),
+the client can allow users to reset their Secret Storage, cross-signing, and
+key backup.
```
TheArcaneBrony, line 58: If a client enforces cross signing, how would cross signing work without access to other devices or recovery key?

Real scenario that happens to me from time to time:
Not all my accounts have the same powerlevel in rooms, sometimes I need to switch accounts on the go to carry out a moderation action. Worse, most of the moderation bot rooms I interact with on the regular use End-to-End Encryption (usually via Pantalaimon). Additionally, I do *not* feel comfortable keeping a copy of my Recovery Keys on my phone, or accessible to my phone, nor want to rely on physical devices (eg. security keys) when eg. going to a party, hanging out with friends at an entertainment park, etc.

Element X Android provides a good example here: while I enjoy using sliding sync, using Element X left me unable to carry out moderation actions due to being unable to switch accounts without cross-signing.
I'm currently using Schildichat Next to bypass cross signing requirements, but having multiple copies installed would exacerbate my out-of-memory issues on my phone, which are *already* preventing me from using it effectively at all.

proposals/4153-invisible-crypto.md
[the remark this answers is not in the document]
tulir: There was some discussion about this in [#element-x-ios:matrix.org](https://matrix.to/#/!amwlFMejXQcoCxNYHH:matrix.org/$U0s2WVXqZqW6fPwEseah7g1KqfkmY03BHn5sbh6gqJc?via=matrix.org&via=element.io&via=beeper.com)

For clarity: the wasteful part is not the device key generation itself, but rather the resulting large number of olm sessions when clients have to encrypt megolm sessions for each ghost individually.

Matthew suggested that the ghost could create a device with no identity key (to prevent olm sessions being created) and use it to sign the bot's device.

I'm thinking ghosts will probably need a cross-signing key (because that's the real root of trust). Maybe ghosts could just upload the bot's cross-signing key? They'd have the wrong `user_id`, which clients could detect to realize it's a ghost user. The key could also have a special `usage` field or something else as an extra ghost identification. Or maybe the `user_id` field should be the ghost, but there could be a new `bridge_bot_id` field which points at the bridge bot.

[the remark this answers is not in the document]
tulir: Oh, there's also another related annoying part: double puppeting, where bridge logs in with your account to bridge messages you send from the remote network using your real account instead of a ghost. Currently it works the same way as ghosts in terms of encryption: create megolm keys with bridge bot, send with double puppet.

The number of double puppets is generally much lower than the number of ghosts, so technically it would be fine to create a full e2ee session (which the user can then cross-sign normally). However, that would mean the bridge receives decryption keys for messages in all your rooms instead of only its own rooms, which is not great.

I don't think it's worth creating a separate mechanism for trusting double puppets, but at least the "upload bot's cross-signing keys to ghost" mechanism wouldn't work for double puppets. Creating a device without identity keys could work for both though

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+In addition, users should not send secrets (via Secret Sharing) to their own
+devices that are not cross-signed.
+
+### Messages from non-cross-signed devices should be ignored by default
+
+Similarly, clients have no assurance that encrypted messages sent from
+non-cross-signed devices were sent by the user, rather than an
+impersonator. Therefore messages sent from non-cross-signed devices cannot be
+trusted and should not be displayed differently to the user. Again, clients
+may be allow the user to override this behaviour for specific devices for
```
bnjbvr, line 97: typo:

```suggestion
may allow the user to override this behaviour for specific devices for
```

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+application services may be less common than in interactive clients.  When a
+client fully implements this proposal, users will be unable to interact with
+bots and application services in encrypted rooms if they do not support
+cross-signing.
+
+TODO: status of cross-signing in bots/application services
+
+## Alternatives
+
+## Security considerations
```
FSG-Cat, line 126: This MSC has some severe T&S related security problems by effectively acting as a availability break. Now once the bot ecosystem has advanced sufficiently in its E2EE implementations this problem can be resolved. Recovery Keys will allow easy self restoration of cross signed status just like we have access to today in normal clients like Element Web.

Thanks to Tulir in the Spec  room for mentioning Recovery Key based workflows as a realistic way to help resolve this conflict.

MTRNord: It also would be nice to point the potential tripwires for headless clients in the spec text later on as an info box. Since this is a thing easily overlooked imho if you aren't proficient in e2ee.

proposals/4153-invisible-crypto.md
dkasak: The paragraph says "may even", so enforcing cross-signing when logging in is not a requirement. Moreover, moderation actions are usually state events, which are not E2EE authenticated at the moment, so the situation there should remain unchanged.

Your unsigned device *should* get ignored for everything that *is* E2EE authenticated though, and for a good reason—the very concept of unsigned devices is unsound in an E2EE setting, because such devices can be freely added by the server, so in what sense is it E2EE at all? I don't think your scenario has much bearing on that fact.

proposals/4153-invisible-crypto.md
FSG-Cat: I personally have no stance on if it needs to be explictly called out. But other than that you have described exactly the solution to the problem as far as i understand said solution to be and what the problem is.

The explicit callout is more of a question for @MTRNord

proposals/4153-invisible-crypto.md
FSG-Cat: The reason why moderation bots is a concern is because they are controlled via regular messages in dedicated rooms. Said room if your using mjolnir is recomended to be Encrypted.

proposals/4153-invisible-crypto.md
MTRNord: > The concern is that this MSC breaks E2EE for bots because they most often don't sign their own device(s). Moreover bots are often headless and often have only a single instance, so there is nothing to perform verification against.
> 
> However, a workable way forward for bots is to give them the Recovery Key in order to let them access Secret Storage (SSSS), from which they will retrieve their cryptographic identity (cross-signing keys) and sign their device. You want this technique to be mentioned explicitly (for example in an info box) because it's not always obvious to implementers.
> 
> Is that an accurate rephrasing?

Yes thats my POV on it. It took some asking to get to this solution. After you see it of course it's fairly obvious, but it would imho make a good info box on the spec to make this more obvious to both sdk implementors that need to support it and bot developers that need to use it then. So I think this is somewhat a request for someone later writing the spec text maybe even.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+a more aggressive warning.
+
+Note that this MSC does not propose a mechanism for remembering previous
+cross-signing keys between devices. In other words if Alice changes her
+cross-signing keys and then Bob logs in a new device, Bob’s new device will not
+know that Alice’s cross-signing keys had changed, even if Bob has other devices
+that were previously logged in. Such a mechanism could be proposed by another
+MSC.
+
+### Room keys and secrets should by default not be sent to non-cross-signed devices
```
morguldir, line 80: I think it would make sense to use a m.room_key.withheld code here, tulir proposed a new code `m.not_cross_signed`

Adding this would make it easier to separate unverified users and insecure devices, like mentioned in https://github.com/element-hq/element-meta/issues/2621

proposals/4153-invisible-crypto.md
richvdh: I believe that EW and EX both implement this, via their respective labs flags.

EW's setting looks like this:

image

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Messages from non-cross-signed devices should be ignored by default
+
+Similarly, clients have no assurance that encrypted messages sent from
+non-cross-signed devices were sent by the user, rather than an
+impersonator. Therefore messages sent from non-cross-signed devices cannot be
+trusted and should not be displayed differently to the user. Again, clients
+may be allow the user to override this behaviour for specific devices for
+development or testing purposes.
+
```
ara4n, line 99:
```suggestion

### Unencrypted devices should not impact E2EE behaviour.

For the sake of clarity: clients without E2EE (e.g. apps like [WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/) which choose to set up E2EE only on demand) should not count as unverified devices.

Therefore, while MSC4153 is rolling out, clients should not warn or refuse to send if unencrypted devices are present in an encrypted room.  Once MSC4153 is fully rolled out, these devices would be silently excluded anyway (especially as there is no way to send to them!)
```

dkasak: "unencrypted devices" sounds unclear to me. I think you mean "Matrix devices which do not have identity keys"?

I think in general we may be missing terminology to differentiate between "non-cryptographic devices" (Matrix devices without identity keys) and "cryptographic devices" (Matrix devices that do have identity keys), to easily differentiate between them. They are completely different beasts.

What do you think about this?

```suggest
### Non-cryptographic devices should not impact E2EE behaviour

For the sake of clarity: non-cryptographic devices (devices which do not have
identity keys uploaded) should not have any impact on a client's E2EE
behaviour. For all intents and purposes, non-cryptographic devices are a
completely separate concept and do not exist from the perspective of the
cryptography layer since they do not have identity keys, so it is impossible to
send them encrypted messages.

In particular, Matrix clients MUST NOT consider non-cryptographic devices to be
equivalent to non-cross-signed cryptographic devices for purposes of enforcing
E2EE policy. For example, clients SHOULD NOT warn nor refuse to send messages
due to the presence of non-cryptographic devices.

The intent of this is to smoothly support and minimise interference from
applications which choose to set up E2EE only on demand (e.g.
[WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/).
Such clients should initially create a non-cryptographic device until they are
ready to set up E2EE. Only when they are ready, they will upload identity keys
to the device, converting it into a cryptographic device and making it subject
to MSC4153 rules.

Comment (richvdh) [reply proposals/4153-invisible-crypto.md:99]: > For example, clients SHOULD NOT warn nor refuse to send messages
> due to the presence of non-cryptographic devices.

non-cryptographic devices aren't even returned by `/keys/query` (and there is no other way to find out about another user's devices), so clients *can't* warn or refuse to send messages due to the presence of non-cryptographic devices.

We could do with finding a phrasing that makes it clear that non-cryptographic devices don't cause a problem here, without implying that they have to be special-cased.

Comment (dkasak) [reply proposals/4153-invisible-crypto.md:99]: You can still find out about non-cryptographic device of the *current* user using [GET /_matrix/client/v3/devices](https://spec.matrix.org/v1.13/client-server-api/#get_matrixclientv3devices). So perhaps WorkAdventure was worried of their non-cryptographic device interfering with other devices on the same user's account?

Otherwise I'm not even sure what we're trying to achieve with this addition.

Comment (moufmouf) [reply proposals/4153-invisible-crypto.md:99]: Hey!

David from WorkAdventure here.

Our current issue is that when we connect to Element X after connecting in WorkAdventure (but without configuring the cross-signing), Element X displays this message:

image

> Such clients should initially create a non-cryptographic device until they are
ready to set up E2EE. Only when they are ready, they will upload identity keys
to the device, converting it into a cryptographic device and making it subject
to MSC4153 rules.

If the message is displayed in Element X, I guess WorkAdventure is treated as a cryptographic device that is not properly cross-signed.

The thing is: I don't understand when WorkAdventure is uploading the identity keys to the device.
We are using the Matrix-JS-SDK. As far as I understand, the upload of identity keys is done in crypto.bootstrapCrossSigning => https://github.com/workadventure/workadventure/blob/cbd3607559b5e15c335b449c376c800f261e6f13/play/src/front/Chat/Connection/Matrix/MatrixSecurity.ts#L66-L97
But even if we don't call this method, WorkAdventure is detected as an unsigned device by Element X. The issue probably boils done to us understanding when the upload of identity keys happens in the Matrix-JS-SDK.

Comment (richvdh) [reply proposals/4153-invisible-crypto.md:99]: > Only when they are ready, they will upload identity keys
> to the device,

I think this is confusingly worded. No keys are "uploaded to the device". Rather: a pair of keys are created; these are uploaded *to the server*. The keys in question are, technically speaking, the device's identity keys, but normally when we refer to "identity keys" we mean the *user's* identity keys. So normally we just refer to the "device keys" here.

These device keys are created, and uploaded to the server (thus making the device a cryptographic device) when you call `MatrixClient.initRustCrypto`.

So: if you don't want your device to appear as a cryptographic device (and hence be subject to the warnings in other cryptographic devices), don't call `initRustCrypto` (or any of the other stuff like `bootstrapCrossSigning`)

Comment (dkasak) [reply proposals/4153-invisible-crypto.md:99]: Here's a tweaked version to try to remove the ambiguity discussed above:

```suggestion
### Non-cryptographic devices should not impact E2EE behaviour

For the sake of clarity: non-cryptographic devices (devices which do not have
device identity keys uploaded to the homeserver) should not have any impact on
a client's E2EE behaviour. For all intents and purposes, non-cryptographic
devices are a completely separate concept and do not exist from the perspective
of the cryptography layer since they do not have identity keys, so it is
impossible to send them encrypted messages.

In particular, Matrix clients MUST NOT consider non-cryptographic devices to be
equivalent to non-cross-signed cryptographic devices for purposes of enforcing
E2EE policy. For example, clients SHOULD NOT warn nor refuse to send messages
due to the presence of non-cryptographic devices.

The intent of this is to smoothly support and minimise interference from
applications which choose to set up E2EE only on demand (e.g.
[WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/).
Such clients should initially create a non-cryptographic device until they are
ready to set up E2EE. Only when they are ready will they create the device
identity keys for the device and upload them to the homeserver, converting the
device into a cryptographic device and making it subject to MSC4153 rules.

Comment (moufmouf) [reply proposals/4153-invisible-crypto.md:99]: > These device keys are created, and uploaded to the server (thus making the device a cryptographic device) when you call `MatrixClient.initRustCrypto`.

Interesting! We were not aware of that. Ping @ydiarra

Do you know if we can call `MatrixClient.initRustCrypto` after the client is started (after we call MatrixClient.startClient)? That would be awesome because it would allow us to "upgrade" to cryptographic device if needed (even better if we find a way to put the crypto stuff in a different Vite chunk to avoid loading it if not required). I it ain't possible, we would have to start the client, detect some cryptography is needed, stop the client, and start it again with Rust crypto enabled.

File: proposals/4153-invisible-crypto.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (uhoreg) [reply proposals/4153-invisible-crypto.md:122]: I think that it's a bug that we're accepting messages encrypted using a key that we received from a different user, because that seems bad from a cryptographic point of view.  So I'm not too keen to try to work around it.  I also don't think that we should have bots upload cross-signing keys of different users.

If the big issue is in receiving encrypted messages, maybe the easier solution would be to add a flag in the device keys object to indicate that the device doesn't want to receive any keys?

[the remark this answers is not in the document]
Comment (tulir) [reply proposals/4153-invisible-crypto.md:122]: The flag to not receive encrypted messages sounds good, that'd at least solve the double puppeting issue of receiving unwanted keys. Sending messages would still require creating olm sessions from every ghost that sends messages though

Technically there's also the option of deciding that appservices were a mistake and switching to MSC4144, but having to build duplicates of a bunch of different mechanisms like member lists and membership actions would be unfortunate

File: proposals/4153-invisible-crypto.md (hunk 9 above)
Comment (richvdh) [reply proposals/4153-invisible-crypto.md:99]: > Right, this is jargon on my part but it's symptomatic of us interchangeably referring to several different concepts as "the device"

Sorry yes, I didn't mean any criticism. A lot of the confusion arises because of inconsistency in the spec, or because we use different words in our products to the spec. MSC4161 is trying to improve the landscape here, but it's slow going.

"Identity key", it turns out, is very much a case where our usage is inconsistent. I think we should update the spec to emphasise "**device** identity key" here. 

Your updated wording is much clearer. @uhoreg: can we get @dkasak's suggestion absorbed into the MSC?

Comment (richvdh) [reply proposals/4153-invisible-crypto.md:99]: @moufmouf:

> Do you know if we can call `MatrixClient.initRustCrypto` after the client is started (after we call MatrixClient.startClient)?

Heh; great question. Certainly you *should* be able to do so: this is exactly the correct model for your usage. As to whether it works in practice: I don't know; I don't think I have tested it. There may well be a couple of edge cases we need to fix in matrix-js-sdk to make it work. If you try it, and have any problems, would you mind opening an issue at https://github.com/matrix-org/matrix-js-sdk/issues?

Comment (moufmouf) [reply proposals/4153-invisible-crypto.md:99]: @ydiarra is going to give this a shot. We'll open an issue if we find a problem!

Comment (uhoreg) [reply proposals/4153-invisible-crypto.md:143]: I've applied @dkasak 's suggestion in https://github.com/matrix-org/matrix-spec-proposals/pull/4153/files#r1936064557 (with a minor wording tweak), so I'm going to resolve this thread.

File: proposals/4153-invisible-crypto.md (hunk 7 above)
Comment (uhoreg) [reply proposals/4153-invisible-crypto.md:184]: I added some words in the "Bots and application services" section under "Potential issues", about how to verify bots.

File: proposals/4153-invisible-crypto.md [hunk 10]
```diff
@@ -0,0 +1,208 @@
...
+### Encrypted to-device messages MUST NOT be sent to non-cross-signed devices
+
+Since non-cross-signed devices don’t provide any assurance that the device
+belongs to the user, and server admins can trivially create new devices for
+users, clients MUST not send encrypted to-device messages, such as room keys or
+secrets (via Secret Sharing), to non-cross-signed devices by default.  When
+sending room keys, clients can use a [`m.room_key.withheld`
+message](https://spec.matrix.org/unstable/client-server-api/#reporting-that-decryption-keys-are-withheld)
+with a code of `m.unverified` to indicate to the non-cross-signed device why it
```
uhoreg, line 100: Moving morguldir's [comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4153#discussion_r1898676103) here, to discuss whether we should use `m.unverified` for this, or use a new code:

> I think it would make sense to use a m.room_key.withheld code here, tulir proposed a new code m.not_cross_signed
>
> Adding this would make it easier to separate unverified users and insecure devices, like mentioned in https://github.com/element-hq/element-meta/issues/2621

proposals/4153-invisible-crypto.md
uhoreg: I added text to suggest using an `m.room_key.withheld` message, and I copied this comment there to discuss whether we should use an existing code, or use a new code.  https://github.com/matrix-org/matrix-spec-proposals/pull/4153/files#r1937871448

escix:
I am a non-technical person, so apologies if I am not making sense in my comments.

Currently, EX uses three identity confirmation methods:

- Use another device= This is the easiest one as all cryptography is already done in another device.
- Enter a recovery key= This, I believe, is the Security key to obtain the encryption key from the server.
- Can't confirm= This is to reset my message history and setup a new Security key.

If I understand properly, this MSC is to make login, identity verification and backup on the first login simple and efficient.

Login: dealt with username and password

Identity verification: Unless the persons are next to each other this will be an out-of-band confirmation. If the persons are next to each other this can be verified by scanning a QR code or something. On an out of bound scenario, this can be a phone call and confirmation of emojis or random numbers. 
However, in the current proposal, in the first use it is proposed to accept the identity by default. This will improve the user engagement, but can an advance section present so the above methods can be used by advanced users?

Backup Security keys: I think, by default, the encryption keys must be saved in the server and also in the device (downloaded as a file). Both can be protected with the same Security key.
 
So on the first login, use the username and password to login and presented with the option for identity verification with a skip button for non-technical users. Then proceed to 'backup encryption keys' with a Security Key.

proposals/4153-invisible-crypto.md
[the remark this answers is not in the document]
uhoreg: I've written [MSC4261](https://github.com/matrix-org/matrix-spec-proposals/pull/4261) to propose adding a flag to mark devices that don't want to receive room keys.

proposals/4153-invisible-crypto.md
richvdh, line 1: Moving @escix 's [question](https://github.com/matrix-org/matrix-spec-proposals/pull/4153#issuecomment-2628406154) to a thread:

> I am a non-technical person, so apologies if I am not making sense in my comments.
> 
> Currently, EX uses three identity confirmation methods:
> 
>   * Use another device= This is the easiest one as all cryptography is already done in another device.
> 
>   * Enter a recovery key= This, I believe, is the Security key to obtain the encryption key from the server.
> 
>   * Can't confirm= This is to reset my message history and setup a new Security key.
> 
> 
> If I understand properly, this MSC is to make login, identity verification and backup on the first login simple and efficient.
> 
> Login: dealt with username and password
> 
> Identity verification: Unless the persons are next to each other this will be an out-of-band confirmation. If the persons are next to each other this can be verified by scanning a QR code or something. On an out of bound scenario, this can be a phone call and confirmation of emojis or random numbers. However, in the current proposal, in the first use it is proposed to accept the identity by default. This will improve the user engagement, but can an advance section present so the above methods can be used by advanced users?
> 
> Backup Security keys: I think, by default, the encryption keys must be saved in the server and also in the device (downloaded as a file). Both can be protected with the same Security key.
> 
> So on the first login, use the username and password to login and presented with the option for identity verification with a skip button for non-technical users. Then proceed to 'backup encryption keys' with a Security Key.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,231 @@
...
+
+The intent of this is to smoothly support and minimise interference from
+applications which choose to set up E2EE only on demand (e.g.
+[WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/).
+Such clients should initially create a non-cryptographic device until they are
+ready to set up E2EE. Only when they are ready will they create the device
+identity keys for the device and upload them to the homeserver, converting the
+device into a cryptographic device and making it subject to the rules given in
+this MSC.
+
```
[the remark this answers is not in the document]
andybalaam: (As per https://github.com/matrix-org/matrix-spec-proposals/pull/4161#discussion_r1980443919 I think it makes sense to leave this material out of MSC4161 and add it when this MSC is merged.)

proposals/4153-invisible-crypto.md
uhoreg: The implementation in Element uses `m.unverified`, and it seems to be working OK, so we may just stick with that for now.  If we find that we want to use a separate code after this MSC has already been accepted, we can add it in a separate MSC.

proposals/4153-invisible-crypto.md
uhoreg: While I'm sympathetic to your usecase, I think that we still need the security properties that this MSC gives.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,231 @@
...
+The following bots support cross-signing:
+
+- [meowlnir](https://github.com/maunium/meowlnir)
+- [Arnie](https://gitlab.com/andybalaam/arnie)
+
+The following bot SDKs support, or plan to support, cross-signing such that any
+bots written using them will support cross-signing:
+
+- [maubot](https://github.com/maubot/maubot) (planned)
+- [mautrix-go](https://github.com/mautrix/go) (planned support for Application Services)
```
HarHarLinks, line 201: Thanks for adding this section, it is what I am most concerned about with the series of invisible crypto MSCs.

It's difficult to do an objective survey of the ecosystem in this regard. I think it would however be prudent to also list here bots/appservices/SDKs assumed to be popular by some measure (github stars? project commit activity? activity in project's rooms?) that do support e2ee but NOT cross-signing, not only those maintained by SCT members.

I do however agree that the mau* suite of projects is very wide spread and one important implementation blocking the ecosystem. Further can be found through the Foundation website https://matrix.org/ecosystem/ and the commonly used Matrix "distro" https://github.com/spantaleev/matrix-docker-ansible-deploy?tab=readme-ov-file#bridges

proposals/4153-invisible-crypto.md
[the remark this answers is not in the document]
richvdh: I've opened https://github.com/element-hq/element-meta/issues/2950 to keep track of this problem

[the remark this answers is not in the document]
tulir: Wrote an MSC to define impersonator devices, which handles both ghost users and double puppets: https://github.com/matrix-org/matrix-spec-proposals/pull/4350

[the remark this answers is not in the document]
uhoreg: https://github.com/matrix-org/matrix-spec-proposals/pull/4153/commits/aeceee38cc02c3c56eff0840755d3be929922e2f adds some more words to this MSC about bridging

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+
+The following bots support cross-signing:
+
+- [meowlnir](https://github.com/maunium/meowlnir)
+- [Arnie](https://gitlab.com/andybalaam/arnie)
+
+The following bot SDKs support, or plan to support, cross-signing such that any
+bots written using them will support cross-signing:
+
+- [maubot](https://github.com/maubot/maubot) (planned)
```
tulir:
```suggestion
- [maubot](https://github.com/maubot/maubot)

The following bot SDKs support, or plan to support, cross-signing such that any
bots written using them will support cross-signing:

```

turt2live:
The author of this MSC has [proposed](https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$DD5rJe6v3VRCbj_E9AMu_ZSj-23Mu3z0GLs6rxYYRqI?via=matrix.org&via=element.io&via=envs.net) that this MSC is ready for FCP.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
+# MSC4153: Exclude non-cross-signed devices
+
+End-to-end encryption was first introduced to Matrix in 2016. Over the years,
```
turt2live: editorial, optional: a link to when this happened (if there is such a thing) would be good.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+In general, this MSC proposes to standardize on using cross-signing as a basis
+for trusting devices.  While a user may be unable to verify every other user
+that they communicate with, or may be unaware of the need to verify other
+users, cross-signing gives some measure of protection and so should be used
+where possible.  One of the goals of this MSC is to reduce the number of
+warnings that users will encounter by taking advantage of cross-signing.
+
+## Proposal
+
+Note: The changes below only apply to clients that support encryption.
```
turt2live: editorial, optional: link to spec/module for added clarity

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Users SHOULD have cross-signing keys
+
+Clients SHOULD create new cross-signing keys for users who do not yet have
+cross-signing keys.
```
turt2live: How should clients go about doing this? If a user has 2+ clients and they both try to create cross-signing keys, surely they'd overwrite each other? 

A link to the spec plus some guidance on what it means to "create new keys" would be helpful.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Users SHOULD have Secret Storage
+
+The spec currently does not give recommendations for what information is stored
+in Secret Storage, or even whether Secret Storage is available to users.  Secret
```
turt2live: editorial: please link to the spec

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+The user's Secret Storage SHOULD have a default key (a key referred to by
+`m.secret_storage.default_key`) that encrypts the private cross-signing keys and
+key backup key (if available).
+
+### Verifying individual devices of other users is deprecated
+
+When one user verifies a different user, the verification SHOULD verify the
+users’ cross-signing keys.  Any flow between different users that does not
+verify the users' cross-signing keys (it verifies only the device keys) is
+deprecated.  Verifying a user’s own device keys is still supported.
```
turt2live, line 60: editorial: please link to the spec so we can see what exactly is being deprecated

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+another MSC.
+
+### Encrypted to-device messages MUST NOT be sent to non-cross-signed devices
+
+Since non-cross-signed devices don’t provide any assurance that the device
+belongs to the user, and server admins can trivially create new devices for
+users, clients MUST not send encrypted to-device messages, such as room keys or
+secrets (via Secret Sharing), to non-cross-signed devices by default.  When
+sending room keys, clients can use a [`m.room_key.withheld`
+message](https://spec.matrix.org/unstable/client-server-api/#reporting-that-decryption-keys-are-withheld)
```
turt2live:
```suggestion
message](https://spec.matrix.org/v1.15/client-server-api/#reporting-that-decryption-keys-are-withheld)
```

unstable spec links can and do change, which will break the historical record.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+this MSC.
+
+### Clients MAY make provisions for encrypted bridges
+
+Some bridges are structured in a way such that only one user controlled by the
+bridge (often called the bridge bot) participates in encryption, and encrypted
+messages from other bridge users are encrypted by the bridge bot.  Thus
+encrypted messages sent by one user could be encrypted by a Megolm session sent
+by a different user.  Clients MAY accept such messages, provided the bridge
+bot's device is cross-signed. However, the client MUST annotate the message with
```
turt2live:
```suggestion
by a different user.  Clients MAY accept such messages, provided the session
creator's device is cross-signed. However, the client MUST annotate the message with
```

to clarify which of the two entities is the "bridge bot" here.

turt2live (approved):
overall lgtm - my only FCP-blocking concern is a description of how clients are supposed to create cross-signing keys.

proposals/4153-invisible-crypto.md
turt2live: I think this is largely resolved, but please feel free to open a new thread on the updated diff as needed.

turt2live:
@mscbot fcp merge

turt2live:
@mscbot concern Guidance to clients on how to create cross-signing keys (links to spec, conflict resolution, etc)

proposals/4153-invisible-crypto.md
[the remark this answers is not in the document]
uhoreg: I don't think anything needs to be done here, since it looks like this MSC will be accepted before 4161, but leaving this open for documentation purposes.

proposals/4153-invisible-crypto.md
dbkr: IMO this already would have needed to be specified before, I don't think it's the job of this MSC to start speccing how to create keys or how to do so atomically between clients.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Users SHOULD have Secret Storage
+
+The spec currently does not give recommendations for what information is stored
+in Secret Storage, or even whether Secret Storage is available to users.  Secret
+Storage allows users to keep secrets on the server so that they are accessible
+when the user logs in to a new device and does not have an existing device that
+can share the secrets with the new device.  Therefore users SHOULD have Secret
+storage set up.
+
```
anoadragon453: Slight clarification:

```suggestion
when the user logs in to a new device. This is necessary to retrieve secrets when
the user does not have an existing device that can then share the secrets with the
new device. Therefore users SHOULD have Secret storage set up to avoid needing
to reset their cryptographic identity in this case.

```

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+The user’s Secret Storage SHOULD contain the user’s cross-signing secret keys
+and the key backup decryption key (if the user is using key backup).  This
+ensures that users use cross-signing and key backup on new devices.
+
+The user's Secret Storage SHOULD have a default key (a key referred to by
+`m.secret_storage.default_key`) that encrypts the private cross-signing keys and
+key backup key (if available).
```
anoadragon453, line 53: As a client, I may think that I need to encrypt the user's private cross-signing keys and key backup key with the default key, before storing everything in Secret Storage. Though there's no point to storing an encrypted blob if I store the decryption key alongside it.

Should the private cross-signing keys and key backup key only be encrypted when stored locally?

anoadragon453 (approved):
Minor clarifications, otherwise lgtm! :lock:

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Devices SHOULD be cross-signed
+
+Clients SHOULD encourage users to cross-sign their devices.  This includes both
+when logging in a new device, and for existing devices.  Clients may even go so
+far as to require cross-signing of devices by preventing the user from using
+the client until the device is cross-signed.  If the user cannot cross-sign
+their device (for example, if they have forgotten their Secret Storage key),
+the client can allow users to reset their Secret Storage, cross-signing, and
+key backup.
```
TheArcaneBrony, line 80: > Clients may even go so far as to require cross-signing of devices by preventing the user from using the client until the device is cross-signed.

I feel like this is too strict - at least for the common clients. ie. Element X does this, and is an absolute deal-breaker for me when I have to switch accounts on the go. I don't keep a copy of my recovery keys on my mobile devices for reasons I'd hope would be fairly self-explanatory, and when I need it most, I generally *don't* have another device available to verify with for hours at a time.

So far I've been getting around this by either using debug builds of EX, or using ie. Neochat Android or the "legacy" Element apps.

As someone who relies on having lots of redundancy, this is an absolutely jarring change.

I didn't want to spam threads, so I'll note this also causes issues with these items:
- `Encrypted messages from non-cross-signed devices SHOULD be ignored`
- `Encrypted to-device messages MUST NOT be sent to non-cross-signed devices`

TheArcaneBrony (changes requested):
Some concerns re: users who rely on the current behavior, even if unintentionally.

proposals/4153-invisible-crypto.md
tulir: That specific sentence doesn't carry any sort of requirement or even recommendation, it's more of an observation: clients can choose to stop the user from using a client in a state where it will not work correctly in encrypted rooms. Clients are still totally free to not stop the user if the developers don't want it.

Also, even if that sentence were removed, clients could still do it, unless you're suggesting that the spec should explicitly forbid it. I don't think the spec should forbid clients from preventing the user doing dumb things.

TheArcaneBrony: I'm not saying that clients can't do that, just that it's a massive problem to ie. not be able to log into another homeserver to write bans when you're away from home and having technical difficulties. I'd be fine not having encryption at all in that case, though it's annoying to not be able to coordinate in rooms that happen to be encrypted.

proposals/4153-invisible-crypto.md
turt2live: I likely agree, but without a link to that pre-existing guidance it's difficult to accept that.

proposals/4153-invisible-crypto.md
KitsuneRal: If you cannot unlock your secret storage how do you assure others that it's actually you and not an impersonator coordinating in encrypted rooms?
I agree that a failure mode with encryption entirely disabled makes some sense. Actually, libQuotient allowed it in a previous version, and we've had that mode in Quaternion (you could login with E2EE entirely disabled). That was mainly a precaution while E2EE was unstable in the library but it might make sense bringing back with this MSC. With that said, I don't see much value in using E2EE without access to cross-signing keys.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+- gomuks
+- NeoChat
+- Nheko
+- pantalaimon
+- Tammy
+- Trixnity Messenger
+
+The following encryption-capable clients do not support cross-signing:
+
+- kazv
```
KitsuneRal, line 208:
```suggestion
- kazv
- Quaternion (versions prior to 0.98)
```

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+impossible to send them encrypted messages.
+
+In particular, Matrix clients MUST NOT consider non-cryptographic devices to be
+equivalent to non-cross-signed cryptographic devices for purposes of enforcing
+E2EE policy. For example, clients SHOULD NOT warn nor refuse to send messages
+due to the presence of non-cryptographic devices.
+
+The intent of this is to smoothly support and minimise interference from
+applications which choose to set up E2EE only on demand (e.g.
+[WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/).
```
clokep:
```suggestion
[WorkAdventure](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/)).
```

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+- kazv
+
+### Bots and application services
+
+This is a special case to the issue above, but seems to be a large enough class
+that it deserves its own mention: support for cross-signing in bots and
+application services may be less common than in interactive clients.  When a
+client fully implements this proposal, users will be unable to interact with
+bots and application services in encrypted rooms if they do not support
+cross-signing.
```
bnjbvr, line 218: Is there some timeline specifying some duration for application services/bot authors to implement support for this? It seems that, if this is implemented and enabled by default too soon, this would prevent use of popular e2be bridges like `mautrix-signal`, etc.

I see the note that support is planned for application services in `mautrix-go` which I _think_ is the basis of `mautrix-signal`, but this would be a big disruption if this MSC were enabled *before* the application services had an opportunity to update.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,249 @@
...
+purposes.
+
+### Non-cryptographic devices SHOULD NOT impact E2EE behaviour
+
+For the sake of clarity: non-cryptographic devices (devices which do not have
+device identity keys uploaded to the homeserver) should not have any impact on
+a client's E2EE behaviour. For all intents and purposes, non-cryptographic
+devices are a completely separate concept and do not exist from the perspective
+of the cryptography layer since they do not have identity keys, so it is
+impossible to send them encrypted messages.
```
foresto: I maintain a message notifier that does not encrypt or decrypt messages, but logs in on behalf of a user and receives encrypted message events in order indicate when unread ones exist. Encrypted messages' content remains opaque to my tool, but their arrival must be seen. With that in mind, this sentence concerns me:

> For all intents and purposes, non-cryptographic devices are a completely separate concept and do not exist from the perspective of the cryptography layer since they do not have identity keys, ***so it is impossible to send them encrypted messages.***

This could be interpreted to mean that a homeserver should avoid sending encrypted message events to non-cryptographic devices. Is that what's intended here? If so, it would render simple notifier clients like mine useless (unless some alternative was provided).

If not, can this be re-worded / clarified, please?

Moreover, it would be helpful if the spec explicitly allowed for clients of this kind, and spelled out what steps they should take and what steps are unnecessary in order to do their job without raising any alarms. The lack of guidance on this so far has forced me into considerable guesswork and experimentation just to build something as commonplace as a message notifier. The requirements shouldn't be mysterious or difficult, and clients like this shouldn't have to face regressions like element-hq/element-web#30709, which seems to result from lack of clarity in this area. Since this MSC seems to be the only document that comes close to addressing the matter, perhaps adding a paragraph of direct guidance would be worthwhile here.

tulir: That sentence is stating a fact (it is impossible to encrypt messages to a device that doesn't have encryption keys), not defining new behavior for servers. Defining that kind of behavior for servers would require wording along the lines of "omit `m.room.encrypted` events when a device without encryption keys calls /sync, /messages or other endpoints returning events".

The entire section is there just to say that devices without e2ee shouldn't affect anything, which is exactly what you're asking for.

I guess for an abundance of clarity, there could be a sentence calling out that non-e2ee-capable devices will still receive encrypted events and can observe any unencrypted metadata.

foresto: > That sentence is stating a fact (it is impossible to encrypt messages to a device that doesn't have encryption keys),

It's not impossible, though. One *can* encrypt and send messages to a device that lacks the keys. The device simply won't be able to decrypt them.

(Sorry if this seems pedantic, but experience has taught me that clarity is important in areas like this.)

How about rephrasing to something like this:

"For all intents and purposes, non-cryptographic devices are a completely separate concept and not expected to interact with the cryptography layer, since they do not have identity keys and therefore cannot read encrypted message content."

> I guess for an abundance of clarity, there could be a sentence calling out that non-e2ee-capable devices will still receive encrypted events and can observe any unencrypted metadata.

I think that might be helpful. It adds context to the next paragraph ("clients SHOULD NOT warn nor refuse to send messages due to the presence of non-cryptographic devices"). I have encountered clients misbehaving in this regard, so the additional guidance brought by this MSC is apparently needed in the spec.

Thank you.

tulir: > It's not impossible, though. One can encrypt and send messages to a device that lacks the keys. The device simply won't be able to decrypt them.

"Encrypting a message for a device" means sharing the Megolm keys through an Olm session to that device specifically. To establish an Olm session, the recipient device must have device keys and an one-time key for you to claim. Without device keys, it is literally impossible to encrypt a message for that device.

You can send an encrypted message in a room that includes a device that has no device keys, but when you do that, you have not encrypted the message *for* that device. You've merely sent an encrypted event the device can receive, but can't decrypt.

proposals/4153-invisible-crypto.md
tulir: No need to worry about mautrix-go, but it is a valid concern for less actively developed libraries. It'd probably be appropriate to define a recommended minimum time before this MSC is enabled by default, with the expectation that at least Element would follow that timeline, then be noisy about it in TWIM and other channels to make bot/library developers aware.

proposals/4153-invisible-crypto.md
uhoreg: I don't think there is anything currently specced regarding this, but in general, I think that clients only create cross-signing keys in response to a user action (e.g. they prompt the user to create cross-signing keys, and then wait for the user to click a button), so it's unlikely to race.  But I can write something to that effect in the doc.

proposals/4153-invisible-crypto.md
uhoreg: I don't understand this question.  The decryption key is not stored alongside any encrypted data.  The cross-signing keys and backup keys should be encrypted and stored in Secret Storage regardless of whether the client stores it locally or not.

proposals/4153-invisible-crypto.md
uhoreg: Element is planning on making an announcement soon about enabling this MSC, but it will probably be a bit before it's enabled by default.  The timing for enabling it by default will depend on somewhat on how well it's supported by other clients, bots, and bridges, and I'd expect that there will be noise made about it before it's enabled by default.  I don't want to specify a timeline, as various factors will affect it, but I would guess that it would be on the order of months.

proposals/4153-invisible-crypto.md
uhoreg: This thread is a duplicate of https://github.com/matrix-org/matrix-spec-proposals/pull/4153#discussion_r1740175938  Encryption without cross-signing is a huge security hole, and it's really bad that we're still allowing non-verified devices to participate in encrypted rooms.  Even though this may make certain use cases harder, I think that it's essential from a security perspective to implement this MSC.

proposals/4153-invisible-crypto.md
uhoreg: I've changed the wording to say "... it is impossible to send them decryption keys." which should hopefully be clearer.

proposals/4153-invisible-crypto.md
uhoreg: I've added a note about clients overwriting each other.  It's actually prevented by the fact that replacing your cross-signing keys require UIAuth.

turt2live:
@mscbot resolve Guidance to clients on how to create cross-signing keys (links to spec, conflict resolution, etc)

proposals/4153-invisible-crypto.md
TheArcaneBrony: Can we consider recommending that clients recommend to restrict access to e2ee chats rather than from using the client wholesale?

uhoreg: If the clients upload device keys, but don't cross-sign, then they'll be seen by other clients as potentially malicious, and could trigger warnings in other clients, so I don't think that would be a good option.  Clients could delay uploading device keys until the user is ready to verify, which is already covered under the "Non-cryptographic devices SHOULD NOT impact E2EE behaviour" section of this MSC.  It is up to the client developer whether they want to take this approach.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,270 @@
...
+ensures that users use cross-signing and key backup on new devices.
+
+The user's Secret Storage SHOULD have a default key (a key referred to by
+`m.secret_storage.default_key`) that encrypts the private cross-signing keys and
+key backup key (if available).
+
+### Verifying individual devices of other users is deprecated
+
+When one user verifies a different user, the verification SHOULD verify the
+users’ cross-signing keys.  Any flow between different users that does not
```
Johennes, line 58: Does "cross-signing keys" refer to the MSKs here? Or are there use cases for verifying USKs or SSKs?

proposals/4153-invisible-crypto.md
Johennes: > Element X does this, and is an absolute deal-breaker for me when I have to switch accounts on the go.

It sounds like the root cause here is lack of multi-account support in clients which feels orthogonal to this proposal.

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,270 @@
...
+far as to require cross-signing of devices by preventing the user from using
+the client until the device is cross-signed.  If the user cannot cross-sign
+their device (for example, if they have forgotten their Secret Storage key),
+the client can allow users to reset their Secret Storage, cross-signing, and
+key backup.
+
+### Clients SHOULD flag when cross-signing keys change
+
+If Alice’s cross-signing keys change, Alice’s own devices MUST alert her to
+this fact, and prompt her to re-cross-sign those devices.  If Bob is in an
```
Johennes, line 85: Does this include USK changes? I'm not sure what the use case would be but I think the USK could change without impacting self-verification?

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,270 @@
...
+defines a device verification method that uses encrypted to-device messages,
+such messages would need to be sent to a user's own non-cross-signed devices, so
+that the user can verify their device to cross-sign it.
+
+### Encrypted messages from non-cross-signed devices SHOULD be ignored
+
+Similarly, clients have no assurance that encrypted messages sent from
+non-cross-signed devices were sent by the user, rather than an
+impersonator. Therefore messages sent from non-cross-signed devices cannot be
+trusted and SHOULD NOT be displayed to the user.
```
Johennes, line 137: Does this include the case when the unsigned device is from the user's own account?

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,270 @@
...
+presents a way for bridge users to indicate that the bridge bot is allowed to
+perform encryption on their behalf.
+
+## Potential Issues
+
+### Client support
+
+If a user has devices that are not cross-signed, they will not be able to
+communicate with other users whose clients implement this proposal completely,
+due to the last two points.  Thus we encourage clients to implement
```
Johennes: I think the section about bridges may have made this not be "the last two points" anymore?

proposals/4153-invisible-crypto.md
```diff
@@ -0,0 +1,270 @@
...
+cross-signing as soon as possible, and to encourage users to cross-sign their
+devices, and clients should delay the implementation of the last two points (or
+make it optional) until most clients have implemented cross-signing.
```
Johennes: It feels slightly unclear what "delay" means for client developers here. I wonder if we need a roll-out plan similar to authenticated media?

turt2live: Authenticated media's rollout plan was "do the thing when you think you're ready", which is what the MSC feels like it's saying too.

Johennes: Hm, it was fixed in terms of spec versions and dates though? I suppose this might potentially be more of a question for spec writing rather than the proposal though.

Screenshot 2025-10-01 at 17 54 56

turt2live: As the text says, that's just an example timeline. We needed something so folks didn't leave unauthenticated media enabled forever, but we weren't overly concerned with the specific dates/targets. There are also some implicit forcing functions with authenticated media which make it a bit unique in terms of deployment/rollout - this MSC doesn't have those same functions.

What kind of guidance are you looking for as a developer?

Johennes: My concern is that it is unclear when clients should start enforcing this proposal. Granted, the rollout strategy probably doesn't need to be as complicated as with authenticated media. But it feels like _some_ strategy would be helpful because I think we cannot put these two statements into the spec at the same time:

> [...] clients MUST not send encrypted to-device messages [...] to non-cross-signed devices

> [...] clients should delay the implementation [...] (or make it optional) until most clients have implemented cross-signing

For instance, it would seem reasonable to me to split this proposal across two spec releases. The first release could include the recommendation to have secret storage and cross signing keys. The second release could include the exclusion of non-cross-signed devices and, building on the prior release, omit any indefinite delays.

proposals/4153-invisible-crypto.md
uhoreg: The verification verifies the MSK, which will indirectly verify the USK and SSK, since they are signed by the MSK.  None of the verification methods have a way of verifying the USK or SSK directly.

proposals/4153-invisible-crypto.md
uhoreg: Usually, the cross-signing keys are all changed together, which is why this MSC talks about them collectively, but it's basically just concerned about the MSK.  If the USK changes, but it's still signed by the MSK, and the MSK hasn't changed, then that would be fine.  If the USK changes, but isn't signed by the MSK, then that would be a separate error.

proposals/4153-invisible-crypto.md
Johennes: Oh, you mean verify through the chain. Got it, that makes sense.

proposals/4153-invisible-crypto.md
uhoreg: Yes.  In most cases, a user would probably be able to remember whether they said something or not, and so be able to tell if the message is fake, but since the device is untrusted, it should ignored.

proposals/4153-invisible-crypto.md
uhoreg: Yes, indeed.  As well as the section on non-cryptographic clients.  I'll fix the wording to point to the two sections by name.

proposals/4153-invisible-crypto.md
uhoreg: I don't think any action needs to be taken on the MSC in response to this thread, but I'm leaving this open for reference purposes.

proposals/4153-invisible-crypto.md
uhoreg: I'm not sure that we should be recommending a timeline in the MSC, but would it help if we changed it to "clients SHOULD not send encrypted to-device messages [...] to non-cross-signed devices", rather than "MUST"?  And then maybe change that to a "MUST" in a future MSC.

turt2live:
(this MSC is being held in its current state pending resolution of https://github.com/matrix-org/matrix-spec-proposals/pull/4153#discussion_r2393597572 )

proposals/4153-invisible-crypto.md
Johennes: I think that would be a little better but maybe I'm overthinking it entirely. Might be good to hear some other thoughts.

uhoreg: I've made the change.  I think that you're correct that a MUST is incompatible with a suggestion to delay implementation.

richvdh: @Johennes are you happy to consider this thread resolved now?

Johennes: Yes and apologies, I should have made that more clear. Thanks everyone for holding this just for me in the first place.

richvdh:
> (this MSC is being held in its current state pending resolution of [#4153 (comment)](https://github.com/matrix-org/matrix-spec-proposals/pull/4153#discussion_r2393597572) )

That thread is now resolved, and this MSC is now considered to have properly finished FCP.

proposals/4153-invisible-crypto.md
foresto: Thank you. That change makes a world of difference for at least one use case.

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2301

turt2live:
Merged 🎉

PR #4156, opened 2024-06-17
MSCXXXX: Migrate server_name to via

Johennes:


proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,57 @@
...
+proposal attempts to standardize on `via`.
+
+
+## Proposal
+
+The `server_name` query parameter on [`/_matrix/client/v3/join/{roomIdOrAlias}`] and
+[`/_matrix/client/v3/knock/{roomIdOrAlias}`] is deprecated and a new parameter `via: [string]` is
+introduced. Clients MUST use `via` when the homeserver they're talking to supports it. To do this, they
+MAY either detect server support through [`/_matrix/client/versions`] or always include both parameters.
+Homeservers MUST prefer `via` if both parameters are supplied.
```
clokep, line 23: "MUST prefer" meaning that only `via` should be used and `server_name` should be ignored? (Hopefully they're identical?)

Johennes: Hm, actually yeah that doesn't make any sense. What I was trying to say is that homeservers should look for `via` first and fallback to `server_name`. Though as you say, when they're identical, this probably doesn't matter to begin with.

proposals/4156-server-name-to-via.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Client

proposals/4156-server-name-to-via.md
clokep: I think there are a few things you could do:

* If you see `via` ignore `server_name`.
* Combine by preferring `via`, then falling back to `server_name` (being careful to not repeat).
* Error if `via` differs from `server_name`.

Not sure any of those are better or worse, but the current text isn't really clear which to do.

Johennes: > * If you see via ignore server_name.

This is what I meant to say but failed to express properly. Have updated the proposal.

The case where both parameters contain different values is a really weird one but I think it may not be worth to specify extra handling for this situation given that we don't actually want clients to use it that way?

proposals/4156-server-name-to-via.md
Johennes: Client: https://github.com/matrix-org/matrix-js-sdk/pull/4247

Johennes: Server: https://github.com/element-hq/synapse/pull/17322

Johennes: @turt2live would the Synapse implementation have to be enabled somewhere (e.g. beta.matrix.org) before this proposal can move ahead?

clokep: Implementations both look reasonable to me.

clokep:
I think this is the reasonable way forward.

@mscbot fcp merge

proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,57 @@
+# MSC4156: Migrate `server_name` to `via`
+
+Room IDs in Matrix are generally not routable on their own. In the [room ID grammar] `!opaque_id:domain`,
+the `domain` is the server that created the room. There is, however, no guarantee that this server is
+still joined to the room at a later time. Therefore, room IDs don't provide a reliable resident server
+to send requests to.
```
turt2live, line 6: Critically, the `domain` is *not* to be used as a routing server. It is purely a namespace.

proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,57 @@
...
+proposal attempts to standardize on `via`.
+
+
+## Proposal
+
+The `server_name` query parameter on [`/_matrix/client/v3/join/{roomIdOrAlias}`] and
+[`/_matrix/client/v3/knock/{roomIdOrAlias}`] is deprecated and a new parameter `via: [string]` is
+introduced. Clients MUST use `via` when the homeserver they're talking to supports it. To do this, they
+MAY either detect server support through [`/_matrix/client/versions`] or always include both parameters.
+Homeservers MUST ignore `server_name` if both parameters are supplied.
```
turt2live, line 23:
```suggestion
Homeservers MUST ignore all `server_name` parameters if any `via` parameters are supplied.
```

proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,57 @@
...
+The terminology difference between `server_name` and `via` can be slightly confusing which is why this
+proposal attempts to standardize on `via`.
+
+
+## Proposal
+
+The `server_name` query parameter on [`/_matrix/client/v3/join/{roomIdOrAlias}`] and
+[`/_matrix/client/v3/knock/{roomIdOrAlias}`] is deprecated and a new parameter `via: [string]` is
+introduced. Clients MUST use `via` when the homeserver they're talking to supports it. To do this, they
+MAY either detect server support through [`/_matrix/client/versions`] or always include both parameters.
```
turt2live, line 22: This feels like it's implying there's a stable flag clients can check for, but it doesn't describe what that is. If clients are meant to be using the spec release version, the MSC should say that.

I also don't think we can make it a `MUST` for clients to do this while we describe server behaviour when both/either parameter is supplied. We can likely reword this to say "Clients SHOULD supply both `via` and `server_name` while `server_name` is deprecated. `server_name` may be removed from the spec in a future MSC."

proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+## Alternatives
+
+None other than accepting status quo.
+
+
+## Security considerations
+
+None.
```
turt2live, line 40: There is a security concern here: if a client supplies different `via` and `server_name` parameters, a different room *could* be served to the user depending on which set of parameters the server uses to resolve the room ID. It may not be a useful or practical exploit, but it should be mentioned. If it is practical, that should be mentioned or discussed with security@matrix.org if the scope expands beyond this proposal.

proposals/4156-server-name-to-via.md
turt2live: it would be nice to see other clients concerned about this issue and expressing that with implementation, but I don't think this needs to be deployed to pass the implementation requirement.

It's currently hard to judge if this is a change we should be making.

turt2live (changes requested):

turt2live:
@mscbot concern Security considerations needs completing
@mscbot concern General clarity
@mscbot concern May require more client implementation (needs debate)

proposals/4156-server-name-to-via.md
Johennes: Have added this to the end of the paragraph.

proposals/4156-server-name-to-via.md
Johennes: > This feels like it's implying there's a stable flag clients can check for, but it doesn't describe what that is. If clients are meant to be using the spec release version, the MSC should say that.

That was my intention, yes. Have spelled it out in the proposal.

> I also don't think we can make it a `MUST` for clients to do this while we describe server behaviour when both/either parameter is supplied. We can likely reword this to say "Clients SHOULD supply both `via` and `server_name` while `server_name` is deprecated. `server_name` may be removed from the spec in a future MSC."

I have downgraded the client requirement to SHOULD. I had originally made it a MUST because I thought that would speed up the migration. However, as we discussed in matrix-spec a while ago, deprecations are generally a MUST for the server and a SHOULD NOT for the client.

proposals/4156-server-name-to-via.md
Johennes: Have added this into the proposal. I don't currently see a way in which a client could be tricked into sending different `via` and `server_name` parameters because the two endpoints touched upon in this proposal are the only places throughout the spec where we have two inputs.

proposals/4156-server-name-to-via.md
```diff
@@ -0,0 +1,64 @@
+# MSC4156: Migrate `server_name` to `via`
+
+Room IDs in Matrix are generally not routable on their own. In the [room ID grammar] `!opaque_id:domain`,
+the `domain` is the server that created the room. There is, however, no guarantee that this server is
+still joined to the room at a later time. Therefore, room IDs don't provide a reliable resident server
+to send requests to. Critically, the `domain` is not to be used as a routing server. It is purely a namespace.
```
dbkr, line 6: This is nice but honestly, for next time, I think it's unnecessary. I think it's fair to assume the audience is already familiar with the problem.

turt2live:
Thanks for the updates :) 

@mscbot resolve Security considerations needs completing
@mscbot resolve General clarity

It appears as though the implementation thread isn't receiving replies, but the MSC is getting quite a bit of traffic, so:

@mscbot resolve May require more client implementation (needs debate)

proposals/4156-server-name-to-via.md
turt2live: having the detail in the introduction I think is important, but agreed the phrasing is slightly off here. Not all readers will have context on the problem, particularly when it's a bit of obscure spec.

proposals/4156-server-name-to-via.md
turt2live: I'm assuming lack of comment means folks are comfortable with the level of implementation here.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1933

turt2live:
Merged 🎉

PR #4160, opened 2024-06-24
Clarify "real name" in contributor requirements

tulir:
This updates the sign-off requirements to match what most other matrix-org and element-hq repos already have. The change was first made in synapse: https://github.com/matrix-org/synapse/pull/3467

Same change in matrix-spec: https://github.com/matrix-org/matrix-spec/pull/1886


ara4n:
lgtm.

PR #4161, opened 2024-06-27
MSC4161: Crypto terminology for non-technical users

andybalaam:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/andybalaam/crypto-terminology/proposals/4161-crypto-terminology.md)

Conflict of Interest declaration: I am employed by Element. This MSC was written as part of my work on the Element Cryptography team.

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4161#issuecomment-4114085696)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4161#issuecomment-5254578690)

### DIFF
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -112,17 +112,17 @@ logged in and confirmed.
 In contrast to some other services, **logging out** (or **signing out**) of a
 Matrix device is quite a significant operation: it means your encryption data on
 the device is lost, and if you log out of all devices you will need to use your
-recovery key to re-establish your identity and regain access to your old
+recovery key to re-establish your digital identity and regain access to your old
 messages.
 
 If using a trusted physical device, the right choice for a user may well be not
 to log out, but simply to close the app or browser and re-open it later. This
-preserves their identity and their access to message history.
+preserves their digital identity and their access to message history.
 
 > "Are you sure you want to log out?"
 
 > "If you log out of all devices, you will lose access to message history and
-> will need to reset your identity."
+> will need to reset your digital identity."
 
 ### Verified user
 
@@ -133,31 +133,31 @@ this if you suspect someone in a room may be using a malicious homeserver.)
 In many contexts, most users are **not verified**: verification is a manual
 step ([scanning a QR code](https://spec.matrix.org/v1.12/client-server-api/#qr-codes) or [comparing emojis](https://spec.matrix.org/v1.12/client-server-api/#sas-method-emoji)). (In future, verification will
 probably become more common thanks to [MSC2882 Transitive Trust](https://github.com/matrix-org/matrix-spec-proposals/pull/2882) or something similar).
-When an unverified contact resets their identity, the user should be warned that
+When an unverified contact resets their digital identity, the user should be warned that
 this happened.
 
-If Alice is verified with Bob, and then Alice's identity changes
+If Alice is verified with Bob, and then Alice's digital identity changes
 (i.e. Alice resets their master cross-signing key) then this is very important to
 Bob: Bob verified Alice because they care about proof that no-one is listening,
 and now someone could be. Bob can choose to **withdraw verification** (i.e.
 "demote" Alice from being verified), or **re-verify** with Alice. Until Bob does
 one or the other, Bob's communication with Alice should contain a prominent and
-serious warning that Alice's **identity has been reset**.
+serious warning that Alice's **digital identity has been reset**.
 
 > "This user is verified."
 
-> "WARNING: Bob's identity has been reset!"
+> "WARNING: Bob's digital identity has been reset!"
 
-> "You verified this user's identity, but it has been reset. Please choose to
+> "You verified this user's digital identity, but it has been reset. Please choose to
 > re-verify them or withdraw verification."
 
 ⚠️ Avoid using "cross-signing", which requires a deeper understanding of
 cryptography to understand.
 
 ⚠️ Avoid using "trust on first use (TOFU)", which is a colloquial name for noting
-the identity of users who are not verified so that we can notify the user if it
+the digital identity of users who are not verified so that we can notify the user if it
 changes. (This is a kind of "light" form of verification where we assume that
-the first identity we can see is trusted.)
+the first digital identity we can see is trusted.)
 
 ⚠️ Avoid confusing verification of users with confirmation of devices: the
 mechanism is similar but the purpose is different. Devices must be confirmed to
@@ -165,54 +165,54 @@ make them secure, but users can optionally be verified to ensure no-one is
 listening in or tampering with communications.
 
 ⚠️ Avoid talking about "mismatch" or "verification mismatch" which is very
-jargony - it is the identity which is mismatched, not the verification process.
-Just say "Bob's identity has been reset".
+jargony - it is the digital dentity which is mismatched, not the verification process.
+Just say "Bob's digital identity has been reset".
 
 ⚠️ Where possible, avoid talking about "cryptographic identity" which is very jargony.
-In many contexts, just the word "identity" is sufficient: the dictionary definition of
+In many contexts, just the phrase "digital identity" is sufficient: the dictionary definition of
 identity meaning that someone is who they claim they are, not someone else. The
-fact we confirm identity cryptographically is usually irrelevant to the user.
+fact we confirm digital identity cryptographically is usually irrelevant to the user.
 
-### Identity
+### Digital Identity
 
-A user's **identity** is proof of who they are, and, if you have verified them,
-proof that you have a secure communication channel with them. Your own identity
+A user's **digital identity** is proof of who they are, and, if you have verified them,
+proof that you have a secure communication channel with them. Your own digital identity
 proves who you are, and gives you access to key storage.
 
-Technical note: we use "identity" here to describe a collection of keys: the
+Technical note: we use "digital identity" here to describe a collection of keys: the
 master signing key, user signing key, device signing key, key storage key and
 others.
 
-Your identity allows you to be identified by other users, and also allows you to
-access key storage and therefore see message history. This identity may be
+Your digital identity allows you to be identified by other users, and also allows you to
+access key storage and therefore see message history. This digital identity may be
 stored on the server by using recovery. The recovery key is not part of your
-identity, but allows you to re-establish your identity if you lose all your
+digital identity, but allows you to re-establish your digital identity if you lose all your
 devices.
 
-> When a non-verified user resets their identity:
-> "Alice's identity has been reset."
+> When a non-verified user resets their digital identity:
+> "Alice's digital identity has been reset."
 >
 > Longer explanation:
 > This can happen if the user lost all their devices and the recovery key, but
 > it can also be a sign of someone taking over the account. To be sure, please
-> verify their identity by going to their profile.
+> verify their digital identity by going to their profile.
 
-> When a verified user resets their identity:
-> "WARNING: Bob's identity has been reset!"
+> When a verified user resets their digital identity:
+> "WARNING: Bob's digital identity has been reset!"
 
 (During login, at the "Confirm it's you" stage):
 
 > "If you don't have any other device and you have lost your recovery key, you
-> can create a new identity. (Warning: you will lose access to your old
-> messages!)" button text (in red or similar): "Reset my identity"
+> can create a new digital identity. (Warning: you will lose access to your old
+> messages!)" button text (in red or similar): "Reset my digital identity"
 
-> "Are you sure you want to reset your identity? You will lose access to your
+> "Are you sure you want to reset your digital identity? You will lose access to your
 > message history."
 
 ⚠️ Avoid saying "master key" - this is an implementation detail.
 
 ⚠️ Avoid saying "Alice reset their encryption" - the change was to the user's
-identity.
+digital identity.
 
 References:
 
@@ -231,7 +231,7 @@ References:
   https://support.signal.org/hc/en-us/articles/6829998083994-Phone-Number-Privacy-and-Usernames-Deeper-Dive#verification
 
 All of the above are describing ways of proving who you are, which is close to
-the use of "identity" here.
+the use of "digital identity" here.
 
 ### Message key
 
@@ -315,7 +315,7 @@ for disaster recovery.
 
 ⚠️ Avoid talking about more keys: "the backup key is stored in the secret
 storage, and this allows us to decrypt the messages keys from key backup".
-Instead, we simply say that both identity and message keys are
+Instead, we simply say that both digital identity and message keys are
 stored in key storage.
 
 ### Recovery
@@ -323,22 +323,22 @@ stored in key storage.
 Recovery is useful when a user loses all their devices (or logs out of them
 all).
 
-If **recovery** is enabled, the user's identity is saved on the server, allowing
+If **recovery** is enabled, the user's digital identity is saved on the server, allowing
 them to recover it if they lose all their devices. This in turn allows them to
-recover their key storage and see message history. To recover their identity the
+recover their key storage and see message history. To recover their digital identity the
 user must enter the **recovery key**.
 
-The server is not able to read or manipulate the saved identity, because it is
+The server is not able to read or manipulate the saved digital identity, because it is
 encrypted using the recovery key.
 
-If a user loses their recovery key, they may **reset** their identity. Unless
+If a user loses their recovery key, they may **reset** their digital identity. Unless
 they have old devices, they will not be able to access old encrypted messages
-because the new identity does not have access to the old key storage.
+because the new digital identity does not have access to the old key storage.
 
 A **recovery key** (or **recovery code**) is a way of re-establishing your
-identity if you lose all your devices. This in turn allows you to access key
-storage, and therefore see message history. If you re-establish your identity
-instead of resetting it, other users won't see "Alice's identity has been reset"
+digital identity if you lose all your devices. This in turn allows you to access key
+storage, and therefore see message history. If you re-establish your digital identity
+instead of resetting it, other users won't see "Alice's digital identity has been reset"
 messages, and you will be able to read your message history, even if you logged
 out everywhere or lost your devices.
 
@@ -353,13 +353,13 @@ or "4S".
 > "Write down your recovery key in a safe place"
 
 > "If you lose access to your devices and your recovery key, you will need to
-> reset your identity, meaning you will lose all your message history"
+> reset your digital identity, meaning you will lose all your message history"
 
 ⚠️ Avoid "4S" or "quad-S" - these are not descriptive terms.
 
 ⚠️ Avoid using "security key", "security code", "master key". A
 recovery key allows "unlocking" the key storage, which is a "box" that is on the
-server, containing your identity and message keys. It is used to
+server, containing your digital identity and message keys. It is used to
 recover the situation if you lose access to your devices. None of these other
 terms express this concept so clearly.
 
@@ -379,29 +379,29 @@ in the spec as the
 #### Losing the recovery key
 
 If the user loses their recovery key, they no longer have a way to recover their
-identity.
+digital identity.
 
 If the user still has a secure device, then that device has its own copy of the
-identity information, so they can **change recovery key** without losing their
-identity, meaning other users will not see "Alice's identity has been reset", and
+digital identity information, so they can **change recovery key** without losing their
+digital identity, meaning other users will not see "Alice's digital identity has been reset", and
 they will be able to continue using key storage to access message history.
 
 Note: users should be encouraged to change their recovery key if they have forgotten
 their recovery key, because they are in a precarious position - if they lose
-access to their device, they will be forced to reset their identity and lose
+access to their device, they will be forced to reset their digital identity and lose
 message history.
 
 If the user does not have a device, or all their devices are insecure, then they
-will need to reset their identity, meaning other users
-see "Alice's identity has been reset", and they lose access to their old key
+will need to reset their digital identity, meaning other users
+see "Alice's digital identity has been reset", and they lose access to their old key
 storage, meaning they cannot read message history.
 
 > "If you lose your recovery key you can generate a new one if you are signed in
 > elsewhere"
 
-⚠️ Distinguish between "Reset identity" and "Change recovery key" - these are
-very different actions: resetting identity is destructive, whereas changing
-recovery key from a device that holds the full identity information is benign.
+⚠️ Distinguish between "Reset digital identity" and "Change recovery key" - these are
+very different actions: resetting digital identity is destructive, whereas changing
+recovery key from a device that holds the full digital identity information is benign.
 
 ## Potential issues
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -84,7 +84,7 @@ clients SHOULD document how they deviate from this document, and why.
 
 Instances of a client are called **devices** or alternatively **sessions**.
 Devices that have not been cross-signed by the user who owns them are
-**unconfirmed**. It is important that devices are confirmed, to prevent
+**unconfirmed**. It is important that devices are **confirmed**, to prevent
 security problems like a compromised server creating fake devices that
 can impersonate users (see
 [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)).
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -99,6 +99,14 @@ cryptography to understand.
 ⚠️ Avoid mentioning "device keys". While a device may have keys, the user
 is only concerned about whether it is secure or not.
 
+When a mechanism for simultaneously logging in and confirming a device
+is available (e.g.
+[MSC4108: QR code login](https://github.com/matrix-org/matrix-spec-proposals/pull/4108))
+then we talk about **linking** the device. After a device is linked, it is
+logged in and confirmed.
+
+> "Link this device by scanning the QR code"
+
 #### Logging out
 
 In contrast to some other services, **logging out** (or **signing out**) of a
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -84,11 +84,14 @@ clients SHOULD document how they deviate from this document, and why.
 
 Instances of a client are called **devices** or alternatively **sessions**.
 Devices that have not been cross-signed by the user who owns them are
-**insecure**.
+**unconfirmed**. It is important that devices are confirmed, to prevent
+security problems like a compromised server creating fake devices that
+can impersonate users (see
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)).
 
-> "This device is not secure. Please verify it to continue."
+> "This device is not confirmed. Please confirm it to continue."
 
-> "Confirm it's you" (when asking to verify a device during login)
+> "Confirm it's you" (when asking to confirm a device during login)
 
 ⚠️ Avoid using "cross-signing", which requires a deeper knowledge of
 cryptography to understand.
@@ -148,8 +151,8 @@ the identity of users who are not verified so that we can notify the user if it
 changes. (This is a kind of "light" form of verification where we assume that
 the first identity we can see is trusted.)
 
-⚠️ Avoid confusing verification of users with verification of devices: the
-mechanism is similar but the purpose is different. Devices must be verified to
+⚠️ Avoid confusing verification of users with confirmation of devices: the
+mechanism is similar but the purpose is different. Devices must be confirmed to
 make them secure, but users can optionally be verified to ensure no-one is
 listening in or tampering with communications.
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -125,8 +125,8 @@ is only concerned about whether it is secure or not.
 #### Logging out
 
 In contrast to some other services, **logging out** (or **signing out**) of a
-Matrix device is quite a significant operation: it means the encryption data on
-this device is lost, and if you log out of all devices you will need to use your
+Matrix device is quite a significant operation: it means your encryption data on
+the device is lost, and if you log out of all devices you will need to use your
 recovery key to re-establish your identity and regain access to your old
 messages.
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -217,7 +217,7 @@ specific meaning from cryptography.
 
 ### Recovery key (and recovery passphrase)
 
-A **recovery** key is a way of regaining access to key storage if the user loses
+A **recovery key** is a way of regaining access to key storage if the user loses
 all their devices. Using key storage, they can preserve their cryptographic
 identity (meaning other people don't see "Alice has reset their encryption"
 messages), and also read old messages using the stored message keys.
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -253,7 +253,7 @@ unable to decrypt it.
 When we expect the key to arrive, we are **waiting for this message**.
 
 > "Waiting for this message" with a button: "learn more" that explains that the message key for
-this message has not yet been received, but that we expect it to
+> this message has not yet been received, but that we expect it to
 > arrive shortly. Further detail may be provided, for instance explaining that
 > connectivity issues between the sender's homeserver and our own can cause
 > key delivery delays.
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -242,8 +242,8 @@ are "locked in a box" by encrypting them, and "unlocked" by decrypting them.
 ⚠️ Avoid using "room key". These keys are used to decrypt messages, not rooms.
 
 Note: this clashes with the term "message key" in the double ratchet. Since the
-double ratchet algorithm is for a very different audience, we think that this is
-not a problem.
+double ratchet algorithm is for a very different audience, this is not considered
+a blocking conflict.
 
 ### Unable to decrypt
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -160,7 +160,7 @@ proof that you have a secure communication channel with them.
 ⚠️ Avoid saying "master key" - this is an implementation detail.
 
 ⚠️ Avoid saying "reset their encryption" - the reason that Alice's identity
-changes could be due to attack rather than because they reset their encryption
+changed could be due to attack rather than because they reset their encryption
 (plus "encryption" is jargony).
 
 ### Message key
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -247,6 +247,13 @@ should provide a way for users to be educated on the terminology change (e.g. a
 tooltip or help link): e.g. "Your recovery key may also have been referred to as
 a security key in the past"
 
+⚠️ Be aware that at time of writing the spec uses
+["recovery key"](https://spec.matrix.org/v1.8/client-server-api/#recovery-key)
+to refer to the private half of the backup encryption key, which is different
+from the usage here. The recovery key described in this section is referred to
+in the spec as the
+[secret storage key](https://spec.matrix.org/v1.8/client-server-api/#secret-storage).
+
 ## Potential issues
 
 Lots of existing clients use a whole variety of different terminology, and many
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -102,8 +102,8 @@ cryptographic proof that no-one is listening in on your conversations. (You need
 this if you suspect someone in a room may be using a malicious homeserver.)
 
 In many contexts, most users are **not verified**: verification is a manual
-step (scanning a QR code or comparing emojis). (In future, verification will
-probably become more common thanks to "transitive trust" or "key transparency").
+step ([scanning a QR code](https://spec.matrix.org/v1.12/client-server-api/#qr-codes) or [comparing emojis](https://spec.matrix.org/v1.12/client-server-api/#sas-method-emoji)). (In future, verification will
+probably become more common thanks to [MSC2882 Transitive Trust](https://github.com/matrix-org/matrix-spec-proposals/pull/2882) or something similar).
 When an unverified user resets their identity, we should warn
 the user, so they are aware of the change.
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -87,7 +87,7 @@ do use the verb "verify" to describe how to make a device secure. By using the
 same verb, we help users understand the confusing fact that verifying devices
 and verifying users are similar processes, but with different outcomes.)
 
-⚠️ Avoid using "cross-signing", which requires a deeper understanding of
+⚠️ Avoid using "cross-signing", which requires a deeper knowledge of
 cryptography to understand.
 
 ⚠️ Avoid mentioning "device keys" - a device is just secure or not.
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -285,6 +285,9 @@ future).
 The keys inside key storage are themselves encrypted, so that the server
 operator is not able to access them and read your messages.
 
+In the spec, key storage is referred to as
+[server-side key backup](https://spec.matrix.org/v1.13/client-server-api/#server-side-key-backups).
+
 > "Allow key storage"
 
 > "Key storage holds the keys that allow you to read your message history."
@@ -327,6 +330,10 @@ out everywhere or lost your devices.
 
 A **recovery passphrase** is an easier-to-remember way of accessing the recovery
 key and has the same purpose as the recovery key.
+
+In the spec, recovery is referred to as
+[secret storage](https://spec.matrix.org/v1.13/client-server-api/#secret-storage),
+or "4S".
 
 > "Write down your recovery key in a safe place"
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -51,10 +51,10 @@ provide, and adapt their user interfaces and documentation to use them. (If this
 MSC is accepted, Element will use it as a reference for English wording in its
 clients.)
 
-Where concepts and terms exactly match existing terms in the Matrix spec, we
-propose changing the spec to use the terms from this document. Where they do not
-match, we are very comfortable with different words being used in the spec,
-given it is a highly technical document, as opposed to a client user interface.
+The Matrix spec SHOULD additionally be updated to use the concepts
+and terms from this proposal, where possible. For example, textual
+descriptions of features should use the terms here, but the API 
+endpoints and event types should *not* be updated. 
 
 We hope that this MSC will:
 
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -148,8 +148,8 @@ this if you suspect someone in a room may be using a malicious homeserver.)
 In many contexts, most users are **not verified**: verification is a manual
 step ([scanning a QR code](https://spec.matrix.org/v1.12/client-server-api/#qr-codes) or [comparing emojis](https://spec.matrix.org/v1.12/client-server-api/#sas-method-emoji)). (In future, verification will
 probably become more common thanks to [MSC2882 Transitive Trust](https://github.com/matrix-org/matrix-spec-proposals/pull/2882) or something similar).
-When an unverified user resets their identity, we should warn
-the user, so they are aware of the change.
+When an unverified contact resets their identity, the user should be warned that
+this happened.
 
 If Alice is verified with Bob, and then Alice's identity changes
 (i.e. Alice resets their master cross-signing key) then this is very important to
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -21,7 +21,30 @@ initiative is referred to as "Invisible Cryptography" and is tracked as:
   Including device keys with Olm-encrypted events, and
 * MSC4161 - this document
 
-## Goals
+## Why is this important?
+
+Use of common terminology should help further these goals:
+
+* **to reduce confusion**: many members of the community are confused by the
+  crypto features in Matrix clients, and the profusion of different words for
+  the same thing makes it much worse. By reducing the *number* of words, and
+  carefully choosing *good* words, we hope to develop a common language which
+  makes Matrix easier to understand, and easier to explain.
+
+* **to ease migration**: one of the key features of Matrix for end-users is the
+  choice of clients, meaning no-one is locked in to a particular piece of
+  software. If each client uses conflicting terminology, it becomes much more
+  difficult to move to a different client, which works against the user's
+  ability to migrate.
+
+This proposal uses "SHOULD" language rather than "MUST", because there are many
+good reasons why a particular client might choose different wording. In
+particular, different clients may have very different audiences who communicate
+in different ways and understand different metaphors. This proposal hopes to
+nudge client developers towards consistency, but never at the cost of their
+unique relationship with their users.
+
+## Outcomes
 
 We hope that Matrix client developers will like the terms and wording we
 provide, and adapt their user interfaces and documentation to use them. (If this
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -179,7 +179,7 @@ Your **message history** is a record of every message you have received or sent,
 and is particularly used to describe messages that are stored on the server
 rather than your device(s)
 
-###Key storage
+### Key storage
 
 **Key storage** means keeping cryptographic information on the server. This
 includes the user's cryptographic identity, and/or the message keys needed to
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -90,10 +90,11 @@ the parts of this section relating to insecure devices should be considered
 non-normative.
 
 Instances of a client are called 'devices' (not 'sessions'). Aligned with
-[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices taking part in encryption have been cross-signed by the
 user who owns them, and we call these **devices**.
 
-Devices which have not been cross-signed by the user are considered an error
+Devices which have published cryptographic keys (thus being visible as "cryptographic devices" to other users)
+but which have not been cross-signed are considered an error
 state, primarily to be encountered during the transition to MSC4153 and/or due
 to buggy/incomplete/outdated clients. These devices are referred to as **not
 secure** or **insecure** and their existence is considered a serious and dangerous error
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -49,7 +49,7 @@ the following terms and concepts.
 ### Devices
 
 Instances of a client are called 'devices' (not 'sessions'). Aligned with
-MSC4153, we take it as granted that all devices have been cross-signed by the
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
 user who owns them, and we call these **devices**.
 
 Devices which have not been cross-signed by the user are considered an error
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -83,12 +83,11 @@ clients SHOULD document how they deviate from this document, and why.
 
 ### Devices
 
-**Note: this section depends on [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)
-- Exclude non-cross-signed devices, which specifies clients should avoid sending
+**Note: this section depends on [MSC4153 ("Exclude non-cross-signed devices")](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), which specifies clients should avoid sending
 and receiving encryption info with devices that are not cross-signed by their
-owner ("insecure" devices in our terminology). While MSC4153 remains unmerged,
+owner ("insecure" devices in our terminology).** While MSC4153 remains unmerged,
 the parts of this section relating to insecure devices should be considered
-non-normative.**
+non-normative.
 
 Instances of a client are called 'devices' (not 'sessions'). Aligned with
 [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -62,7 +62,7 @@ condition, similar to an invalid TLS certificate.
 
 > "Ignoring 5 messages that were sent from a device that is not secure."
 
-> "Confim it's you" (when asking to verify a device during login)
+> "Confirm it's you" (when asking to verify a device during login)
 
 ⚠️ Avoid saying "secure device". All devices are considered secure by default;
 the user doesn't typically need to worry about the fact that insecure devices
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -180,6 +180,26 @@ Note: this clashes with the term "message key" in the double ratchet. Since the
 double ratchet algorithm is for a very different audience, we think that this is
 not a problem.
 
+### Unable to decrypt
+
+When we have an encrypted message but no message key to decrypt it, we are
+unable to decrypt it.
+
+When we expect the key to arrive, we are **waiting for this message**.
+
+> "Waiting for this message" button: "learn more" which explains that the key to
+> decrypt this message has not yet been received, but that we expect it to
+> arrive shortly. Further detail may be provided, for instance explaining that
+> connectivity issues between the sender's homeserver and our own can cause
+> key delivery delays.
+
+When the user does not have the message key for a permanent and well-understood
+reason, for example if it was sent before they joined the room, we say **you
+don't have access to this message**.
+
+> "You don't have access to this message" e.g. if it was sent before the user
+> entered the room, or the user does not have key storage set up.
+
 ### Message history
 
 Your **message history** is a record of every message you have received or sent,
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -1,0 +1,1 @@
+# MSC4161: Crypto terminology for non-technical users
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -73,8 +73,8 @@ example](https://github.com/matrix-org/matrix-spec/pull/1819/files#diff-8b25d378
 
 ## Proposal
 
-When communicating about cryptography with non-technical users, we propose using
-the following terms and concepts.
+When communicating about cryptography with non-technical users, the following
+terms and concepts SHOULD be used.
 
 When referring to concepts outlined in this document in their user interface,
 clients SHOULD use the language specified, except where their own users are
--- a/proposals/4161-crypto-terminology.md
+++ b/proposals/4161-crypto-terminology.md
@@ -81,30 +81,30 @@ the user doesn't typically need to worry about the fact that insecure devices
 are a thing, given they should only ever occur in error (or transitional)
 scenarios.
 
-⚠️ Avoid saying "trusted device" or "verified device". Devices are not people,
-and it is helpful to use different language for people vs. devices. (However, we
+⚠️ Avoid saying "trusted device" or "verified device". Devices are not users,
+and it is helpful to use different language for users vs. devices. (However, we
 do use the verb "verify" to describe how to make a device secure. By using the
 same verb, we help users understand the confusing fact that verifying devices
-and verifying people are similar processes, but with different outcomes.)
+and verifying users are similar processes, but with different outcomes.)
 
 ⚠️ Avoid using "cross-signing", which requires a deeper understanding of
 cryptography to understand.
 
 ⚠️ Avoid mentioning "device keys" - a device is just secure or not.
 
-⚠️ Avoid "session" to mean device. Device better describes what most people
+⚠️ Avoid "session" to mean device. Device better describes what most users
 encounter, and is more commonly used in other products.
 
-### Verified person
+### Verified user
 
-When you verify a person they become **verified**. This means that you have
+When you verify a user they become **verified**. This means that you have
 cryptographic proof that no-one is listening in on your conversations. (You need
 this if you suspect someone in a room may be using a malicious homeserver.)
 
-In many contexts, most people are **not verified**: verification is a manual
+In many contexts, most users are **not verified**: verification is a manual
 step (scanning a QR code or comparing emojis). (In future, verification will
 probably become more common thanks to "transitive trust" or "key transparency").
-When an unverified person resets their cryptographic identity, we should warn
+When an unverified user resets their cryptographic identity, we should warn
 the user, so they are aware of the change.
 
 If Alice is verified with Bob, and then Alice's cryptographic identity changes
@@ -115,24 +115,24 @@ and now someone could be. Bob can choose to **withdraw verification** (i.e.
 one or the other, Bob's communication with Alice should contain a prominent and
 serious warning that Alice's **verified identity has changed**.
 
-> "This person is verified."
+> "This user is verified."
 
 > "WARNING: Bob's verified identity has changed!"
 
-> "You verified this person's identity, but it has changed. Please choose to
+> "You verified this user's identity, but it has changed. Please choose to
 > re-verify them or withdraw verification."
 
 ⚠️ Avoid using "cross-signing", which requires a deeper understanding of
 cryptography to understand.
 
 ⚠️ Avoid using "trust on first use (TOFU)", which is a colloquial name for noting
-the identity of people who are not verified so that we can notify the user if it
+the identity of users who are not verified so that we can notify the user if it
 changes. (This is a kind of "light" form of verification where we assume that
 the first identity we can see is trusted.)
 
-⚠️ Avoid confusing verification of people with verification of devices: the
+⚠️ Avoid confusing verification of users with verification of devices: the
 mechanism is similar but the purpose is different. Devices must be verified to
-make them secure, but people can optionally be verified to ensure no-one is
+make them secure, but users can optionally be verified to ensure no-one is
 listening in or tampering with communications.
 
 ⚠️ Avoid talking about "mismatch" or "verification mismatch" which is very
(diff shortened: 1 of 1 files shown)


proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Proposal
+
+When communicating about cryptography with non-technical users, we propose using
+the following terms and concepts.
+
+### Devices
+
+Instances of a client are called 'devices' (not 'sessions'). Aligned with
+MSC4153, we take it as granted that all devices have been cross-signed by the
```
richvdh, line 52:
```suggestion
[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
```

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+this if you suspect someone in a room may be using a malicious homeserver.)
+
+In many contexts, most people are **not verified**: verification is a manual
+step (scanning a QR code or comparing emojis). (In future, verification will
+probably become more common thanks to "transitive trust" or "key transparency").
+When an unverified person resets their cryptographic identity, we should warn
+the user, so they are aware of the change.
+
+If a verified person's cryptographic identity changes (i.e. they reset their
+recovery key) then this is very important: we verified them because we care
```
richvdh, line 99: As a technical point of detail: it is not necessarily the reset of their recovery key that means their identity changes, but rather their master cross-signing key.

It is technically possible to replace your master cross-signing key without changing your recovery key, and vice-versa. It just happens to be that the most likely reason to rotate your MSK is because you've lost your recovery key.

```suggestion
If a verified person's cryptographic identity changes (i.e. they reset their
master cross-signing key) then this is very important: we verified them because we care
```

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+about proof that no-one is listening, and now someone could be. The user can
+choose to **withdraw verification** (i.e. "demote" them from being verified), or
+**re-verify** them. Until they do one or the other, communication with them should
+contain a prominent and serious warning that the user's **verified identity has
+changed**.
```
richvdh, line 104: a lot of confusion about which "user"/"person" we are talking about. Rephrase as "Alice/Bob"?

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+the first identity we can see is trusted.)
+
+⚠️ Avoid confusing verification of people with verification of devices: the
+mechanism is similar but the purpose is different. Devices must be verified to
+make them secure, but people can optionally be verified to ensure no-one is
+listening in or tampering with communications.
+
+⚠️ Avoid talking about "mismatch" or "verification mismatch" which is very
+jargony - it is the identity which is mismatched, not the verification process.
+Just say "Bob's verified identity has changed".
```
richvdh, line 128:
```suggestion
Just say "Bob's [verified] identity has changed".
```

may not be verified

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+unfamiliar to non-technical users.
+
+This document provides a list of concepts and explanations that are intended to
+be suitable for use in Matrix clients that are aimed at non-technical users.
+
+Ideally, encryption in Matrix should be entirely invisible to end-users (much as
+WhatsApp or Signal users are not exposed to encryption specifics) - as outlined
+in [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153).
+However, consistent terminology is still required, especially during the
+transition period to Invisible Cryptography.
```
clokep, line 16: MSC4153 is no longer called "Invisible Cryptography" -- is this simply referring to that MSC or is there something else? If something else, this really needs a definition or link.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+Where concepts and terms exactly match existing terms in the Matrix spec, we
+propose changing the spec to use the terms from this document. Where they do not
+match, we are very comfortable with different words being used in the spec,
+given it is a highly technical document, as opposed to a client user interface.
+
+We hope that this MSC will:
+
+* Cause small changes in the spec (as described in the previous paragraph), and
+* Become a document, blessed by the spec core team and available on the spec web
+  site, but probably separate from the spec proper.
```
clokep, line 34: This is pretty vague -- maybe an appendix would be appropriate?

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+> can create a new identity. (Warning: you will lose access to your old
+> messages!)" button text (in red or similar): "Reset my identity"
+
+⚠️ Avoid saying "master key" - this is an implementation detail.
+
+⚠️ Avoid saying "reset their encryption" - the reason that Alice's identity
+changes could be due to attack rather than because they reset their encryption
+(plus "encryption" is jargony).
+
+### Message key
```
ara4n, line 159: We're missing the official terminology for UTDs, and this feels like a great place to specify it (rather than UISIs or UTDs) etc. e.g. "Waiting for message" or "No access to decrypt this message" or whatever.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,276 @@
...
+Where changes are made in the spec, we suggest that notes be added mentioning
+the old name, as in [this
+example](https://github.com/matrix-org/matrix-spec/pull/1819/files#diff-8b25d378e077f18eb06ebdb9c376e194c8a4c8b95cf909fca6448659a627f283R1326).
+
+## Proposal
+
+When communicating about cryptography with non-technical users, we propose using
+the following terms and concepts.
+
+### Devices
```
greentore, line 49: I really don't like calling sessions devices. It may work well for other, mobile first networks, but with matrix it's not all that uncommon to run multiple sessions on the same device.

uhoreg: The problem is that "sessions" is used to refer to other things in the spec.  Also, the Matrix spec already refers to things such as "device IDs" that use the "device" terminology.

proposals/4161-crypto-terminology.md
pmaier1: Yup, indeed. I'd suggest "Waiting for this message" with a "Learn more" guiding to a FAQ page that explains why users see this and gives some guidance.

proposals/4161-crypto-terminology.md
pmaier1: [MSC4153](https://github.com/matrix-org/matrix-doc/pull/4153) has been renamed to "Exclude non-cross-signed devices" to make the intention clear. We refer to Invisible Crypto is an umbrella term to summarize quite a couple of things, including this MSC. It also includes other efforts to make crypto more 'invisible' to users like [MSC4048](https://github.com/matrix-org/matrix-spec-proposals/pull/4048), [MSC4147](https://github.com/matrix-org/matrix-spec-proposals/pull/4147) and [MSC4153](https://github.com/matrix-org/matrix-doc/pull/4153).

proposals/4161-crypto-terminology.md
MTRNord: How about explicitly naming them "Crypto-Sessions" or something in this direction? I agree with the initial mention about devices. This is already these days due to some UIs a bit confusing.

andybalaam: My personal take is that most people are familiar with "devices", and non-technical users will probably only have one crypto session per device. Further, where people do have multiple sessions on a device I think is it relatively easy to explain that they look like multiple devices to Matrix even though they exist on the same physical device.

In contrast, when I started using Matrix I had no idea what a "session" was, and as I learned more I became further confused because the word is used for several other concepts, including e.g message keys.

So I plan to propose this as-is.

proposals/4161-crypto-terminology.md
andybalaam: I made an attempt in de285aa0

proposals/4161-crypto-terminology.md
andybalaam: Updated in a4b95b46

proposals/4161-crypto-terminology.md
andybalaam: I like it. Proposed in 0499900a

proposals/4161-crypto-terminology.md
andybalaam: First draft added in b919630c

proposals/4161-crypto-terminology.md
TheArcaneBrony: as much as i hate the "devices" terminology, this is also how it's named over on eg. Discord.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+⚠️ Avoid saying "master key" - this is an implementation detail.
+
+⚠️ Avoid saying "reset their encryption" - the reason that Alice's identity
+changed could be due to attack rather than because they reset their encryption
+(plus "encryption" is jargony).
+
+### Message key
+
+A **message key** is used to decrypt a message. The metaphor is that messages
+are "locked in a box" by encrypting them, and "unlocked" by decrypting them.
```
Saiv46, line 169: Better term could be **encryption key**, as it's more common in news.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+
+If a user loses their recovery key, they may **reset** their key storage. Unless
+they have old devices, they will not be able to access old encrypted messages
+because the message keys are stored in key storage, and their cryptographic
+identity will change, because it too is stored in key storage.
+
+> "Allow key storage"
+
+> "Key storage holds your cryptographic identity on the server along with the
+> keys that allow you to read your message history."
```
Saiv46, line 221: > ⚠️ Avoid talking about "cryptographic identity" which is very jargony.

Saiv46 (approved):

Saiv46:
This should be given more priority. User-friendliness of clients is a must

proposals/4161-crypto-terminology.md
andybalaam: Hmm, yeah. Just "Key storage holds your identity"? Or maybe "Key storage holds your identity information"? @pmaier1 any thoughts?

proposals/4161-crypto-terminology.md
turt2live, line 1: Overall review: The MSC appears to describe features and requirements of Element, and assumes that all clients in the ecosystem have similar or comparable features/needs. The assumptions read fine when applying this MSC to an Element client, but when comparing to the spec there feels like there's a gap in understanding. 

Significant review from other client authors/developers, and their respective product/design teams, I think would be extremely valuable to help ensure this MSC does not solely apply to Element. This kind of review may need to be chased down as it's somewhat uncommon for a product-y MSC to be in the spec process.

turt2live (changes requested):
This MSC is something I think would be extremely helpful in the spec. It's important to me that users get a relatively consistent experience regardless of the client they choose to use, but still allowing those clients to express their unique features/competitive advantage to users. Clients should also have capacity to participate in different markets outside of chat.

The MSC also appears very nearly ready for FCP, with only a few blocking comments from my side:
* The MSC could do with a bit stronger rationale for *why* this is important. 
* The MSC appears to have a heavy focus on Element clients, which is natural given the authors. 
* To help with the "why" of the MSC, "implementation" in the form of user research would help. Ideally from a broad spectrum of users and clients, demonstrating the terms chosen carry well between different markets, audiences, brands, etc.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+same verb, we help users understand the confusing fact that verifying devices
+and verifying people are similar processes, but with different outcomes.)
+
+⚠️ Avoid using "cross-signing", which requires a deeper understanding of
+cryptography to understand.
+
+⚠️ Avoid mentioning "device keys" - a device is just secure or not.
+
+⚠️ Avoid "session" to mean device. Device better describes what most people
+encounter, and is more commonly used in other products.
```
richvdh, line 91: Can we be explicit that "insecure device" (rather than "not secure device") is permitted?

proposals/4161-crypto-terminology.md
dkasak: The problem is there are loads of things in Matrix that could be called an "encryption key". This is specifically a "message encryption key", shortened to "message key".

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+> "You don't have access to this message" e.g. if it was sent before the user
+> entered the room, or the user does not have key storage set up.
+
+### Message history
+
+Your **message history** is a record of every message you have received or sent,
+and is particularly used to describe messages that are stored on the server
+rather than your device(s)
+
+### Key storage
```
dkasak, line 207: I would call this "Secret storage".

The mechanism in question can store arbitrary data, not all of which are keys. The next paragraph already has to perform significant linguistic gymnastics in order to explain that the user's identity is also stored in the *key* storage, even though it is called an *identity*, not a *key*. (The fact that it is a key is an unimportant technical detail from a user perspective and is very jargony.)

On the other hand, explaining that a key is a certain type of secret should be very natural.

proposals/4161-crypto-terminology.md
pmaier1: > Key storage holds your identity

Yeah, keep it simple, makes sense to me.

dkasak: See https://github.com/matrix-org/matrix-spec-proposals/pull/4161/files#r1778410962 for my concerns about this. "*Key* storage holds your *identity*" sounds extremely confusing to me if I imagine myself as a non-technical newcomer to Matrix.

proposals/4161-crypto-terminology.md
pmaier1: > as much as i hate the "devices" terminology, this is also how it's named over on eg. Discord.

or Whatsapp or Signal or other platforms. Which is why we suggest it in the first place.

> Has user research shown that devices are preferred over sessions? Does Element have some data on their existing transition from devices to sessions?

Not sure whether we have something dedicated to exactly this comparison. @americanrefugee might know more.

proposals/4161-crypto-terminology.md
pmaier1: > I would call this "Secret storage".

We discussed this. The challenge we see is that it could be mixed up. It could either mean
1. a storage for secrets
2. a storage that is secret

That's why we discarded it and propose "Key storage". For a user it is irrelevant if something is a secret or a key or a token or a hash or whatnot, technically speaking. From a users point of view it should be a safe place that holds _keys_ to open up stuff on new devices or in case of an emergency.

dkasak: Also, the problem will be further exacerbated if/when new non-key things start getting stored in secret storage.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+storage, and this allows us to decrypt the messages keys from key backup".
+Instead, we simply say that both cryptographic identity and message keys are
+stored in key storage.
+
+⚠️ Avoid using "key backup" to talk about storing message keys: keeping things on
+the server is not a "backup", but a reliable, cross-device place where this
+information is stored. The word "backup" implies a redundant way to recover lost
+information, but if the user loses their recovery key, this information is lost.
+Clients and servers may wish to offer additional backup services that provide
+true redundancy and disaster recovery, but key storage is not this.
```
dkasak, line 238: I don't entirely agree with this paragraph as it misses some important nuance. If a user loses their recovery key, they will still have a copy of some (potentially all) keys on their signed in clients. It's entirely possible to run a functioning client without key storage.

> a reliable, cross-device place where this information is stored

Moreover, this sounds like a description fitting of the word "backup".

> The word "backup" implies a redundant way to recover lost information, but if the user loses their recovery key, this information is lost.

And the two clauses here don't sound as a contradiction to me. Backups can certainly be gated behind encryption and authentication. This seems to imply that a backup is only truly a backup if no additional information is required to access it.

To be clear, I agree with phasing out the term "backup" for "storage", but IMO the paragraph above does not provide a good argument for this.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+
+⚠️ Avoid "4S" or "quad-S" - these are not descriptive terms.
+
+⚠️ Avoid "private key" - this is an implementation detail and a term with
+specific meaning from cryptography.
```
dkasak, line 243: I don't understand this paragraph in the context of its section. Avoid the term "private key" when talking about... what? Key storage? If so, I've never heard someone refer to "key storage" as "private key".

proposals/4161-crypto-terminology.md
Xiretza: > We discussed this. The challenge we see is that it could be mixed up. It could either mean
> 
>     1. a storage for secrets
> 
>     2. a storage that is secret

Wouldn't "Secrets storage" fix that problem?

proposals/4161-crypto-terminology.md
afranke: Gmail uses “session”. I’m sure others do too. I don’t think it’s such a foreign term for laypeople, or that it can’t be understood by them.

image

Whatsapp and Signal are more or less tied to phones and don’t expect any other client than “the one app” you have on said phones. In their context, device works. Comparing to them doesn’t work.

afranke: > The problem is that "sessions" is used to refer to other things in the spec. So it would be more confusing to call something else "sessions".

This MSC is about terms to use for non-technical users as far as I can see. I reckon it is not a problem to use a word for something in a given context, and a different word for the same thing in another context. The spec can call these crypto-sessions as was suggested above, or even devices if further disambiguation is deemed necessary. In a technical context, it can be clear that a “device” is not a piece of equipment, whereas it is really not clear for non-technical users.

Without even thinking about multiple distinct clients on the same device, a single client can become a new session if access to the previous session is lost for some reason. And so now you have two “devices” listed even though you used a single computer and a single client.

andybalaam: Element's product people favour "devices", as do Discord, Signal and WhatsApp. We have a counter-example in Google, but I personally find much of Google's UI quite confusing, especially gmail.

As for having two devices listed even though you only used one: the proposed wording seems helpful to me - the mismatch between the wording and reality indicates a problem: they should remove the old device.

Perhaps the product designers from other products could chip in, but at the moment the scale seems to be tipping towards "devices".

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+We hope that this MSC will:
+
+* Cause small changes in the spec (as described in the previous paragraph), and
+* Become an appendix in the spec, with a description that makes clear that the
+  intended audience is different from most of the spec, meaning different words
+  are used from the main spec body.
+
+Clients may, of course, choose to use different language. Some clients may
+deliberately choose to use more technical language, to suit the profiles of
+their users. This document is aimed at clients targeting non-technical users.
```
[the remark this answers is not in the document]
andybalaam: I wrote some words in 56b5daf1 - let me know what you think.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+cryptography to understand.
+
+⚠️ Avoid mentioning "device keys" - a device is just secure or not.
+
+⚠️ Avoid "session" to mean device. Device better describes what most people
+encounter, and is more commonly used in other products.
+
+### Verified person
+
+When you verify a person they become **verified**. This means that you have
```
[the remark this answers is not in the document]
andybalaam: Thanks! Fixed in fba4f5ca

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+
+> "This person is verified."
+
+> "WARNING: Bob's verified identity has changed!"
+
+> "You verified this person's identity, but it has changed. Please choose to
+> re-verify them or withdraw verification."
+
+⚠️ Avoid using "cross-signing", which requires a deeper understanding of
+cryptography to understand.
```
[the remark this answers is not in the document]
andybalaam: Fixed in 9246021e

proposals/4161-crypto-terminology.md
andybalaam: Thanks, yeah not sure it's useful. Removed in 4bcee330

proposals/4161-crypto-terminology.md
andybalaam: Yep, good points. I had another go here: f654c2e6 - let me know what you think.

proposals/4161-crypto-terminology.md
andybalaam: "Secrets storage" is feels unnatural to say in English. The justification for "key storage" is that the most important things (from a non-technical user's point of view) that are stored in this storage are message keys.

The fact that other things also get stored there doesn't need to prevent us naming it after the most important thing. For example, if I have a school locker for P.E. kit that I commonly call a "kit locker" I think it would be perfectly natural for me to say "I keep my Sony Walkman and Rubik's cube in my kit locker"...

proposals/4161-crypto-terminology.md
andybalaam: Fixed in 3c7695f2. Originally I avoided "insecure" because I thought it might cause confusion with e.g. a device that was out of date in terms of security fixes, but in retrospect I think that potential confusion exists either way, and avoiding "insecure" made some of the language clunky.

proposals/4161-crypto-terminology.md
Xiretza: > The fact that other things also get stored there doesn't need to prevent us naming it after the most important thing

It doesn't *need* to, but I think it should, considering that the whole point of this exercise is to make the language used less confusing.

proposals/4161-crypto-terminology.md
dkasak: I think this is much better. Thanks.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,308 @@
...
+
+Your **message history** is a record of every message you have received or sent,
+and is particularly used to describe messages that are stored on the server
+rather than your device(s)
+
+### Key storage
+
+**Key storage** means keeping cryptographic information on the server. This
+includes the user's cryptographic identity, and/or the message keys needed to
+decrypt messages.
```
andybalaam, line 216: I want to split this into 2 separate concepts: I thought we could run storage of message keys together with storage of identity keys etc (recovery) but recently I realised they a really separate, so we probably need something like this:

* **Room key storage** - store room keys on the server, protected by a room key storage key. This room key storage key can be passed around between devices even if you don't have recovery storage.
* **Recovery storage** - store identity keys and the room key storage key on the server, protected by a recovery code. This allows us to recover both our identity and our room key storage key even if we lose all our devices.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,308 @@
...
+
+⚠️ Avoid using "key backup" to talk about storing message keys: this is too
+easily confused with exporting keys or messages to an external system.
+
+⚠️ Avoid "4S" or "quad-S" - these are not descriptive terms.
+
+### Recovery key (and recovery passphrase)
+
+A **recovery key** is a way of regaining access to key storage if the user loses
+all their devices. Using key storage, they can preserve their cryptographic
```
tulir, line 246: For the record, Beeper settled on "recovery code" after considering other options including "recovery key", but I don't remember the exact reasoning since the switch was made 2 years ago

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,308 @@
...
+fact we confirm identity cryptographically is irrelevant to the user;
+cryptography should be invisible.
+
+### Identity
+
+A user's **identity** is proof of who they are, and, if they are verified,
+proof that you have a secure communication channel with them.
+
+> "Warning: Alice's identity appears to have changed" (when a non-verified
+> user resets their recovery key)
```
andybalaam, line 154: We have been trialing the "appears to have changed" wording in Element clients, and it is not popular at all, because it suggest we are unsure what has happened.

The intention of the wording was to make clear that although we are sure that the _cryptographic_ identity has changed, we can't be sure whether this is because the real person did a reset, or some malicious actor is impersonating them.

But this is clearly not working, and we need to work on different wording.

andybalaam: I wonder whether "Alice's identity has changed: there is a chance someone took over their account - to be sure, you should verify their identity in their profile" or something else action-orientated would be helpful.

BillCarsonFr: I think the chances that someone took over the account are in practice very low, and it is alarming to say it up-front. I'd go with `Alice probably re-installed the app and lost her recovery key, you can always verifiy to ensure the identity is legitimate`

richvdh: +1 to "Alice's identity has changed". I think most people will get that. Especially if we can couple it with "there is a chance someone took over their account".

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,308 @@
...
+⚠️ Remember that users may have historically been trained to refer to these
+concepts as "security key" or "security passphrase", and so user interfaces
+should provide a way for users to be educated on the terminology change (e.g. a
+tooltip or help link): e.g. "Your recovery key may also have been referred to as
+a security key in the past"
+
+⚠️ Be aware that at time of writing the spec uses
+["recovery key"](https://spec.matrix.org/v1.8/client-server-api/#recovery-key)
+to refer to the private half of the backup encryption key, which is different
+from the usage here. The recovery key described in this section is referred to
```
richvdh, line 276: This got fixed in https://github.com/matrix-org/matrix-spec/pull/1819

proposals/4161-crypto-terminology.md
tulir: Asked for the reasoning internally:

> Code sounded more friendly than key (in an unscientific review).
> e.g. people know what a pin code is. Code is a numerical thing but a 'key' to most people is a physical object

proposals/4161-crypto-terminology.md
dkasak: Why not copy what other messengers do?

> Alice's identity has changed. [This can happen if the user lost all their devices and the recovery key, but it can also be a sign of someone taking over the account].

Clients are suggested to always display something to the effect of the first sentence. The rest (what's in the brackets) should be available but can be hidden behind a "Tap for more info." interaction or similar.

ara4n: +1 to "Alice's identity has changed." I do think we should be action-focused too, though, so: "Please verify them".

dkasak: Note that this thread is specifically about the non-verified case. It makes sense to mention verification but probably in a non-binding way. I still think there is room for a short single sentence explanation of why this happened, especially if hidden behind an interaction.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,308 @@
...
+
+Instances of a client are called 'devices' (not 'sessions'). Aligned with
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
+user who owns them, and we call these **devices**.
+
+Devices which have not been cross-signed by the user are considered an error
+state, primarily to be encountered during the transition to MSC4153 and/or due
+to buggy/incomplete/outdated clients. These devices are referred to as **not
+secure** or **insecure** and their existence is considered a serious and dangerous error
+condition, similar to an invalid TLS certificate.
```
andybalaam, line 71: Proposed new wording: "Encryption login" - your device is insecure because you have not completed the encryption login.

andybalaam: encryption-login

Mock-up of how this might be communicated to user.

andybalaam: Hmm, lots of push-back on this wording in our team, but we like the diagram.

We are playing around with something like "verify it's you".

proposals/4161-crypto-terminology.md
richvdh: It's worth noting that the EX designs (https://www.figma.com/design/qTWRfItpO3RdCjnTKPu4mL/Settings?node-id=1653-29941&node-type=frame&t=LeUe8iJVmzG8srM5-0 etc) refer to "key storage" to cover both of these concepts. I think that's ok; we just need to be clear that "key storage" is used for both *room keys* and *recovery*, and that they are set up separately.

richvdh: > Room key storage - store room keys on the server, protected by a room key storage key.

NB they are *message* keys, not room keys.

pmaier1: IMO it would be great if we could abstract the details to the user. It should just be one feature from a user POV. Either you want the server to store some information for you and get benefits from it or you don't. I don't see a reason why the user should need to understand the exact concepts behind it.

richvdh: Well, it would be quite a change from the way the apps currently work. Currently EX sets up key backup on registration, but doesn't set up recovery until you actively go through the steps (and EW is even more flexible). That's also why the designs above have separate entries for "Allow key storage" and "set up recovery".

Possibly we could change that, and force the user to set up recovery during the registration/login flow; but it would go against the desire to get people using the app quickly. (Also, this MSC is intended to cover matrix clients in general: even if the Element clients force you to set up recovery at the same time as key backup, other clients may not, and having the terminology available to refer to that situation is important.)

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,344 @@
...
+clients SHOULD document how they deviate from this document, and why.
+
+### Devices
+
+Instances of a client are called 'devices' (not 'sessions'). Aligned with
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), we take it as granted that all devices have been cross-signed by the
+user who owns them, and we call these **devices**.
+
+Devices which have not been cross-signed by the user are considered an error
+state, primarily to be encountered during the transition to MSC4153 and/or due
```
andybalaam, line 91: See the comments in [WorkAdventure's blog post](https://workadventu.re/article-en/managing-e2e-encryption-with-matrix-in-a-simple-way/) under "Compatibility with Element X (and the invisible crypto plan)".

We should make it clear that only unsigned clients _that are doing encryption_ are a problem. Clients that simply don't use any encrypted rooms are fine without being signed.

dkasak: See also https://github.com/matrix-org/matrix-spec-proposals/pull/4153#discussion_r1928557275

proposals/4161-crypto-terminology.md
richvdh: I think it's worth pointing out that the spec and protocol *already* uses the term "device" for this concept. See in particular:
 * https://spec.matrix.org/v1.13/#devices which discusses the fact that there is not a 1:1 relationship between Matrix devices and physical devices
 * The fact that `/login` returns a `device_id`
 * The fact that [send-to-device messages](https://spec.matrix.org/v1.13/client-server-api/#send-to-device-messaging) are called that.
 
[@andybalaam: we could probably do with an Alternatives section in the MSC, which talks a bit more about why we're recommending "device" rather than "session": that would be a good place to include some of those links].

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+to buggy/incomplete/outdated clients. These devices are referred to as **not
+secure** and presence of them are considered a serious and dangerous error
+condition, similar to an invalid TLS certificate.
```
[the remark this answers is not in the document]
richvdh: > Maybe we could include it, but note that it is describing a post-[MSC4153](https://github.com/matrix-org/matrix-doc/pull/4153) world?

It would be much easier if we could just land MSC4153, but in the meantime, yes, this sounds like the best way.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,344 @@
...
+
+Your **message history** is a record of every message you have received or sent,
+and is particularly used to describe messages that are stored on the server
+rather than your device(s)
+
+### Key storage
+
+**Key storage** means keeping cryptographic information on the server. This
+includes the user's identity, and/or the message keys needed to
+decrypt messages.
```
richvdh, line 243: I think it would be helpful to root this in terms of the underlying protocol:

```suggestion
**Key storage** means keeping cryptographic information on the server. In
terms of the underlying protocol, this consists of two parts: [server-side
key backups](https://spec.matrix.org/v1.13/client-server-api/#server-side-key-backups), and
[server-side key storage](https://spec.matrix.org/v1.13/client-server-api/#key-storage).
```

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,344 @@
...
+
+> "Allow key storage"
+
+> "Key storage holds your identity on the server along with the
+> keys that allow you to read your message history."
+
+> "Message history is unavailable because key storage is disabled."
+
+⚠️ Avoid distinguishing between "secret storage" and "key backup" - these are
+both part of key storage.
```
richvdh, line 258: This feels somewhat confusing as written.

Basically, "secret storage" and "recovery" are the same mechanism, and the Element clients absolutely do distinguish between "Key storage" and "recovery". 

I think what we probably want to say here is "avoid the terms 'key backup' and 'secret storage'".

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,344 @@
...
+key and has the same purpose as the recovery key.
+
+**Losing the recovery key**: if the user loses their recovery key, they can
+"reset" it, which means re-storing the identity information in the server,
+encrypted with a new recovery key. If the user has a verified client, then that
+is holding the identity information locally, so they can reset their recovery
+key without losing access to key storage. If they don't have a verified client
+and they lose their recovery key, then they need to reset key storage as well as
+recovery key (since the identity information is needed to read from key
+storage), meaning they lose access to old messages.
```
richvdh, line 287: We now have text in both the "Key storage" and the "Recovery key" sections which talk about what happens if the user loses the recovery key, and they say slightly different things. I think we need to rationalise the two.

proposals/4161-crypto-terminology.md
richvdh: I don't really think this is there yet, tbh. See other comments about it, but I feel like it's hard to set this out sensibly without making it clear that there are two parts to Key storage: message keys, and recovery. 

Sure the recovery key gives you access to key storage, but only if you have previously set up recovery. We have to be able to talk about "setting up recovery" as a separate step to "enabling key storage".

proposals/4161-crypto-terminology.md
andybalaam: Added an alternatives section in c13bd45e

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
andybalaam: Added a note in d0878852

proposals/4161-crypto-terminology.md
andybalaam: I did a rewrite in a9383723 . Better?

proposals/4161-crypto-terminology.md
andybalaam: This is not directly relevant after my rewrite. Did you still want to relate Key Storage with "server-side key backup" and Recovery with "server-side secret storage (4S)"?

proposals/4161-crypto-terminology.md
andybalaam: This is not relevant since my re-write.

proposals/4161-crypto-terminology.md
andybalaam: I think I addressed this in the rewrite a93837230

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
richvdh: @turt2live are you happy to resolve this thread?

proposals/4161-crypto-terminology.md
```diff
@@ -83,6 +83,13 @@ clients SHOULD document how they deviate from this document, and why.
 
 ### Devices
 
+**Note: this section depends on [MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)
+- Exclude non-cross-signed devices, which specifies clients should avoid sending
+and receiving encryption info with devices that are not cross-signed by their
+owner ("insecure" devices in our terminology). While MSC4153 remains unmerged,
+the parts of this section relating to insecure devices should be considered
+non-normative.**
```
richvdh, line 91: Are the `**`s supposed to make this bold? They aren't working.

richvdh: oh I see, that "-" is supposed to be a dash, not the start of a list.

```suggestion
**Note: this section depends on [MSC4153 ("Exclude non-cross-signed devices")](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), which specifies clients should avoid sending
and receiving encryption info with devices that are not cross-signed by their
owner ("insecure" devices in our terminology).** While MSC4153 remains unmerged,
the parts of this section relating to insecure devices should be considered
non-normative.
```

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,310 @@
...
+a security key in the past"
+
+⚠️ Be aware that at time of writing the spec uses
+["recovery key"](https://spec.matrix.org/v1.8/client-server-api/#recovery-key)
+to refer to the private half of the backup encryption key, which is different
+from the usage here. The recovery key described in this section is referred to
+in the spec as the
+[secret storage key](https://spec.matrix.org/v1.8/client-server-api/#secret-storage).
+
+## Potential issues
```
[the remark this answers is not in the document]
richvdh: @turt2live are you happy to resolve this thread?

proposals/4161-crypto-terminology.md
richvdh: Will close this thread in favour of others.

proposals/4161-crypto-terminology.md
richvdh: It seems like we're now proposing "Key storage" to specifically mean key backup, rather than also lumping 4S into it, which seems like it makes this thread outdated?

proposals/4161-crypto-terminology.md
richvdh: I think it might be helpful, yes; doing so will (a) help readers familiar with the current paradigms; (b) help people correlate these recommendations to the concepts in the spec.

proposals/4161-crypto-terminology.md
richvdh: actually we need to tweak the previous paragraph too.

proposals/4161-crypto-terminology.md
andybalaam: Done in ecfd1ad4

proposals/4161-crypto-terminology.md
andybalaam: Yes, that's right. The lumping-together of those 2 ideas didn't work, because whereas key storage can be set up without user interaction, recovery requires the user to acknowledge and write down the recovery key, so the UI must separate them, meaning we get no benefit from lumping them together.

So now:

key storage = storing message keys on the server

recovery = storing "identity" info on the server, where identity means all the private keys: backup key, cross-signing key, dehydrated device key, etc.

richvdh: LGTM. I think we can resolve this thread.

richvdh (approved):

KitsuneRal (changes requested):

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
+# MSC4161: Crypto terminology for non-technical users
```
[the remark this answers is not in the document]
andybalaam: Lots to think about here but I wanted to say straight away: thank you - you raise lots of very important points and I think I agree with almost everything you've said.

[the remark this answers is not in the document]
ajbura: >  I'm quite surprised by the session -> device one

It actually make more sense as "session" as terminology should not be exposed to general user. Session as UI term is more driven from developers use of "session tokens".

Generally explaining UI/UX reasoning is hard but let me break it down with screenshot.
Screenshot 2025-02-12 at 9 25 18 AM
whenever user login new client:
-  it should be considered as a login *entry* (not session) (marked in blue) in UI.
- At this point we can't say the login entry is done by *you*. 
- To verify it we should perform some check that the device that is used, is owned by *you*?
- Prompt the user to verify the device identity.

Overall login entries (session) are subset of devices, but verification involves proving ownership of devices to grant secure access.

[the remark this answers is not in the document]
ajbura: No, you are not wrong. Out of overall cohesive UI, individual login entry should not need to be called as *device*(or anything) as it has data associated with it, which makes itself explanatory.
But when do we call them as device? obviously when we are referencing individual entry in ui. like:
- "A new unverified *device* (login) appears: *xyz (macbook)*" here we should purposely call it a device because of security measures which involves *Device verification* (not session verification). Until we verify, it *is* a new device login in overall secure system.
- "Changing *device* display name" It is better to call it device as the display name contains info that help us associate login with specific device.
- "Deleting choosen *devices*" better to call device as login entries are selected by reading device display names.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Devices
+
+**Note: this section depends on [MSC4153 ("Exclude non-cross-signed devices")](https://github.com/matrix-org/matrix-spec-proposals/pull/4153), which specifies clients should avoid sending
+and receiving encryption info with devices that are not cross-signed by their
+owner ("insecure" devices in our terminology).** While MSC4153 remains unmerged,
+the parts of this section relating to insecure devices should be considered
+non-normative.
+
+Instances of a client are called 'devices' (not 'sessions'). Aligned with
```
richvdh, line 92: A question that arose in discussion with the SCT: do we actually *need* to specify "sessions" vs "devices"? Maybe this is something that we can leave up to individual clients, and users will figure it out?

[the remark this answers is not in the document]
dkasak: There's a thing that's been bugging me about this. Are "devices" and "sessions" even the same thing? MAS is primarily talking about server-side sessions which correspond to access tokens. But when we are talking about cryptography, we don't mean that but something which is exclusively client-side and corresponds to cryptographic key material.

Sure, there is often a correspondence between the two, but they are not the same thing. They are not even necessarily 1-to-1 given there are client instances which have no cryptography enabled.

So is the terminology problem partly caused by our stumbling over this difference without realising it?

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
turt2live: Having discussed this briefly with the SCT, the actual implementation requirements are more along the lines of requiring Element to implement this MSC. This is because of the client's large user base, allowing us to determine if the terms appeal to users in that kind of environment. 

Other clients can and should implement this MSC still, however. Those implementation would further add data to the terms proposed in this MSC, either favourably or otherwise :)

turt2live: I don't think this comment is helpful at this stage in the proposal, so resolving. 

We should continue trying to solicit opinions from other developers/teams, but appear to be getting limited responses so far.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+⚠️ Avoid mentioning "device keys" - a device is just secure or not.
+
+⚠️ Avoid "session" to mean device. Device better describes what most users
+encounter, and is more commonly used in other messaging apps.
+
+#### Logging out
+
+In contrast to some other services, **logging out** (or **signing out**) of a
+Matrix device is quite a significant operation: it means the encryption data on
+this device is lost, and if you log out of all devices you will need to use your
```
turt2live, line 132:
```suggestion
Matrix device is quite a significant operation: it means your encryption data on
the device is lost, and if you log out of all devices you will need to use your
```

nit: change of voice/ownership.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+When you verify a user they become **verified**. This means that you have
+cryptographic proof that no-one is listening in on your conversations. (You need
+this if you suspect someone in a room may be using a malicious homeserver.)
+
+In many contexts, most users are **not verified**: verification is a manual
+step ([scanning a QR code](https://spec.matrix.org/v1.12/client-server-api/#qr-codes) or [comparing emojis](https://spec.matrix.org/v1.12/client-server-api/#sas-method-emoji)). (In future, verification will
+probably become more common thanks to [MSC2882 Transitive Trust](https://github.com/matrix-org/matrix-spec-proposals/pull/2882) or something similar).
+When an unverified user resets their identity, we should warn
+the user, so they are aware of the change.
```
turt2live, line 155: it's not immediately clear which user we're warning here, and I don't have good suggestions for how to fix the sentence without adding more terminology. 

Examples of what I'm thinking of:

> When an unverified user/contact resets their identity, the user should be warned that this happened.

> When an unverified user resets their identity, all the users they verified should be warned that this happened.

> When an unverified user resets their identity, the observing user should be warned that this happened.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+## Proposal
+
+When communicating about cryptography with non-technical users, we propose using
+the following terms and concepts.
+
+When referring to concepts outlined in this document in their user interface,
+clients SHOULD use the language specified, except where their own users are
+known to understand different terms more easily. When making such exceptions,
+clients SHOULD document how they deviate from this document, and why.
```
turt2live, line 82:
```suggestion
When referring to concepts outlined in this document in their user interface,
clients SHOULD use the language specified in **bold**, except where their own users are
known to understand different terms more easily. When making such exceptions,
clients SHOULD document how they deviate from this document, and why.
```

editorial thing: somewhere the MSC needs to say that bolded terms are normative, and this seems like a fine place.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+> "Store message keys on the server."
+
+⚠️ Avoid saying "key" without a previous word saying what type of key it is.
+
+⚠️ Avoid using "room key". These keys are used to decrypt messages, not rooms.
+
+Note: this clashes with the term "message key" in the double ratchet. Since the
+double ratchet algorithm is for a very different audience, we think that this is
+not a problem.
```
turt2live, line 249:
```suggestion
double ratchet algorithm is for a very different audience, this is not considered
a blocking conflict.
```

editorial: assertive/proposal language

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Unable to decrypt
+
+When we have an encrypted message but no message key to decrypt it, we are
+unable to decrypt it.
+
+When we expect the key to arrive, we are **waiting for this message**.
+
+> "Waiting for this message" with a button: "learn more" that explains that the message key for
+this message has not yet been received, but that we expect it to
```
turt2live, line 259:
```suggestion
> this message has not yet been received, but that we expect it to
```

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+⚠️ Avoid saying "key" without a previous word saying what type of key it is.
+
+⚠️ Avoid using "room key". These keys are used to decrypt messages, not rooms.
+
+Note: this clashes with the term "message key" in the double ratchet. Since the
+double ratchet algorithm is for a very different audience, we think that this is
+not a problem.
+
+### Unable to decrypt
```
turt2live, line 251: For the MSC to consider: Can we say "unable to decrypt" if there was a technical error during decryption? ie: the message key was received, but it's not quite in the right format, causing an exception to be raised. Or, the message key was received but the client's IndexedDB exploded.

Suggestion: Clients should distinctly identify technical errors from lack of received message key. For example, by saying "something went wrong with this message" instead of "unable to decrypt". "Decrypt" is considered jargon.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+operator is not able to access them and read your messages.
+
+In the spec, key storage is referred to as
+[server-side key backup](https://spec.matrix.org/v1.13/client-server-api/#server-side-key-backups).
+
+> "Allow key storage"
+
+> "Key storage holds the keys that allow you to read your message history."
+
+> "Message history is unavailable because key storage is disabled."
```
turt2live, line 295: For the MSC to consider: it may be alarming to users that keys aren't stored by default? Adding the context that they are stored *securely* server-side may help ensure that people make an informed decision to turn this on or off.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+
+A **recovery key** (or **recovery code**) is a way of re-establishing your
+identity if you lose all your devices. This in turn allows you to access key
+storage, and therefore see message history. If you re-establish your identity
+instead of resetting it, other users won't see "Alice's identity has changed"
+messages, and you will be able to read your message history, even if you logged
+out everywhere or lost your devices.
+
+A **recovery passphrase** is an easier-to-remember way of accessing the recovery
+key and has the same purpose as the recovery key.
```
turt2live, line 332: For the MSC to consider: a third term for the same thing is confusing to me as a reader. It's not clear from the description here what "accessing the recovery key" means. What is the thing I put in my password manager?

turt2live (approved):
Thanks for your continued work on this MSC! It's shaping up to be extremely well written, and may be possible to include verbatim in the spec itself (when we get there). 

This review is a bit large because it contains primarily editorial considerations, but within there is a suggestion to skip the device vs session debate by allowing both, and preferring device. Ultimately, compared to the other terms in this proposal, I don't feel the difference in terminology would measurably add confusion for users. Provided of course that someone doesn't start calling them "apps" or "connections".

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
andybalaam: Updated in 00993d93 to allow either device or session. It's clear that there is no consensus in the community about which word to use, so we take the slight hit of having 2 words, but we avoid clients having to diverge from this proposal.

As for @dkasak 's comments: I think there might be something very helpful there, but I think it's too low-level for this proposal.

andybalaam: I updated this MSC here 00993d93 with similar language to the above, with a slightly more even-handed wording. Given the strength of feeling on this issue I don't want to describe 'sessions' as a second choice, but rather an alternative.

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
andybalaam: Hopefully 00993d93 gives the necessary flexibility to use either word as needed.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+We hope that this MSC will:
+
+* Cause small changes in the spec (as described in the previous paragraph), and
+* Become an appendix in the spec, with a description that makes clear that the
+  intended audience is different from most of the spec, meaning different words
+  are used from the main spec body.
+
+Clients may, of course, choose to use different language. Some clients may
+deliberately choose to use more technical language, to suit the profiles of
+their users. This document is aimed at clients targeting non-technical users.
```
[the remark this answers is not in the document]
andybalaam: I agree, but I want to make sure the scope of this document is as clear as possible, to avoid unnecessary debate about issues that are more technical than the issues addressed here.

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
dkasak: @andybalaam Do you have any suggestions where else we can take it? It's a bit of a fear of mine that the wording used in this MSC, should it get accepted, will interfere with the established technical wording in the spec.  Both this MSC and the spec talk about "devices", but they really mean (related but ultimately) different things, don't they?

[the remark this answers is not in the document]
andybalaam: I actually think devices here are quite close to what the spec means when it says devices. Where do you see a difference?

In general, I'm not sure where we should take your comments. If you want to propose a change to the wording used in the main spec, I suppose a separate MSC would be the best way.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,451 @@
...
+in different ways and understand different metaphors. This proposal hopes to
+nudge client developers towards consistency, but never at the cost of their
+unique relationship with their users.
+
+## Outcomes
+
+We hope that Matrix client developers will like the terms and wording we
+provide, and adapt their user interfaces and documentation to use them. (If this
+MSC is accepted, Element will use it as a reference for English wording in its
+clients.)
```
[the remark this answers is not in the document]
andybalaam: Updated in 40c268d3

proposals/4161-crypto-terminology.md
andybalaam: Good point - added in cc83a417

proposals/4161-crypto-terminology.md
andybalaam: I'm not yet sure what you are suggesting. Possibilities:

* Because it may be alarming that keys aren't stored by default: we should rename key storage to "server key storage" or similar to make it easier to understand that keys are already stored locally
* Because we want to encourage users that turning on key storage is safe: we should refer to key storage as "secure"

proposals/4161-crypto-terminology.md
andybalaam: Thanks - I attempted to address this here: 30acaacb

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
andybalaam: > Speaking of this MSC: I would appreciate having more external references
> ...
> Or consider "identity" - I don't know if it's good or bad because I don't see the research behind the choice...

This is a very good point. I have attempted to provide some external references in 38e2f257 - if you like the general style of this I can try to do it for some other words.

In some cases I think what we are doing in the MSC is just standardising words that are already in use, so maybe the references should rather be to existing Matrix clients - what do you think?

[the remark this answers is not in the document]
andybalaam: > I would suggest that we actually spec the glossary, leaving specific phrases to a lighter "implementation guide" kind of a document.

We could definitely do a split like this if there is support for it. So far I have thought of what we are doing as a glossary, with the words in bold as normative, and the additional words used for explanation of the glossary, not necessarily an implementation guide. So I am not yet personally convinced that we need a split, but maybe we could make it clearer that the examples are for explanation purposes?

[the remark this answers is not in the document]
andybalaam: > I'm not sure we at SCT have experience and expertise on the subject. Being one of those covering the clients domain area, I may have to go for some research and self-education before I can responsibly make decisions on this. And/or we might have to call someone in who has appropriate background. Fortunately, Matrix is used by some UI-conscious communities (GNOME, in particular) so we might not need to search too far away.

I'd certainly welcome input from people with other expertise - this is so far mainly constructed from observing words already in use in multiple Matrix clients, and with expert input from product management and design professionals within Element. We have made adjustments based on input from other clients and community members too, but have not managed to get a huge amount of engagement.

However, I'd be keen not to add an impossible barrier to this MSC. Even if it needs improvements later (which I agree could be painful) I think it can improve the lives of users soon after it is merged, and if it is never merged we will never see those improvements.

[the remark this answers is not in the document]
andybalaam: >     1. Mainly to SCT: can we make sure this MSC does not end up a single shot and establish a way to further contribute in the same direction?

I believe https://github.com/matrix-org/matrix-spec-proposals/pull/4270 is an attempt to address this comment by establishing a context within which this MSC can be applied.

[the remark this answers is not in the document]
andybalaam: > For example: next to others, I'm not a fan of "devices", I really believe it's a misnomer we borrowed from centralised mobile-first platforms. I can still grudgingly accept it if we couldn't find a better alternative. But the only discussed alternative was "session", which I agree is (even) weaker. I was surprised that "clients" (taken from e-mail), or "applications/apps" (desktop and mobile environments) are not even namechecked.

Thanks, I added more alternatives in fa893e15 . I personally don't think there are better words than "device", but we can try! btw I'm not sure how "devices" is related to being centralised? (I do see how it's from mobile-first platforms, and I think we should embrace that a little since it is how so many people engage with messaging.)

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
dkasak: I think my complaint is primarily that the existing treatment of devices in the spec is ambiguous and somewhat confused. This MSC purports to improve the terminology but is not actually disentangling this ambiguity but rather doubling down on it.

In particular, we do not properly disentangle between:

1. Physical devices or client installations.
2. Cryptographic devices, i.e. the data structures representing the root of trust and authenticity of a given installation via cryptographic means. This roughly corresponds to the `DeviceKeys` struct in the spec.
3. Server-side or MAS-side sessions.

There are some loose business logic relations between these concepts but they are *distinct* concepts. The current spec treats them somewhat interchangeably, ambiguously referring to all of them as "devices". My argument is that each of these concepts should get its own term and that the relationships between them ought to be unambiguously described using those terms.

escix:
After watching Andy's conference (two passwords), these are my comments:

Most of the users don't care whether they are linking/combining or whatever they are doing other than login and get the messages. So, from my point of view, it is all about how am I getting the messages from the server. This is really the login flow management and wordings.

So the flow is:
1. the user login using their login credentials
2. they need to get their messages from the server, this can be done via verifying their new login using other existing logins or using the recovery key.

My suggestion is 'Login Verification'. This means for the user, they are just verifying the new login. They can use either other logged in devices or 'Access Code'.

'Access Code' is the 'recovery key'. From a user point of view, this means to access the messages from the server (i.e. recover the key) they need the 'Access Code'.

For the user verification, we can also use 'Contact Verification'.

Summary:

1. Login Credentials
2. Login Verification (using other logins or Access Code)
3. Access Code
4. Contact Verification

And it is much easier to tell others 'Mate, if you lose your access code, then you will lose your access to the messages 😄'

image



One more thought:

What if the verification from other logins is disabled altogether. Alternatively, if a verified login is available, is it possible to get the 'Access Code' from that login and use it to verify the new login?

This maybe a security issue, but ~~if the Access Code can be retrieved by entering the login credentials then~~ I can't see this being any different from emoji verification.

EDIT 20251104: This also removes the confusion between devices/sessions, as the requirement is to verify each and every 'Login'.

andybalaam:
After some informal user testing and feedback from the 2025 Matrix conference, I have pushed two significant changes today.

* "Verified" devices has been changed to "Confirmed" (and a note added about "linking" devices when we are simultaneously doing login and confirmation)
* "Identity" has been changed to "Digital Identity".

Feedback welcome, but I think this now represents my best ideas about what we should do.

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,464 @@
...
+Instances of a client are called **devices** or alternatively **sessions**.
+Devices that have not been cross-signed by the user who owns them are
+**unconfirmed**. It is important that devices are **confirmed**, to prevent
+security problems like a compromised server creating fake devices that
+can impersonate users (see
+[MSC4153](https://github.com/matrix-org/matrix-spec-proposals/pull/4153)).
+
+> "This device is not confirmed. Please confirm it to continue."
+
+> "Confirm it's you" (when asking to confirm a device during login)
```
turt2live, line 94: For clarity: If the [device](https://spec.matrix.org/v1.17/#devices) does *not* support encryption, is it confirmed or unconfirmed?

(this should be addressed in the MSC)

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,464 @@
...
+
+When you verify a user they become **verified**. This means that you have
+cryptographic proof that no-one is listening in on your conversations. (You need
+this if you suspect someone in a room may be using a malicious homeserver.)
+
+In many contexts, most users are **not verified**: verification is a manual
+step ([scanning a QR code](https://spec.matrix.org/v1.12/client-server-api/#qr-codes) or [comparing emojis](https://spec.matrix.org/v1.12/client-server-api/#sas-method-emoji)). (In future, verification will
+probably become more common thanks to [MSC2882 Transitive Trust](https://github.com/matrix-org/matrix-spec-proposals/pull/2882) or something similar).
+When an unverified contact resets their digital identity, the user should be warned that
+this happened.
```
turt2live, line 137:
```suggestion
When a verified contact resets their digital identity, the user should be warned that
this happened.
```

surely?

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,464 @@
...
+> key delivery delays.
+
+#### You don't have access to this message
+
+When the user does not have the message key for a permanent and well-understood
+reason, for example if it was sent before they joined the room, we say **you
+don't have access to this message**.
+
+> "You don't have access to this message" e.g. if it was sent before the user
+> entered the room, or the user does not have key storage set up.
```
turt2live, line 274: presumably this also includes when the sender refuses to provide the key (using a withheld error code)?

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
turt2live: I'm not seeing implementation attached to this MSC, so I've gone through to identify precise implementation requirements. Please provide screenshots and/or PR links which demonstrate each of these checkboxes. Ideally, implementation for each checkbox is provided for *each* of Element's clients (including Classic), though at a minimum *all* of these requirements *must* be met by at least a single Element client (preferably Web over X over Classic).

In short, though the MSC states that clients *SHOULD* use the terminology of the MSC, the implementation requirements upon Element are a *MUST* (at least while this proposal is pre-acceptance - Element can return to a *SHOULD* after successful FCP).

Implementation evidence required:

* [ ] Element using "devices" or "sessions" to refer to what the spec calls devices.
* [ ] Element using the word "unconfirmed" for devices which aren't cross-signed.
* [ ] Element using the term "linking" to refer to logging in & confirming a device concurrently.
* [ ] Element using "log out" or "sign out".
* [ ] Element shows other users as "verified" and "not verified".
* [ ] Element shows warnings that digital identities have been reset.
  * If the MSC intended to show this when *verified* contacts reset their identity, then those users need to produce such warnings. Alternatively, *unverified* users need to produce the warnings. If *both* are intended, then Element needs to do both as well.
  * [ ] Element offers to "withdraw verification" or to "re-verify"
  * [ ] When for a verified contact, Element's warning cannot be dismissed and states that the other user's "digital identity has been reset".
* [ ] Element does not use the term "cross-signing" in its UI.
* [ ] Element does not use the term "Trust On First Use (TOFU)" in its UI.
* [ ] Element does not use the term "cryptographic identity" in its UI.
* [ ] Element uses "digital identity" to refer to a device's cross-signing keys/identity and key backup/storage.
* [ ] Element uses "message key" to talk about keys required to decrypt a message/event.
* [ ] Element uses "unable to decrypt" to describe the following scenarios:
  * [ ] "Waiting for this message" when message keys are expected but not delivered (yet, hopefully).
  * [ ] "You don't have access to this message" when keys are *not* expected and *not* delivered (ever, hopefully).
  * [ ] When some error occurred. Exact wording is not defined by the MSC, though "unable to decrypt" is to be used.
* [ ] Element uses "message history" to describe decryptable events.
* [ ] Element uses "key storage" (ideally in combination with "message history") to talk about key backup.
* [ ] Element uses "recovery" when a user loses access to all of their devices (or has zero encryption-capable devices).
  * [ ] Element uses "recovery key" or "recovery code" or "recovery passphrase" to talk about how to fix this.
  * [ ] If using "recovery key/code", Element is talking about a machine-generated key a user can upload or enter.
  * [ ] If using "recovery passphrase", Element is talking about a user-understandable phrase or password-like string to represent the machine-generated key.
  * [ ] When a user loses their recovery key, Element says they can "reset" it.
  * [ ] Element does not use "secret storage" or "quad S".
  * If Element uses tooltips to say "your recovery key may have been called a security key in the past" (or similar), this is fine.
* [ ] Element allows users to "change their recovery key" rather than reset it.
* [ ] Element distinguishes between resetting digital identity and changing recovery key.

turt2live (changes requested):
This MSC is currently in queue for proposed-FCP, though I feel there's a couple more steps that need to happen before it can confidently enter that state. The MSC checklist I'll add shortly after this review will cover the points in more detail.

The things I feel need addressing before this is ready for proposed-FCP are:

* **Resolution of outstanding comments (*prior to this review*).** These threads are either long, old, or both. They should be incorporated into the proposal text or marked as resolved in GitHub if already addressed. 
  * The exception is (always) the "implementation requirements" thread - that takes place of implementations in the description.
* **Evidence of implementation.** See my comment on the "implementation requirements" thread.

When these two points, and any other unchecked boxes from the checklist or concerns from other SCT members, are addressed, please re-raise this MSC for proposed-FCP in the [SCT Office](https://matrix.to/#/#sct-office). Questions regarding threads are best asked in those threads. Questions for the SCT which are outside of existing threads should be raised in the SCT Office (SCT members might not see individual/team pings).

Overall though, in case it's not totally clear from the small number of comments, this proposal is *extremely close* to proposed-FCP. I'm looking forward to it getting pushed to that state!

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4161-crypto-terminology.md
```diff
@@ -0,0 +1,464 @@
...
+⚠️ Avoid mentioning "device keys". While a device may have keys, the user
+is only concerned about whether it is secure or not.
+
+When a mechanism for simultaneously logging in and confirming a device
+is available (e.g.
+[MSC4108: QR code login](https://github.com/matrix-org/matrix-spec-proposals/pull/4108))
+then we talk about **linking** the device. After a device is linked, it is
+logged in and confirmed.
+
+> "Link this device by scanning the QR code"
```
CrazyNicc, line 108: Should we really introduce another term here? I feel like `Log-in and confirm this device by scanning the QR code` would be more fitting, and use already established terms in this document, although it is a bit longer. I wouldn't immediately know what `linking` a device would mean.

CrazyNicc (commented):
Just a minor suggestion for improving the clarity, and using one less term.

proposals/4161-crypto-terminology.md
tulir: Signal, WhatsApp and Telegram all use "linking" to refer to logging in with QR codes, so I'd expect users to know what it means even though it's not an established term in Matrix.

CrazyNicc: Alright, that makes sense

proposals/4161-crypto-terminology.md
turt2live: I'm also not sure what I'm suggesting - closing.

proposals/4161-crypto-terminology.md
turt2live: it's unconfirmed - accounts shouldn't be mixing encryption support and not-encryption support.

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
turt2live: This proposal is now sufficiently implemented in at least Element where deviations are either bugs or deliberate, so I don't think the above checklists are helpful. 

Empirically the terminology has been working, and judging by 👍 reactions on this proposal, people generally favour it.

Resolving this thread because it's long and scary (sorry).

proposals/4161-crypto-terminology.md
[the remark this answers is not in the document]
turt2live: @KitsuneRal has this thread been sufficiently addressed? If not, can we break it down whatever is outstanding to dedicated threads for discussion please?

turt2live:
This looks ready to go with some relatively minor editorial changes, and has been proven in practice rather than cited code implementations.

@mscbot fcp merge

PR #4162, opened 2024-07-02
MSC4162: One-Time Key Reset Endpoint

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/otk-reset/proposals/4162-otk-reset.md)

### DIFF
--- a/proposals/4162-otk-reset.md
+++ b/proposals/4162-otk-reset.md
@@ -39,7 +39,7 @@ In practice, OTKs desync for a myriad of reasons, some of which are outlined bel
  }
  ```
 - If `all` is `true`, all stored one-time keys for the calling device are deleted on the server, and `key_ids` is ignored.
-- If `key_ids` is not empty, every specified key ID is deleted from the database. An example key ID is `signed_curve25519:AAAAHQ`.
+- If `all` is `false`, every key specified in `key_ids` is deleted from the database. An example key ID is `signed_curve25519:AAAAHQ`.
 
 The server will then return the following response:
 ```js


proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ - Server databases may become corrupted due to hardware failure.
+ - Server databases may be rolled back (e.g during a bad upgrade process).
+
+#### Protocol issues
+ - Clients only store a limited set of OTKs. When they delete excess OTKs they have no mechanism to tell the server to delete those old OTKs.
+ - Multiple users could hit `/keys/claim` at the same time as a client is uploading OTKs, which creates race conditions which can cause the same key to be sent to multiple users. [[issue](https://github.com/matrix-org/matrix-spec/issues/1124)]
+
+### Proposal
+
+#### Reset endpoint
```
BillCarsonFr, line 30: There is a list of why OTKs could get desync, and an end-point to reset and re-resync.

But how technically do we detect that there is a desync? Unless I missed it I don't think it's noted here.

Would a new end-point that would allow to check that client/server are in sync would be helpfull?
Or will this be always detected to late?

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
+## MSC4162: One-Time Key Reset Endpoint
```
BillCarsonFr, line 1: nit: Looks like this MSC is covering more than just reset.

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+#### Claiming keys ordering
+In addition to these changes, the semantics around `/keys/claim` has been slightly modified. Currently, the semantics are simply:
+
+> Claims one-time keys for use in pre-key messages.
+
+This does not define a sort ordering (i.e _which_ key should be given out to the caller). This MSC proposes tweaking this wording to be:
+
+> Claims one-time keys for use in pre-key messages. The key claimed MUST be the lowest lexicographically sorted key ID for that algorithm. For example, given two key IDs of `AAAAAu` and `AAAAHg`, `AAAAAu` MUST be returned before `AAAAHg`.
```
BillCarsonFr, line 60: I think there have been discussion for the KeyId to be the public key and not an "increment" like now. Would it be possible to have conflicts here? or it's not important? What about just using a timestamp ?

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Rationale
+
+The primary motivation of this endpoint is to provide a way for a client and server to resync OTKs. To do this, a client would call `/keys/reset` with `all: true` to delete all OTKs on the server. Once this 200 OKs, the client is then safe to perform `/keys/upload` with a fresh set of OTKs. This functionality is particularly important for existing clients who currently have desynced OTKs.
```
BillCarsonFr, line 64: I wonder if we don't want something a bit different for fallback keys as they have a longer life time (option to not reset them?). As it may interfer with https://github.com/matrix-org/matrix-spec-proposals/pull/4081

BillCarsonFr (commented):
Some comments

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
+## MSC4162: One-Time Key Reset Endpoint
+
+
+### Context
+One-time keys (OTKs) consist of two parts: a public part which is stored on the server and a private part which is stored on the client device. One-time keys can be uploaded to the server with the `/keys/upload` endpoint. They can then be consumed via the `/keys/claim` endpoint. The client is kept informed of the total number of OTKs stored on the server via the `/sync` endpoint, with the JSON key `device_one_time_keys_count`. It is critical that the set of keys on the client and server remain in-sync, otherwise an encrypted channel between two devices cannot be established. "In-sync" is defined by the set of public keys on the server having their matching private keys on the client.
```
turt2live, line 5: Please wrap lines to ~100 characters

proposals/4162-otk-reset.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+
+OTKs can be "in-flight". This describes the scenario where a OTK has been claimed by another device via `/keys/claim`, but a message encrypted with this key has not been received yet. There are many valid reasons for why the message has not been received yet, but it is also a potential attack vector if the client is forced to remember all keys for all time.
+
+### Desync causes
+In practice, OTKs desync for a myriad of reasons, some of which are outlined below.
+
+#### Client issues
+
+ - If the request to `/keys/upload` fails, the client must retry (including over restarts) the _same set of keys_ as it is not possible to know if the server received the request and has stored them already. If this doesn't happen, the server may return HTTP 400 errors indicating this. [[issue](https://github.com/matrix-org/matrix-rust-sdk/issues/1415)]
```
turt2live, line 14: This can probably be fixed without a whole new endpoint?

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ - If the client runs across multiple processes, both processes can upload different sets of OTKs in response to some trigger event (e.g being notified that a OTK has been consumed via `/sync`). [[issue](https://github.com/matrix-org/matrix-rust-sdk/issues/3110)]
+ - Clients should only delete a OTK once the key has been successfully used. [[issue](https://github.com/matrix-org/matrix-rust-sdk/issues/1761)]
```
turt2live, line 16: These two feel like implementation bugs rather than spec concerns.

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ - Client databases may become corrupted due to hardware failure.
+ - Client databases may be rolled back (e.g full phone backups).
+
+#### Server issues
+ - Server databases may become corrupted due to hardware failure.
+ - Server databases may be rolled back (e.g during a bad upgrade process).
+
+#### Protocol issues
+ - Clients only store a limited set of OTKs. When they delete excess OTKs they have no mechanism to tell the server to delete those old OTKs.
```
turt2live, line 25: Why are they deleting them? If it's just good hygiene, do we add expiry timestamps to the `/upload` endpoint instead?

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ - Multiple users could hit `/keys/claim` at the same time as a client is uploading OTKs, which creates race conditions which can cause the same key to be sent to multiple users. [[issue](https://github.com/matrix-org/matrix-spec/issues/1124)]
```
turt2live, line 26: This feels like an implementation issue, though the spec may need to be clarified how the locking works.

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Proposal
```
turt2live, line 28: The problem feels under-described: why is a reset required?

turt2live (commented):
Early review while I'm here triaging - I haven't reviewed the whole thing yet.

proposals/4162-otk-reset.md
kegsay: > The primary motivation of this endpoint is to provide a way for a client and server to resync OTKs. [...] This functionality is particularly important for existing clients who currently have desynced OTKs.

There's existing devices who have desynced OTKs. This is guaranteed to cause UTDs when those keys get used. It would be nice to fix that.

proposals/4162-otk-reset.md
kegsay: Clients delete them to not be forced into storing an unbounded set of OTKs. Clients try to keep to a number (typically 50 or 100) and when the server says the count is < this the client uploads more. But it can't delete any key as there may be in-flight requests. Hence, you can cause clients to use an unbounded amount of disk space by repeatedly telling the client that the OTK count is 0. This is bad. JS SDK and Rust SDK have maximum bounds to the number of OTKs it will store, currently this is set to [5000](https://github.com/matrix-org/vodozemac/blob/main/src/olm/account/one_time_keys.rs#L45).

proposals/4162-otk-reset.md
kegsay: There needs to be a way to sort the keys and perhaps more importantly, to generate new keys > old keys. Base 64 public keys don't allow for this, so that would make sorting impossible. We'd probably need to use timestamps.

proposals/4162-otk-reset.md
kegsay: We can't detect desyncs because the client and server are not continually checking (and they _do_ need to continually check, as DB corruption / rollbacks can happen at arbitrary points in time). Only `/keys/upload` does some basic sync checks by ensuring that if the client says key ID AAAAA is bytes X that it remains the case on the next upload (assuming the key hasn't been used).

proposals/4162-otk-reset.md
kegsay: Fallback keys are not one-time keys, so wouldn't be reset here.

proposals/4162-otk-reset.md
kegsay: Yes, but it outlines a common failure mode which causes OTKs to desync.

proposals/4162-otk-reset.md
kegsay: Yes, but it outlines a common failure mode which causes OTKs to desync.

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ ```js
+ {
+    "all": true|false
+    "key_ids": [
+        "<algorithm>:<key_id>", ...
+    ]
+ }
+ ```
+- If `all` is `true`, all stored one-time keys for the calling device are deleted on the server, and `key_ids` is ignored.
+- If `key_ids` is not empty, every specified key ID is deleted from the database. An example key ID is `signed_curve25519:AAAAHQ`.
```
andybalaam, line 42:
```suggestion
- If `all` is `false`, every key specified in `key_ids` is deleted from the database. An example key ID is `signed_curve25519:AAAAHQ`.
```

proposals/4162-otk-reset.md
richvdh, line 14: I don't really follow this. Are you saying "clients might forget to re-upload keys if they do not get a successful response"? Why would that cause a 400?

richvdh: It would probably be helpful to break this down into "implementation bugs which could theoretically be fixed" vs "real protocol problems" (such as backup rollback)

proposals/4162-otk-reset.md
richvdh, line 16:
```suggestion
 - Clients may incorrectly delete the private part of an OTK before it is successfully used. [[issue](https://github.com/matrix-org/matrix-rust-sdk/issues/1761)]
```

proposals/4162-otk-reset.md
richvdh: +1 for timestamps here.

proposals/4162-otk-reset.md
```diff
@@ -0,0 +1,90 @@
...
+ ```js
+ {
+    "all": true|false
+    "key_ids": [
+        "<algorithm>:<key_id>", ...
+    ]
+ }
+ ```
+- If `all` is `true`, all stored one-time keys for the calling device are deleted on the server, and `key_ids` is ignored.
+- If `all` is `false`, every key specified in `key_ids` is deleted from the database. An example key ID is `signed_curve25519:AAAAHQ`.
```
richvdh, line 42: This feels like two separate endpoints to me. Indeed maybe we should make the selective-delete endpoint a separate MSC: it's quite a different usecase

proposals/4162-otk-reset.md
richvdh: As I see it, there are basically three usecases for a full reset:

 * We know we have bugs in clients; we should fix those, but in the meantime we can improve resilience for selected deployments by implementing a healing mechanism.
 * We know we have bugs in clients; we will fix those in time, but that doesn't help people who are already broken. We will need a healing mechanism for those users.
 * There are some problems (in particular, client or server rollback from backup) that aren't really bugs, but will result in getting out of sync. Again, we need a healing mechanism.

Our best idea for resolving all three situations is to blow away all the OTKs on the server and start again.

It would be good if the MSC spelled this out.

proposals/4162-otk-reset.md
richvdh, line 5: > It is critical that the set of keys on the client and server remain in-sync

I feel like this is misleading. What we really mean is "the set of keys on the server is a *subset* of those on the client". It's not a problem if the client has extra keys (generally, we assume they've been claimed but not used, but in any case they are harmless, modulo the fact they take up slots in the client).

PR #4164, opened 2024-07-06
MSC4164: Leave all rooms on deactivation

Kladki:
While most servers already do this, I think it would be good to actually recommend this in the spec directly, so that future servers know to implement this as well.

[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/leave-all-rooms-on-deactivation/proposals/4164-leave-all-rooms-on-deactivation.md)

Implementations:
- Synapse: https://github.com/matrix-org/synapse/pull/3201
- Dendrite: https://github.com/matrix-org/dendrite/pull/2545
- Conduit: https://gitlab.com/famedly/conduit/-/commit/b81939841b272206991041ce7643a1c572590db3

### DIFF
--- a/proposals/4164-leave-all-rooms-on-deactivation.md
+++ b/proposals/4164-leave-all-rooms-on-deactivation.md
@@ -1,9 +1,5 @@
 # MSC4164: Leave all rooms on deactivation
 
-There can never be enough templates in the world, and MSCs shouldn't be any different. The level
-of detail expected of proposals can be unclear - this is what this example proposal (which doubles
-as a template itself) aims to resolve.
-
 When an account is
 [deactivated](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3accountdeactivate),
 it removes all ability for the user to login again. However, as the account is not removed from rooms


proposals/4164-leave-all-rooms-on-deactivation.md
```diff
@@ -0,0 +1,36 @@
+# MSC4164: Leave all rooms on deactivation
+
+There can never be enough templates in the world, and MSCs shouldn't be any different. The level
+of detail expected of proposals can be unclear - this is what this example proposal (which doubles
+as a template itself) aims to resolve.
```
Johennes, line 5:
```suggestion

```

proposals/4164-leave-all-rooms-on-deactivation.md
```diff
@@ -0,0 +1,36 @@
...
+
+When an account is
+[deactivated](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3accountdeactivate),
+it removes all ability for the user to login again. However, as the account is not removed from rooms
+it is in, this causes the server to still be sent events for rooms that account is in, even if all
+of the accounts in that room are deactivated, and hence no user is able to see them.
+
+## Proposal
+
+Servers SHOULD make the account leave all rooms it is currently in when the account is deactivated.
```
Johennes, line 15: I'm curious if this could even be a MUST potentially?

proposals/4164-leave-all-rooms-on-deactivation.md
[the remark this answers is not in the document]
turt2live: Implementations:
- Synapse: https://github.com/element-hq/synapse/commit/bf98fa0864fe62dbf1303acc778912d69fd4a738
- Dendrite: https://github.com/matrix-org/dendrite/pull/2545
- Conduit: https://gitlab.com/famedly/conduit/-/commit/b81939841b272206991041ce7643a1c572590db3

proposals/4164-leave-all-rooms-on-deactivation.md
clokep: If we can't come up with a usecase than I suspect it makes sense to be strict.

The main use-case I can think of is stranding a room without an admin. E.g. you might not want to remove the user from the room if they're the user with the highest power level. (See also your MSC4165.) By leaving the user in the room the server admin could "recover" the room.

Half-Shot: I think this leaves the deactivation in a bit of an unknown state and makes the whole thing a bit sticky. I'd be minded to instead have your client, assuming it's got a full set of powerlevels across rooms, to inform *you* if you're about to leave some rooms stranded.

That way, you can take action in advance of the deactivation to fix the rooms...or ignore it and deactivate anyway. Leaving a room without an admin isn't a problem if the user consents to doing so, imo.

PR #4165, opened 2024-07-06
MSC4165: Remove own power level on deactivation

Kladki:
[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/remove-own-power-level-on-deactivation/proposals/4165-remove-own-power-level-on-deactivation.md)

### DIFF
--- a/proposals/4165-remove-own-power-level-on-deactivation.md
+++ b/proposals/4165-remove-own-power-level-on-deactivation.md
@@ -27,8 +27,8 @@ that the account cannot then be moderated by the server administrator, unlike no
 ## Alternatives
 
 This same process can instead occur when rooms are left, which also fixes this issue for rooms which the
-account is not in at the time of deletion. However, leaving a room is not a perminant action, and the user
-may want to re-join, whereas account deactivation is perminant, so it may not make sense to do this process
+account is not in at the time of deletion. However, leaving a room is not a permanent action, and the user
+may want to re-join, whereas account deactivation is permanent, so it may not make sense to do this process
 when the room is left instead.
 
 ## Security considerations


proposals/4165-remove-own-power-level-on-deactivation.md
tulir, line 1: Implementation requirements:

* [ ] Server

proposals/4165-remove-own-power-level-on-deactivation.md
```diff
@@ -0,0 +1,44 @@
...
+This adds additional complexity to the account deletion process, but as this has the potential to mitigate
+the effects of certain attacks, it is deemed worth it.
+
+This also can cause issues where the deactivated account is the only account able to moderate a room, meaning
+that the account cannot then be moderated by the server administrator, unlike now.
+
+## Alternatives
+
+This same process can instead occur when rooms are left, which also fixes this issue for rooms which the
+account is not in at the time of deletion. However, leaving a room is not a perminant action, and the user
```
Johennes, line 30:
```suggestion
account is not in at the time of deletion. However, leaving a room is not a permanent action, and the user
```

proposals/4165-remove-own-power-level-on-deactivation.md
```diff
@@ -0,0 +1,44 @@
...
+may want to re-join, whereas account deactivation is perminant, so it may not make sense to do this process
```
Johennes, line 31:
```suggestion
may want to re-join, whereas account deactivation is permanent, so it may not make sense to do this process
```

Johennes, line 31:
```suggestion
account is not in at the time of deletion. However, leaving a room is not a permanent action, and the user
may want to re-join, whereas account deactivation is permanent, so it may not make sense to do this process
```

proposals/4165-remove-own-power-level-on-deactivation.md
girlbossceo: conduwuit implementation: https://github.com/girlbossceo/conduwuit/pull/571/commits/e9e5fe2176e8be611db91bbd38e3f8d4e92a3e3b

tcpipuk: I've tested the conduwuit implementation working - I was able to deactivate my account on fost.re when migrating to tomfos.tr and it correctly dropped its own powerlevels in all rooms it was able to do so. This is a great feature to have, and saved a lot of manual changes! Thanks so much! 🎉

proposals/4165-remove-own-power-level-on-deactivation.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Proposal
+
+When an account is deactivated, the server SHOULD do the following in all rooms the account is currently in:
+- If there is a `m.room.power_levels` state, and in there is a key for the current user inside the `users`
+field:
+  - If the auth rules for the room allow doing so, send a new `m.room.power_levels` event identical to the
+  current state, with the exception of the field for the current user inside `users` removed.
+
+## Potential issues
```
girlbossceo, line 19: I'm not sure how much of a problem it is in practice, but currently the MSC leaves it open to design that user DMs will also drop power levels.

I didn't consider this factor in the conduwuit implementation until after real-world usage of it on transfem.dev, and because conduwuit is very fast it's not really a problem. But I'm not sure if anyone has any input on whether we should be filtering out DMs or not.

The only situation I can think of where it could be relevant is if the account gets reactivated, and instead of making a new DM the other side invites the reactivated account to the same room. But even then power level 100 isn't required for a DM to function normally.

Either way, just something I noticed after implementing and using this. Maybe it would make Synapse slightly happier (load-wise) if it filters out DMs.

PR #4167, opened 2024-07-06
MSC4167: Copy bans on room upgrade

Kladki:
[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/copy-bans-on-room-upgrade/proposals/4167-copy-bans-on-room-upgrade.md)

Implementations:
- Synapse: https://github.com/matrix-org/synapse/pull/4642
- Dendrite: https://github.com/matrix-org/dendrite/pull/2307

### DIFF
--- a/proposals/4167-copy-bans-on-room-upgrade.md
+++ b/proposals/4167-copy-bans-on-room-upgrade.md
@@ -7,6 +7,11 @@ to the new room due to technical limitations of servers not being able to impers
 homeservers". While this is most notably true for `join` membership events, this is not true for `ban`
 membership, which users likely want copied over to the new room.
 
+This behavior [has been part of Synapse](https://github.com/matrix-org/synapse/pull/4642) since only a
+few months after [support for room upgrades was added](https://github.com/matrix-org/synapse/pull/4091),
+with Dendrite doing this
+[ever since they had support for room upgrades](https://github.com/matrix-org/dendrite/pull/2307).
+
 ## Proposal
 
 When a room upgrade is performed, servers SHOULD copy membership states with the `ban` `membership` to


proposals/4167-copy-bans-on-room-upgrade.md
```diff
@@ -0,0 +1,35 @@
...
+## Proposal
+
+When a room upgrade is performed, servers SHOULD copy membership states with the `ban` `membership` to
+the new room, to ensure that users banned in the old room are still banned in the new room.
+
+## Potential issues
+
+In rooms with many banned users, it may take a while to perform the upgrade, as all the bans will need
+to be performed. However, not doing this would allow previously banned users to access the room, which
+is why this is deemed as worth it.
```
TheArcaneBrony, line 19: Additional potential issue: remote servers iirc are informed of the membership change, causing the banned user to potentially receive "You were banned from <room>: <reason>" prompts.

proposals/4167-copy-bans-on-room-upgrade.md
```diff
@@ -0,0 +1,35 @@
+# MSC4167: Copy bans on room upgrade
+
+When a [room upgrade](https://spec.matrix.org/v1.11/client-server-api/#room-upgrades) is performed,
+many state events are copied over to the new room, to minimize the amount of work the user has to do
+after the upgrade. However, the spec currently states that "Membership events should not be transferred
+to the new room due to technical limitations of servers not being able to impersonate people from other
+homeservers". While this is most notably true for `join` membership events, this is not true for `ban`
```
clokep, line 7: So synapse is not spec compliant here? It might be worth linking to Synapse directly here instead of in the par description, which might be lost.

It was only 5 years ago so sounds like maybe the spec was never updated to match behavior?

clokep: Ok, but somewhere it could be helpful as historical precedence for 2 homeservers to ignore the spec and do this already.

proposals/4167-copy-bans-on-room-upgrade.md
```diff
@@ -0,0 +1,35 @@
...
+membership, which users likely want copied over to the new room.
```
FSG-Cat, line 8: There is the argument to be made that users want the ability to control this behavior.

Why? Because if the room in question uses a Bot like Draupnir or Mjolnir they have no need to clone the bans potentially as they do all their bans via the bot. If all the bans are done via the bot why clutter the room state the room might argue.

proposals/4167-copy-bans-on-room-upgrade.md
turt2live, line 1: Implementation requirements:

* [ ] Server

turt2live: Implementations:
- Synapse: https://github.com/matrix-org/synapse/pull/4642
- Dendrite: https://github.com/matrix-org/dendrite/pull/2307

proposals/4167-copy-bans-on-room-upgrade.md
```diff
@@ -0,0 +1,40 @@
...
+few months after [support for room upgrades was added](https://github.com/matrix-org/synapse/pull/4091),
+with Dendrite doing this
+[ever since they had support for room upgrades](https://github.com/matrix-org/dendrite/pull/2307).
+
+## Proposal
+
+When a room upgrade is performed, servers SHOULD copy membership states with the `ban` `membership` to
+the new room, to ensure that users banned in the old room are still banned in the new room.
+
+## Potential issues
```
turt2live, line 20: There are potential safety concerns with this behaviour, fwiw. Namely:

* Banned users are informed that they're banned from the new room too, which leaks the room ID. (mentioned here: https://github.com/matrix-org/matrix-spec-proposals/pull/4167/files#r1667731522 )
* Starting the room off with a bunch of abuse-containing state events isn't great

proposals/4167-copy-bans-on-room-upgrade.md
```diff
@@ -0,0 +1,40 @@
+# MSC4167: Copy bans on room upgrade
```
turt2live, line 1: It's unclear to me if we want to be more prescriptive in the room upgrade process: https://github.com/matrix-org/matrix-spec/issues/1906 exists for discussion.

proposals/4167-copy-bans-on-room-upgrade.md
Gnuxie: if you have `@alice:matrix.org` present within a room, and `@bob:matrix.org` becomes banned, then `@bob:matrix.org` will be force joined to the room and notified that they are banned because there is another user on their server in the room. Which is an issue because now bob is aware or has been reminded of the new room's existence.

Gnuxie: >Just to check, this only happens with users on the same server as the user upgrading the room, right? 

Haven't tested. 

>seems like strange behavior anyways, and I don't think it is something that should happen.

It does happen though until there is some work to clarify or make sure that this doesn't happen (if it is even possible to do so since it might be quite hard for a server to work out the extent that a user was participating in the room or not).

proposals/4167-copy-bans-on-room-upgrade.md
FSG-Cat: To not dig too deep into the intracies of #3909 and #4106 they have a problem in that being muted under them is counted as being a room member. Now if you copied the leave mute states that could work without issue. And you could in theory allow preemptively setting it via a trivial from my perspective change to #3909 that would be inherited by #4106

Current day mutes as in mutes before 3909 are coppied as they are baked into the powerlevels event.

proposals/4167-copy-bans-on-room-upgrade.md
FSG-Cat: Its good that Draupnir and Mjolnir allows you to avoid these issues if you are willing to accept the tradeoff that comes with reactive sanctions instead of preemptive sanctions.

PR #3385, opened 2021-09-07
[WIP] Bulk small improvements to room upgrades

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/upgrade-state/proposals/3385-refresh-room-upgrades.md)

PR #4168, opened 2024-07-06
MSC4168: Update `m.space.*` state on room upgrade

Kladki:
[Rendered](https://github.com/Kladki/matrix-spec-proposals/blob/update-space-state-on-room-upgrade/proposals/4168-update-space-state-on-room-upgrade.md)

Implementations:
- Synapse (partial, just copying `m.space.child` to the new room): https://github.com/element-hq/synapse/commit/63f28e4a0c22c8133a6da175c7986e389fc781e2
- Continuwuity: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/907

---

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4168#issuecomment-3544966952)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4168#issuecomment-3544963695)

### DIFF
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -38,6 +38,217 @@ the upgrade, regardless of its previous content. This is because the server's li
 previous `via` field may not have joined the upgraded room yet, and thus servers may not be able to
 join through them.
 
+### Examples
+Given the following initial rooms:
+```mirmaid
+flowchart TB
+    !project_space -->|child| !feedback_space
+    !project_space -->|child, order: "1"| !support
+    !project_space -->|child, suggested| !development
+
+    !feedback_space -->|parent| !project_space
+    !support -->|parent| !project_space
+
+    !feedback_space -->|child| !beta_users
+    !feedback_space -->|child| !suggestions
+
+    !beta_users -->|parent, canonical| !feedback_space
+    !suggestions -->|parent| !feedback_space
+```
+
+Here is the changed & new state after (assuming a user on `example.org` is upgrading the following rooms):
+### Upgrading `!project_space` to `!upgraded_project_space`
+
+#### `!upgraded_project_space`:
+```diff
++{
++    "type": "m.space.child",
++    "state_key": "!feedback_space",
++    "content": {
++        "via": ["example.org"]
++    }
++}
++{
++    "type": "m.space.child",
++    "state_key": "!support",
++    "order": "1",
++    "content": {
++        "via": ["example.org"]
++    }
++}
++{
++    "type": "m.space.child",
++    "state_key": "!development",
++    "suggested": true,
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+#### `!feedback_space` & `!support`:
+```diff
++{
++    "type": "m.space.parent",
++    "state_key": "!upgraded_project_space",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+Additionally, if the server implementation decides to remove the previous space's `via`:
+
+```diff
+ {
+     "type": "m.space.parent",
+     "state_key": "!project_space",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
+
+### Upgrading `!support` to `!upgraded_support`
+
+#### `!upgraded_support`:
+```diff
++{
++    "type": "m.space.parent",
++    "state_key": "!project_space",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+#### `!project_space`:
+```diff
++{
++    "type": "m.space.child",
++    "state_key": "!upgraded_support",
++    "order": "1",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+Additionally, if the server implementation decides to remove the previous room's `via`:
+
+```diff
+ {
+     "type": "m.space.child",
+     "state_key": "!upgraded_support",
+     "order": "1",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
+
+### Upgrading `!feedback_space` to `!upgraded_feedback_space`
+
+#### `!project_space`:
+```diff
++{
++    "type": "m.space.child",
++    "state_key": "!upgraded_feedback_space",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+Additionally, if the server implementation decides to remove the previous space's `via`:
+
+```diff
+ {
+     "type": "m.space.child",
+     "state_key": "!feedback_space",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
+
+#### `!upgraded_feedback_space`:
+```diff
++{
++    "type": "m.space.child",
++    "state_key": "!beta_users",
++    "content": {
++        "via": ["example.org"]
++    }
++}
++{
++    "type": "m.space.child",
++    "state_key": "!suggestions",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+#### `!beta_users`:
+```diff
++{
++    "type": "m.space.parent",
++    "state_key": "!upgraded_feedback_space",
++    "canonical": true,
++    "content": {
++        "via": ["example.org"]
++    }
++}
+ {
+     "type": "m.space.parent",
+     "state_key": "!feedback_space",
+-    "canonical": true,
+     "content": {
+         "via": ["example.org", "another.domain", "yet-another.domain"]
+     }
+ }
+```
+
+Additionally, if the server implementation decides to remove the previous space's `via`:
+
+```diff
+ {
+     "type": "m.space.parent",
+     "state_key": "!feedback_space",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
+
+#### `!suggestions`:
+```diff
++{
++    "type": "m.space.parent",
++    "state_key": "!upgraded_feedback_space",
++    "content": {
++        "via": ["example.org"]
++    }
++}
+```
+
+Additionally, if the server implementation decides to remove the previous space's `via`:
+
+```diff
+ {
+     "type": "m.space.parent",
+     "state_key": "!feedback_space",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
 
 ## Potential issues
 
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -11,6 +11,19 @@ In the following sentences, "relevant space state events" refer to `m.space.pare
 room types, in addition to `m.space.child` events for rooms with a type of
 [`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
 
+In addition, "updating relevant space state events" refers to the following:
+- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the
+  space event pointing to the room to be upgraded has `canonical` set to `true` in `content`,
+  homeservers SHOULD update that space event to set `canonical` to `false`, while setting it to
+  `true` in the space event pointing to the new room.
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
+event pointing to the previous room, to signal to clients that they shouldn't join the upgraded
+room.
+
 When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
 to the new room. The sender field in the new event should be set to the user who performed the
 upgrade.
@@ -20,9 +33,10 @@ In addition, servers SHOULD update relevant space state events in rooms that ref
 user (or any other user on the same homeserver, if the implementation decides to use any user it
 can) has the power to do so.
 
-The `via` field of each new state event SHOULD only contain the server name of the server,
-regardless of its previous content. This is because the server's listed in the previous `via` field
-may not have joined the upgraded room yet, and thus servers may not be able to join through them.
+The `via` field of each new state event SHOULD only contain the server name of the server doing
+the upgrade, regardless of its previous content. This is because the server's listed in the
+previous `via` field may not have joined the upgraded room yet, and thus servers may not be able to
+join through them.
 
 
 ## Potential issues
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -40,7 +40,7 @@ join through them.
 
 ### Examples
 Given the following initial rooms:
-```mirmaid
+```mermaid
 flowchart TB
     !project_space -->|child| !feedback_space
     !project_space -->|child, order: "1"| !support
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -7,14 +7,23 @@ be copied over, as well as these events being updated in other rooms.
 
 ## Proposal
 
-When a room upgrade is performed, servers SHOULD replicate `m.space.child` and `m.space.parent` state
-in the new room.
+In the following sentences, "relevant space state events" refer to `m.space.parent` events for all
+room types, in addition to `m.space.child` events for rooms with a type of
+[`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
 
-In addition, `m.space.child` and `m.space.parent` events in other rooms, where the caller of `/upgrade`
-has the permission to do so, servers SHOULD send a new state event for the old room, removing the `via` field
-from it, and send a new event of the same type, with `state_key` being the new room ID, and `via` just
-being the server of the sender user. Like when these events are copied into the upgraded room, the `sender`
-should be set to the user calling the endpoint.
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The sender field in the new event should be set to the user who performed the
+upgrade.
+
+In addition, servers SHOULD update relevant space state events in rooms that reference the old room
+(in their `state_key` field). Note that this will only be possible in rooms where the upgrading
+user (or any other user on the same homeserver, if the implementation decides to use any user it
+can) has the power to do so.
+
+The `via` field of each new state event SHOULD only contain the server name of the server,
+regardless of its previous content. This is because the server's listed in the previous `via` field
+may not have joined the upgraded room yet, and thus servers may not be able to join through them.
+
 
 ## Potential issues
 
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -43,7 +43,7 @@ Given the following initial rooms:
 ```mermaid
 flowchart TB
     !project_space -->|child| !feedback_space
-    !project_space -->|child, order: "1"| !support
+    !project_space -->|child, order: '1'| !support
     !project_space -->|child, suggested| !development
 
     !feedback_space -->|parent| !project_space
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -12,7 +12,7 @@ room types, in addition to `m.space.child` events for rooms with a type of
 [`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
 
 Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
-event pointing to the previous room, to signal to clients that they shouldn't join the upgraded
+event pointing to the previous room, to signal to clients that they shouldn't join the previous
 room.
 
 When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -33,10 +33,10 @@ In addition, servers SHOULD update relevant space state events in rooms that ref
 user (or any other user on the same homeserver, if the implementation decides to use any user it
 can) has the power to do so.
 
-The `via` field of each new state event SHOULD only contain the server name of the server doing
-the upgrade, regardless of its previous content. This is because the server's listed in the
-previous `via` field may not have joined the upgraded room yet, and thus servers may not be able to
-join through them.
+The `via` field of each new relevant space state event pointing to the upgraded room SHOULD only
+contain the server name of the server doing the upgrade, regardless of its previous content. This is
+because the server's listed in the previous `via` field may not have joined the upgraded room yet,
+and thus servers may not be able to join through them.
 
 ### Examples
 Given the following initial rooms:
@@ -65,7 +65,7 @@ Here is the changed & new state after (assuming a user on `example.org` is upgra
 +    "type": "m.space.child",
 +    "state_key": "!feedback_space",
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 +{
@@ -73,7 +73,7 @@ Here is the changed & new state after (assuming a user on `example.org` is upgra
 +    "state_key": "!support",
 +    "order": "1",
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 +{
@@ -81,7 +81,7 @@ Here is the changed & new state after (assuming a user on `example.org` is upgra
 +    "state_key": "!development",
 +    "suggested": true,
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 ```
@@ -118,7 +118,7 @@ Additionally, if the server implementation decides to remove the previous space'
 +    "type": "m.space.parent",
 +    "state_key": "!project_space",
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 ```
@@ -181,14 +181,14 @@ Additionally, if the server implementation decides to remove the previous space'
 +    "type": "m.space.child",
 +    "state_key": "!beta_users",
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 +{
 +    "type": "m.space.child",
 +    "state_key": "!suggestions",
 +    "content": {
-+        "via": ["example.org"]
++        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
 +}
 ```
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -11,15 +11,6 @@ In the following sentences, "relevant space state events" refer to `m.space.pare
 room types, in addition to `m.space.child` events for rooms with a type of
 [`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
 
-In addition, "updating relevant space state events" refers to the following:
-- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
-  should be sent, copying the `order` and `suggested` field from the `content` of the
-  `m.space.child` with `state_key` of the previous room ID.
-- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
-  ID should be sent. If the
-  space event pointing to the room to be upgraded has `canonical` set to `true` in `content`,
-  homeservers SHOULD update that space event to set `canonical` to `false`, while setting it to
-  `true` in the space event pointing to the new room.
 Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
 event pointing to the previous room, to signal to clients that they shouldn't join the upgraded
 room.
@@ -28,8 +19,17 @@ When a room upgrade is performed, servers SHOULD copy relevant space state event
 to the new room. The sender field in the new event should be set to the user who performed the
 upgrade.
 
-In addition, servers SHOULD update relevant space state events in rooms that reference the old room
-(in their `state_key` field). Note that this will only be possible in rooms where the upgrading
+In addition, servers SHOULD create new relevant space state events pointing to the upgraded room in
+rooms that reference the old room (in their `state_key` field). In practice, this means:
+- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the space event pointing to the room to be upgraded has `canonical` set to
+  `true` in `content`, homeservers SHOULD update that space event to set `canonical` to `false`,
+  while setting it to   `true` in the space event pointing to the new room.
+
+Note that this will only be possible in rooms where the upgrading
 user (or any other user on the same homeserver, if the implementation decides to use any user it
 can) has the power to do so.
 
@@ -140,7 +140,7 @@ Additionally, if the server implementation decides to remove the previous room's
 ```diff
  {
      "type": "m.space.child",
-     "state_key": "!upgraded_support",
+     "state_key": "!support",
      "order": "1",
 -    "content": {
 -        "via": ["example.org", "another.domain", "yet-another.domain"]
@@ -187,6 +187,13 @@ Additionally, if the server implementation decides to remove the previous space'
 +{
 +    "type": "m.space.child",
 +    "state_key": "!suggestions",
++    "content": {
++        "via": ["example.org", "another.domain", "yet-another.domain"]
++    }
++}
++{
++    "type": "m.space.parent",
++    "state_key": "!project_space",
 +    "content": {
 +        "via": ["example.org", "another.domain", "yet-another.domain"]
 +    }
--- a/proposals/4168-update-space-state-on-room-upgrade.md
+++ b/proposals/4168-update-space-state-on-room-upgrade.md
@@ -1,0 +1,46 @@
+# MSC4168: Update `m.space.*` state on room upgrade
+
+When a [room upgrade](https://spec.matrix.org/v1.11/client-server-api/#room-upgrades) is performed,
+many state events are copied over to the new room, to minimize the amount of work the user has to do
+after the upgrade. However, the spec doesn't currently recommend that `m.space.child` or `m.space.parent`
+be copied over, as well as these events being updated in other rooms.
+
+## Proposal
+
+When a room upgrade is performed, servers SHOULD replicate `m.space.child` and `m.space.parent` state
+in the new room.
+
+In addition, `m.space.child` and `m.space.parent` events in other rooms, where the caller of `/upgrade`
+has the permission to do so, servers SHOULD send a new state event for the old room, removing the `via` field
+from it, and send a new event of the same type, with `state_key` being the new room ID, and `via` just
+being the server of the sender user. Like when these events are copied into the upgraded room, the `sender`
+should be set to the user calling the endpoint.
+
+## Potential issues
+
+This proposal does not attempt to update `m.space.*` state in rooms where the user upgrading the room
+is not able to update them, as this not only likely requires something like
+[MSC4049](https://github.com/matrix-org/matrix-spec-proposals/pull/4049), but also adds other additional
+complications (e.g. which server should update the state?).
+
+Users may also not want to update events in other rooms. Hopefully this proposal can be used to determine
+if there is a use-case for not updating events in other rooms. If there is one, then a query parameter can
+be added to toggle this feature.
+
+## Alternatives
+
+As above, utilizing [MSC4049](https://github.com/matrix-org/matrix-spec-proposals/pull/4049) to update
+`m.space.*` events in other rooms could be an alternative, but due to additional complications in addition
+to requiring this MSC to be merged, it is deemed as not the best solution (for now).
+
+## Security considerations
+
+None considered.
+
+## Unstable prefix
+
+None required, as currently no new endpoints or fields are being proposed.
+
+## Dependencies
+
+None.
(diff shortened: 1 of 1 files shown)


proposals/4168-update-space-state-on-room-upgrade.md
tulir, line 1: Implementation requirements:

* [ ] Server

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+be copied over, as well as these events being updated in other rooms.
+
+## Proposal
+
+When a room upgrade is performed, servers SHOULD replicate `m.space.child` and `m.space.parent` state
+in the new room.
+
+In addition, `m.space.child` and `m.space.parent` events in other rooms, where the caller of `/upgrade`
+has the permission to do so, servers SHOULD send a new state event for the old room, removing the `via` field
+from it, and send a new event of the same type, with `state_key` being the new room ID, and `via` just
```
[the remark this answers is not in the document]
clokep: The alternative is that the old room then gets kicked "out" of the space, which I think is more confusing? Element clients don't show rooms which are upgraded if you're also in the upgraded room, I'm unsure how common that is in clients though.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
+# MSC4168: Update `m.space.*` state on room upgrade
```
turt2live, line 1: It's unclear to me if we want to be more prescriptive in the room upgrade process: https://github.com/matrix-org/matrix-spec/issues/1906 exists for discussion.

proposals/4168-update-space-state-on-room-upgrade.md
timedoutuk: continuwuity has a complete implementation of this MSC: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/907
updated second pass: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1807

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+When a [room upgrade](https://spec.matrix.org/v1.11/client-server-api/#room-upgrades) is performed,
+many state events are copied over to the new room, to minimize the amount of work the user has to do
+after the upgrade. However, the spec doesn't currently recommend that `m.space.child` or `m.space.parent`
+be copied over, as well as these events being updated in other rooms.
```
TheArcaneBrony, line 6: I don't agree that this is a good solution - infact, copying far too much stuff is exactly why I avoid the entire upgrade endpoint in all circumtances, and manage tombstones/creation myself... I would be more interested in making the entire copy events thing opt-in instead (and configurable via passing either a list of event IDs or type/key pairs to replicate, or a DSL to pick and transform items...)

Kladki: I think that would be better suited in another MSC (however, this (your) suggestion has it's own issues, e.g. especially the first bullet point [here](https://github.com/matrix-org/matrix-spec/issues/1906#issuecomment-3092536667), but all of them apply to some extent).

proposals/4168-update-space-state-on-room-upgrade.md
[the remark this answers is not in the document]
MadLittleMods: The other problem with doing this is it prevents future people from joining if it uses `restricted` (inherited) `m.room.join_rules`.

I also created a dedicated spec issue around this problem: https://github.com/matrix-org/matrix-spec/issues/2194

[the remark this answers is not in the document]
Kladki: This can be instead addressed by adding `m.room.tombstone` to the list of recommended stripped state events.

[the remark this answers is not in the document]
Kladki: Sorry, I meant adding room upgrade details to room summaries/space hierarchies, but the above is something that should probably be done anyways.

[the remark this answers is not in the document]
turt2live: I'd suggest leaving whether to kick the children out of the space to the implementation as an implementation detail for now. It's likely that Synapse will do so, but other servers might not. If we find we need a way to control that behaviour, we can always open another MSC.

turt2live (approved):

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge
@mscbot concern TravisR - Decide to either always update children of spaces, never update, or leave it as an implementation detail (my preferred choice)

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Proposal
+
+When a room upgrade is performed, servers SHOULD replicate `m.space.child` and `m.space.parent` state
+in the new room.
```
clokep: I think we only want this done for spaces, like Synapse does:

```suggestion
When a room upgrade is performed on a [room with a type of `m.space`](https://spec.matrix.org/v1.16/client-server-api/#types),
servers SHOULD replicate `m.space.child` and `m.space.parent` state to the new room.
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+
+In addition, `m.space.child` and `m.space.parent` events in other rooms, where the caller of `/upgrade`
```
clokep: I'm having trouble following this paragraph, can we break it down into a series of bullet points of the operations a server should perform?

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: If it's only done for `m.space` room types, it means that if you upgrade a regular room, the new room won't have the `m.space.parent` pointing to the space if the previous room had it.

clokep: But `m.space.child` has no meaning outside of a space, so it doesn't make sense to copy IMO.

Kladki: Ok, so we can just copy `m.space.child` only in spaces, but still copy `m.space.parent` in all room types.

clokep: Sounds good to me.

proposals/4168-update-space-state-on-room-upgrade.md
clokep: Synapse has a partial implementation (it copies `m.space.child`): https://github.com/element-hq/synapse/commit/63f28e4a0c22c8133a6da175c7986e389fc781e2

proposals/4168-update-space-state-on-room-upgrade.md
anoadragon453: Likewise. Perhaps:

> When a room upgrade is performed, servers SHOULD copy `m.space.child` and `m.space.parent` state events from the old room to the new room. The `sender` field in the new event should be set to the user who performed the upgrade.
>
> In addition, servers SHOULD update `m.space.child` and `m.space.parent` events in rooms that reference the old room (in their `state_key` field). Note that this will only be possible in rooms where the upgrading user has the power to do so.
> 
> The `via` field of each new state event SHOULD only contain the server name of the server, regardless of its previous content. This is because the server's listed in the previous `via` field may not have joined the upgraded room yet, and thus servers may not be able to join through them.

anoadragon453: Note that rooms can also be upgraded by sending an `m.room.tombstone` into the room. We shouldn't exclude copying over state events in this case, just because the user didn't use `/upgrade`.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+from it, and send a new event of the same type, with `state_key` being the new room ID, and `via` just
+being the server of the sender user. Like when these events are copied into the upgraded room, the `sender`
+should be set to the user calling the endpoint.
+
+## Potential issues
+
+This proposal does not attempt to update `m.space.*` state in rooms where the user upgrading the room
+is not able to update them, as this not only likely requires something like
+[MSC4049](https://github.com/matrix-org/matrix-spec-proposals/pull/4049), but also adds other additional
+complications (e.g. which server should update the state?).
```
anoadragon453, line 279: The homeserver *could* update those events if it has *any* user in the room that's able to send them. This might look odd to said user, but we may want to leave the possibility open for homeserver's to do anyways.

proposals/4168-update-space-state-on-room-upgrade.md
MadLittleMods: :+1: For more context on this point, there is a [recent case](https://github.com/element-hq/synapse/issues/19199) in Synapse where we don't handle this kind of thing.

clokep: I don't think that's an 'upgrade' necessarily though? I'm not sure servers should be doing special logic in that case.

You might send a tombstone to redirect to a different room or "close" it.

turt2live: We historically have not considered manually sending tombstones as "upgrades" in the protocol. The spec wants callers to perform upgrades using the endpoint, not by sending events manually.

MadLittleMods: There is some nuance here.

The whole problem is that from the remote homeservers perspective, it doesn't matter whether you've used the `/upgrade` endpoint or not. It just looks like a `m.room.tombstone` was put in the room either way.

In the [case I mentioned around push rules](https://github.com/element-hq/synapse/issues/19199) and where I think @anoadragon453's point is spawning from, I think it does make sense to copy push rules over whenever the homeserver notices the room was upgraded (`m.room.tombstone` and new room has a `predecessor`). Copying push rules is something that can only be done by the local homeserver for the user. And it doesn't make sense to handle this for remote rooms but not local rooms. This kind of discussion is probably why that clarification probably deserves it's own MSC.

(same with copying over local aliases and [room tags](https://spec.matrix.org/v1.13/client-server-api/#room-tagging)) -- Anything that the local homeserver has to do.

Whereas in this case, a local homeserver doesn't need to take any action about space state in the room when it notices a `m.room.tombstone`. All of the action is from the person doing the "upgrade" (or sending the state) and can manually transfer state as they see fit.

**In shorter words,** I don't think we should copy over `m.space.child` or `m.space.parent` when a `m.room.tombstone` event is sent.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Resolved.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Done.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Included this as an option now.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+
+In the following sentences, "relevant space state events" refer to `m.space.parent` events for all
+room types, in addition to `m.space.child` events for rooms with a type of
+[`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
+
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The sender field in the new event should be set to the user who performed the
+upgrade.
+
+In addition, servers SHOULD update relevant space state events in rooms that reference the old room
```
clokep: What does update mean here? The old reference embeds the old room ID in the state key and can’t really be updated.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,55 @@
...
+(in their `state_key` field). Note that this will only be possible in rooms where the upgrading
+user (or any other user on the same homeserver, if the implementation decides to use any user it
+can) has the power to do so.
+
+The `via` field of each new state event SHOULD only contain the server name of the server,
```
clokep:
```suggestion
The `via` field of each new state event SHOULD only contain the server name of the server doing the upgrade,
```

turt2live:
I don't actually know what thread this concern is attached to, so resolving:

@mscbot resolve TravisR - Decide to either always update children of spaces, never update, or leave it as an implementation detail (my preferred choice)

(if it is attached to a thread, I'm not sure it's so important to block FCP anyway)

uhoreg:
> I don't actually know what thread this concern is attached to, so resolving:
> 
> @mscbot resolve TravisR - Decide to either always update children of spaces, never update, or leave it as an implementation detail (my preferred choice)
> 
> (if it is attached to a thread, I'm not sure it's so important to block FCP anyway)

Looks like it would be attached to https://github.com/matrix-org/matrix-spec-proposals/pull/4168#discussion_r1667766217

clokep:
@mscbot concern Lack of details about updating references to replaced room

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Addressed.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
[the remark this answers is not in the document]
Kladki: Mentioned that servers MAY do this.

proposals/4168-update-space-state-on-room-upgrade.md
clokep: Does this need to be documented as a potential issue?

MadLittleMods: Not in this case.

Generally, this is something to keep in mind with proposals for room upgrade behaviors though. My reply above explains the nuance.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,69 @@
...
+
+In addition, "updating relevant space state events" refers to the following:
+- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the
+  space event pointing to the room to be upgraded has `canonical` set to `true` in `content`,
+  homeservers SHOULD update that space event to set `canonical` to `false`, while setting it to
+  `true` in the space event pointing to the new room.
```
clokep: For parent space events, wouldn't it just get copied to the new room with the same state key? Or is the room being upgraded the parent space?

This all sounds simple but gets complicated. Maybe we can have some examples?

Kladki: This is talking about space members (when space parents are being upgraded), since the space parent itself won't have any `m.space.parent` state (unless it is a member of a space itself). I will add examples though.

Kladki: Examples have been added, hopefully there will be no more confusion. :slightly_smiling_face:

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,280 @@
...
++    }
++}
+```
+
+Additionally, if the server implementation decides to remove the previous room's `via`:
+
+```diff
+ {
+     "type": "m.space.child",
+     "state_key": "!upgraded_support",
```
clokep: I think this is a typo?

```suggestion
     "state_key": "!support",
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,280 @@
...
+     "type": "m.space.child",
+     "state_key": "!feedback_space",
+-    "content": {
+-        "via": ["example.org", "another.domain", "yet-another.domain"]
+-    }
++    "content": {}
+ }
+```
+
+#### `!upgraded_feedback_space`:
```
clokep, line 192: I think this is missing the parent space?

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,280 @@
...
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the
+  space event pointing to the room to be upgraded has `canonical` set to `true` in `content`,
+  homeservers SHOULD update that space event to set `canonical` to `false`, while setting it to
+  `true` in the space event pointing to the new room.
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
+event pointing to the previous room, to signal to clients that they shouldn't join the upgraded
+room.
```
clokep: Typo, I think?

```suggestion
Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
event pointing to the previous room, to signal to clients that they shouldn't join the previous
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,280 @@
...
+
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The sender field in the new event should be set to the user who performed the
+upgrade.
+
+In addition, servers SHOULD update relevant space state events in rooms that reference the old room
+(in their `state_key` field). Note that this will only be possible in rooms where the upgrading
+user (or any other user on the same homeserver, if the implementation decides to use any user it
+can) has the power to do so.
```
clokep:
```suggestion
In addition, servers SHOULD create new relevant space state events in rooms that reference the old room
(in their `state_key` field). Note that this will only be possible in rooms where the upgrading
user (or any other user on the same homeserver, if the implementation decides to use any user it
can) has the power to do so.
```

I don't like the use of "update" here since a new event has to be created (since it has a different `state_key`).

clokep (commented):
Thanks for adding examples. I left a few comments which I think are all typos/copy&paste errors, if so, then I think this is good. 👍

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,46 @@
...
+This proposal does not attempt to update `m.space.*` state in rooms where the user upgrading the room
+is not able to update them, as this not only likely requires something like
+[MSC4049](https://github.com/matrix-org/matrix-spec-proposals/pull/4049), but also adds other additional
+complications (e.g. which server should update the state?).
+
+Users may also not want to update events in other rooms. Hopefully this proposal can be used to determine
+if there is a use-case for not updating events in other rooms. If there is one, then a query parameter can
+be added to toggle this feature.
+
+## Alternatives
```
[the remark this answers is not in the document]
clokep: @Kladki Please either close this thread as not relevant or incorporate it into the MSC.

proposals/4168-update-space-state-on-room-upgrade.md
clokep: @Kladki I think we need a conclusion to this thread in the MSC. The link in the MSC implies it happens for all room upgrades (including just a `m.room.tombstone`), if this only applies to the room upgrade endpoint then we should include that info.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,280 @@
...
+
+In addition, "updating relevant space state events" refers to the following:
+- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the
+  space event pointing to the room to be upgraded has `canonical` set to `true` in `content`,
+  homeservers SHOULD update that space event to set `canonical` to `false`, while setting it to
+  `true` in the space event pointing to the new room.
```
richvdh: I have to say, that setting out this (very complicated) definition here, and then referring back to it later, makes the proposal very hard to read. I'd suggest trying to make a single bullet list of actions that a server should carry out.

I can't approve an MSC that is this unclear.

proposals/4168-update-space-state-on-room-upgrade.md
richvdh: I don't think this is rendering the way you expect it to. Note that reviewers will be working with the rendered text, so please review the rendered text yourself to make sure it says what you want.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: My bad, fixed now.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: My bad, fixed now.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: From my side it looks like that'd just be duplication:

Image

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: [MSC3385](https://github.com/matrix-org/matrix-spec-proposals/pull/3385) introduces the concept of doing local changes for a remote upgrade, and mentions that Synapse already does this. So I think it's best for that WIP MSC to be expanded upon later than making it part of this MSC.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: I got rid of this definition, and moved the explanation to the point where that definition was previously referenced.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Fixed.

proposals/4168-update-space-state-on-room-upgrade.md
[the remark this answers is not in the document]
Kladki: See my response below [this comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4168#discussion_r2818428309) (can't link directly to mine because that's how reviews work apparently).

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: I see what you mean now, that `upgraded` should be `previous`. Fixed.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Proposal
+
+In the following sentences, "relevant space state events" refer to `m.space.parent` events for all
+room types, in addition to `m.space.child` events for rooms with a type of
+[`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
+
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
+event pointing to the previous room, to signal to clients that they shouldn't join the previous
+room.
```
richvdh: Additionally to what? What previous room? What event? When do servers do this?

In short: this paragraph feels like it's got lost. Please try and make sure that the document makes sense for someone reading it through to understand your proposal.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The sender field in the new event should be set to the user who performed the
```
richvdh:
```suggestion
to the new room. The `sender` field in the new event should be set to the user who performed the
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+upgrade.
+
+In addition, servers SHOULD create new relevant space state events pointing to the upgraded room in
+rooms that reference the old room (in their `state_key` field). In practice, this means:
+- For `m.space.child` events, a new `m.space.child` event with `state_key` set to the new room's ID
+  should be sent, copying the `order` and `suggested` field from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
```
richvdh:
```suggestion
- In rooms that reference the old room via `m.space.child` events (roughly speaking:
  spaces which are parents of the upgraded room), the upgrading server
  SHOULD send a new `m.space.child` event with `state_key` set to the new room's ID,
  copying the `order` and `suggested` fields from the `content` of the
  `m.space.child` with `state_key` of the previous room ID.
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+- For `m.space.parent` events, a new `m.space.parent` event with `state_key` set to the new room's
+  ID should be sent. If the space event pointing to the room to be upgraded has `canonical` set to
+  `true` in `content`, homeservers SHOULD update that space event to set `canonical` to `false`,
+  while setting it to   `true` in the space event pointing to the new room.
```
richvdh:
```suggestion
- In rooms that reference the old room via `m.space.parent` events (roughly speaking: child rooms
  of an upgraded space), the upgrading server SHOULD send a new `m.space.parent` event with 
  `state_key` set to the new room's ID. If the previous `m.space.parent` event has `canonical` set to
  `true` in `content`, homeservers SHOULD update the old state event to set `canonical` to `false`,
  while setting it to `true` in the newly-sent `m.space.parent` event.
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
+event pointing to the previous room, to signal to clients that they shouldn't join the previous
+room.
+
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The sender field in the new event should be set to the user who performed the
+upgrade.
+
+In addition, servers SHOULD create new relevant space state events pointing to the upgraded room in
+rooms that reference the old room (in their `state_key` field). In practice, this means:
```
richvdh:
```suggestion
In addition, servers SHOULD send new space state events pointing to the upgraded room in
rooms that reference the old room via `m.space.parent` or `m.space.child` events. In practice, this means:
```

proposals/4168-update-space-state-on-room-upgrade.md
richvdh: should the old `m.space.child` event be left in place? it sounds like it should, but we should be explicit.

proposals/4168-update-space-state-on-room-upgrade.md
richvdh: We have specified the `sender` of the new state events in the new room; what should the `sender` be for these events?

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+  `true` in `content`, homeservers SHOULD update that space event to set `canonical` to `false`,
+  while setting it to   `true` in the space event pointing to the new room.
+
+Note that this will only be possible in rooms where the upgrading
+user (or any other user on the same homeserver, if the implementation decides to use any user it
+can) has the power to do so.
+
+The `via` field of each new relevant space state event pointing to the upgraded room SHOULD only
+contain the server name of the server doing the upgrade, regardless of its previous content. This is
+because the server's listed in the previous `via` field may not have joined the upgraded room yet,
```
uhoreg: nit:

```suggestion
because the servers listed in the previous `via` field may not have joined the upgraded room yet,
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
+# MSC4168: Update `m.space.*` state on room upgrade
+
+When a [room upgrade](https://spec.matrix.org/v1.11/client-server-api/#room-upgrades) is performed,
+many state events are copied over to the new room, to minimize the amount of work the user has to do
+after the upgrade. However, the spec doesn't currently recommend that `m.space.child` or `m.space.parent`
+be copied over, as well as these events being updated in other rooms.
```
richvdh, line 6: You don't really explain why this is a problem, and the fact that this has not been implemented in Synapse suggests to me (without intending to throw shade on Continuwuity) that this is not exactly a highly-demanded change to the behaviour of the protocol.

Is there any evidence that this does, in fact, improve the user experience?

richvdh (changes requested):

richvdh:
> @mscbot concern Lack of details about updating references to replaced room

@clokep are you happy to resolve this?

richvdh:
@mscbot concern lack of evidence that this is a desirable change

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Proposal
+
+In the following sentences, "relevant space state events" refer to `m.space.parent` events for all
+room types, in addition to `m.space.child` events for rooms with a type of
+[`m.space`](https://spec.matrix.org/v1.16/client-server-api/#types).
+
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of the
```
uhoreg: The spec currently says that the `via` field is required in `m.space.parent` and `m.space.child` events, and says that if `via` isn't present, then the parent/child relationship is considered to not be there.  So I don't think that removing `via` is something that should be done.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Applied.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: I've now mentioned this after the optional `via` removal.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: You're right, however, the spec also explicitly says you can omit the `via` field if you want to remove a room from a space:
> Child rooms can be removed from a space by omitting the `via` key of `content` on the relevant state event, such as through redaction or otherwise clearing the content.

https://spec.matrix.org/v1.17/client-server-api/#mspacechild-relationship

There is a spec issue about this: https://github.com/matrix-org/matrix-spec/issues/2293

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Same as the ones in the new room, I have clarified this now.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: Sorry, this has now been fixed.

proposals/4168-update-space-state-on-room-upgrade.md
Kladki: I have explained why it's a problem now. As for evidence, I'm not sure what you're looking for, since stuff like user studies aren't usually needed for MSCs.

clokep:
@mscbot resolve Lack of details about updating references to replaced room

proposals/4168-update-space-state-on-room-upgrade.md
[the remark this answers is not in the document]
clokep: I think leaving this as a `MAY` isn't really that useful as it is the current behavior. It also means that clients will need to handle either situation to sanely display space trees.

Maybe we should defer this to a separate MSC though.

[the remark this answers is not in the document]
Kladki: > I think leaving this as a `MAY` isn't really that useful as it is the current behavior.

True, but at least now it would be documented.

> It also means that clients will need to handle either situation to sanely display space trees.
> 
> Maybe we should defer this to a separate MSC though.

As mentioned in https://github.com/matrix-org/matrix-spec/issues/2194, this would probably need changes to the space hierarchy/room summary format (the addition of something like an `upgraded_room_id` field), which I think like you suggested, would be better in a separate MSC.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,301 @@
...
+  copying the `order` and `suggested` fields from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- In rooms that reference the old room via `m.space.parent` events (roughly speaking: child rooms
+  of an upgraded space), the upgrading server SHOULD send a new `m.space.parent` event with
+  `state_key` set to the new room's ID. If the previous `m.space.parent` event has `canonical` set
+  to `true` in `content`, homeservers SHOULD update the old state event to set `canonical` to
+  `false`, while setting it to `true` in the newly-sent `m.space.parent` event.
+
+Like the events sent in the new room, these events sent in existing rooms SHOULD be set to the user
+who performed the upgrade.
```
richvdh, line 40: what does it mean to "set an event to a user"?

richvdh: assuming you mean that the *sender* should be set to that user, what happens if the user who did the upgrade is not in the room?

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,301 @@
...
+would be met with a screen telling them that the conversation has moved somewhere else. On top of
+that, clients which organize rooms by the space they're in wouldn't organize upgraded space
+members, as they would not be part of the space. While the space members can be updated manually,
+this can be very tedious and prone to error (i.e. missing specific rooms from the manual process),
+especially for spaces with many rooms and sub-spaces.
+
+## Proposal
+
+In the following sentences, "relevant space state events" refer to `m.space.parent` events for all
+room types, in addition to `m.space.child` events for rooms with a type of
```
richvdh, line 18:
```suggestion
In the following sentences, "relevant space state events" refer to [`m.space.parent`](https://spec.matrix.org/v1.17/client-server-api/#mspaceparent) events for all
room types, in addition to [`m.space.child`](https://spec.matrix.org/v1.17/client-server-api/#mspacechild) events for rooms with a type of
```

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,301 @@
...
+Like the events sent in the new room, these events sent in existing rooms SHOULD be set to the user
+who performed the upgrade.
+
+Note that this will only be possible in rooms where the upgrading
+user (or any other user on the same homeserver, if the implementation decides to use any user it
+can) has the power to do so.
+
+Additionally, for both event types, homeserver implementations MAY remove the `via` field of
+relevant space events referencing the previous room, to signal to clients that they shouldn't join
+the previous room. Otherwise, said events SHOULD remain unchanged.
```
richvdh, line 48: This doesn't appear to match what the reference implementation does. I'm not familiar with continuwuity's codebase, but lines 343-360 of https://forgejo.ellis.link/continuwuation/continuwuity/pulls/907/files appear to remove *all* fields of the `m.space.child` event, not just the `via` field.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,301 @@
...
+
+When a room upgrade is performed, servers SHOULD copy relevant space state events from the old room
+to the new room. The `sender` field in the new event SHOULD be set to the user who performed the
+upgrade.
+
+In addition, servers SHOULD send new relevant space state events pointing to the upgraded room in
+rooms that reference the old room via `m.space.parent` or `m.space.child` events. In practice, this
+means:
+- In rooms that reference the old room via `m.space.child` events (roughly speaking:
+  spaces which are parents of the upgraded room), the upgrading server
```
richvdh, line 29: > In rooms that reference the old room via `m.space.child` events 

This doesn't seem to quite match the behaviour in https://forgejo.ellis.link/continuwuation/continuwuity/pulls/907. That implementation seems to look at rooms which are the target of the old room's `m.space.parent` events, rather than all rooms which reference the old room via `m.space.child`.

proposals/4168-update-space-state-on-room-upgrade.md
```diff
@@ -0,0 +1,301 @@
...
+  SHOULD send a new `m.space.child` event with `state_key` set to the new room's ID,
+  copying the `order` and `suggested` fields from the `content` of the
+  `m.space.child` with `state_key` of the previous room ID.
+- In rooms that reference the old room via `m.space.parent` events (roughly speaking: child rooms
+  of an upgraded space), the upgrading server SHOULD send a new `m.space.parent` event with
+  `state_key` set to the new room's ID. If the previous `m.space.parent` event has `canonical` set
+  to `true` in `content`, homeservers SHOULD update the old state event to set `canonical` to
+  `false`, while setting it to `true` in the newly-sent `m.space.parent` event.
```
richvdh, line 37: Is this implemented in https://forgejo.ellis.link/continuwuation/continuwuity/pulls/907? I don't see it, but I could be missing it.

proposals/4168-update-space-state-on-room-upgrade.md
richvdh: some evidence of input from product managers or designers is generally preferred, but fair enough, I'll clear my concern

richvdh:
@mscbot resolve lack of evidence that this is a desirable change

proposals/4168-update-space-state-on-room-upgrade.md
uhoreg: Right, that's what I meant by "and [the spec] says that if via isn't present, then the parent/child relationship is considered to not be there."  But what you've written sounds like you're trying to keep the space relationship, but have the room non-joinable via the space relationship, which would mean that you'd need to change the spec for what it means when `via` isn't present.

If you're actually wanting to remove the relationship, then I'd suggest explicitly saying so, e.g.

```
Additionally, for both event types, homeserver implementations MAY remove the `via` field of
relevant space events referencing the previous room, effectively removing the parent/child
relationship. Otherwise, said events SHOULD remain unchanged. 
```

uhoreg: And if you're suggesting that the relationship is removed, it would also be good to have something listing the tradeoffs between removing or not removing the relationship.

richvdh:
@mscbot concern awaiting changes per review comments

turt2live:
@Kladki are you able to take a look at Rich's comments?

Kladki:
As I have exams coming up, I probably won't until early June. Sorry for the inconvenience.

turt2live:
This looks like it's not quite able to pick up checkboxes at the moment. Cancelling FCP until the MSC is ready for re-review.

@mscbot fcp cancel

To submit for re-review, please visit the [SCT Office](https://matrix.to/#/#sct-office:matrix.org).

proposals/4168-update-space-state-on-room-upgrade.md
timedoutuk: The implementation linked was unintentionally partial, there's an updated implementation at https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1807 which updates both room parents and space children.

PR #4169, opened 2024-07-07
MSC4169: Backwards-compatible redaction sending using `/send`

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/send-redact-backwards-compatibility/proposals/4169-send-redact-backwards-compatibility.md)

### DIFF
--- a/proposals/4169-send-redact-backwards-compatibility.md
+++ b/proposals/4169-send-redact-backwards-compatibility.md
@@ -1,0 +1,59 @@
+# MSC4169: Backwards-compatible redaction sending using `/send`
+[MSC2174] moved the `redacts` key from the top level to inside `content`. It
+also defined forwards- and backwards-compatibility for receiving by defining
+that servers will copy the key to both locations when serving events to clients.
+The `/redact` endpoint is also forwards-compatible because it works in all room
+versions. However, `/send` was not made backwards-compatible. This means that
+clients can't switch to using `/send` unless they keep track of room versions.
+
+## Proposal
+The proposed solution is to have the server adjust the content of
+`/send/m.room.redaction` calls based on the room version. To future-proof the
+backwards compatibility for [MSC2244], conversion between arrays and strings is
+also introduced.
+
+Additionally, the `/redact` endpoint is deprecated, as there is no longer any
+reason to use it.
+
+[MSC2244]: https://github.com/matrix-org/matrix-spec-proposals/pull/2244
+
+* When `/send` is called with a string value for `redacts` in pre-v11 room,
+  versions the server will move the `redacts` field from the content (request
+  body) to the top level of the event being formed.
+* When `/send` is called with a string value for `redacts` in post-MSC2244 room
+  versions, the server will replace the `redacts` field with an array containing
+  the string value provided by the client.
+* When `/send` is called with an array value for `redacts` in pre-MSC2244 room
+  versions, the behavior depends on the number of items in the array:
+  * if the array contains a single item, the server will replace the `redacts`
+    field with the only item inside the array.
+    * if the call is in a pre-v11 room, the field is also moved to the top
+      level.
+  * if the field contains multiple items, the server will return a standard
+    error with `M_MASS_REDACTION_UNSUPPORTED` as the errcode and HTTP 400 as
+    the status code.
+
+## Potential issues
+
+## Alternatives
+### Client-side switching
+Clients could remember room versions themselves and adjust the endpoint based
+on that. This works fine for thick clients which cache data locally anyway, but
+does not work for things like bots which have minimal local state.
+
+### Just use `/redact`
+`/redact` works fine for all room versions. We could discourage using `/send`
+for redactions and prefer `/redact` instead. Extending `/redact` to allow an
+array may also feel less weird than the server modifying content in `/send`.
+
+## Security considerations
+Clients may accidentally send invalid redaction events if they try to redact
+using `/send` in an old room on a server that does not implement this MSC.
+
+## Unstable prefix
+
+## Dependencies
+This MSC defines additional behavior for [MSC2244], which has been accepted,
+but not yet implemented.
+
+[MSC2174]: https://github.com/matrix-org/matrix-spec-proposals/pull/2174


proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,56 @@
...
+[MSC4140]: https://github.com/matrix-org/matrix-spec-proposals/pull/4140
+[MSC2174]: https://github.com/matrix-org/matrix-spec-proposals/pull/2174
+
+## Proposal
+The proposed solution is to have the server move the `redacts` key from the
+content to the top level of the event when `/send/m.room.redaction` is called
+in a pre-v11 room.
+
+Additionally, the `/redact` endpoint is deprecated, as there is no longer any
+reason to use it.
```
turt2live: There's also no reason to use `/ban`, but we haven't deprecated it. It feels useful to keep the endpoint around as a sugar API.

tulir: Some of the member endpoints do have additional validation (e.g. preventing unban via /kick or kick via /unban) or behavior like profile copying, so one could argue `/ban` should also exist for consistency even though it doesn't check prev_membership nor set profile fields. `/redact` wouldn't have either (no consistency reason and no additional behavior)

I don't feel strongly about it though, so I can remove the deprecation if we want to keep it

turt2live: The `/send` method for those membership transitions is also enforced. I'm more +1 to keeping the `/redact` endpoint around, or at least calling out the deprecation in the title so it's a little more obvious.

tulir: > The /send method for those membership transitions is also enforced.

I meant that /unban and /kick are both transitions to leave, but they only work with the correct previous membership. `/send` doesn't care whether the previous membership is `ban` or `join` as long as you have enough permission to send the event.

Anyway, I've moved the deprecation to alternatives

tulir:

Checklist:
  - [x] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/v1.8/appendices/#pagination)?
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

tulir:
Probably no reason to delay this, it's very straightforward

@mscbot fcp merge

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,59 @@
+# MSC4169: Backwards-compatible redaction sending using `/send`
+[MSC2174] moved the `redacts` key from the top level to inside `content`. It
```
anoadragon453:
```suggestion
[MSC2174] moved the `redacts` key from the top level to inside `content` for `v11` and later room versions. It
```

This is an important detail.

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,59 @@
...
+also defined forwards- and backwards-compatibility for receiving by defining
+that servers will copy the key to both locations when serving events to clients.
```
anoadragon453:
```suggestion
that servers will copy the key to both locations when serving events to clients, regardless of room version.
```

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,59 @@
...
+The `/redact` endpoint is also forwards-compatible because it works in all room
+versions. However, `/send` was not made backwards-compatible. This means that
```
anoadragon453, line 6: I don't see MSC2174 specifically calling out either `/redact` or `/send`. Was this an issue with server implementation?

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,59 @@
...
+clients can't switch to using `/send` unless they keep track of room versions.
+
+Additionally, [MSC4140] currently only defines delaying events for `/send`.
+While it could be extended to support `/redact` for self-destructing messages,
+the MSC can also work as-is if `/send` supported redactions in all room
+versions.
```
anoadragon453: I feel like you're circling the use case you're trying to support without explicitly saying it.

I think this could be better expressed as "This inconsistency makes it difficult to implement features such as self-destructing messages (utilising MSC4140's delayed events functionality, which is only implemented for `/send`). As clients cannot reliably use `/send` to send redactions across all room versions while supporting older clients."

tulir: The original reason why I wrote this MSC in 2024 was actually abstracting event sending in a client SDK without special-casing redactions, MSC4140 just brought it up again and was harder to work around

tulir: Have reworded that paragraph

proposals/4169-send-redact-backwards-compatibility.md
tulir: I don't think there was any issue with `/redact`, but servers did obviously have to make changes to the endpoint in room v11 even though the MSC didn't call it out explicitly.

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Alternatives
+### Client-side switching
+Clients could remember room versions themselves and adjust the endpoint based
+on that. This works fine for thick clients which cache data locally anyway, but
+does not work for things like bots which have minimal local state.
+
+### Just use `/redact`
+`/redact` works fine for all room versions. We could discourage using `/send`
+for redactions and prefer `/redact` instead.
```
KitsuneRal: For completeness sake, can you comment why you discarded this alternative?

tulir: The reason is what's mentioned in the introduction (preferring less special-casing for sending events), but I guess I should repeat it in the alternative

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -34,7 +34,10 @@ does not work for things like bots which have minimal local state.
 
 ### Just use `/redact`
 `/redact` works fine for all room versions. We could discourage using `/send`
-for redactions and prefer `/redact` instead.
+for redactions and prefer `/redact` instead. However, this would require more
+special-casing for various things that send events, like MSC4140. It is
+preferable to reduce unnecessary special-casing by allowing `/send` in all room
+versions.
```
KitsuneRal, line 40: One more thing here I just noticed: MSC4084 [forbids using `/send` for mass redactions](https://github.com/matrix-org/matrix-spec-proposals/pull/4084/files#diff-70bb76b49fd213aa12db3caecbb8b7b40e685b8a1594c4de68b95640597e42e2R79-R83). The consequence being that, for one, delayed redactions (if done using the MSC4140 mechanism) would only be usable per event. Not sure if it's intended; but for backwards compatibility that should be okay, I guess?

KitsuneRal: Wait, it's even worse, looks like these two MSCs are just at odds with each other, as MSC4084 introduces the new `/send` endpoint. Should we harmonise the two somehow?

tulir: I think you might have misread it? MSC4084 forbids using /v3/redact and /v3/send for mass redactions, instead it introduces /v4/send which can require UIA. However, the whole MSC is blocked on figuring out how to UIA with OAuth2

Also, MSC4343 changes mass redactions to use a new event type, so there shouldn't be any conflicts with future reauth requirements for that

KitsuneRal: I don't see MSC4084 forbidding to use `/v3/redact` for mass redactions? I actually don't see it saying anything about `/redact` at all, aside from mentioning it as a fallback for individual redactions.

As for `/send` - I think I understand now: by introducing the new endpoint, MSC4084 effectively pushes clients to still update the code; while this MSC extends support in the existing endpoint (even though MSC4084 will eventually deprecate it), so there's no conflict. Did I get it right?

tulir: Yeah that sounds right, plus assuming we do MSC4343, then MSC4084 will only apply to the new mass redaction event type, not the individual redaction type that this MSC is touching.

The original version of this MSC had separate fallbacks for mass redactions (see first commit), but I removed that with the assumption that MSC4343 will be applied before mass redactions actually happen

> I don't see MSC4084 forbidding to use /v3/redact for mass redactions?

Oh right, it's not explicitly forbidden, it just follows from *not* updating /redact to allow arrays as a parameter

KitsuneRal: 🤦 I knew I was missing something obvious: it was never proposed to support event  id arrays in `/redact`.

proposals/4169-send-redact-backwards-compatibility.md
```diff
@@ -0,0 +1,62 @@
+# MSC4169: Backwards-compatible redaction sending using `/send`
+[MSC2174] moved the `redacts` key from the top level to inside `content` for `v11` and later room versions. It
+also defined forwards- and backwards-compatibility for receiving by defining
+that servers will copy the key to both locations when serving events to clients, regardless of room version.
+The `/redact` endpoint is also forwards-compatible because it works in all room
+versions. However, `/send` was not made backwards-compatible. This means that
+clients can't switch to using `/send` unless they keep track of room versions.
```
dbkr, line 7: Is there an example of this actually being a problem in practice or is this purely theoretical?

tulir: There are multiple special-cased code paths along the lines of

```python
if type == "m.room.redaction":
  api_call("/redact")
else:
  api_call("/send")
```

which can be removed after this MSC lands. This one for example: https://github.com/mautrix/go/blob/main/bridgev2/matrix/intent.go#L51-L58 (which already contains an implementation of this MSC by using the standard path when the server advertises support). Beeper's client SDK has a similar send method

The use case that pushed me to actually finish this MSC was MSC4140, which is mentioned in the second paragraph

dbkr: Fair enough - thanks.

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2298

tulir:
Merged :tada:

PR #3550, opened 2021-12-07
MSC3550: Allow HTTP 403 as a response to profile lookups

H-Shay:
This is a MSC proposal generated in response to the feedback on https://github.com/matrix-org/matrix-doc/pull/3530.

Essentially, servers (such as Synapse) may not allow profile lookup over federation. For example, Synapse will return an HTTP 403 in response to `GET /_matrix/client/v3/profile/{userId}` if `allow_profile_lookup_over_federation` is set to `False`. This behavior is not currently documented in the spec. This proposal aims to add HTTP 403 to the list of responses to `GET /_matrix/client/v3/profile/{userId}`. 

Signed off by: H.Shay hillerys@element.io

Rendered: https://github.com/matrix-org/matrix-doc/blob/msc_403/proposals/3550-allow-403-response-profile-lookup.md

PR where this behavior was implemented:  https://github.com/matrix-org/synapse/pull/9203

Preview: https://pr3550--matrix-org-previews.netlify.app

PR #4170, opened 2024-07-08
MSCXXXX: 403 error responses for profile APIs

Johennes:
---


proposals/4170-profile-403.md
```diff
@@ -0,0 +1,99 @@
+# MSC4170: 403 error responses for profile APIs
+
+Matrix currently defines the following [client-server APIs] for profile look-ups:
+
+- GET /_matrix/client/v3/profile/{userId}
+- GET /_matrix/client/v3/profile/{userId}/avatar_url
+- GET /_matrix/client/v3/profile/{userId}/displayname
```
turt2live, line 7: links to the APIs please :)

(applies to the federation endpoint too)

proposals/4170-profile-403.md
turt2live, line 1: Implementation requirements:

* [ ] Server emitting error
* [ ] Existing clients being okay with that error (or expecting it)

clokep: It sounds like Synapse already does this behind a config flag? Is that correct @Johennes?

proposals/4170-profile-403.md
Johennes: Good call. I had linked to the respective sections above the list but links on each endpoints are probably better. Have made it so.

proposals/4170-profile-403.md
Johennes: This is going to be slightly convoluted, sorry. 😕

### Synapse

#### CS endpoints

[`c0e0740bef0db661abce352afaf6c958e276f11d`](https://github.com/element-hq/synapse/commit/c0e0740bef0db661abce352afaf6c958e276f11d#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602R312-R321) implemented a config flag `require_auth_for_profile_requests` that, if enabled and among others, denies profile look-up on the CS endpoints with 403 `M_FORBIDDEN` when the users don't share any rooms.

[`bfb95654c97a8d3aa164eff96ecc13755c1c326d`](https://github.com/element-hq/synapse/commit/bfb95654c97a8d3aa164eff96ecc13755c1c326d#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602R304-R307) later moved this behavior behind a dedicated `limit_profile_requests_to_users_who_share_rooms` config flag.

Looking at the currently [latest commit](https://github.com/element-hq/synapse/commit/393429d6928ce5cef6ced647567b7ec899d69ecc#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602R459-R477), this seems to still be in effect.

#### SS endpoint

[`9bc74743d53f4795b8bbd674d2945962a9bb29f9`](https://github.com/element-hq/synapse/commit/9bc74743d53f4795b8bbd674d2945962a9bb29f9#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602R313-R322) added a config flag `allow_profile_lookup_over_federation` that, if disabled, denies profile look-up on the SS endpoint with 403 `M_FORBIDDEN`.

Looking at the currently [latest commit](https://github.com/element-hq/synapse/commit/393429d6928ce5cef6ced647567b7ec899d69ecc#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602R376-R381), this seems to still be in effect.

#### Summary

Notably the case where a user doesn't share a room with the requester but is in a public room known to the homeserver is not covered when `limit_profile_requests_to_users_who_share_rooms` is enabled. If we remove this part from the proposal, I think Synapse should be conforming.

It looks like it's currently possible to configure Synapse so that it restricts profile look-up on the CS endpoints while allowing unrestricted profile look-up on the SS endpoint. This is probably an issue with Synapse rather than the proposal though.

### Element Web / matrix-react-sdk

[`a05c0f9214833e119f09a0f7742a156482dd88f0`](https://github.com/matrix-org/matrix-react-sdk/commit/a05c0f9214833e119f09a0f7742a156482dd88f0) added logic to the invite process that [maps](https://github.com/matrix-org/matrix-react-sdk/commit/07542b0c405f964f017b9805f81132ce8a79f8c6#diff-823d4e301d9386baca76b33e509a3ab83dafe5f4d02d615c12a934d47dab63e5R177-R178) `M_FORBIDDEN` errors on the CS endpoints to `M_PROFILE_UNDISCLOSED` which is then used to display a ["User may or may not exist" message](https://github.com/matrix-org/matrix-react-sdk/commit/07542b0c405f964f017b9805f81132ce8a79f8c6#diff-823d4e301d9386baca76b33e509a3ab83dafe5f4d02d615c12a934d47dab63e5L243-L245).

This is done regardless of status code though I would argue that clients are likely to catch any 4xx error and turn it into a generic error message rather than just handling the documented 404 exclusively.

clokep: > If we remove this part from the proposal, I think Synapse should be conforming.

This is up to you as the author -- if you wish for this to have an implementation with no implementation work on your part, then sounds like removing that might be the easiest way forward.

Johennes: Hm, it would be great to get the part that only documents reality into the spec timely. Then again, if we leave out the public-room-membership condition, things would still be inconsistent because you could then find profiles via the user directory endpoint that you cannot query via the profile APIs (see Synapse's implementation [here](https://github.com/element-hq/synapse/blob/develop/synapse/storage/databases/main/user_directory.py#L1050-L1054)). So I think I'd rather leave it in which would then mean that Synapse's implementation of `limit_profile_requests_to_users_who_share_rooms` needs to be extended to conform with this proposal.

Johennes: Wait, Synapse also has a `user_directory_search_enabled` config flag that if disabled, [denies](https://github.com/element-hq/synapse/blob/develop/synapse/rest/client/user_directory.py#L70) user directory searches entirely. That doesn't seem spec-compliant either because [`POST /_matrix/client/v3/user_directory/search`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3user_directorysearch) is documented with:

> The homeserver may determine which subset of users are searched, however the homeserver MUST at a minimum consider the users the requesting user shares a room with and those who reside in public rooms (known to the homeserver). The search MUST consider local users to the homeserver, and SHOULD query remote users as part of the search.

I'm a little lost on how to proceed here. Technically, when `limit_profile_requests_to_users_who_share_rooms` is disabled in Synapse, it conforms with this proposal because it doesn't limit profile look-ups at all. When the flag is enabled, Synapse doesn't conform with this proposal because it doesn't consider users in public rooms. It does already have other config flags like `user_directory_search_enabled` that make it behave in a non-complying way though.

I think it would be good to leave the public-room-membership condition in the proposal because the user directory and profile endpoints should behave consistently. I'm unsure what, if any, implementation would be required to support that though. 🤔

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,111 @@
...
+
+Unfortunately, the concrete semantics of when to respond with 403 are not fully
+spelled out in the spec and understanding prior proposals' intention requires some
+archeology.
+
+Initially, [MSC1301] introduced a 403 `M_USER_NOT_PUBLIC` response to disallow
+profile look-up for users that the requester does not share a room with. This was
+implemented in Synapse[^1] behind a config flag but only for the client-server
+endpoints and with an error code of `M_FORBIDDEN`. As for the spec, while the
+[user directory] section received a note about similar room membership conditions,
```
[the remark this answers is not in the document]
richvdh: It's also not obvious, to the casual reader, what this has to do with the proposal: this is the first and only mention of the "user directory". If you want to cite the user directory as precedent or something to be aligned with, we could probably do with an explanation of how they interact.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,111 @@
...
+> 403 	The server is unwilling to disclose whether the user exists and/or has profile information.
+
+Unfortunately, the concrete semantics of when to respond with 403 are not fully
+spelled out in the spec and understanding prior proposals' intention requires some
+archeology.
+
+Initially, [MSC1301] introduced a 403 `M_USER_NOT_PUBLIC` response to disallow
+profile look-up for users that the requester does not share a room with. This was
+implemented in Synapse[^1] behind a config flag but only for the client-server
+endpoints and with an error code of `M_FORBIDDEN`. As for the spec, while the
```
richvdh, line 38:
```suggestion
profile look-up for users that the requester does not share a room with. This was
partially implemented in Synapse[^1] behind a config flag: it was only implemented for the client-server
endpoints, and an error code of `M_FORBIDDEN` was used rather than `M_USER_NOT_PUBLIC`. As for the spec, while the
```

Is that correct? Please amend if not.

proposals/4170-profile-403.md
KitsuneRal: The way the proposal is framed now is quite permissible, as I read it. I guess the question here is mainly whether it's desirable to keep it this way? As a client author, I wouldn't mind: the key point for me is that I have to handle 403 `M_FORBIDDEN` reasonably (Quotient-based clients do). As an end user, I might appreciate more prescriptive language but I think it's alright if some implementations/configurations are overly protective. In short, I think the current Synapse can be deemed compliant with the current MSC language; if there's desire we can split hair over further but I don't think it's necessary.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,111 @@
...
+
+homeservers MUST at a minimum allow profile look-up for users that either share a room
+with the requester or reside in a public room known to the homeserver. In all other
+cases, homeservers MAY deny profile look-up by responding with 403 `M_FORBIDDEN`.
+
+If a remote user is queried through the client-server endpoints and the query is not
+denied per the preceding paragraph, homeservers SHOULD query the remote server for the
+user's profile information.
+
+Homeservers MAY disable profile look-up over federation by responding with 403 `M_FORBIDDEN`
```
richvdh, line 67:
```suggestion
Homeservers MAY deny profile look-up over federation by responding with 403 `M_FORBIDDEN`
```

proposals/4170-profile-403.md
richvdh: So I think you're saying that, in its default configuration (ie, `require_auth_for_profile_requests: false`, `limit_profile_requests_to_users_who_share_rooms: false`, and `user_directory_search_enabled: true`), Synapse complies with this proposal?

As far as the implementation requirements for an MSC go, it would mean that we've already proven this is a workable requirement: clearly there are compliant, working implementations in the ecosystem. That's all that's needed.

Now, if we're proposing requirements which would make some (common) configurations of Synapse non-compliant, that's worth calling out in the text of the MSC, to make sure that readers have relevant context. Ultimately if the MSC is accepted, it would become a thing that Synapse would have to fix before declaring itself compliant with the next spec version.

[Also: if you believe `user_directory_search_enabled` makes synapse non-spec-compliant, I'd suggest filing a synapse bug.]

proposals/4170-profile-403.md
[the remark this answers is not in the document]
Johennes: This was meant to refer to:

> the homeserver MUST at a minimum consider the users the requesting user shares a room with [...]

Let me see if I can clarify this and the general relation between the profile and user directory APIs.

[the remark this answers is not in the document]
Johennes: I have updated the proposal accordingly: https://github.com/matrix-org/matrix-spec-proposals/pull/4170/commits/3baad1ac13852162123e4ba2c79e18ab30c67114

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,111 @@
...
+## Proposal
+
+For the endpoints in the client-server API
+
+- [`GET /_matrix/client/v3/profile/{userId}`]
+- [`GET /_matrix/client/v3/profile/{userId}/avatar_url`]
+- [`GET /_matrix/client/v3/profile/{userId}/displayname`]
+
+homeservers MUST at a minimum allow profile look-up for users that either share a room
+with the requester or reside in a public room known to the homeserver. In all other
```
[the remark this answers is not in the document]
Johennes: > Could we leave this problem open for a future proposal and just state that "public room" in this proposal has the same (albeit unspecified) semantics that the user directory API has today?

I've went this way with https://github.com/matrix-org/matrix-spec-proposals/pull/4170/commits/c1aafd8432ea278708fd5fb72da7f948d1816019 but please let me know if you think this is insufficient.

proposals/4170-profile-403.md
Johennes: Thanks a lot for the input all. 🙏

> Now, if we're proposing requirements which would make some (common) configurations of Synapse non-compliant, that's worth calling out in the text of the MSC, to make sure that readers have relevant context. Ultimately if the MSC is accepted, it would become a thing that Synapse would have to fix before declaring itself compliant with the next spec version.

Makes sense. I have called this out in the "Potential issues" section.

> [Also: if you believe user_directory_search_enabled makes synapse non-spec-compliant, I'd suggest filing a synapse bug.]

Good call. I have reported this in https://github.com/element-hq/synapse/issues/17534.

With the above, it then seems that no additional implementations should be required for this proposal.

richvdh: Have removed the `needs-implementation` label.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+## Proposal
+
+For the endpoints in the client-server API
+
+- [`GET /_matrix/client/v3/profile/{userId}`]
+- [`GET /_matrix/client/v3/profile/{userId}/avatar_url`]
+- [`GET /_matrix/client/v3/profile/{userId}/displayname`]
+
+homeservers MUST at a minimum allow profile look-up for users that either share a room
+with the requester or reside in a public[^3] room known to the homeserver[^4]. In all other
```
richvdh, line 64: Navigating the footnotes here is a bit fiddly.

```suggestion
homeservers MUST at a minimum allow profile look-up for users that either share a room
with the requester or reside in a public room known to the homeserver (i.e, the same
requirements as [`POST /_matrix/client/v3/user_directory/search`])[^3]. In all other
```

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+Later, [MSC1301] introduced a 403 `M_USER_NOT_PUBLIC` response on all four profile
+endpoints to optionally disallow profile look-up for users that the requester does
+not share a room with. This did not end up being documented in the spec but in 2019
+was partially implemented by Synapse[^1]: it was only implemented for the client-server
+endpoints, and an error code of `M_FORBIDDEN` was used rather than `M_USER_NOT_PUBLIC`.
+
+In 2021, Synapse implemented[^2] a switchable feature to disable profile look-up
+over federation via a 403 `M_FORBIDDEN` response. [MSC3550] picked up on this
+feature and introduced this response type in the spec though only on the
+`GET /_matrix/client/v3/profile/{userId}` endpoint in the client-server API.
```
richvdh, line 49: could you move this stuff to a separate "History" section towards the end, or something? It's useful context, but a bit hard-going in the introduction to the MSC.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+      definition.
+[^4]: This matches the existing requirements for [`POST /_matrix/client/v3/user_directory/search`].
+
+[`GET /_matrix/client/v3/profile/{userId}`]: https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3profileuserid
+[`GET /_matrix/client/v3/profile/{userId}/avatar_url`]: https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3profileuseridavatar_url
+[`GET /_matrix/client/v3/profile/{userId}/displayname`]: https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3profileuseriddisplayname
+[`GET /_matrix/federation/v1/query/profile`]: https://spec.matrix.org/v1.11/server-server-api/#get_matrixfederationv1queryprofile
+[`POST /_matrix/client/v3/user_directory/search`]: https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3user_directorysearch
+[2017]: https://github.com/matrix-org/matrix-spec-proposals/pull/1096/files#diff-332ce28a7277b9375050644632f99c0e606acb751adc54c64c5faabf981ac7edR35
+[MSC1301]: https://docs.google.com/document/d/1G7JjyTuJlZHieuAflGFWmdKyNViGGLRTWON7AMl0wrM/edit
```
richvdh, line 121: link to the issue please, so we can see the discussion. It's easy to go from the issue to the doc, but hard to go the other way.
```suggestion
[MSC1301]: https://github.com/matrix-org/matrix-spec-proposals/issues/1301
```

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+Unfortunately, the concrete semantics of when to respond with 403 are not fully
+spelled out in the spec and understanding prior proposals' intention requires some
+archeology.
+
+In [2017], the user directory API, [`POST /_matrix/client/v3/user_directory/search`],
+which closely relates to the profile APIs, was introduced into the spec. Since its
+inception, it contained the requirement for servers to consider (at least) users from
+shared and public rooms in the search.
+
+Later, [MSC1301] introduced a 403 `M_USER_NOT_PUBLIC` response on all four profile
```
richvdh, line 40:
```suggestion
Later, [MSC1301] proposed a 403 `M_USER_NOT_PUBLIC` response on all four profile
```

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+endpoints to optionally disallow profile look-up for users that the requester does
+not share a room with. This did not end up being documented in the spec but in 2019
```
richvdh, line 42:
```suggestion
not share a room with. This MSC was never accepted, but in 2019
```

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+
+For the endpoints in the client-server API
+
+- [`GET /_matrix/client/v3/profile/{userId}`]
+- [`GET /_matrix/client/v3/profile/{userId}/avatar_url`]
+- [`GET /_matrix/client/v3/profile/{userId}/displayname`]
+
+homeservers MUST at a minimum allow profile look-up for users that either share a room
+with the requester or reside in a public[^3] room known to the homeserver[^4]. In all other
+cases, homeservers MAY deny profile look-up by responding with 403 `M_FORBIDDEN`.
```
richvdh, line 65: To be clear: it remains implementation-defined as to whether a non-existent user should result in 403 or 404?

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,124 @@
...
+
+If a remote user is queried through the client-server endpoints and the query is not
+denied per the preceding paragraph, homeservers SHOULD query the remote server for the
+user's profile information.
+
+Homeservers MAY deny profile look-up over federation by responding with 403 `M_FORBIDDEN`
+to [`GET /_matrix/federation/v1/query/profile`].
```
richvdh, line 72:
```suggestion
Homeservers MAY deny profile look-up over federation by responding with 403 `M_FORBIDDEN`
to [`GET /_matrix/federation/v1/query/profile`]. To be clear: there is no requirement to return
profiles of users in public or shared rooms over the federation API.
```

proposals/4170-profile-403.md
Johennes: Hm, if the server generally denies profile look-up in all but the required cases, then I thought 403 would be better because it doesn't even disclose whether or not the user exists (see further down in line 76). If the server doesn't restrict profile look-up at all, then 404 seems more natural. There's something in-between these two extremes where the server denies profile look-up for some users only. Maybe it's best to remove the paragraph at line 76 and leave this as an implementation choice, as you suggest.

proposals/4170-profile-403.md
Johennes: Yes, will do but won't be able to make the change before next week as I'm out without access to my laptop this week. Hope that's ok.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,126 @@
...
+None.
+
+
+[^1]: https://github.com/element-hq/synapse/commit/c0e0740bef0db661abce352afaf6c958e276f11d
+[^2]: https://github.com/matrix-org/synapse/pull/9203/files#diff-2f70c35b9dd342bfdaaed445847e0ccabbad63aa9a208d80d38fb248cbf57602L311
+[^3]: As stated in https://github.com/matrix-org/matrix-spec/issues/633, the spec currently
+      doesn't clearly define what a public room is. This proposal does not aim to solve this
+      problem and instead only requires that the user directory and profile APIs use the same
+      definition.
+[^4]: This matches the existing requirements for [`POST /_matrix/client/v3/user_directory/search`].
```
richvdh, line 115: My fault, but this is now orphaned

```suggestion
```

proposals/4170-profile-403.md
richvdh: FWIW, I didn't read the para at L76 as particularly restricting a server's choice about whether to return a 403 or a 404 in the case of a non-existent user.

turt2live: I think we should maintain the implementation's ability to choose, fwiw. We can certainly surround it with advice, but I don't think we need to make the choice for them in this case.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+## Security considerations
+
+None.
+
+
+## Unstable prefix
+
+None.
```
turt2live, line 101: The proposal should explain why an unstable prefix isn't required/possible.

proposals/4170-profile-403.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+## Alternatives
+
+None.
+
+
+## Security considerations
+
+None.
```
turt2live, line 96: Privacy and security are synonymous here - how is the newly proposed behaviour impacting privacy?

turt2live (approved):
overall the concept looks great - just a few small details to clarify before this is set for FCP imo.

proposals/4170-profile-403.md
Johennes: Ok, sounds good. Have reworded to paragraph accordingly: https://github.com/matrix-org/matrix-spec-proposals/pull/4170/commits/76391ea3e2bf8939cbb5eaad7ca410de843968ae

Johennes:
> I think this would fix [matrix-org/matrix-spec#168](https://github.com/matrix-org/matrix-spec/issues/168)? @Johennes: might be good to mention that in the proposal text.

@richvdh I think this issue has actually already been addressed through [MSC3550](https://github.com/matrix-org/matrix-spec-proposals/pull/3550). Have [commented](https://github.com/matrix-org/matrix-spec/issues/168#issuecomment-2295831016) on the issue.

proposals/4170-profile-403.md
Johennes: Have addressed this in https://github.com/matrix-org/matrix-spec-proposals/pull/4170/commits/0e1c81eba92d01300ba847727ebb633f9a15ad1a.

proposals/4170-profile-403.md
Johennes: Have tried to explain it with https://github.com/matrix-org/matrix-spec-proposals/pull/4170/commits/1ab5780805aaa6a99a528fb20c333d3c97a967a2.

thoraj:
We're building a multitenant solution based on matrix (homeserver is Synapse). 

An important part of this is delegating user_directory/search to our own backend to enforce user discoverability. 

If a user is discoverable (according to rules enforced in our custom user_directory) it must also  be  possible to invite the user to a room. 

Currently Element require a profile lookup to be able to so. The solution is to allow profile lookups WITHOUT requiring users to share a room (using the synapse setting mentioned above).

AFAIK nothing in the spec mandates such a flag, but IMO the spec shoul make a mention that HS implementation MAY (or SHOULD) allow opting into  profile lookup (non 403/404 responses) even if a user is not in a public or shared room. 

Unrelated to this specific MSC, but the spec should perhaps also spell out the requirements  to be allowed to invite someone into a room?

Johennes:
Thanks for the feedback from an actual use case! 🙏

> Currently Element require a profile lookup to be able to do so. The solution is to allow profile lookups WITHOUT requiring users to share a room (using the synapse setting mentioned above).
> 
> AFAIK nothing in the spec mandates such a flag, but IMO the spec shoul make a mention that HS implementation MAY (or SHOULD) allow opting into profile lookup (non 403/404 responses) even if a user is not in a public or shared room.

Leaving servers the freedom to allow profile look-ups in more, possibly even all cases was definitely the intention of this proposal. It's implicitly captured in [this paragraph](https://github.com/matrix-org/matrix-spec-proposals/pull/4170/files#diff-37ce311403147e79dd0359167ade41ccfea477f04c5b3b02680b2104f8cb87e8R47-R50). The room membership conditions are only a "minimum" requirement and servers "MAY" (not MUST) deny profile queries if these conditions are unmet.

Whether it's worth spelling this out further is probably a discussion for the matrix-spec pull request[^1] that follows if and when this proposal is accepted.

[^1]: I'll probably just revive https://github.com/matrix-org/matrix-spec/pull/1867

> Unrelated to this specific MSC, but the spec should perhaps also spell out the requirements to be allowed to invite someone into a room?

Interesting question. This is probably best covered in an issue on https://github.com/matrix-org/matrix-spec.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1867

PR #4171, opened 2024-07-19
MSC4171: Service members

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/service-members/proposals/4171-service-members.md)

Implementations:

* ~~Servers (server excluding service members in /sync heroes)~~
  * ~~Synapse: https://github.com/element-hq/synapse/pull/17866~~
* Clients:
  * Element Web: https://github.com/matrix-org/matrix-js-sdk/pull/1771 / https://github.com/matrix-org/matrix-react-sdk/pull/9124 / https://github.com/matrix-org/matrix-react-sdk/pull/12157
  * Element Android: https://github.com/element-hq/element-android/pull/8700
  * Element iOS: https://github.com/element-hq/element-ios/pull/4643
  * gomuks web: https://github.com/tulir/gomuks/blob/e9abcd50d1e4f735790617667b296cea85bf530b/pkg/hicli/sync.go#L778-L798
  * Rust SDK: https://github.com/matrix-org/matrix-rust-sdk/pull/4335
* Bridges & bots (generating service member lists):
  * mautrix-go: https://github.com/mautrix/go/blob/v0.19.0/bridgev2/portal.go#L2346-L2349


proposals/4171-service-members.md
```diff
@@ -0,0 +1,51 @@
...
+while due to the complexity of the system. This proposal is an intermediate
+solution for bots before canonical DMs are fully implemented.
+
+[Canonical DMs]: https://github.com/matrix-org/matrix-spec-proposals/pull/2199
+
+## Proposal
+The proposed solution is a new state event: `m.functional_members`. The state
+event contains a field called `service_members` which is a list of user IDs.
+
+TODO: should the key be the same as the event type?
```
clokep, line 15: It is a bit weird that the state event type refers to the "real" members but then lists the bot members in the room.

proposals/4171-service-members.md
tulir, line 1: Apparently there was already a doc for this, probably should've looked for that before writing it myself 🙈 https://github.com/element-hq/element-meta/blob/develop/spec/functional_members.md

proposals/4171-service-members.md
turt2live: With an Element hat: The event type reflects the feature name rather than the contents, for better or worse. It's a minified version of the Canonical DMs concept of "unimportant" users, where the remaining "important" users are called "functional" members.

This MSC would need a comparison to MSC2199 if it's not already there (I'm replying here from the conversation view, not the diff view)

Half-Shot: I suspect it would have been better to have picked something like `m.member_hints` or something, and then allowed room for several types of hints to be given to the client. Potentially not too late to change if we decide the Unstable prefix uses a different naming scheme but is functionally identical (to stop existing state from breaking)

proposals/4171-service-members.md
```diff
@@ -0,0 +1,51 @@
...
+has two non-service members, then any additional DM-specific treatment should
+be applied as well (for example, using 1:1 semantics for VoIP calls).
+
+To prevent hiding encrypted message recipients, service members MUST still be
+visible in the member list and membership events they send MUST be rendered if
+other users' membership events are rendered.
+
+Clients may include some special label for service members in the member list
+to make it clearer why they're not included in the room name.
+
```
Half-Shot, line 28: If it wasn't obvious, the client should ignore any members not present in the room but in this list (i.e. don't get fancy a show a ghosted member or anything in the member list).

proposals/4171-service-members.md
```diff
@@ -0,0 +1,51 @@
...
+## Potential issues
+Clients that don't support this MSC would still include the service members in
+the room name calculation. It may be possible to work around this by excluding
+service members from `m.heroes` on the server side, but that may also cause
```
Half-Shot, line 32: I wonder if the MSC should explicitly direct servers to filter out these members from the `m.heros` selection. The heros section isn't used. As the spec says:

>`m.heros` -> The users which can be used to generate a room name if the room does not have one. Required if the room’s m.room.name or m.room.canonical_alias state events are unset or empty..

So reasonable to say this is equivalent to the client doing it, and we should probably not have a disagreement between server and client.

proposals/4171-service-members.md
```diff
@@ -0,0 +1,51 @@
...
+other issues.
+
+## Alternatives
+### Inverted user list
+The state event could be inverted to list real members instead of service
+members. This approach was not chosen for backwards-compatibility reasons: if
+a user is invited with a client that doesn't know about this MSC, it's better
+for bots to show up in the name than it is for real users to be hidden from the
+name.
```
Half-Shot, line 41: It would also mean updating this state event all the time, for example a hidden bot in a large room would mean we'd need to send an updated event upon each join. Not to mention the event size limitations.

Half-Shot (commented):
Largely I think the text is fine, I just had a few quibbles.

proposals/4171-service-members.md
```diff
@@ -0,0 +1,51 @@
...
+Any users (service members) listed there should not be considered when
+computing the room name or avatar based on the member list. If the room only
+has two non-service members, then any additional DM-specific treatment should
+be applied as well (for example, using 1:1 semantics for VoIP calls).
+
+To prevent hiding encrypted message recipients, service members MUST still be
+visible in the member list and membership events they send MUST be rendered if
+other users' membership events are rendered.
+
+Clients may include some special label for service members in the member list
```
HarHarLinks, line 26: could go beyond the member list and also mark them in the timeline etc, basically where ever they appear in the room

HarHarLinks (approved):

proposals/4171-service-members.md
HarHarLinks: maybe one could say that functional_members that are not room members (maybe previous room members is ok uhm i don't remember, see the element code) should be simply disregarded.

proposals/4171-service-members.md
clokep: That sounds like a better name. No reason to update the unstable version if functionality is the same.

proposals/4171-service-members.md
Half-Shot: Yes: Any listed mxid that is not currently a member of a room can be safely ignored.

tulir: Added some words about that

proposals/4171-service-members.md
tulir: Renamed it to `m.member_hints` (and kept the unstable prefix as-is). Also added canonical DMs to the alternatives section to explicitly point out which part of MSC2199 is equivalent to this MSC (unimportant users)

proposals/4171-service-members.md
tulir: I guess it would be more technically correct, I'm just not sure if all clients like that. For example, I could imagine clients getting confused and thinking they have the full member list even though a service member was omitted

proposals/4171-service-members.md
tulir: Added some words about that

proposals/4171-service-members.md
tulir: Added `and/or other places in the room` to the sentence

proposals/4171-service-members.md
```diff
@@ -0,0 +1,69 @@
...
+Any listed users who are not room members can be safely ignored. Simply being
+on the list without actually being a room member has no special behavior.
+
+## Potential issues
+Clients that don't support this MSC would still include the service members in
+the room name calculation. It may be possible to work around this by excluding
+service members from `m.heroes` on the server side, but that may also cause
+other issues.
+
+## Alternatives
```
HarHarLinks, line 38: also https://github.com/matrix-org/matrix-spec-proposals/pull/4015 I think?

proposals/4171-service-members.md
Half-Shot: Having looked at some of the bugs from *not* implementing this, I think that room summaries should probably implement this. `m.heroes` should probably exclude any mentioned members. This feels like a fairly harmless change, as clients should still attempt to load the full member list when handling a room.

The summary is there to provide a brief overview for the purpose of naming a room, which is ultimately what service_members need excluding from.

tulir: Oh there's already an implementation, I'll add that then

proposals/4171-service-members.md
```diff
@@ -0,0 +1,69 @@
+# MSC4171: Service members
```
Half-Shot, line 1: N.B. Element implements this in their "auditbot" integration, although the code is closed source.

proposals/4171-service-members.md
clokep: Probably just add this as a link somewhere in the MSC?

EDDOEDDO (approved):

PR #4172, opened 2024-07-25
drop real/legal name requirement from DCO

joshsimmons:
Removing the "real" or "legally identifiable" name requirement in line with updated Foundation contribution policy. Also removing the "private sign off" option as that's no longer necessary.


turt2live (approved):
thanks!

PR #4173, opened 2024-07-31
MSC test pusher

p1gp1g:
[rendered](https://github.com/p1gp1g/matrix-spec-proposals/blob/test-pusher/proposals/xxxx-test-pusher.md)


turt2live:
Thanks for opening this MSC! If it's accepted under the process, we'll need to have Sign Off on the changes: https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md

This can be done anytime between now and acceptance, though is typically better earlier.

p1gp1g:
I've added sign off on the commit

PR #4175, opened 2024-08-05
MSC4175: Profile field for user time zone

clokep:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/profile-tz/proposals/4175-profile-field-time-zone.md)

### DIFF
--- a/proposals/4175-profile-field-time-zone.md
+++ b/proposals/4175-profile-field-time-zone.md
@@ -1,0 +1,91 @@
+# MSC4175: Profile field for user time zone
+
+Knowing another user's time zone is useful for knowing whether they are likely
+to respond or not. Example uses include:
+
+* Showing a user's time zone or time zone offset directly.
+* Showing a user's local time (with hints of whether it is day or night there).
+
+
+## Proposal
+
+Profiles can provide an optional `m.tz` field with values equal to names from the
+[IANA time zone  database](https://data.iana.org/time-zones/theory.html#naming).
+Clients can set and fetch this via the [normal API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles).
+
+* Servers MAY validate that the value is a valid IANA time zone. If deemed invalid
+  they MUST return a 400 error with error code `M_INVALID_PARAM`.
+* Clients MUST handle invalid or unknown values.
+
+If the field is not provided it SHOULD be interpreted as having no time zone information
+for that user.
+
+
+## Potential issues
+
+Clients may need to understand IANA time zone names to show useful information to users.
+Some languages make this easy, e.g. JavaScript can handle this using
+[`Date.toLocaleString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString).
+This may cause clients to bundle the IANA time zone database (and thus also keep it
+up to date).
+
+Clients will need to manually update the profile field when the user changes time zone.
+This could be automated by clients based on location, or left as a manual change to
+users.
+
+
+## Alternatives
+
+The time zone offset could be included directly (in minutes/seconds or in `[+-]HH:MM` form).
+This would require clients to manually update the profile field during daylight
+savings. Using the IANA time zone name is robust against this.
+
+[RFC7095: jCard The JSON Format for vCard](https://datatracker.ietf.org/doc/html/rfc7095)
+format could be used instead, but this doesn't make much sense unless the entire
+profile was replaced.
+
+(Note there's an alternative [jCard](https://microformats.org/wiki/jCard) format
+which is a non-standard derivative of [hCard](https://microformats.org/wiki/hcard).)
+
+
+## Competitive analysis
+
+Slack's [`users.info` API call](https://api.slack.com/methods/users.info) includes
+3 separate fields:
+
+* `tz`: the time zone database name (e.g. `"America/New_York"`)
+* `tz_label`: a friendly name (e.g. `"Eastern Daylight Time"`)
+* `tz_offset`: offset in seconds as an integer (e.g. `-14400`)
+
+
+XMPP directly uses either:
+
+* [XEP-054](https://xmpp.org/extensions/xep-0054.html) uses vCard
+  ([RFC2426](https://datatracker.ietf.org/doc/html/rfc2426)) converted to XML via
+  [draft-dawson-vcard-xml-dtd-01](https://datatracker.ietf.org/doc/html/draft-dawson-vcard-xml-dtd-01)
+* [XEP-0292](https://xmpp.org/extensions/xep-0292.html) uses xCard: vCard XML Representation
+  ([RFC6351](https://datatracker.ietf.org/doc/html/rfc6351)), see also vCard4
+  ([RFC6351](https://datatracker.ietf.org/doc/html/rfc6351))
+
+Rocket.Chat provides a user's [time zone offset](https://developer.rocket.chat/docs/user)
+in the `utcOffset` field.
+
+Mattermost [returns an object](https://api.mattermost.com/#tag/users/operation/GetUser)
+with the user's manual and/or automatic IANA time zone name.
+
+Discord, Twitter, blah don't provide a user's time zone.
+
+## Security considerations
+
+Showing a user's time zone gives some information to their location. There is currently
+no way to limit what profile fields other users can see.
+
+
+## Unstable prefix
+
+`us.cloke.msc4175.tz` should be used in place of `m.tz`. 
+
+
+## Dependencies
+
+Requires [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).


proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,91 @@
...
+
+Mattermost [returns an object](https://api.mattermost.com/#tag/users/operation/GetUser)
+with the user's manual and/or automatic IANA time zone name.
+
+Discord, Twitter, blah don't provide a user's time zone.
+
+## Security considerations
+
+Showing a user's time zone gives some information to their location. There is currently
+no way to limit what profile fields other users can see.
```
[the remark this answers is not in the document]
clokep: My assumption was you would just not set it!

I would think clients would already compare to the local clock and not to any information stored in the user's profile.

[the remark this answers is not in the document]
sumnerevans: Right. I was mainly pointing out that this security consideration is not much of an issue as it can easily just not be set.

If educating users is the concern, we could add a recommendation that clients warn users that the tz will be visible to everyone.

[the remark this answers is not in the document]
clokep: Fair enough! I'll expand it a bit to a client recommendation. Thank you!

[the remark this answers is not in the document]
clokep: Hopefully I've cleared it up.

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,94 @@
...
+
+clients may wish to warn users when providing a timezone and give
```
Half-Shot, line 83:
```suggestion
Clients may wish to warn users when providing a timezone and give
```

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,94 @@
...
+the option to kot incoude it in their profile.
```
Half-Shot, line 84:
```suggestion
Clients may wish to warn users when providing a timezone and give
the option to not include it in their profile.
```

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,94 @@
...
+  they MUST return a 400 error with error code `M_INVALID_PARAM`.
+* Clients MUST handle invalid or unknown values.
+
+If the field is not provided it SHOULD be interpreted as having no time zone information
+for that user.
+
+
+## Potential issues
+
+Clients may need to understand IANA time zone names to show useful information to users.
```
Half-Shot, line 26: It may be worth nodding to the fact that clients shouldn't cache this value, as it's liable to change (frequently, if tracking the user's location)? I suppose there is a problem in that because this data isn't (loosely) synchronised to a state event you can't really wait for the change to come down sync, so clients should refetch this value upon user interaction?

clokep: Added!

Half-Shot:
#12967 is an implementation of this.

girlbossceo:
conduwuit server-side partial implementation: https://github.com/girlbossceo/conduwuit/commit/f163ebf3bbaa9656bbf1d78355220e1d859b17c7#diff-6f6ce4833edde4b67f9f5c950adce5338f3e3a19b6d287c33ff4e9d290b94fa1

ruwuma (our ruma fork) partial implementation for conduwuit:
- https://github.com/girlbossceo/ruwuma/commit/bf93f9caa610bbabc0ab5dab9231505253d0c198
- https://github.com/girlbossceo/ruwuma/commit/0aa4cc4f9c5febacbdba7325129dffe75135890b
- https://github.com/girlbossceo/ruwuma/commit/83a9bf73668aa459ffcf5ca6f95639835c0bac79

The partial implementation only supports GET/PUT/DELETE for just the `us.cloke.msc4175.tz` field. More thought into supporting "generic" K-V fields is being done, but at least gets a basic implementation through the door that can be usable in any future clients that implement the timezone field (which seems likely soon 😇).

clokep:
Judging by this having an implementation I think this isn't really a draft anymore.

tcpipuk:
This is also implemented in gomuks: https://github.com/tulir/gomuks/pull/574

clokep:
This has a couple of implementations and has proven useful already.

I think it is reasonable and doesn't have too much room for abuse.

@mscbot fcp merge

turt2live:

Checklist:
- [ ] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal)
- [ ] For each new endpoint that is introduced:
  - [ ] Have authentication requirements been specified?
  - [ ] Have rate-limiting requirements been specified?
  - [ ] Have guest access requirements been specified?
  - [ ] Are error responses specified?
    - [ ] Does each error case have a specified `errcode` (e.g. `M_FORBIDDEN`) and HTTP status code?
      - [ ] If a new `errcode` is introduced, is it clear that it is new?
- [ ] Will the MSC require a new room version, and if so, has that been made clear?
  - [ ] Is the reason for a new room version clearly stated? For example,
  - [ ] If the endpoint utilises pagination, is it consistent with
- [ ] An introduction exists and clearly outlines the problem being solved.
- [ ] While the exact sections do not need to be present,
  - [ ] Introduction
  - [ ] Proposal text
  - [ ] Potential issues
  - [ ] Alternatives
  - [ ] Dependencies
- [ ] Stable identifiers are used throughout the proposal, except for the unstable prefix section
  - [ ] Unstable prefixes [consider](README.md#unstable-prefixes) the awkward accepted-but-not-merged state
  - [ ] Chosen unstable prefixes do not pollute any global namespace (use “org.matrix.mscXXXX”, not “org.matrix”).
- [ ] Changes have applicable [Sign Off](CONTRIBUTING.md#sign-off) from all authors/editors/contributors
- [ ] There is a dedicated "Security Considerations" section which detail

turt2live:
@mscbot concern checklist incomplete

clokep:
I finished the couple missing bits in the checklist. Thank you for remembering to post!

turt2live:
@mscbot resolve checklist incomplete

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,127 @@
...
+
+* Showing a user's time zone or time zone offset directly.
+* Showing a user's local time (with hints of whether it is day or night there).
+
+
+## Proposal
+
+Profiles can provide an optional `m.tz` field with values equal to names from the
+[IANA Time Zone Database](https://www.iana.org/time-zones).
+Clients can set and fetch this via the [normal API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles).
```
turt2live:
```suggestion
Clients can set and fetch this via the [normal API endpoints](https://spec.matrix.org/v1.14/client-server-api/#profiles).
```

(unstable can and does change, breaking links for the future)

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,127 @@
...
+
+* Servers MAY validate that the value is a valid IANA time zone. If deemed invalid
+  they MUST return a 400 error with error code `M_INVALID_PARAM`.
+* Clients MUST handle invalid or unknown values.
```
turt2live:
```suggestion
* Clients MUST handle invalid or unknown values. One approach may be processing the value as though it was never set.
```

... or something to actually guide implementations on what "MUST" means.

turt2live (approved):
overall lgtm - thanks!

Both of my comments are before-FCP-finishes urgency, and are non-blocking to the start of FCP.

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,127 @@
...
+  they MUST return a 400 error with error code `M_INVALID_PARAM`.
+* Clients MUST handle invalid or unknown values. One approach may be processing the value as though it was never set.
+
+The rationale for somewhat loose validation is that different clients/servers may have
+different understanding of valid time zones, e.g. different versions of the time zone
+database.
+
+If the field is not provided it SHOULD be interpreted as having no time zone information
+for that user.
+
```
dbkr, line 26: I know it might seem fairly obvious but I feel like an example would still be nice here, eg. maybe just the PUT or GET of a profile timezone.

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Alternatives
+
+The time zone offset could be included directly (in minutes/seconds or in `[+-]HH:MM` form).
+This would require clients to manually update the profile field during daylight
+savings. Using the IANA time zone name is robust against this.
+
+
+### Delegate profile fields
```
ara4n, line 80: I have to say i am quite uneasy about this MSC - it feels like a slippery slope towards bloating out the Matrix spec with an overlapping and incomplete implementation of JSContact (RFC9553).  

But on the other hand, RFC9553 simply doesn't contain a "current timezone for user" attribute - it's only available as an attribute on physical addresses.  And RFC9553 is for describing addressbook entries... not profile information (which is subtly semantically different: the data you expect to see in your Contacts app in your phone is definitely different shape to the data you expect to see on someone's LinkedIin profile for instance).

So perhaps the best bet here is to just bite the bullet and spec our own field types like this. But I really hope we end up finding a way to put them in their own registry rather than bulk out the rest of the spec - and to indirect to the likes of RFC9553 where we conceivably can.

clokep: > I have to say i am quite uneasy about this MSC - it feels like a slippery slope towards bloating out the Matrix spec with an overlapping and incomplete implementation of JSContact (RFC9553).

This is the reason I let this sit for so long without pushing this harder after MSC4133 was merged. I'm not really sure how I feel about it but didn't see a better option, especially given there is no explicit time zone field in RFC9553 as you mention below.

I was hoping other SCT members would shout if this seemed to go in the direction of NIH.

> which is subtly semantically different: the data you expect to see in your Contacts app in your phone is definitely different shape to the data you expect to see on someone's LinkedIin profile for instance

Agreed that it is different, much of the info overlaps as it is describing a person, but the context is different.

> So perhaps the best bet here is to just bite the bullet and spec our own field types like this. But I really hope we end up finding a way to put them in their own registry rather than bulk out the rest of the spec - and to indirect to the likes of RFC9553 where we conceivably can.

We could have a field contains a JScontact (or a link to one), but that's a bit beyond this MSC. I agree though that we should not redefine solved things.

proposals/4175-profile-field-time-zone.md
clokep: I'll add some examples.

proposals/4175-profile-field-time-zone.md
```diff
@@ -0,0 +1,127 @@
...
+up to date).
+
+Using the IANA time zone name has the downside that it does now allow arbitrary offsets,
+which may be required for time zones which are not internationally recognized.
+
+Clients will need to manually update the profile field when the user changes time zone.
+This could be automated by clients based on location, or left as a manual change to
+users.
+
+Clients may wish to periodically fetch the time one of other users as it is
```
uhoreg:
```suggestion
Clients may wish to periodically fetch the time zone of other users as it is
```

proposals/4175-profile-field-time-zone.md
dbkr: Did you get a chance to do this? Not super important but would be good to resolve the pending stuff.

clokep: Done! Sorry I missed this.

proposals/4175-profile-field-time-zone.md
clokep: @ara4n Let me know if you want any updates for this or not, I'm unsure if the alternatives section should be fleshed out more?

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2206

turt2live:
Merged 🎉

PR #4176, opened 2024-08-13
MSC: Translatable Errors

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/translated-errors/proposals/4176-translatable-errors.md)

### DIFF
--- a/proposals/4176-translatable-errors.md
+++ b/proposals/4176-translatable-errors.md
@@ -1,0 +1,83 @@
+# MSC4176: Translatable Errors
+
+[Errors in Matrix](https://spec.matrix.org/v1.11/client-server-api/#standard-error-response) contain
+a "human-readable" message which is often shown to users verbatim. Typically, these errors are written
+in English and hardcoded by the server authors, which can be a challenge for users who don't understand
+English or for servers with primarily non-English users.
+
+This proposal clarifies the requirements around `error` alongside a structure for serving different
+language variations of the message for a client to choose from.
+
+## Proposal
+
+`error` is currently described as:
+
+> The `error` string will be a human-readable error message, usually a sentence explaining what went wrong.
+
+This definition is expanded to say it is a *UTF-8-encoded Unicode string* and that only printable
+characters should be used. Clients *should* be aware of Right-to-Left (RTL) and other similar markers
+being used, which may affect their rendering of the error.
+
+The [standard error response](https://spec.matrix.org/v1.11/client-server-api/#standard-error-response)
+is additionally expanded to contain an *optional* mapping object of language code to error message.
+This map *should* be used by clients to pick the most applicable representation of the error for the
+user, falling back to `error` when not available. It is left as an implementation detail for how to
+pick the "most applicable" representation. This map is found under `messages`, a new property:
+
+```jsonc
+{
+  "errcode": "M_WHATEVER",
+  "error": "You can't send that message.",
+  "messages": {
+    "en-US": "You can't send that message.",
+    "fr": "Tu ne peux pas envoyer ce message."
+  }
+}
+```
+
+The language of `error` is left undefined, though is anticipated to remain as typically English.
+
+The language codes under `messages` use the same specification as the `policies` map's language codes
+on [`m.login.terms`](https://spec.matrix.org/v1.11/client-server-api/#definition-mloginterms-params).
+That is, language codes *should* be formatted as per [Section 2.2 of RFC 5646](https://datatracker.ietf.org/doc/html/rfc5646#section-2.2),
+though *may* contain an underscore instead of a dash (`en_US` instead of `en-US`).
+
+A client rendering the error to the user might use the following implementation-specific logic:
+
+1. Check if `messages[user.lang]` is a string. If true, render that. If false, continue.
+2. Check if `messages[config.default_lang]` is a string. If true, render. If false, continue.
+3. Render `error`.
+
+## Potential issues
+
+Servers might end up incorporating relatively complex translation frameworks, or require extensive
+changes to support multiple written languages.
+
+This could potentially be an undesirable feature. Users could instead prefer to have `error` simply
+use their language.
+
+## Alternatives
+
+Request headers may give hints as to what language the user/client prefers, and could be used to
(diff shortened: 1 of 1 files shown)


proposals/4176-translatable-errors.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Client (to demonstrate interest/need)

proposals/4176-translatable-errors.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Potential issues
+
+Servers might end up incorporating relatively complex translation frameworks, or require extensive
+changes to support multiple written languages.
+
+This could potentially be an undesirable feature. Users could instead prefer to have `error` simply
+use their language.
+
+## Alternatives
```
clokep, line 59: Is having more specific `errcode` values another alternative?

turt2live: For some use cases, yes. I'm expecting to publish an MSC soon which uses translatable errors, but a more specific error code would not help. 

(sorry, that future MSC's content is a bit sensitive, so avoiding direct mentions for the moment)

proposals/4176-translatable-errors.md
```diff
@@ -0,0 +1,83 @@
...
+
+Request headers may give hints as to what language the user/client prefers, and could be used to
```
uhoreg, line 61: [`Accept-Language`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) seems like an appropriate header for this use.

MTRNord, line 61: Another Alternative could be an optional query Parameter for the requests. Not the prettiest since it would have to be everywhere(ish) but it's also a thing I have seen elsewhere before.

proposals/4176-translatable-errors.md
```diff
@@ -0,0 +1,83 @@
...
+1. Check if `messages[user.lang]` is a string. If true, render that. If false, continue.
+2. Check if `messages[config.default_lang]` is a string. If true, render. If false, continue.
+3. Render `error`.
+
+## Potential issues
+
+Servers might end up incorporating relatively complex translation frameworks, or require extensive
+changes to support multiple written languages.
+
+This could potentially be an undesirable feature. Users could instead prefer to have `error` simply
```
Half-Shot, line 56: I'd be a bit concerned it's going to increase the per-request (for a given amount of errors in a clients session) footprint by quite a bit. I'd be tempted to say that if your client provides an [Accept-Language](https://github.com/matrix-org/matrix-spec-proposals/pull/4176/files#r1714567960) header then we should populate `messages` with that (and keep the original error as English for backwards compat). If the server doesn't support any langauges suggested by your client then it falls back to English via just providing `error`.

My understanding is the header usage is fairly widespread?

proposals/4176-translatable-errors.md
```diff
@@ -0,0 +1,83 @@
+# MSC4176: Translatable Errors
```
ara4n, line 1: I'm dubious about this. The original intention for these error codes is that the client would expand them out clientside - if anything, shouldn't we be providing structured errors (like structured logging) to do so?  Worst case, folks could optionally implement Accept-Language in order to internationalise the errors on the server, but do we really want servers to have to sprout a translation framework, and then get out of sync with the supported translations on the client; etc?  It feels *much* cleaner for the client to worry entirely about the presentation problem of i18n.

proposals/4176-translatable-errors.md
turt2live: @ara4n [says](https://github.com/matrix-org/matrix-spec-proposals/pull/4176#discussion_r1715054514) (in context of the whole MSC):

> I'm dubious about this. The original intention for these error codes is that the client would expand them out clientside - if anything, shouldn't we be providing structured errors (like structured logging) to do so? Worst case, folks could optionally implement Accept-Language in order to internationalise the errors on the server, but do we really want servers to have to sprout a translation framework, and then get out of sync with the supported translations on the client; etc? It feels _much_ cleaner for the client to worry entirely about the presentation problem of i18n.

proposals/4176-translatable-errors.md
turt2live: moved to other thread: https://github.com/matrix-org/matrix-spec-proposals/pull/4176#discussion_r1715407793

proposals/4176-translatable-errors.md
MTRNord: > @ara4n [says](https://github.com/matrix-org/matrix-spec-proposals/pull/4176#discussion_r1715054514) (in context of the whole MSC):
> 
> > I'm dubious about this. The original intention for these error codes is that the client would expand them out clientside - if anything, shouldn't we be providing structured errors (like structured logging) to do so? Worst case, folks could optionally implement Accept-Language in order to internationalise the errors on the server, but do we really want servers to have to sprout a translation framework, and then get out of sync with the supported translations on the client; etc? It feels _much_ cleaner for the client to worry entirely about the presentation problem of i18n.

the thing with that is that errors currently are server defined freetext often as so many things are just "M_UNKNOWN". Making it really annoying to error handle that. So the current way isnt feasible. Though I agree that lets say a 1MB overhead on every response also feels wrong if it were a dict of possible translations or having "non deterministic" responses across endpoints. Especially as it would be kinda annoying to get an error report with errors in russian when you dont speak it.

neilalexander: IMO the truly correct and hygienic answer here is to identify every place where the existing `M_` error codes are insufficient and define proper error codes for every single one of them. 

Presentation of those well-known errors should be the responsibility of clients. Perhaps at some point some centrally agreed translations of those error codes can be provided open-source to all client authors, but it definitely shouldn't be down to the server to dictate what those are.

It's already a pretty horrible layering violation that server-dictated errors are surfaced directly to the user like this and server-side translations would just cement this mistake into place rather than address it.

The only time `M_UNKNOWN` with freeform text should happen is if the server needs to surface an error that is actually specific to the implementation and not to spec-driven behaviour, i.e. a database connection or transaction error that can't be translated, for example.

PR #4177, opened 2024-08-13
Add upload location hints proposal

Fizzadar:


proposals/4177-upload-location-hints.md
```diff
@@ -0,0 +1,63 @@
...
+
+How do clients/servers interpret location hints? Is it sensible to leave this undefined/open. We 
+could provide explicit options, or a baseline of options, "it is recommended location hints use
+geographical ISO codes" for example.
+
+## Alternatives
+
+Server could use request IP to achieve this without any hints, though inaccurate and invalid when
+a user is abroad. This is reasonable fallback behaviour for a server when clients do not provide
+any location hint.
```
uhoreg, line 49: I don't have a whole lot of experience with CDNs, but would it be reasonable to expose an endpoint that would allow the client to measure the network speed, e.g. by uploading/downloading a small file, from each location?  And then it can use the location with the shortest time.  It might not be entirely accurate, but might be better than guessing based on IP address, or trying to divine the right location based on a location code?

proposals/4177-upload-location-hints.md
```diff
@@ -0,0 +1,63 @@
+# MSC4177: Upload Location Hints
```
uhoreg, line 1: Is this only useful for uploads, or is would something similar also be useful for downloads?

proposals/4177-upload-location-hints.md
```diff
@@ -0,0 +1,63 @@
...
+bucket matching that location. Clients must not expect this to be guaranteed and ultimately the
+server decides.
+
+## Potential issues
+
+How do clients/servers interpret location hints? Is it sensible to leave this undefined/open. We 
+could provide explicit options, or a baseline of options, "it is recommended location hints use
+geographical ISO codes" for example.
+
+## Alternatives
```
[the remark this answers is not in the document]
Mikaela: I think the term you are looking for is EDNS Client Subnet. https://en.wikipedia.org/wiki/EDNS_Client_Subnet

What Wikipedia doesn't mention is NextDNS (and DNS0) and Adguard having more private implementations of it.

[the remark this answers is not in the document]
clokep: Kind of, but I'm talking about how authoritative servers use that + the original source IP to serve close servers, like the [GeoIP backend of PowerDNS](https://doc.powerdns.com/authoritative/backends/geoip.html) or [Roite53 geolocation routing](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-geo.html).

Maybe that's one in the same, but more the application instead of the underlying technology.

proposals/4177-upload-location-hints.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4177-upload-location-hints.md
```diff
@@ -0,0 +1,63 @@
...
+
+When clients request a location the server MAY use that location hint to upload the media to a
+bucket matching that location. Clients must not expect this to be guaranteed and ultimately the
+server decides.
+
+## Potential issues
+
+How do clients/servers interpret location hints? Is it sensible to leave this undefined/open. We 
+could provide explicit options, or a baseline of options, "it is recommended location hints use
+geographical ISO codes" for example.
```
Johennes, line 43: Is the idea that a client would auto-select the most appropriate hint or would the user have to manually pick from the hints the server presents? For auto-selection to work, I think at least a recommendation of hint formats would be needed.

proposals/4177-upload-location-hints.md
```diff
@@ -0,0 +1,63 @@
...
+## Alternatives
+
+Server could use request IP to achieve this without any hints, though inaccurate and invalid when
+a user is abroad. This is reasonable fallback behaviour for a server when clients do not provide
+any location hint.
+
+## Security considerations
+
+None? Reveals bucket locations (to logged in users), but servers are aware of this and there's
+no security issue (I think).
```
Johennes, line 54: Since the hints are expected to be more accurate than IPs, the server would possibly gain more metadata about the client.

proposals/4177-upload-location-hints.md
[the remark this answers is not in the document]
Fizzadar: > Can using something like Anycast routing or geo-based DNS help find the closest location?

Definitely, although requires a more complicated setup. But also using request IP would satisfy most use cases here I think.

My thought process for the PR was to give explicit control/negotiation of locations over to clients/servers rather than the server guessing the best location for a client to upload files to.

Now that I’ve said that I’m not sure there’s really much improvement with this change..

proposals/4177-upload-location-hints.md
Fizzadar: Both really - specified at upload time but yields benefits on download (assuming others downloading are in the same location/region).

proposals/4177-upload-location-hints.md
Fizzadar: Yeah a test endpoint per location would be a handy addition here so clients can make informed decisions.

proposals/4177-upload-location-hints.md
Fizzadar: That’s the idea, perhaps with a speed test endpoint as mentioned above.

PR #4178, opened 2024-08-16
MSC4178: Additional Error Codes for requestToken endpoint

dbkr:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/dbkr/threepid-medium-not-supported/proposals/4178-threepid-medium-not-supported.md)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4178#issuecomment-2297064094)

# Impls
Element web: https://github.com/matrix-org/matrix-react-sdk/pull/12896
Synapse: https://github.com/element-hq/synapse/pull/17578

### DIFF
--- a/proposals/4178-threepid-medium-not-supported.md
+++ b/proposals/4178-threepid-medium-not-supported.md
@@ -12,9 +12,9 @@ Secondly, the supplied third party identifier may be invalid.
 ## Proposal
 
 Firstly, Add the `M_THREEPID_MEDIUM_NOT_SUPPORTED` code to be returned by both
-[`/account/3pid/email/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidemailrequesttoken)
+[`POST /account/3pid/email/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidemailrequesttoken)
 and
-[`/account/3pid/msisdn/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidmsisdnrequesttoken),
+[`POST /account/3pid/msisdn/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidmsisdnrequesttoken),
 defined to mean that the homeserver does not support adding a third party identifier of that medium.
 
 Secondly, allow these endpoints to also return `M_INVALID_PARAM`, to indicate that the third party identifier parameter
@@ -22,17 +22,17 @@ was not a valid identifier for that medium (eg. not a valid phone number).
 
 ## Potential issues
 
-None forseen
+None foreseen.
 
 ## Alternatives
 
-A better UX would be for servers to advertise what third party identifiers that support adding so that clients can
+A better UX would be for servers to advertise what third party identifiers they support adding so that clients can
 inform users before they try to do so. This should be in addition rather than as alternative though: the clearest
 possible API will come from having both.
 
 ## Security considerations
 
-None forseen
+None foreseen.
 
 ## Unstable prefix
 
--- a/proposals/4178-threepid-medium-not-supported.md
+++ b/proposals/4178-threepid-medium-not-supported.md
@@ -1,4 +1,4 @@
-# MSC0000: Error code for Third Party Medium Not Supported
+# MSC4178: Error code for Third Party Medium Not Supported
 
 Homeservers may not always support adding email addresses or phone numbers to a user's account,
 however, there is no error code to signal this situation. Synapse currently returns `M_UNKNOWN`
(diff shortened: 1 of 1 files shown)


dbkr:
@ara4n has given an LGTM in the spec office room, so in hope that this can be a swift and easy win...

@mscbot fcp merge

proposals/4178-threepid-medium-not-supported.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Potential issues
+
+None foreseen.
+
+## Alternatives
+
+A better UX would be for servers to advertise what third party identifiers they support adding so that clients can
+inform users before they try to do so. This should be in addition rather than as alternative though: the clearest
+possible API will come from having both.
```
clokep, line 33: Is there any reason to not just do this? Seems like it would be a small change to `/capabilities`?

proposals/4178-threepid-medium-not-supported.md
```diff
@@ -0,0 +1,46 @@
...
+## Proposal
+
+Firstly, Add the `M_THREEPID_MEDIUM_NOT_SUPPORTED` code to be returned by both
+[`POST /account/3pid/email/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidemailrequesttoken)
+and
+[`POST /account/3pid/msisdn/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidmsisdnrequesttoken),
+defined to mean that the homeserver does not support adding a third party identifier of that medium.
+
+Secondly, allow these endpoints to also return `M_INVALID_PARAM`, to indicate that the third party address
+was not valid for that medium (eg. not a valid phone number).
```
clokep, line 21: Is this meant to be local validation only? Or should this pipe back errors from whatever API your homeserver uses to implement SMS/email?

proposals/4178-threepid-medium-not-supported.md
KitsuneRal: Having it in capabilities would certainly help but error codes are necessary anyway, aren't they?

dbkr: I don't think there's any reason not to do it but it also probably doesn't need to be in the same MSC.

proposals/4178-threepid-medium-not-supported.md
dbkr: The HS will have to actually send an email so presumably the latter? Would this make a difference to what we say in the spec?

clokep: I suppose it wouldn't really make a difference, hopefully any emails and such are sent in the background so maybe there isn't much you can do there anyway.

proposals/4178-threepid-medium-not-supported.md
```diff
@@ -0,0 +1,46 @@
...
+
+Secondly, the supplied third party identifier may be invalid.
+
+## Proposal
+
+Firstly, Add the `M_THREEPID_MEDIUM_NOT_SUPPORTED` code to be returned by both
+[`POST /account/3pid/email/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidemailrequesttoken)
+and
+[`POST /account/3pid/msisdn/requestToken`](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3account3pidmsisdnrequesttoken),
+defined to mean that the homeserver does not support adding a third party identifier of that medium.
```
richvdh, line 18: slightly wondering if we should extend this to `POST /account/3pid/<unknown_medium>/requestToken`, but perhaps that's a job for another day.

proposals/4178-threepid-medium-not-supported.md
```diff
@@ -0,0 +1,46 @@
...
+
+Secondly, allow these endpoints to also return `M_INVALID_PARAM`, to indicate that the third party address
+was not valid for that medium (eg. not a valid phone number).
+
+For both of these codes, HTTP status code 400 should be used.
```
richvdh, line 23: Have you considered extending this to the identity server endpoints? `/_matrix/identity/v2/validate/email/requestToken` etc?

proposals/4178-threepid-medium-not-supported.md
dbkr: Yeah, the whole 3pid management spec could be generalised to arbitrary media, arguably. Might make more sense to do if we ever add another medium.

proposals/4178-threepid-medium-not-supported.md
dbkr: I hadn't really, although only because the need is more obvious for homeservers which commonly may or may not support verifying certain types of identifier. Happy to add this for ID servers too if we'd rather keep symmetry and do it all in one go.

anoadragon453: Given FCP has now started, it's probably best to leave this for a future MSC.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/1944

PR #4179, opened 2024-08-19
MSC0000: Moderation event hiding

tranquillity-codes:


proposals/4179-moderation-hidden.md
```diff
@@ -0,0 +1,47 @@
+## Motivation
+When an account with an offensive profile (profile picture, username...) is removed from a room,
+many clients will render the profile with a message such as "<offensive name> was banned".
+This essentially cements the string in the room history.
```
turt2live, line 4: Redactions are intended to solve this, removing the display name details from the event. This can leave the user ID exposed though, which is not redactable.

Sending clients can also (arguably) decide to not send the display name.

proposals/4179-moderation-hidden.md
turt2live, line 1: Having a comparison to https://github.com/matrix-org/matrix-spec-proposals/pull/3531 would be appreciated.

turt2live (commented):
non-urgent comments, noting the draft status.

turt2live:
When you get a chance, please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes so this MSC is eligible for inclusion in the spec.

proposals/4179-moderation-hidden.md
tranquillity-codes: I am unsure I follow, redacting what exactly ? A ban is rendered by Element, and membership writes are non-redactable. The display name is included in the membership state but is only a part of the problem as the mxid could also be problematic.

tranquillity-codes:
Signed-off-by: Tranquil Ity <tranquillitycodes@proton.me>

proposals/4179-moderation-hidden.md
turt2live: Membership events are redactable because the profile information is there. The profile information is not critical to the protocol and can be redacted safely.

The user ID part is a bit more of a concern though, yes.

proposals/4179-moderation-hidden.md
tranquillity-codes: Based on a quick look, the differences I noticed:
The motivation is different. #3531 is primarily for hiding *message* events pending moderation, while #4179 is for hiding *state* events that *have already happened* but include content that one might not wish to be visible (such as a user with an offensive MXID getting banned)
#3531 is less granular. Its `visibility: hidden` corresponds to the `hidden` level in #4179, but there is no equivalent of the `spoiler` level. Additionally, #3531 uses a free-form string, which cannot be automatically matched on by a client, while #4179 uses a tag list (allowing someone to customize what events they are okay with seeing, and allowing optional fine-grained control over the completeness vs safety of room history per-client)
#3531 requires the sending of a separate event, while #4179 has no such race condition, plus piggy-backs off of edit events for changing the state.

I will note this down in the MSC (and should probably mention that it should support edits, unsure if it requires a mention or not)

proposals/4179-moderation-hidden.md
tranquillity-codes: I tested redacting a membership event - importantly, while `content` was stripped outside of `membership`, `prev_content` was left untouched, and as an example, Element renders `prev_content` even if  `content` has been stripped.

proposals/4179-moderation-hidden.md
tranquillity-codes: Both have been noted in the MSC contents now

tranquillity-codes:
Notice: I'll probably have to rework this due to not being able to figure out a way to handle this properly without introducing a new room version, so I'll redo it with a new room version (auth rules adjustment).

Unsure whether I should edit this MSC or make a new one. This could still be fixed up, but it'd have limited usability as-is (due to only being able to be present when the membership event is sent, and being stripped if the membership event gets redacted). It's primarily oriented at displayname hiding, so I am thinking of introducing an `m.room.name_hidden` state event, and adjusting auth rules so that 8. doesn't reject MXID state keys.

PR #4184, opened 2024-08-28
MSC4184: Dynamic Notification Suppression

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/mention-spam-suppress/proposals/4184-dynamic-notification-suppression.md)

### DIFF
--- a/proposals/4184-dynamic-notification-suppression.md
+++ b/proposals/4184-dynamic-notification-suppression.md
@@ -14,7 +14,7 @@ acceptable number. Further, spammers can somewhat trivially work around the numb
 events 1 step below the limit.
 
 This proposal instead creates a provision in the specification for a server (or client) to decide
-to supress a notification on a per-event basis. Servers can then protect their users by adjusting
+to suppress a notification on a per-event basis. Servers can then protect their users by adjusting
 an X value described above, or implement other suppression mechanisms depending on the situation.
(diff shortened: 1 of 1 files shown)


proposals/4184-dynamic-notification-suppression.md
```diff
@@ -0,0 +1,84 @@
...
+* The notification would expose the user to illegal material.
+
+When the server skips push rule execution on an event, it adds `m.push_rules_executed: false` to the
+`unsigned` object when serving that event over [`/sync`](https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3sync).
+The value may be `true` and included on events served from other endpoints as well, though doesn't
+serve significant utility to the client.
+
+Clients SHOULD NOT execute push rules locally on events with `m.push_rules_executed: false` as this
+may cause notification counts to desynchronize from the server, leading to 'stuck notifications' for
+the user.
```
Johennes, line 43: Clients that don't support this proposal yet, won't act on `push_rules_executed` if a server implements and enables this proposal. Therefore, these clients might run higher risk of hitting 'stuck notifications' which might be worth calling out in the potential issues section.

turt2live:
Having re-discovered this, I don't think it's all that helpful. If we need something like it in the future then this can be re-opened.

PR #4185, opened 2024-08-29
MSC4185: Event Visibility API

tadzik:
[Rendered](https://github.com/tadzik/matrix-spec-proposals/blob/tadzik/msc/event-visibility/proposals/4185-event-visibility-api.md)

Fixes https://github.com/matrix-org/matrix-spec/issues/1747

Synapse implementation: https://github.com/element-hq/synapse/pull/17973
IRC bridge implementation: https://github.com/matrix-org/matrix-appservice-irc/pull/1833

### DIFF
--- a/proposals/4185-event-visibility-api.md
+++ b/proposals/4185-event-visibility-api.md
@@ -1,0 +1,62 @@
+# MSC4185: Event visibility API
+
+Matrix events will often relate to other events.
+A user can send an event A with a relation to event B while not having access to event B in the first place.
+While it is harmless in regular Matrix usage, integrations -- such as bridges
+-- may then find it hard or impossible to figure out how to handle such a situation.
+
+Consider an event that replies to another event.
+An integration may want to process it into a form with an excerpt from the original event, followed by the reply.
+While sensible in the general case, this does open an opportunity for an information leak,
+if the user sending a reply does not in fact have access to the original event.
+
+Integrations have a limited capacity for figuring out whether or not user does have access to a given event.
+As mentioned in [the issue that provoked this proposal](https://github.com/matrix-org/matrix-spec/issues/1747),
+while this information can be derived from the room state, it's not very practical.
+Matrix servers do have that information, and it'd be nice of them to share it.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


Half-Shot (commented):
Looks very sensible to me and fixes a long standing issue we've had with visibility overall. I'd be curious to hear from the homeserver developer community if there are pitfalls with requesting this information frequently.

proposals/4185-event-visibility-api.md
```diff
@@ -0,0 +1,62 @@
...
+## Proposal
+
+This MSC proposes a new API endpoint that'd expose the information that the server already has:
+can user X access the event Y.
+
+It could take form of a endpoint such as
+
+    GET /_matrix/client/v1/can_user_see_event/<room_id>/<user_id>/<event_id>
+
+That responds with a boolean, revealing if `<user_id>` has access to `<event_id>` in `<room_id>`.
```
MadLittleMods, line 27: What determines if a user has access to an event?

I assume this has to do with room membership and history visibility but I feel we should probably explain this explicitly.

And for example, room encryption has nothing to do with this since that happens out of band on the client.

proposals/4185-event-visibility-api.md
```diff
@@ -0,0 +1,62 @@
...
+The endpoint is only accessible to users that do have access to the given event themselves.
+
+## Potential issues
+
+While the homeserver is likely to have event visibility information cached for its local users,
+the operation could be end up costly when performed for remote users.
+Still, even for that case, the homeserver is probably in the best position to calculate this.
+
+## Alternatives
```
MadLittleMods, line 36: One alternative would be to just try fetching the event [`GET /_matrix/client/v3/rooms/{roomId}/event/{eventId}`](https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3roomsroomideventeventid).

> Responses
>
>Status | Description
>--- | ---
>`200`	| The full event.
>`404`	| The event was not found or you do not have permission to read this event.

Application services can already request on the behalf of another user using the `?user_id=@example:example.org` query parameter.

PR #4188, opened 2024-09-04
MSC4188: Recognising HTTP 410 Gone status during discovery

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/gone/proposals/4188-http-gone-status.md)


tcpipuk:
Oops

proposals/4188-http-gone-status.md
turt2live, line 1: Implementation requirements:

* [x] Server advertising `410 Gone`
* [ ] Server respecting `410 Gone`

proposals/4188-http-gone-status.md
```diff
@@ -0,0 +1,59 @@
...
+
+This will avoid unnecessary DNS lookups and requests to servers that no longer wish to federate.
+
+### Server-Server API Changes
+
+When a homeserver receives a `410 Gone` response from the `/.well-known/matrix/server` endpoint,
+it should consider the server offline and halt discovery.
+
+The period until discovery is re-attempted can be determined by the homeserver's implementation,
+such as using exponential backoff, waiting a period of days, or until the next restart.
```
clokep, line 27: The spec should probably give some guidance at least? I think well known responses are supposed to be cached for 24 hours. Does that make sense here?

tcpipuk: This feature is designed for people who want to semi-permanently retire their server, so I didn't want to tell servers to check a *maximum* of 24 hours after the lack attempt... would it be feasible to say "no less than 24 hours" or should I include some sort of maximum like 7 days?

clokep: I had mixed up minimum/maximum I guess. Anyway, I think some sort of guidance makes sense. "For at least 24 hours" might be a decent spot to draw a line. 🤷

proposals/4188-http-gone-status.md
```diff
@@ -0,0 +1,59 @@
...
+## Potential Issues
+
+If the suspension period is too long, it may delay the detection of a server that has come back
+online. Conversely, too short and it may lead to unnecessary requests.
+
+## Alternatives
+
+- **Permanent Suspension**: Instead of a configurable suspension period, homeservers could
+  permanently stop discovery attempts for servers that return a `410 Gone` status. However,
+  this approach may lead to issues if the server comes back online.
```
Johennes, line 47: Since the period is configurable, an administrator wishing to permanently halt discovery on `410 Gone` could simply choose an extremely large period which would virtually achieve their goal?

proposals/4188-http-gone-status.md
```diff
@@ -0,0 +1,59 @@
...
+When a homeserver receives a `410 Gone` response from the `/.well-known/matrix/server` endpoint,
+it should consider the server offline and halt discovery.
+
+The period until discovery is re-attempted can be determined by the homeserver's implementation,
+such as using exponential backoff, waiting a period of days, or until the next restart.
+
+### Implementation Details
+
+- Homeservers should log the `410 Gone` response and the suspension period for transparency and
+  debugging purposes.
```
Johennes, line 32: I'm curious if this is really something the spec needs to cover? On a quick glance we don't appear to have any other places in the SS spec that provide guidelines for internal logging.

tcpipuk: Good point, removed! https://github.com/matrix-org/matrix-spec-proposals/pull/4188/commits/50a55996cc881ed8f8c3b30b63eb6d00d5b819d8

proposals/4188-http-gone-status.md
tcpipuk: Absolutely, yes. I listed it as an alternative because the MSC *could have* specified we never check a server again after receiving the 410 response once, but I'm wary that we should expect the occasional misconfiguration/mistake.

I'm going to add recommend timeframes (at Patrick's request above), but the proposed method (just skipping the rest of discovery and waiting a while) would lower the burden of unwanted requests on both sides, so should already have a significant net positive if this behaviour is standardised :slightly_smiling_face:

proposals/4188-http-gone-status.md
clokep: Additionally I think this requires a quick survey of existing servers to see if it would have any false positives? Would be curious if any deployed servers e.g. properly have DNS SRV delegation setup, but (for some reason) return HTTP 410 for .well-known.

proposals/4188-http-gone-status.md
tcpipuk: Hopefully this looks about right? https://github.com/matrix-org/matrix-spec-proposals/pull/4188/commits/1ad261dc5eb98c7692b924b719685712b642bd07

proposals/4188-http-gone-status.md
tcpipuk: Initially I've pulled the member list of Matrix HQ, created a unique list of 9550 servers, pulled the .well-known for each one, and recorded the HTTP response status.

There were 55 that returned 410 - I've only checked a handful so far and found them all to fail https://federationtester.matrix.org (no SRV, and either timeout or reject port 8448) but I haven't checked them all yet. 

I'll find time soon to check them all and provide a conclusive answer, but my initial testing suggests ex-admins are actually already using this code to signal their server is offline, so this feature may offer an immediate benefit to live servers!

tcpipuk: Sorry about the delay!

Yeah, I haven't found a single server in HQ that returns 410 and appears to be a live Matrix server. There's a wide variety of return codes being used by servers in that room, but it would seem servers currently advertising 410 are trying to signal that they are no longer Matrix servers - there are none I could find that accept port 8448 or have working SRV delegation.

I'm not sure if that'd technically count as an existing implementation of advertising?

proposals/4188-http-gone-status.md
```diff
@@ -0,0 +1,60 @@
...
+  based on their specific needs and network conditions.
+- Homeservers should resume discovery attempts after the suspension period has elapsed, or when new
+  PDUs or EDUs are received from the server, as receiving new activity signals the server is online.
+
+## Potential Issues
+
+If the suspension period is too long, it may delay the detection of a server that has come back
+online. Conversely, too short and it may lead to unnecessary requests.
+
+## Alternatives
```
sumnerevans, line 44: I host my `.well-known` files statically via GitHub Pages[^1] so it would be nice if I could specify something in the JSON to indicate that the homeserver is gone.

For example, maybe:

```json
{
  "m.gone": true
}
```

The issue with this is that it's not really backwards compatible I think.

[^1]: I know this isn't recommended, since it doesn't set the `Content-Type` header correctly, but servers are required to be resilient to this.

tcpipuk: I like this as a possible solution, though as you say, it's technically a second separate one.

Perhaps for sanity you could propose additions/changes to the PR through a review to cover this extra type of "gone" reporting?

Then, not only can I make sure it properly caters to your use case, it'd cause you to be listed as a co-contributor on the PR 🙂

proposals/4188-http-gone-status.md
timedoutuk: > Server advertising `410 Gone`

Does this need to be a server implementation, or can any old server be serving it? I have a few domains now returning `410 Gone` that I've simply thrown into a section in my reverse proxy config. I'd assume it'd be a bit silly to require an implementation to return this code given the whole point is that the implementation won't be there anymore/much longer, but the wording isn't clear.

tcpipuk: My interpretation was that we simply need a server that respects the status code, and at least one server that advertises it - it wouldn't need to be a real Matrix server, just a server that correctly advertises that status code, as the objective is to advertise that there *isn't* a Matrix server here 🙂

timedoutuk: Ah well in that case, both `dendrite.nexy7574.co.uk` and `construct.nexy7574.co.uk` are returning HTTP410 (granted with some spice in the response body to try and further ward off servers that don't respect 410 yet) and I plan to decommission further servers into this same response block later for the foreseeable future. If you'd like to count that as a server responding.

clokep: Looks like this just needs a PR for a server honoring the 410 and we can call FCP on this.

proposals/4188-http-gone-status.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Proposal
+
+Currently, when a Matrix homeserver attempts to discover another server, as a first step it checks
+that server's `/.well-known/matrix/server` endpoint for possible delegation. Any status other than
+`200 OK` is currently considered an error and should result in continuing the discovery process,
+checking for SRV records and/or trying the default port.
+
+This proposal suggests that homeservers should interpret a `410 Gone` status code as an indication
+that the server is permanently offline and should suspend further discovery attempts for a period
```
timedoutuk: I'm confused - here, it is stated that servers should treat other servers as *permanently* offline if they receive a 410 code, but in a later review you said that it was intended for people who want to *semi-permanently* retire, and then further down in the proposal it is recommended that implementations should only honour the 410 status code *temporarily* (for 1-30 days).

Is this supposed to be used to permanently mark a server as "I'm never coming back", or "I'm not here anymore but I might be back later"?

tcpipuk: A bit of both!

The intention is to say "I'm permanently gone" but no one can actually own a DNS domain *permanently*, so we should never store that a domain is permanently un-federatable.

The intent is to offer a mechanism for someone to block federation but also potentially change their mind months/years in the future.

timedoutuk: Hm, in that case, could the "maximum" cache time be upped to something like 90 days? At 30 days, if you're federating with 25k servers (which isn't uncommon - my tiny synapse in 50 odd rooms already has 10k destinations), that's still potentially 300,000 requests a year best-case, but it'd be a third of that if servers only asked every quarter.

Also, correct me if I'm wrong, but if servers received a new event from a server they previously marked offline, they'll discard that "offline" status and resume standard federation operations with them, right? In that case, I don't think an extended cache lifetime will do much harm if a server decides to come back prematurely.

Of course, if this is better suited for a followup MSC, let me know. Just trying to see if I can find a happy balance between being able to mark my servers as never coming back (which they are), giving servers plenty of notice before they start getting TLS cert errors (since I won't be renewing them anymore), and also still having the opportunity to spin up a new server under the same server name later on down the line.

tcpipuk: The quoted numbers are mainly just a recommendation rather than a requirement - it's up to a server to decide how often (if ever) it wants to communicate with another, but I probably wouldn't _recommend_ longer than 30 days in the MSC as that's already quite a long time. 

The main thing I was worried about was that some implementations might not reset the timer, but you make a good point that shouldn't really be a concern. I'll revisit the wording when I'm logged in tomorrow morning!

tcpipuk: @nexy7574 how does https://github.com/matrix-org/matrix-spec-proposals/pull/4188/commits/bc23843225a4056dca2a5ea81c8ebaae000b02a7 look?

timedoutuk: Looks perfect, thanks for the clarifications @tcpipuk :smile:

PR #4190, opened 2024-09-12
Device management for application services

sandhose:


proposals/4190-as-device-management.md
turt2live, line 1: Implementation requirements:
* [ ] Appservice
* [ ] Server

onestacked: Appservice implementation: https://github.com/mautrix/go/pull/288

anoadragon453 (commented):
Overall lgtm, just a couple sticking points.

tulir:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

tulir:
This is implemented in Synapse and most mautrix bridges. It'd be nice if it was in the same spec release as the rest of next-gen auth

@mscbot fcp merge

tulir:
@mscbot concern checklist not complete

proposals/4190-as-device-management.md
girlbossceo: conduwuit homeserver implementation: https://github.com/girlbossceo/conduwuit/commit/24be5794774b7585b6ec1e3dbaa901967d241972

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,48 @@
...
+
+### **`DELETE /_matrix/client/v3/devices/{deviceId}`**
+
+This endpoint no longer requires User-Interactive Authentication for application services.
+
+### **`POST /_matrix/client/v3/delete_devices`**
+
+This endpoint no longer requires User-Interactive Authentication for application services.
+
+### **`POST /_matrix/client/v3/register`**
```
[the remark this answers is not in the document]
tulir: Actually, I'd guess most appservices don't care about the returned token even if they don't set `inhibit_login`, so returning errors would unnecessarily break those. Maybe I'll just add it as a potential issue

tulir:
@mscbot resolve checklist not complete

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,88 @@
...
+
+This endpoint no longer generates a new access token for application services,
+as if `inhibit_login` was set to `true` in the request by default.
+
+## Potential issues
+
+The change to `/v3/register` is technically backwards-incompatible. However,
+most application services likely ignore the returned access token, and next-gen
+auth will break the token generation anyway. It's better to stop returning a
+token than break the endpoint entirely by returning an error.
```
turt2live, line 63: An error would be preferred, as appservices which rely on the token will subtly break rather than a clear an obvious signal.

turt2live (changes requested):

turt2live:
@mscbot concern Breaking change on register endpoint

proposals/4190-as-device-management.md
tulir: Updated to return an error with a new `M_APPSERVICE_LOGIN_UNSUPPORTED` errcode. It could also do `M_MISSING_PARAM` since technically the error is "missing ?inhibit_login=true", but probably better to have a new explicit error code.

richvdh: @turt2live are you happy to remove your concern on this now?

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,93 @@
...
+## Alternatives
+
+A new set of endpoints dedicated to application services could be added to the
+specification, like `GET|PUT|DELETE /_matrix/client/v3/appservices/{appId}/devices/{deviceId}`.
+
+This would have the advantage of not changing the behavior of existing endpoints.
+
+## Dependencies
+
+None. While this MSC is meant for next-gen auth, it can be used independently.
```
tulir, line 85: I just realized that this actually depends on MSC3202 for the device masquerading bits. Creating devices isn't very useful if you don't get a token and don't have any other way to use the device...

3202 might benefit from being split up so OTK count/device list/etc signaling is in one MSC and device masquerading is in another

tulir:
@mscbot concern This actually depends on MSC3202 device masquerading to be useful

proposals/4190-as-device-management.md
onestacked: Splitting off the masquerading seems like a good idea to me.

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,93 @@
...
+
+### **`DELETE /_matrix/client/v3/devices/{deviceId}`**
+
+This endpoint no longer requires User-Interactive Authentication for application services.
+
+### **`POST /_matrix/client/v3/delete_devices`**
+
+This endpoint no longer requires User-Interactive Authentication for application services.
+
+### **`POST /_matrix/client/v3/register`**
```
dbkr: This feels unrelated to what this MSC says its doing?

tulir: The purpose of the MSC is next-gen auth compatibility. /register by default creates a new access token like /login, which isn't possible when using MAS. Unlike /login, /register has other uses (creating the row in the users table), so it's not removed entirely, just the incompatible part is disabled by mandating inhibit_login.

dbkr: Yep, this is the point I was trying to make: the title refers to device management and next gen auth is only referenced down in the issues section. I guess this does mention device IDs, but I would find it useful if this explained that device creation for ASs was being divorced from this endpoint to explain why they must call it with the inhibit_login flag to only create a user - or something to that effect maybe?

tulir: Might be clearer now

dbkr: Thanks, seems better to me.

proposals/4190-as-device-management.md
turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4326

turt2live:
@mscbot resolve Breaking change on register endpoint

turt2live (approved):
I guess I should press this button too

noelportillo (approved):

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
+# MSC4190: Device management for application services
+
+[MSC3202] allows application services to handle and send encrypted events.
+One part of [MSC3202] is the ability to masquerade devices using the `device_id`
```
richvdh: these links are broken

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+## Proposal
+
+This MSC proposes to extend existing endpoints to allow application services to
+create and delete devices for their users without relying on the `/login` and
+`/logout` mechanisms.
+
+As all changes here only apply to application services, guest access is not
+relevant.
+
+### **`PUT /_matrix/client/v3/devices/{deviceId}`**
```
richvdh: it would be helpful to link to the existing specs for existing endpoints, so we can see what's changing.

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+However, application services don't have an endpoint to create devices for their
+users, which means that, in practice, encrypted application services still use
+`/login` with the `m.login.application_service` login type to create devices for
+their users.
+
+Consequently, such application services leave many unused but active access
+tokens for those users.
+
+Furthermore, if [MSC3861] were adopted, the `/login` endpoint would no longer be
+available for application services to use.
```
richvdh: MSC3861 has been adopted, so this sentence is outdated.

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+
+### **`POST /_matrix/client/v3/delete_devices`**
+
+This endpoint no longer requires User-Interactive Authentication for application services.
+
+### **`POST /_matrix/client/v3/register`**
+
+Currently, the default behavior for `/register` is to create a new device and
+access token (i.e. login) in addition to creating the user. Similar to `/login`,
+creating an access token would no longer be possible with [MSC3861]. However,
```
richvdh: This bit has also been overtaken by the events of MSC3861.

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+creating users via the endpoint is still required, so unlike `/login`, `/register`
+will not be removed entirely.
+
+Therefore, application services MUST call the endpoint with `inhibit_login=true`.
+Calls without the parameter, or with a different value than `true`, will return
+HTTP 400 with a new `M_APPSERVICE_LOGIN_UNSUPPORTED` error code.
```
richvdh: I don't see this in the [sample implementation](https://github.com/element-hq/synapse/pull/17705/), and it feels like an unnecessary breaking change. Are you sure it is necessary to break compatibility with existing app services? Has this been tested in an implementation?

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Potential issues
+
+The change to `/v3/register` is technically backwards-incompatible, but it will
+break when switching to next-gen auth in any case, so a new endpoint version
+would not be useful.
```
richvdh, line 89: ah, so it is already broken on those servers that have switched to next-gen auth?

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Dependencies
+
+In order to use the devices created using this MSC, appservices need to be able
+to use device IDs as a part of identity assertion, as defined by [MSC4326].
+
+## Unstable prefix
+
+Until this MSC is stable, application services must opt-in to the new behavior
+by setting the `io.element.msc4190` flag to `true` in their registration file.
```
richvdh: why is opting in necessary?

proposals/4190-as-device-management.md
tulir: An explicit error rather than silently forcing inhibit_login was added following @turt2live's concern about silent breakage. Looks like it's missing implementation though, so that needs to be fixed in synapse

proposals/4190-as-device-management.md
tulir: Yeah, the endpoint has silently stopped returning access tokens and device ids regardless of whether inhibit_login is set

proposals/4190-as-device-management.md
tulir: I think it opts into disabling /login and requiring inhibit_login on /register so it can be tested even without MAS? Not entirely sure, but it says it's only opt-in until the MSC is stable, so probably doesn't matter

proposals/4190-as-device-management.md
tulir: Actually looks like synapse doesn't have code for that case (MAS enabled + no MSC4190 + no inhibit_login) :thinking: I wonder if it's just exploding or returning broken access tokens instead

proposals/4190-as-device-management.md
richvdh: It does matter, because the fact that there is an opt-in implies that someone is worried it will cause breakage on existing applications, and that logic still applies after the MSC is stable.

proposals/4190-as-device-management.md
richvdh: I would definitely like to see this compatibility story clarified in the MSC.

richvdh:
@mscbot concern lack of clarity over backwards compatibility

proposals/4190-as-device-management.md
tulir: I've changed the proposal so that the registration flag is required after the MSC is accepted too, but is *not* required after a server switches to OAuth2 (because at that point, the breaking changes will happen regardless of the flag, so may as well default to true)

proposals/4190-as-device-management.md
tulir: It should be clearer now with the stable opt-in flag. I've also updated the Synapse implementation to actually return the errors this MSC defines instead of having undefined behavior: https://github.com/element-hq/synapse/pull/18946

tulir:
I've made some more changes to the MSC:

* The wording has been updated as OAuth2 is in the spec now.
* The registration flag will now be used even after the MSC is accepted, which means old appservices will not break until the server switches to OAuth2 (the flag is forced to `true` when the server switches, as at that point the old login method would break anyway)
* UIA is no longer required for resetting cross-signing keys. It's not exactly a part of device management, so if there are concerns with including it here, I can extract it into a new very tiny MSC. It's easier to keep it here though.

https://github.com/element-hq/synapse/pull/18946 should finish the synapse implementation (adds the error responses that were missing previously as well as forcing the flag when MAS is enabled and allowing the new cross-signing UIA bypass)

anoadragon453 (approved):
> UIA is no longer required for resetting cross-signing keys. It's not exactly a part of device management, so if there are concerns with including it here, I can extract it into a new very tiny MSC. It's easier to keep it here though.

I think this is fine to keep in this MSC to avoid unnecessary process overhead.

anoadragon453:
@mscbot concern checklist not complete

Missing "Are all MSCs that this MSC depends on already accepted?" while MSC4326 is not yet accepted.

anoadragon453:
@mscbot concern checklist not complete again

tulir:
I was keeping the MSC3202 concern to wait for dependencies, but I guess blocking on the checklist is more accurate (especially now that the dependency is MSC4326, not MSC3202) so I'll resolve mine

@mscbot resolve This actually depends on MSC3202 device masquerading to be useful

proposals/4190-as-device-management.md
```diff
@@ -0,0 +1,130 @@
...
+
+However, application services don't have an endpoint to create devices for their
+users, which means that, in practice, encrypted application services still use
+`/login` with the `m.login.application_service` login type to create devices for
+their users.
+
+Consequently, such application services leave many unused but active access
+tokens for those users.
+
+Furthermore, the `/login` endpoint is longer available for application services
```
AndrewFerr:
```suggestion
Furthermore, the `/login` endpoint is no longer available for application services
```
right?

tulir:
MSC4326's FCP finished:

@mscbot resolve checklist not complete again

now just waiting for 1 more tick here and confirmation from @richvdh that the backwards compatibility story is sufficient

proposals/4190-as-device-management.md
```diff
@@ -29,7 +28,12 @@ create and delete devices for their users without relying on the `/login` and
 As all changes here only apply to application services, guest access is not
 relevant.
 
-### **`PUT /_matrix/client/v3/devices/{deviceId}`**
+To opt into this new behavior, the appservice registration has a new
+`device_management: true` flag. However, if the homeserver has switched to
+OAuth2, it MUST treat all appservices as having the flag enabled, as the old
+login methods will not work in any case.
```
richvdh: I still don't really understand why an opt-in is required.

It looks like we're agreed that the flag is redundant when the HS has switched to OAuth2, so let's consider an HS that has not yet switched. It seems like in that case, the flag has the following effects:

 * Logins with the `m.login.application_service` type return `M_APPSERVICE_LOGIN_UNSUPPORTED` error code
   * Why is this required? Why not have `m.login.application_service` continue to be supported unless the HS has switched to OAuth2?
 * Appservices MUST call ` /_matrix/client/v3/register` with `inhibit_login=true`.
   *  Again, why is this a necessary requirement? Why not make `inhibit_login=false` continue to work unless the HS has switched to OAuth2?

In short, having the opt-in means that we have to spec and support three different modes of operation (legacy auth without opt-in, legacy auth with opt-in, OAuth2), whereas I feel like we only really need two (legacy auth, OAuth2).

Is the argument just "I want to be able to test that my appservice will work correctly on a homeserver that has switched to OAuth2"? That feels to me like a thing that can and should be handled via an implementation-specific setting, rather than complicating the spec. I suspect we'll be stuck with the legacy APIs for quite a while, and hence ending up needing to support the opt-in flag for a while.

tulir: Good point, I've removed the opt-in entirely in https://github.com/matrix-org/matrix-spec-proposals/pull/4190/commits/ab1060479fafef597384fee8fa3c1bff2f5031bf + implemented in https://github.com/element-hq/synapse/pull/19031

All other changes except deleting /login and requiring inhibit_login on /register are available for all appservices, those 2 breaking changes are reserved to OAuth2 homeservers or implementation-specific opt-ins (synapse will presumably just keep `io.element.msc4190` in appservice registrations)

richvdh (approved):
LGTM

richvdh:
@mscbot resolve lack of clarity over backwards compatibility

cloudrac3r:
I didn't like your breaking changes on this one

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2267

turt2live:
Merged 🎉

PR #4191, opened 2024-09-12
MSC4191: Account management for OAuth 2.0 API

sandhose:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/quenting/account-deeplink/proposals/4191-account-deeplink.md)

Implementions:

- Server
  - Synapse via [matrix-authentication-service](https://github.com/element-hq/matrix-authentication-service)
    - [x] Exposes `account_management_uri` and `account_management_actions_supported` values in metadata. e.g. https://matrix-client.matrix.org/_matrix/client/v1/auth_metadata
    - [x] Provides a UX at the advertised URIs with support for the `action` and `device_id` parameters
- Clients
  - [x] Element Web
    - Supported actions include: `org.matrix.session_end`
  - [x] Element X
    - Supported actions include: `org.matrix.profile`, `org.matrix.sessions_list`

To do:

- [x] Update references to MSC3861 as it has been accepted
- [x] Add clarification about being the replacement for delete devices API endpoints when using OAuth 2.0 API
- [x] Check for consistency with MSC3824

---

I am a Software Engineer at Element. This proposal was written and published as an Element employee.

----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4191#issuecomment-3330262419)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4191#issuecomment-3330263076)

### DIFF
--- a/proposals/4191-account-deeplink.md
+++ b/proposals/4191-account-deeplink.md
@@ -38,7 +38,7 @@ The account management URL (as provided above) may accept the following addition
 
 For example, if a user wishes to sign out a session for the device `ABCDEFGH` where the advertised
 `account_management_uri` was `https://account.example.com/myaccount` the client could open a link to
-`https://account.example.com/myaccount?action=org.matrix.session_end&device_id=ABCDEFGH`.
+`https://account.example.com/myaccount?action=org.matrix.device_delete&device_id=ABCDEFGH`.
 
 Not all actions need to be supported by the account management URL, and the client should only use the actions
 advertised in the `account_management_actions_supported` array from above.
@@ -82,7 +82,7 @@ The reason given in the spec is:
 > **WARNING**: Since this endpoint uses User-Interactive Authentication, it cannot be used when the access token
 > was obtained via the OAuth 2.0 API.
 
-As an alternative to the proposed `org.matrix.session_end` action the above device deletion endpoints could be instead
+As an alternative to the proposed `org.matrix.device_delete` action the above device deletion endpoints could be instead
 be used *if* the [User-interactive API](https://spec.matrix.org/v1.15/client-server-api/#user-interactive-api-in-the-rest-api)
 was compatible with the OAuth 2.0 API.
 
@@ -104,17 +104,17 @@ The metadata could be advertised in a metadata endpoint separate to the
 
 ## Security considerations
 
-For the `org.matrix.session_end` action the
+For the `org.matrix.device_delete` action the
 [security considerations](https://spec.matrix.org/v1.15/client-server-api/#security-considerations-6) that are already
 stated for device management apply.
 
-Consequently for the destructive `org.matrix.session_end` and `org.matrix.account_deactivate` actions the server SHOULD
+Consequently for the destructive `org.matrix.device_delete` and `org.matrix.account_deactivate` actions the server SHOULD
 require re-authentication (or authentication through an additional factor) before allowing the user to complete the
 action.
 
 Because the format of these links are public and the device IDs are visible to other Matrix users it is trivial for an
 attacker to craft a "malicious" link that targets a destructive action. To mitigate this risk, for the destructive
-actions (`org.matrix.session_end` and `org.matrix.account_deactivate`) the server MUST inform the user of what the
+actions (`org.matrix.device_delete` and `org.matrix.account_deactivate`) the server MUST inform the user of what the
 action is prior to it being executed (in addition to any re-authentication as above).
 
 ## Unstable prefixes


proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,66 @@
...
+- `org.matrix.account_deactivate` - The user wishes to deactivate their account.
+- `org.matrix.cross_signing_reset` - The user wishes to reset their cross-signing keys.
+
+Subsequent MSCs may extend this list.
+
+### Account management URL parameters
+
+The account management URL may accept the following additional query parameters:
+
+- `id_token_hint` - An ID Token that was previously issued to the client; the issuer uses it as a hint for which user is requesting to manage their account.
```
t3chguy, line 37: This doesn't work for OIDC-Aware (Compat) sessions, unlike MSC4198

Just encountered the issue where I had Account_A logged into Element, but Account_B logged into MAS. Tried to manage a session to log it out, and got an awful error and then saw most of my sessions were missing. Took me a few minutes to figure out the accounts were not the same. Informing the IdP which account the user expects is crucial IMO.

image

tonkku107: That parameter also won't be very helpful with OIDC-native (or should we call them Next Gen Auth -native now?) clients that don't request the `openid` scope since that's no longer necessary

hughns: > OIDC-native (or should we call them Next Gen Auth -native now?)

I suggest calling them OAuth 2.0 native clients.

hughns: I was discussing with @sandhose and thoughts are that MSC4198 would be a better way to do this. So, I am going to remove `id_token_hint` from this MSC.

tonkku107: Regarding the comment on MSC4198:
> Propose that the `login_hint` can also be added to the MSC4191 account management URL query parameters

Wouldn't the best place for that be in this proposal with a dependency to MSC4198?

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,110 @@
...
+added in v1.15 of the spec.
+
+The current spec makes reference to the "homeserver's web UI" in several places without defining how that is discovered
+nor what you can use it for.
+
+This proposal defines how that discovery happens and a way to deep-link to the account management capabilities of the
+homeserver to allow the user to complete the account management operations in a browser independently of the client.
+
+In doing so, this proposal also resolves the issue that the is currently no specced way to delete devices using the
+OAuth 2.0 API (becuase the API endpoints to delete devices are not supported when using the OAuth 2.0 API).
```
hughns:
```suggestion
OAuth 2.0 API (because the API endpoints to delete devices are not supported when using the OAuth 2.0 API).
```

turt2live:
@mscbot fcp merge
@mscbot concern Pending preliminary approval from oauth-ext-review to use the new names - https://mailarchive.ietf.org/arch/msg/oauth-ext-review/6RQXusbqFTy-XZCGwLSW-RIa9Dg/

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,110 @@
...
+This is why the metadata fields are not prefixed with `org.matrix.`, but the actions themselves are.
+
+### Possible actions
+
+The following actions are defined by this MSC:
+
+- `org.matrix.profile` - The user wishes to view/edit their profile (name, avatar, contact details).
+- `org.matrix.sessions_list` - The user wishes to view a list of their sessions.
+- `org.matrix.session_view` - The user wishes to view the details of a specific session.
+- `org.matrix.session_end` - The user wishes to end/logout a specific session.
```
clokep: Why session and not device to match current terminology?

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,110 @@
...
+
+The account management URL (as provided above) may accept the following additional query parameters:
+
+- `action` - Can be used to indicate the action that the user wishes to take, as defined above.
+- `device_id` - This can be used to identify a particular session for `org.matrix.session_view` and
+  `org.matrix.session_end`. This is the Matrix device ID.
+
+For example, if a user wishes to sign out a session for the device `ABCDEFGH` where the advertised
+`account_management_uri` was `https://account.example.com/myaccount` the client could open a link to
+`https://account.example.com/myaccount?action=org.matrix.session_end&device_id=ABCDEFGH`.
```
clokep: Maybe it is covered in security sections already but this feels like potential for phishing or malicious users creating links? Device IDs, the homeserver URL are public, so one can easily craft links to have someone log out devices.

The security section mentions needing to reauthenticate, but maybe also some user action needs to be required to avoid a user authenticating without really knowing why?

hughns: I agree this should be explicit and is in 0b27bdd2124cdc1e1fe5f7db9d298f103246b7aa

The implementation of this in MAS does indeed require confirmation of the action 👍

proposals/4191-account-deeplink.md
hughns: I believe the original rationale for this is that OAuth 2.0 refers to sessions, whereas devices is a Matrix concept.

@sandhose can you provide any other context on this?

hughns: @sandhose has said that originally these values didn't have a `org.matrix` prefix and it felt right to use the OAuth terminology of sessions.

However, now that these values are prefixed it does seem entirely reasonable to revisit them and give them names that make sense for Matrix.

How about:

- `org.matrix.devices_list`
- `org.matrix.device_view`
- `org.matrix.device_delete` (`delete` doesn't feel quite right, but I can't think of anything better right now)

Given that these are just value renames rather than anything semantic, could they be adjusted later in the spec process rather than requiring the implementations to be updated prior to FCP?

clokep: That would be fine, yes.

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,115 @@
...
+This proposal builds on the [OAuth 2.0 API](https://spec.matrix.org/v1.15/client-server-api/#oauth-20-api) that was
+added in v1.15 of the spec.
+
+The current spec makes reference to the "homeserver's web UI" in several places without defining how that is discovered
+nor what you can use it for.
+
+This proposal defines how that discovery happens and a way to deep-link to the account management capabilities of the
+homeserver to allow the user to complete the account management operations in a browser independently of the client.
+
+In doing so, this proposal also resolves the issue that the is currently no specced way to delete devices using the
```
uhoreg: nit:

```suggestion
In doing so, this proposal also resolves the issue that there is currently no specced way to delete devices using the
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,115 @@
...
+[server metadata discovery](https://spec.matrix.org/v1.15/client-server-api/#server-metadata-discovery):
+
+- `account_management_uri`: the URL where the user is able to access the account management capabilities of the
+   homeserver. This is what is currently referred to as the "homeserver's web UI"
+- `account_management_actions_supported`: a JSON array of actions that the account management URL supports
+
+Note that the intent of this proposal is for these values to potentially end up in a OpenID Connect specification, or at
+least formally registered in the
+[IANA Registry for Server Metadata](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#authorization-server-metadata).
+This is why the metadata fields are not prefixed with `org.matrix.`, but the actions themselves are.
```
uhoreg, line 29: I assume this is also the reason why the actions are using `org.matrix.` rather than the `m.` that is normally used in the Matrix spec?

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,115 @@
...
+- `org.matrix.sessions_list` - The user wishes to view a list of their sessions.
+- `org.matrix.session_view` - The user wishes to view the details of a specific session.
+- `org.matrix.session_end` - The user wishes to end/logout a specific session.
+- `org.matrix.account_deactivate` - The user wishes to deactivate their account.
+
+Subsequent MSCs may extend this list.
+
+### Account management URL parameters
+
+The account management URL (as provided above) may accept the following additional query parameters:
```
uhoreg, line 33: Do clients need a way to determine whether the management URL does accept the query parameters?

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,115 @@
...
+`https://account.example.com/myaccount?action=org.matrix.session_end&device_id=ABCDEFGH`.
+
+Not all actions need to be supported by the account management URL, and the client should only use the actions
+advertised in the `account_management_actions_supported` array.
+
+## Alternatives
+
+### Add UIA support to OAuth 2.0 API so that the existing delete devices endpoints can be used
+
+n.b. This isn't a complete alternative to this full MSC and is instead just considers an alternative way for device
```
uhoreg: nit:

```suggestion
n.b. This isn't a complete alternative to this full MSC and instead just considers an alternative way for device
```

proposals/4191-account-deeplink.md
hughns: It already does for the `action` (via the `account_management_actions_supported`). I've added a note to say that the server *should* support `device_id` for `org.matrix.device_view` and `org.matrix.device_delete`

proposals/4191-account-deeplink.md
hughns: I've updated the values in 9e1fbf5ac22720efcf65fc8e60c9128ea389b08d and left this thread unresolved to reflect that the implementation in MAS is using the non-stable values.

proposals/4191-account-deeplink.md
hughns: Yes. I've added a further comment about this in https://github.com/matrix-org/matrix-spec-proposals/commit/9e1fbf5ac22720efcf65fc8e60c9128ea389b08d#diff-a71172f1154cf40383ff506bd6a10e6bd97689fd5d60af1a5ad2e2cd9b6676c6R61-R63

proposals/4191-account-deeplink.md
uhoreg: Ah, OK.  I missed that the actions given in there were the actions that were accepted by the query param.  I was thinking that it was just a generic "The system supports doing these things somehow".

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Add UIA support to OAuth 2.0 API so that the existing delete devices endpoints can be used
+
+n.b. This isn't a complete alternative to this full MSC and instead just considers an alternative way for device
+deletion to happen.
+
+The following endpoints are currently not supported when using the OAuth 2.0 API:
+
+- `POST /_matrix/client/v3/delete_devices`
+- `DELETE /_matrix/client/v3/devices/{deviceId}`
```
richvdh: links are always helpful

```suggestion
- [`POST /_matrix/client/v3/delete_devices`](https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3delete_devices)
- [`DELETE /_matrix/client/v3/devices/{deviceId}`](https://spec.matrix.org/v1.16/client-server-api/#delete_matrixclientv3devicesdeviceid)
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Possible actions
+
+The following account management actions are defined by this MSC:
+
+- `org.matrix.profile` - The user wishes to view/edit their profile (name, avatar, contact details).
+- `org.matrix.devices_list` - The user wishes to view a list of their devices.
+- `org.matrix.device_view` - The user wishes to view the details of a specific device.
+- `org.matrix.device_delete` - The user wishes to delete/logout a specific device.
+- `org.matrix.account_deactivate` - The user wishes to deactivate their account.
```
Johennes, line 57: Since https://github.com/matrix-org/matrix-spec-proposals/pull/4312 has already landed and is being spec'ed in https://github.com/matrix-org/matrix-spec/pull/2234, we might want to add the contained action type here.

```suggestion
- `org.matrix.account_deactivate` - The user wishes to deactivate their account.
- `org.matrix.cross_signing_reset` - The user wishes to reset their cross-signing identity.
```

hughns: Yes, this makes sense. I will add a reference to https://github.com/matrix-org/matrix-spec-proposals/blob/ea0aef0aa32ee979b4db14bb869938a597777346/proposals/4312-x-signing-reset-with-nextgen-auth.md?plain=1#L55-L58 too so that we don't forget the detail when writing the spec PR.

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,133 @@
...
+
+The account management URL (as provided above) may accept the following additional query parameters:
+
+- `action` - Can be used to indicate the action that the user wishes to take, as defined below.
+- `device_id` - This can be used to identify a particular session for `org.matrix.device_view` and
+  `org.matrix.device_delete`. This is the Matrix device ID.
+
+For example, if a user wishes to sign out a session for the device `ABCDEFGH` where the advertised
+`account_management_uri` was `https://account.example.com/myaccount` the client could open a link to
+`https://account.example.com/myaccount?action=org.matrix.session_end&device_id=ABCDEFGH`.
```
tonkku107:
```suggestion
`https://account.example.com/myaccount?action=org.matrix.device_delete&device_id=ABCDEFGH`.
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,133 @@
...
+
+- [`POST /_matrix/client/v3/delete_devices`](https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3delete_devices)
+- [`DELETE /_matrix/client/v3/devices/{deviceId}`](https://spec.matrix.org/v1.16/client-server-api/#delete_matrixclientv3devicesdeviceid)
+
+The reason given in the spec is:
+
+> **WARNING**: Since this endpoint uses User-Interactive Authentication, it cannot be used when the access token
+> was obtained via the OAuth 2.0 API.
+
+As an alternative to the proposed `org.matrix.session_end` action the above device deletion endpoints could be instead
```
tonkku107:
```suggestion
As an alternative to the proposed `org.matrix.device_delete` action the above device deletion endpoints could be instead
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,133 @@
...
+instead.
+
+### Advertise the account management capabilities in a C-S API endpoint
+
+The metadata could be advertised in a metadata endpoint separate to the
+[server metadata discovery](https://spec.matrix.org/v1.15/client-server-api/#server-metadata-discovery) mechanism.
+
+## Security considerations
+
+For the `org.matrix.session_end` action the
```
tonkku107:
```suggestion
For the `org.matrix.device_delete` action the
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,133 @@
...
+[security considerations](https://spec.matrix.org/v1.15/client-server-api/#security-considerations-6) that are already
+stated for device management apply.
+
+Consequently for the destructive `org.matrix.session_end` and `org.matrix.account_deactivate` actions the server SHOULD
```
tonkku107:
```suggestion
Consequently for the destructive `org.matrix.device_delete` and `org.matrix.account_deactivate` actions the server SHOULD
```

proposals/4191-account-deeplink.md
```diff
@@ -0,0 +1,133 @@
...
+require re-authentication (or authentication through an additional factor) before allowing the user to complete the
+action.
+
+Because the format of these links are public and the device IDs are visible to other Matrix users it is trivial for an
+attacker to craft a "malicious" link that targets a destructive action. To mitigate this risk, for the destructive
+actions (`org.matrix.session_end` and `org.matrix.account_deactivate`) the server MUST inform the user of what the
```
tonkku107:
```suggestion
actions (`org.matrix.device_delete` and `org.matrix.account_deactivate`) the server MUST inform the user of what the
```

tonkku107 (changes requested):
`org.matrix.session_end` was renamed to `org.matrix.device_delete` but the rest of the proposal didn't follow

turt2live:
Per recent emails sent to oauth-ext-review on our thread, it seems possible to allow this MSC to go through without formal registration. When we get to the spec PR, we should verbosely describe the fields so we can point to them in a more complete registration application. 

@mscbot resolve Pending preliminary approval from oauth-ext-review to use the new names - https://mailarchive.ietf.org/arch/msg/oauth-ext-review/6RQXusbqFTy-XZCGwLSW-RIa9Dg/

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2270

turt2live:
Merged 🎉

PR #4155, opened 2024-06-13
MSC4155: Invite filtering

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/invite-filtering/proposals/4155-invite-filtering.md)

---

In line with https://github.com/matrix-org/matrix-spec/issues/1700, the following disclosure applies:

I am a Systems Architect at [gematik](https://www.gematik.de/), Software Engineer at [Unomed](https://www.unomed.ch/), Matrix community member and former Element employee. This proposal was written and published with my gematik hat on.

----

SCT stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4155#issuecomment-2920520258)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4155#issuecomment-2920521338)

PR #4133, opened 2024-04-19
MSC4133: Extending User Profile API with Key:Value Pairs

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/main/proposals/4133-extended-profiles.md)

Signed-off-by: Tom Foster <tom@tcpip.uk>

---

Known Implementations:
- Clients:
   - Complete: element-hq/matrix-react-sdk#20
   - Complete: tulir/gomuks#574
- Servers:
   - Complete: element-hq/synapse#17488
   - Complete: girlbossceo/conduwuit#571
- SDKs:
   - Complete: matrix-org/matrix-js-sdk#4391
   - Complete: mautrix/go#337

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4133#issuecomment-2569794477)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4133#issuecomment-2571377851)

PR #4192, opened 2024-09-13
Comparison of proposals for ignoring invites

Johennes:
This aims to aid progression on the topic of ignoring invites by comparing the various existing proposals. I've decided to submit this as a pull request rather than an issue because this allows for easy commenting. Obviously, there is no intention to merge this pull request ever.

[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/ignoring-invites-meta/proposals/xxxx-ignoring-invites.md)

---

### DIFF
--- a/proposals/xxxx-ignoring-invites.md
+++ b/proposals/xxxx-ignoring-invites.md
@@ -1,0 +1,285 @@
+# Comparison of proposals for ignoring invites
+
+Matrix supports ignoring specific users via the eponymous [module] and the
+`m.ignored_user_list` account data event.
+
+```json5
+{
+  "type": "m.ignored_user_list"
+  "content": {
+    "ignored_users": {
+      "@someone:example.org": {}
+    }
+  },
+}
+```
+
+Servers MUST NOT forward invites from ignored users to clients. For the
+purposes of ignoring invites, this feature is, however, quite limiting
+because:
+
+* Ignoring users also hides their events in existing rooms which might not
+  be desirable
+* There's no support for globs or wildcards
+* The module only supports block-list but no allow-list semantics
+* Users cannot review or unignore specific ignored invites
+* There's no built-in concept of sharing ignore lists between users
+
+Several attempts at designing a scheme for ignoring invites specifically
+have been made so far. This documents provides a brief comparison of the
+different proposals and serves as an entry point for further research.
+
+
+## [MSC2270] – Proposal for ignoring invites
+
+A new account data event `m.ignored_room_list` is introduced. The event
+has the exact same semantics as `m.ignored_user_list` but tracks room IDs
+rather than user IDs.
+
+```json5
+{
+  "type": "m.ignored_room_list"
+  "content": {
+    "ignored_rooms": {
+      "!something:example.org": {}
+    }
+  },
+}
+```
+
+A key drawback of this scheme is that it's not possible to blanketly ignore
+invites from specific users or servers. Spammers could, therefore, send
+invites repeatedly.
+
+Additionally, the ignore list is subject to the 64 KiB event size limit.
+
+
+## [MSC3840] – Ignore invites
+
+A new account data event `m.ignored_invites` is introduced. The event has
+separate fields for ignoring invites:
+
+- From specific users
+- From specific servers
+- Into specific rooms
+- Into specific rooms by specific users
+
+```json5
+{
+  "type": "m.ignored_invites"
+  "content": {
+    "ignored_user_ids": {
+      "@someone:example.org": {
+        "ts": 1726138169160,
+        "reason": "..."
+      },
+    },
+    "ignored_servers": {
+      "example.org": {
+        "ts": 1726138169160,
+        "reason": "..."
+      },
+    },
+    "ignored_room_ids": {
+      "!something:example.org": {
+        "ts": 1726138169160,
+        "reason": "..."
+      },
+    },
+    "ignored_invites": {
+      "!something:example.org": {
+        "@someone:example.org": {
+          "ts": 1726138169160,
+          "reason": "..."
+        },
+      },
+    },
+  },
+}
+```
+
+A key improvement over [MSC2270] is the ability to ignore entire users
+and servers.
+
+Similar to [MSC2270], however, the ignore configuration is still subject
+to the 64 KiB event size limit.
+
+Contrary to `m.ignored_user_list` and [MSC2270], `m.ignored_invites` is
+handled exclusively on the client side. This allows users to review and
+unignore ignored invites but also means that ignored invites will be
+repeated in `/sync` which can cause performance issues over time.
+
+
+## [MSC4155]: Invite filtering
+
+A new account data event `m.invite_permission_config` is introduced to
+maintain a configuration for invite filtering with:
+
+- A global default to allow / block all invites
+- Per-user & per-server exceptions that override the default
+
+```json5
+{
+  "type": "m.invite_permission_config",
+  "content": {
+    "default": "allow | block",
+    "user_exceptions": {
+      "@someone:example.org": {},
+      ...
+    },
+    "server_exceptions": {
+      "example.org": {},
+      ...
+    }
+  }
+}
+```
+
+Similar to [MSC3840], this scheme allows to ignore entire users and servers.
+A key advantage though is the ability to create both block-list and allow-list
+configurations.
+
+Unlike [MSC3840], processing the ignore configuration can happen on either the
+client or the server.
+
+Similar to [MSC2270] and [MSC3840], the ignore configuration is subject to
+the 64 KiB event size limit.
+
+
+## [MSC3847] – Ignoring invites with policy rooms
+
+The invite filtering configuration is tracked in [moderation policy lists].
+These are already part of the spec and allow formulating glob rules against
+users, rooms and servers via state events in a policy room.
+
+```json5
+{
+  "type": "m.policy.rule.user | m.policy.rule.room | m.policy.rule.server",
+  "state_key": "...arbitrary...",
+  "content": {
+    "entity": "@someone:example.org | !abcdefg:example.org | example.org | *example.org",
+    "reason": "...",
+    "recommendation": "m.ban"
+  }
+}
+```
+
+Users can hook into policy rooms for invite filtering via a new account data
+event `m.policies` which defines:
+
+- A room into which the user will write their own custom policy rules (`target`)
+- A list of policy rooms to be executed against invites (`sources`). This
+  should generally include `target` but may also list other rooms such as
+  policy rooms recommended by the home server.
+
+```json5
+{
+  "type": "m.policies",
+  "content": {
+    "m.ignore.invites": {
+      "target": "!my-policies:example.org",
+      "sources": [
+        "!my-policies:example.org",
+        "!bobs-policies:example.org"
+      ]
+    }
+  }
+}
+```
+
+Unlike the previous proposals, the ignore configuration is not subject to
+the 64 KiB event size limit because each policy rule is stored in a separate
+event.
+
+Additionally, ignore configurations can easily be shared and reused across
+different users.
+
+Another key advantage is the ability to match users, rooms and servers against
+glob expressions. This could probably be added into the other proposals as well,
+however.
+
+Similar to [MSC3840], the ignore configuration is processed exclusively on the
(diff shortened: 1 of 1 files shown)


turt2live:
for lack of a better system, I'm tracking this as a proposal.

proposals/xxxx-ignoring-invites.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+Unlike [MSC3847], the ignore configuration is subject to the 64 KiB event size
+limit because all rules are stored in a single event.
+
+Similar to [MSC3847], however, glob expressions as well as block-list and
+allow-list configurations are supported.
+
+
+## Summary
```
Gnuxie, line 260: Something that's kind of missing from the review is that most of these proposals focus on how a user can filter invitations from their client to filter them from one device or or from `/sync` across their devices. The focus is on what an individual user can do, for their own experience and no one else's. There obviously needs to be a way to ignore invitations, but advanced filtering such as globs or rules could be missing the point, if Matrix users are having to set these advanced systems up, and everyone has to do that individually, then that's no good. The bar is too high, and that's a lot of duplicated work that everyone is doing. 


The reason for that is because there's no dedicated moderation team in Matrix. If this was a normal social network then there would be a team that would be fed information for each time a user rejects an invitation and from that they would be able to generate the metrics to be able to create these complex rules. That would then be applied system wide for all users of the network.

So how do we get an equivalent to that in Matrix that works?

Gnuxie: So to an extent, homeservers can already collect data whenever one of their resident users rejects an invitation. And tools can in theory be developed to protect resident users from those invitations. But there's a problem with this because the target of the invitations on single user homeservers and small homeservers will get exposed to the invitations before there's enough data to protect everyone on the server who is a target.

Gnuxie: The other route is to skip the homeserver entirely, and have users publish data about the invitations that they have rejected directly (but again, there needs to be consideration about how not to leak unnecessary information and at the minimum identifiers will have to be hashed). The good news there is that the upcoming profiles [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133) actually provides an avenue for policy lists to be discoverable.

Johennes: Thanks for the input, these are good thoughts.

My focus writing this up was definitely on the end user taking action. gematik is currently using [MSC4155](https://github.com/matrix-org/matrix-spec-proposals/pull/4155) in a private federation to let e.g. doctor's decide what patients, other doctors, etc. are allowed to contact them. The other proposals mostly focus on the invitee taking action, too, though.

Since all of the proposals depend on account data in some form, I believe they would conceptually support the homeserver injecting a certain base configuration – just like you're getting default push rules from the server. We'd just have to put modifying the account data event an API.

In order not to let the scope explode, I'm wondering if we should make a separation here. This proposal could focus on providing capabilities for users to set up personalized invite filtering and for servers to inject default configurations. The topic of sharing invite spam information among servers could then be handled separately in other proposals?

proposals/4192-ignoring-invites.md
```diff
@@ -0,0 +1,285 @@
...
+    }
+  },
+}
+```
+
+A key drawback of this scheme is that it's not possible to blanketly ignore
+invites from specific users or servers. Spammers could, therefore, send
+invites repeatedly.
+
+Additionally, the ignore list is subject to the 64 KiB event size limit.
```
kegsay, line 54: I don't think this is true. Account data events != room events, which has the 64KB limit enforced at a federation level. Because account data isn't federated, I think there is no limit. If there was, this would prove problematic given DMs / ignore users today are account data events.

proposals/4192-ignoring-invites.md
```diff
@@ -0,0 +1,285 @@
...
+client which comes with the same advantages and disadvantages.
+
+Finally, when combined with [MSC4150], this scheme also allows for allow-list
+configurations.
+
+
+## [MSC3659] – Invite Rules
+
+A system akin to push rules is defined. Rules are stored as a flat list
+in a new account data event `m.invite_rules`.
```
kegsay, line 211: Given how complex push rules are to implement correctly, I have zero desire to reproduce this here. The complexity issue should be mentioned as a negative.

proposals/4192-ignoring-invites.md
```diff
@@ -0,0 +1,285 @@
...
+allow-list configurations are supported.
+
+
+## Summary
+
+The table below compares the proposals in terms of different features. It's
+worth noting that some features such as glob matching could easily be extended
+to all proposals.
+
+|                      | [MSC2270] | [MSC3840] | [MSC4155]     | [MSC3847]           | [MSC3659] |
```
kegsay, line 266: I never really thought about expressing this as moderation policy lists but I actually love it. This is 100% in the spirit of Matrix as a protocol, where we can share our views of who to ignore _and consciously choose to opt-in_. 

There is the question over which room should your own personal blocklists go, to which profiles as rooms would be the obvious candidate if it ever landed.

proposals/4192-ignoring-invites.md
```diff
@@ -0,0 +1,285 @@
...
+configurations.
+
+Unlike [MSC3840], processing the ignore configuration can happen on either the
+client or the server.
+
+Similar to [MSC2270] and [MSC3840], the ignore configuration is subject to
+the 64 KiB event size limit.
+
+
+## [MSC3847] – Ignoring invites with policy rooms
```
kegsay, line 149: I never really thought about expressing this as moderation policy lists but I actually love it. This is 100% in the spirit of Matrix as a protocol, where we can share our views of who to ignore _and consciously choose to opt-in_. 

There is the question over which room should your own personal blocklists go, to which profiles as rooms would be the obvious candidate if it ever landed.

proposals/xxxx-ignoring-invites.md
kegsay: >  I believe they would conceptually support the homeserver injecting a certain base configuration

If your server is a corporation with their own rules to follow, this makes total sense. However, I somewhat strongly feel that this must not be applied universally for all servers (like push rules are today). A core part of Matrix is that you have the agency to _choose_ which filter bubbles you're taking part in, and it should be visible to the end user.

> This proposal could focus on providing capabilities for users to set up personalized invite filtering and for servers to inject default configurations. The topic of sharing invite spam information among servers could then be handled separately in other proposals?

I don't think this is a good separation of concerns because well, they concern the same thing. Say we do this and decide to implement MSC3659 – Invite Rules - for whatever reason. Months later, we want to share the rules with others. Oh no. Now we have two similar but not quite the same systems dealing with allow/blocklists for different categories of things, when if we had a holistic view from the start we may not have made the same decision.

proposals/4192-ignoring-invites.md
Johennes: You're right. Looks like I (and some of the proposals) got that wrong. Have removed all the references to event size limits.

proposals/4192-ignoring-invites.md
Johennes: How would you say the complexity compares to the policy list proposal above? My hunch is the policy lists could appear less complex only because they currently have fewer matching and action constructs. If the invite rules proposal would be stripped down to be on par feature-wise, would you still consider it a lot more complex to implement?

I'm not necessarily in favor of this proposal. Just trying to make sure I'm not discounting it unfairly.

proposals/xxxx-ignoring-invites.md
Johennes: Point taken. Maybe we should clarify what exactly we want to solve for here:

* Let users manage who can invite them (this is what I originally had in mind when writing this up and what the existing proposals largely focus on)
* Let users share their invite configurations with each other?
* Let users share their invite configurations with their server?
* Let servers share default or recommended invite configurations with their users?
* Let servers share invite configurations with each other?
* Maybe others?

proposals/4192-ignoring-invites.md
joshqou: @kegsay You can read the full MSC3659 spec for more details. It's just a list of rules which are evaluated in series. I've implemented the spec myself and the implementation is not that complicated.

Even if it comparatively complicated, compared to `m.ignore_users`, approving and implementing it would give Matrix the critical user safety tools it currently lacks. If we can have spaces and threads, both vastly more complicated, surely we can be able to properly block dms from harassers. If you have any individual critiques about specific rule types then I'd be open to accepting changes

proposals/xxxx-ignoring-invites.md
Gnuxie: > If your server is a corporation with their own rules to follow, this makes total sense.

I'm not so sure, it should make sense everywhere. There should be defaults because a majority of users will never play around with the settings much. That does not mean that power users shouldn't be able to customize or deviate from the default though.

proposals/4192-ignoring-invites.md
FSG-Cat: Cat also likes the policy based solution and well Cat would say that the personal blocklist should be written to a dedicated MSC3784 policy room but said room can be mapped to via whatever profile system you prefer. This comes with the benefit of maintaining backwards compatibility with tooling that is only focused on policy manipulation.

Plus it allows trivially having multiple personal lists. This allows stuff like private and public opinions and friends only. 

Since i might not want to share that i dislike X but i might be fine with sharing i dislike Y. Yes Policy based ofc has privacy implications but its also very matrix to do this stuff in this way.

proposals/4192-ignoring-invites.md
turt2live, line 1: With a combination of SCT and T&S hat, I think there are two separate problems which might help lead to a path forward:

1. Preventing an invite from reaching a user.
2. Dealing with unwanted invites once they've already reached a user.

For the first, prototyping the desirable constraints with MSC4155 I feel is best - the object is extensible, and can be experimented with quickly. Later, we can add fields to incorporate policy lists (a la MSC3847) to make configurations sharable/even more extensible. The "plan" would be to land MSC4155, expand upon it via MSCs, then land MSC3847 with modifications to account for MSC4155+extensions behaviour. We would also ideally have an extension MSC to introduce the difference between "ignore" and "block", which I might write up to clarify the concept in Matrix generally. 

For the second, I think MSC2270 is the most generic as it doesn't just deal with invites: it deals with rooms more generally. This can enable a feature where someone can be nominally joined to a room, but otherwise elect to not see contents. For example, as an out-of-hours measure: events continue flowing through to the room, but the user doesn't see them until 8am on Monday. (We would have to refine the way ignoring works through another MSC to fully make that possible, though).

MSC3840 I think accomplishes features from both columns, but I'm not sure we actually want to combine these two features as it further leads to confusion about what "ignore" can mean. I'd recommend closing this one in favour of other MSCs, though taking the semantics and considering them for extensions on top of the other MSCs. 

MSC3659 is neat, but I think might be a bit complicated to drive/get wired up to a user interface. Buckets of conditions (like MSC4155) are easier to represent to users than sequences.

MSC4264 I'm considering as adjacent to the problem space for now, especially because it needs a lot more thought and consideration for how invites overall should work. The effort in this comparison is more, in my opinion, to reduce harm from the existing system's mechanics over designing new mechanics.

turt2live (commented):
Thanks for putting this together - it really helps highlight what each MSC does, and that there's interest in solving this particular problem space. 

To summarize my comment (which is mostly an opinion piece):

* MSC4155 should be put on the faster track for FCP.
* MSC3847 should stay on the FCP track, but intentionally put behind extension MSCs which don't yet exist for MSC4155. It would stay in this state until we're reasonably comfortable with MSC4155+extensions's behaviours, then we can codify them into policy lists.
* MSC2270 should also be put on the faster track for FCP.
* MSC3840 should be closed in favour of MSC2270+MSC4155.
* MSC3659 should be closed in favour of MSC4155.
* MSC4264 should be considered external to this stream of work.
* A new MSC gets written to clarify the difference between block and ignore, and should be put on the faster track for FCP. (I plan to do this)
* This comparison MSC can be closed in favour of executing the above plan if agreeable.

proposals/4192-ignoring-invites.md
turt2live: [MSC4283](https://github.com/matrix-org/matrix-spec-proposals/pull/4283) now exists to distinguish between ignore and block.

clokep: From a quick read through @turt2live's plan sounds reasonable!

PR #4194, opened 2024-09-15
MSC0000: User Redaction draft

Gnuxie:
[rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/user-redaction-api/proposals/0000-user-redaction.md)

special thanks to @tulir for discussing this with me and providing inspiration from [meowlnir](https://github.com/maunium/meowlnir).

### DIFF
--- a/proposals/0000-user-redaction.md
+++ b/proposals/0000-user-redaction.md
@@ -1,0 +1,138 @@
+# MSC0000: User redaction
+
+Targeted attacks are a problem in Matrix. Typically throwaway users
+are used to target matrix rooms with abuse in the form of multiple
+messages.  The intent of the spammer is to fill the room timeline view
+in clients with abuse.
+
+Typically, a moderator will be pinged by a bystander, and the
+moderator will respond by first banning the throwaway account and then
+issuing redactions for each event the throwaway account sent in their
+client.
+
+Due to the architecture of various Matrix APIs, there are multiple
+issues that can occur in the moderator's flow.
+
+1. Calling `/messages` is required for some clients to accurately get
+   the full list of `event_ids` the throwaway account sent.  This is a
+   compounding problem because homeservers can be slow to respond to a
+   filtered `/messages` request and then the client must still process
+   the response and issue a redaction for each event returned.
+   This is how both Mjolnir and Draupnir source events for redaction.
+
+2. The client must call `/rooms/redact/event` for each target event,
+   requiring the client to make multiple requests, further impeding
+   response time. Homeservers typically rate limit the number of
+   redactions that a client can send to `/redact` evenly with `/send`,
+   which further compounds response time.
+   [MSC2244](https://github.com/matrix-org/matrix-spec-proposals/pull/2244)
+   tried to alleviate part of the problem by allowing the redactions
+   to be sent in one event. But there currently are no endpoints that
+   allow you to create this event[^create-mass-redaction].
+
+3. Soft failure of events sent by the throwaway account after a ban
+   stop them from being visible to the matrix client, and there is no
+   way for a moderator to redact them. This is a huge problem that
+   occurs far more frequently than it should.
+
+[^create-mass-redaction]: I think?
+## Proposal
+
+This proposal introduces a very simple client-server endpoint to
+alleviate all three of the concerns surrounding the moderators
+redaction flow.
+
+### Redacting a users's events
+
+A new endpoint is introduced into the client-server spec.
+
+`POST /_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`
+
+This endpoint redacts the target matrix user's unredacted events in
+reverse chronological order, starting from the most
+chronologically-recent visible event in the room history for the
+requesting user.
+
+This endpoint is blind to the distinction between normal and
+soft-failed events, and will cause redactions to be issued
+for any soft-failed events that match the scope of the
+request.
+
+It is left to an implementation detail whether a `m.room.redaction`
+event is created for each event, or
+[MSC2244](https://github.com/matrix-org/matrix-spec-proposals/pull/2244)
+is employed. We expect that for now, most implementations will
+issue one `m.room.redaction` event for each event under
+the scope of the request.
+
+#### Rate limiting
+
+This API SHOULD be rate limited, but the number `m.room.redaction`
+events emitted SHOULD NOT be the subject of the rate limit.
+But instead the number of events that have been redacted overall.
+This is to cover semantics of MSC2244.
+Servers SHOULD be more liberal in the number of events that
+can be redacted in comparison to rate limits for `/send` when
+the endpoint is being used to redact another user's events.
+Rather than their own events.
+
+#### Query parameters
+
+`limit`: `integer` - The maximum number of events to redact. Default: 25.
+
+#### Respone
+
+`is_more_events`: `boolean` - Whether there are more events outside of
+the scope of the request sent by the user that could be redacted, but
+have not been because of the `limit` query parameter.  If
+`_is_more_events` is true, then the server should expect that the
+client can optionally be called again, to redact more even more events.
+
+`redacted_events`: `integer` - The number of events that have been redacted, including soft failed events.
+
+`soft_failed_events`: `integer` - The number of soft failed events that have been redacted.
+
+```
+200 OK
+Content-Type: application/json
+
+{
+  "is_more_events": false,
+  "redacted_events": 5
+  "soft_failed_events": 1
+}
+```
+
+##### Limited response
+
+Servers may wish to limit the maximum size of the limit that can be specified.
+
+```
+{
+  "errcode": "M_LIMIT_TOO_LARGE",
+  "error": "The limit parameter specified for this request is too large.",
+  "max_limit": 100
+}
+```
+
+## Potential issues
+
+None so far.
+
+## Alternatives
+
+
+## Security considerations
+
+### Use case for self redaction
+
+Implementors should be cautious over the use of this API for self
+redaction. We might omit that use case from the MSC if there are concerns.
+
+## Unstable prefix
+
+Uhm help?
+
+## Dependencies
+
+None


proposals/4194-user-redaction.md
tulir, line 1: Implementation requirements:

* [ ] Server implementing the endpoint
* [ ] Client or bot using the endpoint

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,138 @@
...
+### Redacting a user's events
+
+A new endpoint is introduced into the client-server spec.
+
+`POST /_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`
+
+This endpoint redacts the target matrix user's unredacted events in
+reverse chronological order, starting from the most
+chronologically-recent visible event in the room history for the
+requesting user.
```
tulir, line 54: This should probably be explicit that already-redacted events should be ignored

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,138 @@
...
+## Security considerations
+
+### Use case for self redaction
+
+Implementers should be cautious over the use of this API for self
+redaction. We might omit that use case from the MSC if there are concerns.
+
+## Unstable prefix
+
+Uhm help?
```
tulir, line 134: The endpoint needs an unstable prefix and a /versions flag should be defined, something like this:

```suggestion
Implementations can use `org.matrix.msc4194` as the unstable prefix and as a
flag in the `unstable_features` section of `/versions`:

* `/_matrix/client/unstable/org.matrix.msc4194/rooms/{roomID}/redact/user/{userID}`
  instead of `/_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`

Once the MSC is approved, the `org.matrix.msc4194.stable` unstable feature flag
can be used to advertise support for the stable version of the endpoint, until
the spec release with the endpoint is advertised.
```

(`org.matrix` can be replaced with a different domain too)

Gnuxie: oops, clobbered the change suggestion one minute

proposals/4194-user-redaction.md
Gnuxie: https://github.com/matrix-org/matrix-spec-proposals/pull/4194/commits/1331f86f4a57226dc5bffc00fed6cf698047454a

proposals/4194-user-redaction.md
Gnuxie: Do you know if `M_LIMIT_TOO_LARGE` exists? or if there's something like that already? I know `M_LIMIT_EXCEEDED` exists but that's in reference to rate limits and not a limit to say `/messages`.

Gnuxie: ok maybe i don't need this since the endpoint is rate limited by number of redactions AND we indicate how many events have been redacted. Servers could just say they only redacted `50` when the limit was `100`. I'll fix that.

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,162 @@
...
+[^create-mass-redaction]: I think?
+
+[^often-soft-failure]: https://github.com/matrix-org/synapse/issues/9329
+## Proposal
+
+This proposal introduces a very simple client-server endpoint to
+alleviate all three of the concerns surrounding the moderators
+redaction flow.
+
+### Redacting a user's events
```
MTRNord, line 48: Not sure if this has to be written out but this is expected to be explicitly a batch Action correct? As in the server should when possible try to batch this rather than treating it as Singular calls (especially as this would make a difference for the database for example)

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,162 @@
...
+Servers SHOULD be more liberal in the number of events that
+can be redacted in comparison to rate limits for `/send` when
+the endpoint is being used to redact another user's events.
+Rather than their own events.
+
+#### Query parameters
+
+`limit`: `integer` - The maximum number of events to redact. Default: 25.
+
+#### Response
```
MTRNord, line 89: Should there be a communication of which Events failed to redact? For example due to a db failure. Similar to how s-s api handles failed events where you get a list of Event ids back. Or should the server retry forever to redact the event?

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,162 @@
...
+
+* An endpoint could be developed to expose soft-failed events to
+  clients or appservices such as in
+ [MSC4109](https://github.com/matrix-org/matrix-spec-proposals/pull/4109).
+
+## Security considerations
+
+### Use case for self redaction
+
+Implementers should be cautious over the use of this API for self
```
MTRNord, line 147: How about limiting self redaction to non dag critical Events? So messages and custom state events but not for example powerlevel Events or similar.

proposals/4194-user-redaction.md
tulir: I don't think it should ever fail to redact individual events 🤔

If the database explodes halfway through, that's the server's problem, it can throw an error on the entire request (but really the server should tell the database to not explode)

proposals/4194-user-redaction.md
FSG-Cat: DAG critical events are supposed to be redaction exempt already for critical fields i thought.

proposals/4194-user-redaction.md
Gnuxie: How could a server treat this as singular calls to the the `rooms/redact/event` endpoint?

I'm struggling how to make this more explicit since we already go into details about how MSC2244 style `m.room.redact` can be used. And there's also quite explicit consideration for that in the section for rate limiting the endpoint.

Gnuxie: Ok, I have tried to clarify https://github.com/matrix-org/matrix-spec-proposals/pull/4194/commits/4927088fa36df2395028fec95436881d13d7eaa3, does that work?

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,179 @@
...
+
+```
+200 OK
+Content-Type: application/json
+
+{
+  "is_more_events": false,
+  "redacted_events": 5,
+  "soft_failed_events": 1
+}
```
Gnuxie, line 121: I'd like some thoughts on the names of these fields. I'd also like to know if people agree that it makes less sense to return say an array of event_ids

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,198 @@
...
+
+It could make sense to add a flag to the endpoint to tell the
+moderator's homeserver to issue redactions on their behalf for the
+newly discovered events. However, this could be complicated for
+servers to implement.  The flag would have to reset if the target user
+is unbanned, and all incoming soft failed events will have to be
+checked against a list of flagged servers.
+
+At the moment we will remain forward compatible with a future
+proposal to add a flag.
```
Gnuxie, line 170: Server implementers: Am I right in making a judgement call that this would be too complicated for the proposal?

tulir: Meowlnir just has a hacky feature where if the last redacted event is <5 minutes old, it'll sleep for a bit and refetch events to redact. The assumption is that the most common soft fail issue is where events race with the ban, which should be covered by waiting a while for the ban to propagate.

A slightly less hacky method might be some way to opt into receiving soft-failed events, but that's probably a different MSC

proposals/4194-user-redaction.md
Johennes: > I'd like some thoughts on the names of these fields. 

This is always going to be subjective but here goes:

`soft_failed_events`might be slightly ambiguous in that it doesn't explicitly express that these events have also been redacted. Maybe the counters could be nested under a `redacted`key?

```json
"redacted": {
  "total": 5,
  "soft_failed": 1
}
```

> I'd also like to know if people agree that it makes less sense to return say an array of event_ids

IIUC the main use case for this API is to redact _all_ of a user's events. The counters in the response could nicely be used to display progress. It might be nice to also return the total number of events to let clients display a completion percentage. Not sure if that's easy to get though?

I can't really think of a reason why the caller would need event IDs for the redacted events.

proposals/4194-user-redaction.md
MTRNord: yup works

proposals/4194-user-redaction.md
Gnuxie: > Maybe the counters could be nested under a redactedkey?

Yeah that sounds like a good idea, changing now

Gnuxie: > I can't really think of a reason why the caller would need event IDs for the redacted events.

Yeah, i'll leave this open for now just in case someone else has thoughts

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,206 @@
+# MSC4194: User redaction
```
clokep, line 1: The name of this MSC doesn't really describe what's happening to me, maybe "Redact all events from a user"?

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,206 @@
...
+### Redacting a user's events
+
+A new endpoint is introduced into the client-server spec.
+
+`POST /_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`
+
+This endpoint redacts the target matrix user's unredacted events in
+reverse chronological order, starting from the most
+chronologically-recent visible event in the room history for the
+requesting user.
```
clokep, line 57: Who issues those redactions? Is this C-S API only valid for local users and the user itself is meant to issue the redactions?

Gnuxie: Does this help? https://github.com/matrix-org/matrix-spec-proposals/pull/4194/commits/8f238c708bda49f050777f718de8e45983adba55

proposals/4194-user-redaction.md
Gnuxie: I've tried to make the intent clearer in the writing https://github.com/matrix-org/matrix-spec-proposals/pull/4194/commits/f01e5a625c573fb00725c36d3a047f1375c60588

Thinking of a way to make sure the title communicates that the redactions are happening within the scope of one room.

proposals/4194-user-redaction.md
clokep: Yes. I failed to realize it was for a particular room. 🤦 So yes, doing it as the requester (and checking the requester can do it) makes sense. 👍

proposals/4194-user-redaction.md
Gnuxie: Is this better? https://github.com/matrix-org/matrix-spec-proposals/pull/4194/commits/cb55d016ec36f4ccb207a5eea78e2916f0e36581

proposals/4194-user-redaction.md
Gnuxie: Yeah, Draupnir could in theory just call the endpoint again after a few minutes. I'd still like to just be able to give a flag to this endpoint, even if servers will only honor it for the next 10minutes or something.

tulir:
@anoadragon453 informed me that Synapse coincidentally just merged an admin API with basically the exact same goal as this MSC: https://github.com/element-hq/synapse/pull/17506

Gnuxie:
@tulir it's not the exact same goal as this MSC, this MSC is explicitly supposed to be used by room moderators who may not be synapse or server admins.

tcpipuk:
There are several other servers than Synapse that also need effective moderation tooling, and I know Conduwuit would be eager to implement this MSC if it helps improve the overall moderation experience, especially increasing compatibility with moderation tooling.

Cadair:
I think the fact that synapse made an admin API for this shows there's a need for it, so standardising it is a good idea, not that it makes this redundant.

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Redacting a user's events
+
+A new endpoint is introduced into the client-server spec.
+The endpoint is introduced with the intention to allow convenient and
+quick bulk redaction of spam events for room moderators.  Who may not
+necessarily have administrative privileges on their homeserver, and
+will also need to redact messages from remote users.
+
+`POST /_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`
```
tulir, line 56: The endpoint should probably take a body similar to single event redaction (i.e. a `reason` field)

proposals/4194-user-redaction.md
Gnuxie: @jevolk [suggested](https://matrix.to/#/%21meowlnir-v11%3Amaunium.net/%249cfZU_zJx5qGlop-AdLXR2LPrU8XrfsE5zqxal8V86A?via=maunium.net&via=matrix.org&via=envs.net) that there should be a future proposal to federate the mass-redact action against users to ensure a cleanup of spam that is soft-failed locally but visible on remote servers.

Gnuxie: I should probably also add the context of why I wanted this consideration, which is that: it's a really common issue for room mods to get complaints that spam hasn't been redacted from remote users, but the mods can't see it

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### Path parameters
+
+`roomID`: `string` - The room to get events to redact from.
+
+`userID`: `string` - The sender of the events to redact.
+
+#### Query parameters
+
+`limit`: `integer` - The maximum number of events to
```
timedoutuk, line 103: This MSC seems to heavily assume that this endpoint will only be called after a user has been banned, however the pagination defined here doesn't really account for a scenario where that isn't the case. How should pagination work if a user is not banned and continues to send events after the endpoint has already been called? Does it reset position to the latest event from them, working back from there (presumably skipping over events that already have been redacted)? Without a pagination token that's the only way I can see this working, but then, that makes this endpoint kinda ineffective if you're trying to redact someone who hasn't been removed from the room, since they can just send another event to interrupt the redaction progress.

timedoutuk: Also, bit of a nitpick, but there's no ceiling suggested here - is it reasonable for a client to assume that a server will usually cap out at 100 events, like /messages typically does?

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,220 @@
...
+quick bulk redaction of spam events for room moderators.  Who may not
+necessarily have administrative privileges on their homeserver, and
+will also need to redact messages from remote users.
+
+`POST /_matrix/client/v1/rooms/{roomID}/redact/user/{userID}`
+
+This endpoint redacts the target matrix user's unredacted events by
+sending redactions on behalf of the requesting user.
+
+The target user's unredacted events are sourced in reverse
```
timedoutuk, line 61: Is ordering a *MUST* in this MSC, or just a *SHOULD*? Assuming the intention is generally to redact all of a user's events (looking at the client implementation), I feel that this is more suitably an optional thing, however clarification would be nice.

proposals/4194-user-redaction.md
```diff
@@ -0,0 +1,220 @@
...
+have not been because of the `limit` query parameter.  If
+`is_more_events` is true, then the server should expect that the
+client can optionally call the endpoint again, to redact even more events.
+
+`redacted_events`: `object` - An object with the following fields:
+
+* `total`: `integer` - The number of events that have been redacted,
+  including soft failed events.
+
+* `soft_failed`: `integer` - The number of soft failed events that
```
timedoutuk, line 128: Is there a reason for telling the requester how many soft failed events were redacted? especially since they're already included in `total`?

proposals/4194-user-redaction.md
timedoutuk: Looking at implementing this into [continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/issues/872), my ideal way to handle this would be either the client re-calling later, or a [MSC4284 policy server](https://github.com/matrix-org/matrix-spec-proposals/pull/4284) would deal with mopping up late-arriving events. Tracking for a limited time in the server would probably come with its own suite of reliability and consistency issues, so having it handled elsewhere is probably both easier and more reliable. Just my two cents.

For context: the continuwuity main room has been dealing with a horrific volume of spam during waves, resulting in the aforementioned inconsistent cleanup. Even with Meowlnir doing the re-checking a few minutes later, we found that events were *still* arriving *days* later, primarily via backfill from unsuspecting servers with `prev_events` and whatnot. We set up a policy server to deal with this, and for a week or two it was still issuing maybe a few redactions per new incoming message, and it has only recently calmed down to maybe one redaction every dozen messages instead. I don't think the server redacting incoming future events, nor re-calling on a timer is going to have the desired effect, aside from in some less aggregious cases.

PR #4143, opened 2024-05-10
MSC4143: MatrixRTC – Real-time communication over Matrix

toger5:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md)

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/master/CONTRIBUTING.md#sign-off)
* [ ] Update the title and file name of your proposal to match this PR's number (after opening).
* [ ] Pull request includes a ['Rendered' link](https://matrix.org/docs/spec/proposals#process) above.
* [ ] Your MSC adheres to each point in the [MSC Checklist](MSC_CHECKLIST.md).
* [ ] Ask in
      [#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) to
      get feedback on this PR.

----

SCT Stuff:

[MSC Checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4143#issuecomment-5169244948)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4143#issuecomment-5169269291)

PR #3417, opened 2021-09-30
Call room room type

SimonBrandner:
Requires https://github.com/matrix-org/matrix-doc/pull/3401
Requires https://github.com/matrix-org/matrix-doc/pull/1840

PR #4196, opened 2024-09-19
MSC4196: Voice and video calling application for MatrixRTC

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/matrixrtc-m-call/proposals/4196-matrixrtc-m-call.md)

---

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4196#issuecomment-5577481625)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4196#issuecomment-5577474723)

### DIFF
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -120,17 +120,6 @@ schema:
 * `sticky_key` — Same as above  
 * `disconnect_reason` as defined below
 
-```
-{
-  "* `disconnect_reason` as defined below
-": {
-    "class": "server_error",
-    "reason": "ice_failed",
-    "description": "Failed to establish peer-to-peer connection via ICE",
-  }
-}
-```
-
 `disconnect_reason` **field explanations:**
 
 | Class | Example Reason | Description / When Used |
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -1,65 +1,193 @@
-# MSC4196: MatrixRTC voice and video conferencing application `m.call`
+# MSC4196: MatrixRTC voice and video calling application `m.call`
 
 ## Proposal
 
-We define a MatrixRTC application type of `m.call`.
+We define a **MatrixRTC application** of type `m.call` for real-time calling, designed to support
+the following use cases:
 
-### `m.rtc.member` state event
+* **1:1 calls in direct messages (DMs)** – Private, peer-to-peer conversations.  
+* **Huddles** – Adding a call to an existing room for spontaneous group discussions.  
+* **Persistent “social” rooms** – Discord-style voice or video spaces that are fixed in type but act
+  as a permanent hangout, where users can drop in and out freely, similar to a pub or watercooler.
 
-A valid `m.rtc.member` state event with application `m.call` has the following fields in addition to the fields defined in [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143):
+### MatrixRTC Slots for voice and video calling
 
-- `session` required object:
-  - `application` required string: `m.call`
-  - `call_id` required string: The call ID of the session. Use `""` for a room level call.
-  - `scope` required string: The scope of the call. One of: `m.room`. More scopes may be added in the future.
+For voice and video calling, MatrixRTC defines **a room-level slot model** to represent calls that
+are scoped to a Matrix room and shared among its participants:
 
-For example:
+* **`m.call#ROOM` — Room-level call slot**  
+  Represents a call instance associated with the entire room. This slot is **managed by a user with
+  sufficient power level** and is intended for calls where **any room member is welcome to join**.
+  Ownership of the call is **shared** — it does not depend on who initiated it. This slot type is
+  also suitable for **1:1 direct message (DM) calls**, including use cases resembling traditional
+  **telephone-style calling semantics**.  
 
-```json5
-// event type: "m.rtc.member"
-// MSC3757 formatted state key: "@user:matrix.domain_xyzABCDEF10123"
+### MatrixRTC Member JSON Object for Application `m.call`
+
+A valid `m.rtc.member` state event with **application type** `m.call` includes the standard fields
+defined in [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143), **plus
+additional fields specific to `m.call`**:
+
+```
 {
-  // standard fields from MSC4143:
-  "member": {
-    "id": "xyzABCDEF10123",
-    "device_id": "DEVICEID",
-    "user_id": "@user:matrix.domain"
-  }
-  "focus_active": { ...focus_1 },
-  "foci_preferred": [
-    { ...focus_1 },
-  ],
-  "session": {
-    "application": "m.call",
+  "application": {
+    "type": "m.call",
     // additional fields for m.call:
-    "call_id": "",
-    "scope": "m.room"
-  }
+    "m.call.id": UUID                                 // optional
+    "m.call.intent": "voice" | "video" | "any-value"  // optional
+  },
 }
 ```
 
-### Ringing with the `m.rtc.notify` room event
+**Field Description**
 
-A valid `m.rtc.notify` event with application `m.call` has the following fields in addition to the fields defined in [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075):
+| Field | Type | Required | Description |
+| :---- | :---- | ----- | :---- |
+| `type` | string | ✅ | Must be `"m.call"`. Identifies the application type. |
+| `m.call.id` | string | ⚪ | Optional call ID of the session. Serves as a unique identifier for the call instance. |
+| `m.call.intent` | string | ⚪ | Optional hint expressing the intended use-case of the session, e.g., `"voice"`, `"video"`, or `"spatial-audio"`. This field is **informational only** and non-authoritative, reflecting the participant's perception of the session. |
 
-- `session` required object: the contents of the `session` from the `m.rtc.member` event.
+Participants MAY include a `m.call.intent` field to express the intended use-case of their `m.call`
+session (e.g. `"voice"`, `"video"`, `"spatial-audio"`). Because of Matrix decentralisation, this
+field is purely informational and non-authoritative, serving only as a hint about how the
+participant perceives the session.
 
-```json5
-// event type: "m.rtc.notify"
+The following is an example of a well-formed `m.rtc.member` event for a participant joining a
+room-level call (`m.call#ROOM`) using the `m.call` application type:
+
+```
+// event type: "m.rtc.member"
 {
-  "content": {
-    // standard fields from MSC4075:
-    "application": "m.call",
-    "m.mentions": {"user_ids": [], "room": true },
-    "notify_type": "notification",
-    "session": {
-      "application": "m.call",
-      // for application = "m.call":
-      "call_id": ""
-    }
+  "slot_id": "m.call#ROOM",
+  "member": {
+    "id": "xyzABCDEF0123"                             // UUID, unique participation instance
+    "claimed_device_id": "DEVICEID",
+    "claimed_user_id": "@user:matrix.domain"
+  },
+  "sticky_key": "xyzABCDEF0123"                       // same as member.id 
+  "application": {
+    "type": "m.call",
+    // additional fields for m.call:
+    "m.call.id": UUID
+    "scope": "m.room"
+    "m.call.intent": "voice" | "video" | "any-value"  // optional
+  },
+  "m.relates_to":{                                    // Reference to original join event; omit if first event
+    rel_type: "m.reference",
+    event_id: "$connect_event_id"
+  },
+  "rtc_transports": [
+    { /* TRANSPORT_1 details */ }
+  ],
+  "versions": [
+    "v0"
+  ],
+}
+```
+
+### Ending a Call and Post-Connect Error Handling
+
+Participation in a call is ended by disconnecting the MatrixRTC slot, as defined in
+[MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143). The
+**`disconnect_reason`** field is used to provide further details about the disconnection and can
+also be used for structured error handling.
+
+A valid `m.rtc.member` event, as a prerequisite for disconnecting from a slot, has the following
+schema:
+
+```
+// event type: "m.rtc.member"
+{
+  "slot_id": "m.call.ROOM",
+  "sticky_key": "xyzABCDEF0123",
+  "m.relates_to":{                  // SHOULD
+    rel_type: "m.reference",
+    event_id: "$connect_event_id"
+  },
+  // Optional
+  "disconnect_reason": { 
+    "class": "server_error",
+    "reason": "ice_failed",
+    "description": "Failed to establish peer-to-peer connection via ICE",
   }
 }
 ```
+
+**Field explanations:**
+
+* `slot_id` — The slot this member belongs to.  
+* `m.relates_to` —  The `m.relates_to` field optionally references the initial connect event  
+* `sticky_key` — Same as above  
+* `disconnect_reason` as defined below
+
+```
+{
+  "* `disconnect_reason` as defined below
+": {
+    "class": "server_error",
+    "reason": "ice_failed",
+    "description": "Failed to establish peer-to-peer connection via ICE",
+  }
+}
+```
+
+`disconnect_reason` **field explanations:**
+
+| Class | Example Reason | Description / When Used |
+| ----- | ----- | ----- |
+| `user_action` | `hangup` | Participant intentionally ended the call after joining. |
+|  | `switch_device` | User moved the session to another device mid-call. |
+| `client_error` | `media_error` | Failed to capture or transmit audio/video after joining. |
+|  | `transport_failure` | Local ICE/DTLS setup failed despite a successful `m.rtc.member` event. |
+|  | `encryption_error` | Failed to set up E2EE for the media channel after connecting. |
+| `server_error` | `ice_failed` | ICE negotiation could not complete due to network/server issues. |
+|  | `dtls_failed` | DTLS handshake failed. |
+|  | `network_error` | Temporary network outage caused the connection to drop. |
+| `redirection` | `call_transferred` | Call was redirected to another slot, device, or user. |
+|  | `moved_temporarily` | Session temporarily moved (e.g., server migration). |
+| `permanent_failure` | `codec_mismatch` | Participant cannot decode/encode the call media. |
+|  | `unsupported_features` | Session requested unsupported capabilities. |
+
+### Call Ringing Using `m.rtc.notification` room event
+
+A valid `m.rtc.notification` event for a MatrixRTC session with application `m.call` MAY have the
+following fields in addition to the fields defined in
+[MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075):
+- `m.call.intent`
+
+```
+// event type: "m.rtc.notification"
+{
+  "type":"m.rtc.notification",  // org.matrix.msc4075.rtc.notification
+  "content": {
+    "sender_ts": 1752583130365,
+    "lifetime": 30000,
+    "m.mentions": {"user_ids": [], "room": true | false},
+    "m.relates_to": {"rel_type":"m.reference", "event_id":"$rtc_member_event_id"},
+    "notification_type": "ring | notification",
+    // additional fields for m.call:
+    "m.call.intent": "voice" | "video" | "any-value"  // optional, 
+  }
+}
+```
+
+### Handling of `m.call.intent`
+
+Clients SHOULD infer a “voice” `m.call.intent` as a voice call, and “video” as a voice and video
+call. 
+
+Clients SHOULD NOT submit a video track if the user has requested a “voice” call initially, although
+this is a soft limitation and users may choose to upgrade to include a video track later in the
+call.
+
+If the intent is not understood, the default value of “video” should be assumed (both voice and
+video tracks may be expected).
+
+### Defaults
+
+[The createRoom preset
+option](https://spec.matrix.org/v1.16/client-server-api/#post_matrixclientv3createroom)
+`trusted_private_chat` should enable a default room slot `m.call#ROOM` for the application `m.call`.
 
 ## Potential issues
 
@@ -69,9 +197,10 @@ A valid `m.rtc.notify` event with application `m.call` has the following fields 
 
 ## Unstable prefix
 
-The `m.call` application type is already within unstable prefixed entries (e.g.
-`org.matrix.msc3401.call.member`) and as such doesn't need its own unstable prefix.
+The `m.call` application type is already within unstable prefixed entries (i.e.
+`org.matrix.msc4143.rtc.member`) and as such doesn't need its own unstable prefix.
 
 ## Dependencies
 
-This MSC builds on [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) and [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075).
+This MSC builds on [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) and
+[MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075).
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -1,4 +1,4 @@
-# MSC0000: MatrixRTC voice and video conferencing application `m.call`
+# MSC4196: MatrixRTC voice and video conferencing application `m.call`
 
 ## Proposal
 
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -2,6 +2,65 @@
 
 ## Proposal
 
+We define a MatrixRTC application type of `m.call`.
+
+### `m.rtc.member` state event
+
+A valid `m.rtc.member` state event with application `m.call` has the following fields in addition to the fields defined in [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143):
+
+- `session` required object:
+  - `application` required string: `m.call`
+  - `call_id` required string: The call ID of the session. Use `""` for a room level call.
+  - `scope` required string: The scope of the call. One of: `m.room`. More scopes may be added in the future.
+
+For example:
+
+```json5
+// event type: "m.rtc.member"
+// MSC3757 formatted state key: "@user:matrix.domain_xyzABCDEF10123"
+{
+  // standard fields from MSC4143:
+  "member": {
+    "id": "xyzABCDEF10123",
+    "device_id": "DEVICEID",
+    "user_id": "@user:matrix.domain"
+  }
+  "focus_active": { ...focus_1 },
+  "foci_preferred": [
+    { ...focus_1 },
+  ],
+  "session": {
+    "application": "m.call",
+    // additional fields for m.call:
+    "call_id": "",
+    "scope": "m.room"
+  }
+}
+```
+
+### Ringing with the `m.rtc.notify` room event
+
+A valid `m.rtc.notify` event with application `m.call` has the following fields in addition to the fields defined in [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075):
+
+- `session` required object: the contents of the `session` from the `m.rtc.member` event.
+
+```json5
+// event type: "m.rtc.notify"
+{
+  "content": {
+    // standard fields from MSC4075:
+    "application": "m.call",
+    "m.mentions": {"user_ids": [], "room": true },
+    "notify_type": "notification",
+    "session": {
+      "application": "m.call",
+      // for application = "m.call":
+      "call_id": ""
+    }
+  }
+}
+```
+
 ## Potential issues
 
 ## Alternatives
@@ -10,4 +69,9 @@
 
 ## Unstable prefix
 
+The `m.call` application type is already within unstable prefixed entries (e.g.
+`org.matrix.msc3401.call.member`) and as such doesn't need its own unstable prefix.
+
 ## Dependencies
+
+This MSC builds on [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) and [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075).
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -15,7 +15,7 @@ the following use cases:
 For voice and video calling, MatrixRTC defines **a room-level slot model** to represent calls that
 are scoped to a Matrix room and shared among its participants:
 
-* **`m.call#ROOM` — Room-level call slot**  
+* **`slot_id`: `m.call#ROOM` — Room-level call slot**  
   Represents a call instance associated with the entire room. This slot is **managed by a user with
   sufficient power level** and is intended for calls where **any room member is welcome to join**.
   Ownership of the call is **shared** — it does not depend on who initiated it. This slot type is
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -192,4 +192,4 @@ The `m.call` application type is already within unstable prefixed entries (i.e.
 ## Dependencies
 
 This MSC builds on [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) and
-[MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075).
+[MSC4354](https://github.com/matrix-org/matrix-spec-proposals/pull/4354).
--- a/proposals/4196-matrixrtc-m-call.md
+++ b/proposals/4196-matrixrtc-m-call.md
@@ -1,14 +1,26 @@
-# MSC4196: MatrixRTC voice and video calling application `m.call`
+# MSC4196: Voice and video calling application for MatrixRTC
+
+[MSC4143] introduces MatrixRTC as an extensible framework for real-time communication in Matrix.
+MatrixRTC uses so called transports to transfer the RTC data between RTC members. Transports are
+then used in what MatrixRTC calls applications to build user experiences for concrete use cases.
+
+This proposal introduces a MatrixRTC application for voice and video calling that is able to power
+a variety of calling use cases including but not limited to classical 1-on-1 and group calling
+as well as Discord-style [Voice Channels]. The application is compatible with the [LiveKit] transport
+introduced in [MSC4195].
+
+Note that call ringing and notifications are not in scope for this proposal. These are covered
+in [MSC4075] and [MSC4310].
+
+[MSC4143]: https://github.com/matrix-org/matrix-spec-proposals/pull/4143
+[Voice Channels]: https://support.discord.com/hc/en-us/articles/19583625604887-Voice-Channels-FAQs
+[LiveKit]: https://github.com/livekit/livekit
+[MSC4195]: https://github.com/matrix-org/matrix-spec-proposals/pull/4195
+[MSC4075]: https://github.com/matrix-org/matrix-spec-proposals/pull/4075
+[MSC4310]: https://github.com/matrix-org/matrix-spec-proposals/pull/4310
+
 
 ## Proposal
-
-We define a **MatrixRTC application** of type `m.call` for real-time calling, designed to support
-the following use cases:
-
-* **1:1 calls in direct messages (DMs)** – Private, peer-to-peer conversations.  
-* **Huddles** – Adding a call to an existing room for spontaneous group discussions.  
-* **Persistent “social” rooms** – Discord-style voice or video spaces that are fixed in type but act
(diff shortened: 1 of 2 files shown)


proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
+# MSC4196: MatrixRTC voice and video calling application `m.call`
```
Johennes, line 1: This could use a short introduction and especially an early backlink to https://github.com/matrix-org/matrix-spec-proposals/pull/4143 as the foundation for this proposal.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+* **Huddles** – Adding a call to an existing room for spontaneous group discussions.  
+* **Persistent “social” rooms** – Discord-style voice or video spaces that are fixed in type but act
+  as a permanent hangout, where users can drop in and out freely, similar to a pub or watercooler.
+
+### MatrixRTC Slots for voice and video calling
+
+For voice and video calling, MatrixRTC defines **a room-level slot model** to represent calls that
+are scoped to a Matrix room and shared among its participants:
+
+* **`m.call#ROOM` — Room-level call slot**
```
Johennes: It might help to reiterate that this is the `slot_id` that goes into the `state_key`?

```suggestion
* **`slot_id`: `m.call#ROOM` — Room-level call slot**  
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+**Field Description**
+
+| Field | Type | Required | Description |
+| :---- | :---- | ----- | :---- |
+| `type` | string | ✅ | Must be `"m.call"`. Identifies the application type. |
+| `m.call.id` | string | ⚪ | Optional call ID of the session. Serves as a unique identifier for the call instance. |
+| `m.call.intent` | string | ⚪ | Optional hint expressing the intended use-case of the session, e.g., `"voice"`, `"video"`, or `"spatial-audio"`. This field is **informational only** and non-authoritative, reflecting the participant's perception of the session. |
```
Johennes, line 48: I'm wondering if this should be an array rather than a single string. Could we possibly foresee any intents that could be combined? Like a video call but you want people to join muted maybe?

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+    "id": "xyzABCDEF0123"                             // UUID, unique participation instance
+    "claimed_device_id": "DEVICEID",
+    "claimed_user_id": "@user:matrix.domain"
+  },
+  "sticky_key": "xyzABCDEF0123"                       // same as member.id 
+  "application": {
+    "type": "m.call",
+    // additional fields for m.call:
+    "m.call.id": UUID
+    "scope": "m.room"
```
Johennes, line 72: This field seems undefined here and in #4143. Could it be that this is a leftover from an earlier design?


```suggestion
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Unstable prefix
+
+The `m.call` application type is already within unstable prefixed entries (i.e.
+`org.matrix.msc4143.rtc.member`) and as such doesn't need its own unstable prefix.
+
+## Dependencies
+
+This MSC builds on [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) and
+[MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075).
```
Johennes: https://github.com/matrix-org/matrix-spec-proposals/pull/4075 is call ringing which depends on this proposal. I suspect this meant to list https://github.com/matrix-org/matrix-spec-proposals/pull/4354 instead given that `sticky_key` is used above.

```suggestion
[MSC4354](https://github.com/matrix-org/matrix-spec-proposals/pull/4354).
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+* **Huddles** – Adding a call to an existing room for spontaneous group discussions.  
+* **Persistent “social” rooms** – Discord-style voice or video spaces that are fixed in type but act
+  as a permanent hangout, where users can drop in and out freely, similar to a pub or watercooler.
+
+### MatrixRTC Slots for voice and video calling
+
+For voice and video calling, MatrixRTC defines **a room-level slot model** to represent calls that
+are scoped to a Matrix room and shared among its participants:
+
+* **`slot_id`: `m.call#ROOM` — Room-level call slot**
```
gewitternacht, line 18: As #4143 [prescribes](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md#slots) that `application_type` and `application_slot_id` are to be "defined by the application's specification", wouldn't it be good to explicitly mention these identifiers, and define them separately?

But later, it also says this, so not sure...
> The grammar for forming slot IDs [...] exists only to namespace the `state_key`, and could be modified in a future proposal.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+**`disconnect_reason`** field is used to provide further details about the disconnection and can
+also be used for structured error handling.
+
+A valid `m.rtc.member` event, as a prerequisite for disconnecting from a slot, has the following
+schema:
+
+```
+// event type: "m.rtc.member"
+{
+  "slot_id": "m.call.ROOM",
```
gewitternacht, line 101:
```suggestion
  "slot_id": "m.call#ROOM",
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+room-level call (`m.call#ROOM`) using the `m.call` application type:
+
+```
+// event type: "m.rtc.member"
+{
+  "slot_id": "m.call#ROOM",
+  "member": {
+    "id": "xyzABCDEF0123"                             // UUID, unique participation instance
+    "claimed_device_id": "DEVICEID",
+    "claimed_user_id": "@user:matrix.domain"
```
gewitternacht, line 65: `claimed_user_id` doesn't exist in [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md#joining-a-slot).

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+| `type` | string | ✅ | Must be `"m.call"`. Identifies the application type. |
+| `m.call.id` | string | ⚪ | Optional call ID of the session. Serves as a unique identifier for the call instance. |
+| `m.call.intent` | string | ⚪ | Optional hint expressing the intended use-case of the session, e.g., `"voice"`, `"video"`, or `"spatial-audio"`. This field is **informational only** and non-authoritative, reflecting the participant's perception of the session. |
+
+Participants MAY include a `m.call.intent` field to express the intended use-case of their `m.call`
+session (e.g. `"voice"`, `"video"`, `"spatial-audio"`). Because of Matrix decentralisation, this
+field is purely informational and non-authoritative, serving only as a hint about how the
+participant perceives the session.
+
+The following is an example of a well-formed `m.rtc.member` event for a participant joining a
```
gewitternacht, line 55:
```suggestion
The following is an example of the well-formed `content` of an `m.rtc.member` event for a participant joining a
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+  "m.relates_to":{                                    // Reference to original join event; omit if first event
+    rel_type: "m.reference",
+    event_id: "$connect_event_id"
+  },
+  "rtc_transports": [
+    { /* TRANSPORT_1 details */ }
+  ],
+  "versions": [
+    "v0"
+  ],
```
gewitternacht, line 84: What's this?

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+    "type": "m.call",
+    // additional fields for m.call:
+    "m.call.id": UUID
+    "scope": "m.room"
+    "m.call.intent": "voice" | "video" | "any-value"  // optional
+  },
+  "m.relates_to":{                                    // Reference to original join event; omit if first event
+    rel_type: "m.reference",
+    event_id: "$connect_event_id"
+  },
```
gewitternacht, line 78: I think this is also outdated, see https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md#chaining-member-events-with-relations.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,195 @@
...
+  "m.relates_to":{                  // SHOULD
+    rel_type: "m.reference",
+    event_id: "$connect_event_id"
+  },
+  // Optional
+  "disconnect_reason": { 
+    "class": "server_error",
+    "reason": "ice_failed",
+    "description": "Failed to establish peer-to-peer connection via ICE",
+  }
```
gewitternacht, line 112: Outdated (also the field explanations below).

gewitternacht (commented):
I only realised while adding comments that this seems to generally be rather outdated. I'm still leaving them here in case they help at all, but if it's planned to revise this in a more general way anyway, feel free to ignore.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: This has since been removed from the proposal. I had the same question and we couldn't find a good definition or use case for this property.

proposals/4196-matrixrtc-m-call.md
Johennes: Yes, you're right. This has been removed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
Johennes: Retracting my comment. I think an enum is fine here after all.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal. The text now explicitly describes the slot ID construction.

proposals/4196-matrixrtc-m-call.md
Johennes: This is now fixed in the latest state of the proposal.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+    },
+    "encryption": {
+      "type": "m.per_member",
+    }
+  },
+  ...
+}
+```
+
+When clients create rooms with `preset = private_chat` in [`/createRoom`], they SHOULD by default
```
robintown: Or also `preset = trusted_private_chat`?

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+include an open `m.rtc.slot` event for `m.call` in `initial_state`. Clients MAY let the user
+override this default behaviour.
```
robintown: I was kind of expecting this to be a SHOULD on the part of the server, i.e. actually part of the preset.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+`encryption` content block on the initial slot event MUST be set to `{ "type": "m.per_member" }`.
+Otherwise, the `encryption` property MUST be omitted.
+
+The default [power levels] assigned during room creation prevent room members other than the room
+creator from sending state events. Including the slot event at room creation time, ensures that
+room members are able to have calls in the room without depending on a room administrator to send
+the slot event later.
+
+For the same reason, homeservers SHOULD include `m.rtc.slot` events for `m.call` applications in
+the set of state events that are replicated during [room upgrades].
```
robintown: Is there a good reason to not have this rule apply to all `m.rtc.slot` events from the get-go in MSC4143?

The one concern that comes to my mind is state bloat, but I can't decide how much weight to give it. The fact that this is a SHOULD implies that future applications could carve out exceptions to the rule if they are designed in a way that is likely to bloat a room's state over time (which, hopefully, they're not).

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+
+Including `m.rtc.slot` events in `initial_state` is not required when other `preset` values are
+used in [`/createRoom`]. With the `trusted_private_chat` preset, all room members get the same power
+level as the room creator. Thus, they can send the `m.rtc.slot` event themselves when needed. In
```
robintown: I think we shouldn't assume that all clients will be fine with marrying the functions of enabling calls and starting a call in a single button, which is what you'd probably want to do to keep the user interface convenient in this case.

If a client's 'start call' button still acts like calls are enabled so long as you *could* enable them, this may prove confusing to an admin of a public room who is looking for evidence that disabling calls in their room has had a tangible effect - or perhaps even prove dangerous, as a single misclick on the button could re-enable calls. Granted, the client could make an exception for rooms that don't look like `trusted_private_chat`, but wouldn't it be simpler to just include an `m.rtc.slot` event in this preset too?

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+[room upgrades]: https://spec.matrix.org/v1.18/client-server-api/#server-behaviour-21
+
+### Membership events
+
+The schema for the `application` content block in `m.rtc.member` events that are joined to `m.call`
+slots, looks as follows:
+
+- `type` (string, required): MUST be `m.call`.
+- `intent` (string): One of `audio`, `video`. Optionally discloses whether the member intends to
+  join the session with audio only or with audio and video. Clients SHOULD set this field when joining
```
robintown:
```suggestion
  join the call with audio only or with audio and video. Clients SHOULD set this field when joining
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+the `intent` property on `m.rtc.member` events. Given that any room member can join the
+slot, this information is effectively obtainable by all room members anyway though. In either
+case, users can opt not to fill `intent` given that it is an optional property.
+
+### Faking intent
+
+A malicious user could set `intent = video` without actually publishing their video stream
+to trick another user into joining the call with their camera on. To mitigate this, clients
+SHOULD default to joining with video disabled regardless of the `intent` values of other
+members. They MAY allow users to override the default setting though.
```
robintown: If I see an incoming video call and consent to answer it, I'd only be a little let down to find out that the caller isn't sharing video. The same goes for if someone genuinely video called me but kept a piece of tape over their webcam.

What's actually important is that clients do something to demonstrate what I'm personally consenting to share when I click 'join' (even if that just amounts to saying 'incoming call' vs. 'incoming video call').

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+When leaving a slot, [MSC4143] allows clients to optionally provide context with regards to the
+reason for leaving in the `leave_reason` property of their `m.rtc.member` event. For `m.call`
+applications, the generic `leave_reason.code` values provided in [MSC4143] are extended with
+the following additional codes:
+
+- `transport_error`: The client failed to negotiate a connection over the chosen transport
+  (e.g. due to an ICE/DTLS setup failure).
+- `media_error`: The client failed to capture or transmit audio and/or video after joining.
+- `codec_mismatch`: The client could not decode/encode the call media.
+- `encryption_error`: The client failed to set up end-to-end encryption for the media channel.
```
robintown: How might that happen? (Only scenario I can imagine is that someone's browser disables web workers, hopefully unusual enough that a generic error would be just fine)

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+a room.
+
+## Alternatives
+
+### Injecting `m.rtc.slot` events on the server
+
+Instead of having clients pass in the initial `m.rtc.slot` event via `initial_state` on
+[`/createRoom`] requests, this logic could also be implemented by the server. This would
+further complicate the already complex steps the server has to run through during room
+creation though.
```
robintown, line 177: Hm, I kind of was expecting this to be the server's responsibility in the case of the `private_chat` or `trusted_private_chat` presets. It seems like exactly what presets are for…

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Security considerations
+
+### Metadata leakage through intent
+
+Some users might not be comfortable with disclosing whether their camera is on or off via
+the `intent` property on `m.rtc.member` events. Given that any room member can join the
+slot, this information is effectively obtainable by all room members anyway though. In either
```
robintown:
```suggestion
session, this information is effectively obtainable by all room members anyway (though the join would at least be a visible choice). In either
```

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
...
+  },
+  ...
+}
+```
+
+When a client joins an `m.call` slot where all other members have set their `intent` to `audio`,
+the joining client SHOULD not publish a video track by default. It MAY allow the user to overrule
+this initial setting both before and after joining though. If the other members use different
+intents or have not disclosed their intent, the joining client SHOULD assume the session to
+represent a video call. This does *not* imply that the client should enter the call with video
```
robintown:
```suggestion
this initial setting both before and after joining though. If any other members have an
`intent` of `video`, the joining client SHOULD assume the session to
represent a video call. This does *not* imply that the client should enter the call with video
```

(Nit, but I think we don't need to recommend any particular response to someone withholding their intent.)

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,201 @@
+# MSC4196: Voice and video calling application for MatrixRTC
+
+[MSC4143] introduces MatrixRTC as an extensible framework for real-time communication in Matrix.
+MatrixRTC uses so called transports to transfer the RTC data between RTC members. Transports are
+then used in what MatrixRTC calls applications to build user experiences for concrete use cases.
+
+This proposal introduces a MatrixRTC application for voice and video calling that is able to power
+a variety of calling use cases including but not limited to classical 1-on-1 and group calling
+as well as Discord-style [Voice Channels]. The application is compatible with the [LiveKit] transport
```
robintown: For the voice channels use case, we could link to MSC3417 (not that the text there has been updated for MatrixRTC yet)

robintown (commented):
Thanks for fleshing this out, it's looking really good to me. Just got a couple suggestions for how to make things less surprising, and nits.

proposals/4196-matrixrtc-m-call.md
Johennes: Good point. 👍

proposals/4196-matrixrtc-m-call.md
Johennes: I made it a SHOULD because replicating state events on room upgrades is generally only a RECOMMENDATION in the spec (see point 3 [here](https://spec.matrix.org/v1.18/client-server-api/#server-behaviour-21)).

It's a good question whether we should make this call-specific or not though. Given that the slot is only the access control for a potential RTC session, I suppose it makes sense to replicate it regardless of application type. Let me move this over to #4143.

Johennes: Added in #4143 with fda56fc07879d37464d5f743b3c29a45006e4c54 and removed here with 15a95f6d6bc578795e1945f8363dc327b3c1d419.

proposals/4196-matrixrtc-m-call.md
Johennes: > If a client's 'start call' button still acts like calls are enabled so long as you _could_ enable them, this may prove confusing to an admin of a public room who is looking for evidence that disabling calls in their room has had a tangible effect - or perhaps even prove dangerous, as a single misclick on the button could re-enable calls.

Hm, but in such a client the call button would show as enabled for the admin regardless of whether the slot exists and is open. This feels like a UX concern unrelated to creating slots at room creation time?

proposals/4196-matrixrtc-m-call.md
Johennes: So how would you change this paragraph? My thinking here was less about not seeing the other person's video but rather that I don't want to share my own video (possibly disclosing my surroundings, allowing the other person to screenshot me, etc.) needlessly.

proposals/4196-matrixrtc-m-call.md
Johennes: The LiveKit Swift SDK has an encryption error type but on further inspection, I'm realizing that it's actually only used internally. I think you're right. We probably don't need this error code split out. Have removed it in 16cb63a33e6fe15ecad2be1910a7dd81be5063d5.

proposals/4196-matrixrtc-m-call.md
Johennes: Added in fb6fd97a36fab62c6f08bb2139177b4cddc67df8.

proposals/4196-matrixrtc-m-call.md
Johennes: I think you could argue the same about enabling room encryption at room creation time. Currently clients have to do this manually by specifying `m.room.encryption` in `initial_state`. I'm not sure what the history of that is to be honest though.

Other than adding to the existing complexity of `/createRoom`, one more potential issue I can see is that slots might contain further configuration options in future. Things like different encryption schemes (such as shared key encryption) or application specific settings (such as voice-only calls). It might be difficult to reflect all of that via presets?

proposals/4196-matrixrtc-m-call.md
robintown: > in such a client the call button would show as enabled for the admin regardless of whether the slot exists and is open

Yes, my point is that's a UX workaround which not all clients may be satisfied with, so I suggest we recommend an `m.rtc.slot` in both presets and thereby avoid giving clients an extra reason to have the workaround.

Johennes: I'm unsure if that will help much. Even if we add it to the preset, there's no guarantee that the slot will actually stay open. So I think clients will have to consider the case of "the slot is closed but I have enough power level to open it" one way of another?

In any case, I don't see any risk with auto-creating the slot in the `trusted_private_chat` preset, too. So given that you see value in that, let me just add that to the proposal.

Johennes: Have done so in 76dce0bca9035cdf82436a6788624db4dffac9a7.

proposals/4196-matrixrtc-m-call.md
robintown: Yeah okay, then how would this sound?

> ### Consent to share media
>
> Before joining a call, clients SHOULD demonstrate to the user what media (audio and/or video) they are about to share so that the user can meaningfully consent to sharing it.
>
> In particular, clients should be careful about the fact that remote users can control a call's intent. If an `intent` of `video` would cause a client to automatically start sharing video upon accepting an incoming call, then it SHOULD clearly differentiate the call from an audio call. This could be achieved for instance by presenting different text labels ('Incoming video call' / 'Join with video') or by showing the user a preview of their video.
>
> Clients MAY of course allow their users to override any default settings before joining a call, so that a user may accept an incoming video call without sharing their own video, for example.

The final paragraph is what relates most to your concern, though I have a hard time describing a further security impact of fake intents as long as the first two paragraphs are upheld.

proposals/4196-matrixrtc-m-call.md
robintown: I suppose with `m.room.encryption` there's the problem that if it were included in the preset, clients would have no way of disabling encryption with those presets. `initial_state` can only add state events, not take them away, and to disable encryption you need the room to never have had an `m.room.encryption` event at all. With `m.rtc.slot` that's not a problem, as you can add one to `initial_state` with `status = closed` to override the defaults.

It would just be a preset, though expecting reconfiguration to be quite common in the future still sounds like a decent argument for not bothering with this. Also, on second thought, maybe calls aren't such a core feature of the protocol that they should be governed by presets that otherwise deal with just access, power, and history visibility.

Maybe let's just wait and see if this comes up with any other reviewers?

proposals/4196-matrixrtc-m-call.md
robintown: > So I think clients will have to consider the case of "the slot is closed but I have enough power level to open it" one way of another?

I expect that disabling/hiding the call button and pointing the user toward a switch that they can flip in the room's settings would work fine for that case

proposals/4196-matrixrtc-m-call.md
Johennes: Sounds good, thank you. Have applied this with d591b5636d75f0dadb0d2a2d01858546e08735d5.

proposals/4196-matrixrtc-m-call.md
Johennes, line 1: Implementation requirements:

- [X] Client

Johennes: - `m.call` application in `m.rtc.slot` events: https://github.com/element-hq/element-web/pull/34392
- `m.call` application in `m.rtc.member` events: https://github.com/matrix-org/matrix-js-sdk/blob/e16b0bcc06d0b17889f3eedef972f3cae9790f08/src/matrixrtc/MembershipManager.ts#L1102-L1125
- Connecting to LiveKit rooms & subscribing: https://github.com/element-hq/element-call/blob/d765267c431685c655c1dad0b6bddc611362c16a/src/state/CallViewModel/remoteMembers/Connection.ts#L188
- Publishing call media to LiveKit: https://github.com/element-hq/element-call/blob/d765267c431685c655c1dad0b6bddc611362c16a/src/state/CallViewModel/localMember/Publisher.ts
- Rendering subscribed video tracks: https://github.com/element-hq/element-call/blob/d765267c431685c655c1dad0b6bddc611362c16a/src/tile/MediaView.tsx#L188-L189
- Rendering subscribed audio tracks: https://github.com/element-hq/element-call/blob/d765267c431685c655c1dad0b6bddc611362c16a/src/livekit/MatrixAudioRenderer.tsx#L56

> [!NOTE]
> The implementation uses `m.call.intent` instead of `intent`.
>
> Supplying an `m.rtc.slot` event for `m.call` in `initial_state` on `/createRoom` is not currently implemented (but trivial). There is an [open discussion](https://github.com/matrix-org/matrix-spec-proposals/pull/4196/changes#r3918135826) about whether this should happen on the client or the server.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+When clients create rooms with a `preset` of `private_chat` or `trusted_private_chat` in
+[`/createRoom`], they SHOULD by default include an open `m.rtc.slot` event for `m.call` in
+`initial_state`. Clients MAY let the user override this default behaviour.
+
+As per [MSC4143], encryption of MatrixRTC sessions is mandatory in encrypted rooms and forbidden
+in unencrypted rooms. Therefore, if [`m.room.encryption`] is also present in `initial_state`, the
+`encryption` content block on the initial slot event MUST be set to `{ "type": "m.per_member" }`.
+Otherwise, the `encryption` property MUST be omitted.
```
gewitternacht, line 67: So, if someone "manually" provides contradicting `m.rtc.slot` and `m.room.encryption` events in `initial_state`, what happens? Does room creation fail, or slot creation fail? Does any of this depend on the order in which the two events are given?
Which reminds me: Did we specify in MSC4143 what happens with unencrypted slots when encryption gets enabled in a previously unencrypted room (I can't find anything like that right now), and does this influence the behaviour here?

(I'm not sure whether all of this needs to be specified here or whether it should be "obvious", I'm just unsure myself.)

proposals/4196-matrixrtc-m-call.md
gewitternacht: My questions [above](https://github.com/matrix-org/matrix-spec-proposals/pull/4196/changes#r3932439118) also made me think whether it wouldn't be cleaner for the server to handle this, but I can very much see the reconfiguration and not-a-core-feature arguments

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Security considerations
+
+### Metadata leakage through intent
+
+Some users might not be comfortable with disclosing whether their camera is on or off via
+the `intent` property on `m.rtc.member` events. Given that any room member can join the
+session, this information is effectively obtainable by all room members anyway (though the
+join would at least be a visible choice). In either case, users can opt not to fill `intent`
+given that it is an optional property.
```
gewitternacht: > Given that any room member can join the session

Is that so? What if they lack the power level to send `m.rtc.member` events? Then they at least aren't "officially" part of the session...

> the join would at least be a visible choice

... but I think this is also not true, because they could also be "unofficially" be part of a session (see [here](https://github.com/hughns/matrix-spec-proposals/blob/hughns/matrixrtc-livekit/proposals/4195-matrixrtc-livekit.md#side-stepping-matrixrtc-membership)).

So I think the whole argumentation here isn't really appropriate. But maybe it isn't even necessary, and 
> In either case, users can opt not to fill `intent` given that it is an optional property.

is enough?

proposals/4196-matrixrtc-m-call.md
Johennes: > So, if someone "manually" provides contradicting `m.rtc.slot` and `m.room.encryption` events in `initial_state`, what happens? Does room creation fail, or slot creation fail? Does any of this depend on the order in which the two events are given?

It would just end up in a non-compliant room configuration. The client passes the events in and the server writes them into the room as they are. I suppose this is comparable to creating a private room without enabling encryption or creating a public room with encryption (with the difference that we have explicitly forbidden that combination for MatrixRTC).

We could technically make the server reject contradicting `m.rtc.slot` and `m.room.encryption` events in `/createRoom`. It'll still be possible to create contradicting events after room creation though. So I'm not sure if it's worth it.

> Which reminds me: Did we specify in [MSC4143](https://github.com/matrix-org/matrix-doc/pull/4143) what happens with unencrypted slots when encryption gets enabled in a previously unencrypted room (I can't find anything like that right now), and does this influence the behaviour here?

We don't seem to have it spelled out explicitly in #4143, true. How about we call it out like this:

> Use of encryption in MatrixRTC is REQUIRED in encrypted rooms. This means that `m.rtc.member` events MUST be encrypted and `m.rtc.slot` events MUST contain an encryption object when sent in an encrypted room. Member / slot events that violate these conditions MUST be considered left / closed. Similarly, clients MUST ignore unencrypted RTC streams when connected to an encrypted slot. **If a slot's use of encryption does not align with the room's use of encryption, clients MUST NOT join the slot, leaving it if needed.**

proposals/4196-matrixrtc-m-call.md
Johennes: Yeah, true. I think we strip it down a little to not get lost in the tangential details.

```suggestion
The `intent` property on `m.rtc.member` events is visible to all room members, not just
those joined to the MatrixRTC session. Some user's might not be comfortable with disclosing
whether their camera is on or off outside the session though. To mitigate this, users can opt
not to fill `intent` given that it is an optional property.
```

proposals/4196-matrixrtc-m-call.md
gewitternacht: > It would just end up in a non-compliant room configuration.

Okay, so the MUST is not something that the server enforces, but something that a client must do in order to end up with usable slots, I see. And the same is true for encryption alignment for any later `m.rtc.slot` events.
Given that, I'd agree that it's probably not worth rejecting a misconfigured `initial_state`.

> How about we call it out like this: ...

That seem like a good addition. I'm not totally sure if it fully covers what I was asking though. Do you mean for this to imply that on an unencrypted->encrypted transition of a room, clients should start to consider the old (unencrypted) slot invalid, and it's the responsibility of the slot creator / encryption enabler / someone with sufficient permissions to send a new `m.rtc.slot` event that includes an encryption object?
(that's all rather implicit then, but again, not sure if it has to be specified)

Johennes: > That seem like a good addition. I'm not totally sure if it fully covers what I was asking though. Do you mean for this to imply that on an unencrypted->encrypted transition of a room, clients should start to consider the old (unencrypted) slot invalid, and it's the responsibility of the slot creator / encryption enabler / someone with sufficient permissions to send a new `m.rtc.slot` event that includes an encryption object?
> (that's all rather implicit then, but again, not sure if it has to be specified)

Yes, that was exactly what I had in mind. Do you have a suggestion for spelling it out a little better?

gewitternacht: I'm still not sure if this is needed at all, but maybe it's good to call this case out so that implementations don't forget to consider it – so maybe something like:

> Note that enabling encryption in a previously unencrypted room consequently invalidates previous `m.rtc.slot` events, and requires new `m.rtc.slot` events with an encryption object to be sent.

(but feel free to reject or adapt to your liking)

Johennes: Have tried to codify it with 5b861835b17cb523c2bdb7efee20e3e963a246c0.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,207 @@
+# MSC4196: Voice and video calling application for MatrixRTC
+
+[MSC4143] introduces MatrixRTC as an extensible framework for real-time communication in Matrix.
+MatrixRTC uses so called transports to transfer the RTC data between RTC members. Transports are
+then used in what MatrixRTC calls applications to build user experiences for concrete use cases.
```
turt2live: This looks copied from the MSC4195 introduction, which sets the reader up for another transport MSC. Some wording tweaks would be good to set the user up for an "application" MSC instead.

Non-blocking for purposes of FCP - we can equally fix it in the spec PR.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,207 @@
...
+
+This proposal introduces a MatrixRTC application for voice and video calling that is able to power
+a variety of calling use cases including but not limited to classical 1-on-1 and group calling
+as well as Discord-style [Voice Channels] (which, previously, were attempted to be introduced
+in [MSC3417]). The application is compatible with the [LiveKit] transport introduced in [MSC4195].
```
turt2live: Is MSC3417 now obsolete in favour of this MSC then? Not sure why it says "previously" when the MSC is still open.

Not blocking for FCP to start, but clarification required for FCP to finish.

proposals/4196-matrixrtc-m-call.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Proposal
+
+A new MatrixRTC application type `m.call` is introduced. For now, only a single instance of `m.call`
+per room is supported. This is sufficient for the majority of use cases and avoids the risk of two
+competing slots being opened for the same call when room administrators race.
+
+### Slot event
+
+The `m.call` application instance MUST use an application-specific slot ID of `room`. The full
```
turt2live:
```suggestion
The `m.call` application instance MUST use an application-specific slot ID literal of `room`. The full
```

(just to clarify that it's not a room ID)

Not really blocking for FCP to start, I think.

turt2live (approved):
Looks reasonable enough to start FCP tickybox gathering, I think. 

I haven't looked into the detail of the MSC too much - instead, I've spent my time looking at process requirements.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge
@mscbot concern Dependency - MatrixRTC (concurrent FCP permitted)

proposals/4196-matrixrtc-m-call.md
Johennes: #3417 has some dependencies on the pre-MatrixRTC group calling mechanism from #3401. I think it will need some updating but shouldn't be obsolete. Have tried to clarify it in 86f490a698d435b62b9cc076df5f24f0a1450fad.

proposals/4196-matrixrtc-m-call.md
Johennes: Yeah, good point. Have tried to rephrase it in 2407455cb5eccfd885aeb89b2e01b7cf5dbc7ea0.

PR #4197, opened 2024-09-22
MSC0000: Copy Paste Hints

anoadragon453:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/anoa/copy_hints/proposals/0000-copy-hints.md)


proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,63 @@
...
+An example is below:
+
+
+```json5
+{
+  "content": {
+    "body": "DO NOT SHARE THIS WITH ANYONE!!! Your 2FA code is: 100000",
+    "m.mentions": {},
+    "msgtype": "m.text",
+    "copy_hint": [23, 29]
```
KitsuneRal, line 23: The example seems to be out of line with the proposal?

KitsuneRal (changes requested):

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,63 @@
+# MSC4197: Copy-Paste Hints
```
KitsuneRal, line 1: Should we extend the scope and also consider removing/redacting the content after copying?

proposals/4197-copy-hints.md
anoadragon453: Oops! Fixed in e2f6046f.

proposals/4197-copy-hints.md
anoadragon453: Matthew says no, so....

anoadragon453: Because this would be in the realm of self-destructing messages (see [MSCXXXX]()).

Also I think this would give a false sense of security. Redactions already exist, [link to spec]().

But thank you for reading my MSC :) :heart:

KitsuneRal: Makes sense, thanks

anoadragon453: Feel free to open a new MSC which builds on top of this one.

florianjacob (changes requested):
I have a security concern regarding copying something else to clipboard than the user expects.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+Clients MAY automatically copy the contents of `copy_hint` to the clipboard without asking the user. This is an implementation detail.
+
+### Potential Issues
+
+None considered!
+
+### Alternatives
```
ara4n, line 44: might want to mention we're deliberately only letting you hint that a single thing is copyable, given you can only put one thing on a pasteboard.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+3. boolean `copy_hint` field (hint to client to copy the whole message)
+4. give up (similar to 2.)
+5. Use indexes in the message
+    * Harder to implement
+    * Use bandwidth
+6. If the event is encrypted, leave the 2FA code out of the encryption
+
+### Security Considerations
+
+TODO.
```
ara4n, line 57: massive spam risk here; need to explicitly advise client implementors to not allow 2FA spam to create an explosion of dialogs or similar

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+### Alternatives
+
+1. `hints` dictionary.
+2. Leave things as is (please don't!)
+3. boolean `copy_hint` field (hint to client to copy the whole message)
+4. give up (similar to 2.)
+5. Use indexes in the message
+    * Harder to implement
+    * Use bandwidth
+6. If the event is encrypted, leave the 2FA code out of the encryption
```
ara4n, line 53: A nice refinement could be delete-after copying semantics, but that should be an extra MSC - probably extending the existing MSC2228 self-destructing messages MSC

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+An example is below:
+
+
+```json5
+{
+  "content": {
+    "body": "DO NOT SHARE THIS WITH ANYONE!!! Your 2FA code is: 100000",
+    "m.mentions": {},
+    "msgtype": "m.text",
+    "copy_hint": "100000"
```
ara4n, line 23: please namespace the field, just as `m.mentions` is... - so `m.copy_hint` once merged

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+    * Use bandwidth
+6. If the event is encrypted, leave the 2FA code out of the encryption
+
+### Security Considerations
+
+TODO.
+
+### Unstable prefix
+
+While this MSC is unstable, all instances of the field name `copy_hint` must be replaced with `org.matrix.msc4197.copy_hint`.
```
ara4n, line 61:
```suggestion
While this MSC is unstable, all instances of the field name `m.copy_hint` must be replaced with `org.matrix.msc4197.copy_hint`.
```

proposals/4197-copy-hints.md
ara4n, line 23:
```suggestion
    "m.copy_hint": "100000"
```

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+In Matrix today, it can be used for communication. One thing that is communicated is two-factor auth codes. In other platforms, one convenience is being able to quickly copy-paste two factor auth codes. This is not possible in Matrix today.
+
+When sending a message, my message could contain a hint to the user's client that they should copy some particular part of the message. I propose to add a new field to `m.room.message` events that will hint to clients what text they could facilitating the user to copy.
+
+### Proposal
+
+* technical details
+* describe the solution (This is the solution. Be assertive.)
+
+`m.room.message` contains a new field, `copy_hint`, under the existing `content` dictionary. This new field will contain a string representing text that the client could present to the user to copy.
```
ara4n, line 12:
```suggestion
`m.room.message` contains a new field, `m.copy_hint`, under the existing `content` dictionary. This new field will contain a string representing text that the client could present to the user to copy.
```

ara4n, line 12: this should probably not be limited to m.room.message. it's effectively an extensible event mix-in to provide an alternative presentation of the message specifically for proposing copy-pasting.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+    "membership": "join",
+    "age": 54359693,
+    "transaction_id": "m1726936182657.27"
+  },
+  "event_id": "$4WvO6_skvEIibdffnDKdTkFHtOKUZaPFLm8HJuXcz7E",
+  "room_id": "!jWkHTegEyVsdPJkjHA:element.io"
+}
+```
+
+Clients MAY automatically copy the contents of `copy_hint` to the clipboard without asking the user. This is an implementation detail.
```
ara4n, line 38:
```suggestion
Clients MAY automatically copy the contents of `m.copy_hint` to the clipboard without asking the user. This is an implementation detail.
```

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Potential Issues
+
+None considered!
+
+### Alternatives
+
+1. `hints` dictionary.
```
ara4n, line 46: Rather than coming up with a new taxonomy for hints, it's better to use extensible events... i.e. you'd do an `m.other_hint_type` if you don't want an `m.copy_hint`.  Although it might be nice for them to have the same shape?

proposals/4197-copy-hints.md
realtyem, line 38: Please don't do this. Beyond the security ramifications of a client copying what it wants to my clipboard, is the fact it means a program can overwrite whatever is in there and I may not notice that it happened. I might have been copying a huge chunk of something important and irreplaceable and now it's gone because someone else triggered a 2FA that is registered to my account

proposals/4197-copy-hints.md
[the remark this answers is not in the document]
TheArcaneBrony: To justify this concern, here's some well-known implementations of this attack:
- This is already common on Discord: [discord.com's safety docs](https://discord.com/safety/common-scams-what-to-look-out-for#title-1)
- Certain types of scams [already do this](https://www.makeuseof.com/tag/anatomy-scam-windows-tech-support-con-examined/) with Windows' Run box ("Press Win+R, press CTRL+V and enter to read this file")

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
+# MSC4197: Copy-Paste Hints
+
+In Matrix today, it can be used for communication. One thing that is communicated is two-factor auth codes. In other platforms, one convenience is being able to quickly copy-paste two factor auth codes. This is not possible in Matrix today.
```
uhoreg, line 3: Please wrap your lines to about 80-120 characters

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+
+When sending a message, my message could contain a hint to the user's client that they should copy some particular part of the message. I propose to add a new field to `m.room.message` events that will hint to clients what text they could facilitating the user to copy.
+
+### Proposal
+
+* technical details
+* describe the solution (This is the solution. Be assertive.)
```
uhoreg, line 10: I think these two lines need to be dropped?

proposals/4197-copy-hints.md
uhoreg, line 46: Can you explain what "`hints` dictionary" is?

proposals/4197-copy-hints.md
uhoreg: I agree.  I don't want my client putting stuff in the clipboard without me telling it to.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. `hints` dictionary.
+2. Leave things as is (please don't!)
+3. boolean `copy_hint` field (hint to client to copy the whole message)
+4. give up (similar to 2.)
+5. Use indexes in the message
+    * Harder to implement
+    * Use bandwidth
+6. If the event is encrypted, leave the 2FA code out of the encryption
+
```
uhoreg, line 54: Another possibility would be to add a tag to a `<span>` element, say something like `Your 2FA code is <span data-mx-copy>123456</span>" which would allow the client to place a "Copy" button next to the text that is to be copied.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
...
+None considered!
+
+### Alternatives
+
+1. `hints` dictionary.
+2. Leave things as is (please don't!)
+3. boolean `copy_hint` field (hint to client to copy the whole message)
+4. give up (similar to 2.)
+5. Use indexes in the message
```
uhoreg, line 50: What do you mean by "indexes"?

proposals/4197-copy-hints.md
dbkr, line 38: Looks like there are plenty of problems pointed out with this but nobody suggesting solutions, but it feels like all the issues are readily solvable by just changing this to suggest something like searching for that text in the message and decorating it with a little copy button.

proposals/4197-copy-hints.md
```diff
@@ -0,0 +1,65 @@
+# MSC4197: Copy-Paste Hints
```
dbkr, line 1: Could possibly be more general, ie. it's just picking out the pertinent information from a message rather than necessarily anything to do with copy/paste.

proposals/4197-copy-hints.md
TheArcaneBrony, line 23: How would I be able to add multiple copy hints? Do I post extra empty messages per hint? Can I set the text to display on the copy button? Thinking for example, a moderation bot might want to provide a way to copy a list of MXIDs, it would be nice to able to say "Copy table", "Copy User IDs", ...

Additionally, i wonder if this might be better implemented using a custom HTML tag/property, eg. `<p>Your 2FA code is <span data-mx-copy="100000">100000</span></p>`, with ommission of a value copying the innerText.

PR #4198, opened 2024-09-24
MSC4198: Usage of OIDC login_hint

tonkku107:
[Rendered](https://github.com/tonkku107/matrix-spec-proposals/blob/oidc-login-hint/proposals/4198-oidc-login-hint.md)

**Implementations:**
- Authorization code flow
  - Homeserver / Authentication service:
    - https://github.com/element-hq/matrix-authentication-service/pull/3343
  - Client:
    - Element X
      - https://github.com/element-hq/element-x-ios/pull/4108
      - https://github.com/element-hq/element-x-android/pull/4752
      - https://github.com/element-hq/element-x-android/pull/7500
 - Device authorization flow
    - Homeserver / Authentication service:
      - https://github.com/element-hq/matrix-authentication-service/pull/5982
    - Client:
      - TBD
 - Account management requests
   - Homeserver / Authentication service:
     - https://github.com/element-hq/matrix-authentication-service/pull/5516
   - Client:
     - TBD
 - Metadata
   - Homeserver / Authentication service:
     - https://github.com/element-hq/matrix-authentication-service/pull/5981

### DIFF
--- a/proposals/4198-oidc-login-hint.md
+++ b/proposals/4198-oidc-login-hint.md
@@ -38,27 +38,29 @@ Example valid hint value: `mxid:@example-user:example.com`
 
 ### Usage
 
-A client can start the login flow by asking the user for their MXID.
-It can then parse the domain, discover the homeserver and its auth metadata ([OAuth 2.0 API Server metadata discovery]),
+A client MAY start the login flow by asking the user for their MXID.
+It SHOULD then parse the domain, discover the homeserver and its auth metadata ([OAuth 2.0 API Server metadata discovery]),
 register itself with the homeserver ([OAuth 2.0 API Client registration])
 and send the user to the authorization endpoint ([OAuth 2.0 API Login flow]), all in one step.
 
-To improve the UX of this flow, the MXID may be sent to the homeserver with the authorization request in the `login_hint`
-query parameter, following the format specified above using the `mxid` hint type.
+To improve the UX of this flow, the MXID MAY be sent to the homeserver with the authorization request in the OPTIONAL
+`login_hint` query parameter from [OpenID Connect Core 1.0], following the format specified above using the `mxid` hint
+type.
 
 Despite the `login_hint` parameter being defined in the OpenID Connect specification, homeservers supporting this proposal
-must handle the parameter even without the `openid` scope.
+MUST handle the parameter even without the `openid` scope.
 
-The homeserver should then assist the user to complete the login flow with the correct account.
+The homeserver SHOULD then assist the user to complete the login flow with the correct account.
 
-The client must be prepared to handle a case where the account that the user signs into will not be the one that was
-initially suggested, especially if the homeserver does not support the `login_hint` parameter and/or
-the user mistakenly uses the wrong credentials.
-The client may inform the user about ending up on a different account than intended and present an option to try again.
+The client MUST be prepared to handle a case where the account that the user signs into will not be the one that was
+initially suggested, especially if the homeserver does not support the `login_hint` parameter and/or the user mistakenly
+uses the wrong credentials.
+The client MAY inform the user about ending up on a different account than intended and present an option to try again.
 
 ### Example authorization request
 
-Expanding on the example authorization request shown in [OAuth 2.0 API Login flow] (broken down into multiple lines for readability),
+Expanding on the example authorization request shown in [OAuth 2.0 API Login flow] (broken down into multiple lines for
+readability),
 with the following additional parameters:
 
 - `login_hint` set to `mxid:@example-user:example.com`
--- a/proposals/4198-oidc-login-hint.md
+++ b/proposals/4198-oidc-login-hint.md
@@ -3,11 +3,14 @@
 This proposal builds on the [OAuth 2.0 API](https://spec.matrix.org/v1.15/client-server-api/#oauth-20-api) that was
 added in v1.15 of the spec.
 
-[OpenID Connect Core 1.0] specifies an optional `login_hint`
-authorization request parameter but leaves the contents of the parameter vague and "to the OP's discretion".
+There are times when a client already knows or has asked for the user's identity before redirecting them to the homeserver's
+account management interface to log in or to manage their devices. It would be useful to pass this information along to
+the homeserver with these requests.
 
-This MSC aims to define a format for use in the context of Matrix which allows for a login flow that starts with the
-user typing their [MXID].
+[OpenID Connect Core 1.0] specifies an optional `login_hint` authorization request parameter which is perfect for this
+purpose but it leaves the contents of the parameter vague and "to the OP's discretion".
+
+This MSC aims to define a format for this parameter for use in the context of Matrix for interoperability.
 
 
 ## Proposal
@@ -36,12 +39,15 @@ The prefix for this hint is `mxid` (all lowercase) and the value is the Matrix u
 
 Example valid hint value: `mxid:@example-user:example.com`
 
-### Usage
+### Usage in authorization requests
 
 A client MAY start the login flow by asking the user for their MXID.
 It SHOULD then parse the domain, discover the homeserver and its auth metadata ([OAuth 2.0 API Server metadata discovery]),
 register itself with the homeserver ([OAuth 2.0 API Client registration])
 and send the user to the authorization endpoint ([OAuth 2.0 API Login flow]), all in one step.
+
+Alternatively, this MAY also be used when the user is already logged in and the client is requesting additional scopes
+(when more granular scopes are defined in a future proposal).
 
 To improve the UX of this flow, the MXID MAY be sent to the homeserver with the authorization request in the OPTIONAL
 `login_hint` query parameter from [OpenID Connect Core 1.0], following the format specified above using the `mxid` hint
@@ -57,7 +63,7 @@ initially suggested, especially if the homeserver does not support the `login_hi
 uses the wrong credentials.
 The client MAY inform the user about ending up on a different account than intended and present an option to try again.
 
-### Example authorization request
+#### Example authorization request
 
 Expanding on the example authorization request shown in [OAuth 2.0 API Login flow] (broken down into multiple lines for
 readability),
@@ -82,6 +88,16 @@ With the line breaks removed and values properly encoded:
 ```
 https://account.example.com/oauth2/auth?client_id=s6BhdRkqt3&response_type=code&response_mode=fragment&redirect_uri=https%3A%2F%2Fapp.example.com%2Foauth2-callback&scope=urn%3Amatrix%3Aclient%3Aapi%3A*+urn%3Amatrix%3Aclient%3Adevice%3AAAABBBCCCDDD&state=ewubooN9weezeewah9fol4oothohroh3&code_challenge=72xySjpngTcCxgbPfFmkPHjMvVDl2jW1aWP7-J6rmwU&code_challenge_method=S256&login_hint=mxid%3A%40example-user%3Aexample.com
 ```
+
+### Usage in [MSC4191] account management
+
+We also add the `login_hint` query parameter for the [MSC4191] `account_management_uri` using the same format.
+
+For example, if `@example-user:example.com` wishes to delete the device `ABCDEFGH` where the advertised
+`account_management_uri` was `https://account.example.com/myaccount` the client could open a link to
+`https://account.example.com/myaccount?action=org.matrix.device_delete&device_id=ABCDEFGH&login_hint=mxid%3A%40example-user%3Aexample.com`.
+
+Similarly to the authorization request, the homeserver SHOULD assist the user to the account the action was intended for.
 
 ### Examples of homeserver assistance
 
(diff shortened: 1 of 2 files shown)


proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,131 @@
...
+
+A client can start the login flow by asking the user for their [MXID].
+The client can then parse the domain, discover the homeserver and its auth issuer ([MSC2965]),
+register itself with the homeserver ([MSC2966])
+and send the user to the authorization endpoint ([MSC2964]), all in one step.
+
+To improve the UX of this flow, the MXID can be sent to the homeserver with the authorization request in the `login_hint`.
+In order to comply with the OpenID Connect specification, the requested scope must also include the `openid` scope.
+
+The value of `login_hint` should be set to the user's MXID, prefixed with `mxid:` (e.g. `mxid:@example-user:example.com`).
```
Johennes, line 22: I'm trying to understand if the prefix is meant to be mandatory or not. Is "should" here the RFC2119 SHOULD and does it relate to inserting the MXID, the `mxid:` prefix or both?

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,131 @@
...
+Prefixing the value allows for more hint value types to be added in the future and for easy detection between them.
```
Johennes, line 23: Just curious, do you already foresee any types of identifiers that would require differentiation via a prefix?

clokep: Maybe you'd want email in some situations?

proposals/4198-oidc-login-hint.md
tonkku107: The prefix is meant to be mandatory, but the inclusion of a value at all is optional, just like in openid if that makes sense. I didn't strictly use RFC2119 keywords (which is why they're not in caps) but I ended up writing in that style, probably thanks to spending too much time reading RFCs

proposals/4198-oidc-login-hint.md
Johennes: Ok, makes sense. Might help to spell this out in the proposal.

proposals/4198-oidc-login-hint.md
Johennes: Ok, gotcha and that sounds sensible. I probably just read it a little to hairsplitting.

proposals/4198-oidc-login-hint.md
Johennes: It's possibly just me and I might be overthinking this but I couldn't find any common standards around login hint prefixes so would have welcomed some verbosity of why we need them or what grammar they might require.

proposals/4198-oidc-login-hint.md
tonkku107: The latest commit should clarify this

Johennes: Thank you. 👍

proposals/4198-oidc-login-hint.md
tonkku107: Yeah, I agree the grammar should be explained in more detail, even if it's just `<anything>:<something>`. I do feel like that one sentence already explains why you would want prefixes quite well but maybe I'm overlooking something.

Johennes: I just failed to think of a colliding example because I didn't consider ActivityPub. But if I'm the only one, then no need making that explicit. 😅

tonkku107: I've done some restructuring to add more detail for the format

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,156 @@
...
+
+The homeserver could filter the login methods presented to only the ones the user has available to them.
+This can help if the user has forgotten if they set a password or used a social login.
+
+**Note:** See security consideration #1.
+
+#### Passing a login_hint to upstream providers
+
+To take full advantage of the benefits of `login_hint`, a suitable value could be forwarded to an upstream provider.
+The vague nature of the values for `login_hint` could make this difficult to implement, however.
```
pixlwave, line 105: Reading this, I'm questioning whether the `mxid:` prefix is helpful or a hinderance.

If for example an upstream provider is expecting a plain username (or their own internal identifier format), what would be the trade-off with supplying the raw hint without a prefix and letting whichever provider (upstream or not) parse that (or throw) however it sees fit? When you consider that the prefix is a hint for a hint, that starts to set alarm bells ringing for me.

The specific use case I'm thinking of here is that some clients already support onboarding by opening a URL such as `client.com/login?homeserver=myserver.net`. If the client wanted to add an additional `login_hint=some_username` query parameter then the `mxid:` adds the following friction points:
- As we add more supported ID types, each server admin would need to check the spec to see if the type they're using is supported before provisioning onboarding links.
- If the type of ID they would like to use doesn't exist (or is hard to derive for the existing types) then they'll need to open a new proposal to add it (which would block onboarding their users).
- Ultimately it seems like someone would then create an `other:` prefix as a catch all, at which point we may as well drop the prefix and just let the provider interpret the hint however it sees fit.

tonkku107: It is not expected that the upstream provider supports the `mxid` hint. It'd have to be up to the person configuring the upstream provider on their homeserver to figure out what type of value the upstream provider expects.

pixlwave: Discussed this with @sandhose, the conclusion was that this MSC doesn't explicitly prevent a value without the `mxid:*` hint so there isn't a concern here :)

proposals/4198-oidc-login-hint.md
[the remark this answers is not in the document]
HarHarLinks: > Authentication service

well... technically just "server"? nobody is prescribing the auth to happen in an extra microservice, right?

noelportillo (commented):
First thing that came out

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,159 @@
...
+value = 1*VCHAR
+```
+
+### The `mxid` hint
+
+This MSC specifies a hint that contains the user's [MXID].
+The prefix for this hint is `mxid` (all lowercase) and the value is the Matrix user ID as specified in the
+[User Identifiers][MXID] section.
+
+Example valid hint value: `mxid:@example-user:example.com`
```
hughns, line 39: A potential point of confusion is that this looks similar to, but is different from, the `matrix:` [URI scheme](https://spec.matrix.org/v1.15/appendices/#matrix-uri-scheme).

An alternative could be to use the `matrix:` URI scheme. e.g. `matrix:@foo:example.com`

Also, can we define an email one? This could be `mailto:foo@example.com` or were you thinking `email:foo@example.com`?

hughns (changes requested):
I suggest:

- Updating to reflect that OAuth 2.0 API now exists in the spec
- Making it explicit that this is an optional parameter for the Authorization code grant
- Propose that that the `login_hint` can also be added to the MSC4191 account management URL query parameters (perhaps prefixed as `org.matrix.msc4198.login_hint`)

proposals/4198-oidc-login-hint.md
tonkku107: The correct `matrix:` URI for a user would be `matrix:u/foo:example.com`.
I think it would be strange to define this as a `matrix:` URI because:
- This is not a URI
- The URI does way more than just point to a user, and all the other types are not relevant to what `login_hint` does.

We could define an email prefix here or in a separate MSC later, but I don't see many matrix clients using the user's email at this point (though it seems Element might be doing something like that internally?)

hughns:
Looking at this again now, I think matrix-rust-sdk supports this for the authorization flow. e.g. https://github.com/matrix-org/matrix-rust-sdk/pull/5012 which is in turn used by https://github.com/element-hq/element-x-ios/pull/4108 and https://github.com/element-hq/element-x-android/pull/4752

Whilst the isn't an implementation for the account management parameter, perhaps that isn't actually needed for the MSC to progress?

hughns (approved):

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,180 @@
...
+
+If multiple concurrent sessions are supported, the homeserver could automatically switch to the correct session before
+showing the consent screen or account management interface.
+
+## Potential issues
+
+TBD
+
+
+## Alternatives
```
hughns, line 180: Another obvious alternative is to use the bare MXID: `login_hint=@example-user:example.com`.

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,180 @@
...
+purpose but it leaves the contents of the parameter vague and "to the OP's discretion".
+
+This MSC aims to define a format for this parameter for use in the context of Matrix for interoperability.
+
+
+## Proposal
+
+### Format
+
+In order to futureproof for new new types of hints in the future and to be able to identify and parse them easily,
```
hughns, line 19: How can a client determine which types of hint a server can support?

Should we add something like a `login_hint_types_supported[]` to the [discovery](https://spec.matrix.org/v1.15/client-server-api/#server-metadata-discovery) response?

e.g. Imagining a future MSC adds `email:example-user@example.com` and `localpart:example-user` hint types:

```
{
...
	"login_hint_types_supported": ["mxid", "email", "localpart"],
...
}
```

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,180 @@
...
+## Unstable prefix
+
+For the authorization request we are borrowing the existing OIDC parameter so an unstable prefix isn't necessary.
+
+For the account management request, `org.matrix.msc4198.login_hint` should be used instead of `login_hint` while in
+development.
+
+## Dependencies
+
+This MSC builds on [MSC4191] (which at the time of writing has not yet been accepted into the spec).
```
hughns: MSC4191 has now been accepted.

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,180 @@
...
+MUST handle the parameter even without the `openid` scope.
+
+The homeserver SHOULD then assist the user to complete the login flow with the correct account.
+
+The client MUST be prepared to handle a case where the account that the user signs into will not be the one that was
+initially suggested, especially if the homeserver does not support the `login_hint` parameter and/or the user mistakenly
+uses the wrong credentials.
+The client MAY inform the user about ending up on a different account than intended and present an option to try again.
+
+#### Example authorization request
```
hughns, line 69: Could we also make this apply to the [device authorisation flow](https://spec.matrix.org/v1.19/client-server-api/#device-authorisation-flow)?

i.e. in the Device Authorisation Request POST request allow the `login_hint` param to be passed.

proposals/4198-oidc-login-hint.md
tonkku107: That was technically accounted for with the mention of the domain being redundant to the server but now that I look at it again the section could use some rewording.

proposals/4198-oidc-login-hint.md
```diff
@@ -0,0 +1,180 @@
...
+The homeserver could filter the login methods presented to only the ones the user has available to them.
+This can help if the user has forgotten if they set a password or used a social login.
+
+**Note:** See security consideration #1.
+
+#### Passing a `login_hint` to upstream providers
+
+To take full advantage of the benefits of `login_hint`, a suitable value that the upstream provider expects could be
+supplied when logging in with one.
+Since this proposal only covers the Matrix protocol, upstream providers likely won't follow the format laid out in this
```
hughns, line 161: Maybe this suggests an alternative to consider: that we use a format that is more common with upstream providers?

If upstreams commonly accept a plain email or username then does that suggest we might want to as well? e.g. `login_hint=example-user@example.com` and `login_hint=example-user`

proposals/4198-oidc-login-hint.md
tonkku107: That's a good point. We probably need to call it `org.matrix.login_hint_types_supported` as it wouldn't make sense to officially register this like we did with `account_management_uri` and `account_management_actions_supported`

proposals/4198-oidc-login-hint.md
tonkku107: Most other parameters also get forgotten about for the device authorization flow in the oauth world. It might be worth looking into something in the future that defines common parameters available in every flow in the spec.

RFC8628 doesn't seem to make any mention of how to go about extending the parameters but it's probably fine to just add `login_hint` as is since that's what OIDC already did for the authorization code flow.

proposals/4198-oidc-login-hint.md
tonkku107: As stated in the introductory section, there is no defined format and every upstream is going to have their own idea of how to use it. There is no guarantee either that the email or username used on the upstream account matches the localpart or email set on someone's matrix account. It's best to leave what's sent to the upstream as a configurable option on the homeserver.

proposals/4198-oidc-login-hint.md
tonkku107: Done in 5f215af

proposals/4198-oidc-login-hint.md
tonkku107: Reworded in aa67dc0

proposals/4198-oidc-login-hint.md
tonkku107: Mentioned that in aa67dc0

proposals/4198-oidc-login-hint.md
tonkku107: Added in a6c1187

proposals/4198-oidc-login-hint.md
tonkku107: Added in 491d18b

PR #4201, opened 2024-09-25
MSCXXXX: Profiles as Rooms v2

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-spec-proposals/blob/FSG-Cat/Profiles-as-rooms/proposals/XXXX-profiles-as-rooms-v2.md)

### DIFF
--- a/proposals/XXXX-profiles-as-rooms-v2.md
+++ b/proposals/XXXX-profiles-as-rooms-v2.md
@@ -28,7 +28,6 @@ To get around the fact that we do not have federated peeking yet this proposal w
 to simply ask the users homeserver about their profiles. This proposal supports per room profiles and viewing
 restricted profiles.
 
-Restricted viewing is done via a check of is the person asking in a room where you are pointing to that profile.
 Public profiles without viewing restrictions are also considered valid as you want to be able to present
 a profile before you share a room if your privacy settings allow for this.
 
@@ -113,8 +112,9 @@ a room is shared `auth_event` must to be omitted.
 
 The user making the request must have history visibility for the `auth_event`
 
-A subset of the profile fields can be requested by putting an array of profile fields a request body. If omitted all
-fields are returned.
+A subset of the profile fields can be requested using the `filter` query parameter type being `string`.
+
+The response to a `GET /_matrix/client/v1/profile/{roomID}` is a set of Stripped state events responsive to the querry.
 
 To Access a profile over federation you use `GET /_matrix/federation/v1/profile/{roomID}`.
 
--- a/proposals/XXXX-profiles-as-rooms-v2.md
+++ b/proposals/XXXX-profiles-as-rooms-v2.md
@@ -226,6 +226,9 @@ MSC informed by Steam. They also have rich profile visibility settings and they 
 not copying all of their settings in this proposal with the right use of this proposal you can achieve mostly similar
 results.
 
+There is the concern that users can ACL ban the mods from their profile and this concern is lessened if moderators are 
+informed of this situation and simply disallow such profiles in their rooms.
+
 The Problems related to that this MSC allows you to essentially ask to be vomited data at are hopefully explained
 away to an acceptable degree.
 
--- a/proposals/XXXX-profiles-as-rooms-v2.md
+++ b/proposals/XXXX-profiles-as-rooms-v2.md
@@ -85,8 +85,8 @@ in a room that has a pointer to this profile.
 For a pointer to be valid said pointer must be issued by the creator of the profile if the profile is restricted.
 This means that invites can only set `public` profiles without causing the server to 404 the request.
 
-To use a given profile you set a `profile` value in your `m.room.member` event for a given room.
-The value of `profile` is the room ID of the profile you want to use in that room.
+To use a given profile you set a `m.profile` value in your `m.room.member` event for a given room.
+The value of `m.profile` is the room ID of the profile you want to use in that room.
 
 `m.profile` events are limited only to all the normal event size rules but clients are allowed to enforce whatever
 limits they find resonable on data that they will render when given this data.
@@ -239,6 +239,7 @@ While this proposal is unstable the following substitutions are to be made.
 `m.profile.privacy` becomes `support.feline.mscXXXX.profile.privacy.v1`
 `m.profile` state event becomes `support.feline.mscXXXX.profile.v1`
 `m.profile` room type becomes `support.feline.mscXXXX.profile`
+`m.profile` value in `m.room.member` becomes `support.feline.mscXXXX.profile.v1`
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


proposals/4201-profiles-as-rooms-v2.md
tulir, line 1: Implementation requirements:

* [ ] Server
* [ ] Client

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+  - [Security considerations](#security-considerations)
+  - [Unstable prefix](#unstable-prefix)
+  - [Dependencies](#dependencies)
+  - [Acknowledgements](#acknowledgements)
+
+## Introduction
+
+Profiles as rooms is a long established dream in the matrix ecosystem that has been declared as the solution
+to a whole bunch of our problems. This MSC aims to realise that dream via packaging up a idea that has been circulating
+in the spec room. Just abuse /profile and the nature of profiles to bypass the Peeking problems.
```
clokep, line 22: This fails to answer "Why?" -- it being a desire isn't enough. Why is this helpful?

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+Public profiles without viewing restrictions are also considered valid as you want to be able to present
+a profile before you share a room if your privacy settings allow for this.
+
+## Proposal
+
+### Profile Rooms
+
+Profile rooms are special rooms with the dedicated purpose of storing profiles. These rooms will be created
+with the `type` being `m.profile` in line with [MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772)
+established precedent on that specialised rooms like these are what room types are for.
```
clokep, line 40: Please link to the spec for merged features.

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+
+Query Parameter of `auth_event` pointing to a `m.room.member` that has a `m.profile` field that matches the profile
+being requested. If the profile is public like in a use case where your presenting profile information before
+a room is shared `auth_event` must to be omitted.
+
+The user making the request must have history visibility for the `auth_event`
+
+A subset of the profile fields can be requested using the `filter` query parameter type being `string`.
+
+The response to a `GET /_matrix/client/v1/profile/{roomID}` is a set of Stripped state events responsive to the query.
```
clokep, line 117: I assume this is something like `filter` is the state keys of events in the profile room? This isn't clear though. The response is the stripped `m.profile` state events?

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Alternatives
+
+The primary alternative to this MSC is [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+
+This MSC sees it self as the better alternative if perfected because it brings Per room and private profiles
+and a path to the original vision of Profiles as Rooms that use Federated peeking. This proposal brings
+Trust and Safety as a Nr 1 priority but ofc this is a very hard thing to deal with as we are essentially
+bringing something akin to a steam profile with this MSC. That's very intentional as Steam profiles are a
+inspiration for this MSC UX wise.
```
clokep, line 202: I don't really see how this helps T&S concerns TBH. Why is it easier to redact events in a room than to update a profile?

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Security considerations
```
clokep, line 204: Having all historical data via state events sounds like a big privacy violation that would be surprising to users?

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+## Proposal
+
+### Profile Rooms
+
+Profile rooms are special rooms with the dedicated purpose of storing profiles. These rooms will be created
+with the `type` being `m.profile` in line with [MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772)
+established precedent on that specialised rooms like these are what room types are for.
+
+They are plain old normal matrix rooms except for that they store profile information as their defined
+purpose and server implementations or future MSCs are free to provide powerlevel templates that reflect this purpose.
```
clokep, line 43: What happens if other events are sent in this room? I'm confused at who is expected to be joined to this room.

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+
+These rooms contain `m.profile` state events to store profile data as the state key determines
+the profile field this data is for.
```
clokep, line 46: An example of this event would be useful!

proposals/4201-profiles-as-rooms-v2.md
```diff
@@ -0,0 +1,264 @@
...
+}
+```
+
+`visibility` of `public` means that anyone is allowed to look up this profile. This value is primarily intended for
+your global profile for when you do not yet share a room with the subject. Like when they are doing a lookup in
+the user directory when starting a DM.
+
+`visibility` of `restricted` restricts who can look up a profile to users who are
+in a room that has a `m.profile` value set to this profile.
+For this value to be legal it must be sent by the creator of the profile.
```
clokep, line 75: I was confused that this is referring to the sender of the `m.room.member` as opposed to in the `m.profile.privacy` event.

proposals/4201-profiles-as-rooms-v2.md
FSG-Cat: Yup you filter based on the state keys you care about and the response is stripped `m.profile` events responsive to the query.

proposals/4201-profiles-as-rooms-v2.md
FSG-Cat: Its not clarified properly but the parties expected to be present in ones profile room is traditionally only the user them self and optionally bots that can update their profile if that isnt handled via scoped access tokens.

proposals/4201-profiles-as-rooms-v2.md
FSG-Cat: Yes this is a bit confusing i can agree. 

The intent is essentially that only the profile owner can create valid pointers for non public profiles.

proposals/4201-profiles-as-rooms-v2.md
TheArcaneBrony: I, as a user on homeserver A, cannot update the profile of a user on homeserver B.

proposals/4201-profiles-as-rooms-v2.md
imbev: > Its not clarified properly but the parties expected to be present in ones profile room is traditionally only the user them self and optionally bots that can update their profile if that isnt handled via scoped access tokens.

Without reading this comment, it's not clear that the room should be joined only by the user and bots. Maybe that should be explicitly specified in the MSC?

PR #4202, opened 2024-09-26
MSC4202: Reporting User Profiles

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/user-reporting/proposals/4202-user-reporting.md)

Signed-off-by: Tom Foster [tom@tcpip.uk](mailto:tom@tcpip.uk)

---

Known Implementations:
- Clients:
- Servers:

----

SCT Stuff:

No MSC checklist

[FCP **close** tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4202#issuecomment-4453200882)

### DIFF
--- a/proposals/4202-user-reporting.md
+++ b/proposals/4202-user-reporting.md
@@ -2,94 +2,108 @@
 
 *This proposal extends [MSC3843](https://github.com/matrix-org/matrix-spec-proposals/pull/3843) to
 allow reporting user profiles over federation. Clients can offer a user interface feature, such as
-a button at the bottom of a user's profile, enabling users to report another user's profile to the
-administrators of the server from which the profile was accessed.*
+a button on a user's profile, enabling users to report another user's profile to the administrators
+of their own homeserver, which can then forward the report over federation if necessary.*
 
 ## Proposal
 
 Building upon the reporting mechanism proposed in
-[MSC3843](https://github.com/matrix-org/matrix-spec-proposals/pull/3843) and the user profile
-information introduced in [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133),
-this MSC proposes modifying the reporting endpoint to allow the use of a Matrix User ID (MXID) in
-place of an event ID. This enables users to report user profiles directly to the homeserver of the
-reported user.
+[MSC3843](https://github.com/matrix-org/matrix-spec-proposals/pull/3843), this MSC proposes that
+clients offer an option in user profiles to report the most recent `m.room.member` event of the
+user in a shared room. This leverages the existing client-server and federation APIs for reporting
+events, allowing servers to handle profile reports without any changes to the APIs.
 
-### Extended Federation Endpoint
+### Client Behaviour
 
-The existing federation endpoint for reporting events is:
+Clients SHOULD offer a user interface element in user profiles (e.g., a "Report User" button) that
+allows users to report problematic profiles. When reporting a profile, clients SHOULD:
 
-```
-POST /_matrix/federation/v1/rooms/{roomId}/report/{eventId}
-```
+- Identify a room (`room_id`) that both the reporting user and the reported user share. This could
+  be any mutual room, including a direct message room. If no mutual room exists, clients MAY
+  display an error to the user indicating that the profile cannot be reported through this mechanism.
 
-This MSC proposes extending this endpoint to allow reporting user profiles by accepting a user ID
-in the `eventId` position, identified by the `@` prefix:
+- Obtain the most recent `m.room.member` event (`event_id`) for the reported user in that room. This
+  event contains the profile information (display name, avatar URL) set by the user in that room.
 
-```
-POST /_matrix/federation/v1/rooms/{roomId}/report/{userId}
-```
+- Use the existing client-server endpoint `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}`
+  to report the `m.room.member` event. The request body SHOULD include:
 
-Where `userId` is a Matrix User ID in the format `@user:domain`.
+  - A `reason` provided by the user, explaining why the profile is being reported.
+  
+  - A `score` indicating the severity, as per the existing specification, though servers MAY ignore
+    this parameter.
+
+  - If reporting content in the user's profile, it is recommended for the client include an MXC URI
+    for a screenshot of the profile (including MXID and offending content) to aid investigation.
 
 #### Example Request
 
 Reporting a user's profile:
 
 ```
-POST /_matrix/federation/v1/rooms/!roomid:example.org/report/@baduser:badserver.com
+POST /_matrix/client/v3/rooms/!roomid:example.org/report/$eventid
+Content-Type: application/json
+
 {
-  "reason": "Inappropriate profile content: mxc://example.org/abcd1234"
+  "reason": "Inappropriate profile content: mxc://example.org/abcd1234",
+  "score": 0
 }
 ```
 
-### Request Body Parameters
+### Homeserver Behaviour
 
-The request body includes the following parameters:
+Upon receiving the report, the homeserver SHOULD process it according to its moderation policies.
+If the event being reported originated from a remote homeserver, and if the homeserver supports
+federation reporting as per MSC3843, it MAY forward the report to the remote homeserver using the
+federation endpoint:
 
-- `reason` (string, **required**): A description explaining why the profile is being reported.
-  This cannot be blank. It is *strongly* recommended that clients include the MXC URI of a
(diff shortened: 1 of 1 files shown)


tcpipuk:
(Oops: fixing the unwanted content!)

tcpipuk:
This MSC is dependent on https://github.com/matrix-org/matrix-spec-proposals/pull/3843 to allow users to report inappropriate users (e.g. offensive MXID, or inappropriate user profile content) rather than a specific event.

proposals/4202-user-reporting.md
tulir, line 1: Implementation requirements:

* [ ] Server (receiving and sending reports)

proposals/4202-user-reporting.md
```diff
@@ -0,0 +1,135 @@
...
+While this MSC is not yet stable, implementations SHOULD use the following endpoint:
+
+```
+POST /_matrix/federation/unstable/uk.tcpipuk.msc0000/rooms/{roomId}/report/{userId}
+```
+
+## Dependencies
+
+- [MSC3843](https://github.com/matrix-org/matrix-spec-proposals/pull/3843): Reporting content over
+  federation
```
tulir, line 135: This would also depend on someone adding a "report user" endpoint to the client-server API

proposals/4202-user-reporting.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Potential Issues
+
+- **Spam Reports**: Servers may receive a high volume of unnecessary or malicious reports. Rate
+  limiting and appropriate error responses can help mitigate this issue.
+- **Cross-Domain Trust**: Servers need to decide how much trust to place in reports received from
+  other homeservers. Policies may vary between servers.
+
+## Alternatives
+
```
tulir, line 118: Instead of defining a new endpoint, one could simply report the profile event of the user, especially since the endpoint proposed here is scoped to a room.

A non-room-scoped user report endpoint may make sense for reporting when you don't have shared rooms though

tulir: Oh, MSC3843 already defines the former:

> To report a user within the context of that room the caller would use the `m.room.member` event for that user.

tcpipuk: Totally agree! It makes no sense to add an extra endpoint for this, so I've now updated to make this a change in client UI/behaviour rather than any direct API changes.

proposals/4202-user-reporting.md
tcpipuk: Agreed, I've removed this requirement now.

proposals/4202-user-reporting.md
tcpipuk: @tulir would it be ok to amend this requirement following the removal of the API change?

proposals/4202-user-reporting.md
```diff
@@ -0,0 +1,141 @@
...
+- **Abuse**: This mechanism could be abused to send spam reports. Homeservers SHOULD implement rate
+  limiting and MAY require additional verification steps before acting on reports.
+
+- **Privacy**: Reporting an `m.room.member` event includes the profile information set by the user
+  in that room. Servers SHOULD handle this data appropriately and respect user privacy.
+
+## Potential Issues
+
+- **Lack of Mutual Rooms**: If the reporting user and the reported user do not share any rooms, the
+  client cannot report the profile using this method.
```
Johennes, line 123: It might be worth expanding how users could be exposed to harmful profiles without sharing a room. I guess only via the user directory? Understanding that might help evaluate how a dedicated endpoint to report user IDs (akin to https://github.com/matrix-org/matrix-spec-proposals/pull/4151) would compare to this proposal.

proposals/4202-user-reporting.md
```diff
@@ -0,0 +1,141 @@
...
+```
+POST /_matrix/federation/v1/rooms/{roomId}/report/{eventId}
+Content-Type: application/json
+
+{
+  "reason": "Inappropriate profile content"
+}
+```
+
+The homeserver SHOULD ensure that the `event_id` being reported corresponds to an `m.room.member`
```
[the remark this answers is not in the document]
tcpipuk: I've fixed that now: https://github.com/tcpipuk/matrix-spec-proposals/commit/d2c4373f67f5cfe1b087bb5e34814a941a07dea0

proposals/4202-user-reporting.md
tcpipuk: Thanks! I've now added to both Potential Issues and Alternatives regarding this: https://github.com/matrix-org/matrix-spec-proposals/pull/4202/commits/605bd83b4f67d0072453f8ee173f7da95b659243

tcpipuk: Sorry, I've also extended this slightly to mention [MSC4151](https://github.com/matrix-org/matrix-spec-proposals/pull/4151)'s solution to this as well: https://github.com/matrix-org/matrix-spec-proposals/pull/4202/commits/534de9470079f63860c87fb1a711a99f504a5ec5

proposals/4202-user-reporting.md
```diff
@@ -0,0 +1,150 @@
...
+  federation. However, since the report is sent from the reporter's homeserver, it may still be
+  possible to infer their identity, especially if the homeserver has a small user base.
+
+- **Abuse**: This mechanism could be abused to send spam reports. Homeservers SHOULD implement rate
+  limiting and MAY require additional verification steps before acting on reports.
+
+- **Privacy**: Reporting an `m.room.member` event includes the profile information set by the user
+  in that room. Servers SHOULD handle this data appropriately and respect user privacy.
+
+## Potential Issues
```
MTRNord, line 120: One to add is that this might (as like with most moderation things on server level) have the issue that server and room rules may differ. A profile might be considered ok by the server admin, while it isn't ok for the room where the user is in. So this proposal only adds server level moderation but still no full room level moderation imho. It solves parts of the issues, but not all of them, imho.

proposals/4202-user-reporting.md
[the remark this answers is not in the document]
Johennes: > The homeserver SHOULD ensure that the event_id corresponds to an event in the room and SHOULD verify that the event was issued by the reported user's homeserver.

I think the second part of this sentence might still be a leftover? The reported user is only identified implicitly via the event ID pointing to a membership event. So I suppose there's no way the server could verify that event against the user that was intended to be reported?

For the first part of the sentence, I'm actually surprised this isn't already spelled out as a MUST in the [spec](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3roomsroomidreporteventid). I suppose this is what the 404 response was meant for. Would there be any cases in which the server should accept an event report for an event that's not in the room? 🤔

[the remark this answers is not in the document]
tcpipuk: Ah, I agree the wording is a bit awkward... I meant to say the server should verify it owns the user that created the event being reported. I'll have another go!

To your second point, I also agree. I suppose it's possible this server may not be aware of the event anymore (e.g. we lost our database) but we could backfill that event if necessary to verify that event comes from this room. I'd hope most implementations will perform those kind of sanity checks, but I didn't want to go so far as to apply a "must" to it.

proposals/4202-user-reporting.md
tcpipuk: Agreed, do you know if there's an MSC for reporting events to room admins so I could perhaps add that as a dependency here too?

proposals/4202-user-reporting.md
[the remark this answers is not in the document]
Johennes: > To your second point, I also agree. I suppose it's possible this server may not be aware of the event anymore (e.g. we lost our database) but we could backfill that event if necessary to verify that event comes from this room. I'd hope most implementations will perform those kind of sanity checks, but I didn't want to go so far as to apply a "must" to it.

Makes sense, yeah. 👍

turt2live:
With my Foundation T&S hat, we're looking at redesigning how reporting works in Matrix to support better communication and be more actionable. With that in mind, we've opened a proposal to generify the reporting API that we're currently working on implementing: https://github.com/matrix-org/matrix-spec-proposals/pull/4457

It's become clear as part of planning Reporting v2 that having dedicated endpoints for each type of report is unwieldy and leads to [inconsistencies](https://github.com/matrix-org/matrix-spec-proposals/pull/4277) with relative ease.

With my SCT hat, I'm suggesting we close this in favour of MSC4457 or something very near to it (single endpoint rather than many). 

@mscbot fcp close

Note: We typically try to reach out to the author out of band before proposing an MSC for closure, but in this case I couldn't find the author's account in the spec rooms to have that conversation. Proposal authors are not required to wait for FCP to start/finish before closing their proposals if they don't want to.

PR #3202, opened 2021-05-18
MSC3202: Encrypted appservices

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/otk-dl-appservice/proposals/3202-encrypted-appservices.md)

PR #2409, opened 2020-01-14
MSC2409: Proposal to send EDUs to appservices 

Sorunome:
[Rendered](https://github.com/Sorunome/matrix-doc/blob/soru%2Bhs/appservice-edus/proposals/2409-appservice-edus.md)

This MSC is a continuation of [MSC1888](https://github.com/matrix-org/matrix-doc/pull/1888) and thus deprecates it.

Implementations:
- SDKs
  - https://github.com/matrix-org/matrix-appservice-bridge/
  - https://github.com/matrix-org/matrix-appservice-node/
  - https://github.com/turt2live/matrix-bot-sdk 
- Bridges
  - https://github.com/matrix-org/matrix-appservice-irc
  - (and other matrix-appservice-bridges generally have support like Slack)
- Homeservers 
  - https://github.com/matrix-org/synapse/pull/8437
  - https://github.com/matrix-org/synapse/pull/11215

Signed-off-by: Sorunome <mail@sorunome.de>

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/2409#issuecomment-2380896215)

PR #4203, opened 2024-09-28
MSC4203: Sending to-device events to appservices

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/appservice-to-device/proposals/4203-appservice-to-device.md)

Part of #3202, split off from #2409

Implementations:

* Bridges: mautrix-*
* Server: https://github.com/matrix-org/synapse/pull/11215

### DIFF
--- a/proposals/4203-appservice-to-device.md
+++ b/proposals/4203-appservice-to-device.md
@@ -1,0 +1,107 @@
+# MSC4203: Sending to-device events to appservices
+This MSC is a part of adding end-to-end encryption support to application services.
+It was split off from [MSC2409]. The parent MSC for encrypted appservices is [MSC3202].
+
+[MSC2409]: https://github.com/matrix-org/matrix-spec-proposals/pull/2409
+[MSC3202]: https://github.com/matrix-org/matrix-spec-proposals/pull/3202
+
+## Proposal
+A new key, `to_device`, is added to the request body of the appservice API
+`/transactions` endpoint. The key contains a list of all to-device events
+sent to users who are covered by the namespaces of the appservice. If there are
+no new to-device events, the key can be omitted.
+
+In addition to the usual keys that are included in to-device events in the
+client-server `/sync` API, appservice to-device events include `to_user_id` and
+`to_device_id` fields to specify which user and device the event was sent to.
+
+```jsonc
+{
+  "to_device": [
+    {
+      "to_user_id": "@appservice_12345:example.com",
+      "to_device_id": "VPLQFIDCIT",
+      "type": "m.room.encrypted",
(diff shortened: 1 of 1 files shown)


proposals/4203-appservice-to-device.md
MTRNord, line 1: Not sure myself but is there any realtion with https://github.com/matrix-org/matrix-spec-proposals/pull/4190 ? I guess that one isnt about the to_device stuff itself but more on the login half of this all :thinking: Am I correct on that?

tulir: You need to be able to create devices to receive to-device events, but there's no direct relation, appservice login already exists in the spec. 4190 is just an appservice login replacement for next gen auth

turt2live, line 1: Implementation requirements:

* [x] Appservice
* [x] Server

tcpipuk: It looks like both of these exist now

tulir:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4203-appservice-to-device.md
```diff
@@ -0,0 +1,109 @@
...
+is not applicable to all scenarios (and thus is an implementation detail). The majority of known
+appservices use exclusive namespaces, which typically also means that those users will not be
+calling `/sync`. Because of this, the server may never get another opportunity to delete the
+messages until it has confirmed that the appservice received the transaction successfully. Inclusive
+namespaces are typically used when the appservice wants to impact a subset of users, but without
+controlling those users explicitly. Typically, inclusive users are also calling `/sync` and so
+the appservice should be CC'd on the to-device messages that would normally go down `/sync`.
+
+For inclusive namespaces, servers may also track who created a device, and not wait for `/sync`
+if the device was created using [appservice login] or [appservice device management].
```
tulir, line 70: Leaving this as an implementation detail might be wrong. I think the approach suggested here also won't work when transitioning from /sync to transactions. There are currently a lot of appservices with exclusive namespaces who will /sync to get their to-device events. If the server starts sending to-device events over /transactions unprompted, the appservices could get events double delivered (if the event is in both /transactions and /sync) or not delivered at all (if they ignore /transactions and the event gets deleted before reaching /sync).

MSC3202 has an `org.matrix.msc3202` opt-in flag, maybe it'd make sense to reuse that for to-device events (because all appservices using this MSC presumably already have that flag set) and also define a stable opt-in flag that encompasses both to-device events and other e2ee-related transactions data

PR #4204, opened 2024-10-02
MSC0000 `m.takedown` moderation policy recommendation

Gnuxie:
⚠️ This MSC is closely related to three other MSCS that are currently being opened, this MSC will be reviewable shortly once I have glued up all 4 PRs ⚠️

### DIFF
--- a/proposals/0000-takedown-recommendation.md
+++ b/proposals/0000-takedown-recommendation.md
@@ -1,0 +1,137 @@
+# MSC0000: `m.takedown` moderation policy recommendation
+
+Currently there is one specific moderation policy recommendation, `m.ban`[^spec-ban-recommendation].
+
+> When this recommendation is used, the entities affected by the rule
+> should be banned from participation where possible. The enforcement
+> of this is deliberately left as an implementation detail to avoid
+> the protocol imposing its opinion on how the policy list is to be
+> interpreted. However, a suggestion for a simple implementation is as
+> follows:
+
+[^spec-ban-recommendation]: https://spec.matrix.org/v1.11/client-server-api/#mban-recommendation
+
+In practice this recommendation is used by the moderation bots to
+store banned rooms, users and servers alongside a reason for the ban.
+
+This has provided the ability to create versatile policy lists that
+can ban matrix users for any recorded reason. Not only for spam,
+illegal content, csam, but softer reasons such as moderation disputes,
+timeouts, inrventions, disagreements.
+
+Unfortunately, because the `m.ban` recommendation has such a broad and
+consequently has the same power for any consequence,
+it is the expectation that a reason is provided for each `m.ban`.
+
+This reason inadvertently allows the subject of the moderation policy
+to be classified as an undesirable consequence. For example, a user called
+`@yarrgh:example.com` could be banned for the reason `piracy`. Which
+would identify `@yarrgh:example.com` as a pirate. Which is problematic.
+
+Moderation bots currently also use the `reason` of the ban to determine
+whether to redact a user's events or not. Which doesn't allow
+for a free-form use of the `reason` where redaction is required.
+
+## Proposal
+
+We introduce a new policy recommendation, `m.takedown`, reserved for
+purging entities and any associated content.  When `m.takedown` is
+used, a `reason` SHOULD NOT be embedded into the same policy event[^bluesky].
+
+The enforcement of the `m.takedown` recommendation deliberately left
+as an implementation detail to avoid the protocol imposing its opinion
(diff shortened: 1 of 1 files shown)


proposals/4204-takedown-recommendation.md
```diff
@@ -0,0 +1,159 @@
...
+A compounding issue is Moderation bots currently use the `reason` of
+the ban to determine whether to redact a user's events or not. Which
+doesn't allow for a free-form use of the `reason` where redaction is
+required.
+
+## Proposal
+
+We introduce a new policy recommendation, `m.takedown`, reserved for
+purging entities and any associated content.  When `m.takedown` is
+used, a `reason` SHOULD NOT be embedded into the same policy event[^bluesky].
```
nexy7574, line 49: Given the implication of a takedown being that all authored content, including messages, should be removed, it would be handy to include a reason in bans so that moderation tools like Draupnir and Meowlnir can auto-redact regardless, yet still include a reason for transparency. I propose that this section is removed or reworded to instead make `reason` *optional*, rather than *discouraged*.

This is useful and a great opportunity because currently, most policy lists dispatch the ban reason `spam` for auto-redaction purposes, which masks the original ban reason, leading to reduced transparency. However, we could use `m.takedown` as "always automatically redact", meaning moderators are free to actually just use whatever reason is relevant without worrying if list subscribers will auto-redact or not.

Gnuxie: So for that reason the use of reasons is discouraged and has to be just so that we don't end up with people making the mistake of classifying content.

Unless you can think of workarounds?

nexy7574: Classifying content was kinda my goal, so I won't think of a workaround with the issues highlighted. I'll probably work with mod bots in the ecosystem to make sure they just always redact for takedowns instead, at least. Thanks for the clarification :)

tcpipuk: This is a technical feature that can be used for many possible purposes, so it doesn't make sense to me that the reason field should be forbidden. 

For example, there could be takedown entries for deactivated accounts (or even servers) to encourage servers to remove content for GDPR purposes - this wouldn't be classifying illegal content, it would simply be stating the purpose of the takedown.

Gnuxie: It's not supposed to be used for many possible purposes, it's specific to taking down abuse. I can try make that more clear in the proposal if you like.

The reason field should be forbidden so that the people do not inadvertently classify content. Even through analyzing which policy rules have a reason property present or absent.

The recommendation is strongly worded as SHOULD NOT because we don't want developers to unaware of the implications of using reasons and get themselves or their users into trouble. Or worse end up doing something irresponsible. And yes, those implications are explained, but evidently people need lots of reinforcement before they learn something.

As for using takedown GDPR or privacy purposes, I don't think a reason field help distinguish that the policy is for someone securing their own privacy vs a moderator purging abuse. And I'm not sure why you would need to make a distinguishment either if you wanted to totally erase a user account or a piece of media or something else? However, If we think that there is a use case and that it needs to be distinguishable, I think we would be better off with a distinct recommendation that cannot be easily confused with takedown of abuse.

Including a reason purely for a privacy use case would weaken the proposal and likely cause a tonne of confusion for implementers. Considering the entire purpose of the recommendation is to avoid providing a reason and is what makes it distinct from `ban`. Please also see https://github.com/matrix-org/matrix-spec-proposals/pull/4207#discussion_r1784697744.

Gnuxie: I've updated the MSC to better explain the purpose of the MSC and thing such as removing the reason. I'm marking the thread as resolved, but please do add a new comment if there's anything that i've missed.

proposals/4204-takedown-recommendation.md
```diff
@@ -0,0 +1,184 @@
...
+
+- Applied to a user: The user and all associated messages, media, invitations
+  are hidden and removed from local storage or caches. If the user enforcing the
+  recommendation is a moderator in the room in common with the target, then
+  messages may be shown with a warning behind spoiler text.
+
+- Applied to a room: The user is banned from the room and all of their recent
+  messages redacted.
+
+- Applied to a server: The user is not allowed to send invites to users on the
```
timedoutuk, line 100: Does "not allowed" mean "forbidden" here (i.e. invite requests are 403'd), or is it more of a soft block?

Gnuxie: It's an implementation detail really and these are meant as loose guidance, not as requirements.

If you do want advice I would suggest 403

PR #4205, opened 2024-10-02
MSC4205: Hashed moderation policy entities

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/sha256-policy-entity/proposals/4205-sha256-policy-entity.md)

Implementations:

* Meowlnir: https://github.com/maunium/meowlnir/pull/4
* matrix-protection-suite: https://github.com/Gnuxie/matrix-protection-suite/pull/88
* Draupnir: https://github.com/the-draupnir-project/Draupnir/pull/761
    + Example protection: https://github.com/the-draupnir-project/Draupnir/pull/777

### DIFF
--- a/proposals/4205-sha256-policy-entity.md
+++ b/proposals/4205-sha256-policy-entity.md
@@ -1,40 +1,39 @@
 # MSC4205: Hashed moderation policy entities
 
-Currently, moderation policies describe the entity they are targeting
-by including a literal identifier in the `entity` field.
+Currently, moderation policies describe the entity they are targeting by
+including a literal identifier in the `entity` field.
 
 This is problematic for multiple reasons[^msc4204].
 
-[^msc4204]: While this MSC is not dependant upon
-[MSC4204](https://github.com/matrix-org/matrix-spec-proposals/pull/4204),
-[MSC4204](https://github.com/matrix-org/matrix-spec-proposals/pull/4204)
-provides a similar context for why hashing entities is desired.
-
+[^msc4204]:
+    While this MSC is not dependant upon
+    [MSC4204](https://github.com/matrix-org/matrix-spec-proposals/pull/4204),
+    [MSC4204](https://github.com/matrix-org/matrix-spec-proposals/pull/4204)
+    provides a similar context for why hashing entities is desired.
 
 #### Propagating abuse
 
-The literal entities can propagate abuse. For example,
-if the user
-`@i.hate.example.com:example.com` is banned, then the
-mxid will be embedded in the policy.
+The literal entities can propagate abuse. For example, if the user
+`@i.hate.example.com:example.com` is banned, then the mxid will be embedded in
+the policy.
 
-Additionally, users have been known to embed or masquerade URLs
-into their mxids.
+Additionally, users have been known to embed or masquerade URLs into their
+mxids.
 
 #### Identifying users before encountering them
 
-Policies can be used as an address book to identify problematic users
-who have not been encountered yet.
+Policies can be used as an address book to identify problematic users who have
+not been encountered yet.
 
-For example, if `@yarrgh:example.com` is banned for `piracy`,
-then it is obvious that `@yarrgh:example.com` could be a pirate.
-Even if the reason was not provided with the policy.
+For example, if `@yarrgh:example.com` is banned for `piracy`, then it is obvious
+that `@yarrgh:example.com` could be a pirate. Even if the reason was not
+provided with the policy.
 
 ## Proposal
 
-We therefore propose a new optional field `hashes` to the top level of
-all moderation policy events.  Embedded within this, we propose a
-simple `sha256` entity hash field.
+We therefore propose a new optional field `hashes` to the top level of all
+moderation policy events. Embedded within this, we propose a simple `sha256`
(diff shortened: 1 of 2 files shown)


proposals/4205-sha256-policy-entity.md
```diff
@@ -0,0 +1,94 @@
...
+Policies can be used as an address book to identify problematic users
+who have not been encountered yet.
+
+For example, if `@yarrgh:example.com` is banned for `piracy`,
+then it is obvious that `@yarrgh:example.com` could be a pirate.
+Even if the reason was not provided with the policy.
+
+## Proposal
+
+We therefore propose a new optional field `hashes` to the top level of
```
Johennes, line 35: The existing `entity` field is currently required. Would you not also have to change it to optional to make the hash masquerading functional?

proposals/4205-sha256-policy-entity.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+In this example, when a moderation tool encounters a new user, or a
+new policy, the tool will calculate the base64 encoded sha256
+of their full mxid `@yarrgh:example.com` to
+match against policies that provide an associated hash.
+
+## Potential issues
+
+None noted.
```
Johennes, line 58: I think hashing wouldn't help for glob-based rules?

proposals/4205-sha256-policy-entity.md
Gnuxie: Yep that's right

Gnuxie: Will this work for you? https://github.com/matrix-org/matrix-spec-proposals/commit/35049664369c6817f7794854a8cde2c5584d4f87#diff-62936d47ad39741850b1935807ef91072db5eccdbad8f3aba2f0bfd86eeeffb1R56-R58

proposals/4205-sha256-policy-entity.md
Gnuxie: Should I add anything else? https://github.com/matrix-org/matrix-spec-proposals/commit/35049664369c6817f7794854a8cde2c5584d4f87#diff-62936d47ad39741850b1935807ef91072db5eccdbad8f3aba2f0bfd86eeeffb1R63-R66

Johennes: No, I think this looks great. 👍

proposals/4205-sha256-policy-entity.md
Johennes: I think this looks good, thank you. 👍

proposals/4205-sha256-policy-entity.md
```diff
@@ -0,0 +1,102 @@
...
+    "hashes": {
+      "sha256": "VPqwbUV7mMMkOVto3kPwsNXXiALMs7VCKWh3OeqqjGs="
+    },
+	"recommendation": "m.takedown",
+  }
+}
+```
+
+In this example, when a moderation tool encounters a new user, or a
+new policy, the tool will calculate the base64 encoded sha256
```
tulir, line 52: Many places in matrix use url-safe and/or unpadded base64, so if this is standard, it should probably be mentioned explicitly

proposals/4205-sha256-policy-entity.md
```diff
@@ -0,0 +1,102 @@
...
+of their full mxid `@yarrgh:example.com` to
+match against policies that provide an associated hash.
+
+Currently the content schema for `m.policy.rule.user` requires the
+`entity` field. In order for the `entity` field to be omitted when a
+hash has been provided, the entity field will have to become optional.
```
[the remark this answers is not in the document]
Gnuxie: So, from my perspective there are policies in the wild that exist with omitted fields already, including the entity field. These were created by the now unmaintained spam police bot in an attempt to keep the context for a policy after unbanning a user.

The omission of entity, and the addition of the unstable hashes property, under the stable `m.policy.rule.room`, has not to my knowledge produced any significant problems for clients that have naive parsing of policy rules. Issues have also not yet been reported in implementation. But maybe there should be a requirement for a client to implement specific rendering for takedowns (gomuks might already) so that we can be sure.

proposals/4205-sha256-policy-entity.md
Gnuxie: Clarified in https://github.com/matrix-org/matrix-spec-proposals/pull/4205/commits/7fc49f27b224d1593d32edd74d333d8c484e4f02

proposals/4205-sha256-policy-entity.md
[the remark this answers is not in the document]
Gnuxie: > I think these events should have a distinct event type. Otherwise parsing the event is dependent on the recommendation, which both excludes ban recommendations from being hashed 

It's not tied to any recommendation. The schema of the policy is changed to make `entity` optional. That's the only change to parsing other than the addition of the `hashes` property. Parsing is not dependent on recommendation.

As for using the policy in an engine (not parsing). If `hashes` is present, then the hashes are used by the policy engine instead of the entity. If `hashes` is omitted then `entity` should be used instead and the policy should be treat as a literal.

PR #4206, opened 2024-10-02
MSC0000: Moderation policy auditing and context.

Gnuxie:
⚠️ This MSC is closely related to three other MSCS that are currently being opened, this MSC will be reviewable shortly once I have glued up all 4 PRs ⚠️

### DIFF
--- a/proposals/0000-policy-reason-auditing.md
+++ b/proposals/0000-policy-reason-auditing.md
@@ -1,0 +1,122 @@
+# MSC0000: Moderation policy auditing and context
+
+Moderation policies provide a `reason` field.
+While this is designed as a freetext field, there are a number
+of reasons why it is not appropriate to do so.
+
+* The `reason` is public, and moderators may not want to provide
+  context about the ban in public. But still need to keep the context
+  on record. They may also need to share context with other
+  moderators, but not the wider community.
+
+* Moderation bots such as Draupnir and Mjolnir use the `reason`
+  field in `m.policy.rule.user` to determine whether the user's
+  events and profile should be redacted. This stops the reason
+  field being used to provide context.
+
+## Proposal
+
+Standard Matrix, encrypted messaging threads are used to provide
+context about moderation policies.
+
+A new policy event `m.policy.rule.context` is introduced.  This is
+almost identical to `m.policy.rule.*` events, except it servers as an
+anchor point to open threads.  The reason why the existing policy is
+not used to act as an anchor point is so that context can be provided
+in a seperate, private, and encrypted policy room. And also so that
+when the original policy is rescinded or expires, the context is still
+available.
+
+```json
+"type": "m.policy.rule.context",
+"content": {
+  "recommendation": "m.ban",
+  "entity": "@yarrgh:example.com",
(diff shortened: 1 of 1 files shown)


proposals/4206-policy-reason-auditing.md
turt2live, line 1: Implementation requirements:
* [ ] Client sending context
* [ ] Client using context

proposals/4206-policy-reason-auditing.md
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+Neighnir then returns a link to the thread as a response
+to Alice's command.
+
+Alice can then send normal Matrix messages in the thread in
+order to provide any context, including screenshots.
+
+## Potential issues
```
Johennes, line 79: If `m.policy.rule.context` is also meant to be a state event, existing chat clients might not be able to render the context thread as they likely won't support threading on state events. This might not be a terrible problem other than being able to browse context rooms from a non-specialized client possibly being convenient.

proposals/4206-policy-reason-auditing.md
```diff
@@ -0,0 +1,116 @@
...
+
+`!neighnir context ban @yarrgh:example.com`.
+
+Neighnir then creates a new `m.policy.rule.context` policy
+in a pre-configured and private policy room. If a context
+policy already exists for the `recommendation` and `entity`
+combination, then this step is skipped.
+
+Neighnir then creates a threaded message referring to the
+policy context. This step is skipped if a thread is already open.
```
Johennes, line 60: Could it be expensive to find existing events in large policy lists given that you have to scan `content` on every context event? The bot could possibly work around that by keeping its own index.

proposals/4206-policy-reason-auditing.md
Gnuxie, line 79: For providing context to moderation policies with hashed entities it will be desirable to include the plain text entity in the context. This is so that the moderators of a common policy list (such as CME) can verify in advance what entity the policy is targeting before they "encounter" the entity themselves and unmask it.

proposals/4206-policy-reason-auditing.md
```diff
@@ -0,0 +1,116 @@
...
+
+Alice is a room moderator, and Neighnir is a moderation bot.
+
+Alice has recently banned `@yarrgh:example.com` and wants to provide
+some context for the ban to the other moderators in her community.
+
+Alice repeats the arguments to the ban command to Neighnir in
+the context command.
+
+`!neighnir context ban @yarrgh:example.com`.
```
tulir, line 52: This current flow would require 4 events: the policy itself, policy context root, start notice, actual context message(s)

I think it'd be better to attach an initial message into the context root itself (like a `reason` field) and create a thread later if discussion or additional info like images are necessary. I'd assume in most cases, moderators will just want to `!neighnir ban @yarrgh:example.com <private context here>`

proposals/4206-policy-reason-auditing.md
tulir: The state key should also be defined if it's meant to be a state event (I'd assume it should copy the policy event's state key)

PR #4208, opened 2024-10-02
Adding User-Defined Custom Fields to User Global Profiles

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/custom-fields/proposals/0000-user-defined-profile-fields.md)

---

Known Implementations:
- Clients:
- Servers:


tcpipuk:
This needs a little work done to it. I've exported the custom field requirements from [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133) and am updating this to reflect that it is a new/separate MSC.

proposals/4208-user-defined-profile-fields.md
```diff
@@ -0,0 +1,227 @@
+# MSC4208: Adding User-Defined Custom Fields to User Global Profiles
```
girlbossceo, line 1: For reference, conduwuit still has the initial implementation of custom fields that was initially in MSC4133. Will review and validate to this new MSC some time in the next few days.

proposals/4208-user-defined-profile-fields.md
```diff
@@ -0,0 +1,227 @@
...
+  enforce size limits on values.
+
+- Homeservers MAY restrict certain keys or values based on server policies.
+
+## Potential Issues
+
+- **Privacy Concerns**: Users need to be aware that custom profile fields are public and visible to
+  anyone who can access their profile.
+
+- **Abuse Potential**: As with any user-generated content, there is potential for misuse. Servers
```
jimmackenzie, line 186: This section needs beefing up considerably - including both client and server considerations for preventing the creation of obviously harmful custom fields, mechanisms for identifying harmful user generated content in these fields, and client controls for choosing whether to display all, some or none custom fields, with sensible defaults.

I'll aim to swing back with suggestions - happy to have conversations to help articulate the concerns more too!

tcpipuk: I totally agree! I've had a go at listing out some talking points (https://github.com/matrix-org/matrix-spec-proposals/pull/4208/commits/a22b9fce1f8b2e42834ad4349fcb24af528ded8e) but would definitely appreciate suggestions on what the Foundation would like to see enshrined in an MSC like this.

proposals/4208-user-defined-profile-fields.md
```diff
@@ -0,0 +1,227 @@
...
+  "u.bio": "I love Matrix!",
+  "u.location": "London"
+}
+```
+
+### Capabilities
+
+A new capability `m.custom_profile_fields` is introduced to control the ability to set custom
+profile fields. It is advertised on the `GET /_matrix/client/v3/capabilities` endpoint. Clients
+should check for this capability before attempting to create or modify custom fields.
```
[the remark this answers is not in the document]
tcpipuk: Thanks! I've had a go at that in the "Server Considerations" section: https://github.com/matrix-org/matrix-spec-proposals/pull/4208/commits/a22b9fce1f8b2e42834ad4349fcb24af528ded8e

proposals/4208-user-defined-profile-fields.md
```diff
@@ -0,0 +1,227 @@
...
+As per [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133), the profile API is
+extended to allow additional fields beyond `avatar_url` and `displayname`. This MSC defines the use
+of the `u.*` namespace for user-defined custom fields:
+
+- **Namespace `u.*`**: Reserved for user-defined custom fields. The portion of the key name after
+  the `u.` defines the display name of this field (e.g., `u.bio`). The values in this namespace
+  must always be UTF-8 strings with content not exceeding 512 bytes.
+
+### Client-Server API Changes
+
```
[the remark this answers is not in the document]
tcpipuk: That is a great point, I've just stripped that out, so we can prioritise the stuff that's going to be a bigger focus in this MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4208/commits/a22b9fce1f8b2e42834ad4349fcb24af528ded8e

proposals/4208-user-defined-profile-fields.md
tcpipuk: Technically I think conduwuit supports this MSC in its entirety thanks to having full support of matrix-org/matrix-spec-proposals#4133 - what will likely come out of this specific MSC is a need for more moderation tooling and/or potentially endpoints to help moderation tools do their job, so I'm going to resolve this thread for now, but only because we both know we'll both be talking about this over the coming months 🙂

tcpipuk:
My expectations at this point for this MSC:

The core API endpoints are defined in #4133 but it's intentionally tightly focused - it doesn't define individual fields because some of them require a much wider conversation. Sometimes it's easy (like #4175) because the content can be standardised and validated, but we need a dependable way for people to know freetext content can be entered without raising moderation concerns, and this MSC is a thorough roadtest of that.

As part of the discussion on this MSC, I'm expecting to identify dependencies (either existing MSCs or new ones) that will block parts of this MSC until they're in place, but by discussing the end-to-end user/admin/T&S experience of freetext fields in this MSC, we can coordinate these problems and ensure the full journey is covered before the functionality is used by the wider community.

I'd recommend other MSCs for specific freetext fields also reference this one, as we can then identify the multitude of uses for string profile fields, ensure they're accounted for in this proposal, and therefore relieve the same burden from other MSCs,

PR #4211, opened 2024-10-05
MSC4211 - WebXDC on Matrix

seija-amanojaku:
[Rendered](https://github.com/seija-amanojaku/matrix-spec-proposals/blob/main/proposals/4211-webxdc.md)


proposals/4211-webxdc.md
tulir, line 1: Implementation requirements:

* [ ] Client

proposals/4211-webxdc.md
```diff
@@ -0,0 +1,156 @@
...
+
+(Please note that the following simply describes how WebXDC calls transmate to 
+Matrix events. More information is available at WebXDC's own specification)
+
+To start a WebXDC event, a client may start by sending a `m.webxdc.start` event, 
+formatted as such:
+```json
+{
+    "content": {
+        "name": "Storm on Mt. Ooe", // optional, must be a string
```
tulir, line 27: The language tag should be `jsonc` or `json5` instead of `json` to allow comments without them turning red in github previews

proposals/4211-webxdc.md
```diff
@@ -0,0 +1,156 @@
+# MSC4211: WebXDC on Matrix
```
MTRNord, line 1: Isnt WebXDC just a widget with less functionality or am I missing something? Not sure if we need a second widget api :thinking:

seija-amanojaku: I've already mentionned this in the "Alternatives" section:
> Widgets(MSC1236), while powerful, also suffer from a similar agnosticity problem with 
other platforms, and has been noted as not having a "[...] canonical document to describe 
[it]" last year. WebXDC also has stricter security concerns(e.g: clients cannot have 
sensitive information leaked, unlike widgets), thus making it safer.

proposals/4211-webxdc.md
uhoreg: Personally, I'd like to see feedback from a web client regarding how difficult this would be to implement, including the required sandboxing.

proposals/4211-webxdc.md
Johennes: I think the lack of a canonical document is a Matrix spec issue rather than a shortcoming of widgets themselves? IIUC, the main selling point of WebXDC appears to be that it doesn't require another server.

seija-amanojaku: > I think the lack of a canonical document is a Matrix spec issue rather than a shortcoming of widgets themselves?

Seeing it like that, yeah, will correct that back
> IIUC, the main selling point of WebXDC appears to be that it doesn't require another server.

It also is. WebXDC is mostly self-contained (and here, the only real requirement is that you can actually get a MXC URI of the actual WebXDC container)

toger5: It super cool to see someone looking into matrix+WebXDC. I have spend some time looking into it as well and liked the potential.

My appraoch during this investigation was to build a WebXDC container widget. It would load the WebXDC content and forward messages back and forth. It could also take care of putting everything into threads (this is the only way we can have proper back pagination with the widget api as of today) and abstract all this "bridging work" away from the client.

As mentioned before this would be sth each client would need to support.

proposals/4211-webxdc.md
```diff
@@ -0,0 +1,158 @@
...
+## Alternatives
+
+Another way to have interactive content in rooms would be to use an integration manager. 
+However, these would require an additional server to be setup(which are not maintained), 
+and would effectively be limited to the Matrix ecosystem only, instead of being platform-
+agnostic.
+Widgets(MSC1236), while powerful, also suffer from a similar agnosticity problem with 
+other platforms, and has been noted as not having a "[...] canonical document to describe 
+[it]" last year. WebXDC also has stricter security concerns(e.g: clients cannot have 
+sensitive information leaked, unlike widgets), thus making it safer.
```
toger5, line 133: I would love to see a WebXDC widget container alternative section.
As described here: https://github.com/matrix-org/matrix-spec-proposals/pull/4211#discussion_r1790732151

> also suffer from a similar agnosticity problem with 
other platforms

Clients would then support WebXDC through the widget api (dont need to implement both and we have WebXDC for some cases as a valid option to reach larger audiences

> would be to use an integration manager

The integration manager is not strictly required. It is possible to also add widgets as a client feature or manually.

> additional server to be setup

this is still true even if the widget setup itself does not require an integration manager the widget needs to be hosted somewhere. A method to load the widget from a matrix media repository would be very interesting.

farooqkz:
Cheers! Looking forward to see WebXDC in Matrix!

proposals/4211-webxdc.md
```diff
@@ -0,0 +1,162 @@
+# MSC4211: WebXDC on Matrix
```
TheArcaneBrony, line 1: 1) How does this interact with clients that do not, or *can't* include a web runtime?
2) I'd been thinking about a similar MSC to be able to define how a given event type would be rendered. This would allow non-chat clients to interoperate with chat clients more clearly, by being able to specify that eg. a bug tracker can have it's events be rendered as "User opened new issue xyz".

proposals/4211-webxdc.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+In Matrix, users may want to send each other interactive content (for example a game
+in a groupchat, or maybe a collaborative board). Previously, integrations were used 
+to provide such content, but they are not in-band (thus needs another server, which 
+tends to be unmaintained as of now). As such, this proposal aims to resolve this by
+adding [WebXDC](https://webxdc.org/) into the Matrix client ecosystem, with no
```
mokhtarmo6, line 8: Ht

mokhtarmo6: Hhj

proposals/4211-webxdc.md
mokhtarmo6: Tt

farooqkz:
What is status of this PR?

hoh:
There were exciting discussions about this topic during the Matrix sprint before FOSDEM this year, has anything progressed on this side?

PR #4213, opened 2024-10-09
MSC4213: Remove server_name parameter from `join` and `knock` endpoints

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/remove-server-name/proposals/4213-remove-server-name.md)

---

I am a Systems Architect at [gematik](https://www.gematik.de/), Software Engineer at [Unomed](https://www.unomed.ch/), Matrix community member and former Element employee. This proposal was written and published with my community member hat on.

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4213#issuecomment-2484386461)

### DIFF
--- a/proposals/4213-remove-server-name.md
+++ b/proposals/4213-remove-server-name.md
@@ -1,0 +1,54 @@
+# MSC4213: Remove `server_name` parameter
+
+[MSC4156] deprecated the `server_name` parameter on [`/_matrix/client/v3/join/{roomIdOrAlias}`]
+and [`/_matrix/client/v3/knock/{roomIdOrAlias}`] in favor of a new parameter `via`. This change
+shipped in [Matrix v1.12]. In line with the [deprecation policy], the `server_name` parameter
+is now eligible for removal from the spec.
+
+
+## Proposal
+
+The deprecated `server_name` parameter is removed from [`/_matrix/client/v3/join/{roomIdOrAlias}`]
+and [`/_matrix/client/v3/knock/{roomIdOrAlias}`].
+
+
+## Potential issues
+
+None. Servers can continue advertising support for earlier versions of the spec that included
+`server_name` via [`/_matrix/client/versions`].
+
+As of writing, the following stable implementations of [MSC4156] are known to the author:
+
+- synapse: https://github.com/element-hq/synapse/pull/17650
+- dendrite: https://github.com/matrix-org/dendrite/pull/3438
+- matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/pull/4381
+- ruma: https://github.com/ruma/ruma/pull/1891
+- trixnity: https://gitlab.com/trixnity/trixnity/-/merge_requests/478
+
+
+## Alternatives
+
+None.
+
+
+## Security considerations
+
+None.
+
+
+## Unstable prefix
+
+None.
+
+
+## Dependencies
+
+None.
+
+
+[`/_matrix/client/v3/join/{roomIdOrAlias}`]: https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3joinroomidoralias
+[`/_matrix/client/v3/knock/{roomIdOrAlias}`]: https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3knockroomidoralias
+[`/_matrix/client/versions`]: https://spec.matrix.org/v1.10/client-server-api/#get_matrixclientversions
+[Matrix v1.12]: https://spec.matrix.org/v1.12/changelog/v1.12/
+[MSC4156]: https://github.com/matrix-org/matrix-spec-proposals/pull/4156
+[deprecation policy]: https://spec.matrix.org/v1.12/#deprecation-policy


proposals/4213-remove-server-name.md
turt2live, line 1: Implementation requirements:

I'm tempted to mirror https://github.com/matrix-org/matrix-spec-proposals/pull/4127#discussion_r1560097122 and say "none", but am second-guessing that MSC's requirements too. It might be beneficial to have evidence that developers are picking up the newer approaches before suggesting removal from the spec.

Thoughts very welcome.

Johennes: It might not count because they were all created by me but there are five implementations linked in the proposal's body, four of which have already landed.

turt2live:
Using the FCP mechanism to encourage comment/explicit blockers:

@mscbot fcp merge

turt2live:
@mscbot concern implementation-needs-checking label

dbkr (commented):
It's possible I'm confused but https://github.com/matrix-org/matrix-js-sdk/pull/4381 removes the unstable prefix from the 'via' param. This MSC is about removing the old `server_name` param, which is a different thing.

proposals/4213-remove-server-name.md
```diff
@@ -0,0 +1,54 @@
...
+## Potential issues
+
+None. Servers can continue advertising support for earlier versions of the spec that included
+`server_name` via [`/_matrix/client/versions`].
+
+As of writing, the following stable implementations of [MSC4156] are known to the author:
+
+- synapse: https://github.com/element-hq/synapse/pull/17650
+- dendrite: https://github.com/matrix-org/dendrite/pull/3438
+- matrix-js-sdk: https://github.com/matrix-org/matrix-js-sdk/pull/4381
```
dbkr, line 24: It's possible I'm confused but https://github.com/matrix-org/matrix-js-sdk/pull/4381 removes the unstable prefix from the 'via' param. This MSC is about removing the old server_name param, which is a different thing.

tulir: The linked implementations are for MSC4156 which added the `via` param. They're listed to show that the old `server_name` parameter is safe to remove

Johennes: Yes, exactly. I think the `server_name` parameter will probably stay in implementations for quite a while unless they consciously choose to only support Matrix >= 1.12. But we don't need it in the spec anymore.

dbkr: Oh right, I see - I guess I was confused by your [comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4213#discussion_r1794055534). In that case I guess I can vouch for the js-sdk impl, in that it now sends the via param.

proposals/4213-remove-server-name.md
```diff
@@ -0,0 +1,54 @@
...
+[MSC4156] deprecated the `server_name` parameter on [`/_matrix/client/v3/join/{roomIdOrAlias}`]
+and [`/_matrix/client/v3/knock/{roomIdOrAlias}`] in favor of a new parameter `via`. This change
+shipped in [Matrix v1.12]. In line with the [deprecation policy], the `server_name` parameter
+is now eligible for removal from the spec.
+
+
+## Proposal
+
+The deprecated `server_name` parameter is removed from [`/_matrix/client/v3/join/{roomIdOrAlias}`]
+and [`/_matrix/client/v3/knock/{roomIdOrAlias}`].
```
clokep, line 12: I don't want to bikeshed too much, but it seems like clients need to look at `/versions` to decide what sort of parameters to give to a versioned endpoint...

tl;dr Should this rev v3 -> v4 for these endpoints since it is a backwards incompatible change?

richvdh: Can't the client just send both params? That's what EW seems to do, empirically.

Johennes: I suppose cutting a new version would be cleaner as it's a breaking change. But then I think clients would still have to check /versions to figure out which version of the endpoint they can call?

clokep: > Can't the client just send both params? That's what EW seems to do, empirically.

True -- I don't feel strongly, just wanted to bring it up as the endpoints *are* versioned for a reason!

KitsuneRal: The client sending both params for backcomp seems like a reasonable way forward, while servers are encouraged to ignore `via` in favour of `server_name`. We could just keep the endpoint the way it is right now (with the deprecated parameter), and the question would not even stand?

clokep: Seems folks are happy with this. 👍

turt2live:
@mscbot resolve implementation-needs-checking label

proposals/4213-remove-server-name.md
KitsuneRal: Sorry, looks like I was misunderstood. What I meant is that the API shape could remain as it is, with `server_name` deprecated; but homeservers would be expected to ignore it from Matrix 1.13 onwards. But it's not a hill for me to die on.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2059

turt2live:
Merged 🎉

PR #4216, opened 2024-10-17
MSC4216: Set Maximum Allowed Tags

lamoboos223:
Enable matrix servers the choice to set a threshold on tags applied by a user, so the user can apply certain tags only to predefined maximum amount.

### DIFF
--- a/proposals/4216-limit-creating-same-tag.md
+++ b/proposals/4216-limit-creating-same-tag.md
@@ -1,33 +1,3 @@
-# MSC4216: Set Maximum Allowed Tags
+## MSC4216 Proposal
 
-[MSC4216](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4216-limit-creating-same-tag.md)
-
-Enable matrix servers the choice to set a threshold on tags applied by a user, so the user can apply certain tags only to predefined maximum amount.
-
-## Proposal
-
-I'm proposing to add config to matrix servers that specify if the matrix server admin want to limit the tags or not. See the example below on limiting the user to apply the tag `u.pin` to only 3 rooms. This is useful because other chat systems use the same logic on restricting some tag. And with this change we can allow matrix servers admins to dynamically set the max to a tag. Also, if the admin wants to create tag `u.archive` and didn't add it to the config file, that tag will not be restricted to a threshold unlike `u.pin` and `u.favorite`.
-
-```yaml
-tags:
-    - name: u.pin
-      max: 3
-    - name: u.favorite
-      max: 10
-```
-
-## Potential issues
-
-NA
-
-## Alternatives
-
-NA
-
-## Security considerations
-
-NA
-
-## Dependencies
-
-This MSC has no dependencies.
+Let matrix server admins limit user-applied tags, e.g., u.pin to 3 rooms. Useful for consistency with other chat systems.
--- a/proposals/4216-limit-creating-same-tag.md
+++ b/proposals/4216-limit-creating-same-tag.md
@@ -1,3 +1,19 @@
 ## MSC4216 Proposal
 
-Give admins the ability to limit user-applied tags, e.g., u.pin to 3 rooms. Useful for consistency with other chat systems.
+Admins should have the ability to limit tag usage, like restricting `u.pin` to **3** rooms, while tags like `u.archive` will remain unrestricted if **not defined**.
+
+## Potential issues
+
+NA
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


proposals/4216-limit-creating-same-tag.md
```diff
@@ -0,0 +1,33 @@
+# MSC4216: Set Maximum Allowed Tags
+
+[MSC4216](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4216-limit-creating-same-tag.md)
+
+Enable matrix servers the choice to set a threshold on tags applied by a user, so the user can apply certain tags only to predefined maximum amount.
+
+## Proposal
+
+I'm proposing to add config to matrix servers that specify if the matrix server admin want to limit the tags or not. See the example below on limiting the user to apply the tag `u.pin` to only 3 rooms. This is useful because other chat systems use the same logic on restricting some tag. And with this change we can allow matrix servers admins to dynamically set the max to a tag. Also, if the admin wants to create tag `u.archive` and didn't add it to the config file, that tag will not be restricted to a threshold unlike `u.pin` and `u.favorite`.
```
clokep, line 9: Please wrap your lines to roughly ~150 characters.

It is best to to write the proposal section as if it has been accepted, e.g.

> Matrix servers can return a new error when a user attempts to tag a room.

lamoboos223: @clokep summarized to 140 characters :)

turt2live (commented):
Please [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) on your changes for eventual FCP consideration.

It also appears as though your proposal has gone missing? I've put it back to draft while it's a work in progress.

proposals/4216-limit-creating-same-tag.md
clokep: Note that by this I meant make sure that each line is at most 150 characters. You can have as many lines as you'd like!

turt2live (commented):
Thanks for submitting this MSC! For the proposal to continue towards acceptance, it will need a more formal description of the error codes and such a client can expect, and an implementation (likely both server and client) to demonstrate the solution works in practice. 

Error code MSCs are typically fairly easy to write, with implementation on the client side often just being evidence that clients don't explode when they hit the error. If they do explode, it's more a sign that the client needs improved error handling generally, rather than a blocker for the proposal's acceptance - we just require evidence that shipping the feature in the spec isn't likely to cause widespread complaint. 

If you need some help writing the MSC language, feel free to visit [#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) on Matrix for advice from the community.

proposals/4216-limit-creating-same-tag.md
```diff
@@ -0,0 +1,19 @@
+## MSC4216 Proposal
+
+Admins should have the ability to limit tag usage, like restricting `u.pin` to **3** rooms, while tags like `u.archive` will remain unrestricted if **not defined**.
```
TheArcaneBrony, line 3: Personally, I see this purely as an implementation detail, not something that should be specified, but there might be value in clarifying semantics for things like M_LIMIT_EXCEEDED in certain contexts (eg. tags in this context).

Additionally: How would migrations canonically work? What happens if the admin lowers the limit to below the count users already have? Are tags even relevant in 2024, with most clients not even supporting custom tags to begin with? Is there a good benefit to limiting usage count of tags across clients?

lamoboos223: Hi
So the reason i was proposing this feature is because to priority tags and lazy loading their content/messages instead of loading the whole rooms with no tags defined. So for example if a user pinned 3 rooms, these 3 rooms will be asked for their content before the rest of the rooms and this will happen on the client side.

  You make a good point about what would happen if the admin lower the amount of a tag, I actually didn’t think it through (I apologies).

@TheArcaneBrony

TheArcaneBrony: I'm not quite sure I see how this would be beneficial? You could just ask the server for all rooms, or add a new joined room on every subsequent request on sync.
Alternatively, you could do the same but request all DMs first? I'm not sure I see a tangible benefit here besides attempting to make initial sync faster, in which case... You could store a sorted list of rooms by member count in account data and handle future logins this way by referencing account data.

I'm not discrediting the idea here, just wondering what the train of thought is versus using heuristics transparently without requiring user involvement.

PR #3765, opened 2022-04-03
MSC3765: Rich text in room topics

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/rich-room-topics/proposals/3765-rich-room-topics.md)

Implementations:

- https://github.com/matrix-org/matrix-events-sdk/pull/5
- https://github.com/matrix-org/matrix-js-sdk/pull/2272
- https://github.com/matrix-org/matrix-react-sdk/pull/8215

---

In line with https://github.com/matrix-org/matrix-spec/issues/1700, the following disclosure applies:

I am a Systems Architect at [gematik](https://www.gematik.de/), Software Engineer at [Unomed](https://www.unomed.ch/), Matrix community member and former Element employee. This proposal was written and published with my community member hat on.

---

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3765#issuecomment-2430107320)

PR #3552, opened 2021-12-07
MSC3552: Extensible Events - Images and Stickers

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/extev/images/proposals/3552-extensible-events-images.md)

Blocked by https://github.com/matrix-org/matrix-doc/pull/1767
Blocked by https://github.com/matrix-org/matrix-doc/pull/3551

Preview: https://pr3552--matrix-org-previews.netlify.app

PR #4221, opened 2024-10-26
MSC4221: Room Banners

88572:
[Rendered](https://github.com/everypizza1/matrix-spec-proposals/blob/room-banners/proposals/4221-room-banners.md)

### Implementations:

- Extera: https://github.com/officialdakari/Extera/commit/3505c06e78fad1d9e2e67326c696c409da62a8a3

### DIFF
--- a/proposals/4221-room-banners.md
+++ b/proposals/4221-room-banners.md
@@ -1,0 +1,41 @@
+# MSC4221: Room Banners
+
+This proposal adds room banners to Matrix, similar to the way Discord has guild banners.
+
+## Proposal
+
+Matrix has ways to add room avatars, but those are limited to a 1:1 aspect ratio. This proposal adds banners to rooms, which are a 16:9 aspect ratio.
+A banner state event should have the type `m.room.banner` and the content should contain the `url` in `mxc://` format.
+Example event:
+```json
+{
+  "type": "m.room.banner",
+  "sender": "@everypizza:chat.blahaj.zone",
+  "content": {
+    "info": {
+      "mimetype": "image/jpeg"
+    },
+    "url": "mxc://…/…"
+  },
+…
+}
+```
+## Potential issues
+
+A malicious user could upload a banner with explicit or illegal content. A future MSC may provide a way to mark room pictures or banners as sensitive using [MSC4193](http://github.com/matrix-org/matrix-spec-proposals/pull/4193).
+
+## Alternatives
+
+None I can think of.
+
+## Security considerations
+
+None I can think of.
+
+## Unstable prefix
+
+While this MSC is unstable, clients wanting to implement this should use `page.codeberg.everypizza.room.banner` instead of `m.room.banner`.
+
+## Dependencies
+
+N/A
--- a/proposals/4221-room-banners.md
+++ b/proposals/4221-room-banners.md
@@ -26,7 +26,7 @@ A malicious user could upload a banner with explicit or illegal content. A futur
 
 ## Alternatives
 
-None I can think of.
+Pinned messages and room topics are in similar areas of the UI in some clients.
 
 ## Security considerations
 


proposals/4221-room-banners.md
tulir, line 1: Implementation requirements:

* [ ] Client

proposals/4221-room-banners.md
```diff
@@ -0,0 +1,41 @@
...
+  "type": "m.room.banner",
+  "sender": "@everypizza:chat.blahaj.zone",
+  "content": {
+    "info": {
+      "mimetype": "image/jpeg"
+    },
+    "url": "mxc://…/…"
+  },
+…
+}
```
tulir, line 21: New events should consider how they fit in with extensible events. For example, https://github.com/matrix-org/matrix-spec-proposals/pull/3765 turns room topics into extensible events

proposals/4221-room-banners.md
```diff
@@ -0,0 +1,41 @@
...
+```
+## Potential issues
+
+A malicious user could upload a banner with explicit or illegal content. A future MSC may provide a way to mark room pictures or banners as sensitive using [MSC4193](http://github.com/matrix-org/matrix-spec-proposals/pull/4193).
+
+## Alternatives
+
+None I can think of.
```
tulir, line 29: Pinned messages and rich text topics are both in the same area of client UI, so they should probably be mentioned even if they aren't the exact same thing

proposals/4221-room-banners.md
everypizza1: Should it become `m.banner` instead of `m.room.banner`?

proposals/4221-room-banners.md
everypizza1: Alright, I added a note about that.

proposals/4221-room-banners.md
officialdakari: :)

officialdakari:
initial support now implemented in Extera

proposals/4221-room-banners.md
```diff
@@ -0,0 +1,41 @@
+# MSC4221: Room Banners
+
+This proposal adds room banners to Matrix, similar to the way Discord has guild banners.
```
Half-Shot, line 3: I wonder how this would render in a Matrix client today. Discord has banners for it's guilds but they only really show up on invite to the guild and in the top right of the client when viewing the guild. They're not *really* tied to individual channels.

For Matrix rooms, where would this go? Obviously I don't expect the spec to start designing the client but I fear there isn't enough information about how this should be presented to the user.

Would this be displayed on invite? (in which case there needs to be recommendations for `m.room.banner` to appear in stripped state (https://spec.matrix.org/latest/client-server-api/#stripped-state).

proposals/4221-room-banners.md
```diff
@@ -0,0 +1,41 @@
...
+      "mimetype": "image/jpeg"
+    },
+    "url": "mxc://…/…"
+  },
+…
+}
+```
+## Potential issues
+
+A malicious user could upload a banner with explicit or illegal content. A future MSC may provide a way to mark room pictures or banners as sensitive using [MSC4193](http://github.com/matrix-org/matrix-spec-proposals/pull/4193).
```
Half-Shot, line 25: Some concerns about having reasonable fallbacks here. Not every client would be able to display a banner (and I'm not sure every user is going to want to be confronted with a banner in their client), so I think there should be some way to fallback to textual representation of the banner.

I don't know how common it is for banners to contain actual information, usually they seem to be more decorative?

proposals/4221-room-banners.md
everypizza1: I would think that banners should be above room info.

proposals/4221-room-banners.md
everypizza1: By "sensitive content", I mean things that would commonly have a content warning on platforms like the Fediverse such as food, lewd stuff, politics, etc.

Half-Shot: Sorry, I commented a line below where I meant to :). This was a separate concern.

officialdakari: > By "sensitive content", I mean things that would commonly have a content warning on platforms like the Fediverse such as food, lewd stuff, politics, etc.

sorry for offtopic, but why food is sensitive content?

Saiv46: > sorry for offtopic, but why food is sensitive content?

@officialdakari Propably because Eating Disorder. It shouldn't be required, but some people on Fediverse go overboard with it. /offtopic

proposals/4221-room-banners.md
HarHarLinks: yes, the type would probably be just `m.banner`. but also [MSC3552](https://github.com/matrix-org/matrix-spec-proposals/pull/3552) content blocks of `m.file` and `m.image_details` (or whatever they end up being defined as) would go into a custom content block like 3765 did:
```
"content": {
  "m.banner": {
    "m.file": {...},
    "m.image_details": {...}
  }
}
```

catfromplan9:
Just checking in, has there been any movement on this?

q-TheDog:
Client implementation: https://github.com/Haven-Organization/haven-desktop
image

proposals/4221-room-banners.md
q-TheDog: Client (Haven Desktop):
- https://github.com/Haven-Organization/haven-desktop/blob/develop/src/apps/social/utils/useUserBanner.ts
- https://github.com/Haven-Organization/haven-desktop/blob/develop/src/apps/social/components/BannerSetting.tsx

nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/15b47291c204c1ac21012429aa207cc0ebd60336

PR #4186, opened 2024-08-30
MSC4186: Simplified Sliding Sync

erikjohnston:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md)

--- 

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4186#issuecomment-3345882856)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4186#issuecomment-3346108416)

PR #4222, opened 2024-10-29
MSC4222: Adding `state_after` to sync v2

erikjohnston:
This is to allow clients to opt-in to a change of the sync v2 API that allows them to correctly track the state of the room.


proposals/4222-sync-v2-state-after.md
turt2live, line 1: Implementation requirements:
* [ ] Server
* [ ] Client

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,192 @@
...
+
+With the proposed API the common case for receiving a state update will cause the event to come down in both the
+`timeline` and `state` sections, potentially increasing bandwidth usage. However, it is common for the HTTP responses to
+be compressed, heavily reducing the impact of having duplicated data.
+
+Clients will not be able to tell when a state change happened within the timeline. This was used by some clients to
+render e.g. display names of users at the time they sent the message (rather than their current display name), though
+e.g. Element clients have moved away from this UX. This behavior can be replicated in the same way that clients dealt
+with messages received via pagination (i.e. calling `/messages`), by walking the timeline backwards and inspecting the
+`unsigned.prev_state` field. While this can lead to incorrect results, this is no worse than the previous situation.
```
toger5, line 155: This kind of usage of the timeline section should be not impacted at all.
We still send all state events (even if they might be wrong) in the timeline section so clients using those state event to render things in the timeline should be not impacted at all.

```suggestion
As before clients will not be able to tell when a state change happened within the timeline. This was used by some clients to
render e.g. display names of users at the time they sent the message (rather than their current display name), though
e.g. Element clients have moved away from this UX.
This MSC allows to compute the current state correctly. It does not fix the behavior for the state history in the timeline. So clients using such an approach still will face the same issues they had before this MSC.

A proper solution for this usecase would be to paginate back state using the `replaces_state` id. With this one can find all actual state event ids. This list of Ids can be used in the timeline history to filter and invalid state changes.
```

erikjohnston: I guess what I wanted to make clear was that clients will need to change how they calculate the "state at each event", and that that behaviour is really considered best-effort at best.

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,192 @@
...
+</tr>
+</table>
+
+Since the current state of the room does not include the new state event, it's excluded from the `state_after` section.
+
+> [!IMPORTANT]
+> Both responses are the same, but the client **MUST NOT** update its state with the event.
+
+
+## Potential issues
```
Gnuxie, line 145: > The format of returned state in `required_state` is a list of events. This does now allow the server to indicate if a "state reset" has happened which removed an entry from the state entirely (rather than it being replaced with another event).

Does this issue from [MSC4186: Simplified Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)  apply to this MSC also

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,192 @@
...
+    } ],
+    "limited": false,
+  },
+  "state_after": {
+    "events": [ {
+      "type": "org.matrix.example",
+      "state_key": ""
+    } ]
+  }
+```
```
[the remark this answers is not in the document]
erikjohnston: well that didn't help :angry:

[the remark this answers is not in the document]
erikjohnston: oh it was missing a bracket!

proposals/4222-sync-v2-state-after.md
erikjohnston: Mmm, that is true. I guess there is an open question as to whether we want to try and fix that here or not :thinking:

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+
+There are no security concerns with this proposal, as it simply encodes the same information sent do clients in a
+different way
+
+## Unstable prefix
+
+| Name | Stable prefix | Unstable prefix |
+| - | - | - |
+| Query param | `use_state_after` | `org.matrix.use_state_after` |
+| Room response field | `state_after` | `org.matrix.state_after` |
```
richvdh, line 189: unstable prefixes normally have the msc number in?

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+the current sync, and clients do not update their view of state based on entries in the timeline.
+
+
+## Proposal
+
+This change is gated behind the client adding a `?use_state_after=true` (the unstable name is
+`org.matrix.use_state_after`) query param.
+
+When enabled, the Homeserver will **omit** the `state` section in the room response sections. This is replaced by
+`state_after` (the unstable field name is `org.matrix.state_after`), which will include all state changes between the
```
anoadragon453, line 23: According to the "Unstable prefix" section below.

```suggestion
This change is gated behind the client adding a `?use_state_after=true` (the unstable name is
`org.matrix.msc4222.use_state_after`) query param.

When enabled, the Homeserver will **omit** the `state` section in the room response sections. This is replaced by
`state_after` (the unstable field name is `org.matrix.msc4222.use_state_after`), which will include all state changes between the
```

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+both suboptimal to just telling the client the correct information in the first place. Since clients will need to be
+updated to handle the new behavior for future sync APIs anyway, there is little benefit from not updating clients now.
+
+We could also do nothing, and instead wait for [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
+(or equivalent) to land and for clients to update to it.
+
+
+## Security considerations
+
+There are no security concerns with this proposal, as it simply encodes the same information sent do clients in a
```
anoadragon453, line 181:
```suggestion
There are no security concerns with this proposal, as it simply encodes the same information sent to clients in a
```

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+mean that a new state event will (likely) appear in both the timeline and state sections of the response.
+
+This is basically the same as how state is returned in [MSC4186 - Simplified Sliding
+Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186).
+
+State events that appear in the timeline section **MUST NOT** update the current state. The current state **MUST** only be
+updated with the contents of `state_after`.
+
+Clients can tell if the server supports this change by whether it returns a `state` or `state_after` section in the
+response.
```
tulir, line 35: `state` isn't currently required in sync responses, but I guess `state_after` will be required in all joined rooms after this change? Should probably make it more explicit that servers MUST return `state_after` with an empty `events` array inside

turt2live:
A team member has request SCT review on this MSC for the following points:

* To understand it doesn't handle where state gets deleted, and if so how to fix, if we indeed want to.
* And depending on that, finalise how it should look spec-wise.

proposals/4222-sync-v2-state-after.md
Gnuxie: Yeah ok but is this easy or hard to fix? If it isn't easy, then will it be harder to fix when the proposal is spec? I don't mind a compromise as long as we're not hurting ourselves later

kegsay: It depends how you classify easy or hard :D

It's technically very easy, as you just need to have some representation of "deleted" events: we've tossed around `content: {}` and `content: null` and a few others like having a new key in the response JSON. It's socially difficult however as it's prone to bikeshedding.

Gnuxie: Ah, i see. `content: {}` is a bit more palatable for me since that's already how a lot of tools send state events to "remove" them from current state, and it doesn't break anyone's event schema. I'd probably pair it with a flag in unsigned or something though.

Gnuxie: Do you intend to show if a state event in current state has been redacted? How would that work?

What about killing two birds in one stone and repeat the event with empty content. Since then it doesn't matter if a client is naive and unaware of the behavior, if they naively update their model then they'll end up with the right state.

Gnuxie: > 1. Mixing deleted state in with the other state means that clients essentially need to handle multiple datatypes in the same list.

> 2. Having a separate field will make it easy for client implementations to miss/forget implementing the extra field during implementation. From my experience, its fairly common for people not to closely read the docs when initially writing clients, and so I think this will happen more often than you might think.

I mean if you really want to avoid both of these ideally you could use a new type for state deltas with a property for the event and a property for the change type. In Draupnir land we call them: Introduced (for unseen state key type pair), Superseded, Redacted, and I guess you'd also need Removed[^1].

Not sure if it's practical or possible for you to change that now though.

[^1]: It's actually a bit more complicated because we try to not keep completely redacted state events around as an optimization and try to tell people exactly how they have changed but eh it's not necessary to copy that https://github.com/Gnuxie/matrix-protection-suite/blob/main/src/StateTracking/StateChangeType.ts#L13-L64

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+When enabled, the Homeserver will **omit** the `state` section in the room response sections. This is replaced by
+`state_after` (the unstable field name is `org.matrix.msc4222.use_state_after`), which will include all state changes between the
+previous sync and the *end* of the timeline section of the current sync. This is in contrast to the old `state` section
+that only included state changes between the previous sync and the *start* of the timeline section. Note that this does
+mean that a new state event will (likely) appear in both the timeline and state sections of the response.
+
+This is basically the same as how state is returned in [MSC4186 - Simplified Sliding
+Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186).
+
+State events that appear in the timeline section **MUST NOT** update the current state. The current state **MUST** only be
```
richvdh, line 31:
```suggestion
Clients **MUST NOT** update their view of the current state based on events that appear in the timeline section.
The current state view **MUST** only be
```

proposals/4222-sync-v2-state-after.md
richvdh:
```suggestion
response. The `state_after` property is **required** by servers that support this change.
```

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+</td>
+</tr>
+</table>
+
+Since the current state of the room does not include the new state event, it's excluded from the `state_after` section.
+
+> [!IMPORTANT]
+> Both responses are the same, but the client **MUST NOT** update its state with the event.
```
richvdh, line 143:
```suggestion
> Both responses are the same, other than the change of `state` to `state_after`, but the client **MUST NOT** update its state with the event.
```

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+with messages received via pagination (i.e. calling `/messages`), by walking the timeline backwards and inspecting the
+`unsigned.prev_state` field. While this can lead to incorrect results, this is no worse than the previous situation.
+
+
+## Alternatives
+
+There are a number of options for encoding the same information in different ways, for example the response could
+include both the `state` and a `state_delta` section, where `state_delta` would be any changes that needed to be applied
+to the client calculated state to correct it. However, since
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) is likely to replace the sync v2 API, we may as
```
richvdh, line 164:
```suggestion
[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) is likely to replace the current `/sync` API, we may as
```

proposals/4222-sync-v2-state-after.md
richvdh: > Any fix we do to communicate deleted state in SSS can easily enough be backported to v2 should it be needed. I see no reason to block the MSC on this point. On the contrary, this MSC _massively_ reduces desync issues compared to v2 without `state_after` because we are now tracking diffs against the current state and not diffs against the last timeline event. For such a small change, this MSC provides immense value.

+1; I'm strongly in favour of *not* blocking this MSC for years until someone has time to wrangle this last nit. I don't think that landing it as-is will make it materially more difficult to fix later.

richvdh:
MSCs proposed for Final Comment Period (FCP) should meet the requirements outlined in the [checklist](https://github.com/matrix-org/matrix-spec-proposals/blob/main/MSC_CHECKLIST.md?rgh-link-date=2025-03-25T05%3A43%3A02Z) prior to being accepted into the spec. This checklist is a bit long, but aims to reduce the number of follow-on MSCs after a feature lands.

- [n/a] Are all MSCs that this MSC depends on already accepted?
- [n/a] For each new endpoint that is introduced:
- [n/a] Will the MSC require a new room version, and if so, has that been made clear?
  - [x] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/v1.8/appendices/#pagination)?
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

richvdh:
There are a couple of open threads here, but I don't think they need to block us from starting to gather FCP ticks.

@mscbot fcp merge

proposals/4222-sync-v2-state-after.md
Gnuxie: Might be way down on the lower end of priorities but this isn't a nit. And it shouldn't be taking years to sort issues out like this. If it is, then that's a more urgent issue to solve because then things are entirely dysfunctional.

richvdh: @Gnuxie you're right, it's not a nit and it is important; I'm sorry I spoke in a dismissive tone about it. You're also right that finding a solution shouldn't take years; I very much hope a resolution can be reached in a much shorter timeframe. 

Nevertheless, given the practicalities of where we are right now in terms of time and money, I think it would be a grave mistake to let perfect be the enemy of good by holding this MSC up while we wait for a solution to come along.

TheArcaneBrony:
LibMatrix now supports MSC4222, including clientside emulation (via /state or /state/x/y?format=event) for servers that don't support it as of [824171](https://cgit.rory.gay/matrix/LibMatrix.git/commit/?id=82417145ff67fae9eac4dbe8137f5b9ef0d222f9)! :tada:

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+Since the current state of the room does not include the new state event, it's excluded from the `state_after` section.
+
+> [!IMPORTANT]
+> Both responses are the same, but the client **MUST NOT** update its state with the event.
+
+
+## Potential issues
+
+With the proposed API the common case for receiving a state update will cause the event to come down in both the
+`timeline` and `state` sections, potentially increasing bandwidth usage. However, it is common for the HTTP responses to
```
KitsuneRal:
```suggestion
`timeline` and `state_after` sections, potentially increasing bandwidth usage. However, it is common for the HTTP responses to
```

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+particularly problematic for use-cases that rely on state being consistent between different clients.
+
+This behavior stems from the fact that the clients update their view of the current state with state events that appear
+in the timeline. To handle gappy syncs, the `state` section includes state events that are from *before* the start of
+the timeline, and so are replaced by any matching state events in the timeline. This provides little opportunity for the
+server to ensure that the clients come to the correct conclusion about the current state of the room.
+
+In [MSC4186 - Simplified Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) this problem is
+solved by the equivalent `required_state` section including all state changes between the previous sync and the end of
+the current sync, and clients do not update their view of state based on entries in the timeline.
```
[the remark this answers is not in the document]
kegsay: The problem is that clients need to know the current state and the current model cannot express the current state. Moving to "state_after" is perhaps a bit of a misnomer as it's actually moving to the _current state_ instead. A trivial example would be this graph (all events are non-conflicted state events):
```
   A
   |
   B
  / \
 C   D

State after C = {A,B,C}
State after D = {A,B,D}
Current State = {A,B,C,D}
```

Implementations which try to think of this as "some start state" + "some linear sequence of events"  = "end state" are fundamentally calculating the wrong thing because forks in the graph (concurrent events) can occur at any point on the linear sequence of events. You might argue that this is commutative so just union the lot (i.e a timeline of [A,B,D,C] or [A,B,C,D] are valid arrival orders and hey the union of either is {A,B,C,D} but the problem is when you then introduce state resolution which is effectively a sorting and filtering algorithm:

```
   A
   |
   B
  / \
 C   C'   <-- C' will win this
  \ / \
   D   E


State after C = {A,B,C}
State after D = {A,B,C'}
State after E = {A,B,C',E}
Current State = {A,B,C',D,E}
```

Now assume you see [A,B,C',E]. The current "state before" does the right thing as does `state_after` E and the "current state" are in agreement as we haven't seen any concurrent events yet: all 3 calculations agree the end state is {A,B,C',E}. Now you get [C,D], and things get _really_ weird:
 - The client thinks the room state is `{A,B,C',E}`.
 - The current /sync impl will have a `state` block of {B} and then have `[C,D]`, meaning we incorrectly replace C' with C.
 - The true state after the last event in the timeline (D) is {A,B,C'} which is wrong as we're missing E. Thankfully, this [isn't what the implementation is doing](https://github.com/element-hq/synapse/pull/17888/files#diff-70c84622edff898486bdc02d7a579db4e52a3e0b26c14020bed0ca1fb8ea2018R1387).
 - The server impl for this MSC is calculating the current state delta, so the delta to {A,B,C',D,E} which is {D} which is correct.

When writing this up, some of this was surprising even to me, so I've backed this up with a complement test you can run which demonstrates the above fork problem in the current sync impl:

<details>
  <summary>Complement test</summary>

```go

func TestSyncDoesRightDelta(t *testing.T) {
	deployment := complement.Deploy(t, 1)
	defer deployment.Destroy(t)
	srv := federation.NewServer(t, deployment,
		federation.HandleKeyRequests(),
		federation.HandleMakeSendJoinRequests(),
		federation.HandleTransactionRequests(nil, nil),
	)
	srv.UnexpectedRequestsAreErrors = false
	cancel := srv.Listen()
	defer cancel()

	alice := deployment.Register(t, "hs1", helpers.RegistrationOpts{})
	bob := srv.UserID("bob")
	ver := alice.GetDefaultRoomVersion(t)
	serverRoom := srv.MustMakeRoom(t, ver, federation.InitialRoomEvents(ver, bob))

	// Join Alice to the new room on the federation server and make Alice a moderator
	alice.MustJoinRoom(t, serverRoom.RoomID, []string{srv.ServerName()})
	since := alice.MustSyncUntil(t, client.SyncReq{}, client.SyncJoinedTo(alice.UserID, serverRoom.RoomID))

	// send the linear sequence B, C', E
	eventB := srv.MustCreateEvent(t, serverRoom, federation.Event{
		Type:     "m.room.B",
		StateKey: b.Ptr(""),
		Sender:   bob,
		Content: map[string]interface{}{
			"event": "B",
		},
	})
	// we need to fudge the origin_server_ts so we win state res, so we need a custom function
	createEventWithTimestamp := func(t *testing.T, s *federation.Server, r *federation.ServerRoom, ev federation.Event, timestamp time.Time) gomatrixserverlib.PDU {
		proto, err := serverRoom.ProtoEventCreator(serverRoom, ev)
		must.NotError(t, "ProtoEventCreator: failed to create proto event", err)
		verImpl, err := gomatrixserverlib.GetRoomVersion(r.Version)
		must.NotError(t, "GetRoomVersion: unknown version", err)
		eb := verImpl.NewEventBuilderFromProtoEvent(proto)
		signedEvent, err := eb.Build(timestamp, spec.ServerName(s.ServerName()), s.KeyID, s.Priv)
		must.NotError(t, "Build", err)
		return signedEvent
	}

	eventCPrime := createEventWithTimestamp(t, srv, serverRoom, federation.Event{
		Type:     "m.room.C",
		StateKey: b.Ptr(""),
		Sender:   bob,
		Content: map[string]interface{}{
			"event": "C'",
		},
		PrevEvents: []string{eventB.EventID()},
	}, time.Now().Add(time.Hour))

	eventE := srv.MustCreateEvent(t, serverRoom, federation.Event{
		Type:     "m.room.E",
		StateKey: b.Ptr(""),
		Sender:   bob,
		Content: map[string]interface{}{
			"event": "E",
		},
		PrevEvents: []string{eventCPrime.EventID()},
	})
	srv.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{eventB.JSON(), eventCPrime.JSON(), eventE.JSON()}, nil)
	time.Sleep(time.Second) // let the events be processed, we could use a sentinel but it's more code and this is just an example

	currentState := make(map[string]string)
	roomID := serverRoom.RoomID

	getCurrentStateForC := func() {
		stateResp := alice.Do(t, "GET", []string{"_matrix", "client", "v3", "rooms", roomID, "state"})
		body := must.ParseJSON(t, stateResp.Body)
		for _, ev := range body.Array() {
			if ev.Get("type").Str == "m.room.C" {
				t.Logf("/state says the current state of C is %v", ev.Get("content").Raw)
			}
		}
	}

	syncUpdatingStateUntilEvent := func(finalEventID string) client.SyncCheckOpt {
		return func(clientUserID string, topLevelSyncJSON gjson.Result) error {
			t.Helper()
			stateField := "rooms.join." + client.GjsonEscape(roomID) + ".state.events"
			timelineField := "rooms.join." + client.GjsonEscape(roomID) + ".timeline.events"
			if len(topLevelSyncJSON.Get(stateField).Array()) > 0 {
				t.Logf("received state field: %v", topLevelSyncJSON.Get(stateField).Raw)
			}
			timeline := topLevelSyncJSON.Get(timelineField).Array()
			for _, ev := range timeline {
				if !ev.Get("state_key").Exists() {
					continue
				}
				currentState[ev.Get("type").Str+"/"+ev.Get("state_key").Str] = ev.Get("content").Raw
				if ev.Get("event_id").Str == finalEventID {
					return nil
				}
			}
			return fmt.Errorf("not seen final event in timeline yet")
		}
	}
	since = alice.MustSyncUntil(t, client.SyncReq{Since: since}, syncUpdatingStateUntilEvent(eventE.EventID()))
	t.Logf("client calculated current state before sending C is: %+v", currentState)
	getCurrentStateForC()

	// now send C,D
	eventC := srv.MustCreateEvent(t, serverRoom, federation.Event{
		Type:     "m.room.C",
		StateKey: b.Ptr(""),
		Sender:   bob,
		Content: map[string]interface{}{
			"event": "C",
		},
		PrevEvents: []string{eventB.EventID()},
	})

	eventD := srv.MustCreateEvent(t, serverRoom, federation.Event{
		Type:     "m.room.D",
		StateKey: b.Ptr(""),
		Sender:   bob,
		Content: map[string]interface{}{
			"event": "D",
		},
		PrevEvents: []string{eventC.EventID(), eventCPrime.EventID()},
	})

	srv.MustSendTransaction(t, deployment, "hs1", []json.RawMessage{eventC.JSON(), eventD.JSON()}, nil)
	time.Sleep(time.Second)

	// now sync until we see event D
	since = alice.MustSyncUntil(t, client.SyncReq{Since: since}, syncUpdatingStateUntilEvent(eventD.EventID()))
	t.Logf("client calculated current state after sending C is: %+v", currentState)
	getCurrentStateForC()

	t.Logf("event B: %v prevs=%v", eventB.EventID(), eventB.PrevEventIDs())
	t.Logf("event C: %v prevs=%v", eventC.EventID(), eventC.PrevEventIDs())
	t.Logf("event C': %v prevs=%v", eventCPrime.EventID(), eventCPrime.PrevEventIDs())
	t.Logf("event D: %v prevs=%v", eventD.EventID(), eventD.PrevEventIDs())
	t.Logf("event E: %v prevs=%v", eventE.EventID(), eventE.PrevEventIDs())
	must.Equal(t, currentState["m.room.C/"], `{"event":"C'"}`, "incorrect value for C")
}
```
</details>

This will produce:
```
    federation_sync_test.go:227: client calculated current state before sending C is: map[m.room.B/:{"event":"B"} m.room.C/:{"event":"C'"} m.room.E/:{"event":"E"}]
    client.go:706: [CSAPI] GET hs1/_matrix/client/v3/rooms/!0-f0MIgRBjwmRJcYz9xV:host.docker.internal:61064/state => 200 OK (3.088667ms)
    federation_sync_test.go:200: /state says the current state of C is {"event":"C'"}
    client.go:706: [CSAPI] GET hs1/_matrix/client/v3/sync => 200 OK (14.688584ms)
    federation_sync_test.go:256: client calculated current state after sending C is: map[m.room.B/:{"event":"B"} m.room.C/:{"event":"C"} m.room.D/:{"event":"D"} m.room.E/:{"event":"E"}]
    client.go:706: [CSAPI] GET hs1/_matrix/client/v3/rooms/!0-f0MIgRBjwmRJcYz9xV:host.docker.internal:61064/state => 200 OK (2.970042ms)
    federation_sync_test.go:200: /state says the current state of C is {"event":"C'"}
    federation_sync_test.go:259: event B: $TZcrwJhi0tgbUq5xtiwu9blCvhWWzetPu7h-azGdNmo prevs=[$oTThzswBLFz3b6mE8e-uDqVNgVYto6nYZsUgNOcNT88]
    federation_sync_test.go:260: event C: $KXHXQy8AgEWBTfy-7e5g5nuytM6PyY3QmtON3gGxEYo prevs=[$TZcrwJhi0tgbUq5xtiwu9blCvhWWzetPu7h-azGdNmo]
    federation_sync_test.go:261: event C': $huJSetAOxfYS3pCh28GaZc5QEgJhkmcEB2eijzSyvNw prevs=[$TZcrwJhi0tgbUq5xtiwu9blCvhWWzetPu7h-azGdNmo]
    federation_sync_test.go:262: event D: $EEzR4Ky58G4Xs4k7BvnM4ZfxWGeVOUOvTuS9vf9T9VI prevs=[$KXHXQy8AgEWBTfy-7e5g5nuytM6PyY3QmtON3gGxEYo $huJSetAOxfYS3pCh28GaZc5QEgJhkmcEB2eijzSyvNw]
    federation_sync_test.go:263: event E: $r5fBu3cXBYuJKcodxMz3QkxzOPKw3XIval1hONFOgno prevs=[$huJSetAOxfYS3pCh28GaZc5QEgJhkmcEB2eijzSyvNw]
    federation_sync_test.go:264: Equal incorrect value for C: got '{"event":"C"}' want '{"event":"C'"}'
```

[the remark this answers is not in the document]
richvdh: That's a great write-up, kegan, thanks! It would be great if it could go somewhere it can be linked from the body of the MSC (or even be added as an appendix to the MSC body?)

clokep:
@mscbot concern Unresolved threads

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,193 @@
...
+`state_after` (the unstable field name is `org.matrix.msc4222.use_state_after`), which will include all state changes between the
+previous sync and the *end* of the timeline section of the current sync. This is in contrast to the old `state` section
+that only included state changes between the previous sync and the *start* of the timeline section. Note that this does
+mean that a new state event will (likely) appear in both the timeline and state sections of the response.
+
+This is basically the same as how state is returned in [MSC4186 - Simplified Sliding
+Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186).
+
+State events that appear in the timeline section **MUST NOT** update the current state. The current state **MUST** only be
+updated with the contents of `state_after`.
```
fkwp:
```suggestion
Clients **MUST** only update their local state using `state_after` and **NOT** consider the events that appear in the timeline section of `/sync`.
```

proposals/4222-sync-v2-state-after.md
fkwp:
```suggestion
response. Servers that support this change **MUST** return the `state_after` property, even if empty.
```

proposals/4222-sync-v2-state-after.md
fkwp:
```suggestion
> Even though both responses look very similar, the client **MUST NOT** update its state with the event from the timeline section when using `state_after`.
```

proposals/4222-sync-v2-state-after.md
fkwp:
```suggestion
As before, clients will not be able to reliably tell when a state change happened within the timeline. Currently, some clients walk through the timeline to
render e.g. display names of users at the time they sent the message (rather than their current display name), though
e.g. Element clients have moved away from this UX.
This MSC allows to compute the current state correctly. It does not fix the behavior for the state history in the timeline. So clients using such an approach still will face the same issues they had before this MSC.
```

turt2live (approved):
lgtm provided these threads are incorporated into the MSC:

* https://github.com/matrix-org/matrix-spec-proposals/pull/4222/files/fbcf1fc8a3f47cc2dc545e09059a9a2af1ee6e62#r2054361343
* https://github.com/matrix-org/matrix-spec-proposals/pull/4222/files#r1821125000

proposals/4222-sync-v2-state-after.md
```diff
@@ -0,0 +1,191 @@
...
+## Potential issues
+
+With the proposed API the common case for receiving a state update will cause the event to come down in both the
+`timeline` and `state` sections, potentially increasing bandwidth usage. However, it is common for the HTTP responses to
+be compressed, heavily reducing the impact of having duplicated data.
+
+As before, clients will not be able to reliably tell when a state change happened within the timeline. Currently, some clients walk through the timeline to
+render e.g. display names of users at the time they sent the message (rather than their current display name), though
+e.g. Element clients have moved away from this UX.
+This MSC allows to compute the current state correctly. It does not fix the behavior for the state history in the timeline. So clients using such an approach still will face the same issues they had before this MSC.
```
andybalaam:
```suggestion
Both before and after this proposal, clients are not able to calculate reliably exactly when in the
timeline the state changed (e.g. to figure out which message should show a user's previous/updated
display name - note that some clients e.g. Element have moved away from this UX). This is because
the accurate picture of the current state at an event is calculated by the server based on the room
DAG, including the state resolution process, and not based on a linear list of state updates.

This proposal ensures that the client has a more accurate view of the room state *after the sync has
finished*, but it does not provide any more information about the *history of state* as it relates
to events in the timeline. Clients attempting to build a best-effort view of this history by walking
the timeline may still do so, with the same caveats as before about correctness, but they should be
sure to make their view of the final state consistent with the changes provided in `state_after`.
```

proposals/4222-sync-v2-state-after.md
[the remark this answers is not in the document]
fkwp: Addressed by https://github.com/matrix-org/matrix-spec-proposals/pull/4222/commits/8d4d12317781f0f4e6eee74ab6aec33363327be0

proposals/4222-sync-v2-state-after.md
fkwp: addressed by https://github.com/matrix-org/matrix-spec-proposals/pull/4222/commits/38b0f091fd2fa51c0df3e4bfa0f6741238633feb

fkwp:
> lgtm provided these threads are incorporated into the MSC:
> 
> * https://github.com/matrix-org/matrix-spec-proposals/pull/4222/files/fbcf1fc8a3f47cc2dc545e09059a9a2af1ee6e62#r2054361343
> * https://github.com/matrix-org/matrix-spec-proposals/pull/4222/files#r1821125000

addressed by:
- https://github.com/matrix-org/matrix-spec-proposals/pull/4222/commits/8d4d12317781f0f4e6eee74ab6aec33363327be0
- https://github.com/matrix-org/matrix-spec-proposals/pull/4222/commits/38b0f091fd2fa51c0df3e4bfa0f6741238633feb

richvdh:
@clokep are you happy to resolve your concern now?

clokep:
@mscbot resolve Unresolved threads

KitsuneRal (commented):
A cosmetic change that got stuck in pending.

proposals/4222-sync-v2-state-after.md
toger5: @erikjohnston I agree this can get committed, we do mention explicitly that the HS has to omit the `state` section (and replace it with `state_after`).

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2187

uhoreg:
Merged! :tada:

PR #4225, opened 2024-11-06
MSC4225: Specification of an order in which one-time-keys should be issued

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/otks_in_upload_order/proposals/4225-one-time-key-ordering.md)

Implementation: https://github.com/element-hq/synapse/pull/17903, https://github.com/element-hq/synapse/pull/17934

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4225#issuecomment-2484390002)

### DIFF
--- a/proposals/4225-one-time-key-ordering.md
+++ b/proposals/4225-one-time-key-ordering.md
@@ -29,13 +29,14 @@ This proposal is not a complete solution to the problem of premature discarding
 of one-time keys: even if the server issues a recent one-time key, it is still
 possible for a to-device message to be delayed so long that the recipient has
 discarded the private part of the one-time key. It is, however, a significant
-improvement. A possible solution is for clients that expect to be used in
-conditions of poor connectivity to keep old OTKs for longer.
+improvement. A possible future solution is for clients that expect to be used
+in conditions of poor connectivity to keep old OTKs for longer.
 
 There are other ways in which the server and client can get out of sync with
 respect to one-time keys, including by a [database
 rollback](https://github.com/element-hq/element-meta/issues/2155), or
-implementation defects.
+implementation defects. It is anticipated that other solutions will be found
+for those situations.
 
 ## Alternatives
 
@@ -53,7 +54,7 @@ implementation defects.
 
 2. [MSC4162](https://github.com/matrix-org/matrix-spec-proposals/pull/4162)
    proposes a mechanism by which a client can inform the server that it is
-   discarding certiain OTKs, so that the server can also remove the public
+   discarding certain OTKs, so that the server can also remove the public
    keys. This seems a heavier-weight solution to the problem.
 
 3. A more invasive alternative would be to design an encryption stack which
@@ -82,7 +83,7 @@ None.
 
 ## Appendix: detailed explanation of the failure mode
 
-### Backgroynd
+### Background
 
 End-to-end encryption in Matrix relies on individual devices sharing
 [Megolm](https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md)
@@ -113,6 +114,8 @@ it can generate new ones when required.
 See [One-time and fallback
 keys](https://spec.matrix.org/v1.12/client-server-api/#one-time-and-fallback-keys)
 which specifies much of this behaviour.
+
+### Problem
 
 Clearly, a device must retain the private part of each one-time key until that
 key is used to establish an Olm session. However, for a number of reasons,
--- a/proposals/4225-one-time-key-ordering.md
+++ b/proposals/4225-one-time-key-ordering.md
@@ -1,0 +1,109 @@
+# MSC4225: Specification of an order in which one-time-keys should be issued
+
+## Background
+
+End-to-end encryption in Matrix relies on individual devices sharing
+[Megolm](https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md)
+message keys via [to-device
+messages](https://spec.matrix.org/v1.12/client-server-api/#send-to-device-messaging)
+which are themselves encrypted using the
+[Olm](https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/olm.md)
+ratchet.
+
+Suppose Alice has wishes to send an Olm-encrypted message to Bob (with whom she
+has not previously established an Olm session). When Bob logged in, his device
+will have created a long-term identity key, as well as a number (typically 50)
+of one-time keys. Each of these keys has a private part, which Bob's device
+retains locally, and a public part, which Bob's device publishes to his
+homeserver via the
+[`/keys/upload`](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3keysupload) endpoint.
+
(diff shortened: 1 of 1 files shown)


turt2live (approved):
only briefly read the MSC, but seems sane at the core

turt2live:
@mscbot fcp merge

proposals/4225-one-time-key-ordering.md
```diff
@@ -0,0 +1,155 @@
...
+2. [MSC4162](https://github.com/matrix-org/matrix-spec-proposals/pull/4162)
+   proposes a mechanism by which a client can inform the server that it is
+   discarding certain OTKs, so that the server can also remove the public
+   keys. This seems a heavier-weight solution to the problem.
+
+3. A more invasive alternative would be to design an encryption stack which
+   does not rely on one-time keys. Setting aside whether the security
+   properties of such a protocol would be sufficient, this is considered well
+   out-of-scope for this proposal.
+
```
dbkr, line 80: Could another alternative be for the server to returns the IDs of all the keys it has rather than just how many? Plus maybe information about which keys have been claimed and when, and a mechanism for the client to reset the OTKs held by the server. It could then detect any anomalies and reset, and enforce a maximum time for a key to be used after being claimed.

That said, I think changing the endpoint to return the keys in the order they're uploaded is a sensible thing to do anyway, so this is probably more academic.

richvdh: > Could another alternative be for the server to returns the IDs of all the keys it has rather than just how many? Plus maybe information about which keys have been claimed and when, 

I guess so. If a server says either "I've still got this key ID" or "I issued this key ID recently" then clients would know not to throw that key away yet. But is there any reason at all to prefer that over specifying that servers issue keys in order, so clients can just throw away the oldest keys first?

>  mechanism for the client to reset the OTKs held by the server.

Would you want to reset? Or just tell a server that you've forgotten about specific keys? Either way, it's MSC4162.

dbkr: > But is there any reason at all to prefer that over specifying that servers issue keys in order, so clients can just throw away the oldest keys first?

Just that then the client would know exactly which ones it need to keep, rather than just reducing the chance of a valid one being discarded?

> Would you want to reset? Or just tell a server that you've forgotten about specific keys?

I think either would be fine, at least for the purposes of this discussion.

richvdh: Either way: if the server issues the keys in order, then it's much easier to tell the client which ones it still has: "everything more recent than key X", as opposed to "key A, B, C, D, etc".

It's then questionable whether the client can *do* anything with that information. If it has to throw some keys away, it has to throw some away, regardless of whether the server still has them. (A reset endpoint would help, but again: it's a separate change).

dbkr: Mmm, yes. I was mostly trying to fill in the gap of what a 'proper' solution would actually look like since this is deliberately a "better but not 100% perfect" change. If the answer is, "have a reset endpoint" then that's fine.

richvdh: Possibly the comment's in the wrong place, but the only real idea I have for fixing the rest of the problem (dropping keys *after* the public key has been claimed, but *before* the to-device message arrives) is described in the "problems" section:

> A possible future solution is for clients that expect to be used in conditions of poor connectivity to keep old OTKs for longer.

proposals/4225-one-time-key-ordering.md
```diff
@@ -0,0 +1,155 @@
...
+produce (6 characters vs 11 characters).
+
+## Potential issues
+
+This proposal is not a complete solution to the problem of premature discarding
+of one-time keys: even if the server issues a recent one-time key, it is still
+possible for a to-device message to be delayed so long that the recipient has
+discarded the private part of the one-time key. It is, however, a significant
+improvement. A possible future solution is for clients that expect to be used
+in conditions of poor connectivity to keep old OTKs for longer.
```
erikjohnston, line 49: AIUI there's almost always going to be situations where the client receives a `to-device` message using an OTK it no longer has. Is there any mechanism to handle this situation, e.g. a reply asking the sender to retry?

proposals/4225-one-time-key-ordering.md
```diff
@@ -0,0 +1,155 @@
...
+   particular, a OTK is much smaller than a Megolm key, and there is no limit
+   to the number of Megolm keys that a client has to retain.
+
+   In short though, this just seems inefficient, compared to specifying that
+   OTKs should be issued in upload order.
+
+2. [MSC4162](https://github.com/matrix-org/matrix-spec-proposals/pull/4162)
+   proposes a mechanism by which a client can inform the server that it is
+   discarding certain OTKs, so that the server can also remove the public
+   keys. This seems a heavier-weight solution to the problem.
```
erikjohnston, line 74: Sort of wondering whether clients should replace all OTKs on the server at once, rather than appending new ones to the list? That way whenever the client replaces all the OTKs its easier to then say "we can discard the previous batch of OTK private keys in N days". 

Just an ideal non-blocking thought.

proposals/4225-one-time-key-ordering.md
```diff
@@ -0,0 +1,155 @@
...
+
+Clearly, a device must retain the private part of each one-time key until that
+key is used to establish an Olm session. However, for a number of reasons,
+ranging from network errors to malicious activity, it is possible for a claimed
+one-time key never to be used to establish an Olm session.
+
+This presents a problem: there is a limit to the number of private one-time
+keys that a client can retain. Over time, as keys are repeatedly claimed,
+replaced with newly-generated keys, but not actually used, the client must
+start to discard older keys.
```
erikjohnston, line 144: OOI how long do clients typically retain these keys?

richvdh: libolm kept 100. Vodozemac keeps 5000.

proposals/4225-one-time-key-ordering.md
richvdh: Not at the moment. We've had thoughts about that sort of thing. TBH though, I think it's pretty unlikely to happen once we fix the ordering problem, unless the sender literally goes offline for a year just after claiming the OTK.

proposals/4225-one-time-key-ordering.md
richvdh: erm... not really following how that would work/help.

dbkr: This is the kind of idea I was toying with in my discussion: replacing the whole lot every time would have the effect of removing any keys that the device had deleted from the server. It's probably a similar conclusion to that conversation.

ara4n:
Let the record show that i was having a slight wobble over some dim memory of whether there was some impact on deniability by being able to predict the order that OTKs are consumed.  Looks like there isn't, but for posterity, the raw discussion was:

Screenshot 2024-11-28 at 14 03 40

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2029

richvdh:
merged!

PR #4226, opened 2024-11-12
MSC0000: Reports as rooms.

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/reports-as-rooms/proposals/0000-reports-as-rooms.md)

Signed-off-by: Gnuxie Gnuxie@protonmail.com

---

This MSC was written in a volunteer capacity without any $funding :feet:

### DIFF
--- a/proposals/0000-reports-as-rooms.md
+++ b/proposals/0000-reports-as-rooms.md
@@ -1,0 +1,286 @@
+# MSC0000: Reports as rooms
+
+Currently it is possible to report users, rooms, and events to the Matrix
+homeserver. However there are missing parts of the specification:
+
+1. Receiving reports in a Matrix client is entirely unspecified.
+2. There is no way to report any content to room moderators.
+
+## Proposal
+
+Compared to the alternative proposals, this MSC requires significantly
+less supporting infrastructure, and is both very simple and
+extensible.
+
+We propose a new room type `m.report` which will be used to present
+reports to Matrix server admins or room moderators.  This method of
+reporting is expected to replace the legacy methods of reporting, via
+the `/report` endpoints in the client-server API.
+
+Although each of these flows are written from the perspective of a
+client, it is an implementation detail whether Alice's client, Alice's
+server, or Mike's server creates the room. This is in order to support
+compatibility with the existing reporting endpoints or those found in
+similar proposals.
+
+### Report to room moderator flow
+
+Alice is participating in a Matrix room dedicated to sharing Cat
+pictures that is moderated by Mike and Laura.
+
+Bob is also in the room, and keeps uploading memes from ifunny.co
+while Mike and Laura are busy.  Alice decides to report Bob.
+
+1. Alice selects Bob's most recent message in her client
+2. Alice selects report.
+3. Alice is presented with a view showing the event that is being
+   reported. Mike and Laura's profiles are also shown in the view and
+   it is clear to Alice they will be the moderators of the report.
+4. Alice explains the situation in the `reason` field and submits the
+   report.
+5. Alice's client creates a new Matrix room with the type
+   `m.room.report`.  Alice's client invites both Mike and Laura as
+   part of the `createRoom` process.
+6. Alice can find the report in her client under a reports tab, and
+   can see that neither Mike or Laura have responded yet.
+
+### Report moderator flow
+
+Mike receives a notification within his client that a new report has
+been received with minimal detail.
+
+1. Mike opens a report view where he can see an overview of the new
+   report. Mike reads the MXID of the report submitter,
+   `@alice:amazing-linux.example.com`, and the entity that is the the
+   subject of the report. In this instance the entity is the mxid
+   `@bob:example.com`. Mike also notes that the report has been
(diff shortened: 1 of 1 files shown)


proposals/4226-reports-as-rooms.md
```diff
@@ -0,0 +1,286 @@
...
+
+Bob is still misbehaving and has now started harassing the room moderators.
+
+Mike decides to escalate the report to bob's server admins.
+
+1. Mike selects `escalate` option in his client, opening a view with the
+   profiles of the report moderators of Bob's server.
+2. Mike's client invites the report moderators of Bob's server to the room.
+
+### Report content mixin
```
Gnuxie, line 119: Jassuko suggested:

> Can you figure a way to allow linking to multiple events instead of just a single event? Some times more context would be nice for decision making and gaining a full picture might need pointers to events that are not immediately next to each other.

Gnuxie: I replied:

> normally one event is enough just to anchor it and we expect moderators to go looking for the surrounding context, but yeah as you say that's a problem when stuff is spread out. it's difficult because i don't know if any client has the ux that would let you select multiple events like that? But I guess it would be possible to either allow them to send more events with context or add an array or something, idk

proposals/4226-reports-as-rooms.md
```diff
@@ -0,0 +1,286 @@
...
+server admins. Mechanisms for better handling of invitations generally
+need to be expanded upon urgently in Matrix.  The only novel
+mitigation presented in this MSC is that the reporter profile is
+hidden from any listing of reports and any notification.
+
+## Alternatives
+
+### [MSC4121](https://github.com/matrix-org/matrix-spec-proposals/pull/4121):  The `m.role.moderator` `/.well-known/matrix/support` role
+
+This role could be reused instead of the proposed `m.role.report_moderator`.
```
Gnuxie, line 257: We should do this and add a flag to the MSC4121 role for whether they should receive reports

proposals/4226-reports-as-rooms.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

deepadmax:
I imagine that a report room could be end-to-end encrypted; the subject may be sensitive. If that's the case, you may not want to store the *reason* unencrypted either. What if the reason would be stored elsewhere? A state event could refer to the event ID of the message to be used as the reason; by default, it would be the first message. Another benefit to storing it as state rather than in the creation event is that the reason can be changed later to better describe the problem.

proposals/4226-reports-as-rooms.md
```diff
@@ -0,0 +1,286 @@
...
+
+#### Extending `/.well-known/matrix/support` (already accepted into spec via MSC1929)
+
+For reporting to server administrators, we propose a new support role,
+`m.role.report_moderator`.
+
+When a report needs to be sent to server administrators, each of the
+the `m.role.report_moderator`'s will be invited to the report room.
+
+## Potential issues
```
Gnuxie, line 225: https://github.com/matrix-org/matrix-spec-proposals/pull/4226#issuecomment-2474477010

> I imagine that a report room could be end-to-end encrypted; the subject may be sensitive. If that's the case, you may not want to store the reason unencrypted either. What if the reason would be stored elsewhere? A state event could refer to the event ID of the message to be used as the reason; by default, it would be the first message. Another benefit to storing it as state rather than in the creation event is that the reason can be changed later to better describe the problem.

(@deepadmax thank you for reading the MSC :) btw it's not a big deal but next time you open a thread somewhere by clicking on the file so that we can reply easily ^^)

Gnuxie: This is true, if the report was sent in messages once the report room was created then it could be end to end encrypted. However, doing so would complicate the MSC. (See power level https://github.com/matrix-org/matrix-spec-proposals/blob/8703e19c65e0aac9b7ff9959dc04a69b006df357/proposals/4226-reports-as-rooms.md?plain=1#L153-L168).

Currently the reporter is supposed to deliberately yield power level in the room in order to give the report moderators control. The client's of the report moderator are supposed to check that the power level has been yielded immediately. The idea being that this would minimize attempts to harass report moderators by flooding them with reports that have abusive content.

Maybe it is overly cautious though. If clients can hide messages behind spoiler text, and can blur images (both things that Nheko and Element web can do for instance) by default, then it might not be so bad to allow the reporter to send messages to the room.

It is probably then worthwhile turning the report context into a timeline event and allowing multiple context events to be sent within one report (so this would replace the mixin `m.report.*` that is present within the `m.room.create` event, and they could be encrypted too).

It would be nice to know if you or anyone else have thoughts about that before I rewrite these parts of the MSC?

Gnuxie: Doing this would close off the route of backwards compatibility though when reports are intended to be created by servers when the legacy reporting client server api endpoints are used. Unless we just allow those to be sent without encryption (because then at least one server in this instance is already able to intercept the contents of the report).

PR #4227, opened 2024-11-15
MSC4227: Audio based quick login

albertotirla:
[Rendered](https://github.com/albertotirla/matrix-spec-proposals/blob/audio-quick-login/proposals/4227-audio-based-quick-login.md)

### Pull Request Checklist

* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/master/CONTRIBUTING.md#sign-off)
* [x] Update the title and file name of your proposal to match this PR's number (after opening).
* [x] Pull request includes a ['Rendered' link](https://matrix.org/docs/spec/proposals#process) above.
* [x] Your MSC adheres to each point in the [MSC Checklist](MSC_CHECKLIST.md).
* [x] Ask in
      [#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) to
      get feedback on this PR.

This MSC gives visually impaired people a way to use the quick login mechanism, without having to scan a QR code, which is impossible to do reliably with any current assistive technologies

### DIFF
--- a/proposals/4227-audio-based-quick-login.md
+++ b/proposals/4227-audio-based-quick-login.md
@@ -269,5 +269,5 @@ not applicable here
 
 ## Dependencies
 
-This MSC builds on MSC4108, which at the time of writing has not yet been accepted into the spec.
+This MSC builds on [MSC4108](https://github.com/matrix-org/matrix-spec-proposals/pull/4108), which at the time of writing has not yet been accepted into the spec.
 


proposals/4227-audio-based-quick-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+The only thing this MSC changes from the protocol outlined in the dependent MSC is the transport of the initial secret, the binary string represented through a qr code. The cryptographic primitives, the insecure session channel, the way in which  it is turned into a secure one, the strings exchanged between the two clients remain unchanged, which means that a lot of the mechanisms are in place already.
+
+Like in the dependent MSC, any device can generate the audio signal, same for recieving it. This means that the login is more likely to succeed, all one needs is a single device which has a microphone, the other having speakers functional is implied because this is providing accessibility to visually impaired people, who definitely have speakers working otherwise they wouldn't have speech output.
+
+The mechanism used for transmitting the audio signal across devices is morse code, as standardised and described by International Telecommunication Union [at this address](https://www.itu.int/rec/R-REC-M.1677-1-200910-I/). Here are afew reason for which this way of communication was chosen, as well as what criteria other protocols would have to satisfy in order to be considered:
```
anoadragon453, line 15: Have you considered other libraries (that include error correction) that use pitch to encode more data per beep? https://github.com/ggerganov/ggwave is a well-known library by the creator of `llamacpp`. It may result in shorter transmission times. However it would be less widely supported than morse code.

There's also ultrasonic protocols - those which aren't audible to regular people: https://github.com/jamesonrader/AudioQR. This has the advantage of not letting others know (unless they're actively searching) that you're broadcasting a code. A curious CCTV operator may go back and try to decode morse code that they happen to pick up in a recording.

However, a "silent" protocol does mean that a blind user won't know when a code start being sent across devices and ends. And if this is announced out loud to the user, then anybody eavesdropping would be made aware of that as well.

thoraj:
While I agree with goals of this MSC, I'm very doubtful it will work as specified. 
* the use of morse code seem strange. Since morse code was invented, there have been numerous innovation of more efficient ways to transmit information over audio channels  (FSK, DTMF or more advanced QAM methods) 
* the transmission must most likely use some form of error correcting code get the desired accuracy
* transmitting in the band 50khz-80khz (ultrasound) will be hard/impossible with standard phone hardware. Android samples at 44,1khz or 48khz afaik, which theoretically limits bandwidth to 22khz (Nyquist limit)
* 

Using more proven methods and protocols rather than inventing something new, seems more likely to succeed .

proposals/4227-audio-based-quick-login.md
albertotirla: those subsonic and ultrasonic protocols have two obvious disadvantages:

* they are more likely to trigger emergency alert protocols from our phones, because a lot of the beyond human hearing stuff in those protocols happen at those kinds of frequencies. It wouldn't be nice for a user's phone to call the police just because they want to login on matrix, or worse, to trigger emergency tones from other nearby devices
* we have to include bad microphones in the equation here, a lot of the ultrasonic protocols out there depend on people having stelar hardware, which is definitely not always the case, meanwhile morse has a history of being sent over very much flawed radio waves

There are similar problems with some of the libraries you listed:

* they must be as widely available as possible. QR code scanning might not be an issue nowadays, but a lot of those fancy protocols are written for only one or afew programming languages, are hardly standardised whatsoever, etc
* they must work with low-powered devices, as much as possible anyway, which means, there shouldn't be heavy computing involved, nor very high-quality sound hardware required, on either end of the spectrum

About people trying to decode a morse code appearing in their recordings, that's not an issue, they'd only be getting base64 encoded, lzma compressed, binary sequences which don't mean anything, plus they expire quickly anyway, so I don't think that's much a problem, except probably a nusence to tv reporters who can't understand why a bit of morse would ever decode to binary haha

albertotirla:
I will look more indepth at FSK, because this one seemns quite promising. The problem is, as always, if we have bindings for this protocol for at least as many programming languages as we have matrix sdks. I looked at something similar before, and the problem with all these is that we don't have enough bindings for this stuff as we have for qr code generation and scanning, which is probably even less than other methods.

proposals/4227-audio-based-quick-login.md
tezlm: Another option worth considering is [dtmf](https://en.wikipedia.org/wiki/DTMF). It's been used in telecommunications for decades, works with bad microphones/speakers, and is standardized.

albertotirla: of all the options people told me about so far, this one seemns by far the best to replace morse, especially because it has two tones per state, so you can directly encode binary sequences by recording the most significant bits in the upper tone, while the least significant bits go in the lower tone, though I doubt that's in the actual standard document. This is a disadvantage compared to morse code, because with morse I can encode a letter directly, but with this I would potentially no longer need to base64 encode the binary sequence at all, dk about lzma compression though, that might have to stay. The problem again is availability, because at a glance I couldn't find many libraries which encode and decode these types of signals, but I found enough of them to still be able to consider it. Yes, telephone touch tones are definitely usable in bad speaker scenarios, and they have been around for a long time, and as long as they don't trigger the dialer of any nearby phones, I'm good lol

proposals/4227-audio-based-quick-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+the binary code, the exact same one which would have been used for creating a QR code from, undergoes the following transformations:
+
+* it gets compressed using the LZMA2 compression algorythm, so that the length of the transmission over morse is as short as possible. The following additional parameters are being used for LZMA:
+  * level, 8
+  * LC, 4
+  * FB, 256
+  * MC, 10000
+  * writeEndMark, 1
+* base64 encoding is being aplied, in order to not make the morse code encoder error out when parsing binary characters resulting from the compression
```
tonkku107, line 33: The message could be a lot shorter if it was transmitted directly as binary. 0 could be a dot and 1 could be a dash. Are there no encoders and decoders out there capable of doing this?

proposals/4227-audio-based-quick-login.md
```diff
@@ -0,0 +1,102 @@
...
+
+Finally, decompression, with the same compression parameters as before, is being aplied, in order to get the initial code. The rest of the login flow is being followed precisely as written in the dependent MSC, so it will not be repeated here
+
+### note regarding accessibility and user convenience
+
+Because the situation where one client only supports qr code, while the other only supports morse is not desirable and should be avoided, client implementors should do the following:
+
+* if a client implements QR code login, it is strongly recommended that it also implements audio login, for accessibility reasons, because even if the other client supports audio login, the VI person still can't do anything to get the information from that qr code
+* if a client implements audio login, it is not required to also implement QR loggin, because audio login is accessible to everyone
```
tonkku107, line 62: I wouldn't include that clients are not required to implement QR code login. As you mention, this is assuming visually impaired people have speakers but that may not be true for the rest. I personally only have a headset, so it'd be awkward to have to shove my phone into the earcups to log in. I think it makes more sense to require both methods.

proposals/4227-audio-based-quick-login.md
```diff
@@ -1,117 +1,102 @@
...
+
+the binary code, the exact same one which would have been used for creating a QR code from, undergoes the following transformations:
+
+* it gets compressed using the LZMA2 compression algorythm, so that the length of the transmission over morse is as short as possible. The following additional parameters are being used for LZMA:
+  * level, 8
+  * LC, 4
+  * FB, 256
+  * MC, 10000
+  * writeEndMark, 1
+* base64 encoding is being aplied, in order to not make the morse code encoder error out when parsing binary characters resulting from the compression
```
akarl10, line 33: Correct me if I am wrong, but isn't morse upper case only (or at least does not distinguish between upper and lower case)? If that is true base64 is not really an option since that requires a-z and A-Z. An alternative might be base32

proposals/4227-audio-based-quick-login.md
akarl10: Agree. Audio login might also be useful for the (unlikely) case where one has a pc without webcam but a bluetooth headset where the mobile phone might be the device that allows the pc to login.

proposals/4227-audio-based-quick-login.md
akarl10: hm. I'm not quite sure about that. Comparing with base32 a letter encodes 5 bits. In morse a letter is  4 beeps or shorter. Numbers in morse are 5 beeps so for them we would be equal, but that's the worst case so on average base32 should be faster

proposals/4227-audio-based-quick-login.md
albertotirla: good catch! I'll revisit that shortly, after I come to a decision regarding whether I should keep using morse code or switch to phone touch tones

akarl10: I think DTMF would work quite well since with one tone we would send 4 bits including ABCD and both +# (https://en.wikipedia.org/wiki/DTMF). There is no need for binary encoding/decoding and the transmission speed is "constant" (4 bits per tone, so if we take 100ms per tone, 25ms pause we would send 4 bytes per second)

proposals/4227-audio-based-quick-login.md
albertotirla: if we're going to use morse to the end of this MSC, then we should use it as specified by its standard. While any implementation which converts data to sound via short beeps would be similarly valid, we're trying to reduce the difficulty for people to actually implement the spec, not increase it. Right now, the decision is between keeping morse and changing to DTMF, because the latter one accounts for error correction at least somewhat and has a wider band for sending more complicated values, which could make our code shorter to transmit and recieve

proposals/4227-audio-based-quick-login.md
albertotirla: yes, that would be quite good, except that I don't know whether the key and the rest of the code fit in a...d and 1...10, and if they don't, we are possibly deviating from the standard. Unless I'm reading the standard wrong?

proposals/4227-audio-based-quick-login.md
albertotirla: noted, so they should implement both by the spec. What about cases where the interface is invisible, or the client is written in such a way that displaying images properly on the screen isn't possible? I will add whenever possible in both instances, because the reverse could be true as well, so leaving clients a bit of room to wiggle out of impossible situations. I only hope this wouldn't be used by clients to clame this isn't possible for them, however as people said before, a spec isn't the law, so that's that, I should expect people to not follow the spec from time to time, whatever I recommend to the contrary.

proposals/4227-audio-based-quick-login.md
akarl10: > yes, that would be quite good, except that I don't know whether the key and the rest of the code fit in a...d and 1...10, and if they don't, we are possibly deviating from the standard. Unless I'm reading the standard wrong?

no, you are right, the 4th column (abcd) might not be in every implementation, although afaik it should be standardized, just normal phones never had that keys. If we stick to 3 columns, then we have 12 possible "tones". That obviosly does not map cleanly to bits (8 tones would be 3 bits, 12 tones "3.5", but by combining two consecutive tones we would have 7 bits with the disadvantage that we need to successfully transmit 2 tones or we loose both. Don't know if that would impact the end result)

albertotirla: we have two options, as I see them:

* we could convert everything to octal, even if we lose two possible bits, and then we use * and # as begin and end of the whole code, * means the recording program should discard everything before that, while # means that the recording program should stop recording because we got all the code
* we deviate from the standard, and consider the first four bits of a byte the first tone, while the second four the other tone. The huge disadvantage here is that this no longer maps to clean numbers in that table, and that we have to assemble the nibbles, aka half bytes, from two consecutive tones, and I dk if any encoder or decoder actually does that

Either way, with LZMA compression aplied, perhaps that will reduce the code enough to the point where it's more easily transmittable

akarl10: I would stick to the first one. it's not that you loose two bits. With the 12 codepoints->3.something (actually 12 does fit quite badly) bits I meant that you can squeeze out some additional bandwith by sending two tones (to explain better some examples)
one tone -> octet/3bit:
```
0 -> 000
1 -> 001
2 -> 010
...
7 -> 111
```
two tone 7 bit (its first tone * 12 + second tone, 0=0, *=10 #=11):
```
0 0 -> 0000000
0 1 -> 0000001
..
0 7 -> 0 000 111
0 8 -> 0 001 000
0 9 -> 0 001 001
0 * -> 0 001 010
0 # -> 0 001 011
...
1 0 -> 0 001 100
1 1 -> 0 001 101
1 # -> 0 010 111
...
* 0 -> 1 111 000
...
* 7 -> 1 111 111
```
as you can see with that not every combination would be vaid, but * * and # # could be used for start and stopp.

albertotirla: so basically

* we use the standard, except it works on groups of 4 bits, not a whole byte
* if we were to compare this to hexadecimal, the * and # characters would be like E and F
* begin and end of message could be ** and ##, but what happens if the byte sequence contains 0xFF somewhere, is that even possible with the actual data we have to encode?

Hmm, perhaps we should look at code like this, in more detail, though I dk if it encodes more than the usual standard

https://github.com/gbonacini/dtmflib.git

albertotirla: actually yeah, let's not overcomplicate things, perhaps it's better to just convert everything to octal, though how do we encode byte separators is another unknown problem, unless we use a * for that

tonkku107: I grabbed a live QR code and tried LZMA compression on its binary data and the output was actually slightly larger (136 bytes originally, 162 bytes compressed)

albertotirla: did you use LZMA compression with the parameters specified in this msc? If so, I thought that may happen, which means this MSC is about to get that much simpler

tonkku107: I was able to configure the level but not the others. not sure how much impact they have

albertotirla: ok, will remove the compression requirement as soon as I make the necesary adjustments to switch to DTMF

albertotirla: alright, I did all that, removed the compression requirement and switched to DTMF, although I'm not sure if it will help matters in the long run because the code might be even longer now than with morse. The endianness might be slightly wrong, so it may not work with the example program I found on github and posted somewhere above, because I think that's actually big endian, but this thread has served its purpose, so I think it's alright to close it, people are free to reopen it anyway

proposals/4227-audio-based-quick-login.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4227-audio-based-quick-login.md
```diff
@@ -0,0 +1,116 @@
+# MSC4227: Audio based quick login
+
+MSC4108 allows the matrix ecosystem to offer users a quicker way to sign in trusted devices, using a QR code based workflow. This is familiar to a lot of users, as they were exposed to this workflow on other networks.
+
+However, the pains for visually impaired users regarding this workflow are the same here as they are on other devices, because it is nearly impossible, with current assistive technologies, to reliably scan a QR code. Because of that, any workflow involving those is unreasonably difficult, to the point that there are lots of workarounds developed for these situations, some working only some of the time, while others not at all.
```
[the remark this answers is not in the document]
albertotirla: I'm not sure how I can do that, because I can't know the status bar of my text editor to see the column number I'm on. There might hopefully be an editor plugin to do this, but I'm not sure of any that's accessible to VI people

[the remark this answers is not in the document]
andybalaam: It seems like you found this, but in case others chance upon this:

```
:set textwidth=100
```

then select everything with `ggVG` and wrap with `gq`.

proposals/4227-audio-based-quick-login.md
anoadragon453: I've updated the comment.

albertotirla: ok, thanks! I'll see what can be done about the client requirement, will try to talk to a client developer who implemented qr login and see what they think about implementing this too

PR #4229, opened 2024-11-20
WIP: MSC4229: Pass through `unsigned` data from `/keys/upload` to `/keys/query`

richvdh:
This is just a placeholder for now


richvdh:
sooo, we're putting this work (https://github.com/element-hq/element-meta/issues/2467) on indefinite hold; I'm therefore closing this MSC for now. If we come back to it, we'll reopen.

PR #4230, opened 2024-11-21
[Placeholder] Flag for animated images

dbkr:
This is a placeholder to reserve an MSC number for a flag indicating whether an image in animated (see https://github.com/element-hq/element-web/issues/28311#issuecomment-2490919849).

### DIFF
--- a/proposals/xxxx-animated-image-flag.md
+++ b/proposals/xxxx-animated-image-flag.md
@@ -1,0 +1,117 @@
+# MSC0000: Template for new MSCs
+
+*Note: Text written in italics represents notes about the section or proposal process. This document
+serves as an example of what a proposal could look like (in this case, a proposal to have a template)
+and should be used where possible.*
+
+*In this first section, be sure to cover your problem and a broad overview of the solution. Covering
+related details, such as the expected impact, can also be a good idea. The example in this document
+says that we're missing a template and that things are confusing and goes on to say the solution is
+a template. There's no major expected impact in this proposal, so it doesn't list one. If your proposal
+was more invasive (such as proposing a change to how servers discover each other) then that would be
+a good thing to list here.*
+
+*If you're having troubles coming up with a description, a good question to ask is "how
+does this proposal improve Matrix?" - the answer could reveal a small impact, and that is okay.*
+
+There can never be enough templates in the world, and MSCs shouldn't be any different. The level
+of detail expected of proposals can be unclear - this is what this example proposal (which doubles
+as a template itself) aims to resolve.
+
+
+## Proposal
+
+*Here is where you'll reinforce your position from the introduction in more detail, as well as cover
+the technical points of your proposal. Including rationale for your proposed solution and detailing
+why parts are important helps reviewers understand the problem at hand. Not including enough detail
+can result in people guessing, leading to confusing arguments in the comments section. The example
+here covers why templates are important again, giving a stronger argument as to why we should have
+a template. Afterwards, it goes on to cover the specifics of what the template could look like.*
+
+Having a default template that everyone can use is important. Without a template, proposals would be
+all over the place and the minimum amount of detail may be left out. Introducing a template to the
+proposal process helps ensure that some amount of consistency is present across multiple proposals,
+even if each author decides to abandon the template.
+
+The default template should be a markdown document because the MSC process requires authors to write
+a proposal in markdown. Using other formats wouldn't make much sense because that would prevent authors
+from copy/pasting the template.
+
+The template should have the following sections:
+
+* **Introduction** - This should cover the primary problem and broad description of the solution.
+* **Proposal** - The gory details of the proposal.
+* **Potential issues** - This is where problems with the proposal would be listed, such as changes
+  that are not backwards compatible.
+* **Alternatives** - This section lists alternative solutions to the same
+  problem which have been considered and dismsissed.
+* **Security considerations** - Discussion of what steps were taken to avoid security issues in the
+  future and any potential risks in the proposal.
+
+Furthermore, the template should not be required to be followed. However it is strongly recommended to
+maintain some sense of consistency between proposals.
+
+
+## Potential issues
+
+*Not all proposals are perfect. Sometimes there's a known disadvantage to implementing the proposal,
+and they should be documented here. There should be some explanation for why the disadvantage is
+acceptable, however - just like in this example.*
+
+Someone is going to have to spend the time to figure out what the template should actually have in it.
+It could be a document with just a few headers or a supplementary document to the process explanation,
+however more detail should be included. A template that actually proposes something should be considered
+because it not only gives an opportunity to show what a basic proposal looks like, it also means that
+explanations for each section can be described. Spending the time to work out the content of the template
+is beneficial and not considered a significant problem because it will lead to a document that everyone
+can follow.
+
+
+## Alternatives
+
+*This is where alternative solutions could be listed. There's almost always another way to do things
+and this section gives you the opportunity to highlight why those ways are not as desirable. The
+argument made in this example is that all of the text provided by the template could be integrated
+into the proposals introduction, although with some risk of losing clarity.*
+
+Instead of adding a template to the repository, the assistance it provides could be integrated into
+the proposal process itself. There is an argument to be had that the proposal process should be as
+descriptive as possible, although having even more detail in the proposals introduction could lead to
+some confusion or lack of understanding. Not to mention if the document is too large then potential
+authors could be scared off as the process suddenly looks a lot more complicated than it is. For those
+reasons, this proposal does not consider integrating the template in the proposals introduction a good
+idea.
+
+
+## Security considerations
+
+**All proposals must now have this section, even if it is to say there are no security issues.**
+
+*Think about how to attack your proposal, using lists from sources like
+[OWASP Top Ten](https://owasp.org/www-project-top-ten/) for inspiration.*
+
+*Some proposals may have some security aspect to them that was addressed in the proposed solution. This
+section is a great place to outline some of the security-sensitive components of your proposal, such as
+why a particular approach was (or wasn't) taken. The example here is a bit of a stretch and unlikely to
+actually be worthwhile of including in a proposal, but it is generally a good idea to list these kinds
+of concerns where possible.*
+
+MSCs can drastically affect the protocol. The authors of MSCs may not have a security background. If they
+do not consider vulnerabilities with their design, we rely on reviewers to consider vulnerabilities. This
+is easy to forget, so having a mandatory 'Security Considerations' section serves to nudge reviewers
+into thinking like an attacker.
+
+## Unstable prefix
+
+*If a proposal is implemented before it is included in the spec, then implementers must ensure that the
+implementation is compatible with the final version that lands in the spec. This generally means that
+experimental implementations should use `/unstable` endpoints, and use vendor prefixes where necessary.
+For more information, see [MSC2324](https://github.com/matrix-org/matrix-doc/pull/2324). This section
+should be used to document things such as what endpoints and names are being used while the feature is
+in development, the name of the unstable feature flag to use to detect support for the feature, or what
+migration steps are needed to switch to newer versions of the proposal.*
+
+## Dependencies
+
+This MSC builds on MSCxxxx, MSCyyyy and MSCzzzz (which at the time of writing have not yet been accepted
+into the spec).


proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,75 @@
...
+to display the image as animated must download the original image for any image format which
+is capabale of being animated. This means they download the full size image for non-animated
+images even when the user never chooses to enlarge the image.
+
+Even clients that animate images on hover must pre-fetch the full size image if they want the
+animation to start without a delay while it's downloaded.
+
+## Proposal
+
+We add an optional boolean flag, `ia_animated` to the `info` object of image events indicating if
```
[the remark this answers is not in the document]
dbkr: I guess just images for now: clarified.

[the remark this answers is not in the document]
sumnerevans: Any reason not to allow it for stickers?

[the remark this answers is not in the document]
dbkr: I'm not really sure to be honest, which is exactly why I'm leaving it for someone else to propose if they think it's appropriate.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
...
+Behaviour when the field is omitted is left up to the client. They might choose to behave as if it is present
+and set to `true`, ensuring backwards compatibility whilst still saving bandwidth for images where the flag
+is present and set to `false`. Perhaps they might decide to change this behaviour once more clients start
+sending the flag.
+
+## Potential issues
+
+Clients may lie about the flag which would cause unexpected behaviour, for example, an image which
+was not presented might then animate when the user clicks to enlarge it, allowing for 'jump scares'
+or similar. Clients may wish to prevent images from being animated if the flag is set to false.
```
HarHarLinks, line 50: Wouldn't be sensible for an implementation to use this flag only as a hint as to whether it should preload the original, and probe the media regardless of when it was downloaded to confirm the animation? I feel like this MSC is recommending a client behaviour that is a tad too "trusting".

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
...
+to display the image as animated must download the original image for any image format which
+is capabale of being animated. This means they download the full size image for non-animated
+images even when the user never chooses to enlarge the image.
+
+Even clients that animate images on hover must pre-fetch the full size image if they want the
+animation to start without a delay while it's downloaded.
+
+## Proposal
+
+We add an optional boolean flag, `is_animated` to the `info` object of `m.image` events indicating if
```
HarHarLinks, line 15: Just to bring it up, we could also give this flag a more general name `prefetch_original` to express "you might wanna prefetch this", and apply it optionally to any attachment type. A client might assume based on the attachment type, mime type, flag, and other metadata combinations, whether it wants to prefetch, e.g. to play an animated image/gif without delay. Then again I fail to come up with other reasons you might want to hints about prefetching attachments.

dbkr: Personally I'm more persuaded by the idea of giving information about the media rather than trying to second guess what the client might want to do.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
...
+    "body": "partyparrot.gif",
+    "file": {
+      [...]
+    },
+    "info": {
+      "w": 640,
+      "h": 480,
+      "mimetype": "image/gif",
+      "size": 35295,
+      "is_animated": "true",
```
Half-Shot, line 34: I assume you meant a boolean `true` rather than the string "true" ?

```suggestion
      "is_animated": true,
```

zecakeh:
This is implemented in Fractal too as of today:

- Ruma: https://github.com/ruma/ruma/pull/1985
- matrix-rust-sdk: https://github.com/matrix-org/matrix-rust-sdk/pull/4503
- Fractal: https://gitlab.gnome.org/World/fractal/-/merge_requests/1889

dbkr:
As per discussion on https://github.com/element-hq/element-web/issues/28311 and since this now has 2 impls, I'm going to

@mscbot fcp merge

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies
  - [ ] Unstable prefixes [consider](README.md#unstable-prefixes) the awkward accepted-but-not-merged state

turt2live:
@mscbot concern Checklist incomplete

(I haven't attempted to pre-populate it this time, sorry)

proposals/4230-animated-image-flag.md
TheArcaneBrony, line 15: Unrelated thread: how would this work with inline emojis?

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
+# MSC4230: 'Animated' flag for images
+
+Since animated images are thumbnailed to a still image, clients need to download the full
+image in order to display it animated. However, there is currently no way of telling that
+an image is animated without downloading the original image. This means that clients wishing
+to display the image as animated must download the original image for any image format which
```
richvdh, line 6: what does "display the image as animated" mean? An annotation in the UI? Or actually making an animated thumbnail?

Either, I guess. Would be good to be clearer though.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
...
+is capabale of being animated. This means they download the full size image for non-animated
```
richvdh, line 7:
```suggestion
is capable of being animated. This means they download the full size image for non-animated
```

proposals/4230-animated-image-flag.md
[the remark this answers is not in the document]
richvdh: +1, let's do it for stickers too.

proposals/4230-animated-image-flag.md
tulir: Emojis can definitely be animated. Might be better to leave adding a new attribute like `data-mx-animated-emoticon` to a different MSC though (perhaps MSC2545, or a completely new one)

richvdh: it might be that we're using different definitions of "emoji".  Are they not just a unicode codepoint?

richvdh: That was definitely not clear from the question! But yes, that sounds like a question for another MSC, given that such things aren't yet specced.

dbkr: It also seems irrelevant since there would be no thumbnail support here anyway.

proposals/4230-animated-image-flag.md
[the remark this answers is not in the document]
dbkr: I'm afraid I am somewhat drowning with other commitments at the moment. If anyone would like to propose an update to this that would encompass stickers then please go ahead, otherwise I will put this on the back burner for a bit.

dbkr:
I think the current state of this is that it's blocked on figuring out what extending it to stickers would involve. However, I have very little knowledge on how stickers work in my head right now, so if anyone would like to chip in with this this would be very welcome. Otherwise, we can consider cancelling it and coming back with a fresh MSC considering both.

proposals/4230-animated-image-flag.md
HarHarLinks, line 15:
```suggestion
We add an optional boolean flag, `is_animated` to the `info` object of `m.image` and `m.sticker` events indicating if
```

HarHarLinks (changes requested):
I think that is all it takes to make this work for stickers.

proposals/4230-animated-image-flag.md
dbkr: Hopefully clearer now.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,77 @@
...
+      "w": 640,
+      "h": 480,
+      "mimetype": "image/gif",
+      "size": 35295,
+      "is_animated": true,
+    }
+```
+
+If this flag is `false`, receiving clients can assume that the image is not animated. If `true`, they should
+assume that it is.
```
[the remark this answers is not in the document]
dbkr: I've changed these, making them both SHOULD as it made more sense to be than MAY, however I've also added explicitly about not trusting the data as I think this is now necessary with the more formal instruction.

proposals/4230-animated-image-flag.md
dbkr: If we can check this to make sure, this seems fine to me.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,79 @@
...
+sending the flag.
+
+## Potential issues
+
+Clients may lie about the flag which would cause unexpected behaviour, for example, an image which
+the client did not tell the user was animated might then animate when the user clicks to enlarge it,
+allowing for 'jump scares' or similar. Clients may wish to prevent images from being animated if the
+flag is set to false.
+
+As above, supporting animated images becomes harder for sending clients because they must work out if
```
Half-Shot, line 53: One thing to call out is that [browsers are dumb](https://github.com/element-hq/element-web/issues/29560) and so there might not be a way for clients to do this reliably, so this flag may end up being a false positive in some cases.

I don't think this really makes the situation any worse, but worth calling out.

dbkr: Done - seem ok?

Half-Shot: Excellent, ta!

proposals/4230-animated-image-flag.md
clokep: What's the status of this?

dbkr: From my PoV I'd just like something concrete showing that this is sufficient and correct for sticker events before adding it.

proposals/4230-animated-image-flag.md
```diff
@@ -0,0 +1,80 @@
...
+
+Even clients that animate images on hover must pre-fetch the full size image if they want the
+animation to start without a delay while it's downloaded.
+
+## Proposal
+
+We add an optional boolean flag, `is_animated` to the `info` object of `m.image` events indicating if
+the image is animated or not. This SHOULD match whether the original image contains animation. Note
+that this will require clients probe the image file for animation. Simpler clients may, therefore,
+choose to not send this value at all, or always set it to false, meaning receiving clients are
```
erikjohnston: What's the rationale for allowing simpler clients to always set to false? Naively it would feel more useful to say simpler clients SHOULD always omit the value? Receiving clients can then decide how to handle that case depending on what they're trying to achieve.

Indeed, further down you say:

> If this flag is `false`, receiving clients SHOULD assume, but not trust, that the image is not animated.

Which isn't something one can assume in the presence of simple clients that always set to false. Then further down:

> Behaviour when the field is omitted is left up to the client. They might choose to behave as if it is present
and set to `true`, ensuring backwards compatibility whilst still saving bandwidth for images where the flag
is present and set to `false`.

The net effect is that this MSC is recommending that receiving clients will have inconsistent behaviours in the presence of different simple clients who chose to do different things (omit vs always set to false).

dbkr: So the idea is that since receiving clients should basically be presenting anything where the flag is false in a way that it can't animate (ie. displaying the thumbnail) although this wouldn't work if it used the original to display it full size. Thinking about it, it's probably much better anyway to say that if a client doesn't know, it should just not set it rather than make up an answer out of nowhere for no reason.

turt2live:
@dbkr this is expected to get put back on the SCT's radar soon - are you able to take a look and resolve some of the open threads?

proposals/4230-animated-image-flag.md
turt2live: this is fine for stickers. It might be nice to link to the spec, but that's hardly an issue: https://spec.matrix.org/v1.17/client-server-api/#msticker

proposals/4230-animated-image-flag.md
erikjohnston: This was done in 81c572a

proposals/4230-animated-image-flag.md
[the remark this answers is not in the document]
turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4230/commits/fbb7e60fd466c83d6c0ad4890040a2d2aee92da4

turt2live:
I've made the executive decision to resolve the only outstanding thread.

@mscbot resolve Checklist incomplete

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2328

turt2live:
Merged 🎉

PR #4233, opened 2024-11-28
MSC4233: Remembering which server a user knocked through

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/knocking-server/proposals/4233-remembering-knock-via.md)

Author: @turt2live, @ara4n 

----

### DIFF
--- a/proposals/4233-remembering-knock-via.md
+++ b/proposals/4233-remembering-knock-via.md
@@ -1,4 +1,4 @@
-# MSC4233: Remembering which server a user knock through
+# MSC4233: Remembering which server a user knocked through
 
 [Knocking](https://spec.matrix.org/v1.12/client-server-api/#knocking-on-rooms) allows users to request
 an invite into a semi-public room from members already in that room. The use cases for this feature
--- a/proposals/4233-remembering-knock-via.md
+++ b/proposals/4233-remembering-knock-via.md
@@ -1,0 +1,113 @@
+# MSC4233: Remembering which server a user knock through
+
+[Knocking](https://spec.matrix.org/v1.12/client-server-api/#knocking-on-rooms) allows users to request
+an invite into a semi-public room from members already in that room. The use cases for this feature
+vary, though it is relatively common that the room isn't already known to the user's homeserver. For
+this reason, the [API endpoint for knocking](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3knockroomidoralias)
+accepts `via` parameters to assist the homeserver in locating another server to knock through.
+
+Later, when the client wishes to show the user's pending knocks, it may wish to show the user more
+information, or possibly even detect that the room became public and offer a direct join button. It
+may do this through an API like the summary endpoint proposed in [MSC3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266).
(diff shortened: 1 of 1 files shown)


proposals/4233-remembering-knock-via.md
turt2live, line 1: Implementation requirements:

* [ ] Client(?)
* [ ] Server

proposals/4233-remembering-knock-via.md
```diff
@@ -0,0 +1,113 @@
...
+{
+  "rooms": {
+    "knock": {
+      "!opaque:example.org": {
+        "knock_state": {
+          "events": [
+            {"type": "m.room.name", "state_key": "", "content": {"name": "My Room"}}
+          ]
+        },
+        "knock_servers": [ // new property
```
ara4n, line 42: i wonder if `via` might be clearer?

proposals/4233-remembering-knock-via.md
```diff
@@ -0,0 +1,113 @@
...
+Simplified sliding sync is currently described as [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
+and is set to overhaul the sync model for clients. This proposal doesn't change MSC4186, but does
+suggest that the `knock_servers` field be similarly kept next to `knock_state` on a room subscription.
+
+## Potential issues
+
+Servers may not have already tracked this, so information will be lacking for already-knocked rooms.
+Clients should expect errors, per usual, when attempting to call the summary API or any other endpoint.
+
+## Alternatives
```
ara4n, line 83: presumably another option would be for the server to silently remember the knock server on behalf of the client and add an implicit via when they provide a raw room id?

proposals/4233-remembering-knock-via.md
turt2live, line 1: This MSC also applies to rescinding knocks: servers [don't know](https://github.com/element-hq/synapse/blob/d80cd57c54427687afcb48740d99219c88a0fff1/synapse/handlers/room_member.py#L2022) what server to `make/send_leave` through remotely.

anoadragon453: This proposal currently doesn't help in the case of the user attempting to rescind a knock (see https://github.com/element-hq/synapse/issues/18030 for some discussion) as [`POST /_matrix/client/v3/rooms/{roomId}/leave`](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3roomsroomidleave) does not have a `via` parameter. It did not need one originally, as before knocking it was assumed the user had to be in the room already in order to `/leave`.

To support rescinding knocks over federation, either this proposal adds a `via` parameter to `/leave` as well, or we just make adding remotes entirely server-side as https://github.com/matrix-org/matrix-spec-proposals/pull/4233/files#r1862610665 describes.

proposals/4233-remembering-knock-via.md
turt2live: The MSC already requires the server to track it, so I guess it's an open question for whether we want to continue using `via` for things, or have servers remember stuff.

proposals/4233-remembering-knock-via.md
turt2live: `/leave` is also used to decline invites, which is also subject to the same sort of issues (if the sender's server is down). 

Overall, if we do want to add `via` to `/leave`, I'd ask that it be a dedicated MSC. If we want to move away from client-provided `via`s, that may also be a different MSC.

PR #4235, opened 2024-12-02
`server_name` query param for hierarchy endpoint

ajbura:


proposals/4235-via-query-param-for-hierarchy.md
```diff
@@ -0,0 +1,35 @@
+# MSC4235: `via` query param for /hierarchy endpoint
+
+To request the hierarchy of a space, clients need to call the
+[/hierarchy](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv1roomsroomidhierarchy) endpoint. This
+endpoint does not require user to join the room before responding with hierarchy details. It instead check whether the
+room is potentially joinable (per [MSC3173](https://github.com/matrix-org/matrix-spec-proposals/pull/3173)) before
```
clokep, line 6: Please link to the spec for merged MSCs.

proposals/4235-via-query-param-for-hierarchy.md
```diff
@@ -0,0 +1,35 @@
...
+responding.
+
+Because it doesn't not require space room to be joined before calling this endpoint, It is possible that the requested
+room is not locally available on the server and server need to request over federation to respond. In that case this
+endpoint does not provide client to express the server name's which server need to request over federation.
```
clokep, line 11: I think there's a couple typos that are making this hard to understand:

* "Doesn't not" -> "does not"
* "and server need to" -> "the server may need to"

proposals/4235-via-query-param-for-hierarchy.md
```diff
@@ -0,0 +1,35 @@
...
+
+The [/hierarchy](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv1roomsroomidhierarchy) endpoint should
+include `via` as query parameter which server can use in case requested room in not locally available on the server.
+
+Following change will allow clients to use this endpoint in more efficient manner, like: 
+1. Requesting the only part of sub-spaces hierarchy which is unknown the the client.
+2. Rendering the preview of space and it's children before joining the space.
+3. Since server does not respect [order](https://spec.matrix.org/v1.9/client-server-api/#mspacechild) of the children in
+   response, client can divide the work of loading child spaces and figuring out order and re-ordering instead of
+   relying on the long paginated response for those action.
```
clokep, line 23: Doesn't the hierarchy endpoint return them in order already?

ajbura: synapse does return them in order but conduit and conduwuit does not. Regardless I removed the wording as it was a server specific issue.

clokep: I think that's a bug in the implementations then.

proposals/4235-via-query-param-for-hierarchy.md
```diff
@@ -0,0 +1,36 @@
+# MSC4235: `via` query param for /hierarchy endpoint
+
+To request the hierarchy of a space, clients need to call the
+[/hierarchy](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv1roomsroomidhierarchy) endpoint introduced
+in [MSC2946](https://github.com/matrix-org/matrix-spec-proposals/pull/2946). This endpoint does not require user to join
```
clokep, line 5: We should link to the spec if possible?

proposals/4235-via-query-param-for-hierarchy.md
```diff
@@ -0,0 +1,36 @@
...
+the room before responding with hierarchy details. It instead check whether the room is potentially joinable (per
+[MSC3173](https://github.com/matrix-org/matrix-spec-proposals/pull/3173)) before responding.
+
+Because it does not require a space room to be joined before calling this endpoint, it's possible that the requested
```
clokep, line 9: How would this happen? From a room directory response? I'm unsure what user interactions would cause this.

proposals/4235-via-query-param-for-hierarchy.md
ajbura: it already has link to the hierarchy spec or I am missing something?

proposals/4235-via-query-param-for-hierarchy.md
ajbura: User interactions are implementation specific, this MSC is trying to fix the issue in the hierarchy spec. An example using the via param would be requesting the only part of sub-spaces hierarchy which is unknown the the client. I have listed some more example of how this could be helpful in the MSC itself.

proposals/4235-via-query-param-for-hierarchy.md
clokep: There's no reason to link to the MSC then, it's confusing.

proposals/4235-via-query-param-for-hierarchy.md
clokep: But it is useful to understand the expected sets of APIs that will be called.

ajbura: I have added the client usage details https://github.com/ajbura/matrix-spec-proposals/blob/server-name-hierarchy/proposals/4235-via-query-param-for-hierarchy.md#rendering-the-space-hierarchy-before-joining and https://github.com/ajbura/matrix-spec-proposals/blob/server-name-hierarchy/proposals/4235-via-query-param-for-hierarchy.md#rendering-subspaces-hierarchy-independently

Let me know if that clears the concern?

clokep:
Overall this looks pretty good! Thanks for writing it up -- do you know if there are any client implementations? (Mostly I'm interested in a client that has the logic for when it sends the vias?)

ajbura:
For client side I am planing to implement this in Cinny once it lands in one of the server implementation.

PR #4237, opened 2024-12-06
Distinguish room state and timeline events in MSC2762

robintown:
Currently, MSC2762 is ambiguous as to whether it gives widgets a way to actually listen to the state of a room. The `read_events` action is said to only return zero or one events if you include a `state_key`, which suggests that the `m.receive.state_event` capability is meant to expose clients to *room state entries* rather than simply, state events that show up in the timeline. However, there seems to be no real reason to withhold timeline events from the widget just because they happen to have a `state_key`—state events may still have semantics in certain applications even if they don't belong to the resolved room state.

The proposal here is to more clearly distinguish room state updates from incoming timeline events by:
1. Saying that `read_events` always returns timeline events, rather than performing room state lookups
2. Introducing an `update_state` to proactively push all room state that the widget is interested in, and communicate any further updates as they occur

The idea is to make the widget API reflect the design of Simplified Sliding Sync and MSC4222 by sending the same state events twice: once to say "this is an event in the timeline" and again to say "this is a state update". In clients this has given us the means to finally track room state reliably, to guide implementations toward the "pit of success", and hopefully it'll have the same effect for widgets too.

Implementations:
- https://github.com/matrix-org/matrix-widget-api/pull/116
- https://github.com/element-hq/element-web/pull/28681
- https://github.com/matrix-org/matrix-js-sdk/pull/4574

### DIFF
--- a/proposals/2762-widget-event-receiving.md
+++ b/proposals/2762-widget-event-receiving.md
@@ -224,7 +224,7 @@ to receive events from. The exact details of these permissions are covered later
 
 When a widget is approved to receive some state events, the client begins syncing all room state
 entries matching the capabilities in rooms where the widget is permitted to receive events. It
-communicates each room state entry by sending a `toWidget` request with action `update_state`.
+communicates the current state by sending a `toWidget` request with action `update_state`.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


proposals/2762-widget-event-receiving.md
```diff
@@ -220,6 +220,52 @@ unless they asked for it (and were approved), and they receive *decrypted* event
...
+
+The `data` is the state event representing the current value of the room state entry for
+(`room_id`, `type`, `state_key`). The widget acknowledges receipt of this request with an empty
+`response` object.
+
+Once the client has finished sending `update_state` actions for all currently approved room state
+entries, it sends a `toWidget` request with action `room_state_synced` and an empty `data` object.
+This informs the widget that any room state entries for which no `update_state` action was sent *do
+not exist*, and the widget acknowledges the request with an empty `response` object. The client
+continues sending `update_state` actions whenever it observes a change in the relevant room state.
```
toger5, line 260: Will it then also send another `room_state_synced` action?
I am not sure I like the concept of room state synced. I think it does not match matrix. It gives the false impression that room state can be in sync but this is nothing we can prove with the matrix protocol.
For example there is no difference for the widget between:
 - all state has been synced and then a ms later new state is received by the server (that is already 2min old) and gets send to the widget
 - not all state is synced, one event is missing (2mins old) and it is synced in the next ms.
 
 Both cases are equivalent but one gives you the state synced before the 2min old event and the other one after.
 A widget using state needs to be able to deal with all possible state changes and states. Since it is required to be reactive to all scenarios one can argue that there is no special logic it can do when all state is loaded.

turt2live:
@robintown it's not clear to me if this is meant to be an MSC, or is a way to allow the team to review some changes before including in the proposal.

robintown:
It's the latter, yes

proposals/2762-widget-event-receiving.md
robintown: Updated this according to our discussion: `update_state` now sends an array so that `room_state_synced` is unnecessary.

proposals/2762-widget-event-receiving.md
```diff
@@ -220,6 +220,54 @@ unless they asked for it (and were approved), and they receive *decrypted* event
...
+        }
+      }
+    ]
+  }
+}
+```
+
+`data.state` is an array of state events representing the current values of the room state for each
+(`room_id`, `type`, `state_key`) tuple. The widget acknowledges receipt of this request with an
+empty `response` object.
```
toger5, line 258: Reading through this again for some cases we might need a pagination api.
I think postmessage is only limited by system memory so it does not really need pagination. But for rooms like matrix HQ the client itself does not hold all the room members.

I am not sure if we even want the widget to be able to force the client to download all the member events with this API.
But maybe we should allow the client to inform the widget that it limited the amount of results.
If the widget receives a state update we might want include a list of keys that got capped.
This would allow the widget to know which keys it does not know entirely and where it might want to do a `read_event` request.

Overall since we do not have a good concept for pagination for messages (state and timeline) in general I don't see this as blocking.
It is not a breaking change that the widget message horizon can be limited by the client. this is already the current behaviour.

toger5 (approved):
This makes sense and is very simple.

proposals/2762-widget-event-receiving.md
robintown: Thanks for calling this out again. As discussed I do want to leave advanced pagination concepts for another time.

PR #4238, opened 2024-12-10
MSC: Pinned events read marker

ajbura:


proposals/4238-pinned-events-read-marker.md
```diff
@@ -0,0 +1,30 @@
+# MSC4238: Pinned events read marker
+
+Currently, messages are pinned using the `m.room.pinned_events` state event (see [specification](https://spec.matrix.org/v1.12/client-server-api/#mroompinned_events)). However, when new messages are pinned by the super user, other users can not detect newly pinned messages.
+
+It primarily happen because `m.room.pinned_events` does not have any information that can help users detect the change. Resulting in users having no visual clues about newly pinned message. To cope with it they either have to mentally remember the total number of pinned message or discovery of `m.room.pinned_events` state event in clients timeline reminds them to check for possibly new pinned messages.
```
turt2live, line 5: Please wrap lines to ~100 characters

proposals/4238-pinned-events-read-marker.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4238-pinned-events-read-marker.md
tulir, line 5: Why is `prev_content` not sufficient?

ajbura: Because `prev_content` doesn't have info regarding the read status of events. Also it contain content of immediate previous whereas new pinned events entry can span multiple events.

proposals/4238-pinned-events-read-marker.md
```diff
@@ -0,0 +1,42 @@
+# MSC4238: Pinned events read marker
+
+Currently, messages are pinned using the `m.room.pinned_events` state event (see
+[specification](https://spec.matrix.org/v1.12/client-server-api/#mroompinned_events)). However, when new messages are
+pinned by the super user, other users can not detect newly pinned messages.
+
+It primarily happen because `m.room.pinned_events` does not have any information that can help users detect the change.
+Resulting in users having no visual clues about newly pinned message. To cope with it they either have to mentally
+remember the total number of pinned message or discovery of `m.room.pinned_events` state event in clients timeline
+reminds them to check for possibly new pinned messages.
```
KitsuneRal, line 10: I generally agree with the problem statement but not sure if a solution really needs one more read marker. The existing `m.read` marker applies to message and state events alike so in theory a client could (should?) check whether any (and which, in fact) `m.room.pinned_events` state events are newer than `m.read`. The same problem exists with `m.room.topic`, for example: without being in a room, you cannot tell whether the room topic has changed unless the client marks the room as unread when a new `m.room.topic` event arrives.

When it comes to the conditions to mark the room as unread, these are effectively defined by push rules. If you add a push rule that says to _notify_ (not highlight) on new `m.room.pinned_events` the respective room will end up being marked as updated. I encourage the author of this MSC to explore this direction; adding a push rule is easier for the spec as well as for clients.

PR #4240, opened 2024-12-11
Make MSC3765 independent of extensible events and new room versions

Johennes:
This is to address https://github.com/matrix-org/matrix-spec-proposals/pull/3765#discussion_r1878711894.

(Unfortunately, I cannot push to the branch myself anymore.)


turt2live (approved):
I think this captures the idea - thanks!

PR #4242, opened 2024-12-17
MSC4242: State DAGs

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/placeholder-1/proposals/4242-state-dags.md)

Synapse impl:
 - [CSAPI](https://github.com/element-hq/synapse/pull/19424)
 - [Federation](https://github.com/element-hq/synapse/pull/19425)

Complement tests: https://github.com/matrix-org/complement/pull/841

No client changes required as it's not client visible.

### DIFF
--- a/proposals/4242-state-dags.md
+++ b/proposals/4242-state-dags.md
@@ -180,7 +180,7 @@ ensures that the event never makes it into the room state because bans are appli
   formed by `prev_state_events` is used to calculate `auth_events` for state resolution.
   This targets the historic authorisation property.
 - Room state is derived from the DAG induced by `prev_state_events`, NOT `prev_events`.
-  Events which specify faulty `prev_state_events` are rejected.
+  Events which specify faulty `prev_state_events` (e.g. referencing non-state events or events in different rooms) are rejected.
 - `prev_state_events` are populated with the current forward extremities of the set of state events in the room.
 - Servers **MUST** have all paths from any state event to the `m.room.create` event. For any
   event received over federation via `/send`, the sending server **MUST** be able to provide the
@@ -296,8 +296,7 @@ The "current state" of the room is given by the resolved state of the state DAG'
 
 https://spec.matrix.org/v1.12/server-server-api/#soft-failure
 
-This proposal removes the clause about not adding the soft-failed event to the server's list of forward extremities
-**for the state DAG only**.
+Events in the state DAG **should be part of the forward extremities of a room** otherwise we run into the problems described below.
 
 State DAGs exist to ensure that all servers rapidly converge on the same set of state events (eventual state event delivery) so they can
 calculate the same current state in the room (room state convergence). Soft failure makes this harder and in certain cases can obstruct
@@ -346,7 +345,7 @@ eventually converge. This will be addressed in a future MSC.
 
 ##### `/get_missing_events`
 
-An additional top-level key `state_dag` is added. If `true`, the server walks back via
+An additional top-level key `state_dag` is added. If `true`, the receiving server walks back via
 `prev_state_events` and not `prev_events`. Default: `false`.
 
 Example:
@@ -363,7 +362,7 @@ Example:
 }
 ```
 
-If the `limit` is reached, the server should continue to walk back until it has discovered *all*
+If the `limit` is reached, the sending server should continue to walk back until it has discovered *all*
 `prev_state_events`. Care must be taken to ensure *all* paths to the `m.room.create` event are walked. Consider:
 
 ```mermaid
@@ -376,7 +375,7 @@ flowchart BT
     F(Latest event) --> E
 ```
 If this endpoint only returns E, then there is a link between the latest event and the last known
-event, but server implementations MUST NOT stop walking the graph as it is not a complete graph since
+event, but sending server implementations MUST NOT stop walking the graph as it is not a complete graph since
 C is still missing. This may result in additional `/get_missing_events` calls. This MSC recommends
 that servers exponentially increase their `limit` until they have seen all events, e.g. `?limit=10`
 then 20, 40, 80... This is a simple approach and there are better alternatives which can be explored
@@ -414,6 +413,8 @@ Walking with `latest_events={D,E}` returns `[B,C]`. Algorithmically, the server 
  - the result list must not be reversed at the end, unlike non-state DAG `/get_missing_events` responses.
 
 ##### `/send_join`
+
+For room versions which implement state DAGs:
 - Do not return `auth_chain`, `state`, `servers_in_room` by default. Instead return `state_dag` which contains the entire state DAG.
 - Verify that you have ALL paths from the `prev_state_events` of the join event to the create event for the room. If you don't, then you did not join the room successfully and the server is faulty. Join via a different server.
 - Faster room joins: if `?omit_members=` is set, include `partial_state_event_ids`. See "Faster remote room joins" for more information.
@@ -436,10 +437,11 @@ The intended server behaviour would be:
 
 1. If type is `m.room.create`:
     1. If it has any `prev_events`, reject.
-    2. **[Changed in this version]** If it has any `prev_state_events`, reject.
-    3. If the domain of the `room_id` does not match the domain of the `sender`, reject.
+    2. If the event has a `room_id`, reject.
+    3. **[Added in this version]** If it has any `prev_state_events`, reject.=
     4. If `content.room_version` is present and is not a recognised version, reject.
-    5. Otherwise, allow.
+    5. If `additional_creators` is present in `content` and is not an array of strings where each string passes the same user ID validation applied to `sender`, reject.
+    6. Otherwise, allow.
 2. **[Added in this version]** Considering the event's `prev_state_events`:
     1. If there are entries which do not belong in the same room, reject.
     2. If there are entries which do not have a `state_key`, reject.
@@ -448,7 +450,7 @@ The intended server behaviour would be:
     1. **[Removed in this version]** If there are duplicate entries for a given `type` and `state_key` pair, reject.
     2. **[Removed in this version]** If there are entries whose `type` and `state_key` don’t match those specified by the auth events selection algorithm described in the server specification, reject.
     3. If there are entries which were themselves rejected under the checks performed on receipt of a PDU, reject.
-    4. **[Removed in this version]** If there is no `m.room.create event` among the entries, reject.
+    4. **[Removed in this version]** If any event in `auth_events` has a `room_id` which does not match that of the event being authorised, reject.
   
  ##### Rationale
 
@@ -472,6 +474,7 @@ but provides very very weak guarantees that at the time the join was issued, tho
 in the room.
 
 State DAG rooms use a different approach. If `?omit_members=true`:
+ - The `/send_join` response still includes _the entire state DAG._ The same amount of data is transferred with and without faster room joins.
  - The `/send_join` response includes a new array: `timeline` which represents the most recent events in the room,
    suitable for presenting in the timeline. How many events is configurable via a new query param `?timeline_limit=` which
    is a non-zero integer, with a hard cap of 64. This is subject to history visibility rules.
@@ -533,7 +536,7 @@ The joining server should then perform the following:
 
 This technique makes several tradeoffs:
  - It sends more response data than faster remote room joins does today, because it is including the entire state DAG in the response. However, the cost of this
-   is reasonably small compared to the cost of verifying the resulting events. By doing this, we guarantee that rooms will transition from being partially joined
+   is reasonably small compared to the cost of verifying the resulting events and storing state snapshots. By doing this, we guarantee that rooms will transition from being partially joined
    to "joined" or "failure" in finite time. If fetching the state DAG was deferred to another round trip, the partially joined room may _never_ transition in the
    case where an attacker has control over the network. This is stronger than what [MSC3902](https://github.com/matrix-org/matrix-spec-proposals/pull/3902) currently
    proposes.
@@ -648,7 +651,8 @@ Non-auth events:
  uk.half-shot.bridge       |      4
  ```
 
-The reason for this surprising result is because the size of the state DAG does NOT scale with the number of members
+The reason for this surprising result is because the amount of _extra events_ included by including
+non-auth state does NOT scale with the number of members
 in the room. This makes intuitive sense, as the rate of topic changes shouldn't change depending on
 if there are 100 or 100,000 members in the room.
 
--- a/proposals/4242-state-dags.md
+++ b/proposals/4242-state-dags.md
@@ -303,7 +303,7 @@ State DAGs exist to ensure that all servers rapidly converge on the same set of 
 calculate the same current state in the room (room state convergence). Soft failure makes this harder and in certain cases can obstruct
 servers from agreeing on the same current state. Consider this scenario (green nodes are the current state at `$vsOy`):
 
-<img width="687" alt="Screenshot 2024-12-19 at 08 30 59" src="https://github.com/user-attachments/assets/43acb34c-497b-4081-aef6-113f5fcd7093" />
+<img width="687" height="300" alt="sf1" src="https://github.com/user-attachments/assets/d4135ffa-b5c8-431e-8314-4a58dd3372d9" />
 
  - there are 3 concurrent events:
    * A) PL100 bans PL75
@@ -321,7 +321,7 @@ servers from agreeing on the same current state. Consider this scenario (green n
    unimportant events never referenced it.
  - Thus, the long dead server does not know the join rules are "invite", _despite it being part of the current room state_.
 
-<img width="626" alt="Screenshot 2024-12-19 at 08 36 20" src="https://github.com/user-attachments/assets/23100dda-7c1c-4664-99de-9d42e8c48bc7" />
+<img width="626" height="279" alt="sf2" src="https://github.com/user-attachments/assets/7ff1effa-9797-46ba-bbe5-b4aa01a04fc0" />
 
 So it is worth considering what value soft-failure has _for state DAGs_. Soft-failure is very useful for filtering
 out messages which would otherwise be delivered to clients, because the stream (arrival) order doesn't care about
--- a/proposals/4242-state-dags.md
+++ b/proposals/4242-state-dags.md
@@ -782,12 +782,15 @@ checks, as not all servers are guaranteed to converge due to reliance on `/state
 
 ## Unstable prefix
 
-- Room version is `org.matrix.msc4242.11`.
+- Room version is `org.matrix.msc4242.12`, based on room version 12.
 - The `/get_missing_events` flag is `org.matrix.msc4242.state_dag`.
 - There is no need to prefix the `/send_join` response key for the state DAG as it's already namespaced via the room version supplied in `/make_join`.
 - There is no need to prefix `prev_state_events` as it's already namespaced via the room version.
 
 ## Dependencies
 
-This MSC has no dependencies.
+This MSC has no dependencies, but would benefit greatly from an MSC which makes events self-verifying,
+such as [MSC4014](https://github.com/matrix-org/matrix-spec-proposals/pull/4014),
+[MSC4243](https://github.com/matrix-org/matrix-spec-proposals/pull/4243),
+[MSC4345](https://github.com/matrix-org/matrix-spec-proposals/pull/4345).
 
--- a/proposals/4242-state-dags.md
+++ b/proposals/4242-state-dags.md
@@ -1,0 +1,792 @@
+# MSC4242: State DAGs
+
+Room state is important: who is joined to a room, who is an admin in the room, and what are the join rules
+(invite/public/restricted) are fundamental properties of a room. The current mechanism to ensure
+eventual consistency over federation for room state is suboptimal and can cause state to
+converge too slowly, or not at all. This MSC proposes a faster-to-converge and more secure
+alternative in the form of state DAGs. A state DAG is only for the state
+events in the room, where edges are defined using `prev_state_events` instead of `prev_events`.
+
+State DAGs fix several vulnerabilities in state resolution. These vulnerabilities are not with the
+algorithm, but with the data provided as inputs to the algorithm: a
+case of garbage in, garbage out. Specifically, the State DAG proposal of this MSC:
+ - _calculates_ the `auth_events` key which is today sender-specified in every event,
+ - _mandates_ that all servers fill in the entire state DAG on receipt of new events with unknown
+   `prev_state_events` by walking back up the DAG to the create event via the `/get_missing_events` endpoint.
+   This obsoletes the `/state`, `/state_ids` and `/event_auth` endpoints.
+ - _calculates_ room state based on `prev_state_events`. This unifies step 4 and step 5 of the PDU
+   checks as the calculated room state is used to calculate the `auth_events`.
+
+## Problem properties: eventually consistent authorisation
+To decide on the authorisation of incoming events, servers in a room need to replicate authorisation
+events (*auth events*), which make up the part of the room state necessary for the authorisation decision
+(*auth state*). Auth events are state events, and defined as all events of types relevant to write access
+to a room, namely `{ m.room.create, m.room.power_levels, m.room.member, m.room.join_rules, m.room.third_party_invite }`.
+Other state events that are relevant for authorisation or security are events of the types
+`{ m.room.history_visibility, m.room.guest_access, m.room.encryption, m.room.server_acl }`. It is critical that these event
+types are synchronised rapidly among all honest servers as they control the fundamental security properties
+in the room.
+
+All room events are partially ordered by hash links to previous events, i.e. the room DAG formed by
+`prev_events`. In analogy with the room DAG, this MSC proposes to add additional hash links to form a
+partial order on state events only, i.e. a state DAG formed by `prev_state_events`.
+
+This MSC aims to improve the reliability of state event replication, as well as the security of
+authorization decisions. In line with scientific research on
+[Conflict-Free Replicated Data Types](https://dl.acm.org/doi/full/10.1145/3695249), a solution needs
+to provide the following properties:
+- Eventual state event delivery: If a state event is added to the room DAG by an honest server, it is eventually added on all honest servers.
+- Room state convergence: Two honest servers that have the same set of state events in their room DAG derive the same room state.
+
+This MSC aims to improve the security of authorization decisions. In line with scientific research on
+[To the Best of Knowledge and Belief: On Eventually Consistent Access Control](https://publikationen.bibliothek.kit.edu/1000176494),
+a solution needs to provide the following properties:
+- Historic authorisation: An event stored by an honest server as part of its room DAG is authorised
+  by the room state resolved from all events in the state DAG before the given event.
+- Concurrent authorisation: If an event affects room state of an honest server, that event is
+  authorised by the room state resolved from all events in the state DAG before *or concurrent to*
+  the given event.
+
+The eventual delivery and convergence properties in conjunction are said to make up the eventual
+consistency guarantee, which is one of the strongest guarantees we can have while staying available
+under partition, and what we want for authorisation in particular. The historic and concurrent
+authorisation properties on the other hand describe *how* we want consistent authorisation to look like.
+Room state convergence and concurrent authorisation are topic of the state resolution algorithm. As
+this MSC assumes that state resolution is correct, these properties are assumed to be fulfilled, and
+the focus is on eventual state event delivery and historic authorisation.
+
+## Background: Authorisation rules today
+
+When a server receives an event from another server, it needs to authorise that event. The
+[checks performed on receipt of a PDU](https://spec.matrix.org/v1.10/server-server-api/#checks-performed-on-receipt-of-a-pdu)
+outline the steps involved. 
+
+The area this MSC concerns itself with is Steps 4 and 5:
+
+> 4) Passes authorization rules based on the event’s auth events, otherwise it is rejected.
+> 5) Passes authorization rules based on the state before the event, otherwise it is rejected.
+
+The `auth_events` are a set of events which "authorise" the event. The idea is that this set of
+events includes all events which "prove" that this event is allowed. For example, to send a message,
+the message must have `auth_events` proving:
+ - The room in question (`m.room.create` event)
+ - The user was joined to the room (`m.room.member` event)
+ - The user was authorised to send into the room (`m.room.power_levels` event)
+
+These authorisation events must also themselves be authorised. For example, to join an
+invite-only room, your `m.room.member` join event must also reference the `m.room.member` invite
+event. This forms an "auth chain" as any given event references auth events which reference auth
+events which reference auth events, all the way to the `m.room.create` event.
+
+```mermaid
+flowchart BT
+    JOIN("m.room.member => join") --> CREATE("m.room.create")
+    PL("m.room.power_levels") --> JOIN
+    MSG("m.room.message") --> PL
+    MSG -.-> PL & CREATE & JOIN
+```
+*A room DAG showing the `auth_events` (dashed arrows) for a `m.room.message`.
+The `prev_events` path is shown with solid arrows.*
+
+```mermaid
+flowchart BT
+    JOIN("m.room.member => Alice join") --> CREATE("m.room.create")
+    PL("m.room.power_levels") --> JOIN
+    JR("m.room.join_rules") --> PL
+    MSG("m.room.message") --> JR
+    INV("m.room.member => Alice invite Bob") --> MSG
+    BOB("m.room.member => Bob joins") --> INV
+    BOB -.-> INV & JR & CREATE & PL
+    INV -.-> JR & CREATE & PL & JOIN
+
+    linkStyle 6 stroke:#2962FF,fill:none
+    linkStyle 7 stroke:#2962FF,fill:none
+    linkStyle 8 stroke:#2962FF
+    linkStyle 9 stroke:#2962FF,fill:none
+    linkStyle 10 stroke:#D50000,fill:none
+    linkStyle 11 stroke:#D50000,fill:none
+    linkStyle 12 stroke:#D50000,fill:none
+    linkStyle 13 stroke:#D50000,fill:none
+```
+*A join event with its `auth_events` in blue, along with the invites `auth_events` in red, forming a "chain".*
+
+However, this check alone is not enough. Despite the intention that they "prove" the event is
+authorised, i.e. keeping up the historic authorisation property, they do no such thing. They are just
+a "sanity check" that _at some point_ this event _could_ be authorised.
+The best illustration of this is to consider a banned user. At one point in
+time, the banned user was joined to the room. Nothing stops the banned user from sending messages
+into the room, referencing their _old_ join event prior to the ban. This would pass **this step** of
+the PDU checks.
+
+```mermaid
+flowchart BT
+    JOIN("m.room.member => Alice join") --> CREATE("m.room.create")
+    PL("m.room.power_levels") --> JOIN
+    JR("m.room.join_rules") --> PL
+    MSG("m.room.message Alice") --> JR
+    INV("m.room.member => Alice invite Bob") --> MSG
+    BOB("m.room.member => Bob joins") --> INV
+    MSG2("m.room.message Bob") --> BOB
+    BAN("m.room.member => Bob is banned") --> MSG2
+    MSG3("m.room.message Bob whilst banned") --> BAN
+    MSG3 -.-> CREATE & BOB & PL
+```
+*Bob attempting to evade a ban by referencing in his `auth_events` his join event from earlier.*
+
+As a result, PDUs need additional checks, including Step 5, which verifies the historic authorisation property:
+_at the state before the event was sent_, the user was allowed to do this action. For non-faulty
+servers, the set of events relating to authorisation will be the same in both `auth_events` and the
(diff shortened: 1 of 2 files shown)


turt2live:
(for clarity: this is using the process defined by https://github.com/matrix-org/matrix-spec/pull/2157)

proposals/4242-state-dags.md
turt2live, line 1: Implementation requirements:

* [ ] Server (unstable room version)
* [ ] Complement tests (for unstable room version)

turt2live:
This MSC now has content.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+ m.room.canonical_alias    |      1
+ m.room.guest_access       |      7
+ m.room.history_visibility |      3
+ m.room.name               |      1
+ m.room.related_groups     |      1
+ m.room.topic              |      8
+ uk.half-shot.bridge       |      4
+ ```
+
+The reason for this surprising result is because the size of the state DAG does NOT scale with the number of members
```
MatMaul: I don't understand that, since up to my knowledge, `m.room.member` IS a state event ?

Kegan answer:
> this is comparing the amount of state events we currently have to fetch, with the potential extra events a state DAG would include. the amount extra does not scale with the number of members in the room.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+in the room.
+
+State DAG rooms use a different approach. If `?omit_members=true`:
+ - The `/send_join` response includes a new array: `timeline` which represents the most recent events in the room,
+   suitable for presenting in the timeline. How many events is configurable via a new query param `?timeline_limit=` which
+   is a non-zero integer, with a hard cap of 64. This is subject to history visibility rules.
+ - The `/send_join` response includes a new field: `partial_state_event_ids` which is a
+   list of event IDs which contain what the join-helper server claims to be the current state, omitting all members _except_
+   the `sender`s of the `timeline` events and non-member state events e.g the `sender` of the `m.room.create` event.
+   This is sent alongside the entire state DAG. For clarity, the join event induced by `/send_join` is also omitted from
```
MatMaul, line 500: So we still receive the membership events since they are state events and will be present in the full state DAG. Is it on purpose ? Why is it still faster then ?

Kegan answer:
> joining rooms is slow because you need to A) download a lot of events, B) verify those events, C) store state snapshots. Faster room joins today makes all these things faster by reducing the amount of events you download initially. State DAGs  with faster room joins takes a different approach and only optimises B and C, it still downloads the same amount of data.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+#### Soft Failure
+
+> When an event is “soft failed” it should not be relayed to the client **nor be referenced by new events** created
+> by the homeserver (i.e. they **should not be added to the server’s list of forward extremities of the room**).
+> Soft failed events are otherwise handled as usual.
+
+https://spec.matrix.org/v1.12/server-server-api/#soft-failure
+
+This proposal removes the clause about not adding the soft-failed event to the server's list of forward extremities
+**for the state DAG only**.
```
MadLittleMods: This phrasing is confusing. 

During my first reading of the *Soft Failure* section, I understood the problem below but then questioned, "are we doing anything about this?". Re-reading the section, I see this sentence but then because of the way it's written, took multiple tries to grok.

To rephrase for my own reference: Events in the state DAG should be part of the forward extremities of a room otherwise we run into the problems described below.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+  "latest_events": [
+    "$event_received_via_send"
+  ],
+  "limit": 10,
+  "state_dag": true
+}
+```
+
+If the `limit` is reached, the server should continue to walk back until it has discovered *all*
+`prev_state_events`. Care must be taken to ensure *all* paths to the `m.room.create` event are walked. Consider:
```
MadLittleMods: "server" is ambiguous

Based on the further paragraphs, this seems to be talking about the homeserver making the request to ask about missing events.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+  \ / |
+   D  E
+```
+Walking with `latest_events={D,E}` returns `[B,C]`. Algorithmically, the server should:
+ - map `earliest_events` to a set of 'seen' event IDs.
+ - sort the provided `latest_events` and put them into a queue, removing any in 'seen'.
+ - while the queue is not empty and the limit has not been reached:
+   * dequeue an event ID and find its `prev_state_events` PSE.
+   * Remove IDs from PSE if they are in 'seen'.
+   * add the remaining event IDs from PSE to the result list and the end of the queue.
```
MadLittleMods: I'm assuming the list of PSE should be sorted before put onto the end of the queue?

(needs clarification)

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+ - do state resolution on A to get the auth state S before the event E,
+ - find the relevant auth events in S which would today be put into `auth_events`,
+ - store them in a calculated `auth_events` field.
+
+#### Auth rules changes
+
+1. If type is `m.room.create`:
+    1. If it has any `prev_events`, reject.
+    2. **[Changed in this version]** If it has any `prev_state_events`, reject.
+    3. If the domain of the `room_id` does not match the domain of the `sender`, reject.
```
MadLittleMods: (no longer applies in [room version 12](https://spec.matrix.org/v1.16/rooms/v12/) which I assume we're building from)

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+    5. Otherwise, allow.
+2. **[Added in this version]** Considering the event's `prev_state_events`:
+    1. If there are entries which do not belong in the same room, reject.
+    2. If there are entries which do not have a `state_key`, reject.
+    3. If there are entries which were themselves rejected under the checks performed on receipt of a PDU, reject.
+4. **[Changed in this version]** Considering the event’s calculated `auth_events`:
+    1. **[Removed in this version]** If there are duplicate entries for a given `type` and `state_key` pair, reject.
+    2. **[Removed in this version]** If there are entries whose `type` and `state_key` don’t match those specified by the auth events selection algorithm described in the server specification, reject.
+    3. If there are entries which were themselves rejected under the checks performed on receipt of a PDU, reject.
+    4. **[Removed in this version]** If there is no `m.room.create event` among the entries, reject.
```
MadLittleMods: The auth rules in [room version 12](https://spec.matrix.org/v1.16/rooms/v12/) (which I assume we're building from) are a bit different than what's laid out here.

Regardless, I'm confused why the `m.room.create` requirement is removed? Seems to be the case even for room version 12. From what I'm reading in the MSC, it seems important to be able to trace everything back to the `m.room.create` event.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+   This reduces the amount of events that need to be checked on the critical path. This is particularly effective on member events as members may be on different
+   servers which will take time to fetch server keys for.
+
+From an implementation perspective, servers which implement [MSC3902](https://github.com/matrix-org/matrix-spec-proposals/pull/3902) should need minimal changes to
+adapt their system to work in state DAG rooms because the underlying data model is the same. Care must be taken to ensure that if the state DAG verification is
+interrupted e.g due to restarting the server, that it is resumed on startup.
+
+#### Consequences of having a full state DAG
+
+`depth` can be calculated and hence trusted for events that are included in the state DAG. This has
```
MadLittleMods, line 569: I assume this is only `depth` (topological order) of the state event in the state DAG?

Separate from `depth` used for all events today.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+ensures that the event never makes it into the room state because bans are applied first.
+
+## Proposal
+
+### High level summary of changes
+- `auth_events` are removed from all events, and replaced with `prev_state_events`. The state DAG
+  formed by `prev_state_events` is used to calculate `auth_events` for state resolution.
+  This targets the historic authorisation property.
+- Room state is derived from the DAG induced by `prev_state_events`, NOT `prev_events`.
+  Events which specify faulty `prev_state_events` are rejected.
```
MadLittleMods: How can a server detect faulty `prev_state_events`? What's faulty?

Is this referring to passing the authorization rules?

MadLittleMods: There are faulty Complement test cases in https://github.com/matrix-org/complement/pull/841 (see `tests/msc4242/msc4242_faulty_events_test.go`)

 - Faulity if no `prev_state_events`
 - Faulty if any of the `prev_state_events` do not exist
 - Faulty if referencing non-state events in `prev_state_events`
 - Authorization rules

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,796 @@
...
+ - map `earliest_events` to a set of 'seen' event IDs.
+ - sort the provided `latest_events` and put them into a queue, removing any in 'seen'.
+ - while the queue is not empty and the limit has not been reached:
+   * dequeue an event ID and find its `prev_state_events` PSE.
+   * Remove IDs from PSE if they are in 'seen'.
+   * add the remaining event IDs from PSE to the result list and the end of the queue.
+ - the result list must not be reversed at the end, unlike non-state DAG `/get_missing_events` responses.
+
+##### `/send_join`
+- Do not return `auth_chain`, `state`, `servers_in_room` by default. Instead return `state_dag` which contains the entire state DAG.
```
MadLittleMods, line 427: We should clarify that this new response only applies for the new room versions. I think that's the implication (assumed)

proposals/4242-state-dags.md
kegsay: Rephrased:

> The reason for this surprising result is because the amount of _extra events_ included by including non-auth state does NOT scale with the number of members in the room.

proposals/4242-state-dags.md
kegsay: Attempted to clarify.

proposals/4242-state-dags.md
kegsay: Removed this sentence and replaced with:

> Events in the state DAG **should be part of the forward extremities of a room** otherwise we run into the problems described below.

proposals/4242-state-dags.md
kegsay: Clarified throughout with sending server vs receiving server.

proposals/4242-state-dags.md
kegsay: Doesn't matter, as it's still deterministic if we use the wire-format ordering.

proposals/4242-state-dags.md
kegsay: It was removed from v12 because the room ID _is_ the create event ID, so it doesn't live in `auth_events` anymore.

It was removed from the state DAGs proposal because it's a calculated field, thus will always be present. This is also why we remove duplicate checks, or selecting tuples which aren't auth events.

proposals/4242-state-dags.md
kegsay: Yes, only for state events, not messages. The sentence does say "for events that are included in the state DAG" so I don't think further clarification is needed.

proposals/4242-state-dags.md
kegsay: Clarified with:

> Events which specify faulty `prev_state_events` (e.g. referencing non-state events or events in different rooms) are rejected.

It's in the high-level changes section so not saying more on this here.

proposals/4242-state-dags.md
MadLittleMods: To explain for myself (assuming), because `room_id` is part of the event, we can verify the `m.room.create` event that way.

proposals/4242-state-dags.md
MadLittleMods: I was thinking this was necessary in order to "walk these events in lexicographical order" :thinking: 

If we don't do this then part-way down the forks, we could wander down in whatever path.

kegsay: Aha, you're right, thanks. Will fix.

kegsay: https://github.com/matrix-org/matrix-spec-proposals/commit/d74137107e80d84f0bb0baaf189693d0e2c692c3

kegsay: Reopened this because this ordering is harder to implement efficiently than other orderings. In particular, when working on the CTE SQL for this, expressing "process one frontier element at a time, query its predecessors, sort them lexicographically then append them to results" results in a subtle ordering which _depends on the processing order of parents_. The CTE SQL works on a hop-based system, finding all events 1 hop away, sorting them, rinse and repeat. This loses the distinction of sorting based on parent order, causing some unit tests to fail.

As an example:
```
  A
  |
  B
 / \
D   C
|   |
E   F  <-- latest={E,F}, earliest={}, limit=2
```

The old order would sort latest=[E,F], then pop E, process it to produce D, then pop F and process it to process C, resulting in [D, C] as the output.

The proposed ordering would sort latest=[E,F] then simultaneously lookup all prevs to process [D, C] (everything 1 hop away) then sort them, resulting in [C, D] as the output.

The main important thing here to have a deterministic output, I don't particularly care which ordering we use, and given the CTE one is faster, that one should probably be the one we do.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,800 @@
...
+converge too slowly, or not at all. This MSC proposes a faster-to-converge and more secure
+alternative in the form of state DAGs. A state DAG is only for the state
+events in the room, where edges are defined using `prev_state_events` instead of `prev_events`.
+
+State DAGs fix several vulnerabilities in state resolution. These vulnerabilities are not with the
+algorithm, but with the data provided as inputs to the algorithm: a
+case of garbage in, garbage out. Specifically, the State DAG proposal of this MSC:
+ - _calculates_ the `auth_events` key which is today sender-specified in every event,
+ - _mandates_ that all servers fill in the entire state DAG on receipt of new events with unknown
+   `prev_state_events` by walking back up the DAG to the create event via the `/get_missing_events` endpoint.
```
gamesguru, line 15: This seems insufficient for guaranteeing fully up-to-date reconciliations. While walking `prev_state_events` guarantees full causal history for an event, it fails to uncover concurrent branches of state events (hidden forward extremities generated during network partitions). Relying on the "forward-push" mechanism (new events being sent) to merge branches means partitions will lead to semi-permanent divergence in inactive rooms. Active partition-healing and discovery mechanisms are essential to the eventual consistency this MSC aims for... relegating them to "follow-up" MSCs feels inadequate. I'm sure there has been some work/investigation here, but I have not seen or come across those proposals. I have some ideas here regarding `/room_digest` endpoints and gossip mechanisms.

If state DAGs are to be their own fully contiguous, self-sufficient sub-graphs, they definitely deserve a dedicated `/_matrix/federation/v2/get_missing_state_events` endpoint rather than overloading `/get_missing_events` with a boolean param. Because this MSC introduces completely different traversal logic when walking the state DAG, overloading the legacy API endpoint is surely an anti-pattern. We need a clear, dedicated endpoint for this specific BFS walk.

The proposed exponential limit markers may result in network bottlenecks when resolving deep partitions. Having recently implemented binary-lifted jump pointers (`shorteventid_statejumppointers`) to alleviate database thrashing under the `O(N)` hop-by-hop walk of large test cases, I feel a pressing need for that same `O(log N)` efficiency over the network wire.

On the network side, I caution against leaning too heavily on Bloom Filters. While convenient at shallow depths, their footprint scales linearly to keep a low false-positive rate. A hybrid approach seems better. Similar to how C++ switches from quicksort to heapsort to bound worst-case overhead, federation syncs should enforce an algorithmic "switch" above a critical divergence depth... explicitly shifting to the `O(log N)` techniques already mentioned (Merkle search trees, Merkle-Patricia tries, Range-based reconciliation) to maintain a healthy and efficient network.

gamesguru (changes requested):

proposals/4242-state-dags.md
kegsay: I don't disagree with most of what you have written, but you don't seem to understand that we have a working protocol and large protocol changes are difficult to land.

I explicitly call out that we don't have an anti-entropy mechanism today and consciously defer it. You may disagree, that's fine. It is worth noting that _in practice_ the lack of such a mechanism hasn't materially impacted the protocol because in practice relying on forward-push is enough. Anti-entropy does have its own costs (constant background chatter gossiping latest forward extremities) so it's not as clear cut as you're portraying.

It's a taste issue if you want a new endpoint or not. Both operations walk a DAG and have the same inputs. If you have any non-taste issues which may lean to one or the other then let me know.

Yes, it can take a while to reconcile with a server which has diverged significantly from you. I'm very open to suggestions to improve this, though you need to bear in mind we are working in a Byzantine environment. Most optimisations assume servers follow the protocol and degrade to a DAG walk (or worse, break entirely). If you have a way to ensure the jump pointers can't be outright malicious, please comment with a suggestion.

We're not using bloom filters in this proposal.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,800 @@
...
+The reason for this surprising result is because the amount of _extra events_ included by including
+non-auth state does NOT scale with the number of members
+in the room. This makes intuitive sense, as the rate of topic changes shouldn't change depending on
+if there are 100 or 100,000 members in the room.
+
+There are exceptions however, which will all be affected by this proposal:
+ - `org.matrix.msc3401.call.member` events. MSC3401 uses per-user room state to synchronise call info.
+This _does_ scale with the number of users in the room and would be negatively affected by this
+proposal, as all historical calls will need to be downloaded prior to joining the room.
+ - `m.space.child` events. Space rooms will be more expensive to join, though this proposal would
```
gamesguru, line 660: It may also be worth analyzing or mentioning `m.space.parent`. This introduces some cross-room validation complexity, per the spec:

> To avoid situations where a room falsely claims it is part of a given space, `m.space.parent` events should be ignored unless one of the following is true:
>  - A **corresponding `m.space.child` event can be found in the supposed parent** space.
>  - The **sender of the `m.space.parent` event has sufficient power level in the supposed parent** space to send `m.space.child` state events (there doesn't need to be a matching child event).
>
> https://spec.matrix.org/v1.2/client-server-api/#mspaceparent-relationships

---

As a detail, `m.room.tombstone` seems to have very low overhead (not having any forward extremities or being referenced). But is worth including for completeness.

Lastly, I'd like to bring up `m.policy.rule.*` and `m.room.policy` events. These events are very common in some rooms, and I'm curious if they have the potential to slow things down at all. Seems these events would carry the cost of standard membership events. Rooms today like the KDE help lounge (with over 500K membership churn due to their timeout/rejoin model) can be almost impossible to join and require significant disk/CPU usage if ever successfully joined.

proposals/4242-state-dags.md
gamesguru: 1. I certainly see the value in not attaching too many subordinate MSCs, and how that could slow things down more. However, I insist on proper separations of concerns. If providing "eventual consistency" (preferably: rapid consistency) falls under the purview of a _different_ card or MSC, that should be clearly admitted, explained, and documented. This can be especially misleading given Synapse's tendency _not_ to re-evaluate anything it already processed. Advertising State DAGs intrinsically as a sufficient standalone solution causes some confusion to the serious reader. My MSC4521 aims optimistically to reconcile differences of 100,000 elements over populations of 10 million in about 5-10 seconds. Smaller reconciliations (like under 1000 differences) are nearly instant, under 200 ms—theoretically fast enough to keep all partitions unified in real-time (except for deliberate Byzantine defectors).

2. `/get_missing_state_events` will cache dedicated ETags and consume a homogeneous JSON structure... also allows homeserver admins to route state-DAG syncs to dedicated worker pools or apply custom reverse proxy rules and rate limits. Better to keep it separate.

3. Merkle proofs can attest to the credibility of a path or series of binary hops. While state resolution must still be run (i.e., the corresponding event chain must be fully fetched and validated), a valid Merkle path is instantly more credible, worthy, and promising as a lead than an opaque, ostensibly Byzantine input.

    A "graceful" fallback to linear walking `O(N)` could just as easily be Byzantine. I could self-sign 10 million PDUs and stream them to you one after another, all from the same domain. Linear walks should probably be a worst-case fallback if all else fails.

    To avoid enforcing the full download back to `m.room.create`, we can leverage the finality concepts you recently outlined in your PaPoC '26 ERA paper. By utilizing "epoch events" or an `m.room.finality` seal issued by the finality arbiter (e.g., the immutable Creator role), we can certify or commit a given snapshot, allowing nodes to act as "thin servers" and trust the cryptographic seal.

    This perfectly aligns with the ERA paper's note that finalized events can be compacted. We could establish a precise cadence (e.g., checkpointing every 5000 state events and adjacent merge points), preemptively compress the deterministic JSON with `xzip`, and seal the bundle cryptographically to maintain high availability indefinitely.

    For room versions or customizations not opting into a finality clause, any rebuilds triggered later by backdated admin events remain the server's responsibility to process from the conflict out to the frontier (existing implementations and runtime characteristics can be substantially improved, to the point of allowing arguably even the largest servers to rebuild huge rooms leisurely).

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,800 @@
...
+
+As servers now know the entire state DAG, they can use this to figure out the "state before the event".
+This replaces the endpoints `/state`, `/state_ids` and `/event_auth` which were all previously used to
+fetch missed events. This fixes a security vulnerability where faulty responses to `/state` or `/state_ids`
+could cause state resets. This same _calculated room state_ is used to calculate the `auth_events` for
+incoming events. This unifies steps 4 and steps 5 of the [PDU checks](https://spec.matrix.org/v1.10/server-server-api/#checks-performed-on-receipt-of-a-pdu):
+
+> 1) Is a valid event, otherwise it is dropped. For an event to be valid, it must contain a `room_id`, and it must comply with the event format of that room version.
+> 2) Passes signature checks, otherwise it is dropped.
+> 3) Passes hash checks, otherwise it is redacted before being processed further.
```
gamesguru, line 284: Hi, I am doing a deep dive on redactions in my state res library lately.

Since a new **redaction algo** was not specified or updated, and `auth_events` were gutted out in favor of `prev_state_events`, this would mean redactions strip and break a lot of things. Am I missing something?

We would need to clarify protected fields and hashing semantics.

Technically just overloading the old `auth_events` field name would be an easier refactor, at the expense of confusion/semantic poisoning (and possible resolution failures if the create event and room version can't be fetched)... for a year or two, mere presence of the `prev_state_events` will be a distinguishing factor (e.g., maps numerically to v13+ and state res v2.2+, where there is no v14 nor v2.3 yet).

I think some edge cases about how redacted state events with `content: {}` can influence fork resolution are worth adding, either as manual notes or formally as unit, or rather I guess Complement tests? I did a little work on an internal PR (against my own Complement fork's master/main). I'll invite you there, we can maybe discuss some ideas at leisure. I may be able to put a small demo together.

I have a feeling some of the unintuitive logic around the _same-event-ID despite redaction_ and the `hashes: { ... }` logic/conundrums can be addressed and simplified by choosing the right Merkle tree construction and combining header leaves correctly, so content can be stripped without mutating `event_root`. Or `content_hash` can be checked if it matches the "empty commitment." This sounds like a big lift, but if redactions require a lot of work anyway we may want to do things a more "intuitive" way and refactor while we are at it (I'm not saying I'm married to the idea of Merkle trees/paths/proofs here, but it's certainly worth considering over the coming months so we can get more understanding of the benefits and any possible limits).

Starting a separate review for this, since I have a hunch redactions may deserve a dedicated conversation thread.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+alternative in the form of state DAGs. A state DAG is only for the state
+events in the room, where edges are defined using `prev_state_events` instead of `prev_events`.
+
+State DAGs fix several vulnerabilities in state resolution. These vulnerabilities are not with the
+algorithm, but with the data provided as inputs to the algorithm: a
+case of garbage in, garbage out. Specifically, the State DAG proposal of this MSC:
+ - _calculates_ the `auth_events` key which is today sender-specified in every event,
+ - _mandates_ that all servers fill in the entire state DAG on receipt of new events with unknown
+   `prev_state_events` by walking back up the DAG to the create event via the `/get_missing_events` endpoint.
+   This obsoletes the `/state`, `/state_ids` and `/event_auth` endpoints.
```
MadLittleMods, line 16: What should happen when you call these endpoints with a MSC4242 state DAG room?

Probably want to respond with some error 🤔

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+replaced with _another DAG_, called the "state DAG". This DAG references the "latest"
+(forward extremities) state events. They are linked together via a new field: `prev_state_events`
+which functions in the same way as `prev_events` does today (it's a list of event IDs). All events
+will have this field, even non-state events, in which case they serve to continually broadcast that
+server's current forward extremities for state events.
+
+Any given server can then _calculate_ what the `auth_events` should be for an event by using the
+`prev_state_events`. This _calculated field_ is then used for state resolution.
+
+Servers MUST NOT reference an event in `prev_state_events` which is not in fact a state event,
```
reivilibre, line 207: low importance: Should it also be forbidden to reference an event twice, or to reference two events A and B where B is a state-DAG-descendant of A (`A -> ... -> B`)?

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+
+## Proposal
+
+### High level summary of changes
+- `auth_events` are removed from all events, and replaced with `prev_state_events`. The state DAG
+  formed by `prev_state_events` is used to calculate `auth_events` for state resolution.
+  This targets the historic authorisation property.
+- Room state is derived from the DAG induced by `prev_state_events`, NOT `prev_events`.
+  Events which specify faulty `prev_state_events` (e.g. referencing non-state events or events in different rooms) are rejected.
+- `prev_state_events` are populated with the current forward extremities of the set of state events in the room.
```
reivilibre, line 184: What happens if there are too many forward state extremities that they don't fit in the event?

It seems difficult to do the same thing that we do for regular `prev_events` (which is pick only a limited number, e.g. 50 of them, in this case when there are too many forward extremities) as it seems doing so could change auth outcome.

Should servers automatically send dummy state events to merge the state DAG, as they do to merge the regular DAG?

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+#### Soft Failure
+
+> When an event is “soft failed” it should not be relayed to the client **nor be referenced by new events** created
+> by the homeserver (i.e. they **should not be added to the server’s list of forward extremities of the room**).
+> Soft failed events are otherwise handled as usual.
+
+https://spec.matrix.org/v1.12/server-server-api/#soft-failure
+
+Events in the state DAG **should be part of the forward extremities of a room** otherwise we run into the problems described below.
+This means only non-state message events are subject to soft-failure.
```
reivilibre, line 300:
```suggestion
This means only non-state events are subject to soft-failure.
```
I think you meant?

reivilibre: I wonder if this is actually what we 'should' do, or whether we should instead mean that soft-failed state events are kept part of the forward extremities for State DAG room versions — but other soft-failure effects still apply?

Soft-failure is also about protecting clients from rubbish that malicious servers are sending. It seems wise to keep some concept of 'non-displayed state event' that does not get added to a client's (`/sync`) timeline when it is received as a real-time event.


Perhaps worth also being explicit whether we mean 'regular' forward extremities or 'state DAG' forward extremities here, I'm not sure of the correct interpretation.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+that servers exponentially increase their `limit` until they have seen all events, e.g. `?limit=10`
+then 20, 40, 80... Servers MAY cap the number of events they return in order to ensure that the HTTP response
+size remain within the maximum limits set by their reverse proxy. This requires no signalling 
+e.g. `limited: true` because
+the termination condition is filling in the state DAG, not based on the number of events returned in any
+one response. This is a simple approach and there are better alternatives which can be explored
+in a future MSC which do fewer round trips and exchange less information.
+ - [Bloom filters](https://arxiv.org/abs/2012.00472)
+ - [Range-based set reconciliation](https://arxiv.org/abs/2212.13567) 
+ - [Merkle search trees](https://ieeexplore.ieee.org/abstract/document/9049566).
```
reivilibre: ([Prolly Trees](https://djhsjtu.github.io/jackyzha0.github.io/thoughts/Prolly-Trees) are also fun, seem related to Merkle Search Trees although I'm unfamiliar with the latter)

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+
+In addition to this, `/get_missing_events` is modified to return events deterministically. This is
+done by first sorting events by the minimum number of hops away they are from `latest_events`, then tie-breaking
+by sorting lexicographically: `$bar` before `$foo`. This enables database engines
+which use ASCII sorting (such as PostgreSQL's C collation) to sort correctly using `ORDER BY hops, event_id`.
```
reivilibre, line 393:
```suggestion
which use ASCII byte-wise lexicographic sorting (such as PostgreSQL's C collation) to sort correctly using `ORDER BY hops, event_id`.
```

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+(not a state event) then the response MUST include the `prev_state_events` for that message event.
+This ensures that a server can perform a single `/get_missing_events` request to optimistically fill in
+the state DAG. Without this, a server would need to make additional `/event` requests to fetch the `prev_state_events`
+in the message event and then use those state events in the `/get_missing_events` request.
+
+Finally, both `earliest_events` and `latest_events` have an independent size limit of 100 elements when
+`state_dag = true`. An HTTP 400 error is returned if a server attempts to set more elements than this.
+Both of these fields are derived from a room's forward extremities, just at different points in time.
+The expected number of forward extremities in a room is much lower than this limit. If a room were to
+genuinely have so many entries, servers would need to break up their `/get_missing_events` requests.
```
reivilibre, line 422: I'm not following the proposed workaround for this (niche, but seemingly possible especially under malice?) sad path:

If an event `$E` has 201 `prev_state_events` `$P1`...`$P201`, how do you query them?

I believe you:

- first query with `latest_events = ["$E"]` and `earliest_events = [(other events)]`, returning `$P1`..`$P100`
- now you query with `latest_events = ["$E"]` and `earliest_events = ["$P1", ..., "$P100"]`, returning `$P101`..`$P200`
- now you still need to query for `$P201` but you seemingly can't do so through `/get_missing_events`...

Do you have to handle this by getting a single event with [`/event/$P201`](https://spec.matrix.org/v1.19/server-server-api/#get_matrixfederationv1eventeventid)
I guess that's fine, not beautiful, but fine — though I guess we would be better acknowledging it.


EDIT: The sensible answer is probably in fact to define a limit on `prev_state_events` and ensure that this doesn't cause trouble for servers getting their events authed by other servers (because they can't refer to all the necessary state)

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+
+#### Faster remote room joins
+
+The state DAG model has implications on "faster remote room joins" which merely returns the auth chain for
+the join event, non-member current state events, a list of servers in the room and nothing else. This tiny subset of events is fast
+but provides very very weak guarantees that at the time the join was issued, those auth events existed
+in the room.
+
+State DAG rooms use a different approach. If `?omit_members=true`:
+ - The `/send_join` response still includes _the entire state DAG._ The same amount of data is transferred with and without faster room joins.
```
reivilibre, line 493: At first, this sounds like a contradiction, but maybe I'm following what you mean now.
Would 'the entire state DAG' not include all membership state events? Therefore `omit_members` surprises me.

I think one thing that got me is that it sounds like you're saying 'we will transfer the same amount of data as one of **today's** faster remote room joins, which appears to not be the case if we send some representing of every member event.

Don't think I hit the best wording but at least:
```suggestion
State DAG rooms use a different approach. If `?omit_members=true`:
 - The `/send_join` response still includes _the entire state DAG._ For state DAG rooms, the same amount of data will be transferred with and without faster room joins.
```

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+ - The `/send_join` response includes a new array: `timeline` which represents the most recent events in the room,
+   suitable for presenting in the timeline. How many events is configurable via a new query param `?timeline_limit=` which
+   is a non-zero integer, with a hard cap of 64. This is subject to history visibility rules.
+ - The `/send_join` response includes a new field: `partial_state_event_ids` which is a
+   list of event IDs which contain what the join-helper server claims to be the current state, omitting all members _except_
```
reivilibre, line 498: Don't think this is the normal term and it caught me off guard. I don't object to calling it that, but at least write the standard term alongside
```suggestion
   list of event IDs which contain what the join-helper (resident) server claims to be the current state, omitting all members _except_
```

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+The state DAG model has implications on "faster remote room joins" which merely returns the auth chain for
+the join event, non-member current state events, a list of servers in the room and nothing else. This tiny subset of events is fast
+but provides very very weak guarantees that at the time the join was issued, those auth events existed
+in the room.
+
+State DAG rooms use a different approach. If `?omit_members=true`:
+ - The `/send_join` response still includes _the entire state DAG._ The same amount of data is transferred with and without faster room joins.
+ - The `/send_join` response includes a new array: `timeline` which represents the most recent events in the room,
+   suitable for presenting in the timeline. How many events is configurable via a new query param `?timeline_limit=` which
+   is a non-zero integer, with a hard cap of 64. This is subject to history visibility rules.
```
reivilibre, line 496: low importance: Should this be non-zero or non-negative?
Could there be a case where a server doesn't want any timeline events?

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+ - The `/send_join` response includes a new field: `partial_state_event_ids` which is a
+   list of event IDs which contain what the join-helper server claims to be the current state, omitting all members _except_
+   the `sender`s of the `timeline` events and non-member state events e.g the `sender` of the `m.room.create` event.
+   This is sent alongside the entire state DAG. For clarity, the join event induced by `/send_join` is also omitted from
+   `partial_state_event_ids` as it is implied. This avoids scenarios where the join-helper server claims the joining server isn't in the room
+   despite just calling `/send_join`.
+ - The `/send_join` response includes another new field: `partial_auth_chain_ids` which is a map of event ID to calculated auth event IDs.
+   The join event and every event in `partial_state_event_ids` MUST be present as a key in this map, with the exception of the `m.room.create` event which MAY
+   be absent. In addition, the auth event IDs for the partial state events MUST be present as keys in this map. Their auth events MUST
```
reivilibre, line 505:
```suggestion
   The join event and every event in `partial_state_event_ids` MUST be present as a key in this map, with the exception of the `m.room.create` event which MUST
   be absent. In addition, the auth event IDs for the partial state events MUST be present as keys in this map. Their auth events MUST
```
Avoid needless ambiguity/fracture. If we write `MAY` we have to add compliance tests to cover both cases that could have been easily avoided by choosing a decision.

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+   Do not verify all events in the state DAG yet. Only verify the partial state event IDs and the timeline events with
+   steps 1, 2, 3 and 4 of the PDU checks (valid event, hashed and signed correctly, with correct `auth_events`).
+ - Persist the state and timeline and unblock the CSAPI `/join` endpoint.
+ - In the background, verify the entire state DAG with all steps of the PDU checks as usual. Ensure the claimed auth chain is deleted
+   and replaced with its calculated counterpart.
+ - If the state DAG is valid:
+    * destroy the state associated with the room and replace it with the verified state. Invalidate client caches.
+    * reverify all events in the `timeline`, as well as any live events which arrived in the interim.
+ - If the state DAG is not valid:
+    * destroy the state associated with the room. Invalidate client caches.
```
reivilibre, line 551: As in, delete the local copy of the room?

proposals/4242-state-dags.md
```diff
@@ -0,0 +1,819 @@
...
+   This reduces the amount of events that need to be checked on the critical path. This is particularly effective on member events as members may be on different
+   servers which will take time to fetch server keys for.
+
+From an implementation perspective, servers which implement [MSC3902](https://github.com/matrix-org/matrix-spec-proposals/pull/3902) should need minimal changes to
+adapt their system to work in state DAG rooms because the underlying data model is the same. Care must be taken to ensure that if the state DAG verification is
+interrupted e.g due to restarting the server, that it is resumed on startup.
+
+#### Consequences of having a full state DAG
+
+`depth` can be calculated and hence trusted for events that are included in the state DAG. This has
```
reivilibre, line 569: This seems incorrect, unless I missed a part where we are redefining `depth` of state events to only consider `prev_state_events`?

It appears to me that state events can still refer to regular events as part of `prev_events` and it is necessary to have the full ancestry of those `prev_events` to truly trust the `depth`.

Whichever explanation is correct here, it feels it should be elaborated in the MSC here as it seems I'm the second to ask

reivilibre (commented):
There's quite a lot here, but despite that it feels quite harmonious to the design of Matrix, so overall I think I quite like it. Some comments.

proposals/4242-state-dags.md
kegsay: The `depth` here refers to the state DAG not the complete room DAG. I can clarify.

PR #4245, opened 2024-12-21
MSC4245: Immutable encryption algorithm

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/immutable-encryption-algo/proposals/4245-immutable-encryption-algorithm.md)

----

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published under my Director of Standards Development hat, with a focus on interoperability between existing systems.

### DIFF
--- a/proposals/4245-immutable-encryption-algorithm.md
+++ b/proposals/4245-immutable-encryption-algorithm.md
@@ -1,0 +1,63 @@
+# MSC4245: Immutable encryption algorithm
+
+Enabling encryption in a room after it's been created works well for some encryption algorithms,
+like Megolm in today's specification, but doesn't work well when aiming to support RFC 9420 MLS in
+Matrix. Rooms using MLS will be required to change their authorization rules and possibly state
+resolution behaviours, which could be breaking if a server misses the [`m.room.encryption`](https://spec.matrix.org/v1.13/client-server-api/#mroomencryption)
+state event being sent. To avoid this situation, this proposal suggests that the encryption algorithm
+can *optionally* be specified in the [`m.room.create`](https://spec.matrix.org/v1.13/client-server-api/#mroomcreate)
+event instead, nullifying the `m.room.encryption` event's `algorithm` field.
+
+
+## Proposal
+
+When the `m.room.create` event's `content` contains an `encryption_algorithm` field, that encryption
+algorithm MUST be used within that room. The `algorithm` field of `m.room.encryption` serves no purpose
+in such rooms.
+
+If the `m.room.create` event's `content` does not have an `encryption_algorithm`, the `m.room.encryption`
+state event operates as per usual.
+
+Other fields/properties of `m.room.encryption` are not modified by this proposal, such as key rotation
+periods.
+
+Note that an empty string, or unrecognized value, may be present in `encryption_algorithm`: this is
+legal, and causes `m.room.encryption`'s `algorithm` to be ignored.
+
+
+## Alternatives
+
+None applicable
+
+
+## Potential issues
+
+From a security perspective, it's entirely possible that the `m.room.encryption` event's `algorithm`
+is different from the `m.room.create`'s `encryption_algorithm`. Clients may (rightly) get confused
+if they don't yet recognize `encryption_algorithm` in the create event, and use a less secure algorithm
+in the room. Receiving clients may further accept these events and attempt to decrypt them instead
+of rejecting due to the wrong algorithm being used. Or, worse, there are effectively two rooms going
+on as half the participants only encrypt for their half of the room.
+
+Solving this is tricky, and may require a room version bump with significant public information
+campaigns to encourage client developers to adopt the change, once the MSC is accepted. The Spec Core
+Team should consider rollout requirements prior to Final Comment Period of `merge` for this MSC.
+
+
+## Security considerations
+
+See potential issues.
+
+
+## Dependencies
+
+This MSC is not dependent on any other MSCs. [MSC4244](https://github.com/matrix-org/matrix-spec-proposals/pull/4244)
+relies on the functionality introduced by this proposal. Others may additionally exist.
+
+
+## Unstable prefix
+
+While this proposal is not considered stable, implementations should use `org.matrix.msc4245.encryption_algorithm`
+in the create event instead of `encryption_algorithm`. Developers should note that rooms created
+using this unstable prefix may break when implementations drop support for the unstable prefix,
+and not expose such rooms to production end users.


proposals/4245-immutable-encryption-algorithm.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server (to prevent clients making mistakes)
* [ ] A rollout plan

proposals/4245-immutable-encryption-algorithm.md
```diff
@@ -0,0 +1,63 @@
+# MSC4245: Immutable encryption algorithm
+
+Enabling encryption in a room after it's been created works well for some encryption algorithms,
+like Megolm in today's specification, but doesn't work well when aiming to support RFC 9420 MLS in
+Matrix. Rooms using MLS will be required to change their authorization rules and possibly state
+resolution behaviours, which could be breaking if a server misses the [`m.room.encryption`](https://spec.matrix.org/v1.13/client-server-api/#mroomencryption)
+state event being sent. To avoid this situation, this proposal suggests that the encryption algorithm
+can *optionally* be specified in the [`m.room.create`](https://spec.matrix.org/v1.13/client-server-api/#mroomcreate)
+event instead, nullifying the `m.room.encryption` event's `algorithm` field.
```
dkasak, line 9: I would like to see immutable encryption in all rooms, not just MLS rooms, and potentially even make it non-optional in future room versions. Changing the encryption protocol for a room from a server-controlled setting is inherently problematic and almost never useful.

Indeed, clients today mostly ignore any changes in the room encryption algorithm after they've observed the first `m.room.encryption` event. This means the existence of the `algorithm` field only risks client splitbrains without actually serving any purpose.

Placing the algorithm in the create event would ensure uniqueness because to change it, you'd have to be in a different room. This also hints at the natural way to change the algorithm: room upgrades.

turt2live: this proposal isn't MLS-specific, fwiw. It's just using MLS as a driver. The `encryption_algorithm` could also be a megolm string, making that room stuck on megolm.

The rationale for optionality is to allow existing rooms to continue "upgrading" from unencrypted to encrypted, while setting a standard for future rooms to use. As we see uptake (and critically, problems/cheers as a result of the immutability), we could remove `m.room.encryption`'s `algorithm` field.

dkasak: > this proposal isn't MLS-specific, fwiw. It's just using MLS as a driver. The encryption_algorithm could also be a megolm string, making that room stuck on megolm.

Great!

> The rationale for optionality is to allow existing rooms to continue "upgrading" from unencrypted to encrypted

I'm not sure that the benefits outweigh the costs, so I'm raising this as an early request to look whether we might want to be more proactive on this front. The encryption algorithm is very much something that should be baked into the room / group, and there should be no opportunities for participants to disagree on whether the room is encrypted or not and using which algorithm.

The modern Matrix approach is to rely on room upgrades more heavily, so it's really hard to come up with a convincing reason why encryption, of all things, should be left malleable, with all the downsides that carries.

neilalexander: > The rationale for optionality is to allow existing rooms to continue "upgrading" from unencrypted to encrypted

... which is fine when all participants are behaving and have accepted the `m.room.encryption` event, but due to the eventually consistent nature of Matrix, that's not guaranteed to happen immediately or even within any reasonable timeframe. Worse is that someone who enables encryption on a room has no visibility whatsoever into whether other participant servers are consistent or not until one of their users sends a message, by which time a leak could have happened in plain-text. 

I can see value in just deprecating `m.room.encryption` altogether in favour of an approach like this.

proposals/4245-immutable-encryption-algorithm.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Alternatives
+
+None applicable
+
+
+## Potential issues
+
+From a security perspective, it's entirely possible that the `m.room.encryption` event's `algorithm`
+is different from the `m.room.create`'s `encryption_algorithm`. Clients may (rightly) get confused
```
neilalexander, line 36: One possibility is to extend the auth rules so that an `m.room.encryption` event will only be accepted if either the create event doesn't specify an algorithm or if the `algorithm` matches the one specified in the create event exactly.

proposals/4245-immutable-encryption-algorithm.md
```diff
@@ -0,0 +1,63 @@
...
+resolution behaviours, which could be breaking if a server misses the [`m.room.encryption`](https://spec.matrix.org/v1.13/client-server-api/#mroomencryption)
+state event being sent. To avoid this situation, this proposal suggests that the encryption algorithm
+can *optionally* be specified in the [`m.room.create`](https://spec.matrix.org/v1.13/client-server-api/#mroomcreate)
+event instead, nullifying the `m.room.encryption` event's `algorithm` field.
+
+
+## Proposal
+
+When the `m.room.create` event's `content` contains an `encryption_algorithm` field, that encryption
+algorithm MUST be used within that room. The `algorithm` field of `m.room.encryption` serves no purpose
```
uhoreg, line 15: FWIW, MSC2883 basically does the same thing, though it uses a different name (`encryption` instead of `encryption_algorithm`).  But it's easy enough to harmonize the naming.  In MSC2883, I did consider making the `encryption` property an object (e.g. `encryption: {algorithm: "...", ...}`), in case we wanted to make other encryption parameters immutable.  But I don't know that there are any other such parameters.

proposals/4245-immutable-encryption-algorithm.md
```diff
@@ -0,0 +1,63 @@
...
+If the `m.room.create` event's `content` does not have an `encryption_algorithm`, the `m.room.encryption`
+state event operates as per usual.
+
+Other fields/properties of `m.room.encryption` are not modified by this proposal, such as key rotation
+periods.
+
+Note that an empty string, or unrecognized value, may be present in `encryption_algorithm`: this is
+legal, and causes `m.room.encryption`'s `algorithm` to be ignored.
+
+
```
uhoreg, line 27: Implementation note: assuming [MSC3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266) is accepted, if this MSC is also accepted, the endpoints that 3266 defines will need to also look at the `m.room.create` event to populate the `encryption` property.

PR #3554, opened 2021-12-07
Extensible Events - Translatable Text

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/extev/translatable/proposals/3554-extensible-events-translatable-messages.md)

Blocked by https://github.com/matrix-org/matrix-doc/pull/1767

PR #4247, opened 2024-12-27
MSC4247: User Pronouns

88572:
[Rendered](https://github.com/everypizza1/matrix-spec-proposals/blob/pronouns/proposals/4247-user-pronouns.md)

Implementations:
SDK:
- https://github.com/mautrix/go/pull/337


proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,42 @@
+# MSCXXXX: User Pronouns
```
turt2live, line 1:
```suggestion
# MSC4247: User Pronouns
```

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,42 @@
...
+
+Many users of Matrix put pronouns in display names. However, that causes screen clutter. This proposal defines a standardized pronouns field on top of MSC4133.
```
turt2live, line 3:
```suggestion
Many users of Matrix put pronouns in display names. However, that causes screen clutter. This proposal defines a standardized pronouns field on top of [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
```

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Proposal
+
+Profiles may have optional `m.pronouns` and `m.pronouns.secondary` fields as a dictionary. These fields can be fetched through the [profile API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles). Clients should parse this and use these instead of they/them where possible.
```
turt2live, line 7: Please wrap lines to about 100 characters

proposals/4247-user-pronouns.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,49 @@
...
+{ 
+    "avatar_url": "…", "displayname": "…",
+    "m.pronouns": {
+        "subject": "it",
+        "object": "it", 
+        "possessive_determiner": "its", 
+        "possessive_pronoun": "its", 
+        "reflexive": "itself"
+    }, 
+    "m.pronouns.secondary": {
```
nexy7574, line 26: Would it not be better to have `m.pronouns` be an ordered array of these objects, that way the user can have more than two sets of pronouns? This also implicitly keeps primary/secondary by nature of array indexes.

everypizza1: The only reason there's secondary pronouns is localization: some languages don't have concepts of some neopronouns like it/its.

proposals/4247-user-pronouns.md
tcpipuk: Conduwuit already supports this by supporting arbitrary fields in #4133 so I'd assume this MSC only requires a client implementation that can read/write the field?

proposals/4247-user-pronouns.md
tcpipuk: Would it be simpler to require they be English (like the rest of the API) then put the onus on the client to present a localised version?

You could include a "display" option to show a string alternative in the user's local language, but using a standard language allows clients to show icons or other indicators to describe the user, e.g. offering contextual notifications like "she said" or "they said".

nexy7574: Right, makes sense for localisation, however, my understanding from the proposal is that this would replace the need to put them in displayname. Many users have more than two sets of pronouns in their display name (or bios in other places), so would not allowing more than two sets of pronouns not be ideal here too?

everypizza1: Just made it an array.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,51 @@
+# MSC4247: User Pronouns
+
+Many users of Matrix put pronouns in display names. However, that causes screen
+clutter. This proposal defines a standardized pronouns field on top of
+[MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+
+## Proposal
+
+Profiles may have an optional `m.pronouns` field as an
+array. These fields can be fetched through the
```
uhoreg, line 10: Can you clarify whether this field is primarily intended for human consumption or for machine consumption?  The format seems more machine-friendly, but the introduction suggests that a human use-case (replacement for putting pronouns in display names).

Also, can you explain how the array should be interpreted?  I'm guessing that it is in order of preference from most preferred to less preferred?

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,51 @@
...
+Clients should parse this and use these instead of they/them where possible.
+
+### Example
+
+```json
+{ 
+    "avatar_url": "…", "displayname": "…",
+    "m.pronouns": [
+        {
+            "subject": "it",
```
uhoreg, line 21: Are the properties in this example all the properties that are used?  Are they all optional/required?  Generally, examples are not normative, so the properties should be defined outside of the example.

Also, might be helpful to have a language indicator somehow?

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,51 @@
...
+pronouns are a basic part of langauge.
+
+## Security issues
+
+None.
+
+## Unstable prefix
+
+Clients and servers wishing to implement this early may use
+`io.fsky.nyx.pronouns` and `io.fsky.nyx.pronouns.secondary`.
```
uhoreg, line 51: Why are there two names here, while the proposal itself defines only one name?

uhoreg (commented):
Seems pretty straightforward.  A few mostly-nitpicky comments.

proposals/4247-user-pronouns.md
everypizza1: That was from an old version, and I forgot to remove that.

proposals/4247-user-pronouns.md
everypizza1: Clients should have as many of these as the language supports, so all are optional.

proposals/4247-user-pronouns.md
everypizza1: Clients should parse the pronouns and return them to users in the format `subject`/`possessive_determiner`.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,60 @@
...
+            "possessive_pronoun": "hers",
+            "reflexive": "herself",
+            "language": "en",
+            "summary": "she/her"
+        }
+    ]
+}
+```
+The example uses it/its pronouns followed by she/her pronouns, both in English.
+The array is ordered by preference, `language` should be a language code, and
```
tulir, line 42: "language code" needs to be defined more precisely, I don't think anything in the current spec uses languages. https://github.com/matrix-org/matrix-spec-proposals/pull/3554 seems to reference [BCP-47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt)

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,60 @@
...
+clients should render the `summary` for the pronouns. Clients may offer
```
tulir, line 43: This sounds like `summary` should be required, and all other fields should be optional? Or maybe `language` should be required too? The other allowed fields should also be explicitly listed outside examples

everypizza1: Yes, summary and language are required, all else is optional.

proposals/4247-user-pronouns.md
everypizza1: Would ISO 639 language codes work?

tulir: I think BCP-47 uses ISO 639, but also allows optional subtags for specifying regions (so both `en` and `en-US` are allowed in BCP-47, while ISO 639 only has the concept of `en`)

everypizza1: I'll switch to BCB-47 then, so there's more flexibility.

proposals/4247-user-pronouns.md
clokep: I think @uhoreg is asking if `subject`/`object`/`possessive...` etc. should be defined in the protocol or not.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,62 @@
...
+```json
+{ 
+    "avatar_url": "…", "displayname": "…",
+    "m.pronouns": [
+        {
+            "subject": "it",
+            "object": "it", 
+            "possessive_determiner": "its", 
+            "possessive_pronoun": "its", 
+            "reflexive": "itself",
```
tulir, line 26: I'm not sure if these fields are a good idea, since they seem to be specific to English and probably aren't compatible with any other language. Also, I'd guess most implementations will simply use `summary` for displaying in the profile and won't even try to apply the other fields.

It might be better to just have the freeform field. Maybe also an enum (non-freeform string) for preferred grammatical gender, although even that could get complicated.

Is there any prior art or research into user-definable pronouns that support internationalization?

If the fields are kept as-is, each of them needs to be defined separately, it's not enough to have them in the example. Keeping the fields may also require an implementation actually using them to show they're useful.

tulir: Thinking more about the enum: it should at least have values `masculine`, `feminine` and `neuter`. The potential complications I can think of are:

* Some languages don't have a real neuter. Maybe that's not a problem though, because clients have to be able to fall back to gender being unknown anyway?
* Some languages have animate and inanimate neuter forms (singular they vs it in English). Does the enum need to have those separately? (`neuter_inanimate`)
* There might be other types in some weird languages. Are there any, and if there are, do they need to be options too? (and then how do other languages handle those options?)

On the implementation side: pronoun information is generally only needed for rendering state events like profile changes ("X changed his/her/their name"), but those state events don't include this profile info. Fetching the full profile for each profile change state event seems like a bad idea.

In any case, a freeform field should exist to display when viewing someones profile. If the other fields can't be used effectively, then it may be best to narrow this down to only have the freeform field and nothing else.

everypizza1: I think we may want to throw out `subject`, `object`, etc and do the `masculine`, `feminine`, and `neuter` with the `animate` and `inanimate` and for languages that don't support some of those, fall back to the closest to neutral.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+clutter. This proposal defines a standardized pronouns field on top of
+[MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+
+## Proposal
+
+Profiles may have an optional `m.pronouns` field as an
+array. These fields can be fetched through the
+[profile API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles).
+Clients should use these instead of they/them where possible. All fields
+within `m.pronouns` are optional, exluding `"language"` and `"summary"`.
```
sumnerevans, line 13:
```suggestion
within `m.pronouns` are optional, excluding `"language"` and `"summary"`.
```

(since there is only one more field, it might be better to reverse the wording of this sentence and say that all are required except grammatical gender)

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Example
+
+```json
+{ 
+    "avatar_url": "…", "displayname": "…",
+    "m.pronouns": [
+        {
+            "grammatical_gender": "inanimate",
```
sumnerevans, line 22: What grammatical genders are available? I think they need to be enumerated. I don't know what grammatical gender "they/them" would be, for example.

I also think they need to ne prefixed (`m.male` for example) and defined what that means in terms of how the UIs should adapt their rendering. This will also allow for future grammatical genders from other languages or linguistic innovations to be defined later (and use unstable prefixes).

proposals/4247-user-pronouns.md
nexy7574: With https://github.com/tulir/gomuks/pull/574, Gomuks supports rendering pronouns (setting them is not implemented in the UI yet, but is supported on the backend).

proposals/4247-user-pronouns.md
tcpipuk: I've tried out a few options across a few languages, and for languages like Russian that modify other words in the sentence depending on the pronouns, I'd probably use something like this so you can specify pronouns per-language (if you want, you could just specify English if you want) but also specify "grammatical_pattern" so software can know which form of words you want it to use around your pronouns:

```json
{
    "m.pronouns": [
        {
            "grammatical_pattern": "feminine",
            "forms": {
                "en": {
                    "subject": "she",
                    "object": "her",
                    "possessive_determiner": "her",
                    "possessive_pronoun": "hers",
                    "reflexive": "herself",
                    "dependent": "her"
                },
                "de": {
                    "subject": "sie",
                    "object": "sie",
                    "possessive_determiner": "ihre",
                    "possessive_pronoun": "ihres",
                    "reflexive": "sich",
                    "dependent": "sie"
                },
                "ja": {
                    "subject": "彼女",
                    "object": "彼女",
                    "possessive_determiner": "彼女の",
                    "possessive_pronoun": "彼女のもの",
                    "reflexive": "自分",
                    "dependent": "彼女"
                },
                "zh": {
                    "subject": "她",
                    "object": "她",
                    "possessive_determiner": "她的",
                    "possessive_pronoun": "她的",
                    "reflexive": "自己",
                    "dependent": "她"
                },
                "fi": {
                    "subject": "hän",
                    "object": "häntä",
                    "possessive_determiner": "hänen",
                    "possessive_pronoun": "hänen",
                    "reflexive": "itsensä",
                    "dependent": "hänen"
                }
            },
            "display": {
                "en": "she/her",
                "de": "sie/ihre",
                "ja": "彼女/彼女の",
                "zh": "她/她的",
                "fi": "hän/häntä"
            }
        }
    ]
}
```

What do you think?

tcpipuk: (also, some languages like Japanese vary the content based on the "formality" so you'd likely need to put some wording in that it's out of scope for this MSC, as public pronouns can't really specify whether _this particular interaction_ has a particular level of formality)

proposals/4247-user-pronouns.md
tcpipuk: Some languages distinguish between "animate" and "inanimate" so @everypizza1 may want to consider whether it's important to consider bot accounts as "inanimate" or whether we just anthropomorphise every user, but from brief research into a few African, Asian, and European languages, I think this should cover most of them:

```json
{
    "m.pronouns": [
        {
            "grammatical_pattern": "masculine",
            "forms": {
                "en": {
                    "subject": "he",
                    "object": "him",
                    "possessive_determiner": "his",
                    "possessive_pronoun": "his",
                    "reflexive": "himself",
                    "dependent": "him"
                }
            },
            "display": { "en": "he/him" }
        },
        {
            "grammatical_pattern": "feminine",
            "forms": {
                "en": {
                    "subject": "she",
                    "object": "her",
                    "possessive_determiner": "her",
                    "possessive_pronoun": "hers",
                    "reflexive": "herself",
                    "dependent": "her"
                }
            },
            "display": { "en": "she/her" }
        },
        {
            "grammatical_pattern": "neuter",
            "forms": {
                "en": {
                    "subject": "it",
                    "object": "it",
                    "possessive_determiner": "its",
                    "possessive_pronoun": "its",
                    "reflexive": "itself",
                    "dependent": "it"
                }
            },
            "display": { "en": "it/its" }
        },
        {
            "grammatical_pattern": "common",
            "forms": {
                "sv": {
                    "subject": "hen",
                    "object": "hen",
                    "possessive_determiner": "hens",
                    "possessive_pronoun": "hens",
                    "reflexive": "sig",
                    "dependent": "hen"
                }
            },
            "display": { "sv": "hen/hens" }
        },
        {
            "grammatical_pattern": "plural",
            "forms": {
                "en": {
                    "subject": "they",
                    "object": "them",
                    "possessive_determiner": "their",
                    "possessive_pronoun": "theirs",
                    "reflexive": "themself",
                    "dependent": "them"
                }
            },
            "display": { "en": "they/them" }
        },
        {
            "grammatical_pattern": "epicene",
            "forms": {
                "en": {
                    "subject": "ze",
                    "object": "zir",
                    "possessive_determiner": "zir",
                    "possessive_pronoun": "zirs",
                    "reflexive": "zirself",
                    "dependent": "zir"
                }
            },
            "display": { "en": "ze/zir" }
        },
        {
            "grammatical_pattern": "none",
            "forms": {
                "fi": {
                    "subject": "hän",
                    "object": "häntä",
                    "possessive_determiner": "hänen",
                    "possessive_pronoun": "hänen",
                    "reflexive": "itsensä",
                    "dependent": "hänen"
                }
            },
            "display": { "fi": "hän/häntä" }
        }
    ]
}
```

There are extra forms not covered here, e.g. some languages have "noble" versions, but this MSC might want to just state an assumption that a chat protocol does not necessarily need or want to record the nobility level of the user 🙂

proposals/4247-user-pronouns.md
tcpipuk: Running the test against more languages, Hungarian and Tamil would be a challenge... they have loads of forms that would be quite difficult to describe without supporting all of them 🤔

```json
{
    "grammatical_pattern": "feminine",
    "forms": {
        "hu": {
            "subject": "ő", // Nominative: She
            "object": "őt", // Accusative: Her (direct object)
            "possessive_determiner": "ővé", // Possessive determiner: Her
            "possessive_pronoun": "övé", // Possessive pronoun: Hers
            "reflexive": "magának", // Reflexive: Herself
            "instrumental": "vele", // Instrumental: With her
            "dative": "neki", // Dative: To/for her
            "locative": "nála", // Locative: At her place
            "ablative": "tőle", // Ablative: From her
            "allative": "hozzá", // Allative: To her
            "elative": "belőle", // Elative: Out of her
            "illative": "bele", // Illative: Into her
            "superessive": "rajta", // Superessive: On her
            "sublative": "alá", // Sublative: Under her
            "delative": "róla", // Delative: From (off) her
            "terminative": "őig", // Terminative: As far as her
            "essive": "őként", // Essive: As her
            "translative": "ővé", // Translative: Becoming her
            "causal-final": "miatta", // Causal-final: Because of her
            "temporal": "őtől fogva", // Temporal: Since her
            "dependent": "őt" // General dependent form: Her
        }
    },
    "display": {
        "hu": "ő/őt"
    }
},
{
    "grammatical_pattern": "feminine",
    "forms": {
        "ta": {
            "subject": "அவள்", // Nominative: She
            "object": "அவளை", // Accusative: Her (direct object)
            "possessive_determiner": "அவளுடைய", // Possessive determiner: Her
            "possessive_pronoun": "அவளுடையது", // Possessive pronoun: Hers
            "reflexive": "தன்னால்", // Reflexive: Herself
            "instrumental": "அவளால்", // Instrumental: With/by her
            "dative": "அவளுக்கு", // Dative: To/for her
            "locative": "அவளிடம்", // Locative: At her place
            "ablative": "அவளிடமிருந்து", // Ablative: From her
            "genitive": "அவளுடைய", // Genitive: Belonging to her
            "vocative": "அவளே", // Vocative: Calling her
            "dependent": "அவளை" // General dependent form: Her
        }
    },
    "display": {
        "ta": "அவள்/அவளை"
    }
}
```

That said, if the main purpose is to just tell people some pronouns, and there's no expectation that clients will have different messages/etc based on the pronoun data, perhaps it could just be stated in the MSC that many of these are not needed for an online chat medium so have been intentionally left out?

It seems to me that the purpose of this MSC is to help people know what pronouns to use when talking to another person, not to entirely describe how a given language's grammatical system works.

proposals/4247-user-pronouns.md
tulir, line 1: @ara4n pointed out JSContact (RFC9553) in https://github.com/matrix-org/matrix-spec-proposals/pull/4175#discussion_r2271344930. Looks like they have a pronoun field and ended up with basically the same design of declaring a grammatical gender and summary: https://www.rfc-editor.org/rfc/rfc9553.html#name-speaktoas

Main difference is that they only allow declaring 1 grammatical gender, which seems quite dumb. It's also a lot more complicated overall (translations are in an entirely separate top-level object and I have no idea what those k19 and k32 keys are)

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
+# MSC4247: User Pronouns
+
+Many users of Matrix put pronouns in display names. However, that causes screen
+clutter. This proposal defines a standardized pronouns field on top of
+[MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+
+## Proposal
+
+Profiles may have an optional `m.pronouns` field as an
+array. These fields can be fetched through the
```
dbkr, line 10: With regards to this being an array, note that JContact is a dictionary with context and preference order which means it's possible to designate different pronouns for different situations, and also to have pronouns of equal preference. However, given that this also has to be easy for a user to input and read, I think I'm leaning towards the solution of an array and assuming that if some are context-dependent then it will be apparent to the reader which to use when. In any case, a machine won't know this context so it's useful to express a default for this. I do think it's worth adding as an alternative though: this proposal lacks an alternatives section and it could probably use one.

proposals/4247-user-pronouns.md
dbkr: To nudge this back on point, I agree that these really could do with being enumerated: free text isn't useful here as clients need to know what values are possible and how to handle them. I don't think they need `m.` prefixes as I doubt we'd ever allow custom reverse DNS values here. I *do* think that expanding on JContact's definitions  by giving the sets of pronouns they're intended to map to (in whatever language) alongside each one, and probably setting a standard for the set of English pronouns to use. For example, it's unclear what "animate" is actually supposed to mean given that in English there would be animate-feminine, animate-masculine, etc.

I'd also argue for giving some hints as to how clients might use these. For example, I'd suggest offering a dropdown of common sets of pronouns and then also allowing free text. If the user picks a common one, the grammatical_gender can be set accordingly. If they enter free text, grammatical_gender gets omitted.

Also also, I'd argue that we don't need to get this 100% complete on the first go. I'd suggest starting off simply with the common English ones: 'neuter' meaning (they/them), 'masculine' (he/him), 'feminine' (she/her) and 'inanimate' (it/its) - anything else is expressed with free text (or a client could even add others to the dropdown if it chooses, it would just leave grammatical_gender unset). Others mappings between pronouns and grammatical_gender can be added in other MSCs.

We could also remind clients that "Not specified" must be an option (indeed, the default!)

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Example
+
+```json
+{ 
+    "avatar_url": "…", "displayname": "…",
+    "m.pronouns": [
+        {
+            "grammatical_gender": "inanimate",
+            "language": "en",
```
dbkr, line 20: I would probably vote for removing 'language' here, at least for a first cut of this feature. It's not 100% clear whether it's describing just the summary field or the grammatical_gender field too, and what the latter would mean. I feel like it's probably a vestige from when this MSC had the full set of pronouns individually. If it is included, I think it would need a fair bit of explanation on how it would actually be applied.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+            "summary": "she/her"
+        }
+    ]
+}
+```
+The example uses it/its pronouns followed by she/her pronouns, both in English.
+The array is ordered by preference, `language` should be a
+[BCP-47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt) language code, and
+clients should render the `summary` for the pronouns. Clients SHOULD use the
```
dbkr: Firstly, this is duplicative with the first paragraph of the proposal. Secondly, this is *nontrivial*. I think we need to give clients some hints on how they'd actually do this, ie. by picking an appropriate translation using the grammatical_gender field, accepting that this might get quite complex in different language, however I think it's okay to leave the specifics of what to use in different languages to the clients. I think it's worth specifying that they should be using grammatical_gender and *not* trying to parse the summary.

Also, I wouldn't use the phrase "state events" here (the events themselves are protocol-level), I would generalise to whenever the client refers to the user as you did above, calling out display name changes as an example.

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Potential issues
+
+Some users may not want to publish pronouns to others, although that is out of
+scope for this MSC. Some users may also complain about "woke", although
+pronouns are a basic part of langauge.
+
+## Security issues
+
+None.
```
dbkr: Potential for abusive content in the free-text field (obviously no more so than display names though, but I think it's worth noting).

31a05b9c: unlike display names, this is expected to inject arbitrary text into completely unmarked parts of the client's interface. clients should be reminded that nothing prevents the pronouns from being set to someone else's name or an entire paragraph which looks suspiciously like the surrounding UI text

proposals/4247-user-pronouns.md
```diff
@@ -0,0 +1,54 @@
...
+        {
+            "grammatical_gender": "feminine"
+            "language": "en",
+            "summary": "she/her"
+        }
+    ]
+}
+```
+The example uses it/its pronouns followed by she/her pronouns, both in English.
+The array is ordered by preference, `language` should be a
```
Merikei: Think what the order is should be defined too (presumably "most preferred first"?)

Example based on `content/client-server-api/modules/search.md`:
> -   `rank`, which returns the most relevant results first.

proposals/4247-user-pronouns.md
ranidspace: From what I can tell, state events are changed based on the `grammatical_gender` field which clients would have to implement each one manually. It should fall back to the default neutral set when encountering a `grammatical_gender` it cant handle, and then any instances of `summary`, as it would appear in a profile or next to a message, can be marked similar to a username. This is my guess though, I feel like the MSC should be more explicit on how to actually handle these fields.

proposals/4247-user-pronouns.md
nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/fb6cb2f76437a9fef75c85e00774f453141ae9de

PR #4250, opened 2025-01-11
MSC: Authenticated media v2 (Cookie authentication for Client-Server API)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/authenticated-media-v2/proposals/0000-authenticated-media-v2-cookies.md)

### DIFF
--- a/proposals/0000-authenticated-media-v2-cookies.md
+++ b/proposals/0000-authenticated-media-v2-cookies.md
@@ -1,0 +1,210 @@
+# MSC0000: Authenticated media v2 (Cookie authentication for Client-Server API)
+
+With the introduction of authenticated media in Matrix 1.11, some browser environments became unable
+to view *unencrypted* media because they were unable to append the required `Authorization` header.
+Specifically, in Firefox Private Browsing Mode (and similar for non-Chromium browsers),
+[Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) are unavailable,
+leaving the client two options:
+
+1. Buffer *all* media into a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob), like they
+   typically do for encrypted media; or
+2. Show errors, infinite spinners, or another indication that the media will never arrive.
+
+Option 2 is obviously not great, but option 1 isn't far off: clients will experience abnormally high
+memory usage just to show user/room avatars and thumbnails, and will not be able to benefit from
+[`Range` requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) to stream data
+to the user. For example, if a user wishes to download a 10mb PDF, the client would need to download
+the entire PDF into a Blob first, then serve that to the user (assuming it doesn't crash due to an
+arbitrary memory limit first). This could take quite a while depending on network conditions too,
+holding even more memory if the user tries to switch to another room while the download is still
+happening.
+
+Ideally, the spec can give the affected clients a third option to allow the non-trivial number of
+affected users a chance at seeing media again. This proposal is heavily inspired by
+[issue #1949](https://github.com/matrix-org/matrix-spec/issues/1949) which explores some potential
+options in this area, and puts forward [Cookies](https://en.wikipedia.org/wiki/HTTP_cookie) as the
+best way to solve the issue. This is an opt-in authentication mechanism for specific Client-Server API
+endpoints, allowing the client to access the protected resources.
+
+## Background
+
+Readers not familiar with authenticated media should read the following resources:
+
+* https://matrix.org/blog/2024/06/26/sunsetting-unauthenticated-media/
+* https://matrix.org/blog/2024/06/20/matrix-v1.11-release/
+* https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3916-authentication-for-media.md
+* https://spec.matrix.org/v1.13/client-server-api/#content-repo-client-behaviour
+
+## Proposal
+
+The Client-Server API's [authentication mechanisms](https://spec.matrix.org/v1.13/client-server-api/#client-authentication)
+are expanded to include an opt-in cookie for specific endpoints. Clients which present a valid cookie,
+valid access token, or both to the affected endpoints are considered authenticated for that request.
+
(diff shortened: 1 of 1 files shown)


proposals/4250-authenticated-media-v2-cookies.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server
* [ ] Evidence that this actually works in Firefox private browsing mode and other similar environments

proposals/4250-authenticated-media-v2-cookies.md
```diff
@@ -0,0 +1,210 @@
+# MSC4250: Authenticated media v2 (Cookie authentication for Client-Server API)
+
+With the introduction of authenticated media in Matrix 1.11, some browser environments became unable
+to view *unencrypted* media because they were unable to append the required `Authorization` header.
+Specifically, in Firefox Private Browsing Mode (and similar for non-Chromium browsers),
+[Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) are unavailable,
+leaving the client two options:
+
+1. Buffer *all* media into a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob), like they
```
benkuly, line 9: Another alternative: Stream the image via plain old fetch into OPFS and then load it from there. OPFS is not always available in private tabs. In that case the media can be streamed into IndexedDb too. That's what Trixnity does.

proposals/4250-authenticated-media-v2-cookies.md
```diff
@@ -0,0 +1,210 @@
...
+
+A user supplying an expired or useless cookie, and no other form of authentication:
+```
+GET /_matrix/client/v1/media/download/example.org/abc123
+Cookie: not_auth=token
+
+-> 401 M_MISSING_TOKEN
+```
+
+## Potential issues
```
ara4n, line 149: How well does this with with CDNs? (especially in comparison with the Auth header approach).  As I assume this is the main reason that S3 & Discord etc use timed/hashed query params...

turt2live: It should have zero impact because we still hit the homeserver before redirection.

proposals/4250-authenticated-media-v2-cookies.md
```diff
@@ -0,0 +1,210 @@
...
+the entire PDF into a Blob first, then serve that to the user (assuming it doesn't crash due to an
+arbitrary memory limit first). This could take quite a while depending on network conditions too,
+holding even more memory if the user tries to switch to another room while the download is still
+happening.
+
+Ideally, the spec can give the affected clients a third option to allow the non-trivial number of
+affected users a chance at seeing media again. This proposal is heavily inspired by
+[issue #1949](https://github.com/matrix-org/matrix-spec/issues/1949) which explores some potential
+options in this area, and puts forward [Cookies](https://en.wikipedia.org/wiki/HTTP_cookie) as the
+best way to solve the issue. This is an opt-in authentication mechanism for specific Client-Server API
```
dbkr, line 26: I think it was more siding with timed URLs, at least as far as my reasoning went (although this is great as an MSC for the cookie auth option regardless).

proposals/4250-authenticated-media-v2-cookies.md
```diff
@@ -0,0 +1,210 @@
...
+implementation-specific authentication cookie. Servers MAY decline to set a cookie for any implementation-
+specific reason, such as seeing the cookie is already retained by the client, or a local security
+policy denies the use of cookies.
+
+When setting cookies, servers SHOULD scope the cookie to a specific domain, and apply a relatively
+short expiration (minutes, not hours). Using the `Path` and `Secure` parameters is also recommended.
+
+Servers SHOULD note that some browsers and cookie jar implementations may restore expired cookies, and
+SHOULD therefore check expiration against an internal database rather than rely on the client-supplied
+value.
```
dbkr, line 58: Not quite sure how questionable cookie jar impls are relevant here: isn't rejecting expired cookies a fundamental security requirement for the server? Or maybe I've misunderstood what this is trying to say.

proposals/4250-authenticated-media-v2-cookies.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Proposal
+
+The Client-Server API's [authentication mechanisms](https://spec.matrix.org/v1.13/client-server-api/#client-authentication)
+are expanded to include an opt-in cookie for specific endpoints. Clients which present a valid cookie,
+valid access token, or both to the affected endpoints are considered authenticated for that request.
+
+Clients opt into the cookie authentication by setting the `?set_auth_cookie=true` query parameter when
+calling [`/sync`](https://spec.matrix.org/v1.13/client-server-api/#get_matrixclientv3sync) or
+[`/whoami`](https://spec.matrix.org/v1.13/client-server-api/#get_matrixclientv3accountwhoami) (for
```
dbkr, line 46: I think I was coming down on the side of a dedicated 'cookie pls' endpoint rather than dual-purposing sync or whoami, since the only benefit would be to save a few requests by tacking it onto sync requests, whereas you would lose precise control on when the cookie got refreshed and add interactions between media and sync that aren't really necessary. For `whoami` I really don't see any benefit to dual-purposing.

proposals/4250-authenticated-media-v2-cookies.md
turt2live: It was a weird thing for the MDN page to say in a brief scroll, so carried here as a recommendation.

proposals/4250-authenticated-media-v2-cookies.md
turt2live: A benefit of dual purpose is the client doesn't have to care about the cookie at all: it'll get appended by their http client automatically, and the server can decide when to override it. This also allows a server to set `HttpOnly` if it wants.

proposals/4250-authenticated-media-v2-cookies.md
turt2live: "this" is meant to be the MSC rather than issue - will clarify

PR #3179, opened 2021-05-04
Clarify that implementations can use stable prefixes once an MSC has merged

anoadragon453:
Fixes #3146.

This PR changes the Matrix Spec Proposals page to clarify that implementations **do not** need to wait until a spec release to use stable prefixes, but that they can do so after the corresponding MSC has been merged. The justification is that once an MSC has been merged, it's fairly guaranteed that it will land in the spec. Yet it will take time for the spec release process to run its course, and we shouldn't make implementations wait for that.

The exception to this is if implementating a feature in a backwards-compatible way requires a new spec version to indicate to clients/servers that a feature has been added/changed. This situation is rare though, and most implementations won't fall into this category.

There's a couple changes at the top of the diff which were just correcting errors in the document.

PR #4253, opened 2025-01-15
MSC4253: Modifying or rejecting accepted MSCs

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/process/modify-accepted/proposals/4253-spec-process-modify-reject-accepted.md)

----

### DIFF
--- a/proposals/4253-spec-process-modify-reject-accepted.md
+++ b/proposals/4253-spec-process-modify-reject-accepted.md
@@ -3,7 +3,7 @@
 The [spec process](https://spec.matrix.org/proposals/) defines the stages and steps an MSC goes through
 before becoming actual specification. That process has an awkward accepted-but-not-merged state where
 the MSC has successfully completed a *merge* FCP, but has not yet had a spec PR merged to introduce
-that MSC to the formal specification. During this time, implementations MAY use stable endpoints when
+that MSC to the formal specification. During this time, implementations MAY use stable identifiers when
 speaking the MSC.
 
 The general theory of this stage is that after several rounds of review, the MSC is unlikely to materially
@@ -26,7 +26,7 @@ For MSCs which are *accepted* but not *merged* (anywhere between `finished-final
    enact this process step. This step is known as "post-acceptance rejection", regardless of target
    state for the affected MSC.
 
-2. Another MSC is required to change the text of another MSC, provided it does two things:
+2. Another MSC is required to change the text of the accepted MSC, provided it does two things:
 
    1. Describe the rationale for the change being made in a dedicated MSC; and
    2. Modify the accepted MSC's actual text in the same GitHub PR. This is to ensure that the change


proposals/4253-spec-process-modify-reject-accepted.md
turt2live, line 1: Implementation requirements: 

None

proposals/4253-spec-process-modify-reject-accepted.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Potential issues
+
+Spec PR reviewers may still miss MSCs which modify another accepted MSC. SCT members should consider
+when it is more appropriate to use post-acceptance rejection instead of modification. For example, it
+may be more correct to pull an MSC back to `in-review` when a modification would be significant or
+easily forgotten.
+
+Stable implementations of accepted MSCs may be severely affected by either process step. The SCT is
+expected to work out a plan for how to address those incompatibilities when performing either step.
```
Johennes, line 48: Possibly a stupid question but what if using stable identifiers was prohibited until the MSC actually lands in a spec release? That is to say treat it as unstable until it is actually released.

turt2live: It'd probably have no effect, as there's no stable implementations. Room versions would fall under this because they require another MSC to actually make them stable.

proposals/4253-spec-process-modify-reject-accepted.md
```diff
@@ -0,0 +1,65 @@
...
+`disposition-merge` to `spec-pr-in-review`, inclusive), the following process steps MAY be taken:
+
+1. The SCT may choose to revert FCP acceptance to bring the MSC back to "open, in review" or pull the
+   MSC to `rejected` instead. The SCT is required to provide guidance to stable implementations, if
+   applicable, for how to handle the change when this happens. It is left as an implementation detail
+   to determine whether an MSC is brought back to `in-review` or pushed to `rejected`, and how to
+   enact this process step. This step is known as "post-acceptance rejection", regardless of target
+   state for the affected MSC.
+
+2. Another MSC is required to change the text of another MSC, provided it does two things:
```
KitsuneRal, line 29:
```suggestion
2. Another MSC is required to change the text of the accepted MSC, provided it does two things:
```

proposals/4253-spec-process-modify-reject-accepted.md
```diff
@@ -0,0 +1,65 @@
+# MSC4253: Modifying or rejecting accepted MSCs
+
+The [spec process](https://spec.matrix.org/proposals/) defines the stages and steps an MSC goes through
+before becoming actual specification. That process has an awkward accepted-but-not-merged state where
+the MSC has successfully completed a *merge* FCP, but has not yet had a spec PR merged to introduce
+that MSC to the formal specification. During this time, implementations MAY use stable endpoints when
```
clokep, line 6:
```suggestion
that MSC to the formal specification. During this time, implementations MAY use stable identifiers when
```

proposals/4253-spec-process-modify-reject-accepted.md
```diff
@@ -0,0 +1,65 @@
...
+2. Another MSC is required to change the text of another MSC, provided it does two things:
+
+   1. Describe the rationale for the change being made in a dedicated MSC; and
+   2. Modify the accepted MSC's actual text in the same GitHub PR. This is to ensure that the change
+      is captured in two ways: with a dedicated, unmerged, MSC and as real rendered text in the event
+      someone is reviewing the accepted MSC's text.
+
+When reviewing spec PRs for accuracy to their respective MSCs, reviewers are encouraged to use the
+accepted, rendered, text as reference rather than the original GitHub PR for the MSC due to PRs not
+updating post-merge.
```
clokep, line 38: Maybe clarify this means "what's in the `main` branch"?

proposals/4253-spec-process-modify-reject-accepted.md
clokep: This used to be the case but it felt it slowed down ecosystem progress too much by tying features to spec releases. It would simplify handling of version flags though.

turt2live: (oh, I failed at reading the question, sorry - clokep's response is much more accurate)

clokep: Maybe with quarterly releases this is no longer necessary, however?

proposals/4253-spec-process-modify-reject-accepted.md
```diff
@@ -0,0 +1,65 @@
...
+2. Another MSC is required to change the text of the accepted MSC, provided it does two things:
+
+   1. Describe the rationale for the change being made in a dedicated MSC; and
+   2. Modify the accepted MSC's actual text in the same GitHub PR. This is to ensure that the change
+      is captured in two ways: with a dedicated, unmerged, MSC and as real rendered text in the event
+      someone is reviewing the accepted MSC's text.
+
+When reviewing spec PRs for accuracy to their respective MSCs, reviewers are encouraged to use the
+accepted, rendered, text as reference rather than the original GitHub PR for the MSC due to PRs not
+updating post-merge.
```
HarHarLinks, line 38: Why would the mess of modifying an accepted MSC through another be preferable over reopening/replacing the original?

I'm not even sure reopening one is great when that does not really fit the tooling (i.e. GitHub workflow/limitations), when it can be replace with a new one for free?

proposals/4253-spec-process-modify-reject-accepted.md
clokep: https://github.com/matrix-org/matrix-spec-proposals/pull/3179 should hopefully be a launch point for context.

clokep: Oddly that change want by an MSC?

turt2live: Process changes aren't required to have an MSC, but it can be helpful depending on the level of feedback someone is after. In that case, the SCT spent significant time discussing it internally before making the PR, so felt an MSC was not required.

proposals/4253-spec-process-modify-reject-accepted.md
turt2live: Github doesn't allow a merged PR to be reopened. Process-wise, we'd end up always rejecting and rewriting if we were to do that, which is high process for little value compared to just writing a small diff.

proposals/4253-spec-process-modify-reject-accepted.md
turt2live: on the actual topic of stable identifiers: the spec process is supposed to already acknowledge that while developers *can* switch to stable identifiers, there is still risk in doing so. Developers also can't assume what version of the spec a feature will land in, which makes some/many stable identifier usages less useful without the "unstable stable flag" on `/versions`.

Johennes: > Maybe with quarterly releases this is no longer necessary, however?

Yeah, it doesn't seem like an extensive benefit anymore.

It also feels a little odd to me to implement something that is not spec'ed as if it was spec'ed, especially when RFC keywords are often not explicitly nailed down in MSCs.

PR #4254, opened 2025-01-17
MSC4254: Usage of RFC7009 Token Revocation for Matrix client logout

sandhose:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/quenting/oauth2-revocation/proposals/4254-oauth2-revocation.md)

Implemented in `matrix-authentication-service`, Element X iOS and Android (through the matrix-rust-sdk) and Element Web

Part of the following proposal:

- https://github.com/matrix-org/matrix-spec-proposals/pull/3861

---

I am a Software Engineer at Element. This proposal was written and published as an Element employee.

----

SCT stuff:

[checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4254#issuecomment-2616673667)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4254#issuecomment-2616673983)

### DIFF
--- a/proposals/4254-oauth2-revocation.md
+++ b/proposals/4254-oauth2-revocation.md
@@ -74,6 +74,15 @@ The client should handle these errors appropriately:
 - If the client is not authorized to revoke the token, the server returns a 401 Unauthorized response
 - For other errors, the server returns a 400 Bad Request response with error details
 
+### Replacement of existing APIs
+
+This proposal replaces the existing [`/_matrix/client/v3/logout`] endpoint for [MSC861]-compatible client.
+Those clients MUST use this mechanism to logout, and clients using the [`/_matrix/client/v3/login`] endpoint to login MUST keep using the existing [`/_matrix/client/v3/logout`] endpoint.
+
+Note that this proposal does not itself provide alternatives to endpoints like [`POST /_matrix/client/v3/login/all`], [`DELETE /_matrix/client/v3/devices/{deviceId}`] or [`POST /_matrix/client/v3/delete_devices`].
+Under the [MSC3861] proposal, management of other devices is not the responsibility of the client, and should instead be provided in a separate user interface by the homeserver.
+
+
 ## Potential issues
 
 The main consideration around token revocation is ensuring proper cleanup of all related tokens and state. The server must:
@@ -105,3 +114,8 @@ Token revocation is a critical security feature that allows users to terminate a
 [RFC7009]: https://tools.ietf.org/html/rfc7009
 [MSC2965]: https://github.com/matrix-org/matrix-spec-proposals/pull/2965
 [MSC3861]: https://github.com/matrix-org/matrix-spec-proposals/pull/3861
+[`/_matrix/client/v3/login`]: https://spec.matrix.org/v1.13/client-server-api/#login
+[`/_matrix/client/v3/logout`]: https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3logout
+[`POST /_matrix/client/v3/login/all`]: https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3loginall
+[`DELETE /_matrix/client/v3/devices/{deviceId}`]: https://spec.matrix.org/v1.13/client-server-api/#delete_matrixclientv3devicesdeviceid
+[`POST /_matrix/client/v3/delete_devices`]: https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3delete_devices
--- a/proposals/4254-oauth2-revocation.md
+++ b/proposals/4254-oauth2-revocation.md
@@ -1,0 +1,100 @@
+# MSC4254: Usage of [RFC7009] Token Revocation for Matrix client logout
+
+This proposal is part of the broader [MSC3861: Next-generation auth for Matrix, based on OAuth 2.0/OIDC][MSC3861].
+
+This MSC specifies how Matrix clients should use OAuth 2.0 Token Revocation as defined in [RFC7009] to implement client logout.
+
+## Proposal
+
+### Prerequisites
+
+This proposal requires the client to know the following authorization server metadata about the homeserver:
+
+- `revocation_endpoint`: the URL where the client is able to revoke tokens
+
+The discovery of the above metadata is out of scope for this MSC, and is currently covered by [MSC2965].
+
+### Token revocation
+
+When a user wants to log out from a client, the client should revoke either its access token or refresh token by making a POST request to the revocation endpoint as described in [RFC7009].
+
+The server must revoke both the access token and refresh token associated with the token provided in the request.
+
+The request includes:
+- The `token` parameter containing either the access token or refresh token to revoke
+- Optionally, the `token_type_hint` parameter, with either the `access_token` or `refresh_token` value. If provided, the server must use this value to determine which token to revoke
+- The `client_id` obtained during client registration
+
+#### Sample flow
+
+Revoking using the access token:
+
+```http
+POST /oauth2/revoke HTTP/1.1
+Host: auth.example.com
+Content-Type: application/x-www-form-urlencoded
+
+token=mat_ooreiPhei2wequu9fohkai3AeBaec9oo&
+token_type_hint=access_token&
+client_id=s6BhdRkqt3
+```
+
+```http
+HTTP/1.1 200 OK
+```
+
+Or equivalently, using the refresh token:
+
+```http
+POST /oauth2/revoke HTTP/1.1
+Host: auth.example.com
+Content-Type: application/x-www-form-urlencoded
+
+token=mar_Pieyiev3aenahm4atah7aip3eiveizah&
+token_type_hint=refresh_token&
+client_id=s6BhdRkqt3
+```
+
+```http
+HTTP/1.1 200 OK
+```
+
+### Handling errors
+
+The server may return an error response as defined in [RFC7009]. The client should handle these errors appropriately:
+
+- If the token is already revoked, the server returns a 200 OK response
+- If the client is not authorized to revoke the token, the server returns a 401 Unauthorized response
+- For other errors, the server returns a 400 Bad Request response with error details
+
+## Potential issues
+
+The main consideration around token revocation is ensuring proper cleanup of all related tokens and state. The server must:
+
+1. Track the relationship between access tokens and refresh tokens
+2. Properly revoke both tokens when either one is provided
+3. Clean up any Matrix device associated with the session
+
+## Alternatives
+
+### OpenID Connect RP-Initiated Logout
+
+OpenID Connect defines a [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) specification that allows clients to initiate a logout through a browser redirect. This would:
+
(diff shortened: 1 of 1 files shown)


turt2live (approved):
nice and simple: call revoke to revoke

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,100 @@
...
+client_id=s6BhdRkqt3
+```
+
+```http
+HTTP/1.1 200 OK
+```
+
+### Handling errors
+
+The server may return an error response as defined in [RFC7009]. The client should handle these errors appropriately:
```
richvdh, line 64: could we be explicit that the response is an RFC6749 error response rather than a Matrix error response?

```suggestion
The server may return an error response as defined in [RFC7009]. Note that RFC7009 mandates a [RFC6749 error response](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2) rather than a Matrix standard error response.

The client should handle these errors appropriately:
```

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Token revocation
+
+When a user wants to log out from a client, the client should revoke either its access token or refresh token by making a POST request to the revocation endpoint as described in [RFC7009].
+
+The server must revoke both the access token and refresh token associated with the token provided in the request.
+
+The request includes:
+- The `token` parameter containing either the access token or refresh token to revoke
+- Optionally, the `token_type_hint` parameter, with either the `access_token` or `refresh_token` value. If provided, the server must use this value to determine which token to revoke
```
dbkr, line 25: Didn't it say above that it revokes both?

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,100 @@
...
+HTTP/1.1 200 OK
+```
+
+### Handling errors
+
+The server may return an error response as defined in [RFC7009]. The client should handle these errors appropriately:
+
+- If the token is already revoked, the server returns a 200 OK response
+- If the client is not authorized to revoke the token, the server returns a 401 Unauthorized response
+- For other errors, the server returns a 400 Bad Request response with error details
```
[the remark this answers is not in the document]
tonkku107: From RFC7009:
> The authorization server responds with HTTP status code 200 if the
   token has been revoked successfully or if the client submitted an
   invalid token.
 
 So it should be status 200 for invalid tokens.
 
 Maybe this change would clarify
 ```suggestion
- If the token is already revoked or invalid, the server returns a 200 OK response
- If the client is not authorized to revoke the token, the server returns a 401 Unauthorized response
- For other errors, the server returns a 400 Bad Request response with error details
```

File: proposals/4254-oauth2-revocation.md [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+### Token revocation
+
+When a user wants to log out from a client, the client should revoke either its access token or refresh token by making a POST request to the revocation endpoint as described in [RFC7009].
+
+The server must revoke both the access token and refresh token associated with the token provided in the request.
+
+The request includes:
+- The `token` parameter containing either the access token or refresh token to revoke
+- Optionally, the `token_type_hint` parameter, with either the `access_token` or `refresh_token` value. If provided, the server must use this value to determine which token to revoke
+- The `client_id` obtained during client registration
```
FSG-Cat, line 26: So after consulting the auth room it makes sense to make client id optional for access tokens atleast. They are live without this information and therefore it makes sense to enable revokation without it.

tonkku107: Specifically to allow revoking a leaked access token which could be used maliciously without a client id, making it harder to do the right thing than the bad thing.

proposals/4254-oauth2-revocation.md
morguldir: I think this specifies which type the token is, so if `token` is an access token and you specified a refresh token hint, it won't work, but if you specified access token, the request will go through and both will be revoked

sandhose: Clarified in 797d26eb

proposals/4254-oauth2-revocation.md
sandhose: I've clarified this in d78fa2a1. wdyt?

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,100 @@
...
+
+```http
+HTTP/1.1 200 OK
+```
+
+### Handling errors
+
+The server may return an error response as defined in [RFC7009]. The client should handle these errors appropriately:
+
+- If the token is already revoked, the server returns a 200 OK response
```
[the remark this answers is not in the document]
sandhose: Changed this to 'revoked or invalid', so that doesn't imply tracking of old tokens

proposals/4254-oauth2-revocation.md
FSG-Cat: I think the commit referenced solves my exact concern.

proposals/4254-oauth2-revocation.md
```diff
@@ -16,14 +16,17 @@ The discovery of the above metadata is out of scope for this MSC, and is current
...
-- The `token` parameter containing either the access token or refresh token to revoke
-- Optionally, the `token_type_hint` parameter, with either the `access_token` or `refresh_token` value. If provided, the server must use this value to determine which token to revoke
-- The `client_id` obtained during client registration
+The request includes the following parameters, encoded as `application/x-www-form-urlencoded`:
+
+- `token`: This parameter MUST contain either the access token or the refresh token to be revoked.
+- `token_type_hint`: This parameter is OPTIONAL, and if present, MUST have a value of either `access_token` or `refresh_token`.  The server MAY use this value to optimize the token lookup process
+- `client_id`: The client identifier obtained during client registration.
+
+If the `client_id` is not provided, or does not match the client associated with the token, the server SHOULD still revoke the token. The server MAY also warn the user that one of their sessions may be compromised in this scenario.
```
richvdh, line 29: I'm not entirely following why a server would want to warn if one of the sessions is compromised if no `client_id` is provided. A sentence or two of rationale might help here.

sandhose: I've explained that better in cd239ce5

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,107 @@
...
+
+### OpenID Connect RP-Initiated Logout
+
+OpenID Connect defines a [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) specification that allows clients to initiate a logout through a browser redirect. This would:
+
+1. Allow the server to clear browser session state
+2. Support single logout across multiple clients
+3. Give visual feedback to the user about the logout process
+
+However, this approach requires a browser redirect which may not be desirable for all clients, especially mobile platforms.
```
ara4n, line 95: I'm not following why a browser redirect would ever be a problem (especially on mobile)?

The main proposal made sense when i was reading it, but then on hitting this RP-Initiated Logout alternative, I thought "huh, but these advantages sound great - why aren't we doing this then?", given I don't understand the disadvantage.

sandhose: I think it's relatively unexpected from a user perspective to be bounced through the browser for logging out. This can be relatively transparent for web clients with a redirect, but for native clients, it means opening a browser window, switching context; and there is no defined way to get back to the client after the RP-initiated logout, so you end up outside the client at the end of the logout

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,107 @@
+# MSC4254: Usage of [RFC7009] Token Revocation for Matrix client logout
```
erikjohnston, line 1: Can you add some words about how this relates to `/logout` please? 

Some questions I have:

1. Does this supersede `/logout`? i.e. if you login via OAuth2 you should use this API?
2. What happens if you call `/logout` having logged into OAuth2?
3. Does this replaces `/logout/all`?

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,107 @@
...
+### Token revocation
+
+When a user wants to log out from a client, the client SHOULD revoke either its access token or refresh token by making a POST request to the revocation endpoint as described in [RFC7009].
+
+The server MUST revoke both the access token and refresh token associated with the token provided in the request.
+
+The request includes the following parameters, encoded as `application/x-www-form-urlencoded`:
+
+- `token`: This parameter MUST contain either the access token or the refresh token to be revoked.
+- `token_type_hint`: This parameter is OPTIONAL, and if present, MUST have a value of either `access_token` or `refresh_token`.  The server MAY use this value to optimize the token lookup process
```
erikjohnston, line 26: Is this something that is part of OAuth2? It feels to me fairly pointless, and a potential source of confusion if a client manages to pass in an access token but call it a refresh token or vice versa.

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,107 @@
...
+- `client_id`: The client identifier obtained during client registration.
+
+If the `client_id` is not provided, or does not match the client associated with the token, the server SHOULD still revoke the token.
+This behavior is meant to help good actors like secret scanning tools to proactively revoke leaked tokens.
+The server MAY also warn the user that one of their sessions may be compromised in this scenario.
```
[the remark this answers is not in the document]
sandhose: Same, part of the RFC7009

proposals/4254-oauth2-revocation.md
sandhose: This is a very good point, clarified in ef03a25c

proposals/4254-oauth2-revocation.md
ara4n: hm, okay. can you add the rationale to the MSC please? :)

proposals/4254-oauth2-revocation.md
```diff
@@ -0,0 +1,121 @@
...
+
+The client should handle these errors appropriately:
+
+- If the token is already revoked or invalid, the server returns a 200 OK response
+- If the client is not authorized to revoke the token, the server returns a 401 Unauthorized response
+- For other errors, the server returns a 400 Bad Request response with error details
+
+### Replacement of existing APIs
+
+This proposal replaces the existing [`/_matrix/client/v3/logout`] endpoint for [MSC861]-compatible client.
```
anoadragon453, line 79:
```suggestion
This proposal replaces the existing [`/_matrix/client/v3/logout`] endpoint for [MSC3861]-compatible clients.
```

zecakeh:
spec PR: https://github.com/matrix-org/matrix-spec/pull/2151

richvdh:
> spec PR: [matrix-org/matrix-spec#2151](https://github.com/matrix-org/matrix-spec/pull/2151)

merged! :tada:

noelportillo (commented):
Very nice work

PR #4256, opened 2025-01-24
MSC: RFC 9420 MLS mode Matrix

dklimpel:
[Rendered](https://github.com/dklimpel/matrix-spec-proposals/blob/mls-RFC-9420/proposals/0000-rfc9420-mls-mode-matrix.md)


proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+The invited user will then receive the invite and welcome message. If the user wants to join the
+room, the client can send a request to the /join endpoint.
+
+As no member events are used in this room version, the direct /state endpoint can not be used to
+change membership in a room
+
+![Invite flow][invite-flow]
+
+## Potential issues
```
benkuly, line 510: What about display name and avatar url of members? Are they stored in the MLS group context too? And if so, how to prevent malicious users from changing them.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+Most state events from other room versions are now invalid. This requires major changes to the
+authorization rules. The mls commit event now defines what servers are allowed to send events into a
+room as well as who is allowed to modify the commit event. Additionally unencrypted content is
+generally disallowed in this room version and the create event defines the encryption algorithm now.
+The below rules describe such behaviour and if there is the option to decide between stricter and
+less strict validation, those rules tend to favour stricter validation. Some of those rules could be
+made more lenient to allow more extensibility if such a need is expected.
+
+1. if state_key is not empty, reject
```
Johennes, line 202: Maybe splitting hairs but empty and missing are considered equivalent here, right?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+   4. If `servers` is empty or any of the entries are not a valid servername, reject
+   5. If `powers` is empty or any of the entries are not a valid servername, reject
+   6. If the `origin` is not listed in `powers` (of the previous `m.mls.commit` event), reject
+   7. If the `powers` has entries not in `servers`, reject
+   8. If the `powers` start not with exactly the same entries in the same order as the subset of
+entries in the previous `powers` above the `origin` , reject
+   9. If the `can_propose` has entries not in `servers`, reject
+   10. Otherwise, allow
+10. If the event type is `m.mls.pending_commit`:
+    1. If the epoch of this event is not exactly the epoch of the previous `m.mls.commit` event + 1, reject
```
Johennes, line 239: "the _current_ `m.mls.commit` event" might be clearer here since `m.mls.pending_commit` doesn't directly replace that event. This could maybe also be applied to 9.2.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+   2. If the epoch of this event is not exactly the epoch of the previous `m.mls.commit` event + 1,
+reject
+   3. If the MLS content is not a commit, reject
+   4. If `servers` is empty or any of the entries are not a valid servername, reject
+   5. If `powers` is empty or any of the entries are not a valid servername, reject
+   6. If the `origin` is not listed in `powers` (of the previous `m.mls.commit` event), reject
+   7. If the `powers` has entries not in `servers`, reject
+   8. If the `powers` start not with exactly the same entries in the same order as the subset of
+entries in the previous `powers` above the `origin` , reject
+   9. If the `can_propose` has entries not in `servers`, reject
```
Johennes, line 236: Would we need to enforce that entries from `powers` also appear in `can_propose` to enable the process described in "Commit handling with multiple servers" below?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+Events are encrypted using plain MLS and sent into a room as m.room.encrypted events. The sender can
+be retrieved using MLS, which allows mapping a decrypted MLS message to a credential.
+
+The encrypted payload should be:
+
+```json
+{
+   "type": "m.room.encrypted",
+   "content": {
+      "algorithm": "m.megolm.v1.aes-sha2",
```
Johennes, line 382: Should this not be something starting with `m.mls.v1`?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+Some of these issues are already explained in the proposal sections. The following list provides
+them and other issues together to allow an easier overview.
+
+### Redactions
+
+Currently redactions are not supported in the outer protocol layer. This has the benefit of making
+redactions invisible to the server, but also prevents any server side aggregation. This also means
+content can’t be deleted. As we require all events to be encrypted, this might not be a major
+problem, but policy makers might disagree. Redactions could be supported in various ways and we
```
Johennes, line 520: Redactions are not only relevant in the context of data privacy but also for moderation. Not being able to delete malicious or illegal content seems like a significant problem. 😕

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+for the room. This would be the MLS cipherset prefixed with `m.mls.v1.` for example
+"m.mls.v1.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519". All future events in the room need to
+have the same algorithm set. To change the algorithm a new room needs to be created possibly via the
+room upgrade endpoint.
+
+A new state event is introduced to track the current MLS commit. This event follows the format for
+encrypted events with only an `algorithm` and `ciphertext` field. It contains the current MLS
+commit as a MLS private message.
+
+The MLS commit contains a CBOR encoded field in its unencrypted authenticated data of the MLS
```
uhoreg, line 149: Can you give some rationale here for why you're using CBOR?

proposals/4256-rfc9420-mls-mode-matrix.md
uhoreg: Yes, it probably needs to be reworded.  Events with a missing `state_key` have to be able to pass this rule.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+   1. If it has any `prev_events` or `auth_events` , reject
+   2. If the room version is not an mls version, reject (or it should have been a different
+algorithm for that room version!)
+   3. If `origin` does not match the domain of the `room_id` , reject
+   4. If the `algorithm` is not present or empty, reject
+   5. Otherwise, allow
+4. Considering the events auth events:
+   1. If there is not exactly one `m.room.create` event (with implicitly an empty state key), reject
+   2. If the event is not the `m.room.create` event (we never get here) or an `m.mls.commit` event
+with epoch == 0 (we do get here), then if there is no `m.mls.commit` event in the auth events,
```
uhoreg, line 214: I don't see anywhere in this MSC that mentions where the epoch comes from.  I think it may be helpful to mention something about that somewhere, since this MSC will be read by people not completely familiar with MLS.  (It comes from the `ciphertext`, which despite being called "ciphertext", contains some cleartext bits, including the epoch.  Which means that servers will be expected to be able to parse the MLS ciphertext, at least enough to read the epoch, group ID, and anything else that's needed.)

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+reject (could move this check to a different section)
+   3. If any additional auth events are present, reject
+   4. If any auth event did not pass the checks done on PDU receipt, reject
+5. If the `content` of the `m.room.create` event in the room state has the property `m.federate` set
+to `false`, and the `origin` of the event does not match the `origin` domain of the create event,
+reject.
+6. If the `origin` is not listed in the `servers` of the previous `m.mls.commit` event (unless it is
```
uhoreg, line 221: I think you need to explain what "previous" means here (presumably, the `m.mls.commit` event that is listed in the event's `auth_events`).

uhoreg (commented):
I've only read through part of this, but here are some (mostly nitpicky) comments so far.  Generally, there's inconsistency between using "mls" and "MLS".  It would be nice to use "MLS" everywhere.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+keys for specific transitions, that are only accessible to specific users, that a server could then
+validate without understanding the content of events. For example specific signatures to be able to
+send invites and allow other users to gain access to the encrypted messages of a room. Other
+approaches might be possible by leveraging some features of MLS more directly.
+
+This feature is left for future MSCs to resolve.
+
+### Room state
+
+Currently all state is stored in one GroupContext extension. This has obvious size limitations.
```
uhoreg, line 569: Just to be clear, I believe that this means that having the state in the GroupContext means that the entire state will be encoded in every single commit.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+MLS requires the delivery service to resolve conflicting commits. This requires some form of global
+synchronization of the system, which is difficult in a decentralised setting.
+
+The MLS protocol always requires server-side approval of a commit before it can be applied by
+clients. Therefore clients need to wait for a commit to appear in the sync response before applying
+it.
+
+There are several options for how a client can send a new commit into a room.
+
+* Users on the first server listed in the "powers" list can send commit state events directly. They
```
uhoreg, line 286: What happens if multiple people on that first server send a commit state event at the same time?  Does he server just select one to accept and return an error to the others?  If so, what error code does it use?

proposals/4256-rfc9420-mls-mode-matrix.md
uhoreg, line 510: I think that the possible delay in applying a commit due to "powers" servers being offline is a significant potential issue that should be listed here.  For example, if the first server listed in "powers" is offline, then when you invite someone, it will take 5 minutes before they can actually be in the room.  (Could they end up being halfway in the room, if the inviter closes their client after they try to invite, but before they see the commit event?)

This, combined with encrypted state, may also cause problems for MatrixRTC, which IIRC uses various state events.  If it takes over 5 minutes to set up a call, or for someone to join a call, that could be problematic.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+value is not 0, a client is allowed to keep the old private keys required to apply a different
+commit around for up to this many seconds. This allows a client to recover automatically from state
+conflicts happening during a very small period.
+
+Having conflicting commits will lead to some undecryptable messages. For short durations that can be
+mitigated by keeping the old state around for seconds to minutes. This is the common case when
+people invite new users at the same time or do otherwise racing modifications on different servers.
+For long durations no automatic recovery is possible and manual communication is necessary. We
+expect those cases to be rare even in larger federations. Long term server outages usually don’t
+include membership modifications at the same time as usually one server is completely offline. Even
```
uhoreg, line 348: Membership modifications aren't the only reasons for sending commits, though.  Group members are expected to periodically send commits (or Updates which are committed by others), which will probably be more frequent than membership changes.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+### Adding users to a room
+
+To invite users into a room the client should first fetch the key packages for that user and add the
+user to the MLS group. The key packages should be validated and should be cross-signed as configured
+in the additional authenticated data of the current MLS commit.
+After this the commit in the room is updated. Once this commit is accepted by the delivery service,
+the inviting user should send out the welcome message as a to_device message and send an invite
+using the /invite endpoint. To distribute the ratchet tree the ratchet tree MLS extension is
+currently required.
```
uhoreg, line 500: In large groups, the `Welcome` message can get quite large, so it may not be appropriate to send the whole thing as a to-device message.  In my MSC2883 work, I put large `Welcome` messages in the media repository, and sent a reference to the media repository as a to-device message.  This has some advantages (e.g. the client can download the large message when it wants to, rather than have it clog up the `/sync`, and the message doesn't need to be base64-encoded).  But it has some other problems (e.g. we don't have a way to delete the media when it's no longer needed, the sender's server could be down when the recipient wants to download it).  We may need to look into some better way of sending large blobs between devices, as this is an issue that affects all MLS-related MSCs, as well as possibly some other things that we're working on.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+}
+```
+
+### Adding users to a room
+
+To invite users into a room the client should first fetch the key packages for that user and add the
+user to the MLS group. The key packages should be validated and should be cross-signed as configured
+in the additional authenticated data of the current MLS commit.
+After this the commit in the room is updated. Once this commit is accepted by the delivery service,
+the inviting user should send out the welcome message as a to_device message and send an invite
```
uhoreg, line 498: I don't see any definition of how the Welcome message is sent as a to-device event (e.g. what event type it uses, what properties the event has).  Can you add that?

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have updated it.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+room as well as who is allowed to modify the commit event. Additionally unencrypted content is
+generally disallowed in this room version and the create event defines the encryption algorithm now.
+The below rules describe such behaviour and if there is the option to decide between stricter and
+less strict validation, those rules tend to favour stricter validation. Some of those rules could be
+made more lenient to allow more extensibility if such a need is expected.
+
+1. if state_key is not empty, reject
+2. if state_key is present and event type is not `m.mls.commit` or `m.room.create`, reject
+3. if the event is `m.room.create` :
+   1. If it has any `prev_events` or `auth_events` , reject
```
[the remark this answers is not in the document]
dklimpel: I have added the rule.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+   3. If any additional auth events are present, reject
+   4. If any auth event did not pass the checks done on PDU receipt, reject
+5. If the `content` of the `m.room.create` event in the room state has the property `m.federate` set
+to `false`, and the `origin` of the event does not match the `origin` domain of the create event,
+reject.
+6. If the `origin` is not listed in the `servers` of the previous `m.mls.commit` event (unless it is
+the first commit), reject
+7. If the room_id or the MLS group_id do not match the room_id of the create event, reject
+8. If the algorithm of the event does not match the create event, reject
+9. If the event type is `m.mls.commit`:
```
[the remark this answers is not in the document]
dklimpel: Yes, I have added this.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Yes.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+allows an attacker to access possibly all past messages. There are various ways message
+synchronization and backup could be supported by introducing an additional layer of indirection.
+This can be based on exporters from the MLS key schedule. But this will require a new protocol with
+a separate security analysis and is left out of this proposal.
+
+### Encrypted state events
+
+This proposal provides no support for generic state events outside of the MLS layer. State events
+inside the MLS layer are agreed upon using a GroupContext extension. The extension id for the group
+state should be 0xF6D0. 0xF000 to 0xFFFF is reserved for private extensions, 6D is a lowercase M in
```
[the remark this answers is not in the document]
dklimpel: Thanks for it.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+}
+```
+
+### Adding users to a room
+
+To invite user's devices into a room the client should first fetch the key packages for that user and add the
+user to the MLS group. The key packages should be validated and should be cross-signed as configured
+in the additional authenticated data of the current MLS commit.
+After this the commit in the room is updated. Once this commit is accepted by the delivery service,
+the inviting user should send out the welcome message as a to_device message and send an invite
```
richvdh, line 500: What would be the impact if the invite was not sent, for example due to a buggy or malicious client, or due to a network error? It seems like the MLS membership list and Network membership list would get out of sync, which could be problematic?

richvdh: Relatedly, is there any risk of a buggy client sending an invite without having sent a welcome message?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+using the /invite endpoint. To distribute the ratchet tree the ratchet tree MLS extension is
+currently required.
+
+The invited user will then receive the invite and welcome message. If the user wants to join the
+room, the client can send a request to the /join endpoint.
+
+As no member events are used in this room version, the direct /state endpoint can not be used to
```
richvdh, line 507: Given that member events are not used, how does network membership (invites, joins, etc) propagate between the homeservers that are participating in the room?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+used for state resolution. This replaces the existing approach of ordering by power level and
+provides a reliable tie break, since no server can have the same index as another server. There are
+specific rules around promotions in the powers list to provide similar guarantees about demoting
+more powerful users/servers as the power levels event did. Neither of those lists can be empty.
+`security.window` specifies the seconds of how long old private keys might be kept to recover from
+mls commit conflicts. Optional, defaults to 0\. Clients should not allow values larger than 7 days
+and should clamp any larger values to that maximum.
+`security.required_verification` specifies the verification required to add new users or clients
+to the MLS state. The sender of such a commit needs to have at least this level of verification with
+the other participant to add them. Possible values are m.direct, m.cross, m.tofu and m.none.
```
richvdh, line 178: I can guess what most of the values mean, but what does m.direct mean?

richvdh: Also, is this setting enforced by the protocol somehow, or is it intended as an indication to clients about how they SHOULD behave?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+servers. Instead the commit is optimistically applied by one homeserver and forwarded to other
+homeservers via federation. If there is a conflict, this is resolved by state resolution. Clients
+are then notified via the sync API about changes in the state they expect.
+
+Only the highest power server may send `m.mls.commit` events directly. Clients on other servers
+should send `m.mls.proposed_commit` events instead. These may then be replayed by servers in the
+"powers" list.
+
+Servers have a 5 minute conflict window to apply `m.mls.proposed_commit` they receive. The
+highest power server can apply the commit in the first 3 minutes after the `origin_server_ts` of
```
richvdh, line 326: Is this not subject to manipulation by a server wishing to put itself in charge of applying commits, by falsifying origin_server_ts?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+the `m.mls.proposed_commit` event. After this a 2 minute grace period follows to allow for some
+clock desynchronization latencies as well as message transfer. The next 5 minutes are for the second
+server listed in "powers" and so on. If the commit referenced by the auth events of a
+`m.mls.proposed_commit` is not the current commit, a commit has already been applied and servers
+should ignore the `m.mls.proposed_commit`.
+
+Clients can know their `m.mls.proposed_commit` failed to be applied if they haven’t received a
+`m.mls.commit` event matching their `m.mls.proposed_commit` after "len(powers)*5" minutes.
+
+State resolution might result in some clients losing access to messages as they threw their private
```
richvdh, line 336: My understanding of this section is that a netsplit of any non-trivial duration can lead to different commits being applied on each side of the split; when the split is eventually resolved, this will result in one side "losing" and any clients on the losing side will be effectively excluded from the room, since they will have discarded the necessary MLS state.

If that's correct, it feels akin to the ["state reset"](https://github.com/element-hq/synapse/issues/8629) problems that current Matrix can suffer from, but much worse. 

This seems very dangerous to me. IMHO it would be better not to allow MLS commits outside the current highest power server at all.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,803 @@
...
+some additional filtering based on key package capabilities that this proposal explicitly left for a
+future MSC.
+
+In other areas the proposals have some significant differences. Detailed comparison is difficult at
+this time as both proposals are still in progress and [MSC4244][MSC4244] includes a lot of open "TODOs":
+
+#### Membership handling
+
+This proposal does not use membership events and gets rid of the "sender" field in events. This
+essentially makes membership opaque to the homeserver and gets Matrix a lot closer to portable
```
richvdh, line 636: I am unclear how membership can be opaque to the server. Doesn't the server still need to know a membership list to distribute events to clients?

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: JSON would have some overhead when putting it directly into the bytes here, so CBOR seemed like the more natural fit, but it isn't really a requirement. We can change that.

Currently there are no other fields in the authenticated data. There is a way to structure extensions using authenticated data, that we aren't using here yet, which could be more appropriate? But so far we decided to go with the simplest approach.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+### State events
+
+This room version removes all existing state events apart from the `m.room.create` event from the
+unencrypted room state. Some state events like `m.room.name` can still be sent inside of the MLS
+encryption layer, but they are not visible to the server. This section focuses on the server visible
+state events, see "Encrypted state events" for state events inside of the MLS layer.
+
+Every valid state event in this room version is required to have an empty state key.
+
+The `m.room.create` event has a mandatory `algorithm` field now to set the encryption algorithm
```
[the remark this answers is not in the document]
dklimpel: I have adapted it.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+disconnected from the other servers in the room for an arbitrary extent of time and the room will
+recover once the connection is restored. This is achieved by using a DAG of messages, that allows
+temporary splits to merge together using state resolution once a connection is restored including
+arbitrary state modifications.
+
+Previous proposals have tried to unify those aspects of MLS and Matrix, but often this has come at
+the expense of some of the security benefits of MLS. At its core having a coherent set of members in
+a room and the ability to change membership at any time without the ability to broadcast these
+changes to other servers in the room are at conflict. If any server can modify historical membership
+of a room, this compromises Post-Compromise Security (PFS). In an earlier proposal, decentralised
```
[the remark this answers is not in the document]
dklimpel: Yes, that was a mistake. I reduced it to "security".

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have changed it to "current".

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Thank you, there will be a section on this here.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+GET /_matrix/client/v3/keys/list
+{
+    "one_time_keys": ["m.mls.v1.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519:AAAAAAAAAAAA"],
+    "fallback_keys": ["m.mls.v1.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519:AAAAAAAAAAAB"]
+}
+```
+
+### Adding users to a room
+
+To invite users into a room the client should first fetch the key packages for that user and add the
```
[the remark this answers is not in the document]
dklimpel: I have updated this.

proposals/images/4256-invite-flow.png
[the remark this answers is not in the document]
dklimpel: The image will be updated.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Technically that is optional, since events can either be sent as a proposed commit or directly. Ideally every server can propose though, although the first power server really doesn't need to.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Yep, exactly what they said and we should put that somewhere

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have added a note.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: That's the regular MLS conflict case. The server selects one based on some criteria (e.g. the one that came first) and sends a rejection message to sender of the other commit.
That needs of course a message. But I don't think we defined the error messages/codes for any of the failure cases yet.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+The MLS protocol always requires server-side approval of a commit before it can be applied by
+clients. Therefore clients need to wait for a commit to appear in the sync response before applying
+it.
+
+There are several options for how a client can send a new commit into a room.
+
+* Users on the first server listed in the "powers" list can send commit state events directly. They
+still have to wait for the commit to sync back to them, before they can consider the commit
+accepted.
+* Other servers in the "powers" list could in theory send commits directly (from an auth rule
```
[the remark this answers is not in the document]
dklimpel: Technically it might be useful to be able to send a commit also on other servers if the primary power server is down for example. This needs more implementation experience to see, if such a mechanism is necessary or we should reject such events on the server side.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: In that case the higher power server could still send a conflicting event, which would win during state.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+Servers have a 5 minute conflict window to apply `m.mls.proposed_commit` they receive. The
+highest power server can apply the commit in the first 3 minutes after the `origin_server_ts` of
+the `m.mls.proposed_commit` event. After this a 2 minute grace period follows to allow for some
+clock desynchronization latencies as well as message transfer. The next 5 minutes are for the second
+server listed in "powers" and so on. If the commit referenced by the auth events of a
+`m.mls.proposed_commit` is not the current commit, a commit has already been applied and servers
+should ignore the `m.mls.proposed_commit`.
+
+Clients can know their `m.mls.proposed_commit` failed to be applied if they haven’t received a
+`m.mls.commit` event matching their `m.mls.proposed_commit` after "len(powers)*5" minutes.
```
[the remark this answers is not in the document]
dklimpel: Yes.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Yes, that is expected. If you want to prevent such conflicts, you should only put a single server into the power list, which achieves the same result as only allowing commits from the highest power server. The option to add more servers in the powers list only exists for the cases where people expect servers to go offline eventually (like in smaller communities, where a friend decides to quit Matrix).

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+State resolution might result in some clients losing access to messages as they threw their private
+state away immediately to provide PCS and forward secrecy. In that case the clients dropped by the
+state event will have to be sent another welcome to rejoin them into the room.
```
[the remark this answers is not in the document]
dklimpel: A client can be re-added by reinviting them. Usually there is no way to know this for certain, but clients can likely guess a client got dropped from the room by looking at the results of state resolution. But that is unreliably to do over federation, as you can't always know what fork a room ended up on on that server.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+
+The duration of this temporary compromise is specified in the security.window in seconds. If this
```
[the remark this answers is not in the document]
dklimpel: Any security window is not secure. The security guidance here is not to use it.
If it is used, it should be as small as possible.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+value is not 0, a client is allowed to keep the old private keys required to apply a different
+commit around for up to this many seconds. This allows a client to recover automatically from state
+conflicts happening during a very small period.
+
+Having conflicting commits will lead to some undecryptable messages. For short durations that can be
+mitigated by keeping the old state around for seconds to minutes. This is the common case when
+people invite new users at the same time or do otherwise racing modifications on different servers.
+For long durations no automatic recovery is possible and manual communication is necessary. We
```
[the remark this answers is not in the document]
dklimpel: Yes, the users must exchange information out of band. This can be by email, face to face or another intact or private room or similar.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: There's no requirement to regularly send commits.
However, it's recommended for security. But that's a membership modification as well because updating the own leaf nodes changes the leaf node and such the membership.
But maybe it should be made clear here that any change of leaf nodes in the tree is meant.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: They are just normal encrypted events:

```json
{
  "type": "m.room.encrypted",
  "content": {
    "algorithm": "m.mls.v1....",
    "ciphertext": "<encrypted_payload_unpadded_base_64>"
  }
}
```

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Only the local server needs to know which of its users are in the room. Invites and joins are handled using the existing federation APIs for invites and joins, but other servers never get told about those users, they only know which servers are participating.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have added a new paragraph: `Size of Welcome Messages`

> Welcome messages can become pretty big when they contain the ratchet tree. To reduce the message
> size of welcome messages, MLS allows distributing the ratchet tree (which is the largest part in
> the message) in other ways. Future MSCs can define other delivery mechanisms for ratchet trees
> when the size of the welcome messages becomes an issue.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+the inviting user should send out the welcome message as a to_device message and send an invite
+using the /invite endpoint. To distribute the ratchet tree the ratchet tree MLS extension is
+currently required.
+
+The invited user will then receive the invite and welcome message. If the user wants to join the
+room, the client can send a request to the /join endpoint.
+
+As no member events are used in this room version, the direct /state endpoint can not be used to
+change membership in a room
+
```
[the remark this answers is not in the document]
dklimpel: Users can add their own devices as part of the verification process during login or other users might propose adding them, when they notice the new device and can verify it.

For leaving users can propose a self removal (if that extension is enabled) or ask someone else to remove them from the group via normal communication in the room.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: This MSC expects that the servers in the powers list are online.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: Yes, but that's a known issue and can be resolved with MLS extensions. Either a custom extension that only sends a diff, or by using mechanisms from the extensions draft.

However if there is no state change, no proposal for that needs to be sent, so it also isn't included in the commit.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+## Alternatives
+
+### MSC2883: Matrix-flavoured MLS
+
+[MSC2883][MSC2883] provides MLS support over Matrix by making extensive changes to MLS (dMLS). Those changes
+compromise Forward Secrecy by making it impossible to delete past private keys. Such a change does
+not provide the security guarantees we are looking for and is incompatible with MLS. If state
+resolution allows an arbitrary long amount of time between one server changing membership and other
+servers being made aware of that change, secret keys to resolve those commits have to be kept around
+for the same amount of time.
```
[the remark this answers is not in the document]
dklimpel: Our goal currently is to use MLS as is. I do think there is some value in finding a more Matrix way to do MLS in the future, but until we can make sure we don't have severe compromises because of that, we think it is a good idea to stick to standard MLS.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+Every valid state event in this room version is required to have an empty state key.
+
+The `m.room.create` event has a mandatory `algorithm` field now to set the encryption algorithm
+for the room. This would be the MLS cipherset prefixed with `m.mls.v1.` for example
+"m.mls.v1.MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519". All future events in the room need to
+have the same algorithm set. To change the algorithm a new room needs to be created possibly via the
+room upgrade endpoint.
+
+A new state event is introduced to track the current MLS commit. This event follows the format for
+encrypted events with only an `algorithm` and `ciphertext` field. It contains the current MLS
```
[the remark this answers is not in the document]
dklimpel: I have added a note.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have adapted the paragraph and defined a default.

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: As mentioned previously in the proposal the server does know about its local users currently, but not about room members on other servers. We amended this paragraph to include that info, since it was only mentioned elsewhere in the proposal before.

proposals/4256-rfc9420-mls-mode-matrix.md
richvdh: ok, but how is the list of participating servers kept in sync with the network membership list?

proposals/4256-rfc9420-mls-mode-matrix.md
richvdh: ok, so a client that receives a welcome needs to wait a while to see if an invite arrives? How long should it wait?

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,822 @@
...
+
+Keys, credentials, signatures, room version and event types from this proposal should use the prefix
+"de.bwi".
+
+As such credentials, key packages and signatures should be prefixed with "de.bwi.mls.v1" instead of
+"m.mls.v1".
+
+The room version should be "de.bwi.mls.v1".
+
+The mls commit event should be "de.bwi.mls.pending_commit".
```
ara4n, line 807: (and `de.bwi.mls.commit` presumably)

proposals/4256-rfc9420-mls-mode-matrix.md
dklimpel: I have added a section for displayname & avatars

proposals/4256-rfc9420-mls-mode-matrix.md
kegsay: If the CBOR is just the JSON object below then you aren't going to save many bytes here, given all the string keys / values need to be represented as-is.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,836 @@
...
+guarantees and as such  should only be used and implemented after a rigorous security analysis on
+the exact impact.
+
+### Verification chains
+
+This proposal relies on cross-signing to add new members and devices. Such a structure might not
+provide sufficient security guarantees. In megolm each sender can decide who can decrypt a message.
+In MLS this is a shared state and relies much more heavily on each individual to provide the same
+trust guarantees. For this reason the minimum requirement is spelled out in the authenticated data
+of the commit events, but any client can in theory violate this requirement. Stricter rules around
```
kegsay, line 573: There is a philosophical aspect here which is worth pointing out.

Design wise, Olm/Megolm put absolute control on who can decrypt a given message into the hands of the _sender_, this is alluded to a few lines above: "In megolm each sender can decide who can decrypt a message.".  In contrast, MLS puts absolute control on who can decrypt a given message into the hands of the _group_.

This becomes problematic when you put this design into a featureful environment like Matrix, which has group access control. It's problematic because certain scenarios require you to subvert MLS to remain secure. Consider the scenario where you are a normal member in a company internal room and become aware that Bob has been compromised. Taking immediate action is critical to avoid leaking information. However, I cannot remove Bob from the rooms I am in because I am not a moderator/admin. As a result, I cannot communicate in those rooms without automatically providing decryption keys to Bob, unless I subvert MLS and purposefully exclude certain keys from the group. This breaks the MLS philosophy of "everyone knows who can decrypt which messages".

Personally, I think having absolute control over who I encrypt messages for is important. MLS makes tradeoffs here which risks leaking information. It would be nice if any MLS implementation preserves the ability to exclude specific keys.

proposals/4256-rfc9420-mls-mode-matrix.md
```diff
@@ -0,0 +1,801 @@
...
+10. If the event type is `m.mls.pending_commit`:
+    1. If the epoch of this event is not exactly the epoch of the previous `m.mls.commit` event + 1, reject
+    2. If the MLS content is not a commit, reject
+    3. If `servers` is empty or any of the entries are not a valid servername, reject
+    4. If `powers` is not the same as the current `m.mls.commit`, reject
+    5. If `origin` is not listed in `can_propose` of the current `m.mls.commit`event, reject
+    6. If the `powers` has entries not in `servers`, reject
+    7. If the `can_propose` has entries not in `servers`, reject
+    8. Otherwise, allow
+11. If the epoch of the `ciphertext` is not the epoch of the current `m.mls.commit` event + 1,
```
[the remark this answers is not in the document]
franziskuskiefer: First I think it's important to clarify what the different pieces are that we talk about.

For MLS we have a `Commit`, which is part of a `PrivateMessage`. On top of that, we have the Matrix event. Note that the `Commit` may also be inside a `PublicMessage`. We didn't do that distinction here but assumed `PrivateMessage`s for now. Hence the term `ciphertext`.

The `PrivateMessage` has an epoch field as part of the public metadata. This, as well as all the framing metadata (encrypted) uses the current epoch the message is being sent in.
The `Commit` that's inside the message however, triggers the move to the next epoch. As such, the commit is built and authenticated with the new epoch (which is only implicit though, there's no epoch field).

I think that's where the confusion may be coming from. The outer epoch of the message is the old epoch n, but the commit itself is built for n+1. So the text here seems to be correct, because the epoch of the event is the epoch of the message framing.

proposals/4256-rfc9420-mls-mode-matrix.md
nyabinary: cc: @dklimpel any plans to resolve this?

nyabinary:
Is this still being worked on?

PR #4258, opened 2025-01-29
MSC4258: Federated User Directory

MatMaul:
[Rendered](https://github.com/tchapgouv/matrix-spec-proposals/blob/fed-user-dir/proposals/4258-federated-user-directory.md)

This proposal has been thought and written by me and people listed below, all employed by the French state for the Tchap project.

@mcalinghee @odelcroi @yostyle @NicolasBuquet

### DIFF
--- a/proposals/4258-federated-user-directory.md
+++ b/proposals/4258-federated-user-directory.md
@@ -19,6 +19,7 @@ It would be authenticated and rate limited.
 #### Request
 ```json
 {
+  "requester": "@foo:origin.org",
   "limit": 10,
   "search_term": "foo"
 }
@@ -34,7 +35,6 @@ It would be authenticated and rate limited.
       "display_name": "Foo",
       "m.tz": "America/New_York",
       "user_id": "@foo:bar.com",
-      "visibility": "local",
     }
   ]
 }
@@ -42,7 +42,12 @@ It would be authenticated and rate limited.
 
 All profile fields (cf [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133)) should be returned here.
 
-On top of that, the server should include a `visibility` field that should mirror the user preference stored in the account data, cf `m.user_directory` definition later on.
+Search should be performed on all the profile fields too.
+
+Field `requester` should be verified by the requested server: since the request is signed, we know that it has been sent by a specific server, so we MUST check that the
+server name of the requester matches the requester server.
+
+`requester` field is useful to be able to filter the results on the requested server directly, depending of the visibility user preference (cf next section).
 
 When an user calls the client user search API, the server should send a federated user search request to all known servers or a subset of it. It would then receive the results and return them to the user.
 Servers must not forward this request to other servers and only return results known locally. This is to avoid infinite loop between servers knowing each other.
@@ -86,7 +91,9 @@ Example of the content of a `m.user_directory` account data:
 
 We propose to introduce a reactive mechanism to allow the server to stream new results to the client.
 
-For that we introduce a `search_token` to the request coming from the response of a previous search(`search_token`). A request containing a `search_token` will stall until new results are available to the server. If some more results are expected to be returned, it may include another `search_token`, and hence.
+For that we introduce a `search_token` to the request coming from the response of a previous search(`search_token`).
+A request containing a `search_token` will stall until new results are available to the server.
+If some more results are expected to be returned, it may include another `search_token`, and hence.
 
 `search_token` is optional within the request so proposed changes are retro-compatible.
 
@@ -115,8 +122,6 @@ This parameter is optional and default to remote.
 ```
 `search_token` :  is a unique identifier that means that more results can be retrieved by querying with this `search_token`. `limited` should be `true` when `search_token` is returned.
 
-The server should strip `visibility` from the results returned by other servers.
-
 ## Potential issues
 
 We may have requests lost or getting timeout from intermediary network equipment, especially since we are using some kind of long polling.
@@ -124,7 +129,9 @@ We think the fact that we use a `search_token` that changes on each request allo
 
 ## Alternatives
 
-We first thought about using an account data, however it has a big caveat: remote servers can't access it, hence remote servers will not be able to honor the visibility when trying to return remote users that are already visible locally to them.
+Previous version of this MSC was using a profile field to store the visibility setting of the user, and hence this setting was visible to everyone.
+By adding `requester` in the federation request, the target server is fully in control of enforcing user setting and we avoid leaking the setting.
+One drawback is that we can't cache the result per target server and results should be cached per user. It feels like an acceptable performance tradeoff.
 
 Rather than using a `search_token`, we could use a `search_id` that will be the same for all subsequent calls.
 This solution is less informative about the progression of the search from the server perspective, cf `Potential issues` section.
@@ -139,7 +146,7 @@ A malicious server could list all user matrix ids that are defined in `remote` o
 
 The federation search endpoint should be rate limited.
 
-We recommend to not answer for `search_term` with less than 3 characters like "a" or "at".
+We recommend to not answer for `search_term` with less than 3 characters like "a" or "at", except when an exact match is available.
 
 #### Trust & Safety recommendations
 
--- a/proposals/4258-federated-user-directory.md
+++ b/proposals/4258-federated-user-directory.md
@@ -90,7 +90,7 @@ For that we introduce a `search_token` to the request coming from the response o
 
 We propose to add a new field in the profile (MSC4133) `m.user_directory.visibility` to give the user the ability to control their visibility in the user directory.
 
-Differents values are possible :
+Different values are possible :
 - `hidden` : not visible to anyone
 - `local` : visible only to local homeserver users
 - `restricted`: visible to any user sharing a room with
@@ -108,7 +108,7 @@ If no value is provided (or it is null), the user hasn't set a preference and th
 
 ## Potential issues
 
-We may have requests lost or getting timeout from intermediary network equipments, especially since we are using some kind of long polling.
+We may have requests lost or getting timeout from intermediary network equipment, especially since we are using some kind of long polling.
 We think the fact that we use a `search_token` that changes on each request allow the server to track correctly if new search results were already received by the client or not.
 
 ## Alternatives
--- a/proposals/4258-federated-user-directory.md
+++ b/proposals/4258-federated-user-directory.md
@@ -1,0 +1,150 @@
+# MSC4258: Federated User Directory
+
+Currently user search can only be done locally, which would at best get a list of all users known to the server.
+
+This proposal aims at introducing a federation endpoint allowing servers to broadcast the search to the current federating servers and get results among all their known users.
+
+Improvement of the client API is also proposed to accommodate the fact that results will arrive asynchronously and to allow users to manage their visibility on search results. 
+
+
+## Proposal
+
+### Federation endpoint
+
+We first propose a new federation endpoint similar to the [current client API](https://spec.matrix.org/v1.12/client-server-api#post_matrixclientv3user_directorysearch).
+It would be authenticated and rate limited.
+
+#### `POST /_matrix/federation/v3/user_directory/search`
+
+#### Request
+```json
+{
+  "limit": 10,
+  "search_term": "foo"
+}
+```
+
+#### Response
+```json
+{
+  "limited": false,
+  "results": [
+    {
+      "avatar_url": "mxc://bar.com/foo",
+      "display_name": "Foo",
+      "m.user_directory.visibility": "local",
+      "user_id": "@foo:bar.com"
+    }
+  ]
+}
+```
+
+All profile fields (cf [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133)) should be returned here.
+
+When an user calls the client user search API, the server should send a federated user search request to all known servers. It would then receive the results and return them to the user.
+Servers must not forward this request to other servers and only return results known locally. This is to avoid infinite loop between servers knowing each other.
+
+However we have a problem here: we can't have expectation on when and even if servers will answer to the search request.
+
+We are hence proposing some changes to the client API to accommodate the need to have a way to stream new results to the client.
+
+Note that `m.user_directory.visibility` is defined further down this proposal.
+
+### Client endpoint changes
+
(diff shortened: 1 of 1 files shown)


MatMaul:
We are planning to implement the MSC in the coming months, however we would happily take early feedback in the meantime.

proposals/4258-federated-user-directory.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## Potential issues
+
+We may have requests lost or getting timeout from intermediary network equipment, especially since we are using some kind of long polling.
+We think the fact that we use a `search_token` that changes on each request allow the server to track correctly if new search results were already received by the client or not.
+
+## Alternatives
+
+We first thought about using an account data, however it has a big caveat: remote servers can't access it, hence remote servers will not be able to honor the visibility when trying to return remote users that are already visible locally to them.
```
clokep, line 116: An alternative to this is to include the requesting user to the federated servers and let them decide whether to provide the user in the result or not?

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+      "user_id": "@foo:bar.com"
+    }
+  ]
+}
+```
+`search_token` :  is a unique identifier that means that more results can be retrieved by querying with this `search_token`. `limited` should be `true` when `search_token` is returned.
+
+#### New profile field to control user visibility in the directory
+
+We propose to add a new field in the profile (MSC4133) `m.user_directory.visibility` to give the user the ability to control their visibility in the user directory.
```
clokep, line 91: This seems like an odd piece of data to allow other users to query.

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+
+We are hence proposing some changes to the client API to accommodate the need to have a way to stream new results to the client.
+
+Note that `m.user_directory.visibility` is defined further down this proposal.
+
+### Client endpoint changes
+
+We propose to introduce a reactive mechanism to allow the server to stream new results to the client.
+
+#### POST /_matrix/client/v3/user_directory/search
```
clokep, line 57: Clients cannot control if the request is only local or not then?

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+We first propose a new federation endpoint similar to the [current client API](https://spec.matrix.org/v1.12/client-server-api#post_matrixclientv3user_directorysearch).
+It would be authenticated and rate limited.
+
+#### `POST /_matrix/federation/v3/user_directory/search`
+
+#### Request
+```json
+{
+  "limit": 10,
+  "search_term": "foo"
```
clokep, line 23: Is there guidance on how the search term is used? Is it the same as the current API?

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+## Proposal
+
+### Federation endpoint
+
+We first propose a new federation endpoint similar to the [current client API](https://spec.matrix.org/v1.12/client-server-api#post_matrixclientv3user_directorysearch).
+It would be authenticated and rate limited.
+
+#### `POST /_matrix/federation/v3/user_directory/search`
```
clokep, line 17: What are the valid error conditions?

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+      "m.user_directory.visibility": "local",
+      "user_id": "@foo:bar.com"
+    }
+  ]
+}
+```
+
+All profile fields (cf [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133)) should be returned here.
+
+When an user calls the client user search API, the server should send a federated user search request to all known servers. It would then receive the results and return them to the user.
```
clokep, line 44: This sounds really really expensive for the server.

tcpipuk: This could benefit from #4259 🙂

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,150 @@
...
+
+We propose to add a new field in the profile (MSC4133) `m.user_directory.visibility` to give the user the ability to control their visibility in the user directory.
+
+Different values are possible :
+- `hidden` : not visible to anyone
+- `local` : visible only to local homeserver users
+- `restricted`: visible to any user sharing a room with
+- `remote` (or federated or public ?): visible to users on local and remote homeservers
+
+If no value is provided (or it is null), the user hasn't set a preference and the server should follow the current expected behavior (visible if sharing a room in common or in public room).
```
Johennes, line 99: "visible if sharing a room in common or in public room" is actually the only the [_minimum_](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3user_directorysearch) requirement.

proposals/4258-federated-user-directory.md
Johennes: Tagging on to this, the [profile federation API](https://spec.matrix.org/v1.13/server-server-api/#get_matrixfederationv1queryprofile) has a 403 to let server admins deny profile look-up. This might be good to have on the user directory API as well.

proposals/4258-federated-user-directory.md
Johennes: Could this maybe go into account data instead? It only needs to be accessed by the local homeserver, right?

proposals/4258-federated-user-directory.md
spaetz: Indeed MSC #4259 could/should be mentioned as a possibility to build upon?

proposals/4258-federated-user-directory.md
MatMaul: I can see several cases:
- an empty list should be returned if all matching users has visibility settings that would hide them.
- federated user directory is fully disabled on the server. 404 could be used here.
- federated user directory is restricted to a set of allowed servers for example. We should probably use 403 then.

proposals/4258-federated-user-directory.md
MatMaul: Current spec is telling to search the Matrix ID and the display name and for now this proposal doesn't change that.

Should we change that to also search all profile fields instead ? I am not sure, opinions welcome here.

proposals/4258-federated-user-directory.md
MatMaul: > This sounds really really expensive for the server.

We should probably relax the requirement to include all known servers, so that implementations can optimize that if needed. `to all known servers or a subset of it` for example ?

I am not sure the impact will be that heavy however : typing in a room with almost all servers like `Matrix HQ` will similarly broadcast events to all of those.

> This could benefit from #4259 🙂
> Indeed MSC https://github.com/matrix-org/matrix-spec-proposals/pull/4259 could/should be mentioned as a possibility to build upon?

Could you elaborate what you both have in mind ? for now it can help a bit to receive profile updates if we have a local search cache, but I don't really see something else.

proposals/4258-federated-user-directory.md
MatMaul: This proposal would change that, and user defined visibility would prevail, as expected by the users :)

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Sensitive Data Exposure
+
+A malicious server could list all user matrix ids that are defined in `remote` or `restricted`.
+
+#### Data Exposure Mitigation recommendations
+
+The federation search endpoint should be rate limited.
+
+We recommend to not answer for `search_term` with less than 3 characters like "a" or "at".
```
spaetz, line 142: so @a:matrix.org could never be found?

proposals/4258-federated-user-directory.md
spaetz: once the user id search term contains (part of) a domain name, we can probably stop asking random servers? Such as search term "@steve:matri" probably does not need to query the etke.cc server for users? I am just a little worried about the performance impact of these searches. (although most search will probably NOT contain domain parts. Not sure really.

spaetz (commented):
Thanks, it is such a **hard** problem (mainly performance and privacy). There is a reason why there is no federated email directory....
A few minor comments inline.

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
+# MSC4258: Federated User Directory
+
+Currently user search can only be done locally, which would at best get a list of all users known to the server.
```
spaetz, line 3:
```suggestion
Currently user search can only be done locally, which lists  - at maximum - all users known to the user's server.
```

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### New account data to control user visibility in the directory
+
+We propose to add a new account data of type `m.user_directory` with a single `visibility` field to give the user the ability to control their visibility in the user directory.
+
+Different values are possible :
+- `hidden` : not visible to anyone
+- `local` : visible only to local homeserver users
+- `restricted`: visible to any user sharing a room with
+- `remote`: visible to users on local and remote homeservers
```
Johennes, line 64: For [rooms](https://spec.matrix.org/v1.14/client-server-api/#listing-rooms), the visibility values are `private` and `public`. This could be mimicked here by using `private`, `local`, `restricted` and `public`. The terms "private" and "public" have been deemed [ambiguous](https://github.com/matrix-org/matrix-spec/issues/633) in the past. At the same time, however, maybe we're making things worse by adding yet another terminology. I don't have a strong opinion myself here to be honest.

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
...
+We propose to introduce a reactive mechanism to allow the server to stream new results to the client.
+
+For that we introduce a `search_token` to the request coming from the response of a previous search(`search_token`). A request containing a `search_token` will stall until new results are available to the server. If some more results are expected to be returned, it may include another `search_token`, and hence.
+
+`search_token` is optional within the request so proposed changes are retro-compatible.
+
+We also propose a new `search_scope` parameter to limit the scope of a search.
+Possible values are:
+- `local` : only search users local to the homeserver, this must not trigger a federated search
+- `restricted`: search users known to this homeserver, this must not trigger a federated search
```
Johennes, line 96: Would the server still have to trigger federated profile queries for external users in this case?

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
...
+```json
+{
+  "limited": false,
+  "results": [
+    {
+      "avatar_url": "mxc://bar.com/foo",
+      "display_name": "Foo",
+      "m.tz": "America/New_York",
+      "user_id": "@foo:bar.com",
+      "visibility": "local",
```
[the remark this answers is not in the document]
MatMaul: This is only in the federation response, we removed it from the client one after switching to account data. It is useful to be able to apply `restricted` visibility on the requester server.
It could be omitted if we pass the requester in the request, then the remote server is able to calculate if it should return the result or not.
We were more thinking about caching when designing the API, since having the visibility allows to cache the request for all users while we can only cache the result per user if we don't have the visibility.

Perhaps we should just return `"visibility": "restricted"` and nothing otherwise as a tradeoff, I am not sure. But yeah current state is too leaky I agree.

proposals/4258-federated-user-directory.md
MatMaul: I think this should be left as an implementation detail: we added `or a subset of it` at L47 so the homeserver can be quite liberal in the caching and requesting behavior.

proposals/4258-federated-user-directory.md
MatMaul: The implementation could just return the exact match but not the rest. I'll add it.

proposals/4258-federated-user-directory.md
MatMaul: I am not sure honestly. I think it may be left as an implementation detail for now at least, and if it turns out that one way or the other is quite superior during implementation we can specify it.

proposals/4258-federated-user-directory.md
MatMaul: I think I have a small preference to avoid the ambiguity, but both are fine for me.

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,162 @@
...
+
+However we have a problem here: we can't have expectation on when and even if servers will answer to the search request.
+
+We are hence proposing some changes to the client API to accommodate the need to have a way to stream new results to the client.
+
+### Client endpoint changes
+
+#### New account data to control user visibility in the directory
+
+We propose to add a new account data of type `m.user_directory` with a single `visibility` field to give the user the ability to control their visibility in the user directory.
```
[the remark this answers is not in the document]
MatMaul: I don't have a strong opinion, let's wait for others opinion on that I think.

proposals/4258-federated-user-directory.md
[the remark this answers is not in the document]
Johennes: Oh, I see. Sorry, I had mistaken this as the CS response. 🤦‍♂️

In that case this seems acceptable to me since the field is required for filtering by the server and being stripped out in the CS response.

proposals/4258-federated-user-directory.md
MatMaul: I've changed the MSC to do so, and put previous behavior in the alternate section.

proposals/4258-federated-user-directory.md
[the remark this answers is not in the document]
MatMaul: I've put the requester in the request, which avoid leaking the visibility setting even between servers.

proposals/4258-federated-user-directory.md
```diff
@@ -0,0 +1,169 @@
...
+
+Different values are possible :
+- `hidden` : not visible to anyone
+- `local` : visible only to local homeserver users
+- `restricted`: visible to any user sharing a room with
```
V02460, line 68: Is `restricted` a superset of `local`? Or will it result in some users on the same HS not be able to find each other? In that case, would another option that combines `local` and `restricted` make sense?

PR #4259, opened 2025-01-30
MSC4259: Profile Update EDUs for Federation

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/patch-1/proposals/4259-bulk-profile-sync.md)

---

Known Implementations:
- Servers:

### DIFF
--- a/proposals/4259-bulk-profile-sync.md
+++ b/proposals/4259-bulk-profile-sync.md
@@ -1,4 +1,4 @@
-# MSC4259: Profile Sync API for Federation
+# MSC4259: Bulk Profile Sync API for Federation
 
 Currently, homeservers must individually request profiles via
 [`/_matrix/federation/v1/query/profile`](https://spec.matrix.org/v1.13/server-server-api/#get_matrixfederationv1queryprofile)
--- a/proposals/4259-bulk-profile-sync.md
+++ b/proposals/4259-bulk-profile-sync.md
@@ -1,0 +1,186 @@
+# MSC4259: Profile Sync API for Federation
+
+Currently, homeservers must individually request profiles via
+[`/_matrix/federation/v1/query/profile`](https://spec.matrix.org/v1.13/server-server-api/#get_matrixfederationv1queryprofile)
+when they need to display user information. This leads to significant inefficiencies in the Matrix
+federation: servers must make separate requests for each profile they need, cannot efficiently
+detect when profiles change, and risk serving stale data to their users. The problem becomes more
+acute as servers cache profile data to reduce federation traffic, requiring careful balance between
+cache duration and data freshness.
+
+This proposal introduces a new federation endpoint that allows homeservers to efficiently sync
+profile data from multiple users in a single request. It builds upon
+[MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133) which introduced extended
+profile fields, and adds a token-based sync mechanism that lets servers track profile changes
+without repeatedly requesting unchanged data.
+
+## Proposal
+
+The current approach to federation profile lookups has several issues:
+
+- Inefficient use of network resources when requesting multiple profiles
+- Increased latency when displaying user information
+- Difficulty in maintaining accurate (non-stale) cached copies of remote profiles
+- No standardised way to efficiently detect when profiles have been updated
+
+This MSC adds a new `/_matrix/federation/v1/profiles` endpoint that accepts batched profile
+requests and returns only changed data on subsequent requests using sync tokens.
+
+### Behaviour
+
+1. The requesting server provides an array of MXIDs it wishes to sync profiles for
+
+2. If `last_batch` is provided, only profiles that have changed since that sync token are returned
+
+3. The responding server:
+   - MAY filter which profiles it returns based on its policies
+   - MAY return partial results and indicate this via `next_batch`
+   - MUST include `next_batch` in successful responses
+   - MUST NOT return profiles unless they were requested
+   - SHOULD indicate users that are not present on the server
+   - SHOULD suggest an appropriate `next_time` for subsequent requests
+
+4. Missing profiles in the response indicate no changes since `last_batch`
+
+5. Error responses for specific profiles suggest excluding them from future requests
+
+### Federation API Changes
+
+#### New Endpoint
+
+- **Method**: `POST`
+- **Endpoint**: `/_matrix/federation/v1/profiles`
+- **Auth**: Standard federation authentication
+- **Rate Limiting**: Implementation-defined, with server guidance via `next_time` response field
+
+#### Request Format
+
+```json
+{
+    "last_batch": "opaque_server_token", // Optional sync token from previous response
+    "accounts": [                        // Required array of MXIDs
+        "@alice:example.com",
+        "@bob:example.com"
+    ]
+}
+```
+
+#### Response Format
+
+```json
+{
+    "content": {
+        "@alice:example.com": {
+            "displayname": "Alice",
+            "avatar_url": "mxc://example.com/alice",
+            "org.example.language": "en-GB"
+        },
+        "@bob:example.com": {
+           "errcode": "M_NOT_FOUND",
+           "error": "User not found"
+        }
+    },
+    "next_batch": "opaque_server_token_2",
+    "next_time": "2024-03-14T12:30:00Z"
+}
+```
+
+#### Error Codes
+
+Inside profile payloads, standard Matrix error codes are used for errors:
+
+- `M_FORBIDDEN`: "User profile is not accessible to the requesting server"
+- `M_NOT_FOUND`: "User does not exist"
+
+For entire-endpoint errors, the following standard Matrix error codes are used:
+
+- **403 response**:
+  - `M_FORBIDDEN`: "Profile lookup over federation is disabled on this homeserver"
+
+- **404 response**:
+  - `M_UNKNOWN_TOKEN`: "Profile sync token unknown or expired"
+
+- **429 response**:
+  - `M_LIMIT_EXCEEDED`: "Too many requests" (should use `retry_after_ms` to indicate time to retry)
+
+### Implementation Notes
+
+- When no `last_batch` parameter is provided, return current profile data for all requested
+  accounts. This is typically used for initial sync or to start over when all tokens have expired.
+
+- Servers MUST maintain a history of sync tokens to allow clients to resume syncs after connection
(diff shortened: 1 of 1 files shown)


proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,186 @@
...
+- **Endpoint**: `/_matrix/federation/v1/profiles`
+- **Auth**: Standard federation authentication
+- **Rate Limiting**: Implementation-defined, with server guidance via `next_time` response field
+
+#### Request Format
+
+```json5
+{
+    "last_batch": "opaque_server_token", // Optional sync token from previous response
+    "accounts": [                        // Required array of MXIDs
```
Half-Shot, line 61: So my immediate worry here is scalability. Imagine a scenario where you might have two matrix.org sized servers cross federating. If each server has a few million accounts and we say for arguments sake that nearly all those users talk to each other, that's an array a few million in size happening every time period...

Granted, I accept that you might not request those profiles all of the time, but in a worst case scenario this would be a traffic nightmare (and probably worse than the EDU-update based approach byte-for-byte).

On the receiving server, it's going to have to maintain a token for your server in perpetuity (or some N seconds?) and with that it will need to store the list of profiles you care about. Not only would that be expensive, but it sounds like a great way to DDoS a server to me :)

Half-Shot (commented):
Some early thoughts.

proposals/4259-bulk-profile-sync.md
turt2live, line 1: Implementation requirements:
* [ ] Server(s)

tcpipuk:
Following a conversation with @Half-Shot in Matrix, I've given this a rewrite to use EDUs rather than a pull-based sync mechanism.

I considered proposing a subscription system and/or other metrics, however EDUs are not really designed for this and more thorough interest/permissions systems are intended in other more thorough solutions such as "rooms as profiles".

proposals/4259-bulk-profile-sync.md
tcpipuk: Thanks a lot for the feedback here. As I've now rewritten the MSC following our recent conversations, I've marked this thread as resolved so we can start afresh with the new version 🙂

proposals/4259-bulk-profile-sync.md
anoadragon453, line 1: A discussion on performance will inevitably come up on this MSC, so I'd like to start that up here (and see the MSC itself address it).

#### Presence Performance

Many reviewing this MSC will immediately think of the performance problems that
plague the [Presence](https://spec.matrix.org/v1.17/client-server-api/#presence)
feature.

The Synapse homeserver's implementation of presence works via the following:

1. Presence information can be set by the client, but the server will use
  requests such as `/sync` and read receipts to determine if the user is
  "currently active" or when they were last active.
2. Presence information is shared between all users who share a room.
3. Online user's presence state is changed to "idle" if they are not currently
  active and their last active time was >5min ago. They are set back to "online"
  when they make a relevant request.
4. When a user joins a room, their presence information is sent to other users,
  and they receive presence information about all other users in the room.

From 2020 data, Element observed in a publicly-federating customer's deployment
that a homeserver with 66 daily active users saw:

1. ~16Hz total (local and remote) presence updates
2. ~3Hz coming from remote servers
3. ~1.5Hz of updates proactively being sent to clients and other servers.

The majority of CPU usage was going towards sending federation transactions
(which dropped from ~20Hz to ~1Hz after disabling presence). Spikes occurred
when a local user joined a room, which involved sending presence to every server
in the room, and receiving presence for every server in the room.

They also of course occurred whenever a user - that was in rooms with many remote
servers - updated their presence.

Much of this can be optimised through implementation and protocol changes.

Perhaps we could weave presence information into `/{make,send}_join`, or have a
separate, single call to bulk push and receive presence information for all
users in the room that belong to a given homeserver.

Implementations can spread updates to different servers out over time,
preventing many transactions being sent in parallel overloading the homeserver.
They can also rate-limit and "debounce" updates (ignoring some that appear in a
quick succession). They can optimise their HTTP clients so TLS setup costs are
light on CPU resources.

I [still think profile fields hold value](https://github.com/matrix-org/matrix-spec-proposals/pull/4426#discussion_r2858697464) in their functionality over presence.
But similar optimisations can likely be applied to both features.

#### Profile Field-specific quirks

That being said, profile updates will generally be much less frequent than
presence updates in general. Profile fields make it easy to set different
rate-limits/debouncing logic for each field ID, e.g. you could limit `m.call`
updates much more strictly than `m.music`. Each field has its own traffic
patterns to consider.

I also think limiting who can see which profile will significantly cut down on
the traffic. A mechanism for allowing anyone to see your avatar/displayname, but
only friends seeing your current song, would go a long way. Presence didn't have
such a granular level of user-controlled recipients. This is best handled in a
separate MSC.

proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,115 @@
...
+This MSC adds a new `m.profile` EDU type that servers generate when users update their profiles,
+allowing real-time notifications of profile changes between servers.
+
+### Behaviour
+
+1. When a user updates their profile via client-server APIs, their homeserver:
+   - Processes and stores the profile update as per normal
+   - MAY generate EDUs to notify other servers
+   - If sending EDUs, SHOULD only send them to servers that share a room with the user,
+     filtering/ratelimiting/de-duplicating the broadcasts as needed to meet their own policies
```
sandhose, line 38: I would downgrade that to a MAY, because I would want us to support the case of "let's broadcast every profile update to this trusted homeserver" as a first-class citizen

proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,115 @@
...
+            "avatar_url": null,  // Signals removal of avatar_url
+            "org.example.language": "en-GB"
+        }
+    }
+}
+```
+
+The EDU contains only fields that have changed. Fields set to `null` should be considered removed.
+Omitted fields should be considered unchanged. Recipients can fetch the full profile using the
+existing federation API if they need to verify the complete state.
```
sandhose, line 67: We need to also support *removing* an entire profile (in case of deactivation) and maybe replacing an entire profile (or maybe remove + set is good enough for this case)

proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,115 @@
...
+- Servers should implement appropriate rate limiting for EDU generation/sending, and MAY delay
+  notifications to de-duplicate/combine multiple field updates into a single EDU
+
+- While this EDU system reduces the need for manual profile requests, implementations should note:
+  - Remote servers may not support or send these EDUs
+  - EDUs can occasionally fail to be delivered
+  - Both servers and clients COULD implement periodic re-fetching of profiles (e.g. weekly or
+    monthly) if they require stronger consistency guarantees
+  - The frequency of such re-fetching should be balanced against available resources, network
+    conditions, and desired data freshness
```
sandhose, line 82: Maybe this should be another MSC, but I think that a way to *backfill* the whole user directory (or at least the part that you're allowed to see) from another homeserver is crucial.

This would solve a very important gap, because this means that once you're aware of a remote homeserver, you're capable of indexing locally all the users that this homeserver wants to disclose to you without having rooms in common

proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,115 @@
...
+   - If sending EDUs, SHOULD only send them to servers that share a room with the user,
+     filtering/ratelimiting/de-duplicating the broadcasts as needed to meet their own policies
+
+2. Remote servers receiving the EDU:
+   - MAY use these EDUs to update cached profile data, and thus hold onto cached profiles for longer
+   - COULD ignore the EDU if they are not interested in updates for this user
+
+3. EDU handling:
+   - Servers MAY cache profile data from EDUs to reduce future federation traffic
+   - Profile information is considered public, so servers MAY broadcast to any known server
```
V02460, line 46: Maybe before declaring all profile information public, we should take a more nuanced stance. E.g. [MSC4258](https://github.com/matrix-org/matrix-spec-proposals/pull/4258) is putting quite some effort in with on-demand searches and rate-limiting  to prevent mass scraping of profile data and aggregating it in one place – even though individual profiles are technically public and can be queried and found via search.

This MSC would amass a huge amount of remote profile data. We should consider options to avoid that and keep our data footprint small.

proposals/4259-bulk-profile-sync.md
mcalinghee: A suggestion that could be considered as a "remove of profile" : 
```json5
{
"type": "m.profile",
"content": {
    "user_id": "@alice:example.com",
    "fields": null
   }
}
```

proposals/4259-bulk-profile-sync.md
```diff
@@ -0,0 +1,115 @@
...
+   - Servers MUST NOT forward EDUs to other servers
```
mcalinghee, line 47: [MSC4258](https://github.com/matrix-org/matrix-spec-proposals/pull/4258) is trying to resolve the search of federated users.
Would it be a good idea to update the user directory when remote servers are receiving those EDUs?
MSC4258 could possibly cover this particular point.

CrazyNicc, line 47: These two statements conflict with each other

proposals/4259-bulk-profile-sync.md
CrazyNicc, line 1: How will this interact with Per-Room Profiles? And will the existing method of broadcasting room profiles via State Events be deprecated here, or will they be limited to the Per-Room Profiles?

proposals/4259-bulk-profile-sync.md
sandhose: They don't really conflict, but it's ambiguous. I think you can read this as:

 - Profile information is considered public, so servers MAY broadcast *the profile information of their local users* to any known server
 - Servers MUST NOT forward EDUs to other servers

if that makes sense?

proposals/4259-bulk-profile-sync.md
sandhose: I would say that for now, it's complementary, as extensible profiles are not part of the `m.room.membership` events. We'll eventually need to think about how the global profile and the per-room profile interact, but that's better tackled in other MSCs, like MSC4466 or MSC4069

proposals/4259-bulk-profile-sync.md
CrazyNicc: yeah, that text explains it better

PR #4261, opened 2025-02-01
MSCxxxx: "Do not encrypt for device" flag

uhoreg:


proposals/4261-do-not-encrypt-for-device-flag.md
```diff
@@ -0,0 +1,51 @@
...
+decrypt a message; sending the room keys to all of the application service's
+puppeted users is redundant.  Another example is bots that only need to send
+messages, but not receive messages.
+
+We propose adding a flag to the device keys to indicate that encrypted messages
+should not be sent to certain devices.
+
+## Proposal
+
+A new optional `do_not_encrypt` property is added to the `DeviceKeys` structure
```
uhoreg, line 15: Suggestions for better names welcome

proposals/4261-do-not-encrypt-for-device-flag.md
turt2live, line 1: Implementation requirements:
* [ ] Client asking to not be encrypted to
* [ ] Client not encrypting to that other client

proposals/4261-do-not-encrypt-for-device-flag.md
```diff
@@ -0,0 +1,51 @@
+# MSC4261: "Do not encrypt for device" flag
+
+Some devices (such as bots and application service puppets) may not need to
+receive encrypted messages.  For example, an application service may have
+multiple puppeted users in a room, but only needs to receive room keys once to
+decrypt a message; sending the room keys to all of the application service's
```
tulir, line 6: https://github.com/matrix-org/matrix-spec-proposals/pull/4350 defines a method for appservices to avoid creating multiple sessions in the first place

uhoreg: #4350 looks like a more complete solution for bridges.  This MSC might still be useful for bots that only need to send messages, but not read messages.  But without the bridge use-case, it might not be worth it.  So I plan on closing this MSC once 4350 gets far enough along.

tulir: Oh yes, this could still be useful for normal bots, especially if combined with a command system like MSC4332 to give bots the ability to say "only encrypt commands for me". Telegram bots have a similar "privacy mode" flag where they either receive all messages or only registered commands

PR #4262, opened 2025-02-03
MSC4262: Sliding Sync Extension: Profile Updates

tcpipuk:
[Rendered](https://github.com/tcpipuk/matrix-spec-proposals/blob/patch-2/proposals/4262-sliding-sync-profiles.md)

---

Known Implementations:
- Clients:
  -  Element X
- Servers:
  - Synapse

### DIFF
--- a/proposals/4262-sliding-sync-profiles.md
+++ b/proposals/4262-sliding-sync-profiles.md
@@ -1,0 +1,121 @@
+# MSC4262: Sliding Sync Extension: Profile Updates
+
+This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) which adds support for receiving profile updates via Sliding Sync. It complements [MSC4259](https://github.com/matrix-org/matrix-spec-proposals/pull/4259) which handles federation-level profile updates.
+
+## Proposal
+
+MSC3575 currently does not include support for receiving profile field updates through the `/sync` endpoint. This extension adds support for receiving profile updates for users who are members of rooms the client is subscribed to.
+
+The proposal introduces a new extension called `profiles`. It processes the core extension arguments `enabled`, `rooms`, and `lists`, and adds the following optional arguments:
+
+```json5
+{
+    "enabled": true,           // sticky
+    "lists": ["rooms", "dms"], // sticky
+    "rooms": ["!abcd:example.com"], // sticky
+    "fields": ["displayname", "avatar_url"], // optional filter for specific profile fields
+    "include_history": false   // optional, defaults to false
+}
+```
+
+If `enabled` is `true`, then the sliding sync response MAY include profile updates in the following format:
+
+```json5
+{
+    "users": {
+        "@alice:example.com": {
+            "displayname": "Alice",
+            "avatar_url": "mxc://example.com/abc123",
+            "org.example.language": "en-GB"
+        },
+        "@bob:example.com": {
+            "displayname": null  // Field removal
+        }
+    }
+}
+```
+
+### Behaviour
+
+1. The extension only returns profile updates for users who are members of rooms that the client is subscribed to via either:
+   - Room IDs explicitly listed in the `rooms` argument
+   - Rooms that fall within the sliding windows specified in `lists`
+
+2. The optional `fields` argument allows clients to filter which profile fields they want to receive updates for. If omitted, all profile field updates are included.
+
+3. The optional `include_history` argument controls whether the initial sync includes recent historical profile changes:
+   - If false (default), only current profile states are sent on initial sync
+   - If true, the server MAY include recent profile changes that occurred before the sync
+
+4. On an initial sync:
+   - Profile data MUST only be sent for rooms returned in the sliding sync response
+   - If `include_history` is false, only current profile states are sent
+   - If `include_history` is true, recent profile changes MAY be included
+
+5. When live streaming:
+   - Profile updates MUST be sent as the server receives them
+   - For rooms which initially appear (`initial: true`) due to direct subscriptions or rooms moving into the sliding window, current profile states MUST be included
+   - A null value for a field indicates the field has been removed
+   - Omitted fields should be considered unchanged
+
+### Implementation Notes
+
+- Servers SHOULD implement appropriate batching and rate limiting of profile updates to prevent overwhelming clients
+
+- While this extension provides real-time profile updates, implementations should note:
+  - Network issues could cause missed updates
+  - Clients MAY implement periodic full profile refreshes if they require stronger consistency guarantees
+  - The frequency of such refreshes should be balanced against resources and desired freshness
+
+- Profile updates are typically infrequent compared to other real-time events like typing notifications, so including them in sliding sync is considered efficient
+
+## Potential Issues
+
+1. Large Initial Sync Payload
+   - With `include_history` enabled, the initial sync could be large for rooms with many members
+   - Servers should consider implementing reasonable limits on historical profile data
+
+2. Update Frequency
+   - Some users might update profiles frequently
+   - Implementations should consider rate limiting and batching updates
+
+## Alternatives
+
+1. Separate Profile Sync API
+   - Could provide more granular control over profile syncing
+   - Would increase complexity by requiring another connection
+   - Would duplicate much of sliding sync's room subscription logic
+
+2. Push-based Profile Updates
+   - Could use a separate WebSocket connection for profile updates
+   - Would increase complexity and connection overhead
+   - Would duplicate room subscription logic
+
+## Security Considerations
+
+1. Profile information is considered public data in Matrix
+
+2. The extension respects existing privacy boundaries:
+   - Only returns updates for users in rooms the client can access
+   - Follows the same authentication and authorization as the main sliding sync endpoint
+
+3. Rate limiting helps prevent abuse
+
+## Unstable Prefix
+
+No unstable prefix as Sliding Sync is still in review. To enable this extension, add this to your request JSON:
+
+```json
+{
+    "extensions": {
+        "profiles": {
+            "enabled": true
+        }
+    }
+}
+```
+
+## Dependencies
+
+This MSC builds on:
+- MSC3575 (Sliding Sync), which is not yet accepted into the spec
--- a/proposals/4262-sliding-sync-profiles.md
+++ b/proposals/4262-sliding-sync-profiles.md
@@ -1,6 +1,7 @@
 # MSC4262: Sliding Sync Extension: Profile Updates
 
 This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575)
+(and its proposed successor [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186))
 which adds support for receiving profile updates via Sliding Sync. It complements
 [MSC4259](https://github.com/matrix-org/matrix-spec-proposals/pull/4259) which handles
 federation-level profile updates.
@@ -132,4 +133,6 @@ request JSON:
 
 This MSC builds on:
 
-- MSC3575 (Sliding Sync), which is not yet accepted into the spec
+- [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) (Sliding Sync) or its
+  proposed successor [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)  
+  (Simplified Sliding Sync), neither of which are yet accepted into the spec
--- a/proposals/4262-sliding-sync-profiles.md
+++ b/proposals/4262-sliding-sync-profiles.md
@@ -134,5 +134,5 @@ request JSON:
 This MSC builds on:
 
 - [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) (Sliding Sync) or its
-  proposed successor [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)  
+  proposed successor [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
   (Simplified Sliding Sync), neither of which are yet accepted into the spec
--- a/proposals/4262-sliding-sync-profiles.md
+++ b/proposals/4262-sliding-sync-profiles.md
@@ -1,12 +1,18 @@
 # MSC4262: Sliding Sync Extension: Profile Updates
 
-This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575) which adds support for receiving profile updates via Sliding Sync. It complements [MSC4259](https://github.com/matrix-org/matrix-spec-proposals/pull/4259) which handles federation-level profile updates.
+This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575)
+which adds support for receiving profile updates via Sliding Sync. It complements
+[MSC4259](https://github.com/matrix-org/matrix-spec-proposals/pull/4259) which handles
+federation-level profile updates.
 
 ## Proposal
 
-MSC3575 currently does not include support for receiving profile field updates through the `/sync` endpoint. This extension adds support for receiving profile updates for users who are members of rooms the client is subscribed to.
+MSC3575 currently does not include support for receiving global profile field updates through the
+`/sync` endpoint. This extension adds support for receiving profile updates for users who are
+members of rooms the client is subscribed to.
 
-The proposal introduces a new extension called `profiles`. It processes the core extension arguments `enabled`, `rooms`, and `lists`, and adds the following optional arguments:
+The proposal introduces a new extension called `profiles`. It processes the core extension
+arguments `enabled`, `rooms`, and `lists`, and adds the following optional arguments:
 
 ```json5
 {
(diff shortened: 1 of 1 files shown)


proposals/4262-sliding-sync-profiles.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+        "@alice:example.com": {
+            "displayname": "Alice",
+            "avatar_url": "mxc://example.com/abc123",
+            "org.example.language": "en-GB"
+        },
+        "@bob:example.com": {
+            "displayname": null  // Field removal
+        }
+    }
+}
```
pixlwave, line 61: I'm looking at implementing this for Element X but it isn't immediately clear to me where in the SSS response this is expected to be. Could you clarify if it is
- as shown here and becomes a new top-level object.
- nested within the `extensions` object (it seems slightly odd that the extension would be called `profiles` but the object returned is then called `users`).
- adding the custom profile fields to an existing `users` e.g. if there is such a thing within `rooms`.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+2. The optional `fields` argument allows clients to filter which profile fields they want to receive
+   updates for. If omitted, all profile field updates are included.
+
+3. The optional `include_history` argument controls whether the initial sync includes recent
+   historical profile changes:
+   - If false (default), only current profile states are sent on initial sync
+   - If true, the server MAY include recent profile changes that occurred before the sync
+
+4. On an initial sync:
+   - Profile data MUST only be sent for rooms returned in the sliding sync response
```
pixlwave: How does this interact with the `lazy_members` flag? Presumably it should only return the profile data for members that have been included in the `rooms` response too, otherwise the initial sync could be massive if it includes e.g. matrix-hq?

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+        "@bob:example.com": {
+            "displayname": null  // Field removal
+        }
+    }
+}
+```
+
+### Behaviour
+
+1. The extension only returns profile updates for users who are members of rooms that the client is
```
pixlwave: Out of interest, is it expected for the user's own profile to be included too? This would provide a nice way to subscribe to the user's global profile for changes from other devices which I don't think exists right now (and maybe it could be included, even if the user isn't part of any rooms). Either way, the MSC should presumably explicitly mention the expected behaviour here.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+{
+    "enabled": true,           // sticky
+    "lists": ["rooms", "dms"], // sticky
+    "rooms": ["!abcd:example.com"], // sticky
+    "fields": ["displayname", "avatar_url"], // optional filter for specific profile fields
+    "include_history": false   // optional, defaults to false
+}
+```
+
+If `enabled` is `true`, then the sliding sync response MAY include profile updates in the following format:
```
pixlwave: I've also been wondering how this extension would interact with the `heroes` portion of the sync response too. Would the idea be that the client keeps track of these global profiles and automatically merges the new fields (i.e. everything but avatar/name) with each room's `heroes`?

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+4. On an initial sync:
+   - Profile data MUST only be sent for rooms returned in the sliding sync response
+   - If `include_history` is false, only current profile states are sent
+   - If `include_history` is true, recent profile changes MAY be included
+
+5. When live streaming:
+   - Profile updates MUST be sent as the server receives them
+   - For rooms which initially appear (`initial: true`) due to direct subscriptions or rooms moving
+     into the sliding window, current profile states MUST be included
+   - A null value for a field indicates the field has been removed
```
zecakeh: This conflicts with the profile endpoints of the CS API where setting a value to `null` doesn't remove the profile field, and might be an acceptable value since extended profiles were merged into the spec.

From [the definition of `PUT /profile/{userId}/{keyName}`](https://spec.matrix.org/v1.18/client-server-api/#put_matrixclientv3profileuseridkeyname):

> Servers MAY reject `null` values. Servers that accept `null` values SHOULD store them rather than treating `null` as a deletion request. Clients that want to delete a field, including its key and value, SHOULD use the `DELETE` endpoint instead.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
+# MSC4262: Sliding Sync Extension: Profile Updates
+
+This MSC is an extension to [MSC3575](https://github.com/matrix-org/matrix-spec-proposals/pull/3575)
+(and its proposed successor [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186))
+which adds support for receiving profile updates via Sliding Sync. It complements
```
anoadragon453: This MSC should be updated to focus on extending MSC4186, now that it has been accepted.

proposals/4262-sliding-sync-profiles.md
anoadragon453: Implementation authors should be aware that their implementations (especially Synapse) may currently differ from the accepted underlying [MSC4186: Simplified Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) document contents. See the bottom of that MSC for a list of implementation differences.

This MSC is written against the accepted version of MSC4186's definition of sliding sync.

proposals/4262-sliding-sync-profiles.md
anoadragon453: I had a [similar question while reviewing MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186#discussion_r3469755640). There, a leave for a given room is only sent down to the client if the room that was left was previously sent to the client in that connection.

Seems this MSC could specify a similar rule when choosing which rooms to include profile updates for.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+   - Room IDs explicitly listed in the `rooms` argument
+   - Rooms that fall within the sliding windows specified in `lists`
+
+2. The optional `fields` argument allows clients to filter which profile fields they want to receive
+   updates for. If omitted, all profile field updates are included.
+
+3. The optional `include_history` argument controls whether the initial sync includes recent
+   historical profile changes:
+   - If false (default), only current profile states are sent on initial sync
+   - If true, the server MAY include recent profile changes that occurred before the sync
```
anoadragon453: Why would a client want previous profile data? Do homeservers even store historical profile state (Synapse doesn't)?

proposals/4262-sliding-sync-profiles.md
anoadragon453: We ended up doing this to avoid performance problems in the legacy sync variant of this MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4429/changes#diff-ff7b92bd96566832eba816741b724860e898458319a5e7aaf3bb4626b629fb07R175

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,138 @@
...
+2. The extension respects existing privacy boundaries:
+   - Only returns updates for users in rooms the client can access
+   - Follows the same authentication and authorization as the main sliding sync endpoint
+
+3. Rate limiting helps prevent abuse
+
+## Unstable Prefix
+
+No unstable prefix as Sliding Sync is still in review. To enable this extension, add this to your
+request JSON:
```
anoadragon453: Different now!

anoadragon453 (changes requested):
@tcpipuk would you be opposed to myself taking over updating and maintaining this MSC? Element are currently looking to implement it as part of our work on [MSC4426: User Status Profile Fields](https://github.com/matrix-org/matrix-spec-proposals/pull/4426).

proposals/4262-sliding-sync-profiles.md
tcpipuk: I believe the objective was to show changes that occurred within the last minute or two before sync to understand whether a field has just updated, but a little time has passed so it's entirely possible I was on crack! 😄

tcpipuk:
Thank you for the offer, @anoadragon453 - I've let this one lapse because of a lot of life events over the last 12 months, so would be all too happy for you to shepherd this one through for Element (and the community's) benefit! 😊

proposals/4262-sliding-sync-profiles.md
pixlwave: https://github.com/matrix-org/matrix-spec-proposals/pull/4262/commits/45ca5e0f89c5dbe62ea7e779f0829c5aa76a9eb5

proposals/4262-sliding-sync-profiles.md
pixlwave: https://github.com/matrix-org/matrix-spec-proposals/pull/4262/commits/45ca5e0f89c5dbe62ea7e779f0829c5aa76a9eb5

proposals/4262-sliding-sync-profiles.md
pixlwave: From https://github.com/matrix-org/matrix-spec-proposals/pull/4262/commits/45ca5e0f89c5dbe62ea7e779f0829c5aa76a9eb5: Yes it is 🎉

proposals/4262-sliding-sync-profiles.md
pixlwave: Referenced in https://github.com/matrix-org/matrix-spec-proposals/pull/4262/commits/45ca5e0f89c5dbe62ea7e779f0829c5aa76a9eb5, I think this is clear how it works now from my side.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,207 @@
...
+            "users": {
+                "@alice:example.com": {
+                    // Fields that were newly set, or updated.
+                    "updated": {
+                        "displayname": "Alice",
+                        "avatar_url": "mxc://example.com/abc123",
+                        "org.example.language": "en-GB"
+                    },
+                    // A field has been removed from a user's profile.
+                    "removed": ["com.example.other_field"]
```
pixlwave: I think the only thing I'm unsure about since https://github.com/matrix-org/matrix-spec-proposals/pull/4262/commits/45ca5e0f89c5dbe62ea7e779f0829c5aa76a9eb5 is whether these `updated` and `removed` fields are optional or required?

anoadragon453: Good point. `removed` is optional, while `updated` is required. Of course the entire value could instead be a `null`, and the user entry itself may or may not exist :)

I'll add this to the doc, good catch!

pixlwave: Super, have updated Ruma, thanks.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,207 @@
...
+
+1. Profile information is considered public data in Matrix
+
+2. The extension respects existing privacy boundaries:
+   - Only returns updates for users in rooms the client can access
+   - Follows the same authentication and authorization as the main sliding sync endpoint
+
+## Unstable Prefix
+
+Before this MSC is accepted, implementations should use
```
jaywink, line 204: The response should also be, until stabilized:

```
extensions:
  org.matrix.msc4262.profiles:
    users: {}
```

I think?

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+```
+
+Profile fields that are either newly added to a user's profile or recently
+updated are found under `users-><user_id>->updated`. Likewise, any field IDs that
+are cleared/removed from a user's profile will appear under
+`users-><user_id>->removed`. Omitted fields are considered unchanged.
+
+Only fields specified by the `fields` request parameter will be included in
```
jaywink, line 74: Above in https://github.com/matrix-org/matrix-spec-proposals/pull/4262/changes#diff-1ca33124821bb0a7204f15b5f2751689eff710a33851a96de07c3c5d694e7367R29 we say this parameter is optional, which conflicts a bit?

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,207 @@
...
+that they may stop tracking this user, as that user has left all shared rooms.
+
+### When are profile updates returned?
+
+An individual sliding sync session (which persists over the lifetime of a
+client's login) contains top-level `rooms` and `lists` parameters. From these,
+the homeserver derives a list of rooms the client wants updates for.
+
+When a room enters this subset in this connection for the first time, all
+requested `fields` from profiles of users in that room MAY be sent down. This
```
jaywink, line 97: This also seems to indicate `fields` is a mandatory parameter.

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,207 @@
...
+
+```json5
+{
+    // extensions is a top-level field in the response body.
+    "extensions": {
+        "profiles": {
+            "users": {
+                "@alice:example.com": {
+                    // Fields that were newly set, or updated.
+                    "updated": {
```
jaywink, line 47: Nit: `updated` feels a bit weird as sometimes the fields may have been set a long time ago, though I'll buy it if it means "new to the client". But at least the comment above should reflect this fact.

Feels like this key could be something like `fields` or `values` 🤔

jaywink: Also, it might be nice to try and have the same structure with MSC4429 in the end, regarding the key for fields/updates/removals, where now the `profile_updates` key started bothering me a bit.. 😁

proposals/4262-sliding-sync-profiles.md
anoadragon453: Although... I suppose this doesn't actually make sense if a field was removed, but no `updates` were made... You'd want to see:

```json
"@alice:example.com": {
    "removed": ["com.example.other_field"]
}
```

So actually, *both* fields should be optional. Sorry for the flip-flopping.

proposals/4262-sliding-sync-profiles.md
anoadragon453: Hmm, yes, this is confusing. Clarified in 25ee654e.

proposals/4262-sliding-sync-profiles.md
anoadragon453: Great catch. Pointed it out explicitly in 37f09ec1.

proposals/4262-sliding-sync-profiles.md
anoadragon453: Clarified in 25ee654e.

proposals/4262-sliding-sync-profiles.md
anoadragon453: Yes, we need to have a `removed` key in MSC4429 as well, which would be another layer of nested dicts (with the existing fields going under `updated`, etc.).

`fields` works, but:

1. Is the same as the request parameter.
2. Implies that `removed` doesn't concern fields, which it does.

I like `updated` because it means both "added" or "changed", and stands in contrast to `removed`.

anoadragon453: I'd consider `updated` -> `updates` and `removed` -> `removals` though.

proposals/4262-sliding-sync-profiles.md
pixlwave: No worries, will update 👍

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,235 @@
...
+scope.
+
+If `enabled` is `true`, then the sliding sync response MAY include profile
+updates in the following format:
+
+```json5
+{
+    // extensions is a top-level field in the response body.
+    "extensions": {
+        "profiles": {
```
reivilibre, line 43: It's the dreaded question again: is there going to be any pagination, so the server doesn't send unboundedly large responses here? If so, how will it work?
https://github.com/matrix-org/matrix-spec/issues/2378

proposals/4262-sliding-sync-profiles.md
```diff
@@ -0,0 +1,253 @@
...
+
+```json5
+{
+    "enabled": true,
+    "fields": ["displayname", "avatar_url", "m.some_field"],
+}
+```
+
+The `fields` parameter is optional. If omitted, *all* profile fields will be in
+scope.
```
anoadragon453, line 34: Note that this differs from the semantics of the `profile_fields.ids` parameter of MSC4429, where omitting the parameter means *no* profile fields will be in scope.

The goal is to not suddenly start sending profile fields to legacy clients. Instead, they must opt-in. In legacy sync, that opt-in is done by providing the `profile_fields.ids` parameter. In sliding sync, clients opt-in to extensions via setting their `*.enabled` boolean. From there, all other parameters assume the client wants profile fields.

Therefore, if we defined an omitted `fields` parameter to mean "no profile updates" (like in MSC4429), you'd have:

```json
{
    "enabled": true
}
```

which wouldn't result in any fields being sent down. And that's practically useless. So instead we define it as "all fields".

I don't have a good use case for "receive updates to all known profile fields" yet, as clients will usually only want to receive updates for the fields they've built UI for already.

Perhaps if you're looking at a user's profile, showing all set fields, even unknown ones, and you want that UI to update in real time? 🤷 Interested in anyone has any other use cases.

nyphong39-svg (approved):
tcpipuk:patch-2

proposals/4262-sliding-sync-profiles.md
dbkr: I don't believe there's any reason an HS couldn't split a profile with many fields up into several sync responses: I don't think there's any meaning attached to them arriving together or separately. Each field should be small.

proposals/4262-sliding-sync-profiles.md
dbkr: This has been languishing a while. If there are strong opinions, voice them now. :)

proposals/4262-sliding-sync-profiles.md
jaywink: Feels like pagination should work similarly to how existing implementations are handling it, for example [thread subscriptions](https://github.com/element-hq/synapse/blob/2229a1b8e48054236c89044fb9192cb1ab872b09/synapse/rest/client/thread_subscriptions.py#L121-L121) / [MSC4308](https://github.com/matrix-org/matrix-spec-proposals/blob/rei/msc_ssext_threadsubs/proposals/4308-sliding-sync-ext-thread-subscriptions.md#companion-endpoint-for-backpaginating-thread-subscription-changes).

proposals/4262-sliding-sync-profiles.md
jaywink: Synapse implementation: https://github.com/element-hq/synapse/pull/20003

PR #4266, opened 2025-02-14
MSC4266: Policies in /.well-known/matrix/support

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/policies/proposals/4266-policies.md)

---

### DIFF
--- a/proposals/4266-policies.md
+++ b/proposals/4266-policies.md
@@ -1,0 +1,82 @@
+# MSC4266: Policies in `/.well-known/matrix/support`
+
+Matrix supports collecting policy consent from users during [account
+registration]. There is, however, no API allowing clients to retrieve the
+policies again at a later point. This requires the user to manually store them
+upon registration to be able to refer to them again.
+
+Furthermore, some policies, such as the privacy policy, might be relevant for
+users of other homeservers. Again, these users' clients have no way to
+programmatically retrieve the server's policies.
+
+The present proposal addresses this situation by publishing the policies in the
+server's support document under [`/.well-known/matrix/support`].
+
+## Proposal
+
+A new optional property `policies` is added to the response of
+[`/.well-known/matrix/support`]. The format is the same one used during
+[registration][account registration].
+
+``` json5
+{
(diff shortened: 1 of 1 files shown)


proposals/4266-policies.md
turt2live, line 1: Implementation requirements:
* [ ] Server publishing this
* [ ] Client using this

proposals/4266-policies.md
```diff
@@ -0,0 +1,82 @@
...
+## Alternatives
+
+It might be debatable whether policies represent "support" information. Instead
+of repurposing the support document, the policies could also be made available
+via a dedicated endpoint.
+
+Instead of querying the server, the client could store the policies in the
+user's account data. If [encrypted] this would prevent the server from tampering
+with the policies the user has consented to. This would, however, not allow
+external users to retrieve the policies.
```
FSG-Cat, line 68: From a homeserver administration and moderation perspective and just a community moderation perspective having access to the policies of remote servers is very useful. Policy documents for a remote server can give an initial impression of if the server has compatible policies with a given community or not.

proposals/4266-policies.md
```diff
@@ -0,0 +1,82 @@
...
+        "url": "https://example.org/somewhere/terms-1.2-fr.html"
+      },
+      "version": "1.2"
+    }
+  }
+}
+```
+
+If the request is authenticated, the server SHOULD respond with the latest
+version of the policies that the user consented to.
```
Johennes, line 53: On second thought, I suspect this might be tricky to do for `.well-known` files because these usually contain static content only. A dedicated `/policies` endpoint might be better after all.

1vel: I think a dedicated policies endpoint would be better. `.well-known` endpoints are typically served by a webserver, not the homeserver.

proposals/4266-policies.md
MTRNord: The Connectivity Tester now has support to display this https://github.com/MTRNord/matrix-connection-tester-ui/releases/tag/v0.5.1

PR #4267, opened 2025-02-17
MSCXXXX: Automatically forgetting rooms on leave

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/auto-forget/proposals/XXXX-auto-forget.md)

Fixes: https://github.com/matrix-org/matrix-spec/issues/447

---


proposals/4267-auto-forget.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

Johennes: Server: https://github.com/element-hq/synapse/pull/18196

Johennes: Client: https://github.com/matrix-org/matrix-js-sdk/pull/4732

proposals/4267-auto-forget.md
```diff
@@ -0,0 +1,70 @@
...
+
+This can limit clients' options to maintain an archive of historic rooms (such
+that they have left *without* forgetting them). Therefore, servers that
+auto-forget rooms MUST advertise that they do so via an `m.leave_without_forget`
+capability.
+
+``` json5
+{
+  "capabilities": {
+    "m.leave_without_forget": {
```
t3chguy, line 32: I got hung up on this naming. `m.leave_implicit_forget` or `m.leave_auto_forget` maybe? Leave without forget sounds like its doing the opposite of 

> Therefore, servers that
auto-forget rooms MUST advertise that they do so via an `m.leave_without_forget`
capability.

Johennes: Yeah, the current naming is a bit odd. I was thinking of the capability as "I can leave a room without forgetting it" but that makes it reciprocal to auto-forgetting rooms on the server.

You could also think of it inversely like in your suggestion as "I can leave and forget a room in one operation". My only concern with this was that when the capability is enabled, a client actually has _less_ choice. It's not actually that you _can_ leave and forget in one operation but rather that you _must_.

Maybe I'm just overthinking it though. 😵‍💫

t3chguy: `m.forget_forced_upon_leave` maybe?

Johennes: I like that. It's not reciprocal and expresses the coercion aspect. 👍

proposals/4267-auto-forget.md
girlbossceo: conduwuit server implementation:
https://github.com/girlbossceo/conduwuit/commit/35981d5aef8785c132d2e2a166cfcde1cd24169e
https://github.com/girlbossceo/conduwuit/commit/c8a730c29e3ec5c9d38028b89f3fd26ed546ef8f

kegsay (approved):
My only concern is that we may be unduly confident because Synapse has had the config option `forget_rooms_on_leave` for a long time now. The problem is it defaults to `false`, so I would like to hear from some people who have been using that option successfully.

proposals/4267-auto-forget.md
ara4n: it's a bit clunky though - how about just `m.forced_forget`?

proposals/4267-auto-forget.md
```diff
@@ -0,0 +1,70 @@
+# MSC4267: Automatically forgetting rooms on leave
```
ara4n, line 1: This looks very sensible to me, fwiw. I don't think we need further proof of implementation between 2x servers, given it's also pretty trivial to envisage how it'd work.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
This MSC has been [proposed](https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$XFWOEbFoyph8GXpuOPJR7Lt5zkHBx18XuLKomYuB6sA?via=matrix.org&via=element.io&via=envs.net) as ready for FCP by the author. The next steps are for an SCT member to review it and the implementation to make the determination if it's ready for FCP, then propose FCP.

turt2live:
@mscbot fcp merge

proposals/4267-auto-forget.md
richvdh, line 1: I guess this is fine, but: do we have actual evidence of the `forget_rooms_on_leave` setting causing client-side brokenness? I'm slightly wondering if this is a non-issue, and we can just say that servers can forget rooms when they want.

richvdh: In fact: I assume the main problem is that clients which *do* choose to offer a "historical rooms" section with "forget" buttons don't know which rooms to list in the historic section (because they don't know whether those historic rooms have been "forgotten").

In which case, I don't really think this solves the problem. Doesn't exactly the same problem occur if a user has two clients, and chooses to forget some rooms on one of them?

It feels like maybe we need a special section of (incremental) syncs which lists rooms that have been forgotten since the previous sync.

Johennes: I think there are two problems in the current spec:

1. A client has no way to know if the server auto-forgets on leave or not. As a result, it cannot be sure whether or not it should display any of its related UI, such as a "forget" button.
2. The act of forgetting rooms doesn't propagate automatically across multiple clients. As a result, clients that do support historical rooms, currently need to perform initial syncs once in a while to determine the current set of forgotten rooms.

This proposal only tries to address 1 while I think your comment aims at 2. I think both should be fixed but I don't think one fixes the other?

richvdh: My thinking was that servers that auto-forget rooms on leave would send out the same notification that a room had been forgotten as when another client did the forgetting.

Presumably clients only show a "forget" button on rooms that have been left but not yet forgotten (a list that would always be empty on servers with auto-forget), so a "forgotten" notification in /sync could solve both problems.

Johennes: First of, I think a mechanism for communicating forgotten rooms via incremental `sync` would be great. We should definitely do this.

I believe some UI usages would still require the capability from this proposal though. A client might want to:

- simultaneously display buttons for "leave" (`/leave`) and "leave & forget" (`/leave` followed by `/forget`)
- offer a setting to toggle what a single "leave" button does (`/leave` followed or not followed by `/forget`)
- explain the finality or non-finality of a single "leave" button before issuing a `/leave` (e.g. in a tooltip or dialog)
- display the empty "historical rooms" view or section alongside some documentation explaining what it is or how it'll be populated

I think these would not be possible or result in confusing UX if the client cannot determine the server's auto-forget behaviour ahead of time?

richvdh: fair enough; if those are real use-cases desired by clients, then I can't argue against a capability for it. Could you add them to the doc so that we know what we're trying to fix (and, as importantly, what we're *not* fixing)?

Johennes: I have tried to capture it in 1771d76084deecd44d1f6c4fec362b8f3c5e4fd5. Thanks for raising this.

To be honest, I have no evidence that these (or any other forget-related) use cases are particularly strong. The concept of differentiating left and forgotten rooms might be confusing or superfluous to begin with and maybe it shouldn't exist in the spec. However, I think the two problems we discussed above make it hard for clients to actually use this feature and experiment with its UX to then verify what's needed.

KitsuneRal: As a client author I wouldn't mind if we didn't even distinguish between left and forgotten rooms on the client side; simplifying the UX and making it clear at least for some servers is the next best thing.
(The "historical rooms" thing effectively never worked right, beyond of the "oh shucks I left the wrong room let me quickly rejoin it" case.)

proposals/4267-auto-forget.md
```diff
@@ -0,0 +1,95 @@
...
+    followed or not followed by `/forget`)
+-   explain the finality or non-finality of a single "leave" button before
+    issuing a `/leave` (e.g. in a tooltip or dialog)
+-   display the empty "historical rooms" view or section alongside some
+    documentation explaining what it is or how it'll be populated
+
+On a server that automatically forgets rooms on leave, these would result in
+confusing UX, however.
+
+The present proposal seeks to standaradize the currently proprietary behaviour
```
ara4n:
```suggestion
The present proposal seeks to standardize the currently proprietary behaviour
```

Johennes: Oh, looks like the spell check job didn't save me this time. 🤦‍♂️ 

Thanks.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2292

PR #4268, opened 2025-02-24
MSC4268: Sharing room keys for past messages

richvdh:
A proposal for sharing the keys to existing room messages when inviting someone to a room.

This reintroduces the functionality previously proposed in MSC3061.

Conflict of Interest declaration: I am employed by Element. This MSC was written as part of my work on the Element Cryptography team.

The functionality discussed here is implemented in current versions of Element Web and Element X.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/encrypted_history_sharing/proposals/4268-encrypted-history-sharing.md)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4268#issuecomment-4040658330)
[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4268#issuecomment-4040661140)

### DIFF
--- a/proposals/4268-encrypted-history-sharing.md
+++ b/proposals/4268-encrypted-history-sharing.md
@@ -135,7 +135,7 @@ The plaintext content of such a message should be:
 The properties within the `content` are defined as:
 
  * `room_id`: the room to which the keys in the key bundle relate. (This is
-   required so that Bob can download the key bundle at the right time.)
+   required so that Bob can defer downloading the bundle until he joins the room.)
 
  * `file`: `EncryptedFile` from the [encrypted attachment
    format](https://spec.matrix.org/v1.13/client-server-api/#extensions-to-mroommessage-msgtypes).
--- a/proposals/4268-encrypted-history-sharing.md
+++ b/proposals/4268-encrypted-history-sharing.md
@@ -27,6 +27,9 @@ future members of the room should have access to. Specifically, those are the
 megolm sessions associated with that room which were marked with
 `shared_history`: see [below](#shared_history-property-in-mroom_key-events).
 
+It is RECOMMENDED that Alice first makes sure she has downloaded any keys for
+the relevant room from the server-side key storage.
+
 The keys are assembled into a JSON object with the following structure:
 
 ```json5
--- a/proposals/4268-encrypted-history-sharing.md
+++ b/proposals/4268-encrypted-history-sharing.md
@@ -1,7 +1,7 @@
 # MSC4268: Sharing room keys for past messages
 
 In Matrix, rooms can be configured via the
-[`m.room.history_visibility`](https://spec.matrix.org/v1.14/client-server-api/#room-history-visibility)
+[`m.room.history_visibility`](https://spec.matrix.org/v1.17/client-server-api/#room-history-visibility)
 state event such that previously-sent messages can be visible to users that
 join the room. However, this is ineffective in encrypted rooms, where new
 joiners will lack the keys necessary to decrypt historical messages.
@@ -57,7 +57,7 @@ The keys are assembled into a JSON object with the following structure:
 
 The properties in the object are defined as:
 
-  * `room_keys`: an array of objects each with the following fields from [`ExportedSessionData`](https://spec.matrix.org/v1.14/client-server-api/#definition-exportedsessiondata):
+  * `room_keys`: an array of objects each with the following fields from [`ExportedSessionData`](https://spec.matrix.org/v1.17/client-server-api/#definition-exportedsessiondata):
 
      * `algorithm`
      * `room_id`
@@ -71,7 +71,7 @@ The properties in the object are defined as:
    implication).
 
  * `withheld`: an array of objects with the same format as the content of an
-   [`m.room_key.withheld`](https://spec.matrix.org/v1.14/client-server-api/#mroom_keywithheld)
+   [`m.room_key.withheld`](https://spec.matrix.org/v1.17/client-server-api/#mroom_keywithheld)
    message, usually with code `m.history_not_shared` (see
    [below](#new-withheld-code)) to indicate that the recipient isn't allowed to
    receive the key.
@@ -79,9 +79,9 @@ The properties in the object are defined as:
 A single session MUST NOT appear in both the `room_keys` and `withheld` sections.
 
 The JSON object is then encrypted using the same algorithm as [encrypted
-attachments](https://spec.matrix.org/v1.14/client-server-api/#sending-encrypted-attachments)
+attachments](https://spec.matrix.org/v1.17/client-server-api/#sending-encrypted-attachments)
 (i.e., with AES256-CTR), and uploaded with [`POST
-/_matrix/media/v3/upload`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixmediav3upload).
+/_matrix/media/v3/upload`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixmediav3upload).
 
 The details of this key bundle are then shared with Bob, as below.
 
@@ -90,13 +90,14 @@ The details of this key bundle are then shared with Bob, as below.
 Having uploaded the encrypted key bundle, Alice must share the details with each of Bob's devices.
 
 She first ensures she has an up-to-date list of his devices (performing a
-[`/keys/query`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3keysquery)
-request if necessary. She then sends a to-device message to each of his devices
+
+[`/keys/query`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixclientv3keysquery)
+request if necessary). She then sends a to-device message to each of his devices
 **which are correctly signed by his cross-signing keys**.
 
 A new to-device message type is defined, `m.room_key_bundle`, which MUST be
 encrypted using
-[Olm](https://spec.matrix.org/v1.14/client-server-api/#molmv1curve25519-aes-sha2).
+[Olm](https://spec.matrix.org/v1.17/client-server-api/#molmv1curve25519-aes-sha2).
 
 The plaintext content of such a message should be:
 
@@ -151,7 +152,7 @@ them.
 
 Suppose Alice and Bob are participating in an encrypted room, and Bob now
 wishes to invite Charlie to join the chat. If the [history
-visibility](https://spec.matrix.org/v1.14/client-server-api/#room-history-visibility)
+visibility](https://spec.matrix.org/v1.17/client-server-api/#room-history-visibility)
 settings allow, Bob can share the message decryption keys for previously sent
 messages with Charlie. However, it is dangerous for Bob to take the server's
 word for the history visibility setting: a malicious server admin collaborating
@@ -161,7 +162,7 @@ been changed over time and it can be difficult for clients to estalish which
 setting was in force for a particular Megolm session.
 
 To counter this, we add a `shared_history` property to
-[`m.room_key`](https://spec.matrix.org/v1.14/client-server-api/#mroom_key)
+[`m.room_key`](https://spec.matrix.org/v1.17/client-server-api/#mroom_key)
 messages, indicating that the creator of that Megolm session understands and
 agrees that the session keys may be shared with newly-invited users in
 future. For example:
@@ -193,10 +194,10 @@ or `world_readable`, or vice versa), then clients MUST rotate their outbound
 megolm session before sending more messages.
 
 In addition, a `shared_history` property is added to the [`BackedUpSessionData`
-type](https://spec.matrix.org/v1.14/client-server-api/#definition-backedupsessiondata)
+type](https://spec.matrix.org/v1.17/client-server-api/#definition-backedupsessiondata)
 in key backups (that is, the plaintext object that gets encrypted into the
 `session_data` field) and the [`ExportedSessionData`
-type](https://spec.matrix.org/v1.14/client-server-api/#definition-exportedsessiondata). In
+type](https://spec.matrix.org/v1.17/client-server-api/#definition-exportedsessiondata). In
 both cases, the new property is set to `true` if the session was shared with us
 with `shared_history: true`, and `false` otherwise.
 
@@ -223,7 +224,7 @@ In all cases, an absent or non-boolean `shared_history` property is treated the 
 ### New "withheld" code
 
 The spec currently
-[defines](https://spec.matrix.org/v1.14/client-server-api/#mroom_keywithheld) a
+[defines](https://spec.matrix.org/v1.17/client-server-api/#mroom_keywithheld) a
 number of "withheld" codes which are used to indicate that a client is
 deliberately *not* sharing a megolm session key with another. Normally these
 codes are used in `m.room_key.withheld` to-device events; as the text above
@@ -285,13 +286,85 @@ be when importing a key export; however:
     MUST show that information to the user, since he has only that user's word
     for the authenticity of those sessions.
 
-TODO: tell the sender we have finished with the bundle, so they can delete it?
+### Out-of-scope (for now, at least)
+
+* Ideally, the bundle would be deleted once the recipient has successfully
+  downloaded it. An implementation is left for a future MSC. This is tracked at
+  https://github.com/matrix-org/matrix-rust-sdk/issues/5113.
+
+* In future, we will need to extend the
+  [`BackedUpSessionData`](https://spec.matrix.org/v1.17/client-server-api/#definition-backedupsessiondata)
+  structure (used for server-side key backups) and
+  [`ExportedSessionData`](https://spec.matrix.org/v1.17/client-server-api/#key-export-format)
+  (used for key exports) to include information about whether the key was
+  forwarded by another user (and if so, by whom). However, this is redundant
+  until we have authenticated backups (MSC4048). This is tracked at
+  https://github.com/matrix-org/matrix-rust-sdk/issues/5110.
 
 ## Potential issues
 
+* As noted in the "Security considerations" section below, current
+  implementations make assumptions about when it is necessary to rotate
+  outgoing megolm sessions that are no longer sufficient.
+
 ## Alternatives
 
+We considered, and dismissed, a number of alternative approaches to solving
+this problem:
+
+* MSC3061, but without the security flaw.
+
+  Where MSC3061 implementations previously shared keys with *all* of the
+  recipient's devices, we could limit this behaviour to only *verified*
+  devices, hence mitigating the security concern.
+
+  However, this approach still had performance problems: sending a to-device
+  message to each recipient device, for each megolm session ever used in the
+  room, is prohibitive.
+
+* Shared key store
+
+  We considered whether it was possible for members in a room to somehow
+  collaborate on building some sort of "shared key store"; new members could
+  then be granted access.
+
+  However, we were unable to find satisfactory solutions to a number of
+  practical questions. For example:
+
+  * should the store be populated by all users, or just admins? If the former,
+    what if a mischeivous user uploads incorrect keys? If the latter, what if
+    no admins are online?
+
+  * What happens if there are users on multiple servers: do we need to
+    replicate the keystore between servers, or risk the keystore being
+    unavailable?
+
+  * How do we orchestrate rotating the keystore when a user leaves the room?
+
+* Request keys from existing room members
+
+  A new joiner could request keys to the room from existing room
+  members. However, this is (a) intrusive to existing room members, and (b)
+  very vulnerable to social-engineering attacks.
+
+These alternatives are also discussed in a presentation made at Matrix Conference
+2025: https://youtu.be/_E4ArQopptM?t=168.
+
 ## Security considerations
+
+* In an encrypted room, senders are required to rotate their outgoing Megolm
+  sessions after a user leaves the room, to ensure that the now-departed user
+  cannot decrypt messages sent after they left.
+
+  Currently, most implementations only do this if they have actually shared the
+  key to the session with the departed user (i.e., typically, they sent at
+  least one message while the user was in the room); they otherwise assume that
+  the user does not have a copy of the session decryption key and it is safe to
+  continue using the same session.
+
+  This latter is no longer a valid assumption.
+
+  TODO: how to solve this? https://github.com/element-hq/element-meta/issues/3078.
 
 * The proposed mechanism allows clients to share the decryption keys for
   significant amounts of encrypted content. Sharing historical keys in this way
@@ -313,7 +386,7 @@ TODO: tell the sender we have finished with the bundle, so they can delete it?
   inviter working in cahoots with a homeserver administrator could make it
   appear as though events sent by one user were in fact sent by another.
 
-  Ultimately, the recipient of a key bundle is taking the world of the sender
+  Ultimately, the recipient of a key bundle is taking the word of the sender
   of that key bundle as to the actual owner of each megolm session. This is an
   inevitable consequence of the deniability property of encrypted messaging.
 
--- a/proposals/4268-encrypted-history-sharing.md
+++ b/proposals/4268-encrypted-history-sharing.md
@@ -45,7 +45,7 @@ The keys are assembled into a JSON object with the following structure:
   "withheld": [
     {
       "algorithm": "m.megolm.v1.aes-sha2",
-      "code": "m.unauthorised",
+      "code": "m.history_not_shared",
       "reason": "History not shared",
       "room_id": "!Cuyf34gef24t:localhost",
       "sender_key": "RF3s+E7RkTQTGF2d8Deol0FkQvgII2aJDf3/Jp5mxVU",
@@ -58,7 +58,7 @@ The keys are assembled into a JSON object with the following structure:
 The properties in the object are defined as:
 
   * `room_keys`: an array of objects each with the following fields from [`ExportedSessionData`](https://spec.matrix.org/v1.14/client-server-api/#definition-exportedsessiondata):
-  
+
      * `algorithm`
      * `room_id`
      * `sender_claimed_keys`
@@ -72,8 +72,9 @@ The properties in the object are defined as:
 
  * `withheld`: an array of objects with the same format as the content of an
    [`m.room_key.withheld`](https://spec.matrix.org/v1.14/client-server-api/#mroom_keywithheld)
-   message, usually with code `m.unauthorised` to indicate that the recipient
-   isn't allowed to receive the key.
+   message, usually with code `m.history_not_shared` (see
+   [below](#new-withheld-code)) to indicate that the recipient isn't allowed to
+   receive the key.
 
 A single session MUST NOT appear in both the `room_keys` and `withheld` sections.
 
@@ -106,7 +107,7 @@ The plaintext content of such a message should be:
     "room_id": "!Cuyf34gef24t:localhost",
     "file": {
       "v": "v2",
-      "url": "mxc://example.org/FHyPlCeYUSFFxlgbQYZmoEoe",     
+      "url": "mxc://example.org/FHyPlCeYUSFFxlgbQYZmoEoe",
       "key": {
           "alg": "A256CTR",
           "ext": true,
@@ -121,7 +122,7 @@ The plaintext content of such a message should be:
     }
   },
   "sender": "@alice:example.com",
-  "recipient": "@bob:example.org", 
+  "recipient": "@bob:example.org",
   "recipient_keys": { "ed25519": "<bob_ed25519_key>" }
   "keys": { "ed25519": "<alice_ed25519_key>" },
   "sender_device_keys": { ... }
@@ -132,7 +133,7 @@ The properties within the `content` are defined as:
 
  * `room_id`: the room to which the keys in the key bundle relate. (This is
    required so that Bob can download the key bundle at the right time.)
- 
+
  * `file`: `EncryptedFile` from the [encrypted attachment
    format](https://spec.matrix.org/v1.13/client-server-api/#extensions-to-mroommessage-msgtypes).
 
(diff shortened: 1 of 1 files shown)


proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,314 @@
...
+to solve a similar problem; however, the mechanism used did not scale well. In
+addition, the implementation in `matrix-js-sdk` was subject to a [security
+vulnerability](https://matrix.org/blog/2024/10/security-disclosure-matrix-js-sdk-and-matrix-react-sdk/)
+which this proposal addresses.
+
+## Proposal
+
+### Room key bundle format
+
+When Alice is about to invite Bob to a room, she first assembles a "room key
```
bradtgmurray, line 24: Does this imply that clients should have all the keys for the room locally for the room at the time of the invite?

I'm thinking of a client that doesn't download it's complete key storage on login, but instead lazily requests keys as needed from it's homeserver. 

Should we be suggesting that clients should download their complete key storage at login? Or that clients that don't download all their keys at login should download all their keys prior to inviting someone using the existing `GET /_matrix/client/v3/room_keys/keys/{roomId}` endpoint?

Additionally, clients may need to defer sending the invite if they're currently catching up on a backlog of to_device events, and there might be a race where recent keys aren't available to the client when they would like to make the invite.

bradtgmurray: I should avoid making comments mid-Matrix Conference presentations, lest they be answered in later slides.

richvdh: lol :grin:  For those following at home, the answer to both questions is "yes", and this might be worth calling out in the text of the MSC

richvdh: (FWIW we do think that clients should be downloading their complete key storage at login, to guard against key loss due to a storage reset operation: https://github.com/element-hq/element-meta/issues/2446)

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,354 @@
...
+### Actions as a receiving client
+
+When Bob's client receives an `m.room_key_bundle` event from Alice, there are two possibilities:
+
+ * If Bob has recently accepted an invite to the room from Alice, the client
+   should immediately download the key bundle and start processing it. Note,
+   however, that this process must be resilient to Bob's client being restarted
+   before the download/import completes.
+
+   TODO: what does "recently" mean?
```
richvdh: Fixing this is still on our todo list; currently tracked as part of https://github.com/matrix-org/matrix-rust-sdk/issues/5112

proposals/4268-encrypted-history-sharing.md
richvdh: added a note to the body of the MSC

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+### Room key bundle format
+
+When Alice is about to invite Bob to a room, she first assembles a "room key
+bundle" containing all of the room keys for megolm sessions that she believes
+future members of the room should have access to. Specifically, those are the
+megolm sessions associated with that room which were marked with
+`shared_history`: see [below](#shared_history-property-in-mroom_key-events).
+
+It is RECOMMENDED that Alice first makes sure she has downloaded any keys for
+the relevant room from the server-side key storage.
```
richvdh, line 31: currently implementation is a WIP: https://github.com/matrix-org/matrix-rust-sdk/issues/5111

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+  significant amounts of data from the media store by sending
+  `m.room_key_bundle` messages pointing to large media files.
+
+  Further, a malicious sender might attempt to make the recipient believe that
+  a megolm session belonged to Alice, whereas it actually belonged to
+  Charlie. Even if the recipient later receives a key bundle from an honest
+  user, they may now have difficulty deciding which user was correct.
+
+  Both problems can be mitigated by only accepting key bundles when accepting
+  an invite from that user.
```
Johennes, line 460: This might be subjective but it feels relatively easy to be tricked into accepting an invite from a malicious user. They may use a reasonable looking profile and room, for instance. Ideally just accepting the invite shouldn't allow them to make my client start downloading arbitrarily large files.

Maybe there could be an intermediate step where the user explicitly confirms the download before it starts?

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+Suppose Alice and Bob are participating in an encrypted room, and Bob now
+wishes to invite Charlie to join the chat. If the [history
+visibility](https://spec.matrix.org/v1.17/client-server-api/#room-history-visibility)
+settings allow, Bob can share the message decryption keys for previously sent
+messages with Charlie. However, it is dangerous for Bob to take the server's
+word for the history visibility setting: a malicious server admin collaborating
+with Charlie could tell Bob that the history visibility was open when in fact
+it was restricted. In addition, the history visibility in a given room may have
+been changed over time and it can be difficult for clients to estalish which
+setting was in force for a particular Megolm session.
```
Johennes, line 166: I'm not sure if its feasible but would it help at all if the history visibility was locked into the room create event so that it's fixed for the lifetime of a room?

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+
+The details of this key bundle are then shared with Bob, as below.
+
+### `m.room_key_bundle` to-device message
+
+Having uploaded the encrypted key bundle, Alice must share the details with each of Bob's devices.
+
+She first ensures she has an up-to-date list of his devices (performing a
+
+[`/keys/query`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixclientv3keysquery)
```
anoadragon453, line 98:
```suggestion
She first ensures she has an up-to-date list of his devices (performing a
[`/keys/query`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixclientv3keysquery)
```

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+  "recipient_keys": { "ed25519": "<bob_ed25519_key>" }
+  "keys": { "ed25519": "<alice_ed25519_key>" },
+  "sender_device_keys": { ... }
+}
+```
+
+The properties within the `content` are defined as:
+
+ * `room_id`: the room to which the keys in the key bundle relate. (This is
+   required so that Bob can download the key bundle at the right time.)
```
anoadragon453: Why is the `room_id` required to download the bundle *at the right time*? Is this related to the DoS issue below?

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+
+Suppose Alice and Bob are participating in an encrypted room, and Bob now
+wishes to invite Charlie to join the chat. If the [history
+visibility](https://spec.matrix.org/v1.17/client-server-api/#room-history-visibility)
+settings allow, Bob can share the message decryption keys for previously sent
+messages with Charlie. However, it is dangerous for Bob to take the server's
+word for the history visibility setting: a malicious server admin collaborating
+with Charlie could tell Bob that the history visibility was open when in fact
+it was restricted. In addition, the history visibility in a given room may have
+been changed over time and it can be difficult for clients to estalish which
```
anoadragon453:
```suggestion
been changed over time and it can be difficult for clients to establish which
```

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+### New "withheld" code
+
+The spec currently
+[defines](https://spec.matrix.org/v1.17/client-server-api/#mroom_keywithheld) a
+number of "withheld" codes which are used to indicate that a client is
+deliberately *not* sharing a megolm session key with another. Normally these
+codes are used in `m.room_key.withheld` to-device events; as the text above
+specifies, we will now also use them in the `withheld` section of the room key bundle.
+
+This MSC proposes the addition of a new withheld code, `m.history_not_shared`,
```
anoadragon453, line 237: Can this code also be used in the aforementioned `m.room_key.withheld` to-device event?

proposals/4268-encrypted-history-sharing.md
anoadragon453: This appears to be referring to "this process must be resilient to Bob's client being restarted
   before the download/import completes." rather than "TODO: what does "recently" mean?".

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+### Actions as a receiving client
+
+When Bob's client receives an `m.room_key_bundle` event from Alice, there are two possibilities:
+
+ * If Bob has recently accepted an invite to the room from Alice, the client
+   should immediately download the key bundle and start processing it. Note,
+   however, that this process must be resilient to Bob's client being restarted
+   before the download/import completes.
+
+   TODO: what does "recently" mean?
```
anoadragon453: > TODO: what does "recently" mean?

I'm not sure if we need to say "recently" (ensuring that the download resumes after an app restart/crash is a separate matter).

Essentially, at the time of learning that an invite has been accepted (either by the current client or another of the user's), the client should either:

* immediately download the bundle
* delay/linearise the download to avoid a DoS

Even if the client is offline for a month and only learns about an invite acceptance much later, the key bundle is still desirable.

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,430 @@
...
+  sessions after a user leaves the room, to ensure that the now-departed user
+  cannot decrypt messages sent after they left.
+
+  Currently, most implementations only do this if they have actually shared the
+  key to the session with the departed user (i.e., typically, they sent at
+  least one message while the user was in the room); they otherwise assume that
+  the user does not have a copy of the session decryption key and it is safe to
+  continue using the same session.
+
+  This latter is no longer a valid assumption.
```
anoadragon453:
```suggestion
  The latter is no longer a valid assumption.
```

proposals/4268-encrypted-history-sharing.md
anoadragon453: Perhaps if the download is over some threshold (10MB? What's considered an outlier for bundle size?), then the client SHOULD display some sort of confirmation to the user, i.e.:

> This room has a large encrypted history (about 86 MB). Download now, download once on Wi-Fi, or skip.

Implying that if they refuse the download, they may miss out on being able to view encrypted history for the room (while potentially giving the option to request it again later?).

anoadragon453 (commented):
Generally on the right track. Just a few small details to iron out.

proposals/4268-encrypted-history-sharing.md
richvdh: Yes. Is this better?

```suggestion
   required so that Bob can defer downloading the bundle until he joins the room.)
```

proposals/4268-encrypted-history-sharing.md
richvdh: It certainly would, and we discussed exactly that as part of developing this feature, but it is a substantial change to the history-visibility model of Matrix, and we felt like driving that change through would be a challenging prerequisite.

proposals/4268-encrypted-history-sharing.md
richvdh: No, it would be meaningless there. I can add some words to clarify this.

proposals/4268-encrypted-history-sharing.md
richvdh: well, it refers to the whole thing.  matrix-org/matrix-rust-sdk#5112 is "be resilient to Bob's client being restarted", but to implement that, we need to make sure we have a persisted store of "recently accepted invites", which means we need to figure out an answer to this TODO.

proposals/4268-encrypted-history-sharing.md
richvdh: To correct an apparent misunderstanding here:

> either by the current client or another of the user's), the client should either

Current implementations only download the bundle if *they were the ones that accepted the invite*. We don't plan to change this: provided the inviter sent the key bundle to all of the user's devices, it should be sufficient.

proposals/4268-encrypted-history-sharing.md
Johennes: Yes, that's fair. I think this can be resolved then, thanks.

proposals/4268-encrypted-history-sharing.md
anoadragon453: Ah, yes! That makes much more sense.

proposals/4268-encrypted-history-sharing.md
anoadragon453: Aha, I didn't realise that the inviting client would sent a to-device message to all other clients of the invitee. Presumably there's only one key bundle, encrypted by one symmetric key, and then you encrypt that key individually to each device?

richvdh: yup.

proposals/4268-encrypted-history-sharing.md
deepadmax: The expected size of the bundle can be calculated from the size of the room, and only if it is disproportionally larger, you're given the option to download or skip. That should cover most usual cases and only flag outliers. Do you see any dangers in that?

Johennes: Do you have a specific heurestic in mind by which the invited client could estimate a reasonable size?

richvdh: >  (10MB? What's considered an outlier for bundle size?)

Some rough numbers:

Currently my room with biggest history has 25800 sessions. Let's suppose a "very big room" is 100000 sessions.

Each entry in the key bundle is ~516 bytes (varying slightly with length of `room_id`). 

So the bundle for a "very big room" would be about 50MB (assuming the overhead of encryption is negligible)

proposals/4268-encrypted-history-sharing.md
richvdh: done in fecc36ba.

proposals/4268-encrypted-history-sharing.md
richvdh: It's also worth saying there is an inherent limit to the size of a key bundle, via the homeserver's max_upload_size. On `matrix.org`, that limit is 100MB, which might be a bit on the large size, but I'd be reluctant to hardcode a limit much below that.

Having discussed this with the team: we note that adding a warning to the receiving user would be an easy extension later on, so in the interest of getting history sharing out the door before the heat-death of the universe, we're inclined to leave this for a future iteration.

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,432 @@
...
+
+  We considered whether it was possible for members in a room to somehow
+  collaborate on building some sort of "shared key store"; new members could
+  then be granted access.
+
+  However, we were unable to find satisfactory solutions to a number of
+  practical questions. For example:
+
+  * should the store be populated by all users, or just admins? If the former,
+    what if a mischeivous user uploads incorrect keys? If the latter, what if
```
richvdh:
```suggestion
    what if a mischievous user uploads incorrect keys? If the latter, what if
```

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,432 @@
...
+   The `shared_history` flag defined below is omitted (it is `true` by
+   implication).
+
+ * `withheld`: an array of objects with the same format as the content of an
+   [`m.room_key.withheld`](https://spec.matrix.org/v1.17/client-server-api/#mroom_keywithheld)
+   message, usually with code `m.history_not_shared` (see
+   [below](#new-withheld-code)) to indicate that the recipient isn't allowed to
+   receive the key.
+
+A single session MUST NOT appear in both the `room_keys` and `withheld` sections.
```
turt2live: if it does, what does the receiver do? Reject the whole bundle, or accept the `room_keys` version?

turt2live (approved):
overall lgtm - thanks!

I'm going to drop this from the SCT's review tracker - if further review is needed, please visit the [SCT Office](https://matrix.to/#/#sct-office:matrix.org) to re-raise it.

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,432 @@
...
+
+The JSON object is then encrypted using the same algorithm as [encrypted
+attachments](https://spec.matrix.org/v1.17/client-server-api/#sending-encrypted-attachments)
+(i.e., with AES256-CTR), and uploaded with [`POST
+/_matrix/media/v3/upload`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixmediav3upload).
```
kegsay, line 89: Some homeservers apply limits to the total amount a single user can upload to the media repo. It would be very surprising to the end user if they try to invite someone and it fails because they have exceeded their quota. It would similarly be surprising to see that they have consumed all their quota simply by inviting a few people to encrypted rooms.

I would prefer to see some distinction between uploads for sharing room keys and uploads for sharing attachments because:
 - it can be quota'd differently
 - it can expire (e.g. after 1 day)
 - it can have different size limits per attachment (or at least a minimum limit)

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,432 @@
...
+The details of this key bundle are then shared with Bob, as below.
+
+### `m.room_key_bundle` to-device message
+
+Having uploaded the encrypted key bundle, Alice must share the details with each of Bob's devices.
+
+She first ensures she has an up-to-date list of his devices (performing a
+[`/keys/query`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixclientv3keysquery)
+request if necessary). She then sends a to-device message to each of his devices
+**which are correctly signed by his cross-signing keys**.
```
kegsay, line 100: I know the intent here but surely a malicious server can trivially bypass this by resetting the invited user's identity? You won't see an alert popup (the main safeguard we have) since they aren't in the room yet. 

Similarly, if you've never interacted with that user you would TOFU attacker-controlled users (e.g. if you typo the invited user ID).

If we restrict it to only share history with those we have verified then the feature becomes much less useful potentially.

richvdh: > I know the intent here but surely a malicious server can trivially bypass this by resetting the invited user's identity? You won't see an alert popup (the main safeguard we have) since they aren't in the room yet.

If this happens with a user that you've previously shared a room with, then we'll be tracking their previous identity and will ignore the new identity until the user acknowledges it.

> Similarly, if you've never interacted with that user you would TOFU attacker-controlled users (e.g. if you typo the invited user ID).

Yeah, this is a danger, and somewhat related to Signalgate. I wonder if the UI should warn you when you're inviting someone you haven't interacted with before. I'll raise it to discuss with the team.

richvdh: https://github.com/element-hq/element-meta/issues/3163

proposals/4268-encrypted-history-sharing.md
turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4425 and similar would probably help with this. Something that MSC doesn't mention is the server could force-expire ephemeral media, effectively treating it as a separate quota.

proposals/4268-encrypted-history-sharing.md
richvdh: :shrug: I'm not sure it materially matters.

proposals/4268-encrypted-history-sharing.md
richvdh: Fixed by https://github.com/matrix-org/matrix-js-sdk/pull/5171 and https://github.com/matrix-org/matrix-rust-sdk/pull/6017

proposals/4268-encrypted-history-sharing.md
richvdh: We figured out an answer of 24 hours. Clarified in b7056c20.

proposals/4268-encrypted-history-sharing.md
richvdh: Added a note about this in the "potential issues" section in b7056c20.

proposals/4268-encrypted-history-sharing.md
richvdh: Hopefully this is clear enough.

proposals/4268-encrypted-history-sharing.md
richvdh: Added a note about this in b7056c20.

proposals/4268-encrypted-history-sharing.md
richvdh: Added a note about this in the "Potential issues" section in b7056c20.

proposals/4268-encrypted-history-sharing.md
richvdh: Added a note about this in the "Security considerations" section in b7056c20.

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,463 @@
...
+### Actions as a receiving client
+
+When Bob's client receives an `m.room_key_bundle` event from Alice, there are two possibilities:
+
+ * If Bob has recently accepted an invite to the room from Alice, the client
+   should immediately download the key bundle and start processing it. Note,
+   however, that this process must be resilient to Bob's client being restarted
+   before the download/import completes.
+
+   The definition of "recently" is left up to clients. (They should consider
```
kegsay, line 278: I think this is a mistake. This value determines how long the key bundle must be stored on the server (e.g. in the case of a partial download before the client was closed). By leaving "recently" undefined, we lose any ability to set reasonable TTLs on this data.

As a general design goal, I do not think we should be writing APIs which allow unbounded data to accumulate. This is what this is doing.

richvdh:
# MSC Checklist

This document contains a list of final checks to perform on an MSC before it
is accepted. The purpose is to prevent small clarifications needing to be
made to the MSC after it has already been accepted.

Spec Core Team (SCT) members, please ensure that all of the following checks
pass before accepting a given Matrix Spec Change (MSC).

MSC authors, feel free to ask in a thread on your PR or in the
[#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) room for
clarification of any of these points.

- [X] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal)
- [n/a] Are all MSCs that this MSC depends on already accepted?
- [n/a] For each new endpoint that is introduced:
  - [n/a] Have authentication requirements been specified?
  - [n/a] Have rate-limiting requirements been specified?
  - [n/a] Have guest access requirements been specified?
  - [n/a] Are error responses specified?
    - [n/a] Does each error case have a specified `errcode` (e.g. `M_FORBIDDEN`) and HTTP status code?
      - [n/a] If a new `errcode` is introduced, is it clear that it is new?
- [no] Will the MSC require a new room version, and if so, has that been made clear?
  - [n/a] Is the reason for a new room version clearly stated? For example,
- [n/a] Are the [endpoint conventions](https://spec.matrix.org/latest/appendices/#conventions-for-matrix-apis) honoured?
  - [n/a] Do HTTP endpoints `use_underscores_like_this`?
  - [n/a] Will the endpoint return unbounded data? If so, has pagination been considered?
  - [n/a] If the endpoint utilises pagination, is it consistent with
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

richvdh:
I think this is ready for FCP.

@mscbot fcp merge

proposals/4268-encrypted-history-sharing.md
turt2live: Imposing a limit in the spec feels like we'd also be limiting use cases. Instead, because clients with slightly obscure requirements (such as availability) are typically attached to servers which are aware of those requirements, we can likely move this timeout in part to the server. 

The client *should* still make its own decisions about what to request, but the server *should* also decide how long it keeps bundles around for. These two values may not be the same, and can lead to download failures: this is acceptable, in my opinion. A future MSC can introduce a re-request mechanism if it becomes a high enough priority issue.

(this is non-blocking for FCP in my opinion: I'm fine with the text as-written, but would appreciate a note that the server might delete things before the client can request it)

turt2live (approved):
no blocking concerns - thank you!

anoadragon453 (approved):

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,468 @@
...
+
+   Delaying the download in this way avoids a potential DoS vector in which an
+   attacker can cause the victim to download a lot of useless data.
+
+Once Bob has downloaded the key bundle, the sessions are imported as they would
+be when importing a key export; however:
+
+  * Only keys for the relevant room should be imported.
+
+  * Bob's client should remember who sent the keys (Alice, in this case), and
```
uhoreg: I think

```suggestion
  * Bob's client should remember who he received the keys from (Alice, in this case), and
```

would be clearer (or maybe "... who sent him the keys ...") to distinguish from the original sender of the keys.

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,468 @@
...
+
+  Currently, most implementations only do this if they have actually shared the
+  key to the session with the departed user (i.e., typically, they sent at
+  least one message while the user was in the room); they otherwise assume that
+  the user does not have a copy of the session decryption key and it is safe to
+  continue using the same session.
+
+  The latter is no longer a valid assumption. Instead, clients MUST observe
+  changes in state in the room, and whenever they see a user leaving the room,
+  assume that the departed user may have access to any existing Megol session,
```
uhoreg: typo

```suggestion
  assume that the departed user may have access to any existing Megolm session,
```

proposals/4268-encrypted-history-sharing.md
richvdh: @turt2live I basically agree with you, but it's worth noting that a server has no means of distinguishing a key bundle from any other piece of encrypted media (other than via side-channels), so it might be hard for a server to apply a sensible timeout.

We *could* find some way to decorate it, if we considered it important.

turt2live: Yea, I'd expect it to be decorated by some timeout or ephemeral flag that could theoretically be used by other media (but won't be)

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,468 @@
...
+* In future, we will need to extend the
+  [`BackedUpSessionData`](https://spec.matrix.org/v1.17/client-server-api/#definition-backedupsessiondata)
+  structure (used for server-side key backups) and
+  [`ExportedSessionData`](https://spec.matrix.org/v1.17/client-server-api/#key-export-format)
+  (used for key exports) to include information about whether the key was
+  forwarded by another user (and if so, by whom). However, this is redundant
+  until we have authenticated backups (MSC4048). This is tracked at
+  https://github.com/matrix-org/matrix-rust-sdk/issues/5110.
+
+## Potential issues
```
clokep, line 322: How big can this bundle get? Will (de)serializing it cause potential memory issues with clients? Do we need to allow for multiple?

richvdh: It's limited only by the size limits on the media store (e.g. 50MB by default for Synapse, 10MB for matrix.org's free tier). Some discussion about this [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4268#discussion_r2698313783).

We're basically assuming that clients can safely decrypt and hold in memory anything that the media store can send them, so I'm not too worried about us blowing out memory limits.

At the other end, it has to be said that matrix.org's paid tiers have been introduced since we started this project, and it's not inconceivable that people will hit the 10MB limit if they invite someone to a massive encrypted room. That said (a) that sounds like the level of usage where they should be either paying or going elsewhere, and (b) even if we want to support multiple bundles in future, I'd much rather leave that until we're sure we need it.

proposals/4268-encrypted-history-sharing.md
richvdh: Ok, so I guess that would mean:

 * Stick a query parameter of some kind (`media_type`, maybe?) on [`/_matrix/media/v3/upload`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixmediav3upload), and put something in the metadata part of the response in [`/_matrix/federation/v1/media/download/{mediaId}`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1mediadownloadmediaid).
 * Leave the actual TTLs up to HS impls, for now at least. 
 * What would happen if a client forgot to set the flag? Should we add another query param on [`/_matrix/client/v1/media/download/{serverName}/{mediaId}`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaid) saying "only give me the media if it has the key-bundle flag set", so that we can enforce it being set?

I guess this might also help solve the incompatibility with MSC3911?

We (Element crypto team) are keen to release this feature soon, rather than get it too bogged down in spec hell, so I'd appreciate a temperature reading on this from SCT members (and @kegsay, as the OP on this thread): maybe you could use the [apache voting scale](https://apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions) to express your feelings on shipping without a way for clients to indicate that a specific media upload is actually a key bundle.

proposals/4268-encrypted-history-sharing.md
clokep: Thanks, I'm not sure if any of this is worth adding to the MSC? Maybe just a "clients may generate a file too large to upload, in which case they should do X"? Not sure it is worth it as a callout.

proposals/4268-encrypted-history-sharing.md
erikjohnston: I personally think it would be beneficial to flag to the server that this is a key bundle. However, we know that MSC3911 is something that we'll (eventually) do and so we'll need to annotate these bundles in some way anyway, so I don't really care if we punt that question until then.

I think we should note in the MSC that servers may delete media based on last usage, and so clients cannot assume that the bundle will be around forever (and downloading may fail).

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,468 @@
...
+  * Bob's client should remember who he received the keys from (Alice, in this case), and
+    MUST show that information to the user, since he has only that user's word
+    for the authenticity of those sessions.
+
+### Out-of-scope (for now, at least)
+
+* Ideally, the bundle would be deleted once the recipient has successfully
+  downloaded it. An implementation is left for a future MSC, such as
+  [MSC4425](https://github.com/matrix-org/matrix-spec-proposals/pull/4425). This
+  is tracked at https://github.com/matrix-org/matrix-rust-sdk/issues/5113.
```
tulir, line 311: Deleting the bundles being out of scope feels very not nice, but I guess it's not the end of the world since it's encrypted with AES-256 which isn't expected to be cracked by quantum computers

proposals/4268-encrypted-history-sharing.md
```diff
@@ -0,0 +1,468 @@
+# MSC4268: Sharing room keys for past messages
```
ara4n, line 1: My two main concerns on this MSC are:

 * Without the ability to delete key bundles once they've been consumed by the invited user, we accumulate encrypted key material in a way which is tantamount to harvest-now-decrypt-later attacks.  However, given MSC4425 (or other ephemeral MSCs) are on the radar, I'm (just about) happy to for this to progress on the understanding that we follow up afterwards rapidly with MSC4425 or similar.

 * In terms of use experience, it's really not obvious to me that the common case should be to share all possible keys with an invited user. Sometimes this will be important (e.g. knowledge transfer to a new user joining a team/org) - but sometimes it will be very undesirable **even if the room has shared history viz**, particularly if the history viz been changed back and forth over time.  I suspect that an approach similar to the WhatsApp idiom of "how much history do you want to share with the new user? nothing? 1 msg? 100 msgs? everything?" might help with this.  However: this is a client UX question and is already supported by the MSC.

So, neither of these points end up being blocking, although both make me quite uncomfortable with the proposal. I'll checkbox it to proceed, but please can we track these concerns on the Element crypto-team agenda so they're not lost.

proposals/4268-encrypted-history-sharing.md
richvdh: @mscbot concern add some notes around deleting media

proposals/4268-encrypted-history-sharing.md
richvdh: I've added a more explicit note in the "Potential Issues" section in 31fa0dd1. I don't have any great suggestions for what clients should do about it right now, so I'm glossing over that bit.

proposals/4268-encrypted-history-sharing.md
richvdh: I've not had screams of outrage from any of the SCT here, and as Erik says we'll probably need to do something here in future, so rather than trying to second-guess what we need, I'm going to punt this.

I added a note reminding clients that media can disappear in 31fa0dd1.

proposals/4268-encrypted-history-sharing.md
richvdh: > * Without the ability to delete key bundles once they've been consumed by the invited user, we accumulate encrypted key material in a way which is tantamount to harvest-now-decrypt-later attacks.  However, given [MSC4425](https://github.com/matrix-org/matrix-doc/pull/4425) (or other ephemeral MSCs) are on the radar, I'm (just about) happy to for this to progress on the understanding that we follow up afterwards rapidly with [MSC4425](https://github.com/matrix-org/matrix-doc/pull/4425) or similar.

This is already tracked at https://github.com/matrix-org/matrix-rust-sdk/issues/5113, though I would say that we'd need to deprioritise some other scheduled work for any followup to be rapid.

> In terms of use experience, it's really not obvious to me that the common case should be to share all possible keys with an invited user. 

My inclination on this is to deliver the implementation we have and see what feedback we get.

proposals/4268-encrypted-history-sharing.md
richvdh: @mscbot resolve add some notes around deleting media

richvdh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2399

turt2live:
Merged 🎉

PR #4269, opened 2025-02-25
MSC4269: Unambiguous mentions in body

heftig:
[Rendered](https://github.com/heftig/matrix-spec-proposals/blob/unambiguous-mentions-in-body/proposals/4269-unambiguous-mentions-in-body.md)

The message `body` should contain user IDs instead of display names.

### DIFF
--- a/proposals/4269-unambiguous-mentions-in-body.md
+++ b/proposals/4269-unambiguous-mentions-in-body.md
@@ -1,0 +1,52 @@
+# MSC4269: Unambiguous mentions in body
+
+The use of display names makes it difficult to parse user mentions from a message body. Clients can
+also make it difficult to type user IDs without having them replaced with a mention.
+
+Some clients, such as bots, have no use for or do not want to deal with the complexity of parsing
+the pseudo-HTML of the `formatted_body`. In addition, the `m.mentions` added by [MSC3952] is not
+enough, either, as the listed user IDs lack the context of the message body.
+
+Translating the display names in the message body back to user IDs is difficult, if not impossible.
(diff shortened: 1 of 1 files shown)


proposals/4269-unambiguous-mentions-in-body.md
```diff
@@ -0,0 +1,52 @@
...
+
+Translating the display names in the message body back to user IDs is difficult, if not impossible.
+However, the reverse operation - translating unambiguous user IDs in the body into display names -
+is trivial in comparison.
+
+Thus it is proposed that the message `body` should contain user IDs instead of display names. 
+
+## Proposal
+
+The specified client behavior for user and room mentions is modified so that the event's `body`
```
Johennes, line 18: Linking to the spec here might help in better understanding the proposal.

```suggestion
The [specified client behavior](https://spec.matrix.org/v1.13/client-server-api/#user-and-room-mentions) for user and room mentions is modified so that the event's `body`
```

proposals/4269-unambiguous-mentions-in-body.md
```diff
@@ -0,0 +1,52 @@
...
+should contain the unambiguous user ID, canonical room alias or room ID instead of the anchor's
+text component.
+
+## Potential issues
+
+Some clients, especially bridges, might rely on the body containing the display name for a good user
+experience. However, it would not be difficult to modify such a client to translate user IDs in the
+message body to display names.
```
Johennes, line 26: I think this would make it impossible to send a literal MXID in `body` without having it transformed to the display name upon display?

It would also be a breaking change which could possibly require a new room version.

PR #4273, opened 2025-03-12
MSC0000: Approve and Disapprove ratings for moderation policies

Gnuxie:

### DIFF
--- a/proposals/0000-approval-disapproval.md
+++ b/proposals/0000-approval-disapproval.md
@@ -1,0 +1,140 @@
+# MSC0000: Approve and Disapprove ratings for moderation policies
+
+Currently, when watching a moderation policy list, there is no way to express
+approval or disapproval of certain policies.
+
+Typically, Matrix moderation tooling requires that moderators accept all
+policies within a moderation policy list in an all-or-nothing approach. We call
+this profile of policy list subscription _direct propagation_: any policy that
+appears on a watched list with _direct propagation_ semantics are directly
+considered and actioned by tooling. Examples of this include Mjonlir & Draupnir.
+
+For example: `charity`, the moderator of `cat-community.example.com`, cannot
+subscrbe to the list, `#bat-coc-bl:bat-community.example.org` without accepting
+all of the existing policies. This is problematic because `charity`'s friend
+`bob` was banned from the list for arguing with a moderator. `charity` respects
+the ban in the bat community, but still needs to be able to interact with their
+friend.
+
+Addtionaly, this all-or-nothing approach increases the amount of trust that
+moderators place onto the curators of moderation policy lists. As watching a
+list gives the list curators the ability to manage members within the
+subscribing community.
+
+For example: `luna`, the moderator of `bat-community.example.com` notices that
+`cat-community.example.com` is watching the list, and uses their power to ban
+more of `charity`'s friends.
+
+These problems lead to situations where communities avoid watching other
+moderation policy lists entirely, either because of a few policies that they
+take issue with, or it is simply unsafe to do so because of the level of trust
+placed on the curators of a list when watching them.
+
+## Proposal
+
+We introduce a new moderation policy meta rating event type,
+`m.policy.rule.approval`. This policy exists purely for moderators to express
+their approval or disapproval for policies in other lists.
+
+Content Schema:
+
+```yaml
+properties:
+  rating:
+    description: |-
+	  A binary rating of either string literals approve or disapprove.
+    type: string
+	enum:
+	  - approve
+      - disapprove
+  event_id:
+    type: string
+    description: |-
+      The event ID of the moderation policy that this is a rating for.
+    example: "$some_event_id"
+type: object
+required:
+  - rating
+  - event_id
+```
+
+Example:
+
+```json
+{
+  "sender": "@luna:bat-community.example.com",
+  "type": "m.policy.rule.approval",
+  "content": {
+    "rating": "approve",
+    "event_id": "$wioefjwoijefo:example.com"
+  },
+  "state_key": "lciRM4XTjq9bj3QMawAbXFH7pCRJEvft34ZmhGBNsxc="
+}
+```
+
+### Discussion
+
+These policies allow for moderation tools to develop _approval
+only_[^approval-only] policy list subscription modes. Whereby policies are only
+considered from a list when they have been approved by a moderator. Please note,
+that due to the _lazy_ nature of matching, the workload on moderators for
+approving policies when watching a list will be minimal[^minimal-workload].
+
+Additionally, subscriptions that use _direct propagation_ can remove policies
+from consideration that have been _disapproved_ by a moderator.
+
+[^approval-only]: https://github.com/the-draupnir-project/planning/issues/4
+
+[^minimal-workload]:
+    They will only have to approve policies that are matching encountered
+    entities. For example, there is no need to approve a policy banning a user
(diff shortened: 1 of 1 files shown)


proposals/4273-approval-disapproval.md
tulir, line 1: Implementation requirements:

* [ ] Client/bot sending approvals
* [ ] Client/bot using approvals

turt2live, line 1: Implementation requirements:

* [ ] Sender
* [ ] Receiver

turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4273#discussion_r1992398325

turt2live, line 1: Overarching question: does this need to be a protocol feature? In the early design of policy lists there was a theory of "multiple confirmations" and manual approval, where the bot wouldn't honour the policy list recommendations until either multiple lists added the same rule, or the operator accepted it. 

This is also why the spec doesn't say that recommendations *must* be followed: the lists are just sources of information. What the parser/receiver does with that information is entirely left to the implementation.

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,140 @@
...
+propagates the approval ratings, then their tool will enact policies from
+`#people-i-hate:bat-community.example.com`.
+
+#### Mitigation
+
+To prevent this, implementations SHOULD use an allowlist of senders when
+considering `m.policy.rule.approval` ratings by default, rather than the
+specifics of a policy list subscription profile.
+
+## Alternatives
```
tulir, line 121: Meowlnir implements ordered policy lists and "unban" policies, so that if an unban policy is found before a ban policy, the unban takes precedence and the ban is ignored. It also has the benefit of not requiring an existing policy to disapprove of.

tulir: For the positive approval use case, bots could just listen to a policy list without applying it, thus allowing the admin to copy approved rules over to a list which *is* applied. Meowlnir has a `dont_apply` flag on watched lists for that purpose

Gnuxie: I've added anti-ban [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4273/files#diff-0bad88e243ad9c016176080fed8cbbca573549c2b90926bb918b1e95aad541a6R129). I think principally the difference is i want these ratings to ultimately be used in recommender systems. So that in the future there can be policy discovery... ie suggestions for bans that you are from lists not being watched at all.

proposals/4273-approval-disapproval.md
Gnuxie: This doesn't seem to be an overarching question, but a question you have about the proposal?

I am unaware of the early work around moderation policy lists. This proposal has been on the cards for a long time now and there is a lot of potential for this proposal to be used beyond moderation purposes as it is generic.

This does need to be documented somewhere for the purpose of interoperability between moderation tools and matrix clients, and the MSC process currently is the best avenue for that.

Gnuxie: > This is also why the spec doesn't say that recommendations must be followed: the lists are just sources of information. What the parser/receiver does with that information is entirely left to the implementation.

Likewise, they are free to do what they want with approval ratings in this proposal

proposals/4273-approval-disapproval.md
Gnuxie: (whoops i didn't see this comment https://github.com/matrix-org/matrix-spec-proposals/pull/4273#discussion_r1992404427)

> copy approved rules over to a list which is applied

This would remove the original context for the policy, including who sent it. I will add this to the alternatives but it is my assessment that this is a very risky propagation strategy and i advise against it in the strongest terms.

Gnuxie: Added all this info the the alternatives section. Please feel free to add new comments there ^^

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,195 @@
...
+	  A binary rating of either string literals approve or disapprove.
+    type: string
+	enum:
+	  - approve
+      - disapprove
+  event_id:
+    type: string
+    description: |-
+      The event ID of the moderation policy that this is a rating for.
+    example: "$some_event_id"
```
Gnuxie, line 54: CC @deepbluev7 

We should instead use the `room_id`, the event `type` and the event `state_key`. As the `state_key` + `type` combination are used to identify a policy within moderation policy lists so that they can be updated or removed.

Gnuxie: On the other hand, without the `event_id` or `entity` the policy can be modified after the fact once agreement has been met to target a different entity that you disagree with

Gnuxie: An approach to this in implementation could be simply to notify that a policy has changed, and ask again for a new rating.

Gnuxie: I'll update the MSC with that recommendation.

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,195 @@
...
+friend.
+
+Addtionaly, this all-or-nothing approach increases the amount of trust that
+moderators place onto the curators of moderation policy lists. As watching a
+list gives the list curators the ability to manage members within the
+subscribing community.
+
+For example: `luna`, the moderator of `bat-community.example.com` notices that
+`cat-community.example.com` is watching the list, and uses their power to ban
+more of `charity`'s friends.
```
Gnuxie, line 26: @deepbluev7 says to just use "mine" and "third party" rather than all these names.

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,195 @@
...
+
+```yaml
+properties:
+  rating:
+    description: |-
+	  A binary rating of either string literals approve or disapprove.
+    type: string
+	enum:
+	  - approve
+      - disapprove
```
Johennes, line 49: I think spaces (or tabs) got slightly out of order here.

```suggestion
    description: |-
      A binary rating of either string literals approve or disapprove.
    type: string
    enum:
      - approve
      - disapprove
```

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,195 @@
...
+`m.policy.rule.approval`. This policy exists purely for moderators to express
+their approval or disapproval for policies in other lists.
+
+Content Schema:
+
+```yaml
+properties:
+  rating:
+    description: |-
+	  A binary rating of either string literals approve or disapprove.
```
Johennes, line 45: What does "binary" refer to here?

proposals/4273-approval-disapproval.md
```diff
@@ -0,0 +1,195 @@
...
+And yet on networks that follow copy propagation, bans are often permanent and
+indefinite. Even after explicit retraction. Moderators are incentivised to keep
+copies active out of an abundance of caution and to create less work.
+
+For this reason, we strongly recommend against using copy propgation not just as
+a strategy for approval, but in all circumstances.
+
+## Security considerations
+
+See ptoential issues.
```
Johennes, line 186:
```suggestion
See potential issues.
```

proposals/4273-approval-disapproval.md
Gnuxie, line 54: What if i want to mark other entities as approved or disapproved? Such as a room or a user?

PR #4274, opened 2025-03-18
MSC4274: Inline media galleries via msgtypes

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/msgtype-galleries/proposals/4274-inline-media-galleries.md)

---

I am a Systems Architect at [gematik](https://www.gematik.de/), Software Engineer at [Filament](https://filament.dm/) and [Unomed](https://www.unomed.ch/), Matrix community member and former [Element](https://element.io/) employee. This proposal was written and published with my Filament hat on.

### DIFF
--- a/proposals/4274-inline-media-galleries.md
+++ b/proposals/4274-inline-media-galleries.md
@@ -54,7 +54,7 @@ the msgtypes `m.image`, `m.video`, `m.audio` and `m.file`. Rather than
 
 ## Potential issues
 
-The size of galleries under this proposal is limitted by the event size limit
+The size of galleries under this proposal is limited by the event size limit
 (65,536 bytes). An encrypted gallery event with a single image, including
 thumbnail, and no caption measures about 2,000 bytes. Each additional image adds
 roughly another 1,000 bytes. Therefore galleries would be capped to about 60
--- a/proposals/4274-inline-media-galleries.md
+++ b/proposals/4274-inline-media-galleries.md
@@ -1,0 +1,111 @@
+# MSC4274: Inline media galleries via msgtypes
+
+Matrix allows sharing individual media files such as images or videos via
+dedicated [`m.room.message` msgtypes]. This is, however, not practical for
+sharing larger collections of files. For one thing, there is no mechanism to set
+a joint caption for several media files. For another, there is no way to inform
+other clients that a set of pictures belongs together and should be grouped
+accordingly in the timeline. These shortcomings call for a built-in media
+gallery feature as it is widely known from other messaging apps.
+
+This proposal defines a compact and pragmatic variant of sharing media galleries
+in a single event.
+
+## Proposal
+
+A new `msgtype` of `m.gallery` is introduced with the following properties in
+`content`:
+
+- `body` (string): The caption of the gallery.
+- `format` (string): The format used in `formatted_body`. Currently only
+  `org.matrix.custom.html` is supported.
+- `formatted_body` (string): The formatted version of the body. Required if
+  `format` is specified.
+- `itemtypes` (array): Ordered array of metadata for each item in the gallery.
+
+The objects in the `itemtypes` array have the same schema as the `content` of
+the msgtypes `m.image`, `m.video`, `m.audio` and `m.file`. Rather than
+`msgtype`, they use `itemtype` though.
+
+``` json5
+{
+  "type": "m.room.message",
+  "content": {
+    "msgtype": "m.gallery",
+    "body": "Checkout my photos from [FOSDEM 2025](https://fosdem.org/2025/)",
+    "format": "org.matrix.custom.html",
+    "formatted_body": "Checkout my photos from <a href=\"https://fosdem.org/2025/\">FOSDEM 2025</a>",
+    "itemtypes": [{
+      "itemtype": "m.image",
+      "body": "filename.jpg",
+      "info": {
+        "h": 398,
+        "mimetype": "image/jpeg",
+        "size": 31037,
+        "w": 394
+      },
+      "url": "mxc://example.org/JWEIFJgwEIhweiWJE",
+      ...
+    }, ...]
+  },
+  ...
+}
+```
+
+## Potential issues
+
+The size of galleries under this proposal is limitted by the event size limit
+(65,536 bytes). An encrypted gallery event with a single image, including
+thumbnail, and no caption measures about 2,000 bytes. Each additional image adds
+roughly another 1,000 bytes. Therefore galleries would be capped to about 60
+images. While this is a limitation, the amount seems practical for most use
+cases.
+
+Additionally, this proposal doesn't prescribe a fallback mechanism for clients
+that don't support `m.gallery` msgtypes. Clients that care about backwards
+compatibility could encode media links or a generic fallback text into `body` or
+`formatted_body`, respectively. It is unclear, however, if a fallback is
+actually desirable or not because unwrapping 60 images into somebody else's
+timeline might be considered inappropriate if not spammy.
+
+## Alternatives
+
+This proposal is similar to the seemingly abandoned [MSC3382]. The latter
+conflates galleries into `m.text` (or even arbitrary) msgtypes and reuses
+`msgtype` inside the items which creates a circular dependency. These issues are
+solved in the present proposal.
+
+Rather than adding a new `msgtype`, galleries could be expressed via extensible
+events by reusing parts of [MSC1767] and related proposals. This would only take
+effect in a future room version, however, and can be covered in a separate MSC.
+
+Another alternative is to define galleries as chains of related events as in
+[MSC2881]. Compared to the inline approach taken in this proposal, this has
+better fallback support, is more flexible and less prone to event size limits.
+At the same time, it is significantly more complex to implement in terms of
+replies, edits, redactions, forwarding, etc.
+
+Finally, rather than defining galleries on the sender's side, receiving clients
+could opportunistically group consecutive images in the timeline into galleries.
+This might need specific rules such as grouping by sender or time windowing.
+Different clients might prefer very different display rules, however. Therefore,
+it seems more practical that the spec annotates galleries clearly while leaving
+their UI treatment to clients themselves.
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+Until this proposal is accepted into the spec, implementations SHOULD refer to
+`m.gallery` as `dm.filament.gallery`.
+
+## Dependencies
+
+None.
+
+  [`m.room.message` msgtypes]: https://spec.matrix.org/v1.13/client-server-api/#mroommessage-msgtypes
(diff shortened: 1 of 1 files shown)


proposals/4274-inline-media-galleries.md
turt2live, line 1: Implementation requirements:

* [ ] Client

proposals/4274-inline-media-galleries.md
```diff
@@ -0,0 +1,111 @@
...
+
+A new `msgtype` of `m.gallery` is introduced with the following properties in
+`content`:
+
+- `body` (string): The caption of the gallery.
+- `format` (string): The format used in `formatted_body`. Currently only
+  `org.matrix.custom.html` is supported.
+- `formatted_body` (string): The formatted version of the body. Required if
+  `format` is specified.
+- `itemtypes` (array): Ordered array of metadata for each item in the gallery.
```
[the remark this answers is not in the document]
Johennes: Hm, I think I landed on this because `itemtypes` is basically the generalization of `msgtype` plus the associated data into an array. I agree that `items` would be clearer though.

ToddCrimson:
Just checking. Is this feature now implemented and available?

Half-Shot:
> Just checking. Is this feature now implemented and available?

This is the specification of the feature. See [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4274#discussion_r2001771722) for any clients that are declaring support for the feature in it's experimental state. I don't see any *client* implementations yet.

ToddCrimson:
> > Just checking. Is this feature now implemented and available?
> 
> This is the specification of the feature. See [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4274#discussion_r2001771722) for any clients that are declaring support for the feature in it's experimental state. I don't see any _client_ implementations yet.

Thanks, OK. So looks like the feature set has been merged in, just still no client that is supporting it yet.

proposals/4274-inline-media-galleries.md
bxdxnn: * Element X Android: https://github.com/element-hq/element-x-android/pull/6519
* Element Web: https://github.com/element-hq/element-web/pull/33111

proposals/4274-inline-media-galleries.md
```diff
@@ -0,0 +1,111 @@
...
+other clients that a set of pictures belongs together and should be grouped
+accordingly in the timeline. These shortcomings call for a built-in media
+gallery feature as it is widely known from other messaging apps.
+
+This proposal defines a compact and pragmatic variant of sharing media galleries
+in a single event.
+
+## Proposal
+
+A new `msgtype` of `m.gallery` is introduced with the following properties in
```
MurzNN, line 16: If we are taking in the scope not only images, but also files (`m.file`), why do we call this a "gallery"? 

From my point of view, the best term for this feature is "attachments", because actually on the technical side we have a regular `m.message` with just some attached entities (images, videos, files, anything else).  

And the term "gallery" is related more to the displaying style, not to the message type, because a message with 5 attached `.zip` files will be hard to display as a gallery ;) Even the same "5 images" can be displayed as a list of file names, not as a gallery, if we have a text-based UI.

So, naming this not as a "gallery" but using a more general term, will extend the scope of such message type, and will allow, in the future, to put into `itemtypes` other interesting things like references to other messages, multiple forwarded messages, etc.

What do you think about this idea?

proposals/4274-inline-media-galleries.md
```diff
@@ -0,0 +1,111 @@
...
+Another alternative is to define galleries as chains of related events as in
+[MSC2881]. Compared to the inline approach taken in this proposal, this has
+better fallback support, is more flexible and less prone to event size limits.
+At the same time, it is significantly more complex to implement in terms of
+replies, edits, redactions, forwarding, etc.
+
+Finally, rather than defining galleries on the sender's side, receiving clients
+could opportunistically group consecutive images in the timeline into galleries.
+This might need specific rules such as grouping by sender or time windowing.
+Different clients might prefer very different display rules, however. Therefore,
```
MurzNN, line 91: I think we should clarify here the displaying rules: where the "body" text should be displayed - above the attached items or below them, or maybe add a field that holds the display style.

Because this is crucial from the user point of view:

- If we display the body above, it will be the main "body" of the message, and the attached images are just an addition (Slack style).

- If we display the body below, the main content of the message will be files, and the "body" will be just a commenting text, in addition to the images (Telegram style).

Both variants of rendering are possible and available in different messengers, but they bear completely different meanings of the message.

MurzNN: I, personally, am a fan of the Slack style, not Telegram style ;)

Just imagine a message:
Variant 1: 
```
Checkout my photos from FOSDEM 2025:
[photo 1]
[photo 2]
[photo 3]
```

Variant 2:
```
[photo 1]
[photo 2]
[photo 3]
Checkout my photos from FOSDEM 2025:
```

The order of things in variant 1 looks more correct, right?

g-work:
In my opinion, this is (from a non-technical point of view) the difference between attachment(s) and gallery.

For a gallery I'd prefer Variant 2

[photo 1]
[photo 2]
[photo 3]
What a nice day at FOSDEM 2025!

I think this is what non professional users expect from a gallery. As far as I remember, Telegram even allows to add captions per image, which is very useful to describe i.e. holiday pictures.

For a bunch of files I'd prefer Variant 1
Here are the files concerning [whatever]:
[file.zip]
[file.odt]
[file.pdf]

proposals/4274-inline-media-galleries.md
MurzNN: Actually, these are two totally different things from the logical point of view:
- Above files - is the "message body" (main content).
- Below files - is the "message caption" (secondary thing: signature, comment, subscription, not the main content).

And if we keep the text named as "body", we should put it above. Or, rename to "caption" then. 

Or, maybe even keep both ones ;) with recommendations what to render where.

MurzNN:
By the way, there is also an MS Teams / Skype style, which is very convenient! Users just type the message text, and can insert images (and any other files) inline directly in the text, at the top, on the bottom, anywhere!

So, this mode meets both approaches - images as annotation to the message text (Slack style) and the message text as annotation to images (Telegram style). 

And Matrix, actually, already supports this mode without any changes! We can insert media as links inside the text using references like this:
```
Check out my photos from FOSDEM 2025:
<img src=\"mxc://example.org/JWEIFJweifjWIeifj\" />" 
<img src=\"mxc://example.org/JWEIFJweifjWIeifk\" />" 
<img src=\"mxc://example.org/JWEIFJweifjWIeifl\" />" 

And especially look at this awesome group photo: 
<img src=\"mxc://example.org/JWEIFJweifjWIeifm\" />" 

What a nice day at FOSDEM 2025!
```

A similar approach of inserting attachments (inline as a link) is used in GitHub, GitLab, and many other services.

So, seems to implement this feature, we don't need any server-side changes; this can be implemented fully on the client side (message composer to handle inserting, and renderer to display thumbnails and popup galleries), right?

proposals/4274-inline-media-galleries.md
MurzNN: So, about the "keep both ones", see my idea in the comment https://github.com/matrix-org/matrix-spec-proposals/pull/4274#issuecomment-4368701640 - I think this will be the best solution to cover all cases with minimal sever-side changes.

MurzNN:
Okay, let me try with a separate MSC describing a universal galleries approach using inline MXC links: https://github.com/matrix-org/matrix-spec-proposals/pull/4475 - who likes the inline media approach, let's move there to discuss!

proposals/4274-inline-media-galleries.md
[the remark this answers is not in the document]
tulir: `items` would definitely be better. Since there seem to be implementations, the unstable prefix could be kept as `itemtypes` and the stable one changed to `items`.

proposals/4274-inline-media-galleries.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Proposal
+
+A new `msgtype` of `m.gallery` is introduced with the following properties in
+`content`:
+
+- `body` (string): The caption of the gallery.
+- `format` (string): The format used in `formatted_body`. Currently only
+  `org.matrix.custom.html` is supported.
+- `formatted_body` (string): The formatted version of the body. Required if
```
tulir, line 22: Behavior of captions inside items should probably be mentioned too. I think it'd make sense to allow them, but with the caveat that clients may choose to only display them when the user expands a specific image (so not in the timeline).

proposals/4274-inline-media-galleries.md
nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/3926de367e268f87c10f77e8f162f7dc44b27d1d

Velin92: Element X iOS:
- https://github.com/element-hq/element-x-ios/pull/5928
- https://github.com/element-hq/element-x-ios/pull/5929
- https://github.com/element-hq/element-x-ios/pull/5932
- https://github.com/element-hq/element-x-ios/pull/5943
- https://github.com/element-hq/element-x-ios/pull/5955
- https://github.com/element-hq/element-x-ios/pull/5957

manuroe:
@Johennes gave me access to his branch so that I can contribute to this MSC (thanks!). I am doing it with my Engineering Manager hat at Element.

We internally discussed the relations-based alternative. I added a paragraph about our conclusions in https://github.com/matrix-org/matrix-spec-proposals/pull/4274/commits/d01e4e0c4eccd0c00c9639a9ea123ad9ed6ffb72. Our approach was from a UX perspective to make the feature as nice as possible for the end users.

proposals/4274-inline-media-galleries.md
```diff
@@ -0,0 +1,119 @@
...
+  `format` is specified.
+- `itemtypes` (array): Ordered array of metadata for each item in the gallery.
+
+The objects in the `itemtypes` array have the same schema as the `content` of
+the msgtypes `m.image`, `m.video`, `m.audio` and `m.file`. Rather than
+`msgtype`, they use `itemtype` though.
```
tulir, line 28: Why? Seems like a pretty unnecessary special case to rename a single field

proposals/4274-inline-media-galleries.md
tulir: Limiting to images and videos would probably make more sense than using a less descriptive name. Multiple other files in a single message are much less common, so it'd also need clients to invent new UI (not sure if any other chat app allows them)

krille-chan: Telegram also supports sending a list of files of different types in a single message bubbe:
grafik

 I would more argue that we should rename this to `m.files` which then could deprecate other message types like `m.file`, `m.audio`, `m.video` and `m.image` as those would just be a `m.files` with only one item in the list.

PR #4276, opened 2025-03-20
MSC: Soft unfailure for self redactions

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/soft-unfailure/proposals/4275-soft-unfail-redactions.md)

Author: @matrix-org/trust-safety 
Shepherd: @matrix-org/trust-safety 

Conflict of interest disclosure: Many members of the T&S team are Element employees and may serve additional roles outside their primary responsibility. Not all members of the T&S team are publicly known and disclosing all conflicts may reveal their identities - they have been excluded for this reason.

### DIFF
--- a/proposals/4275-soft-unfail-redactions.md
+++ b/proposals/4275-soft-unfail-redactions.md
@@ -1,0 +1,48 @@
+# MSC4275: Soft unfailure for self redactions
+
+When a user is removed from a room, the server may issue several redactions to their messages to clean
+up rooms. Users may also use similar functionality, if supported by their server, to remove their own
+messages after being removed.
+
+These self redactions can end up being [soft failed](https://spec.matrix.org/v1.13/server-server-api/#soft-failure)
+due to the authorization rules preventing the redaction event from being validated, despite being part of the
+DAG at a legal point.
+
+This proposal suggests that servers be less strict about soft failing self-redactions in particular.
+
+## Proposal
+
+When evaluating for soft failure, servers SHOULD permit `m.room.redaction` events which target events
+by the same `sender` to go through normally, avoiding soft failure. Servers MAY impose limits on how
+many events can bypass soft failure. Limits similar to the following may be of interest to developers:
+
+* Only the first 300 redactions from the same `sender` bypass soft failure.
+* Only the first redaction targeting a given event bypasses soft failure. "Duplicate" redactions are
+  subject to soft failure.
+* Only the redactions received within 1 hour of the most recent membership event change can bypass
+  soft failure.
+
+Applying some or all of the conditions may help servers avoid event spam abuse from making it to
+local users over `/sync`.
+
+## Potential issues
+
+This may allow banned users to spam rooms with redaction events - the limits proposed above are potential
+ways to mitigate the issue.
+
+## Alternatives
+
+Another approach could be to modify auth rules to exempt same-sender `m.room.redaction` events from the requirement
+to pass authorization at the current resolved state. This approach may not work well with [how redactions work](https://spec.matrix.org/v1.13/rooms/v11/#handling-redactions).
+
+## Security considerations
+
+See Potential Issues.
+
+## Unstable prefix
+
+Not applicable.
+
+## Dependencies
+
+This proposal has no direct dependencies.


proposals/4276-soft-unfail-redactions.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4276-soft-unfail-redactions.md
```diff
@@ -0,0 +1,48 @@
+# MSC4276: Soft unfailure for self redactions
+
+When a user is removed from a room, the server may issue several redactions to their messages to clean
+up rooms. Users may also use similar functionality, if supported by their server, to remove their own
+messages after being removed.
```
Gnuxie, line 5: What is the context for a remote server or remote user redacting their own messages?

Gnuxie, line 5: What does `remove` mean explicitly? Does it mean kicked, banned or both?

Gnuxie, line 5: If it means kicked or banned, then is good faith being assumed in either remote users and servers?

proposals/4276-soft-unfail-redactions.md
```diff
@@ -0,0 +1,48 @@
...
+
+When evaluating for soft failure, servers SHOULD permit `m.room.redaction` events which target events
+by the same `sender` to go through normally, avoiding soft failure. Servers MAY impose limits on how
+many events can bypass soft failure. Limits similar to the following may be of interest to developers:
+
+* Only the first 300 redactions from the same `sender` bypass soft failure.
+* Only the first redaction targeting a given event bypasses soft failure. "Duplicate" redactions are
+  subject to soft failure.
+* Only the redactions received within 1 hour of the most recent membership event change can bypass
+  soft failure.
```
Gnuxie, line 23: What tooling is expected to be deployed by remote users or servers such that they send redactions within 1 hour?

proposals/4276-soft-unfail-redactions.md
```diff
@@ -0,0 +1,48 @@
...
+
+Applying some or all of the conditions may help servers avoid event spam abuse from making it to
+local users over `/sync`.
```
Gnuxie, line 26: This is what the proposal is really about. And yet this is the only context about why this is a proposed solution.

proposals/4276-soft-unfail-redactions.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Potential issues
+
+This may allow banned users to spam rooms with redaction events - the limits proposed above are potential
+ways to mitigate the issue.
+
+## Alternatives
+
+Another approach could be to modify auth rules to exempt same-sender `m.room.redaction` events from the requirement
+to pass authorization at the current resolved state. This approach may not work well with [how redactions work](https://spec.matrix.org/v1.13/rooms/v11/#handling-redactions).
```
Gnuxie, line 36: [MSC4194](https://github.com/matrix-org/matrix-spec-proposals/pull/4194) is an alternative that does not require good faith in remote users and servers.

proposals/4276-soft-unfail-redactions.md
Gnuxie: Ok, I guess the context is something like this: a user has registered on a server for use as a spam/brigade throwaway. This user has been banned by a bunch of remote servers. The server admins discover the throwaway and send redactions from the same user (hijacking the account) to clean and try undo some of the damage. In the rooms where the user has been banned, the redactions will have to use prior state to authorize and soft fail.

So I guess "good faith" here means both discovery of the user throwaway by server admins, and then also their cooperation. But that's not the right way to frame the situation. This is really a proposal to assist server admins in cleaning up accounts that have violated their tos.

turt2live: Some servers apply redactions as an erasure technique.

Gnuxie, line 5: I'd change the framing of the whole MSC to be from the perspective of a server admin decommissioning a user that was used for abuse. Because currently the MSC is written as though it is doing a service for room level moderation and this is really misleading.

turt2live: I'm not following the concern here, sorry.

proposals/4276-soft-unfail-redactions.md
turt2live: This is to deal with possible federation delays, not tooling delays. The redactions may take a little while to send.

proposals/4276-soft-unfail-redactions.md
turt2live: This line is trying to prevent redactions from causing client performance concerns. If there's an actionable comment here, please open a new thread.

proposals/4276-soft-unfail-redactions.md
turt2live: MSC4194 requires this MSC in order to work - it's not an alternative.

proposals/4276-soft-unfail-redactions.md
turt2live: The level of context currently feels appropriate to me, as there are a variety of reasons why the server might issue redactions for their users beyond abuse. If you have concrete suggestions, please use the inline suggestions tool or a PR if there's a lot of them.

jimmackenzie: This is very useful feedback - the intent was always that this is a tool for server administrators, and that complementary mechanisms such as [MSC4194](https://github.com/matrix-org/matrix-spec-proposals/pull/4194) could exist for the room moderator perspective. We'll look at clarifying that.

proposals/4276-soft-unfail-redactions.md
Gnuxie: No it does not. MSC4194 is for use by room moderators, using the room moderator to send redactions, not the target user.

proposals/4276-soft-unfail-redactions.md
Gnuxie: I guess the problem i'm trying to communicate to you is that if the user parts from a room an hour before the server admin responsible for them discovers that the account was used for abuse then the redactions will be soft failed.

proposals/4276-soft-unfail-redactions.md
Gnuxie: But why are the servers erasing messages? Can this please be added to the context in the MSC?

proposals/4276-soft-unfail-redactions.md
TheArcaneBrony: Additionally, not all server admins respond within 1 hour, the user may already have been banned hours and hours before addressed by a server admin, potentially leaving soft failed media (hi, race conditions) in place for other servers in the room.

proposals/4276-soft-unfail-redactions.md
```diff
@@ -0,0 +1,48 @@
...
+
+When a user is removed from a room, the server may issue several redactions to their messages to clean
+up rooms. Users may also use similar functionality, if supported by their server, to remove their own
+messages after being removed.
+
+These self redactions can end up being [soft failed](https://spec.matrix.org/v1.13/server-server-api/#soft-failure)
+due to the authorization rules preventing the redaction event from being validated, despite being part of the
+DAG at a legal point.
+
+This proposal suggests that servers be less strict about soft failing self-redactions in particular.
```
Saiv46, line 11: I think servers should be **more** strict about self-redactions due to legal requirements, room policy (it would _suck_ if someone will remove useful information or frame someone by removing context), or moderation purposes (reporting messages after ban, machine learning, etc).

Saiv46 (changes requested):

turt2live:
While this solution works, it opens up doors to undesirable effects:

* This is effectively partial soft failure, which may be tricky to get right or reliably maintain.
* Banned/left users can still participate in the room indefinitely, even if it is just through redactions.
* Authorization rules may be affected, and thus may require a new room version, slowing down adoption of this behaviour.

For the specific use case of cleaning up spam in a room, especially after a ban, MSC4194-style endpoints may wish to support using a room moderator's account instead of the subject user to send redactions. MSC4293 may also be of interest to some communities. For the cases where the server is trying to be a good citizen and cleaning up spam in all rooms the user posted to, the server may wish to withhold the leave event (from deactivation or similar) until the user's server-sent self redactions are fully processed. 

I'm closing this MSC in favour of those alternative approaches.

PR #2414, opened 2020-01-18
MSC2414: Make reason and score parameters optional for reporting content

iinuwa:
[Rendered](https://github.com/iinuwa/matrix-doc/blob/master/proposals/2414-optional-content-reporting-reason.md)
Signed-off-by: Isaiah Inuwa <isaiah.inuwa@gmail.com>

PR #919, opened 2017-05-18
Add GeoJSON & accuracy to m.location.

Half-Shot:

PR #4277, opened 2025-03-21
MSCXXXX: Harmonizing the reporting endpoints

Johennes:
[Rendered]()

---

I am a Systems Architect at [gematik](https://www.gematik.de/), Software Engineer at [Filament](https://filament.dm/) and [Unomed](https://www.unomed.ch/), Matrix community member and former [Element](https://element.io/) employee. This proposal was written and published with my gematik hat on.


proposals/4277-reporting-harmonization.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,77 @@
...
+    it either.
+
+These differences seem unnecessary and were likely introduced by accident only.
+The present proposal, therefore, seeks to align the three endpoints.
+
+## Proposal
+
+On all three endpoints:
+
+1.  The `reason` parameter is made optional.
```
turt2live, line 34: `reason` is made required on rooms and users to limit spam, and is deliberate. It's expected that a future MSC will require it to be non-empty too (possibly combined with other not-yet-described fields).

turt2live (changes requested):

proposals/4277-reporting-harmonization.md
turt2live: ideally, yes, but it's harder to go optional -> required than it is required -> optional

I think we'd make different decisions now than we would have 5 years ago.

proposals/4277-reporting-harmonization.md
Johennes: 200 regardless of event / room existence: https://github.com/element-hq/synapse/pull/18263

I'm unsure if the random delay / constant time part requires an implementation. The spec already mentions this for the event reporting endpoint but Synapse doesn't appear to implement it and it's not part of https://github.com/element-hq/synapse/pull/18120 either.

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,77 @@
...
+
+2.  Servers MAY add a random delay or use constant time functions when
+    processing responses to combat enumeration attacks.
+
+All of these changes appear applicable regardless of the reported subject and it
+is not clear why the spec should differentiate the endpoints here.
+
+## Potential issues
+
+The `reason` field on the room and user reporting endpoints is currently
```
Half-Shot, line 47: Honestly, I find it bizarre that reason is required but may be empty string. I suppose this might not be the MSC to change it, but I'd rather we just made it optional.

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,74 @@
+# MSC4277: Harmonizing the reporting endpoints
+
+Matrix currently has three reporting endpoints:
+
+- [`/_matrix/client/v3/rooms/{roomId}/report/{eventId}`] ([added] in Matrix
```
Half-Shot, line 5: This also still has a `score` field, which I've always regarded as bonkers because I've never seen any client attempt to ask the user "how offensive is this from a score of -100 to 0"

proposals/4277-reporting-harmonization.md
Johennes: I think I understand the value of requiring it to be filled for spam protection. The combination of required but allowed to be empty strikes me as slightly odd, too, but it's hard to judge without knowing T&S's future plans for the field.

Generally though, I suppose if we later go from required but possibly empty to required and not empty, it'll be just as much a breaking change for clients as if we went from optional to required and not empty.

proposals/4277-reporting-harmonization.md
Johennes: Oh, good point. This was made optional with https://github.com/matrix-org/matrix-spec-proposals/pull/2414. Given that this proposal landed in 2020 and the field is missing in the other two endpoints, I suppose we could safely remove it now.

proposals/4277-reporting-harmonization.md
Half-Shot: It's hard to say. Even if the field was required people could just keyboard smash the field. I'm not really going to die on this hill though, I just think we should unify it in one direction.

My feeling for T&S that any report is a useful report, and that tooling is smart enough these days that we can make use of the room_id / event_id to do preliminary investigations even if the `reason` context isn't useful.

proposals/4277-reporting-harmonization.md
Half-Shot: I would say so, on the grounds that I don't think any client has ever offered users a choice on the value. IIRC Element Web doesn't even fill the field.

Johennes: Added into the proposal with b3ac83f70844391f7d82d2c1cabfbe515dc63c38.

proposals/4277-reporting-harmonization.md
Johennes: I've added the removal of `score` into https://github.com/element-hq/synapse/pull/18263. Synapse didn't do anything else than assert the type anyway.

timedoutuk: Implemented in continuwuity [#919](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/919).

turt2live:
This MSC has been put forward as potentially ready for FCP. T&S is aware of some features which may need mirroring to the trio of endpoints, and may want to include them in this MSC (the MSC hasn't been reviewed to see if they're already addressed).

Next steps are for T&S to review the MSC, then send it for SCT review. Feel free to ping me on either if the process takes too long.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
This seems ready for broader review. With my T&S hat, these changes are non-conflicting with other reporting endpoint changes we intend to make later in the year.

@mscbot fcp merge

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,82 @@
...
+  0.4.0, no MSC known)
+- [`/_matrix/client/v3/rooms/{roomId}/report`] ([added][1] in Matrix 1.13 as per
+  [MSC4151])
+- `/_matrix/client/v3/users/{userId}/report` (to be [added][2] to Matrix as per
+  [MSC4260])
+
+The spec contains a number of subtle differences for these endpoints:
+
+1.  The user reporting endpoint [allows] servers to respond with 200 even if the
```
dbkr: as a minor markdown nit, 'allows' doesn't seem like a great identifier for a link so this might be better inline?

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,82 @@
...
+2.  Servers MAY add a random delay or use constant time functions when
+    processing responses to combat enumeration attacks.
+3.  The `score` property is removed.
+
+All of these changes appear applicable regardless of the reported subject and it
+is not clear why the spec should differentiate the endpoints here.
+
+## Potential issues
+
+Some Clients may send `score` on event reports today. These clients are not
```
dbkr:
```suggestion
Some clients may send `score` on event reports today. These clients are not
```

proposals/4277-reporting-harmonization.md
Johennes: This links to the respective part of the MSC because it hadn't yet landed at the time of writing. Now that it was merged, we can actually just link to the spec from the endpoint in the introduction and don't need this link here anymore. Have made this change in 2b8172f5180c0d9708f9eedab40f47fab43d3e7c.

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,82 @@
...
+  [MSC4260])
+
+The spec contains a number of subtle differences for these endpoints:
+
+1.  The user reporting endpoint allows servers to respond with 200 even if the
+    user doesn't exist to deny enumerating users. This option is not allowed in
+    the event and room reporting endpoints.
+2.  The user and event reporting endpoints allow servers to add a random delay
+    when generating responses. This is a more sophisticated measure against
+    enumeration attacks. While the spec doesn't explicit forbid this technique
```
uhoreg: nit:

```suggestion
    enumeration attacks. While the spec doesn't explicitly forbid this technique
```

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,82 @@
...
+`score` property.
+
+## Alternatives
+
+None.
+
+## Security considerations
+
+Enumeration attacks are likely more severe for users than for rooms or events.
+There are still not irrelevant, however.
```
uhoreg: nit:

```suggestion
They are still not irrelevant, however.
```

proposals/4277-reporting-harmonization.md
```diff
@@ -0,0 +1,82 @@
+# MSC4277: Harmonizing the reporting endpoints
+
+Matrix currently has three reporting endpoints:
+
+- [`/_matrix/client/v3/rooms/{roomId}/report/{eventId}`] ([added] in Matrix
+  0.4.0, no MSC known)
```
richvdh, line 6: it was MSC739, fwiw. That's what MSCs used to look like, back in the day.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2311

turt2live:
Merged 🎉

PR #2448, opened 2020-02-27
MSC2448: Using BlurHash as a Placeholder for Matrix Media

anoadragon453:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/anoa/blurhash/proposals/2448-blurhash-for-media.md)

Implementations (and iterations):

* [Ruma](https://github.com/ruma/ruma/pull/363)
* [matrix-media-repo](https://github.com/matrix-org/matrix-doc/pull/2448#issuecomment-593061702)
* [Nheko](https://github.com/Nheko-Reborn/nheko/pull/137)
* [Element Web](https://github.com/matrix-org/matrix-react-sdk/pull/5099)
* https://github.com/matrix-org/matrix-react-sdk/pull/5099
* https://github.com/turt2live/matrix-media-repo/pull/327
* https://github.com/turt2live/matrix-media-repo/pull/328
* https://github.com/turt2live/matrix-sticker-manager/pull/22
* https://github.com/turt2live/matrix-dimension/pull/420

**[FCP status comment](https://github.com/matrix-org/matrix-doc/pull/2448#issuecomment-811983072)**

PR #4278, opened 2025-03-25
MSC4278: Media preview controls

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/media-preview-switch/proposals/4278-media-preview-controls.md)

Implementations:
 - Element Web https://github.com/element-hq/element-web/pull/29582 complete
 - Element X Android https://github.com/element-hq/element-x-android/pull/4962
 - Element X iOS https://github.com/element-hq/element-x-ios/pull/4142

Conflict of Interest declaration: I am employed by Element. This MSC was written as part of my work on improving safety for Matrix clients on behalf of the org.

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4278#issuecomment-4114113387)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4278#issuecomment-4114125282)


proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,161 @@
...
+
+- Acme Corp <"on">
+  - Engineering <"private">
+    - Room A <"off">
+    - Room B <no-value>
+  - Support <no-value>
+    - Room C <"off">
+  - Room D <"off">
+  - Room F <no-value>
+- Room F
```
jimmackenzie, line 76:
```suggestion
- Room E
```

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,161 @@
...
+
+The result would be:
+ - Room A, Room C and Room D would be "off".
+ - Room B would be "private".
+ - Room E would be "on".
```
jimmackenzie, line 81:
```suggestion
 - Room F would be "on".
```

proposals/4278-media-preview-switch.md
Cadair, line 1: Would it be worth extending the global preferences to include room invite images?

Half-Shot: Oh yes, that's something I entirely forgot to mention having tunnel visioned on timeline media all day.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+Bad actors can easily send a DM to a user (which would pass the `private` check) containing unwanted
+content.
+
+### Levels
+
+The account data may exist at both the global and room level. The global setting defines the preference for
+all rooms, unless a per-room setting overrides it.
+
+It is also possible for rules to cascade via spaces. A top level space may set a specific rule, and child
+rooms may set their own rules. When this is the case, the strictest rule must always be applied.
```
clokep, line 73: The below example should probably explore rooms that are in multiple spaces.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+### Requires adjustments to well-known
+
+This proposal requires administrators to adjust their well-known, which means that bad-actor homeservers
+may potentially set the value of `content_trust_level` to `trusted` even if the content is not trusted.
+
+This is mitigated by users being able to override settings, but does however mean that unsuspecting users
+may be more exposed to content than they should be.
+
+
+## Alternatives
```
clokep, line 120: Maybe another alternative is to key on sender, not on room type? Ie sharing a private room with a user might be enough to trust their media?

I think at some point someone had explored the idea of "trusted users"...

proposals/4278-media-preview-switch.md
turt2live, line 1: Implementation requirements:
* [x] Client

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+This key contains the following content.
+
+```json
+{
+  "media_previews": "off|on|private",
+  "invite_avatars": "off|on"
+}
+```
+
+`media_previews` refers to media in the room timeline, that may be thumbnailed.
```
olivia-fl, line 31: Does this include user avatars, or media in reactions from [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027)? The text should probably spell that out explicitly. If they are not included, we should consider separate fields to control them. User avatars, in particular, are a common abuse vector in my experience. Most current clients that have a "hide media" option do not hide user avatars.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+Users may individually hide media, and this preference must be respected over the room defaults.
+
+#### `private`
+
+Previws for media MAY be shown in "private" rooms without a prompt. A private room is any room where:
+  - The `m.room.join_rules` state exists.
+  - The `join_rule` key of this state is `invite`, `knock`, `restricted`, or `knock_restricted`.
+
+If any other `join_rule` is set, or cannot be determined by the client then the assumption MUST be that the
+room is public and previews should be shown. Future join rules may be added to this list, but it's critical
```
manuroe, line 58: Are we missing a "not" here?

```suggestion
room is public and previews should not be shown. Future join rules may be added to this list, but it's critical
```

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+that clients adopt a safety first approach here.
+
+This value is the **default** setting when no account data exists on the user's account.
+
+Note that this setting has no effect for `invite_avatars`. Avatars can only be `off` or `on` for all invites.
```
manuroe, line 63: Can we state what should be the default value for `invite_avatars`?

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+Most graphical Matrix clients display media in the timeline of a room with a preview. Often these are delivered
+without any consent prompt, at least by default. This presents a risk to users who may be joining the network
+and do not have the same safety barriers that experienced users may have configured.
+
+With that in mind, this MSC provides a simple switch for users and homeservers to configure a global media preview
+preference.
+
+For the purposes of this MSC, the term "undesirable content" refers to images that the user does not want to see,
+be it illegal, abusive or otherwise wanted.
```
Xiretza, line 15:
```suggestion
be it illegal, abusive or otherwise unwanted.
```

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+`media_previews` refers to media in the room timeline, that may be thumbnailed.
+
+`invite_avatars` refers to any *room* avatar rendered in the client, where the client has a `membership` of `invite`.
+
+The fields `media_previews` and `invite_avatars` may be one of three values.
+
+#### `off`
+
+The client MUST not show any previews for any media, in the room.
```
Xiretza, line 39:
```suggestion
The client MUST NOT show any previews for any media in affected rooms.
```

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+Users may individually consent to seeing media, for example by clicking on a prompt to show a preview.
+
+If consent is given, the client SHOULD then track that consent and show the media again in the future.
+
+#### `on`
+
+Media MAY be shown in any room without a prompt. 
+
+Users may individually hide media, and this preference must be respected over the room defaults.
```
Xiretza, line 49:
```suggestion
Users may individually hide media, and this preference MUST be respected over any defaults defined in `m.media_preview_config`.
```

"the room defaults" sounds like it only includes the room-level account data.

proposals/4278-media-preview-switch.md
Xiretza: Should actually be a "MUST NOT", I think.

Xiretza, line 58: This should clarify that for (presumed-)public rooms, the opt-in behaviour is the same as for `off` ("Users may individually consent to seeing media[...]") and for private rooms, the opt-out behaviour is the same as for `on` ("Users may individually hide media[...]").

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+ - Room A, Room C and Room D would be "off".
+ - Room B would be "private".
+ - Room E would be "on".
+ - Room F is out of the space, and would default to the user's global rule.
+
+### Notes
+
+It's important here that this account data MUST be configurable by a user.
+
+Homeservers COULD specify a default value ahead of time for the user, by setting a default
```
Xiretza, line 97:
```suggestion
Homeservers MAY specify a default value ahead of time for the user, by setting a default
```

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+Not all clients will respect this configuration initially, and many clients will continue to support
+their own variant of this setting in the short term.
+
+## Potential issues
+
+### The field is extremely coarse 
+
+This provides a blanket setting for all rooms and content received, rather than this applying to rooms
+specifically (via a state key or otherwise).
```
Xiretza, line 108: "The account data may exist at both the global and room level" seems to contradict this, or am I misunderstanding something?

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,170 @@
...
+specifically (via a state key or otherwise). 
+
+
+### Requires adjustments to well-known
+
+This proposal requires administrators to adjust their well-known, which means that bad-actor homeservers
+may potentially set the value of `content_trust_level` to `trusted` even if the content is not trusted.
+
+This is mitigated by users being able to override settings, but does however mean that unsuspecting users
+may be more exposed to content than they should be.
```
Xiretza, line 117: This seems to be from a different proposal entirely

proposals/4278-media-preview-switch.md
Half-Shot: +1 to spelling it out. 

It's a good question. This originally focused on timeline media, either m.image/m.video, m.sticker, or inline media but reactions and avatars are other possible abuse vectors.

My feeling is that for reactions, it might just be suitable to show the alt text in public rooms. For user avatars, I think losing all avatars in a public room might be a bit extreme. One way we could go about this is allowing users to click-to-consent to individual avatars in public rooms like we have for Element's previews, and then that is tracked. We'd need some way to cascade rules from spaces so that people in my trusted circle are automatically permitted.

And of course the problem with all this is the client needs to clearly tell you *why* an avatar is/isn't displaying to prevent it looking like a bug.

proposals/4278-media-preview-switch.md
Half-Shot: Errmhphh, yes this was from the previous one and should have gone in the bin.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,157 @@
...
+
+Homeservers COULD specify a default value ahead of time for the user, by setting a default
+value internally for the account data. The user *must* be able to mutate this value.
+
+Not all clients will respect this configuration initially, and many clients will continue to support
+their own variant of this setting in the short term.
+
+## Potential issues
+
+TODO: Write this up.
```
Half-Shot, line 105: @Half-Shot needs to write this up.

My thoughts are primarily that this is more things the client needs to track, and the cascading model is quite heavy for clients to build. in an OOB convo we remarked that spaces are already quite heavy.

This also presents a minor UX challenge for clients, they will need to expose this setting at least 3 times:

 - In the global client settings
 - In spaces
 - In individual rooms

and then keep those settings in sync with other clients. This is a tall order, but I think we do need the granularity because if it's too coarse people will just switch it on/off and have a bad time, or at worst stop federating.

proposals/4278-media-preview-switch.md
Half-Shot: that's a bad one :sweat:

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,157 @@
...
+Homeservers MAY specify a default value ahead of time for the user, by setting a default
+value internally for the account data. The user *must* be able to mutate this value.
+
+Not all clients will respect this configuration initially, and many clients will continue to support
+their own variant of this setting in the short term.
+
+## Potential issues
+
+TODO: Write this up.
+
```
Half-Shot, line 109: Additionally "on"|"off" may be too coarse for invites.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,160 @@
...
+The client MUST NOT show any previews for any media in affected rooms.
+
+Users may individually consent to seeing media, for example by clicking on a prompt to show a preview.
+
+If consent is given, the client SHOULD then track that consent and show the media again in the future.
+
+This value is the **default** setting for `invite_avatars` when no account data exists on the user's account.
+
+#### `on`
```
BillCarsonFr, line 48: Other platforms (discord for example) offers more than just on/off. It offers on/blur/block.
I think that it is a good addition. Clients that do not support blur could just block in that case.

The blured image gives a rough idea of the image without exposing the user to the real image.

Without that how is a user expected to react to a `off` image? Never click on it or take the risk and click on it?

proposals/4278-media-preview-switch.md
olivia-fl: Yeah, that all makes sense, and I think points towards having a separate field for hiding avatars. Personally, there are some rooms that get frequent spam with abusive avatars where I would 100% turn on "hide all avatars" if it was an option in the client. The ability to selectively show avatars from trusted users (or even manually enabling avatars for everybody in the room periodically when there isn't spam happening) would be preferable. I wouldn't want to bog down this MSC with the complexity of per-user switches though, and am happy with incremental improvement.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,160 @@
...
+Bad actors can easily send a DM to a user (which would pass the `private` check) containing unwanted
+content.
+
+### Levels
+
+The account data may exist at both the global and room level. The global setting defines the preference for
+all rooms, unless a per-room setting overrides it.
+
+It is also possible for rules to cascade via spaces. A top level space may set a specific rule, and child
+rooms may set their own rules. When this is the case, the strictest rule must always be applied.
```
Johennes, line 76: Could this selective inheritance lead to confusing situations where you cannot set per-room controls anymore because a parent space has set stricter controls?

proposals/4278-media-preview-switch.md
MatMaul: The main trouble with `blur` is that the picture is fetched and ends up in both the client and the server cache. It's quite bad if it's illegal content.

We may still want the `blur` option for private federations however ? I wouldn't recommend that as a default for the public federation because of ^.

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,160 @@
...
+```json
+{
+  "media_previews": "off|on|private",
+  "invite_avatars": "off|on"
+}
+```
+
+`media_previews` refers to media in the room timeline, that may be thumbnailed.
+
+`invite_avatars` refers to any *room* avatar rendered in the client, where the client has a `membership` of `invite`.
```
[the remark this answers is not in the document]
Half-Shot: Ah yes, should be clear it's not the user avatar but *any* third party media used in the presentation of a room you are invited into. Will clarify.

proposals/4278-media-preview-switch.md
Half-Shot: I think clients would need to figure out how to convey that sanely (perhaps greying out the options that *would* be overuled).

Equally, this section might be up for the chop if we can't find a sane way to design it into clients..

proposals/4278-media-preview-switch.md
Half-Shot: blur*hashes* do not, as they're part of the event's metadata https://github.com/matrix-org/matrix-spec-proposals/pull/2448 (unless it's user avatars, which is a different kettle o fish)

But I think it's probably subjective whether there is enough information in a blurred image to tell you whether something is NSFW. Also, clients can *lie* about the hash so if anything it's likely to be abused as much as avatars.

I'd probably say blurs are really just useful as loading indicators but otherwise should be for trusted media.

Half-Shot:
https://github.com/element-hq/element-web/pull/29582 is now a complete implementation minus the spaces support, which I am tempted to reconsider in the face of client complexity. Will update the MSC tomorrow.

proposals/4278-media-preview-switch.md
t3chguy: A setting based on whether you already share a (joined) DM with the inviter could be quite powerful as a trivial form of (cheap) "trust" where checking is just cross-referencing the `m.direct` global account_data

proposals/4278-media-preview-switch.md
t3chguy: See my suggestion above https://github.com/matrix-org/matrix-spec-proposals/pull/4278/files#r2037916930 for a cheaper approach

proposals/4278-media-preview-switch.md
Half-Shot: I've chopped the cascading model.

Half-Shot:
I've updated the proposal with the latest thinkings:

 - Spaces are now out, for the moment. We did have an honest look at putting it into the client, but the complexity was too great. In the interest in not blocking what is a fairly critical safety piece, this will land with the two levels of control. Further MSCs could build upon this though.
 - There is hopefully more clarity about precisely what we consider a room avatar and a media preview.

proposals/4278-media-preview-switch.md
Half-Shot: This has been added to the alternatives section. The idea is nice albeit I think I'd like to include that in another MSC.

proposals/4278-media-preview-switch.md
Half-Shot: I think this is covered by alternatives now?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+  "invite_avatars": "off|on"
+}
+```
+
+### `media_previews`
+
+`media_previews` refers to any media that may be automatically rendered in the context of a room without user
+interaction, like a thumbnail.
+
+Clients should show show or hide media depending upon this setting, on at least the following content types:
```
turt2live:
```suggestion
Clients SHOULD show or hide media depending upon this setting, on at least the following content types:
```

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+
+ - Media message types `m.image`, `m.video`, `m.file` which all can include thumbnail information or be thumbnailed
+ by the home server.
+ - The `m.sticker` event type.
+ - Inline media via `img`.
+ - Any event which includes thumbnail information or may be thumbnailed by the home server.
```
turt2live:
```suggestion
 - Any event which includes thumbnail information or may be thumbnailed by the homeserver.
```

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+to see the media.
+
+Users may individually consent to seeing media, for example by clicking on a prompt to show a preview.
+
+If consent is given, the client SHOULD then track that consent and show the media again in the future.
+
+#### `private`
+
+Previews for media MAY be shown in "private" rooms without a prompt. A private room is any room where 
+the `m.room.join_rules` state exists AND `join_rule` key of this state is `invite`, `knock`, `restricted`, or `knock_restricted`.
```
turt2live: I don't expect the MSC to mention this, but FYI the T&S team is (slowly) working on a formal technical definition for the word "private". It's very close to what this MSC describes at the moment, but is a little different.

This MSC can do what it's currently doing in my opinion. T&S will "fix" it later, if needed.

turt2live (approved):
Generally, this looks broadly ready for FCP to me. My comments here are non-blocking for FCP purposes. Threads prior to my review should also be resolved, per the MSC checklist I'll add shortly.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge

proposals/4278-media-preview-controls.md
clokep: This is closely related to what I was going to ask -- is this defining a new term which conflicts with the existing usage of "private" in the spec?

proposals/4278-media-preview-switch.md
```diff
@@ -0,0 +1,184 @@
...
+Clients should show show or hide media depending upon this setting, on at least the following content types:
+
+ - Media message types `m.image`, `m.video`, `m.file` which all can include thumbnail information or be thumbnailed
+ by the home server.
+ - The `m.sticker` event type.
+ - Inline media via `img`.
+ - Any event which includes thumbnail information or may be thumbnailed by the home server.
+
+The above list is not exhaustive as future event types may increase the scope of media in the context of a room,
+but offers a reference point for clients.
```
Johennes: This feels potentially contradictory. A client can only filter instances of media that it knows about. If new types of renderable media are introduced in future, I think the respective MSCs should clarify whether or not they fall under the media filtering introduced by this proposal?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+room is public and previews MUST not be shown. Future join rules may be added to this list, but it's critical
+that clients adopt a safety first approach here.
+
+Note that this setting has no effect for `invite_avatars`. Avatars can only be `off` or `on` for all invites. 
+Bad actors can easily send a DM to a user (which would pass the `private` check) containing unwanted
+content.
+
+#### `on`
+
+Media MAY be shown in any room without a prompt.
```
Johennes: Should this not rather be a SHOULD? If it set the property to `on`, I would expect my client to show media unless it has a very good reason not to. MAY means the client could just blanketly ignore my choice.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+
+Note that this setting has no effect for `invite_avatars`. Avatars can only be `off` or `on` for all invites. 
+Bad actors can easily send a DM to a user (which would pass the `private` check) containing unwanted
+content.
+
+#### `on`
+
+Media MAY be shown in any room without a prompt. 
+
+Users may individually hide media, and this preference MUST be respected over any defaults defined in `m.media_preview_config`.
```
Johennes: A feature to hide individual media isn't currently defined by the spec or this proposal. It feels slightly odd to add normative wording about how such a feature is to behave when it is still hypothetical?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+
+This value is the **default** setting for both properties when no value is set, to keep with the previous defaults.
+
+### Levels
+
+The account data may exist at both the global and room level. The global setting defines the preference for
+all rooms, unless a per-room setting overrides it.
```
Johennes, line 119: How should `invite_avatars` work in room account data? You would probably only write room account once you've joined a room. But then you're already beyond the invite stage making the setting pointless?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+### Notes
+
+Homeservers MAY specify a default value ahead of time for the user, by setting a default
+value internally for the account data. The user *must* be able to mutate this value, as it's
+considered a safety feature.
```
Johennes, line 154: Maybe this should be a SHOULD rather than MUST?

If the server defaults previews to `off`, that seems like the safest possible option and it feels acceptable to deny users turning previews on again. This might, for instance, be a plausible setting for corporate environments when the room in question is federated.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+### Spaces
+
+A previous version of this MSC also explored the idea of the setting cascading through a space tree, but
+it was difficult to design for both because it required clients to hold more state about the room (e.g. traversing
+a large amount of state locally to determine the rules) as well as challenges in rendering the setting to users.
+
+A future MSC may explore the possibility of this applying to spaces, but this initial version will operate only
+globally and per-room.
+
+However, not having this feature means many room may need apply their own rules.
```
Johennes:
```suggestion
However, not having this feature means many rooms may need to apply their own rules.
```

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+### Levels
+
+The account data may exist at both the global and room level. The global setting defines the preference for
+all rooms, unless a per-room setting overrides it.
+
+
+### Notes
+
+Homeservers MAY specify a default value ahead of time for the user, by setting a default
+value internally for the account data. The user *must* be able to mutate this value, as it's
```
hughns: As the reason for allowing mutation is for safety of the user, then should it be that the server MUST allow the user to choose a more restrictive setting, but the homeserver MAY enforce a minimum.e.g. A server policy could be set as follows: allowed media_previews values: off, private (default)

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,184 @@
...
+This proposal introduces a new account data key, `m.media_preview_config`.
+
+This key contains the following content.
+
+```json
+{
+  "media_previews": "off|private|on",
+  "invite_avatars": "off|on"
+}
+```
```
Johennes, line 29: The proposal and some of the threads already touch upon possible future extensions of the config. Do we need to account for a case where more than one mode can be active at a time?

For instance, lets say a new mode `moderated` is introduced in future to allow media previews in rooms that are governed by a moderation bot. I now may want to allow previews in both `private` and `moderated` rooms but the structure of `m.media_preview_config` only allows one at a time.

proposals/4278-media-preview-controls.md
clokep: In particular `private` is a reserved join rule, which makes this kind of confusing IMO.

proposals/4278-media-preview-controls.md
uhoreg: Should images from URL previews be included in this list?  (especially [previews generated by the event sender](https://github.com/matrix-org/matrix-spec-proposals/pull/4095))

tulir: I'd say yes, since even without MSC4095, an attacker can trivially make a website that previews whatever malicious media they want

MSC2448 also defines blurhashes for URL previews, which may be useful for clients that want to hide the preview images

proposals/4278-media-preview-switch.md
Half-Shot: Cascading over spaces was dropped in this MSC, so closing this.

proposals/4278-media-preview-switch.md
Half-Shot: Tried to tidy this up in 2e2573fe519d8e9596e23d65b22384c1f57a914c

proposals/4278-media-preview-switch.md
Half-Shot: Fixed in f1283f05d3c99601b634cfa18b7242eec3e0feff

proposals/4278-media-preview-switch.md
Half-Shot: Covered in b979b87b2df20804486ff3b802b07a08caf515e6

proposals/4278-media-preview-controls.md
Half-Shot: @clokep @turt2live I'm confused by this feedback, should I be changing the word "private" so as not to clash?

proposals/4278-media-preview-switch.md
Half-Shot: I ended up saying as much:

>The above list is not exhaustive. Other MSCs SHOULD take this setting into account when rendering content that could
be considered a media preview, such as [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027). 

The burden is shifted onto other MSCs. It may be useful if the SCT would keep this spec in mind (when it lands) when reviewing other MSCs which include media.

proposals/4278-media-preview-controls.md
Half-Shot: Updated 08ca88744090b9a1185bf5e7e46582b4ab24cd39

proposals/4278-media-preview-controls.md
Half-Shot: I think it's still essential to allow clients to do this, but it's not a hard requirement.

> The mechanism for tracking opt-in/opt-out of media is left as an implementation detail.

A future MSC could de-implementation detail it, but for the moment this system is working out quite well for the Element clients.

proposals/4278-media-preview-controls.md
Half-Shot: Yeah, I kinda figured it was obvious that this setting is only really useful at the global level, but I also didn't feel like it was worth complicating the MSC by stating that the fields have different levels. 

My expectation is that implementations will just resolve the correct account for global/room each time they evaluate a piece of media and pick the first value. I hadn't included that resolving piece though. edc287ed4d5dd00b14877746241098fc0ee5e494 expands my thinking.

proposals/4278-media-preview-controls.md
Half-Shot: I'm unsure on this one. To me this is a safety setting for clients and as a user I should retain control over how my client *chooses* to render media. I would find it quite scary if my homeserver has forced any value in a safety settings control.

 Whether or not the homeserver provides the media I requested based off it's own rules is a different system to me (or rather, this feels like the wrong place to put in corporate controls).

@matrix-org/trust-safety may have opinions here

proposals/4278-media-preview-controls.md
Half-Shot: I think this is the same concern as https://github.com/matrix-org/matrix-spec-proposals/pull/4278#discussion_r2979813108?

proposals/4278-media-preview-controls.md
Half-Shot: My main concern is the power in this MSC is the simplicity, which makes it easier for clients to protect their users. We could instead have an array of "room types" and "on" / "off", but it becomes much harder to build a UI around that based on our previous attempts.

I've mentioned in the MSC that if there is an extension, then clients that do not support the new extension should treat the value as `off` for safety. 

My feeling is that we can and should consider more nuanced cased for previews but it should be pushed to the next MSC so that we can get this out and give users a coarse setting to protect themselves (as unfortunately this has already sat around far too long).

proposals/4278-media-preview-controls.md
Half-Shot: The EWeb implementation does, so we should!

proposals/4278-media-preview-controls.md
clokep: I don't know the solution besides that this is defining a new term. That is probably ok if it doesn't conflict with future T&S or other plans.

proposals/4278-media-preview-controls.md
jimmackenzie: Emitting an @matrix-org/trust-safety opinion: we love sensible defaults, we support folks having freedom to choose a value that works for them, so we'd back keeping the MUST. The argument that servers could choose whether to provide requested media feels sound.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+
+ - Media message types `m.image`, `m.video`, `m.file` which all can include thumbnail information or be thumbnailed
+ by the homeserver.
+ - The `m.sticker` event type.
+ - Inline media via `img`.
+ - Any event which includes thumbnail information or may be thumbnailed by the homeserver.
+ - URL Preview thumbnails
+
+The above list is not exhaustive. Other MSCs SHOULD take this setting into account when rendering content that could
+be considered a media preview, such as [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027).
```
uhoreg:
```suggestion
be considered a media preview, such as [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027) (Custom Images in Reactions). 
```

proposals/4278-media-preview-controls.md
clokep: This might need @jimmackenzie inout also, but "private" means many things to different people. There's been a bunch of previous conversations around what a "private room" is.

I don't really want to block this on naming stuff, but this term is also pretty loaded.

From my POV I'm looking for input from other SCT members here.

turt2live: No changes required from my perspective - just flagging that the definition might change later.

proposals/4278-media-preview-controls.md
anoadragon453: @Johennes switching `media_previews` to something like an array (e.g. `["private", "moderated", "public", ...]`) would allow for more future extensibility. But it does lead to potentially confusing combinations (a user sets the value to `["private", "moderated"]`. Should previews be shown in public, but moderated rooms?).

On the UI discussion, we currently have the following in Element:

Image

You could switch this to checkboxes:

Image

But the fact that "moderated" can overlap with "public" and "private" leads to it being difficult to predict what should happen in a given room.

Perhaps something like a slider view would be better:

Image

And to support "private and moderated", you can define a new "private_moderated" (or just "moderated") value for `media_previews`. No array needed.

Personally I like the idea of each value opening up more rooms where previews would be shown by default. The simple string type for `media_previews` works in that case.

---

@Half-Shot I imagine you've already looked at this, but Discord takes a separate approach, allowing for lots of extensibility:

Image

I would model this as:

```json
{
  "msm": {
    "direct_messages": "blur",
    "direct_messages_friends": "block",
    "server_channels": "blur"
  },
  "gm": {
    "direct_messages": "blur",
    "direct_messages_friends": "block",
    "server_channels": "blur"
  },
}
```

Though this seems overkill for our use case, and I think the slider above is much more user-friendly (though the UI that Element has today is fine).

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+## Proposal
+
+This proposal introduces a new account data key, `m.media_preview_config`.
+
+This key contains the following content.
+
+```json
+{
+  "media_previews": "off|private|on",
+  "invite_avatars": "off|on"
```
anoadragon453, line 27: What do you think about:

* `off` -> `hide`
* `on` -> `show`

"off" sounds to me like the avatar should not be shown. Whereas instead we expect clients to initially hide them until the user has consented to viewing them.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+A private room is any room where:
+
+ - The `m.room.join_rules` state exists
+ - The `join_rule` key of this state is `invite`, `knock`, `restricted`, or `knock_restricted`.
+
+If any other `join_rule` is set, or cannot be determined by the client then the assumption MUST be that the
+room is public and previews MUST not be shown. Future join rules may be added to this list, but it's critical
+that clients adopt a safety first approach here.
+
+The user MAY still opt-in to media if the room is "public", and opt-out if the room is "private".
```
anoadragon453: What do you mean by "opt-out"? Showing a "hide" button?

I now realise such an option exists in Element, but as a user of other messaging platforms where such a feature does not exist, it would be helpful to introduce that concept within the proposal (perhaps in the introduction).

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+ - URL Preview thumbnails
+
+The above list is not exhaustive. Other MSCs SHOULD take this setting into account when rendering content that could
+be considered a media preview, such as [MSC4027](https://github.com/matrix-org/matrix-spec-proposals/pull/4027) (Custom Images in Reactions). 
+
+The one exception to the rule for now is user avatars in rooms the user has joined. For the purposes of this MSC,
+by joining a room the user *has* consented to seeing the user avatars contained within. See the potential issues
+section for a deeper explanation.
+
+### `invite_avatars`
```
anoadragon453, line 52: I assume we're using "off" and "on" strings for extensibility in the future, otherwise I would suggest switching this to a boolean.

If that is the case, could you say so explicitly in the document?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+
+Media SHOULD be shown in any room without a prompt.
+
+Users SHOULD be able to individually hide media, and this preference MUST be respected over any defaults defined in `m.media_preview_config`.
+
+This value is the **default** setting for both properties when no value is set, to keep with the previous defaults.
+
+### Other values
+
+Clients MUST treat any unknown property value in either defined property as `off`, for safety reasons.
```
anoadragon453: Could you expand on those reasons?


```suggestion
Clients MUST treat any unknown property value in either defined property as `off`. This would
prevent, for example, a user firing up an old client and suddenly seeing media after using a
newer client that has set a newer restrictive property type.
```

proposals/4278-media-preview-controls.md
anoadragon453: It'd be worth adding a sentence or two at the end of this section explicitly noting that "clients may not always have access to room-level account data when making a decision (e.g. upon being invited to a new room). In this case, they should only use the global account data when deciding how to present media".

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+  "media_previews": "on"
+  "invite_avatars": "off"
+}
+```
+
+
+### Notes
+
+Homeservers MAY specify a default value ahead of time for the user, by setting a default
+value internally for the account data. The user *must* be able to mutate this value, as it's
```
anoadragon453:
```suggestion
value internally for the account data. The user MUST be able to mutate this value, as it's
```

proposals/4278-media-preview-controls.md
anoadragon453: @Johennes in a corporate environment, I think it'd be enough that a custom client would prevent users from changing this option, despite the users technically being able to do so via an API call.

proposals/4278-media-preview-controls.md
clokep: Another option could be to have an array of join rules to allow, but that trades flexibility for probably a worse UI of some sort. 🤔 I'm still a bit concerned about adding a new term to the spec that's overloaded.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+
+The client MUST NOT show any previews for any media in affected rooms. Clients will hide the media entirely,
+behind a click-to-view prompt, or some other mechanism where a user is either prevented entirely or must consent
+to see the media.
+
+Users MAY individually consent to seeing media, for example by clicking on a prompt to show a preview.
+If consent is given, the client SHOULD then track that consent and show the media again in the future. The
+mechanism for tracking opt-in/opt-out of media is left as an implementation detail.
+
+#### `private`
```
richvdh, line 80: Non-blocking comment: I'm unenthusiastic about the use of the keyword `private` here, due to the general confusion about what exactly is a "private" room. That said, I don't have a better suggestion, and I'm happy to see you've spelt out what it means in this instance.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+
+Previews for media MAY be shown in "private" rooms without a prompt.
+A private room is any room where:
+
+ - The `m.room.join_rules` state exists
+ - The `join_rule` key of this state is `invite`, `knock`, `restricted`, or `knock_restricted`.
+
+If any other `join_rule` is set, or cannot be determined by the client then the assumption MUST be that the
+room is public and previews MUST not be shown. Future join rules may be added to this list, but it's critical
```
richvdh:
```suggestion
room is public and previews MUST NOT be shown. Future join rules may be added to this list, but it's critical
```

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,240 @@
...
+Users MAY individually consent to seeing media, for example by clicking on a prompt to show a preview.
+If consent is given, the client SHOULD then track that consent and show the media again in the future. The
+mechanism for tracking opt-in/opt-out of media is left as an implementation detail.
+
+#### `private`
+
+Previews for media MAY be shown in "private" rooms without a prompt.
+A private room is any room where:
+
+ - The `m.room.join_rules` state exists
```
richvdh, line 85: It has to be said that a room with no `join_rules` event is, in fact, invite-only (cf https://github.com/matrix-org/matrix-spec/issues/657#issuecomment-2514253692), so perhaps we want to treat it accordingly.

That said, it feels like the chance of a client encountering a room where there is genuinely no `join_rules` is lower than that of a client encountering some bug where it can't see the `join_rules` for some reason. Seems like erring on the side of safety is the right call.

Might be worth making that line of reasoning explicit, though?

richvdh (approved):

turt2live:
@mscbot concern Lots of unresolved threads - author needs to address commentary, including bringing details into the proposal itself where applicable.

proposals/4278-media-preview-controls.md
Half-Shot: Resolving as a future addition for a future MSC.

proposals/4278-media-preview-controls.md
Half-Shot: Hm, I think the symmetry with `media_previews` is more useful than defining more enums. I agree insofar that the language isn't perfect, but I think `off`/`on` is plenty clear enough.

proposals/4278-media-preview-controls.md
Half-Shot: We talked about extensibility in a sync today.

For what it's worth my feeling is that for minor changes such as new enums, the current behaviour in the spec is fine insofar that unknown enums are treated as `off` which means it fails-safe. As clients update, they can enjoy the complexity.

I feel that for larger extensions, we don't need to account for this in the current MSC. There are options such as creating a more featureful event type and using this as the fallback for older clients, but that's out of scope here.

proposals/4278-media-preview-controls.md
Half-Shot: Covered in 9dad69de

proposals/4278-media-preview-controls.md
Half-Shot: Agreed. Maybe one day when our T&S approach is more spelled out we might have some common definitions to lean on.

proposals/4278-media-preview-controls.md
Half-Shot: Done in 06883081

proposals/4278-media-preview-controls.md
Half-Shot: Clarified in bf6e1d664fd70a8e35fa4cc2bfeada62f8836e7e

proposals/4278-media-preview-controls.md
Half-Shot: Done in d3a05d3b

proposals/4278-media-preview-controls.md
Half-Shot: Noted in 3b00107d

proposals/4278-media-preview-controls.md
Half-Shot: Resolving as I don't think I can action this.

turt2live:
@mscbot resolve Lots of unresolved threads - author needs to address commentary, including bringing details into the proposal itself where applicable.

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,252 @@
...
+### Levels
+
+The account data may exist at both the global and room level. The global setting defines the preference for
+all rooms, unless a per-room setting overrides it.
+
+If the client doesn't have access to room level account data, such as for invites then it should just read
+from the global account data.
+
+If `m.media_preview_config` is set at both global and room levels, the client MUST prefer using a defined value
+over a default value. E.g. account data of:
```
dbkr: Can you clarify how you know what value is a default value here?

proposals/4278-media-preview-controls.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Trust users you share a DM with
+
+We could seek to trust media if we share a room with a user (e.g. the `m.direct`), which might be a cheap way to
+establish a basis of trust for users.
+
+### Blurring previews
+
+One suggestion would be to blur untrusted previews rather than keeping them as off. There are a few ways that this
+could be achieved, but each have downsides.
```
tulir, line 229: Technically this seems less like an alternative and more like an implementation detail of how to hide media? I'd assume any client that supports MSC2448 will simply use the sender-provided blurhashes when available instead of an ugly blank box

dbkr:
@mscbot concern default value

proposals/4278-media-preview-controls.md
Half-Shot: Have attempted to clarify. I think the "default" was a total misnomer

proposals/4278-media-preview-controls.md
Half-Shot: I think there was a concern that even blurhashes are not *great* if you are being sent harmful material and an umambigious blank box was preferrable.

Aside from the fact that we have now spent the last few years training users that blurhashes are a *loading* state, so using blurhashes is likely to confuse users into thinking the application is still loading the image.

dbkr:
@mscbot resolve default value

clokep:
I still don't love the `private` term used in here without a broader definition, but not enough to block this.

tulir:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2440

PR #4279, opened 2025-03-27
MSC: Server notice rooms

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/reporting-v2/00-user-server/proposals/4279-service-notice-rooms.md)

### DIFF
--- a/proposals/4279-service-notice-rooms.md
+++ b/proposals/4279-service-notice-rooms.md
@@ -1,0 +1,268 @@
+# MSC4279: Server notice rooms
+
+[Server Notices](https://spec.matrix.org/v1.13/client-server-api/#server-notices) are already a concept
+within Matrix, though they are often perceived as [phishing attempts](https://github.com/element-hq/element-meta/issues/1759),
+limiting their usefulness. At the time of their introduction, event-based server notices were primarily
+intended to make users aware of Terms of Service changes on homeservers, especially as GDPR legislation
+was starting to take effect. Over the years since, a few sharp edges have been found:
+
+1. As already mentioned, the server sends a regular room invite to the user, which looks a lot like
+   a phishing attempt. Users have no indication that they can trust that the invite is valid.
+2. Despite being an invite, Synapse's implementation doesn't allow the user to reject it. Users get
+   confusing/annoying error messages which build some trust in the invite, but erode trust in the
+   server/admin/protocol's ability to function.
+3. The rooms are easily lost, and can't be "closed". This leads to data being abandoned forever on
+   the server.
+4. The events use a custom msgtype for enhanced UI, though clients rarely implement such behaviour
+   and instead rely on the fallback.
+
+This proposal addresses these core concerns by replacing the event-based system with a room-based
+mechanism. Rooms are created using presets (though could be created manually still, if desired) and
+have a specific room type that clients can use to render them properly. They are still technically
+sent as invites to the user, however this detail is expected to be hidden from the user in clients.
+There are also some events and behaviours specified to allow the room to be cleaned up, and signal
+to the client/user what actions they can take.
+
+By improving the mechanism, the T&S team at the Matrix.org Foundation hopes to be able to use the
+functionality to communicate with users who report content about the status of their reports, and
+with users who have moderation action taken against them. These rooms can also support two-way
+communication, allowing users to appeal the team's decision with useful context.
+
+
+## Future MSC work
+
+As a feature, server notice rooms can expand in scope until they consume all available matter in the
+universe. To help limit this proposal's scope, allowing it to more easily move through the process,
+the following non-exhaustive list of things are deliberately left for future proposals to tackle:
+
+* Integrating server notice rooms into report/ticket flows. One possible option here is to extend the
+  appservice room registration to include specific [room types](https://spec.matrix.org/v1.13/client-server-api/#types),
+  which would allow an appservice to bridge between ticketing and Matrix. A related MSC may be to
+  extend the various reporting APIs to (optionally) return a room ID to ease the user experience of
+  sending a report and immediately getting a feedback "we got your report!" message.
+* Interactions with sending notices to users over federation. This proposal recommends that servers
+  either take measures to prevent clients from showing such rooms over federation as official, or
+  outright refuse to participate. A future MSC could introduce a way to add trust decorations to the
+  room, somehow.
+
+
+## Proposal
+
+The existing [`m.server_notice message type`](https://spec.matrix.org/v1.13/client-server-api/#server-notices)
+is deprecated for removal by a future MSC. It is replaced by this MSC's mechanism.
+
+A new `m.server_notice` [room type](https://spec.matrix.org/v1.13/client-server-api/#types) is
+established, denoting a room as "Notice Room". Servers SHOULD limit creation of server notice rooms
+to an implementation-specific definition of "administrators". When the server encounters the room
+type over federation, the server SHOULD either strip the room type before serving the room to the
+client, or refuse to participate. Stripping the room type prevents clients from getting confused
+about how 'official' the notice is, though leads to subpar user experience.
+
+There are two major kinds of Notice Rooms that a server may wish to use:
+
+1. Notices the invited user(s) can reply to.
+2. Readonly notices.
+
+Both of these kinds are determined by the power levels in the room. If the user has permission to
+send events of an appropriate type, the user may reply.
+
+To make creation easier, two new [room creation presets](https://spec.matrix.org/v1.13/client-server-api/#creation)
+are introduced to match the two notice kinds: `notice_readonly` and `notice`. They have the following
+events set by default:
+
+For `notice_readonly`:
+
+* [`m.room.create`](https://spec.matrix.org/v1.13/client-server-api/#mroomcreate) has `content`:
+  ```jsonc
+  {
+    "m.federate": false,
+    "type": "m.server_notice"
+  }
+  ```
+
+* [`m.room.history_visibility`](https://spec.matrix.org/v1.13/client-server-api/#mroomhistory_visibility)
+  has `content`:
+  ```jsonc
+  {
+    "history_visibility": "shared"
+  }
+  ```
+
+* [`m.room.join_rules`](https://spec.matrix.org/v1.13/client-server-api/#mroomjoin_rules) has `content`:
+  ```jsonc
+  {
+    "join_rule": "invite"
+  }
+  ```
+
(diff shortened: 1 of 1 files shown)


proposals/4279-service-notice-rooms.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+
+The existing [`m.server_notice message type`](https://spec.matrix.org/v1.13/client-server-api/#server-notices)
+is deprecated for removal by a future MSC. It is replaced by this MSC's mechanism.
+
+A new `m.server_notice` [room type](https://spec.matrix.org/v1.13/client-server-api/#types) is
+established, denoting a room as "Notice Room". Servers SHOULD limit creation of server notice rooms
+to an implementation-specific definition of "administrators". When the server encounters the room
+type over federation, the server SHOULD either strip the room type before serving the room to the
+client, or refuse to participate. Stripping the room type prevents clients from getting confused
+about how 'official' the notice is, though leads to subpar user experience.
```
Johennes, line 59: What would be a use case for allowing server notice rooms to federate? So far, I had thought these notices to be local by nature. If that is the case we could just disallow them to federate entirely via `m.federate`.

If we do have a use case for federating these rooms, why would a remote client not want to know that this is a server notice room? I think they could still differentiate the room from local notice rooms via the sender on `m.room.create`.

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+
+1. All server notices MUST be sent as *invites*, though clients SHOULD intercept the room invite to
+   render it with native UI.
+2. Users MUST NOT reject the invite. This is enforced by the server.
+3. Users MUST NOT [leave](https://spec.matrix.org/v1.13/client-server-api/#leaving-rooms) or forget
+   the room until the creator allows it (how this works is described later in this section). This is
+   enforced by the server.
+4. Clients SHOULD render the user's current Notice Rooms in a dedicated section of their UI rather
+   than part of a room list.
+5. Notice rooms SHOULD be created with suitable encryption enabled.
```
Johennes, line 154: What does "suitable" mean here?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+6. Servers SHOULD NOT allow notice rooms to be created without the use of the presets defined by this
+   proposal. This is to force developers to *really *think about whether they need to override the
```
Johennes, line 156:
```suggestion
   proposal. This is to force developers to *really* think about whether they need to override the
```

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+   default properties and change the behaviour of the room. They can still set things like `name`,
+   `power_level_content_override`, etc.
```
Johennes: For clarity, this means servers SHOULD reject `/createRoom` requests with `type = m.server_notice` in `creation_content` unless `preset` is set to either `notice` or `notice_readonly`? But if one of the presets is set, clients can override all aspects of the preset via the parameters of `/createRoom`?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+7. When a user leaves a notice room, the notice is considered "archived". Archived notices may be
+   [forgotten](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3roomsroomidforget)
+   to "delete" them.
```
Johennes, line 161: I think this works the same way for any room type already?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+8. If the client the user is leaving the notice room through doesn't support archived notices, the
+   client SHOULD automatically [forget](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3roomsroomidforget)
+   the room too.
```
Johennes, line 164: Possibly outside the scope of this proposal: Would it be worthwhile to apply this to all room types?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+9. Clients SHOULD NOT offer the ability to report users, rooms, or events in notice rooms.
+10. Servers SHOULD use standard event types when communicating with users. For example, `m.room.message`
+    with `msgtype: m.text` rather than custom or obscure types.
+11. Clients SHOULD NOT show a member list, presence, typing notifications in notice rooms. Clients
+    SHOULD NOT send typing notifications, either.
+12. With respect to read receipts, clients SHOULD NOT show read receipts for any other user than the
+    one that is logged in. Clients SHOULD send read receipts, if the user's settings permit it.
+13. Clients SHOULD disable slash commands and sending threaded messages in notice rooms.
```
Johennes, line 172: The spec doesn't currently define slash commands which makes it a bid odd to prohibit them. I'm also not sure how these relate to threaded messages?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Security considerations
+
+This proposal can theoretically allow users to create rooms which appear trusted to other rooms, if
+the server does not properly filter out rooms it didn't create before sending them to users. This is
+most possible while an older server and newer client are in use: the server doesn't know it's supposed
+to be filtering the room, and the client is simply responding to input it was given.
+
+An approach for fixing this would be for the client to check the supported specification versions by
+the server, determining whether the filtering code *should *be present. If the server doesn't support
```
Johennes, line 219:
```suggestion
the server, determining whether the filtering code *should* be present. If the server doesn't support
```

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+anyone with suitable power level) allows it. To do this, the creator sends an `m.room.leave_rules`
+state event with an empty string for a state key. This is a new event type. The `content` would be
+very similar to the join rules event:
+
+The `leave_rule` may be either `allow` or `deny`, and has a different default depending on the room
+type. For `m.server_notice` rooms, the default is `deny`. For all other rooms, the default is `allow`.
+The leave rule is *not *enforced by auth rules and therefore does not require a new room version. A
+future MSC may change this.
+
+Once the leave rule is set to `allow`, the user may be permitted to leave the room.
```
Johennes, line 187: Since the alternatives section is blank: How does this compare to adding `leave` in `m.room.power_levels`?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+  }
+  ```
+
+For `notice`, the same events are sent with the `m.room.power_levels` having `events_default: 0`
+instead of `events_default: 100`.
+
+For both presets, the following other things *should* happen too:
+
+* The `is_direct` flag on [`/createRoom`](https://spec.matrix.org/v1.13/client-server-api/#post_matrixclientv3createroom)
+  defaults to `true`.
```
Johennes, line 132: Is this actually needed given that clients are also encouraged to render these rooms outside the room list?

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+An approach for fixing this would be for the client to check the supported specification versions by
+the server, determining whether the filtering code *should *be present. If the server doesn't support
+the relevant specification version (or newer), the client should disable server notice room support.
+
+A future MSC may also create ways to verify the creator/sender, allowing the server to drop its filter.
+[MSC4145](https://github.com/matrix-org/matrix-spec-proposals/pull/4145) is a step towards this,
+though without federated trust.
+
+
+## Potential issues
```
Johennes, line 227: This proposal is adding more complexity to `/createRoom`. I personally find this to be one of the most complicated endpoints in the spec already today due to the multitude of parameters and the different ways in which they interact or override each other.

proposals/4279-service-notice-rooms.md
```diff
@@ -0,0 +1,268 @@
...
+within Matrix, though they are often perceived as [phishing attempts](https://github.com/element-hq/element-meta/issues/1759),
+limiting their usefulness. At the time of their introduction, event-based server notices were primarily
+intended to make users aware of Terms of Service changes on homeservers, especially as GDPR legislation
+was starting to take effect. Over the years since, a few sharp edges have been found:
+
+1. As already mentioned, the server sends a regular room invite to the user, which looks a lot like
+   a phishing attempt. Users have no indication that they can trust that the invite is valid.
+2. Despite being an invite, Synapse's implementation doesn't allow the user to reject it. Users get
+   confusing/annoying error messages which build some trust in the invite, but erode trust in the
+   server/admin/protocol's ability to function.
```
deepbluev7, line 13: Wouldn't it be better to just spec, that the server should autojoin the users? Why do we need invites for server notice rooms?

proposals/4279-service-notice-rooms.md
deepbluev7, line 59: Wouldn't it be better to keep the current tag mechanism? Tags can't be set by anyone but the server admin or the user. As such it is harder to abuse. Currently I can create a room with a server notice room type and invite users into it. As such there is no way for clients to trust the server, that the room is actually a server notice room, if we rely on the room type.

The tag otoh could be manually set by an admin or the user itself, but that would only lead to the user or admin phishing the user, which would be possible anyway. What is the reason behind using a room type? It seems a lot less reliable at first glance.

proposals/4279-service-notice-rooms.md
deepbluev7: Yeah, this seems like an odd requirement. Why would you not use threads when reporting abuse? And slash commands can do a lot of things like joining or switching rooms. That sounds rather odd as a requirement.

TheArcaneBrony: This also sounds like a dangerous thing for accessibility. I can see 2 cases where this may be a problem:
- Clients where slash commands are the primary way to manage rooms (ie. gomuks-cli)
- Users who use slash commands as a replacement for UI elements (ie. vision-impaired users who may wish to use /part etc)

proposals/4279-service-notice-rooms.md
TheArcaneBrony: Seems to me like it's provisionary for making rooms forgotten on leave by default, ie. as per MSC #4267

PR #4282, opened 2025-04-02
MSC4280: Interactive /rooms/ROOM_ID/messages (Client-Server API)

bnjbvr:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/bnjbvr/interactive-roommessag/proposals/4280-interactive-room-message.md).

TL;DR: in the client-server API, `/rooms/ROOM_ID/messages` can be slow, and clients need a way to express that they want a quick response for an interactive user session.

---

Disclosure: I am a Matrix community member, and also employed by Element to work on the [Matrix Rust SDK](https://github.com/matrix-org/matrix-rust-sdk). This proposal is written and published as an Element employee.


proposals/4282-interactive-room-message.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4282-interactive-room-message.md
```diff
@@ -0,0 +1,91 @@
...
+
+However, some server implementations, including
+[Synapse](https://github.com/element-hq/synapse/blob/5c84f258095535aaa2a4a04c850f439fd00735cc/synapse/handlers/pagination.py#L575-L584),
+[Conduit](https://gitlab.com/famedly/conduit/-/blob/a7e6f60b41122761422df2b7bcc0c192416f9a28/src/api/client_server/message.rs#L201)
+and
+[Conduwuit](https://github.com/girlbossceo/conduwuit/blob/0f81c1e1ccdcb0c5c6d5a27e82f16eb37b1e61c8/src/api/client/message.rs#L94-L101),
+may generate, under some implementation-specific conditions, federation requests to
+[backfill](https://spec.matrix.org/v1.14/server-server-api/#backfilling-and-retrieving-missing-events)
+the room timeline, and fetch more events from other servers. This slows down reception of the
+response in the client, since it now be blocking on the server waiting for the federation responses
```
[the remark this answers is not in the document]
bnjbvr: Thanks!

bnjbvr:
Added a note that the hint may be ignored in a specific case to avoid wasting CPU and bandwidth resources, after a discussion with @deepbluev7 (thanks!).

proposals/4282-interactive-room-message.md
```diff
@@ -0,0 +1,104 @@
...
+to come. Moreover, the time spent retrieving those responses is theoretically unbounded, so the
+homeserver and the clients may have to wait forever for such requests to complete.
+
+We need a more responsive way to fetch older events from the server, without having to wait for
+federation responses to come back. This is the point of this MSC.
+
+## Proposal
+
+It is proposed that the `/rooms/{room_id}/messages` endpoint be modified to allow clients to
+specify a new boolean query parameter `interactive`, which indicates that the client is interested
```
Johennes, line 44: I think you also need to define the unstable prefix to use by implementations while this MSC has not been accepted. It's typically done in a section at the very end of the proposal.

proposals/4282-interactive-room-message.md
```diff
@@ -0,0 +1,104 @@
...
+
+This MSC doesn't resolve this problem, and it may make it more apparent on the contrary, if *all*
+`/messages` requests end up *not* causing any federation backfill. The most likely consequence of
+this is that events might be more frequently misordered across clients.
+
+## Alternatives
+
+Instead of an additional query parameter, this MSC could mandate that this becomes the expected
+behavior of all the implementations. This would be an implicit breaking change, and it may inhibit
+use cases where clients might prefer a perfectly backfilled room over a quick response time.
```
MadLittleMods, line 92: Related MSC for indicating gaps in the timeline: [MSC3871](https://github.com/matrix-org/matrix-spec-proposals/pull/3871)

If we indicated gaps in the `/messages`, we could respond quickly always as a default and clients can handle displaying and filling in the gaps as necessary.

proposals/4282-interactive-room-message.md
```diff
@@ -0,0 +1,104 @@
...
+- if the homeserver has reached its known end of the room, and must backfill older events from
+  federation, then it shall:
+    - either block on the backfill request to complete, before returning the response to the
+      client,
+    - or race the completion of the backfill request with a timeout, and return an empty response
+      if the backfill request didn't complete in time. In that case, clients are expected to retry
+      the back-pagination request later.
+- otherwise, the server shall immediately return events it had in its local state, and if needed it
+  shall start a backfill request in the background (so the next request has chances to complete
+  quickly, be it interactive or not).
```
erikjohnston, line 63: The problem is that there might be gaps, and we currently have no way of telling clients that there might be a gap (nor then allowing the client to request the server fetch messages in the gap).

We could have a flag where we immediately stop if there is a gap, which could speed up the `/messages` done after a SSS `initial` room, but I don't think would help in the general case of backpagination.

ara4n: msc3871 is the msc for relaying gaps to clients

proposals/4282-interactive-room-message.md
MadLittleMods: This was explored further during the recent Element Hackathon ("How Hard Could It Be?") with an initial implementation in Synapse -> https://github.com/element-hq/synapse/pull/18873


> For the hack day, I'll be tackling the backfill problem with this strategy:
> 
>  - **Default to fast responses with gaps**: As a default, we can always respond quickly and indicate gaps ([MSC3871](https://github.com/matrix-org/matrix-spec-proposals/pull/3871)) for clients to paginate at their leisure.
>  - **Fast back-pagination**: Clients back-paginate with `/messages?dir=b&backfill=false`, and Synapse skips backfilling entirely, returning only local history with gaps as necessary.
>  - **Explicit gap filling**: To fill in gaps, clients use `/messages?dir=b&backfill=true` which works just like today to do a best effort backfill.
> 
> This allows the client to back-paginate the history we already have without delay. And can fill in the gaps as they see fit.
> 
> Gaps can be represented in the UI with a message like "We failed to get some messages in this gap, try again 🗘.", giving users clear feedback. Regardless of clients trying to fill in gaps automatically, I would still suggest to display gaps so people can tell what's happening.
> 
> This is basically a simplified version of [MSC4282](https://github.com/matrix-org/matrix-spec-proposals/pull/4282) leveraging [MSC3871: Gappy timelines](https://github.com/matrix-org/matrix-spec-proposals/pull/3871) to get proper client feedback to indicate where the gaps are so we can skip backfill without worrying. For reference, skipping backfill without letting clients know where the gaps are just means they won't ever know that they are missing messages.
>
> *-- @MadLittleMods, https://github.com/element-hq/how-hard-can-it-be-2025/issues/47#issuecomment-3234339497*

proposals/4282-interactive-room-message.md
```diff
@@ -0,0 +1,104 @@
+# MSC4282: Hint that a /rooms/{room_id}/messages request is interactive
+
+The endpoint [/rooms/{room_id}/messages](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3roomsroomidmessages)
```
turt2live, line 3: "latest" changes and may become inaccurate/unusable as a link after some changes. It's always preferable to link to a specific spec version instead.

```suggestion
The endpoint [/rooms/{room_id}/messages](https://spec.matrix.org/v1.16/client-server-api/#get_matrixclientv3roomsroomidmessages)
```

turt2live (changes requested):
This MSC is on the stack for FCP, but appears to be missing a few details. The important missing bits, I think, are:

1. Lack of an unstable prefix section
2. Implementations aren't linked

There also appears to be debate on a few pieces of the MSC.

I'll post the MSC checklist shortly after this review - once the author/authoring team feels the MSC is ready for FCP again, please bring it up in the [SCT Office](https://matrix.to/#sct-office:matrix.org) for re-review. We may not see github pings for review without a mention in the office.

turt2live:

- [ ] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal) specified in the MSC’s PR description?
- [ ] For each new endpoint that is introduced:
  - [ ] Have authentication requirements been specified?
  - [ ] Have rate-limiting requirements been specified?
  - [ ] Have guest access requirements been specified?
  - [ ] Are error responses specified?
    - [ ] Does each error case have a specified `errcode` (e.g. `M_FORBIDDEN`) and HTTP status code?
      - [ ] If a new `errcode` is introduced, is it clear that it is new?
- [ ] Will the MSC require a new room version, and if so, has that been made clear?
  - [ ] Is the reason for a new room version clearly stated? For example, modifying the set of redacted fields changes how event IDs are calculated, thus requiring a new room version.
  - [ ] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/latest/appendices/#pagination)?
- [ ] An introduction exists and clearly outlines the problem being solved. Ideally, the first paragraph should be understandable by a non-technical audience.
- [ ] While the exact sections do not need to be present, the details implied by the proposal template are covered. Namely:
  - [ ] Introduction
  - [ ] Proposal text
  - [ ] Potential issues
  - [ ] Alternatives
  - [ ] Dependencies
- [ ] Stable identifiers are used throughout the proposal, except for the unstable prefix section
  - [ ] Unstable prefixes [consider](https://github.com/matrix-org/matrix-spec-proposals/blob/main/README.md#unstable-prefixes) the awkward accepted-but-not-merged state
  - [ ] Chosen unstable prefixes do not pollute any global namespace (use “org.matrix.mscXXXX”, not “org.matrix”).
- [ ] Changes have applicable [Sign Off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) from all authors/editors/contributors
- [ ] There is a dedicated "Security Considerations" section which detail any possible attacks/vulnerabilities this proposal may introduce, even if this is "None.". See [RFC3552](https://datatracker.ietf.org/doc/html/rfc3552) for things to think about, but in particular pay attention to the [OWASP Top Ten](https://owasp.org/www-project-top-ten/).

PR #4283, opened 2025-04-02
MSC: Distinction between Ignore and Block

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/safety-glossay/ignore-block/proposals/4283-ignore-block-distinction.md)

### DIFF
--- a/proposals/4283-ignore-block-distinction.md
+++ b/proposals/4283-ignore-block-distinction.md
@@ -1,0 +1,91 @@
+# MSC4283: Distinction between Ignore and Block
+
+When discussing features around user safety, terminology like *ignore*, *block*, *filter*, and *mute*
+is typically used. Though the terminology is consistent, the definitions are not.
+
+In a similar effort to [MSC4161](https://github.com/matrix-org/matrix-spec-proposals/pull/4161), this
+proposal aims to clarify the definitions used from a technical perspective. Clients can, but SHOULD
+NOT, use different terminology for their users.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


proposals/4283-ignore-block-distinction.md
turt2live, line 1: Implementation requirements:
* [ ] Some sort of demonstration that these terms make sense (ie: evidence that "block" is defined correctly, given that "ignore" is already used)

clokep (approved):

proposals/4283-ignore-block-distinction.md
```diff
@@ -0,0 +1,91 @@
...
+
+Other platforms and protocols may call *ignores* "mutes" or "filters". Avoid these terms as they
+conflict with other parts of the Matrix Protocol (namely, with notifications and `/sync`).
+
+**Block** means to prevent the sender from completing an action. This is typically done server-side,
+but with some work, can be done client-side. Reusing the invite example, the sender would get explicitly
+told "you cannot send this invite" instead of it appearing successful with an *ignore*.
+
+Other platforms may call *blocks* "ignores", which is mildly confusing, but swapping the terminology
+would involve significant edits to the specification and its implementations.
```
turt2live, line 28: Out of band (summarized) thoughts: Blocks are meant to be bidrectional too. For example, if Alice blocks Bob, Bob can't see what Alice posts (and Alice can't see Bob's posts). Bob knows this is happening.

Supporting this in Matrix is difficult, but not impossible. One option may be to further distinguish what subtype of block we support (ie: public or private blocks).

PR #4286, opened 2025-05-06
MSCxxxx: App store compliant handling of payment links within events

hughns:


manuroe:
Note that massige changes are happening on the App Store topic. Since last week for USA, and only for the USA at this moment:

*As part of the ruling, the judge says that Apple cannot:
    - Impose “any commission or any fee on purchases that consumers make outside an app”
    - Restrict developers’ style, formatting, or placement of links for purchases outside of an app
    - Block or limit the “use of buttons or other calls to action”
    - Interfere with consumers’ choice to leave an app with anything beyond “a neutral message apprising users that they are going to a third-party site”*
 https://www.theverge.com/news/659246/apple-epic-app-store-judge-ruling-control

Hopefully it will extend worldwide soon and we will not need this MSC.

proposals/4286-external-payment-details.md
```diff
@@ -0,0 +1,149 @@
+# MSC4286: App store compliant handling of payment links within events
```
MTRNord, line 1: I am unsure if we should “approve” questionable legality practices of apple in Matrix Spec. Especially considering the recently lost court cases, I believe this is the wrong move and instead, at most should be handled on an implementation basis rather than spec.

See for example, https://techcrunch.com/2025/04/30/epic-games-just-scored-a-major-win-against-apple/ as a randomly chosen source for this.

proposals/4286-external-payment-details.md
```diff
@@ -0,0 +1,149 @@
...
+    "msgtype": "m.text",
+    "format": "org.matrix.custom.html",
+    "body": "Please use a web browser to manage your account",
+    "formatted_body": "Please use a web browser to manage your account<span data-mx-external-payment-details> or click <a href=\"https://account.example.com/plan\">here</a></span>"
+}
+```
+
+On a client that is aware of the new attribute and chooses to not render, it would appear as:
+
+> Please use a web browser to manage your account
```
MTRNord, line 54: I feel like this is a major downgrade in UX and arguably makes the app unusable. Imho I rather would see the APIs standardized to be native to the app rather than this.

ara4n: i suspect you are missing the point of the MSC, which is that some appstores don't let you manage purchases with native code in the app *or* with links to external web services.

ara4n:
> Hopefully it will extend worldwide soon and we will not need this MSC.

unfortunately we can't count on this, plus it might revert in future. so far Apple have only changed things in the US, and based on other regulatory behaviour they may apply regulation strictly to the geographies where they've been regulated.

proposals/4286-external-payment-details.md
ara4n: Matrix is built for the real world, and the sad reality is that today in most jurisdictions Apple still has this rule in place and has had it for many years - even if there's an injunction against it at the time of writing in the US.  The fact that it's legally questionable is orthogonal to the fact that exists and impacts every Matrix client on the iOS App Store, and therefore needs to be solved in a consistent cross-client manner - hence being in the spec.

proposals/4286-external-payment-details.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Background
+
+Developers of Matrix clients that are distributed via the Apple [App Store](https://www.apple.com/app-store/)
+must comply with the [App Review Guidelines](https://developer.apple.com/app-store/review/guidelines/).
+
+These guidelines include restrictions on which payment methods are offered and how they are accessed.
+
+If a homeserver operator requires a paid plan for usage of some capabilities then a conflict of
+interests can arise between the homserver operator and the application developer:
```
jaller94:
```suggestion
interests can arise between the homeserver operator and the application developer:
```

proposals/4286-external-payment-details.md
KitsuneRal: I don't see it as any sort of "approval" but I feel like we are speccing a loophole. If a given homeserver doesn't comply with this MSC, clients will have a hard time complying with Apple's practices anyway. And on the other hand, since the content is still making its way to the client even with the `span` markup, will it be sufficient for Apple? The last thing I would want is Apple coming back to the Foundation for endorsing a practice they see as violating their T&Cs. The link is not even related to the client, it's made by the server which, usually in Matrix, is formally independent from the client; therefore I'm not sure the quoted part from the review guidelines actually applies. Can/Do we have a legal review of this?

PR #4287, opened 2025-05-08
MSC4287: Sharing key backup preference between clients

andybalaam:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/andybalaam/shared-key-backup-preference/proposals/4287-shared-key-backup-preference.md)

Implementations:
 * Element Web: https://github.com/element-hq/element-web/pull/29290
 * matrix-rust-sdk (and hence EX): https://github.com/matrix-org/matrix-rust-sdk/pull/2842

I am employed by Element and a Matrix community member. This proposal was written and published with my Element client developer hat on.

---

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4287#issuecomment-3785858098)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4287#issuecomment-3785860383)


proposals/4287-shared-key-backup-preference.md
turt2live, line 1: Implementation requirements:
* [x] Client setting value
* [x] Client using value

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+### A proper unstable prefix
+
+Because an existing implementation exists, the proposed unstable prefix has
+different semantics from the final proposal. If this is deemed unacceptable, a
+more normal unstable prefix could be used.
+
+## Security considerations
+
+Unencrypted account data is under the control of the server, so a malicious
+server could:
```
Half-Shot, line 100: Something that could offset this a bit is strong requirements for clients to announce the setting to users. Either as a toast or dialog or something attention-grabby.

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+  are enabled for a user.
+
+* If this event type does not exist in account data, or if it does not contain
+  the `enabled` property, or if the value of `enabled` is not a boolean value,
+  clients MUST ignore the existing value and:
+
+    * EITHER choose whether to perform key backup (possibly based on user input)
+      and update account data to reflect the choice,
+
+    * OR turn key backup off.
```
richvdh, line 53: Given that this is a fresh signin, it's less "turn key backup off", and more "not enable key backup".

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+differently, while also allowing clients to choose default behaviour suitable
+for their audience.
+
+### Dynamically reacting to changes
+
+As specified, this only affects client behaviour when a user signs in. We could
+specify that clients must watch the value of this account data and dynamically
+change their key backup behaviour when it changes. This would increase the
+severity of the security issues discussed below. Whether this behaviour would be
+more or less surprising for the user is a potential discussion point.
```
richvdh, line 89: It's worth noting that clients should already notice key backup being turned OFF, because attempts to upload new keys start failing. (EW has a `NewRecoveryMethodDialog` and a `RecoveryMethodRemovedDialog`, neither of which work very well, but they exist).

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Security considerations
+
+Unencrypted account data is under the control of the server, so a malicious
+server could:
+
+* increase the user's attack surface by tricking clients into performing key backups
+  against the user's will, or
+
+* cause data loss by tricking clients into not performing key backups.
```
richvdh, line 105: ... but servers have much bigger hammers to wield at that particular nail

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+    "m.key_backup": { "enabled": true }
+}
+```
+
+If `enabled` is `true`, key backup for new sign-ins is on. If `enabled` is
+`false`, key backup for new sign-ins is off. Otherwise, the value is
+undetermined and the client should either treat it as off, or make a choice and
+update the value to reflect it.
+
+### Behaviour on sign-in
```
richvdh, line 38: Is it correct that we only check the flag on sign-in? Or do we check it at other times too?

proposals/4287-shared-key-backup-preference.md
andybalaam: https://github.com/element-hq/element-web/pull/29912 implements this in Element Web.

proposals/4287-shared-key-backup-preference.md
andybalaam: Noted in b7f2324a

andybalaam: I'm not sure whether clients should be required to bring this setting to users' attention on sign-in. Could be reasonable?

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+
+When a user signs in:
+
+* If this event type exists in account data and contains the specified property
+  in the correct format, clients MUST use it to determine whether key backups
```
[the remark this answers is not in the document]
andybalaam: Do you have an intuition for why? I'm thinking that if a client supports key backup but doesn't obey this flag it will really confuse other clients, so MUST seems appropriate, but I'm not sure of the conventions here.

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,129 @@
...
+  are enabled for a user.
```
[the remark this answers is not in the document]
andybalaam: Done in c6a6ffc8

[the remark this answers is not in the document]
andybalaam: > When we say "backups are enabled": do we mean the process of backing up new keys, the process of downloading existing keys, or both?
>
> I guess probably "both" but I think you could make a strong argument that downloading old keys could happen even if uploading new keys is disabled.

I was thinking of both, but I your comment makes me think we should only ever opt out of uploading, and downloading should always be on? Maybe that's a wider discussion though.

proposals/4287-shared-key-backup-preference.md
andybalaam: Fixed in 3d926bd2 except I talked about it being deleted, because I think that is what will cause us to notice, right?

proposals/4287-shared-key-backup-preference.md
andybalaam: Noted in d249eb20

proposals/4287-shared-key-backup-preference.md
andybalaam: That is correct in the current implementation.

proposals/4287-shared-key-backup-preference.md
richvdh: with my SCT hat on, I believe we have adequate implementations of this.

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,136 @@
...
+`m.key_backup` whose contents consist of a single field `enabled` whose value is
+boolean.
+
+For example, if I `GET
+/_matrix/client/v3/user/{userId}/account_data/m.key_backup`, the response looks
+like:
+
+```
+{
+    "m.key_backup": { "enabled": true }
```
richvdh: `GET account_data` just returns the content of the account data


```suggestion
    "enabled": true
```

proposals/4287-shared-key-backup-preference.md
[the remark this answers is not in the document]
richvdh: fair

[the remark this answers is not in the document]
richvdh:
```suggestion
  in the correct format, clients which support key backup MUST use it to determine whether key backups
```

proposals/4287-shared-key-backup-preference.md
richvdh: yes, I think so

proposals/4287-shared-key-backup-preference.md
richvdh: I'm happy with this being left up to implementations (as I think your proposal does)

proposals/4287-shared-key-backup-preference.md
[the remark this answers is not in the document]
richvdh: so the implication is that, in this situation, we enable backup even if the account data has `enabled: false`?

I don't see that as a show-stopper, but maybe we should call it out as acceptable?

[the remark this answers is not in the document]
andybalaam: Done in 09ce593b

richvdh:

- [X] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal) specified in the MSC’s PR description?
- [X] Will the MSC require a new room version, and if so, has that been made clear?
- [X] Are backwards-compatibility concerns appropriately addressed?
- [X] An introduction exists and clearly outlines the problem being solved. Ideally, the first paragraph should be understandable by a non-technical audience.
- [X] All outstanding threads are resolved
  - [X] All feedback is incorporated into the proposal text itself, either as a fix or noted as an alternative
- [X] While the exact sections do not need to be present, the details implied by the proposal template are covered. Namely:
  - [X] Introduction
  - [X] Proposal text
  - [X] Potential issues
  - [X] Alternatives
  - [X] Dependencies
- [X] Stable identifiers are used throughout the proposal, except for the unstable prefix section
  - [X] Unstable prefixes [consider](https://github.com/matrix-org/matrix-spec-proposals/blob/main/README.md#unstable-prefixes) the awkward accepted-but-not-merged state
  - [X] Chosen unstable prefixes do not pollute any global namespace (use “org.matrix.mscXXXX”, not “org.matrix”).
- [X] Changes have applicable [Sign Off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) from all authors/editors/contributors
- [X] There is a dedicated "Security Considerations" section which detail any possible attacks/vulnerabilities this proposal may introduce, even if this is "None.". See [RFC3552](https://datatracker.ietf.org/doc/html/rfc3552) for things to think about, but in particular pay attention to the [OWASP Top Ten](https://owasp.org/www-project-top-ten/).

richvdh:
This seems to me to be ready for FCP.

@mscbot fcp merge

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,144 @@
...
+undetermined and the client should either treat it as off, or make a choice and
+update the value to reflect it.
+
+### Behaviour on sign-in
+
+When a user signs in:
+
+* If this event type exists in account data and contains the specified property
+  in the correct format, clients which support key backup MUST use it to determine whether key backups
+  should be enabled.
```
turt2live: Not all clients will need this, and the "MUST...should" language can be slightly confusing. 
```suggestion
When a user signs in to a client which supports encryption and key backup:

* If this event type exists in account data and contains the specified property
  in the correct format, clients which support key backup MUST use it to determine whether key backups
  are enabled.
```

@mscbot concern TravisR - logic appears to apply to clients which don't need/support it

proposals/4287-shared-key-backup-preference.md
turt2live: I don't think the proposal adequately says that. It says clients *must* use the setting to enable key backups, with no mention of implementation specific behaviour.

@mscbot concern TravisR - Proposal appears to force clients to use server-controlled setting

turt2live (approved):
otherwise lgtm - thanks!

proposals/4287-shared-key-backup-preference.md
turt2live: uh, mscbot seems to have gotten confused and overwrote the concern 🤦 

@mscbot concern TravisR - Proposal appears to force clients to use server-controlled setting

proposals/4287-shared-key-backup-preference.md
uhoreg: The text already says "clients which support key backup" on line 43 (presumably clients that support key backup also support encryption).  And clients that don't support key backup satisfy the second bullet point because not using key backup is an option.

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,144 @@
...
+visible to users, especially at the time when it affects behaviour (on sign-in).
+
+## Unstable prefix
+
+This is currently partially implemented in Element clients using a
+reversed-sense format like:
+
+```
+{
+    "m.org.matrix.custom.backup_disabled" { "disabled": true }
```
turt2live, line 129: ftr, this should *not* have been prefixed with `m.`

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,144 @@
...
+
+    * OR not perform key backup.
+
+### Behaviour on setting change
+
+If the user turns on key backups, clients MUST set this event type in account
+data, to `"enabled": true`.
+
+If the user turns off key backups, clients MUST set this event type in account
+data, to `"enabled": false`.
```
clokep: Do we specify anything else about this today? Should clients delete any existing backups or anything of that nature?

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,144 @@
...
+server could:
+
+* increase the user's attack surface by tricking clients into performing key backups
+  against the user's will, or
+
+* cause data loss by tricking clients into not performing key backups. (But
+  servers can delete data from key backups at will, so this seems unimportant.)
+
+This can be mitigated if clients make the setting, or any change to the setting,
+visible to users, especially at the time when it affects behaviour (on sign-in).
```
clokep: How would they make it visible if they aren't monitoring the setting? 🤔

proposals/4287-shared-key-backup-preference.md
clokep: (This might also be seen as a request for a link to the current spec around this. 😄 )

proposals/4287-shared-key-backup-preference.md
turt2live: I've re-read it, but still find the wording confusing. Duplicating some text I think is fine here for the added clarity.

Another option to introduce clarity is to draw up a flow chart.

proposals/4287-shared-key-backup-preference.md
andybalaam: Reading https://spec.matrix.org/v1.17/client-server-api/#server-side-key-backups I don't see anything related to what the client should do if the user turns off key backups,

proposals/4287-shared-key-backup-preference.md
andybalaam: I guess the most sensible thing is for the client to make this visible to users at the time the information is used, on sign-in.

clokep: Maybe just remove the "any change" then?

```suggestion
This can be mitigated if clients make a mismatch in the setting visible at
the time when it affects behaviour (on sign-in). In the future it may be possible
to monitor the setting and inform users of any change to the setting.
```

clokep:
@mscbot concern Checklist incomplete

proposals/4287-shared-key-backup-preference.md
andybalaam: @turt2live I think/hope that dd149139 addresses this concern.

proposals/4287-shared-key-backup-preference.md
andybalaam: Noted in e4de388b

andybalaam:
I believe @turt2live 's concerns have been addressed. I am not sure what @clokep is referring to with "Checklist incomplete" - I think all the unchecked items are not applicable but I might be missing something.

clokep:
> "Checklist incomplete" - I think all the unchecked items are not applicable but I might be missing something.

I'd expect them to be checked even if they don't apply, but I might be misremembering our process and will defer to @turt2live about that.

tulir:
The checklist is indeed supposed to be fully checked, even the items that aren't applicable (it says so at the top). Looks like all the unchecked items were not applicable, so I went ahead and checked them

@mscbot resolve Checklist incomplete

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,151 @@
...
+{
+    "m.org.matrix.custom.backup_disabled" { "disabled": true }
+}
+```
+
+so we propose to use this form as an unstable prefix, and reverse the sense of
+the boolean when this feature is stabilised.
+
+NOTE: the use of the `m.` prefix for this property name was a mistake since
+[identifiers starting with the characters m. are reserved for use by the official Matrix specification](https://spec.matrix.org/latest/appendices/#common-namespaced-identifier-grammar).
```
turt2live:
```suggestion
[identifiers starting with the characters m. are reserved for use by the official Matrix specification](https://spec.matrix.org/v1.17/appendices/#common-namespaced-identifier-grammar).
```

(we prefer to link to a specific version, because latest can and does break links over time)

turt2live:
@mscbot resolve TravisR - Proposal appears to force clients to use server-controlled setting
@mscbot resolve TravisR - logic appears to apply to clients which don't need/support it

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,151 @@
...
+Unencrypted account data is under the control of the server, so a malicious
+server could:
+
+* increase the user's attack surface by tricking clients into performing key backups
+  against the user's will, or
+
+* cause data loss by tricking clients into not performing key backups. (But
+  servers can delete data from key backups at will, so this seems unimportant.)
+
+This can be mitigated if clients make a mismatch in the setting visible at
```
richvdh: Could you maybe clarify what you mean by "a mismatch in the setting"? Mismatched compared to what?

andybalaam: I attempted a fix in 7ae226ef

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,151 @@
...
+
+### A proper unstable prefix
+
+Because an existing implementation exists, the proposed unstable prefix has
+different semantics from the final proposal. If this is deemed unacceptable, a
+more normal unstable prefix could be used.
+
+## Security considerations
+
+Unencrypted account data is under the control of the server, so a malicious
```
ara4n, line 109: Heads up that I tripped over this fairly badly while reviewing; my thought process was: "autoenabling backup sounds scary; warning users sounds even scarier - what would the warning even say?". @tulir pointed out however that in practice current clients enable backup automatically at login - so in practice this doesn't significantly change the attack surface. 

Instead, paranoid clients which let the user opt out of backup should warn the user when they are enabling backup whatever the scenario - including at login.

Unsure if this needs to be written up on the MSC, but am commenting it here so I have something to refer back to in future.

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,151 @@
...
+  clients may wish to confirm the user's intention.)
+
+* If this event type does not exist in account data, or if it does not contain
+  the `enabled` property, or if the value of `enabled` is not a boolean value,
+  clients MUST ignore the existing value and:
+
+    * EITHER choose whether to perform key backup (possibly based on user input)
+      and update account data to reflect the choice,
+
+    * OR not perform key backup.
```
Johennes: I know this is not the spec PR but EITHER and OR are not [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) keywords and if they are used like this here, they _may_ enter the spec by accident later. 🙈 

I think this is basically saying "if you cannot read the field, do whatever you want"? In RFC2119 speak that would probably use "MAY", e.g. like this:

```
* If this event type does not exist in account data, or if it does not contain
  the `enabled` property, or if the value of `enabled` is not a boolean value,
  clients MAY decide whether or not to perform key backup (possibly based
  on user input).
```

proposals/4287-shared-key-backup-preference.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Behaviour on setting change
+
+If the user turns on key backups, clients MUST set this event type in account
+data, to `"enabled": true`.
+
+If the user turns off key backups, clients MUST set this event type in account
+data, to `"enabled": false`.
```
Johennes: In the case where the account data entry doesn't exist, it's possibly the client, not the user, that decides whether or not to enable key backups. Should that be captured here?

```
If a client turns on key backups, it MUST set this event type in account
data, to `"enabled": true`.

If a client turns off key backups, it MUST set this event type in account
data, to `"enabled": false`.
```

proposals/4287-shared-key-backup-preference.md
andybalaam: Thanks, fixed in 76d77b9d

proposals/4287-shared-key-backup-preference.md
andybalaam: Yes, fixed in 395b98ac

andybalaam: Apologies, @richvdh pointed out that this probably does constitute a substantive change, and we are in FCP so it's not valid for us to make a substantive change at this point. I will revert it. I also think there is some reasonable justification for the previous wording, so I will attempt that in a separate comment.

andybalaam: Reverted in 6f580fd4

andybalaam: Justification for the original wording (I am interested @Johennes' response):

This property is intended to capture the user's intent, so that other clients can reflect that intent in their behaviour and UI. If the client sets the property without interacting with the user, there is no need to reflect that in the setting.

Does that make sense?

Johennes: Ok, I suppose that makes sense. And yes, shame on me for commenting _after_ FCP. I only noticed after I had put in the comments, sorry. 🙈

andybalaam:
Introducing stable identifiers:

* https://github.com/ruma/ruma/pull/2457
* https://github.com/matrix-org/matrix-rust-sdk/pull/6410
* https://github.com/matrix-org/matrix-js-sdk/pull/5258
* https://github.com/element-hq/element-web/pull/33034

andybalaam:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2354

tulir:
Looks like this was merged last month, but the 🌮 comment was forgotten

PR #4291, opened 2025-05-20
[WIP] Placeholder

ara4n:


turt2live:
(for clarity: this is using the process defined by https://github.com/matrix-org/matrix-spec/pull/2157)

proposals/4291-room-ids-as-hashes.md
turt2live, line 1: Implementation requirements:

* [x] Server - https://github.com/element-hq/synapse/pull/18685
* [x] Evidence of clients not crashing

Some clients to test/verify (incomplete list):

* [x] matrix.to web
* [x] Element Web (/Desktop)
* [x] Legacy Element Android
* [x] Legacy Element iOS
* [x] matrix-rust-sdk
  * [x] Element X Android
  * [x] Element X iOS
  * [x] Fractal
* [x] Cinny
* [ ] Nheko
* [x] Quotient/Quaternion
* [x] NeoChat
* [ ] Rory&::LibMatrix
* [ ] matrix-bot-sdk (both Element fork and upstream; covers most moderation tooling/bots)
* [ ] Draupnir
* [x] Hydrogen

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+ - It’s used as a last resort `?via=` candidate [in Synapse](https://github.com/element-hq/synapse/blob/v1.133.0/synapse/handlers/room_member.py#L1177).
+
+It is also desirable to know the creator's server name without having to join and get the full create event
+for Trust & Safety purposes. The create event in invites can be spoofed, as can the room ID, but spoofing the
+room ID makes the invite unusable whereas spoofing the create event does not. We could rely on the `sender`
+instead, but for this to be effective it would be important for servers to use the `sender` domain for routing
+purposes, such that if the `sender` was invalid it would cause the invite to be unusable. On the contrary, users
+unfamiliar with Matrix may see a room ID like `!OGEhHVWSdvArJzumhm:matrix.org` and think that `matrix.org` hosts/moderates/
+is somehow responsible for the room in some way. By dropping the domain, we clearly express that the creator domain
+may not be responsible for the contents of the room.
```
turt2live, line 109: This section is mostly trying to say "you couldn't trust the domain in the room ID for invites before, but at least it was available". Tools which filter invites will need to use another metric for determining whether the invite is allowable. 

Edit: the comment this was reply to was seemingly deleted.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+search the old room state for the `m.room.tombstone` event and then jump to that.
+
+### Potential Issues
+
+It is not immediately obvious which server created the room in the first place. See the "Alternatives" section
+for a discussion on the pros/cons of this.
+
+The loss of `content.predecessor.event_id` may negatively impact the way clients stitch together timelines in upgraded rooms.
+
+### Alternatives
```
Gnuxie, line 72: Can't you just ammend `/invite` and `/join` to first fetch the create event?

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+The problem with this approach is that we cannot guarantee that servers will actually treat the keypair
+as ephemeral: they could reuse the same keypair to create a spoofed room.
+
+### Security Considerations
+
+This fixes a class of issues caused by confusable room IDs. Servers can currently perform a form of
+eclipse attack when other servers join via them by presenting a completely different room to some
+servers. This isolates the target server, controlling information that the target server sees. By
+linking together the room ID (which is user-supplied e.g via `matrix:` URIs) we can guarantee that
+all servers agree on the same create event, and thus the subsequent hash DAG that forms.
```
Gnuxie, line 126: This issue does not seem to be addressed by the proposal, and even with the changes in this MSC, it will be possible to sploof rooms with various unaddressed vectors.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Alternatives
+
+#### Keeping the domain suffix
+
+We could keep the colon+server suffix as it contains useful information about the creator's server,
+and has historically been used as routing information / auth checks in several places:
+ - event auth for `m.federate` [in Synapse](https://github.com/element-hq/synapse/blob/9d43bec/synapse/event_auth.py#L312), though the specification [states]((https://spec.matrix.org/v1.15/rooms/v11/#authorization-rules)) this should be applied to the `sender` instead - Step 3.
+ - It’s used as a last resort `?via=` candidate [in Synapse](https://github.com/element-hq/synapse/blob/v1.133.0/synapse/handlers/room_member.py#L1177).
+
```
TheArcaneBrony: Same thing applies to Rory&::LibMatrix as well: https://cgit.rory.gay/matrix/LibMatrix.git/tree/LibMatrix/RoomTypes/GenericRoom.cs?h=0a82b7e2acc18def93818d8e405bf620c328975e#n223

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+may not be responsible for the contents of the room.
+
+However, by adding the colon+server suffix we would allow malicious servers to create room IDs with _earlier room versions_
+which have the same ID as the new proposed format. For example, a malicious server who knows of an existing
+vNext room `!31hneApxJ_1o-63DmFrpeqnkFfWppnzWso1JvH3ogLM:example.com` could create a v11 room with the exact
+same ID, causing confusion. We therefore need to create a new namespace for vNext room IDs which cannot conflict
+with earlier room versions. This is why the colon+server suffix has been removed. We could alternatively encode the
+room version along with the domain to form a new namespace e.g `!localpart:domain@version` where `@` has never been
+allowed in the set of `domain` characters, but there isn't a strong enough reason to encode such information in
+every event.
```
TheArcaneBrony: Suggestion: could this be `!localpart:version@domain` instead? That way, the version would end up in the `user` field of HTTP URLs for clients/servers that naively pass through the server name (also see synapse's via= or Rory&::LibMatrix' join handling)

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
+## MSC4291: Room IDs as hashes of the create event
+
+Currently rooms are identified by a room ID [made from a random string](https://spec.matrix.org/v1.14/appendices/#room-ids)
+selected by the creating server, namespaced to the origin server. This allows a malicious server to
+create a room whose ID collides with an existing ID, causing security issues if the rooms are confused
+together.
```
Gnuxie, line 6: Can we be very specific that this proposal fixes exactly one security issue. Which is that it prevents a malicious server crafting multiple valid rooms with the same room_id using its own users and domain to do so. In order to mask the appearance of a specific room, and direct remote users into different rooms based on origin.

This proposal does not fix an issue where a server can spoof an existing room ID using a room with an invalid create event. And yet the first reading suggests this is the intent of the proposal.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+occurred, which can be relevant when there are lots of leave events after the tombstone. Clients that wish to preserve this behaviour may instead
+search the old room state for the `m.room.tombstone` event and then jump to that.
+
+### Potential Issues
+
+It is not immediately obvious which server created the room in the first place. See the "Alternatives" section
+for a discussion on the pros/cons of this.
+
+The loss of `content.predecessor.event_id` may negatively impact the way clients stitch together timelines in upgraded rooms.
+
```
TheArcaneBrony: Didn't this already affect client-driven room upgrades as they can't predict room IDs? IIRC the way that's historically been solved was to refer to the event *right before* the tombstone.

proposals/4291-room-ids-as-hashes.md
Gnuxie, line 92: When a malicious server that is eclipsing its own rooms with multiple joinable rooms is detected, the origin can be parsed from the `room_id` and all rooms that originate from the server can be blocked[^1].

This is a workaround for this specific problem. But it is also a simple method for managing the rooms created by one server that will be removed or made more complicated if this MSC goes forward, because there will still be other reasons to block a room based on its origin.  

[^1]: Yes this is illegal for compatibility with future room versions, but this is how safety tooling already protects homeservers from rooms created by malicious servers.

Gnuxie: The answer is no, because this proposal does not seem to solve this problem, see https://github.com/matrix-org/matrix-spec-proposals/pull/4291/files#r2198254526.

proposals/4291-room-ids-as-hashes.md
Gnuxie: Perhaps it would be better to propose the MSC under a different premise to security, because it now seems clear that the security issue being addressed by the proposal is not very severe and has existing workarounds[^1]. 

The premise instead could be that the origin on the room_id sucks because it confuses people into believing that the room is "hosted somehwere".

[^1]: https://github.com/matrix-org/matrix-spec-proposals/pull/4291#discussion_r2198286527

proposals/4291-room-ids-as-hashes.md
FSG-Cat: As far as i am concerned its a non issue. as in keep the current behavior that is in the spec as it will help with timeline fusing  / stitching and stuff like Element Web Jump to that uses this field and works perfectly for me atleast.

Like with the current state of /upgrade its an API that a lot of the community recomends avoiding as if its the plauge due to that its only operating mode is Carbon Copy of the predecessor room including bans and powerlevels. But the community often has all bans be Bot managed and Powerlevels often need to be changed to more flexible schemas than the homeserver default templates allow for.

You dont need to do the dance where you create the tombstone event already knowing the room ID of the sucessor room so you can use the tombstone it self as the anchor. Its a dance Synapse uses that is fancy but fragile. Doing the dance the community does with just pick a good event to declare the "end" and then use that is better.

Also removing this field i dont want to see that break Element iOS because i can say it will unless Element iOS gets hardened against this as it validates the upgrade before declaring it valid.

proposals/4291-room-ids-as-hashes.md
Gnuxie: Can we be very specific that this proposal can prevent a server from creating multiple eclipsed and joinable rooms originating from itself[^1].

And then also that the proposal does NOT address an issue where servers can eclipse existing rooms originating from other servers  by using an invalid or non existent create event. Whether that be to present a different room to servers that may or may not be joinable. Or to target a existing room by presenting it alongside other faked events. 

[^1]: ie independent rooms that all work and have a valid/authorizable dag and auth chain. But are independent of each other, yet share the same room_id.

proposals/4291-room-ids-as-hashes.md
Gnuxie: This is especially useful if we want to determine the origin of a room without needing to build a dag or perform signature checks.

proposals/4291-room-ids-as-hashes.md
Gnuxie: Draupnir will crash, and please do not assume that ticking off the matrix-bot-sdk will cover us in future. We have our own infrastructure for parsing matrix events and matrix identifiers.

- https://github.com/the-draupnir-project/matrix-basic-types includes parsing and type brands for matrix identifiers (potentially illegally in the case of room identifiers, but we do so because it is important for us to be able to identify the origin of rooms).
- https://github.com/Gnuxie/matrix-protection-suite/ parses events, and will crash if `room_id` is removed from the top level of events. Not sure whether that specifically is illegal or not. We have a tolerant strategy for handling other fields that don't match the schema (including all content), but this isn't the case for `room_id` and `event_id`.

turt2live:
There's still some work to be done to get the MSC into an mergeable state, namely implementation, but to remove some barriers to process timelines we're starting the proposal for FCP now. This comment will be followed by some concerns to block FCP on certain action items required to actually start FCP.

@mscbot fcp merge

turt2live:
@mscbot concern Server implementations not shared
@mscbot concern Incorporate feedback  shared (both publicly and privately) up to July 11, 2025 
@mscbot concern Client testing is incomplete

proposals/4291-room-ids-as-hashes.md
spaetz: But truly malicious servers can already now use any server name they want for the room id, can't they? So, this is a protection only against stupid malicious servers?

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+<img width="793" alt="Creator chicken egg problem" src="images/4291-creator-chicken-egg.png" />
+
+
+To break this cycle, this MSC removes the `content.predecessor.event_id` field from the `m.room.create` event. There appear to be no security
+reasons why this field exists. This field is sometimes used in clients to jump back to the correct part in the old room's timeline when the tombstone
+occurred, which can be relevant when there are lots of leave events after the tombstone. Clients that wish to preserve this behaviour may instead
+search the old room state for the `m.room.tombstone` event and then jump to that.
+
+### Potential Issues
```
Gnuxie, line 73: Implementations need to make changes to their API's to make sure safety tooling has access to the room origin now that it can't be extracted from the `room_id`. This is so that they can enforce moderation policies targeting servers with the intent of blocking all their rooms at the server level. 

In particular synapse's module API for `user_may_invite` and `user_may_join_room`. And also the `rooms` table synapse has https://github.com/element-hq/synapse/issues/18563.

proposals/4291-room-ids-as-hashes.md
zecakeh: Fractal doesn't crash with the changes in this MSC.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,143 @@
...
+event with the room ID sigil `!` instead of the event ID sigil `$`. For example, if a create event ID is
+`$31hneApxJ_1o-63DmFrpeqnkFfWppnzWso1JvH3ogLM`, the room ID is
+`!31hneApxJ_1o-63DmFrpeqnkFfWppnzWso1JvH3ogLM`.
+*This effectively restricts the room ID grammar to be `!` + 43x unpadded urlsafe base64 characters.*
+
+This binds a given room to a specific single `m.room.create` event, eliminating the risk of confusion.
+This can be seen as an extension of [MSC1659](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1659-event-id-as-hashes.md),
+which replaced confusable `event_id`s with a hash of the event.
+
+The `room_id` field is removed entirely from the `m.room.create` event, but is reintroduced on
```
jevolk, line 26: What is the proposed migration path for creating rooms via [PUT /{roomId}/state/m.room.create](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey) now that the path component is only known from the response?

tulir: I'd assume none as that's not a supported use case. Weird things like that can be done with implementation-specific interfaces.

jevolk: Why is it unsupported? The spec doesn't say 404 is returned when the `{roomId}` doesn't exist.

jevolk (changes requested):
The ability to make direct requests to [`PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}`](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey) has been summarily deleted from the specification by this proposal and there are no instructions proposed to note that in the changelog.

Without specifying a replacement endpoint e.g. `PUT /_matrix/client/v3/rooms/state/{eventType}/{stateKey}` this proposal breaks the existing ability to PUT state content for `m.room.create` and the ability to PUT state or send content for _any_ event in rooms which may have other initial previous events auto-generated by one-shot infallible requests.

The changelog should be updated to note this.

proposals/4291-room-ids-as-hashes.md
tulir: https://spec.matrix.org/v1.15/client-server-api/#creation

> To create a room, a client has to use the following API.

jevolk: > https://spec.matrix.org/v1.15/client-server-api/#creation
> 
> > To create a room, a client has to use the following API.

Is that normative? Is there a technical reason why an implementation cannot create a room by receiving state?

turt2live:
@mscbot resolve Server implementations not shared

proposals/4291-room-ids-as-hashes.md
turt2live: matrix.to appears to overly validate the room ID:
image

proposals/4291-room-ids-as-hashes.md
turt2live: I'd be inclined to list it as a potential issue for now, but not attempt to fix it. It's a pretty obscure part of the spec, and I wouldn't expect that clients try to use it.

proposals/4291-room-ids-as-hashes.md
turt2live: I think this is addressed by https://github.com/matrix-org/matrix-spec-proposals/pull/4311 so resolving. If not, please feel free to open new threads (I may miss replies because of how github notifications work).

proposals/4291-room-ids-as-hashes.md
KitsuneRal: In NeoChat as well, as of this writing.

proposals/4291-room-ids-as-hashes.md
KitsuneRal: I wouldn't really call it an issue - it's been like this, by design, since long ago. A brief clarification that it's not proper to create rooms merely by `PUT`ting an `m.room.create` event would do.

proposals/4291-room-ids-as-hashes.md
KitsuneRal: First off, `domain` in the room id is misleading and we want to get rid of it, partially for that reason. Second, adding components to identifiers increases complexity for more than one project (at least all homeservers would have to implement it). It looks like the language of the spec hasn't been harsh enough to discourage parsing `room_id` fields beyond the initial sigil character; however, parsing a room id _as if it were a URL_ has always been plain wrong; that it happened to work is a mere coincidence. Please let's fix the code that does it instead.

kegsay: `!localpart:version@domain` or `!localpart:domain@version` are both valid alternatives. We aren't planning on doing that though.

Part of the problem with encoding additional data in the room ID is that it simply can't be trusted without the matching create event. Let's assume we go with your `!localpart:version@domain` for a moment. For a create event with `$create_event_id` and `sender: @alice:example.com, content: { room_version: 1234 }` the room ID _should_ be `!create_event_id:1234@example.com` but nothing stops you from saying the room ID is `!create_event_id:555@evil.com`. In theory, this should just cause the room ID to be useless, but as we can see from the comments on this MSC, the room ID is being introspected _in isolation_ to determine allow/reject policies. This is exacerbated by the fact that invite requests have no full-formed create event for servers to check.

Design wise it's far more secure to not add redundant data and to have a single source of truth.

proposals/4291-room-ids-as-hashes.md
kegsay: Added.

proposals/4291-room-ids-as-hashes.md
kegsay: Added some prose.

proposals/4291-room-ids-as-hashes.md
kegsay: > Didn't this already affect client-driven room upgrades as they can't predict room IDs?

No, the `/upgrade` endpoint handles this for you by doing a convoluted dance of:
 - Create a new bare room ID,
 - Use that ID to make the tombstone event,
 - Make the new create event with the tombstone event ID.

kegsay: Added comment for Element iOS.

proposals/4291-room-ids-as-hashes.md
kegsay: Added:

> Furthermore, despite the specification stating that room IDs are opaque identifiers, both clients
and servers in the wild introspect the room ID for routing or moderation purposes. This is fundamentally
_unsafe_ as there are no guarantees the room ID is accurate without some verified events in the room.
By removing the domain from the room ID we ensure real world clients and servers cannot be manipulated
by an attacker who may claim domains they do not control.

proposals/4291-room-ids-as-hashes.md
kegsay: This approach just opens up more problems though since an attacker can take advantage of the blocking behaviour.

I've added this option to the MSC:

> #### Blocking servers which eclipse their own rooms
>
> We could make honest servers block other servers when they detect that the other server has reused a
> room ID. However, this is a reactive approach compared to the proactive approach this proposal takes.
> By relying on servers to "detect" when a room ID is reused, malicious servers can selectively disclose
> the duplicate room to target servers. This would allow the malicious server to force servers with
> admins in the room to no longer participate in the room, effectively removing the ability to moderate
> the room.

proposals/4291-room-ids-as-hashes.md
kegsay: I think you're referring to MSC4311?

kegsay: Added this to Potential Issues:

> This proposal relies on the server being able to verify the create event in the room. This generally
means the server must be joined to the room. In particular, invites/knocks are not protected against
room confusion by this proposal alone. The security guarantees of this proposal are enhanced with
[MSC4311: Ensuring the create event is available on invites and knocks](https://github.com/matrix-org/matrix-spec-proposals/pull/4311).

proposals/4291-room-ids-as-hashes.md
kegsay: > Is there a technical reason why an implementation cannot create a room by receiving state?

Clients can't have unrestricted control over the way room IDs are generated by their own server. If they could, they could cause collisions. The [PUT /{roomId}/state/m.room.create](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey) endpoint needs a room ID as a path parameter. The client cannot possibly know this, and they can't mint new room IDs as per the above statement.

Even if we ignore this, it wouldn't work since endpoints under `PUT /rooms/{roomID}/state` are authorised by "are you joined to this `roomID`?" which happens prior to processing the request. 

If you had a _new_ endpoint which just consumed events _and_ you restricted the way clients can craft room IDs (such as this proposal does) then in theory yeah you could feasibly create rooms by just sending the create event. But that's getting a bit ahead of ourselves and isn't relevant to this proposal.

proposals/4291-room-ids-as-hashes.md
FSG-Cat: Client driven upgrades skip the /upgrade API. They explicitly do NOT want it because they want to skip the server side logic like ban cloning.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,160 @@
...
+`content` so such extra keys become part of the event ID and thus the room ID. Servers may also
+prefer to apply strong rate limits to prevent this scenario.
+
+The removal of enforced server domain namespaces means event IDs truly are _global_. This means we
+rely on the security of the hash function to avoid collisions (which we've always relied on _within_ a
+room). This has always been implied due to the federation endpoint `/event/{eventID}`, but this change
+makes the data model rely on global uniqueness. See [MSC2848](https://github.com/matrix-org/matrix-spec-proposals/pull/2848)
+for more discussion.
+
+### Dependencies
```
Gnuxie, line 173: Could [MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311) be explicitly stated as a dependency for security reasons? 

https://github.com/matrix-org/matrix-spec-proposals/pull/4291#discussion_r2198437019
https://github.com/matrix-org/matrix-spec-proposals/pull/4291#discussion_r2201720395

Gnuxie: Specifically because with current room versions we can block by origin with the domain in the room ID. But with the new MSC we would need the create PDU to do that.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,160 @@
+## MSC4291: Room IDs as hashes of the create event
+
```
kegsay, line 2: Responding to https://github.com/matrix-org/matrix-spec-proposals/pull/4291#pullrequestreview-3016620822 in a thread for ease of managing feedback:

> The ability to make direct requests to [PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey) has been partially deleted from the specification by this proposal and there are no instructions proposed to note that in the changelog.

It has not been partially deleted. This is in the relation to workflows which may _create rooms_ using [PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)  (so `eventType='m.room.create'` and `stateKey=''`). You can't make valid requests in this way as you don't know the `roomID`.  The specification provides a separate endpoint to _create rooms_ and states clients have to use it: https://spec.matrix.org/v1.15/client-server-api/#creation :

> To create a room, a client has to use the following API.

If anything, this is a spec clarity issue that [PUT /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey) cannot be used to create rooms by specifying `m.room.create` as the event type and `""` as the state key.

>
> Without specifying a replacement endpoint e.g. PUT /_matrix/client/v3/rooms/state/{eventType}/{stateKey} this proposal breaks the existing ability to PUT state content for m.room.create and the ability to PUT state or send content for any event in rooms which may have other initial previous events auto-generated by one-shot infallible requests.
>
> The changelog should be updated to note this.

It's never been valid by both the spirit and letter of the law, so there is nothing broken here.

proposals/4291-room-ids-as-hashes.md
kegsay: Whilst I suspect we will want to get MSC4311 in to aid moderation, I don't think this is a hard dependency.

The moderation tooling around using the creator domain isn't particularly strong when you have a public federation: attackers can create rooms on "safe" domains and then use that to avoid detection. It does admittedly remove one useful additional datapoint though, which is mentioned in the Potential Issues section.

Some of the linked rationale is a bit weak:

> the proposal does NOT address an issue where servers can eclipse existing rooms originating from other servers by using an invalid or non existent create event. 

Today, nothing stops a malicious server giving a "safe" room ID with abusive stripped state. This proposal doesn't materially change this. Even with MSC4311 it doesn't change this since what you actually want is to verify the stripped state events (name, avatar, etc) not the create event.

In other words, I see no technical reason to block this MSC on MSC4311. I do see this MSC negatively affecting abuse detection used by moderation tooling, albeit the detection appears to be easily circumventable. If this would critically affect the functioning of moderation tooling I could see the justification for it being a hard dependency, but I just don't see it currently.

Gnuxie: > Today, nothing stops a malicious server giving a "safe" room ID with abusive stripped state. This proposal doesn't materially change this. Even with [MSC4311](https://github.com/matrix-org/matrix-doc/pull/4311) it doesn't change this since what you actually want is to verify the stripped state events (name, avatar, etc) not the create event.

This is true, however the reason to block rooms by origin is because of situations where you want to prevent users on your homeserver encountering rooms from other known communities. That in a lot of cases are not even operating maliciously, but their content is inappropriate for your server or is otherwise abusive and illegal.

Gnuxie: This is a feature that we wanted to have available in Draupnir, but we haven't been able to roll this out yet due to a setback with how we extracted origin, we used to do it from the creator of the room, but this information was unreliable (See https://github.com/element-hq/synapse/issues/18563, though note this issue only counts for rooms that have already been joined to). So we moved to extracting it from the room_id. And that's the route we'd take for blocking rooms via synapse's `user_may_join_room` callback.

JadedBlueEyes: It's already an implicit dependency as servers that don't implement this won't be able to invite servers that do in affected versions:

> If the m.room.create event is not present, not a PDU, or not for the room ID specified, the server MUST fail to continue processing the invite or knock.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,165 @@
...
+   Step 1 of the [Checks performed on receipt of a PDU](https://spec.matrix.org/v1.14/server-server-api/#checks-performed-on-receipt-of-a-pdu)
+   is modified to remove the `room_id` as a universal example of:
+   > Is a valid event, otherwise it is dropped. For an event to be valid, it must ~~contain a room_id, and it must~~ comply with the event format of that room version.
+ - Concerning the [event format](https://spec.matrix.org/v1.14/rooms/v11/#event-format-1):
+   * The "Description" field for `auth_events` is adjusted to state:
+     > Required: Event IDs for the authorization events that would allow this event to be in the room.
+     >
+     > Must contain less than or equal to 10 events. Note that if the relevant auth event selection rules are used,
+     > this restriction should never be encountered.
+     > **NEW: MUST NOT contain the create event ID (room ID).**
```
anoadragon453:
```suggestion
     > **NEW: MUST NOT contain the create event ID. This is already implied by the `room_id` field.**
```

as justification.

anoadragon453 (approved):

proposals/4291-room-ids-as-hashes.md
turt2live: The SCT is currently considering MSC4311 dependent on this MSC, but also does not currently consider MSC4311 as a hard blocker for the stream of work. We intend to bring MSC4311 through the process as rapidly as we can, but there are a number of hidden constraints which may make that hard/impossible.

turt2live:
@mscbot resolve Incorporate feedback shared (both publicly and privately) up to July 11, 2025

proposals/4291-room-ids-as-hashes.md
jevolk: > You can't make valid requests in this way as you don't know the roomID.

Except that a client could create a "custom" `roomID` this way. The only thing preventing implementations from creating rooms via this path is the rather low-key statement as quoted. I submit that's enough to resolve this and not require an entry in the changelog.

turt2live:
With the expectation that matrix.to will be fixed, I'm personally satisfied with the level of client testing showing that they won't crash for most users. Users may see some pretty awful errors in their timelines, however the client remains standing in most cases.

@mscbot resolve Client testing is incomplete

joshqou:
This proposal on it's own is fine if changed slightly, however packaging it alongside a mandatory security-related version bump is very neglectful behaviour towards both paid and unpaid maintainers of matrix software. There are many legitimate reasons as to why origin homeservers should be identifiable just from their room id, some of which are specifically outlined in this proposal, however are ignored in favour of breaking many moderation tools for... a very specific edge-case where a server gains knowledge of a room which could only be created by a malicious server?

I don't see any reason as to why this new format wouldn't be able to retain all of the existing information while also providing protection against impersonation or id "downgrading". i.e: `!:example.com@$createid`

Given those at the foundation and at element have given their approval I doubt anything can be said or done that would change this MSC in any way, however I am leaving this comment in spite of that to voice my own disapproval.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,165 @@
...
+This creates a chicken/egg problem when upgrading rooms because:
+ - The `m.room.tombstone` event needs a `content.replacement_room`.
+ - The `content.replacement_room` ID is now the create event of the new room.
+ - The create event of the new room needs the tombstone event ID in `content.predecessor.event_id`.
+ - The `m.room.tombstone` event needs ...
+
+<img width="793" alt="Creator chicken egg problem" src="images/4291-creator-chicken-egg.png" />
+
+
+To break this cycle, this MSC removes the `content.predecessor.event_id` field from the `m.room.create` event. There appear to be no security
```
HarHarLinks: To be clear: setting this field will not become forbidden when creating a room manually from the API?

It would be unfortunate to break existing room upgrade scripts, and I don't see a good reason to.

proposals/4291-room-ids-as-hashes.md
TheArcaneBrony: Resolved in Rory&::LibMatrix as of https://cgit.rory.gay/matrix/LibMatrix.git/commit/?id=c39d18efd0e6a610d2dc29832407e99e0079bc13

proposals/4291-room-ids-as-hashes.md
TheArcaneBrony: To me it sounds like event_id could simply be redefined to point to the last event as of creating the new room - which is what non-server-driven room upgrades have always done to begin with, though this information could also be derived by users by getting the event id with an extra `GET /_matrix/client/v3/rooms/:roomId/state/m.room.tombstone/?format=event` call (which would cost some overhead, which isn't nice).

TheArcaneBrony: (Besides, that's not immutable, so this doesn't stop a malicious admin from simply moving the tombstone later on)

turt2live: > To be clear: setting this field will not become forbidden when creating a room manually from the API?

Forbidden, no (at least shouldn't be). Unspecified/deprecated, yes.

tulir: > event_id could simply be redefined to point to the last event as of creating the new room

There's no need to redefine anything, that's how the spec already defines it

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,165 @@
...
+reasons why this field exists. This field is sometimes used in clients to jump back to the correct part in the old room's timeline when the tombstone
+occurred, which can be relevant when there are lots of leave events after the tombstone. Clients that wish to preserve this behaviour may instead
+search the old room state for the `m.room.tombstone` event and then jump to that.
+
+### Potential Issues
+
+It is not immediately obvious which server created the room in the first place. See the "Alternatives" section
+for a discussion on the pros/cons of this.
+
+The loss of `content.predecessor.event_id` may negatively impact the way clients stitch together timelines in upgraded rooms.
```
FSG-Cat, line 78: It would essentially break the current way of doing it with having a clear end point and start point based on the anchor that this forms in the old room. And there was absolutely never a need to do a chicken and egg problem by having to know the room ID of the successor room to create it due to sending the tombstone before the room even existed. (What you do instead is you pick the last event you know or even just send a dummy event like a reaction. Therefore no Chicken and egg problem)

Personally i dont see why this change is needed at all. It just breaks a thing that already has been implemented for no good reason and makes UX for room upgrades worse. So either we will just have to reinvent this in v13 or we get it right and keep it is how i see it. Unless it turns out this is the wrong format as far as timeline stitching / fusion is concerned.

HarHarLinks: I think the above may be slightly flawed for DAG reasons (the "last" event is not necessarily the last event for every onlooker's timeline order).

Alternatively, we could spec that a successor room finds the tombstone in the last room my filtering for room upgrade events and picking the last one (by some reliable order?) that refers to the successor room ID.

FSG-Cat: Personally i think that we should resolve any problems like that when the MSC that focuses on improving that feature gets proposed.

Like the point of my concern is that the field removal is completely unjustified as its a imaginary problem and things have worked pefectly on the community side for years without the chicken and egg problem solution that Synapse uses.

The community due to the fact the /upgrade API cant be controlled at all and you cant reliably predict what is going to happen doesnt use this API if your smart. We do manual room upgrades because the Cloning that the API does is not useful enough to outweigh the fact you get a carbon copy of your source room flaws and all. Including a bunch of spam due to all your bans being re applied when we just use Moderation bots for that.

Like we have years upon years of proof that there is no need for this removal as things work just great with the just pick a event and be happy method. And this change will break a Element Web feature that while unreliable for some users works perfectly for me on Synapse on both feline.support and matrix.org of jumping to the specific event listed in the create event. Be that the tombstone it self or a event close to it. (Helps massively with the massive block of leaves that comes after usually)

CobaltCause: I'm in agreement that this field is useful and shouldn't be removed on the basis that each room has exactly 1 create event but may have >=0 tombstone events. If a room wants to declare that it supersedes another room from a precisely known and stable ending point, a field on the create event like this is the only way to do it. My understanding is that this field was never required by anything to point to a tombstone event in particular, so I don't believe it's necessary to change anything.

turt2live:
(FCP is actually *not* finished due to outstanding comments)

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Alternatives
+
+#### Keeping the domain suffix
+
+We could keep the colon+server suffix as it contains useful information about the creator's server,
+and has historically been used as routing information / auth checks in several places:
+ - event auth for `m.federate` [in Synapse](https://github.com/element-hq/synapse/blob/9d43bec/synapse/event_auth.py#L312), though the specification [states]((https://spec.matrix.org/v1.15/rooms/v11/#authorization-rules)) this should be applied to the `sender` instead - Step 3.
+ - It’s used as a last resort `?via=` candidate [in Synapse](https://github.com/element-hq/synapse/blob/v1.133.0/synapse/handlers/room_member.py#L1177).
+ - It's used as a last resort `?via=` candidate in [Rory&::LibMatrix](https://cgit.rory.gay/matrix/LibMatrix.git/tree/LibMatrix/RoomTypes/GenericRoom.cs?h=0a82b7e2acc18def93818d8e405bf620c328975e#n223), NeoChat and others.
```
Half-Shot, line 100: It's also a **required** parameter for `m.space.child` https://spec.matrix.org/latest/client-server-api/#mspacechild.

proposals/4291-room-ids-as-hashes.md
```diff
@@ -0,0 +1,165 @@
...
+
+This supplants [MSC4051](https://github.com/matrix-org/matrix-spec-proposals/pull/4051) with more information.
+
+### Proposal
+
+We redescribe the room ID to be the event ID of the room's `m.room.create`
+event with the room ID sigil `!` instead of the event ID sigil `$`. For example, if a create event ID is
+`$31hneApxJ_1o-63DmFrpeqnkFfWppnzWso1JvH3ogLM`, the room ID is
+`!31hneApxJ_1o-63DmFrpeqnkFfWppnzWso1JvH3ogLM`.
+*This effectively restricts the room ID grammar to be `!` + 43x unpadded urlsafe base64 characters.*
```
Half-Shot, line 20: Some APIs require the user of a server_name, such as spec.matrix.org/latest/client-server-api/#mspacechild requiring there to be a valid set of server names for the `via` field. Some applications do this by splitting the room_id as a practical (if technically unsound) way to do this.

The only alternatives I see are to split the creator's user_id instead, or to store the `home_server` parameter from the login (which isn't practical for appservices or any application currently running). Are we happy to say that splitting the `sender` is the right approach?

tulir: If you're adding a room you're in to a space, then your own homeserver is the obvious first candidate for `via`s. Parsing servers from members in the room is also a common way (most clients that want more than 1 server do that).

I guess if you somehow only have the create event and no other information, you could parse the creator user ID, but that seems like a weird case. If you want to add a room that you're not in and have no information beyond the room ID, then that is no longer possible (but why would you be doing that in the first place).

> which isn't practical for appservices

Surely appservices always know name of the server they're connected to? They need to be able to form full user IDs to specify which user they're acting as

turt2live: It's safe (and expected) that clients parse user IDs to get server names, especially for places like creating `via` candidates for permalinks, so parsing `sender` (or the appservice's own user ID) is safe.

turt2live: (will add the case to the MSC though)

proposals/4291-room-ids-as-hashes.md
turt2live: (added to MSC)

proposals/4291-room-ids-as-hashes.md
turt2live: On balance, the arguments for and against removing the field I think are leaning slightly more towards removing the field. I've adjusted the MSC to clarify that it's *deprecated* rather than *removed*, which can still allow it to be set, but clients MUST NOT expect it to be set.

proposals/4291-room-ids-as-hashes.md
turt2live: addressed elsewhere in the MSC

turt2live:
Review feedback from the last several days, including from various rooms, has been addressed and as such the FCP is now finished.

turt2live:
Spec PR is a WIP out of band by me.

turt2live:
Spec PR is now in review (and open) out of band.

turt2live:
For clarity: this MSC has been assigned to room version 12.

turt2live:
Merged 🎉

PR #4292, opened 2025-05-26
MSC: Handling incompatible room versions in clients

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/client-room-version-caps/proposals/4292-incompatible-client-room-versions.md)

### DIFF
--- a/proposals/4292-incompatible-client-room-versions.md
+++ b/proposals/4292-incompatible-client-room-versions.md
@@ -1,0 +1,150 @@
+# MSC4292: Handling incompatible room versions in clients
+
+[Room versions](https://spec.matrix.org/v1.14/rooms/) are the mechanism where breaking
+behaviours to rooms are managed. These changes usually only impact server implementations,
+but there are exceptionally rare cases where clients may face breaking changes.
+
+In the case of a strongly-typed client, they may experience significant incompatibility
+with room versions which change core event type schemas, such as `m.room.redaction` in
+[MSC2244 (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2244-mass-redactions.md).
+
+This proposal introduces a mechanism for clients to express their compatible room versions
+to the server, so they are less likely to encounter issues when a user joins that room. The
+user experience of showing "incompatible rooms" is left as a client implementation detail.
+
+## Proposal
+
+A new query parameter and response section are added to [`/sync`](https://spec.matrix.org/v1.14/client-server-api/#get_matrixclientv3sync). Clients can add any number of `supported_room_version`
+query parameters to indicate what room versions it supports to the server. This is preferred over
+a filter to avoid clients needing to re-initial sync to change that filter, especially as new room
+versions are introduced in future.
+
+The server takes the unique set of `supported_room_version` parameters and determines which room
+versions are compatible with that set. The compatible set may be larger than the supported set if
+there are room versions which don't contain breaking changes for clients. These room versions will
+be identified in the spec, and currently include the following:
+
+* [Room Version 2](https://spec.matrix.org/v1.14/rooms/v2/) - No client changes.
+* [Room Version 5](https://spec.matrix.org/v1.14/rooms/v5/) - No client changes.
+* [Room Version 6](https://spec.matrix.org/v1.14/rooms/v6/) - While the redaction algorithm did change,
+  it is not backwards incompatible for clients. The effects of redactions may be greater than intended,
+  however can be relatively easily fixed after a client bugfix.
+* [Room Version 7](https://spec.matrix.org/v1.14/rooms/v7/) - Knocking was added, but not in a way
+  which requires clients to implement it right away.
+* [Room Version 8](https://spec.matrix.org/v1.14/rooms/v8/) - Restricted join rules were added, but
+  like in v7, not in a way which requires implementation.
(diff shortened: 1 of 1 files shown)


proposals/4292-incompatible-client-room-versions.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

proposals/4292-incompatible-client-room-versions.md
```diff
@@ -0,0 +1,150 @@
...
+  however can be relatively easily fixed after a client bugfix.
+* [Room Version 7](https://spec.matrix.org/v1.14/rooms/v7/) - Knocking was added, but not in a way
+  which requires clients to implement it right away.
+* [Room Version 8](https://spec.matrix.org/v1.14/rooms/v8/) - Restricted join rules were added, but
+  like in v7, not in a way which requires implementation.
+* [Room Version 9](https://spec.matrix.org/v1.14/rooms/v9/) - Adds to the redaction algorithm, making
+  the effect similar to v6.
+* [Room Version 10](https://spec.matrix.org/v1.14/rooms/v10/) - Adds another join rule, similar to
+  v7 and v8.
+* [Room Version 11](https://spec.matrix.org/v1.14/rooms/v11/) - Reduces the set of redacted fields on
```
FSG-Cat, line 40: V11 as far as I can recall removes the creator field from m.room.create and broke element iOS pre patch.

proposals/4292-incompatible-client-room-versions.md
```diff
@@ -0,0 +1,150 @@
...
+The server takes the unique set of `supported_room_version` parameters and determines which room
+versions are compatible with that set. The compatible set may be larger than the supported set if
+there are room versions which don't contain breaking changes for clients. These room versions will
+be identified in the spec, and currently include the following:
+
+* [Room Version 2](https://spec.matrix.org/v1.14/rooms/v2/) - No client changes.
+* [Room Version 5](https://spec.matrix.org/v1.14/rooms/v5/) - No client changes.
+* [Room Version 6](https://spec.matrix.org/v1.14/rooms/v6/) - While the redaction algorithm did change,
+  it is not backwards incompatible for clients. The effects of redactions may be greater than intended,
+  however can be relatively easily fixed after a client bugfix.
```
clokep, line 31: The double negative here is confusing?

```suggestion
* [Room Version 6](https://spec.matrix.org/v1.14/rooms/v6/) - While the redaction algorithm did change,
  it is backwards compatible for clients. The effects of redactions may be greater than intended,
  however can be relatively easily fixed after a client bugfix.
```

Or maybe "mostly backwards compatible"?

proposals/4292-incompatible-client-room-versions.md
```diff
@@ -0,0 +1,150 @@
...
+A new query parameter and response section are added to [`/sync`](https://spec.matrix.org/v1.14/client-server-api/#get_matrixclientv3sync). Clients can add any number of `supported_room_version`
+query parameters to indicate what room versions it supports to the server. This is preferred over
+a filter to avoid clients needing to re-initial sync to change that filter, especially as new room
+versions are introduced in future.
+
+The server takes the unique set of `supported_room_version` parameters and determines which room
+versions are compatible with that set. The compatible set may be larger than the supported set if
+there are room versions which don't contain breaking changes for clients. These room versions will
+be identified in the spec, and currently include the following:
+
```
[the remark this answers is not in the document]
FSG-Cat: I was thinking about this in the sense of that we should definetively document problems clients have run in to in the past when we define the known ways a client might have to adapt to the existing 11 room versions.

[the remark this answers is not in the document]
TheArcaneBrony: To respond to the *initial* comment here: [LibMatrix](https://cgit.rory.gay/matrix/LibMatrix.git/) breaks on any schema change, most room versions are currently *compatible* if no funky stuff has happened (ie string/bool PLs).

Generally, I tend to implement support for something only when either I need something or something breaks. I would be okay opting out of everything except the latest room version, though on the other hand having to update the SDK to add *explicit* support for a new room version can be quite annoying.

proposals/4292-incompatible-client-room-versions.md
```diff
@@ -0,0 +1,150 @@
...
+A new query parameter and response section are added to [`/sync`](https://spec.matrix.org/v1.14/client-server-api/#get_matrixclientv3sync). Clients can add any number of `supported_room_version`
+query parameters to indicate what room versions it supports to the server. This is preferred over
+a filter to avoid clients needing to re-initial sync to change that filter, especially as new room
+versions are introduced in future.
+
+The server takes the unique set of `supported_room_version` parameters and determines which room
+versions are compatible with that set. The compatible set may be larger than the supported set if
+there are room versions which don't contain breaking changes for clients. These room versions will
+be identified in the spec, and currently include the following:
```
tulir, line 25: The current spec technically doesn't allow clients to assume that vX+1 is compatible with vX. We probably need to explicitly define that clients are allowed to assume that any unknown stable room versions are equivalent to the latest stable room version, because the ones that aren't equivalent get moved to `incompatible_rooms`.

For example, if clients advertise support for rooms up to v12, then v13 is added with no client-facing changes, clients need to be allowed to assume that v13 has creator power and other v12 changes included.

(It could be a follow-up MSC rather than part of this, but necessary for a complete solution anyway)

PR #4293, opened 2025-05-27
MSC4293: Redact on kick/ban

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/user-redact/proposals/4293-redact-on-ban.md)

----

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4293#issuecomment-2988694462)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4293#issuecomment-3009757102)

### DIFF
--- a/proposals/4293-redact-on-ban.md
+++ b/proposals/4293-redact-on-ban.md
@@ -57,7 +57,7 @@ redactions" to ensure the room is cleaned up, though those senders can begin to 
 as best effort (currently, moderation bots in particular work incredibly hard to ensure they get *every*
 event redacted, but can run into delivery, reliability, and completeness issues with 1:1 redactions).
 
-It's also important to note that this proposal intends to compliment mass redactions and coexist in
+It's also important to note that this proposal intends to complement mass redactions and coexist in
 whatever room version mass redactions land in. This proposal has narrow scope and is fairly blunt as
 a measure, which may not be desirable in all situations. For example, only a user's last dozen or so
 messages may be worth redacting rather than the hundreds they sent prior to the ones which got them


proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,103 @@
...
+    "redact_events": true
+  }
+}
+```
+
+The new field is proxied through to the event by the [`/kick`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3roomsroomidkick)
+and [`/ban`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3roomsroomidban)
+sugar APIs, like `reason` is.
+
+## Potential issues
```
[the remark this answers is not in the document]
Gnuxie: So just to be clear by the server do we mean that the sender of the ban also used to send redactions or is that just intentionally ambiguous because it is one of many possible options?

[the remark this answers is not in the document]
turt2live: The server which receives (and soft fails) the event would apply redaction locally. No actual redaction events are sent out by this MSC - they're all locally applied.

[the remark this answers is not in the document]
Gnuxie: ah nice one

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,103 @@
...
+
+It's a little annoying that the flag is redacted when the membership event is redacted, however it's
+extremely rare for a moderator/admin to redact a kick or ban event. We can fix this in a room version
+trivially if it proves to be an issue.
```
Gnuxie, line 64: In consideration with:

>  or another membership event removes the field.

There could be an issue when joining a room with a long history with participating servers that do not implement this MSC. And the newly joining server backfills from one of the servers that has unredacted events sent by a user that has since rejoined/or their ban event was redacted. It is a rare circumstance but it's worth thinking about a little to see if there are any other implications. Dealing with old servers that do not implement the MSC is going to be an issue for sure, but hopefully that won't cause too many problems.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,103 @@
...
+other event.
+
+## Alternatives
+
+Mass redactions are the cited major alternative, where a single event can target approximately 1500
+other events in the room. New rooms can benefit from that functionality, especially for cases not
+covered by this proposal, while existing rooms can be given an option to protect their users with
+relative ease.
+
+## Security considerations
```
turt2live, line 80: We still need to make sure that the sender can actually do redactions in the room

proposals/4293-redact-on-ban.md
turt2live, line 1: Implementation requirements:
* [x] Server - https://github.com/element-hq/synapse/pull/18540
* [x] Sending client - https://github.com/matrix-org/mjolnir/pull/612
* [x] Receiving/applying client:
  * https://github.com/matrix-org/matrix-js-sdk/pull/4867
  * https://github.com/gomuks/gomuks/pull/626

Other non-qualifying (as of writing) implementations:
* https://github.com/matrix-org/matrix-rust-sdk/pull/5273

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,117 @@
...
+This flag is a boolean and, when `true`, causes servers (and clients) to redact all of the user's events
+as though they received an [`m.room.redaction`](https://spec.matrix.org/v1.14/client-server-api/#mroomredaction),
+including adding [`redacted_because`](https://spec.matrix.org/v1.14/client-server-api/#redactions) to
+`unsigned` where applicable. An `m.room.redaction` event is not actually sent, however.
+
+**Note**: This also means that if the user was kicked/banned with a `reason`, that event is automatically
+compatible with the redaction `reason` field and shows up accordingly.
+
+Similar to regular redactions, if the sender of the membership event can't actually redact the target's
+events, the redaction doesn't apply. This means having a power level higher than or equal to `redacts`
```
anoadragon453:
```suggestion
events, the redaction doesn't apply. This means having a power level higher than or equal to `redact`
```

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,117 @@
...
+to existing room versions.
+
+This proposal suggests adding a new flag to membership events to indicate to clients and servers that
+all of that user's events should be redacted in addition to being kicked or banned. The flag isn't
+protected from redaction itself, so may have some consistency issues, but overall should still provide
+relatively high amounts of protection to rooms.
+
+This proposal is exploratory and subject to change. Implementations may validate the idea through
+early feature support, but MUST expect that things will change (or become completely rejected).
+
```
turt2live: for when the MSC is ready for FCP:

```suggestion
```

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,117 @@
...
+    "redact_events": true
+  }
+}
+```
+
+The new field is proxied through to the event by the [`/kick`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3roomsroomidkick)
+and [`/ban`](https://spec.matrix.org/v1.14/client-server-api/#post_matrixclientv3roomsroomidban)
+sugar APIs, like `reason` is.
+
+## Potential issues
```
Gnuxie, line 199: Moderation tooling that naively adds the flag to ban events can make itself think it has already sent real redactions for events when going over `/messages` pagination

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,117 @@
...
+(unban) event. Or, when the user rejoins after a kick, the flag is implicitly dropped.
+
+Events which are delivered after the ban are likely [soft failed](https://spec.matrix.org/v1.14/server-server-api/#soft-failure)
+and are still redacted if the current membership event in the room has a valid `redact_events`
+field.
+
+Other membership states with the flag no-op, such as joins, knocks, and invites.
+
+Moderation bots and similar MAY still wish to issue (mass) redactions upon kick/ban to protect users
+on servers or clients which don't have this feature.
```
Gnuxie: It would be nice to have servers do this automatically so that bots don't have to paginate `/messages` and call `/redact` especially since this can trigger the homeserver to backfill and take a very long time to respond to the request.

proposals/4293-redact-on-ban.md
turt2live: This sounds like a software bug in that tooling rather than the MSC?

Gnuxie: The MSC should take steps to prevent people writing bugs

turt2live: where possible, yes. For a bug like this though, unit/integration tests should catch it rather than any function the MSC can provide.

Did this kind of bug happen somewhere? If so, it'd be good to have a link so I can see if there's something that can be done to avoid it.

proposals/4293-redact-on-ban.md
williamkray: yes, i would agree that if the server is already taking steps to make it _seem_ like these messages are redacted without sending the full redaction event, they should be responsible for handling a background job to "true up" that reality. 

i don't know if this is accurate, but even if the bot fetches the /messages endpoint, i think there's still a potential that bot may or may not recognize the pseudo-redactions, and similarly get stuck in endless loops of thinking it needs to redact messages that have already been redacted.

proposals/4293-redact-on-ban.md
Gnuxie: Mjolnir and Draupnir have a complicating behavior with the `--limit` option on the redact command and redaction behavior. If the redact command is used to redact a user's events within a room, but they have sent more events than the limit, using the redact command again causes the bots to paginate `/messages` up to the same point. This is because currently they do not ignore already redacted events and so calls the `/redact` endpoint for them again. This means that the user of the redact command has to use a workaround of increasing the `--limit` option. When the expected behavior for most users is simply to run the command again with no change to the options. If we were to change this behavior to avoid the unintuitive workaround by skipping already redacted events. Then when the redact on ban flag is used, then the redact command (and the bots' general redaction functionality) will no longer work.

It's fair enough to consider that this behavior is intentional in Mjolnir but it has been in the cross hairs for Draupnir for some time and does come up somewhat frequently https://github.com/the-draupnir-project/Draupnir/issues/426.

turt2live: Can the bot inspect the `redacted_because` field to determine *why* it was redacted?

Gnuxie: Yes, that would be a way of solving this issue. I was unaware that `redacted_because` includes the entire event and assumed that it only included the reason. My bad.

turt2live: Alright, I'll add advice similar to that to the MSC. Thanks!

proposals/4293-redact-on-ban.md
turt2live: I'll put this in as a SHOULD for servers, as there may be a reason for them to not do it (project scope, implementation effort, etc)

turt2live:
With my SCT hat, I've verified the implementations (though welcome second/third opinions)

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,148 @@
...
+server supports the feature or unredacted events otherwise.
+
+To (primarily) help protect users on unsupported *clients*, implementations SHOULD continue to try
+sending individual redaction events in addition to the redact-on-ban flag. They MAY cease to do so
+once they are comfortable with the level of adoption for this proposal. Servers in particular SHOULD
+assist clients and send individual redaction events on their behalf, meaning clients SHOULD wait a
+little bit before trying to issue redactions themselves. For example, a client may ban a user, wait
+a minute, then start sending redactions if it hasn't seen an `m.room.redaction` event targeting some
+of the banned user's events. Servers MAY deduplicate redactions to lower federation load, as they
+always could.
```
Gnuxie: Would it be possible to add a flag to client capabilities if the server has implemented the behavior to redact on the client's behalf?

The idea of the client needing to implement a watchdog and enabling one after each ban before issuing redactions in a time sensitive context (CSAM + pings) is problematic. I also worry that this will be quite a brittle thing to implement in the context of Matrix sync and complicated if it wants to avoid a resource leak.

turt2live: Discussed elsewhere, I'd like to avoid a capability because it's expected that client adoption in particular will be relatively quick. Capabilities are long-term things which can imply to clients/servers that the functionality is optional when our intention is only to make it optional to *send* (ie: via bots), but not *receive*. 

Clients/bots can additionally just send duplicate redactions without a timer. This solution may be preferred by bots which expect to drop the legacy redaction behaviour in relatively short time.

Those bots may still have to hit `/messages`, causing other problems, but the hope is that this remains temporary and they can 'upgrade' to some form of mass redactions sooner rather than later.

I'll aim to clarify all of this in the MSC.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,148 @@
...
+Similar to regular redactions, if the sender of the membership event can't actually redact the target's
+events, the redaction doesn't apply. This means having a power level higher than or equal to `redacts`
+*and* `events["m.room.redaction"]` (if set). We maintain the `events` check despite not actually sending
+events of that type to keep the same expectations within rooms. If the sender doesn't have permission
+to redact an event normally, no redaction is applied.
+
+If the sender is allowed to redact, the redaction behaviour continues until the membership event itself
+is redacted (thus removing the field) or another membership event removes the field. For example, if
+the user is unbanned, the moderator MAY NOT choose to carry the `redact_events` flag to that kick
+(unban) event. Or, when the user rejoins after a kick, the flag is implicitly dropped.
```
turt2live: This needs to clarify that already-redacted events are not unredacted. Once the `redact_events` flag causes an event to be redacted, it stays redacted.

turt2live:
This has gone through quite a few rounds of review in various T&S circles, and has healthy code review to show that teams are interested in supporting this feature. 

With my SCT hat, I think this is ready for FCP, especially to help encourage more client implementations.

@mscbot fcp merge

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+compatible with the redaction `reason` field and shows up accordingly.
+
+Similar to regular redactions, if the sender of the membership event can't actually redact the target's
+events, the redaction doesn't apply. This means having a power level higher than or equal to `redacts`
+*and* `events["m.room.redaction"]` (if set). We maintain the `events` check despite not actually sending
+events of that type to keep the same expectations within rooms. If the sender doesn't have permission
+to redact an event normally, no redaction is applied.
+
+If the sender is allowed to redact, the redaction behaviour continues until the membership event itself
+is redacted (thus removing the field), another membership event removes the field, or the flag is set
```
clokep: Why redact the event itself instead of just stopping at it?

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+little bit before trying to issue redactions themselves. For example, a client may ban a user, wait
+a minute, then start sending redactions if it hasn't seen an `m.room.redaction` event targeting some
+of the banned user's events. Servers MAY deduplicate redactions to lower federation load, as they
+always could.
+
+**Note**: It is possible due to implementation and real-world constraints that not all individual
+redactions will "make it" over federation to another server. This is why mass redaction approaches
+are preferred, as they are significantly more reliable.
+
+## Potential issues
```
clokep, line 199: Doesn't this potentially have weird behavior if a server redacts the events, redacts the membership event itself, then backfills events? It is unclear to me if this is covered the text below (or above in fallback behavior?)

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+where applicable. An `m.room.redaction` event is not actually sent, however.
+
+**Note**: This also means that if the user was kicked/banned with a `reason`, that event is automatically
+compatible with the redaction `reason` field and shows up accordingly.
+
+Similar to regular redactions, if the sender of the membership event can't actually redact the target's
+events, the redaction doesn't apply. This means having a power level higher than or equal to `redacts`
+*and* `events["m.room.redaction"]` (if set). We maintain the `events` check despite not actually sending
+events of that type to keep the same expectations within rooms. If the sender doesn't have permission
+to redact an event normally, no redaction is applied.
```
clokep: Can a user send this when leaving a room? Not in the case of kick but to self redact all.

proposals/4293-redact-on-ban.md
turt2live: err, sorry: the membership event is not redacted by the autoredaction stuff. It's just that since we're doing this without a room version, the flag is at risk of being redacted off the event (technically). 

We otherwise keep redacting after the ban because of eventual consistency and delayed receipt of events.

proposals/4293-redact-on-ban.md
turt2live: it should be covered in the potential issues, but I'll aim to clarify in the text. The short version is: yes, it's weird, but not any more weird than usual. A server which requests events from a server that redacted them will receive redacted copies. This infectious behaviour is a feature, though if a server *really* wanted to it could ask other servers until it finds an unredacted copy.

proposals/4293-redact-on-ban.md
turt2live: No, self-leaves are not permitted by this MSC. Rationale being it opens a can of worms I'd rather not get this MSC stuck in.

proposals/4293-redact-on-ban.md
clokep: I feel like this says the opposite?

> until the membership event itself is redacted

proposals/4293-redact-on-ban.md
clokep: I'm worried about the opposite -- receiving unredacted events and not knowing they should be redacted.

proposals/4293-redact-on-ban.md
clokep: Ok, I think that's unclear that the behavior only applies if the sender and state key don't match then.

proposals/4293-redact-on-ban.md
turt2live: I can see how that might be interpreted that way, but it is not the intention. It's trying to clarify to server implementations that they won't be able to see the `redact_events` flag on the event if the membership event itself is redacted, and therefore the autoredact behaviour also ceases at that point in time.

proposals/4293-redact-on-ban.md
turt2live: That should already be addressed in the MSC, as a known limitation.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
+# MSC4293: Redact on ban
+
+[MSC2244 (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2244-mass-redactions.md)-style
+mass redactions are incredibly helpful for cleaning up large volumes of spam, but still require sending
+a dedicated event to clean up the spam. In a typical case, a user will get kicked/banned from a room
+and the moderators will further redact some or all of their messages. Mass redactions have more use
+cases, but the specific case of "redact everything upon ban" is something which may be easily backported
+to existing room versions.
```
dbkr: I don't think I'm quite getting this proposal. So it's just a temporary measure to do mass redactions in old room versions? Why not just a separate 'mass redact' endpoint? Or, for that matter, have the HS look for mass redact events sent into old room versions and do them manually exactly as it would do here?

Gnuxie: MSC4194 is a prior exploration of a mass redact endpoint for this use case. A reason for collecting the responsibilities into a flag on the ban event are because of the following complications:

- The homeserver applying redactions would also need to send redactions for future soft failed events as well as already received historical events.
  + Toggling this behavior is complicated: should a user rejoin, the behavior would probably need to be tied to membership in the same way as in this proposal.
- The endpoint will always be used after a room member is banned. There isn't another use case with exception of self redaction which is intentionally being ignored by both proposals.
- Without mass redaction events, federating individual redaction events takes longer than a single ban event which can flag to clients to hide all content associated with a user before redaction events arrive. Which is critical in spam attacks[^1].

[^1]: And also consider that soft failure is common and it's not likely that there will be just "one" mass redact event needed to cleanup per spam incident.

proposals/4293-redact-on-ban.md
clokep: I've read this a handful of times now and I can't interpret it any other way.

```suggestion
If the sender is allowed to redact, the redaction behaviour continues until a membership event for
that user is reached which does not have the flag set or has the flag set
```

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+relatively high amounts of protection to rooms.
+
+## Proposal
+
+A new flag is added to [`m.room.member`](https://spec.matrix.org/v1.14/client-server-api/#mroommember)
+events where the target user is kicked or banned: `redact_events`. This flag is a boolean and, when
+`true`, causes servers (and clients) to redact all of the user's events as though they received an
+[`m.room.redaction`](https://spec.matrix.org/v1.14/client-server-api/#mroomredaction), including
+adding [`redacted_because`](https://spec.matrix.org/v1.14/client-server-api/#redactions) to `unsigned`
+where applicable. An `m.room.redaction` event is not actually sent, however.
```
timedoutuk: Later on in the proposal, it's said several times that sending redactions as fallback support is intended, so I'm getting mixed signals here. When this flag is passed, should the server send *actual* redactions or not? I assume it's "yes, send actual redactions, until greater adoption makes them redundant" from later paragraphs, however it is not clear.

Gnuxie: > "yes, send actual redactions, until greater adoption makes them redundant"

If this is the implication then it shouldn't be. Redactions will always need to be sent in legacy room versions. And if the proposal has been written with otherwise then that's a pretty serious issue. @turt2live could you please clarify?

My understanding was that redaction events would have to be sent for all room versions prior to the inclusion of [MSC4298](https://github.com/matrix-org/matrix-spec-proposals/pull/4298).

turt2live: TLDR: "yes, send actual redactions ***if you want to***, until greater adoption makes them redundant" (added emphasis mine) is accurate.

----

Fallback behaviour is a bit of an awkward case for MSCs. In this case, the intended change is meant to be additive to the spec with no actual requirement to send redaction events (even in existing room versions), but there are several implementations of Matrix clients and servers which already exist today that won't know how to handle this new behaviour, and thus senders are *encouraged* (not required) to send "fallback redactions". 

The MSC does give some guidance on what fallback behaviour might look like, but leaves removal of that fallback behaviour up to the individual projects. One project may, for example, decide that pre-MSC4298 rooms will forever get fallback redactions while another decides that 3 months from spec release there is enough client implementations to safely drop the fallback support. Some implementations may also choose somewhere in the middle of that spectrum: instead of chasing events and retrying redactions, it could do a single query and send redactions once, leaving any errors to hopefully be picked up by the redact-on-ban flag.

Neither this MSC or MSC4298 make the fallback behaviour required, either for existing room versions or implementations of Matrix.

proposals/4293-redact-on-ban.md
turt2live: As Gnuxie mentions, we can't/shouldn't be sending 1:1 redaction events anymore, especially post-ban (the original sender can't "self-redact", and the moderator who did the ban may have left). This MSC specifically tries to reduce the number of redaction events to zero if it can, or one in the worst case - mass redactions themselves would be the "one" case. 

The issue with mass redact events is that they aren't backwards compatible to existing room versions, which will cause breakage. We could spend some time trying to fix this (fallback fields, different event types, etc), but those solutions don't look as great (in my opinion) and slow down the rollout of the feature due to broader use cases needing consideration. 

We (T&S) expect that both redact-on-ban and mass redactions will be used in future room versions concurrently for different use cases, and require a more immediate solution to 1:1 redaction event sending.

I'll cover this in the alternatives section when I loop back around for an edit.

turt2live:
@mscbot concern General clarity/understanding of goals and alternatives

proposals/4293-redact-on-ban.md
dbkr: I'm afraid I may be more confused if anything. I was assuming the HS would do something to make this backwards compatible but now I wonder if I've misunderstood that. What makes this backwards compatible given a mass redaction isn't? If this is not a short-term measure, in what circumstances would this be used vs a mass redaction? In any case, hopefully the edit should make things clearer. The info in the responses here probably needs to go into the MSC.

proposals/4293-redact-on-ban.md
anoadragon453: This proposal should be more explicit about clients needing to implement watching for the `redact_events` field and redacting events locally upon seeing it. I could see a client developer walking away from reading this proposal thinking that it will magically work once the homeserver implements support for it.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+protected from redaction itself, so may have some consistency issues, but overall should still provide
+relatively high amounts of protection to rooms.
+
+## Proposal
+
+A new flag is added to [`m.room.member`](https://spec.matrix.org/v1.14/client-server-api/#mroommember)
+events where the target user is kicked or banned: `redact_events`. This flag is a boolean and, when
+`true`, causes servers (and clients) to redact all of the user's events as though they received an
+[`m.room.redaction`](https://spec.matrix.org/v1.14/client-server-api/#mroomredaction), including
+adding [`redacted_because`](https://spec.matrix.org/v1.14/client-server-api/#redactions) to `unsigned`
```
zecakeh: The content of `redacted_because` is not clear. It is the corresponding `m.room.member` event, right?

turt2live: correct. I'll attempt to clarify with an example in the MSC.

In short:

```json
{
  "type": "m.room.message",
  "other": "fields",
  "unsigned": {
    "redacted_because": {
      "type": "m.room.member",
      "other": "fields",
      "content": {
        "membership": "ban",
        "redact_events": true,
        "reason": "spam"
      }
    }
  }
}
```

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,167 @@
...
+already auto-redacted to avoid duplicate work.
+
+With respect to the fallback behaviour, a client might not know if a server is applying fallback
+redactions and may not wish to wait an arbitrary amount of time to see if it does. One solution would
+be to have the server expose a [capability](https://spec.matrix.org/v1.15/client-server-api/#capabilities-negotiation),
+however such a flag would be longer lived than the fallback behaviour itself (hopefully). Instead,
+clients which don't implement watchdog functionality SHOULD send redactions anyway, even if it
+duplicates the server's fallback efforts. Further, as already mentioned above, server MAY deduplicate
+redactions to lower their federation load, though this is closer to a SHOULD considering clients are
+already sending their own redaction events (like in the case of Mjolnir).
```
Gnuxie: Just documenting that `/messages` is not nice for clients to use at the moment in this scenario and why the fallback behavior is a problematic https://github.com/element-hq/synapse/issues/18829

turt2live: this MSC does apply to existing room versions, and is in use already in the Foundation's public rooms.

proposals/4293-redact-on-ban.md
turt2live: I believe I've made it clearer in b06afd7d6d3f54a1939d907b4afdd8c2668627f1 what this proposal's purpose is, and the direction it intends to go in. If there's still clarity issues, please open a new thread so I don't miss it 😅

proposals/4293-redact-on-ban.md
turt2live: I believe this should be clarified in 244082afac23617e0857df2728e60ef1e4263945

proposals/4293-redact-on-ban.md
turt2live: Clarified in a78f536bb944163dacd29bff5eec6fae4feba4e3

proposals/4293-redact-on-ban.md
turt2live: I believe this is clarified in 244082afac23617e0857df2728e60ef1e4263945

turt2live: should be clearer in 244082afac23617e0857df2728e60ef1e4263945

proposals/4293-redact-on-ban.md
turt2live: clarified in 244082afac23617e0857df2728e60ef1e4263945

proposals/4293-redact-on-ban.md
turt2live: The MSC now acknowledges that it's not-great, but continues to recommend it in 244082afac23617e0857df2728e60ef1e4263945

turt2live:
I believe this should be ready for re-review from folks, with an aim towards FCP. There have been no major changes to the core proposal, though the fallback behaviour has shifted in favour of "continue trying your best to send redactions alongside the ban, if you want". 

@mscbot resolve General clarity/understanding of goals and alternatives

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,266 @@
...
+## Proposal
+
+A new flag is added to [`m.room.member`](https://spec.matrix.org/v1.14/client-server-api/#mroommember)
+events where the target user is kicked or banned: `redact_events`. Discussed later, this new flag has
+no effect when applied to self-leaves or other membership states. This flag is a boolean and has two
+effects when `true`:
+
+1. Clients apply the redaction algorithm to events sent by that user which are cached locally, up to
+   the point of the user's previous membership event. No [`m.room.redaction`](https://spec.matrix.org/v1.14/client-server-api/#mroomredaction)
+   events are sent by the client, but the user's events are redacted as though there was such an event.
```
Gnuxie, line 75: What about if a moderator bans a user, and then another moderator wants to redact their events by banning again with the flag?

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,266 @@
...
+3. Alice leaves the room.
+4. For whatever reason, Alice rejoins the room.
+5. Alice sends events D, E, and F.
+6. Bob bans Alice with `redact_events: true`.
+7. Clients and servers apply redactions to events D, E, and F, but *not* A, B, or C. No actual
+   `m.room.redaction` events are sent in this example.
+8. Alice's event F arrive's late to Bob's server.
+9. Bob's server soft fails F because it fails current auth state (Alice is banned), and redacts it
+   because `redact_events: true` is set on Alice's ban. There are still no actual `m.room.redaction`
+   events sent here.
```
Gnuxie: What if spammers exploit this mechanic by leaving and rejoining while spamming a room?

turt2live: identified in 209cad309df660521fa161135e9261ef8ade7a65 - it's not something this MSC currently intends to fix.

proposals/4293-redact-on-ban.md
turt2live: identified in 209cad309df660521fa161135e9261ef8ade7a65 - it's not something this MSC currently intends to fix.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+          "reason": "spam",
+          "redact_events": true
+        }
+      }
+    }
+   }
+   ```
+
+   Note that because `m.room.redaction` supports a `reason` field in the same place as `m.room.member`,
+   clients which look up that reason by going `event["unsigned"]["content"]["reason"]` will still get
```
anoadragon453, line 107: Do you mean ``event["unsigned"]["redacted_because"]["content"]["reason"]``?

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+   a renderable, human-readable, string for their UI regardless of how the event was actually redacted.
+
+When `redact_events` is `false`, it acts as though it was not specified: no redactions are applied,
+exactly like how bans work prior to this proposal.
+
+Prior to applying the effects of redaction above, clients and servers MUST ensure that the sender of
+the kick or ban has power level to redact the target user's events. This means having a power level
+higher than or equal to `redact` *and* `events["m.room.redaction"]` (if set). We maintain the `events`
+check despite not actually sending events of that type to keep the same expectations within rooms. If
+the sender doesn't have permission to redact an event normally, no the `redact_events` flag is ignored
```
anoadragon453:
```suggestion
the sender doesn't have permission to redact an event normally, the `redact_events` flag is ignored
```

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+  "sender": "@mod:example.org",
+  "content": {
+    "membership": "ban",
+    "reason": "flooding", // this is copied to `redacted_because`, leading to clients showing it
+    "redact_events": true
+  }
+}
+```
+
+An example sccenario would be:
```
anoadragon453:
```suggestion
An example scenario would be:
```

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+
+1. Alice joins the room.
+2. Alice sends events A, B, and C.
+3. Alice leaves the room.
+4. For whatever reason, Alice rejoins the room.
+5. Alice sends events D, E, and F.
+6. Bob bans Alice with `redact_events: true`.
+7. Clients and servers apply redactions to events D, E, and F, but *not* A, B, or C. No actual
+   `m.room.redaction` events are sent in this example.
```
anoadragon453, line 164: What happens if Alice simply sends another `join`?

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+   of thing happening. Moderation teams are also encouraged to develop operating procedures which
+   limit the opportunity for accidentally encountering this case.
+
+8. A spammer may attempt to work around this MSC's effects by joining and leaving the room during
+   their spam. This has relatively high cost (the impact of spam is lesser when they aren't joined
+   to the room, and re-joining will hit a more restrictive rate limit on most servers).
+
+   Moderation bots and similar community safety tools are encouraged to add restrictions to the number
+   of join+leave cycles a user may perform in a short window. This will further reduce the effectiveness
+   of such an attack. Issue 7 above also applies here.
```
anoadragon453, line 248: How does one effectively limit the number of joins a given user can perform? If the spammer is on their own homeserver, then there are no CS API rate limits.

What's to stop a user joining a room, waiting 24hrs to clear any join rate-limits, then spamming and sending another join? Relying on banning a user before they finish spamming and sending a state event seems very hopeful, especially when spam is typically reported via bystanders.

Why not just redact all of a user's event history since the creation of the room?

anoadragon453: @mscbot concern The proposal doesn't prevent users from spamming and sending an `m.room.member` event to protect the spam event.

turt2live: Originally the MSC shied away from doing so for performance concerns, but it feels increasingly valuable to say "anything before the ban", especially to simplify the implementations.

Thoughts from more folks would be helpful for steering the MSC in any particular direction.

proposals/4293-redact-on-ban.md
turt2live: considering a duplicate of https://github.com/matrix-org/matrix-spec-proposals/pull/4293#discussion_r2334164734 for the sake of discussion.

proposals/4293-redact-on-ban.md
tulir: My client implementation is at least just "if current state says redact_events, hide all events sent by user". I'm assuming that if a spammer is unbanned, the events have either been redacted explicitly or weren't actually spam and can be shown again.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+by a server to their local database will redact whatever events the server has already seen, or will
+see in the future.
+
+Though this MSC isn't breaking for clients in the same way that MSC2244 is, clients (and to a degree,
+servers) may not support the MSC right away. This means that senders may still have to send "fallback
+redactions" to ensure the room is cleaned up, though those senders can begin to treat those redactions
+as best effort (currently, moderation bots in particular work incredibly hard to ensure they get *every*
+event redacted, but can run into delivery, reliability, and completeness issues with 1:1 redactions).
+
+It's also important to note that this proposal intends to compliment mass redactions and coexist in
```
uhoreg: Nitpick:

```suggestion
It's also important to note that this proposal intends to complement mass redactions and coexist in
```

proposals/4293-redact-on-ban.md
uhoreg: Couldn't a malicious user also get around the redactions by constantly sending displayname changes or no-op `m.room.membership` events?  I'd be in favour of changing it to "anything before the ban".  It seems to me like the most common case is that if someone gets banned and their messages redacted, it's likely that we'd prefer to err on the side of redacting too much, rather than redacting too little.

proposals/4293-redact-on-ban.md
```diff
@@ -0,0 +1,283 @@
...
+considered "best effort" by the sender as they may encounter delivery issues, especially when using
+1:1 redactions instead of mass redactions. "Best effort" might mean using endpoints like
+[MSC4194](https://github.com/matrix-org/matrix-spec-proposals/pull/4194)'s batch redaction, or using
+the less reliable `/messages` endpoint to locate target event IDs to redact. Senders SHOULD note that
+this fallback behaviour will only target events they can see (or be made to see via MSC4194) and might
+not be the same events that a receiving client or server sees.
+
+Senders are encouraged to evaluate when they can cease sending fallback redactions like those described
+above to minimize the event traffic involved in a ban. For moderation bots this may mean waiting for
+sufficiently high *client* implementations existing in their communities.
```
Gnuxie, line 197: How does a moderation bot provide fallback if the server it is using implements this MSC?

turt2live: The server's support is a lot less important than the client support I think. The general fallback behaviour is for the bot to do what it normally does alongside setting the flag, though individual projects can (and should) make their own decisions.

turt2live:
This needs work before it's ready to resume FCP discussions.

@mscbot fcp cancel

PR #4264, opened 2025-02-11
MSCXXXX: Tokens for Contacting Accounts

thomasfiala:
---
name: WIP Proposal
about: A proposal that isn't quite ready for formal review yet.
title: '[WIP] Your Proposal Title'
labels: proposal
assignees: ''

---

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/master/CONTRIBUTING.md#sign-off)
* [ ] Update the title and file name of your proposal to match this PR's number (after opening).
* [ ] Pull request includes a ['Rendered' link](https://matrix.org/docs/spec/proposals#process) above.
* [ ] Have a look at the [MSC Checklist](MSC_CHECKLIST.md) for guidelines on various aspects of your MSC.

PR #4294, opened 2025-05-29
Add ignore invites proposal.

grinapo:
Adding ignore-invites proposal.

### DIFF
--- a/proposals/0666-ignore-invites.md
+++ b/proposals/0666-ignore-invites.md
@@ -1,0 +1,64 @@
+# MSC0666: Ignore and mass ignore invites
+
+
+## Proposal
+
+
+### Introduction 
+
+Invite spam is getting more problematic and clients have no good way to handle massive invite spamming without server support.
+
+**Ignoring** users seem to help this but it has two serious problems:
+
+- completely igoring users make them completely disappear for the user, and if the user is also a moderator/adminstrator it makes impossible to detect and handle further abuse
+- servers are not _required_ to clean up [invites] after ignored users (by the spec).
+
+**Rejecting ignores** also creates problem for clients:
+
+- Invite rejections are _room based_, so when a user is invited to massive amount of rooms by `spammer` then rejections require numerous `leave` API calls
+- API calls are rate limited so massive rejections may be impossible without seriously disturbing user communication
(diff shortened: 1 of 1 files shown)


grinapo:
There is a long history of ~~ignored~~ open ignore invites proposals. There is also a [_summary_ MSC](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/ignoring-invites-meta/proposals/4192-ignoring-invites.md) to list all the MSCs about the topic.
- https://github.com/matrix-org/matrix-spec-proposals/pull/2270
- https://github.com/matrix-org/matrix-spec-proposals/pull/3840
- https://github.com/matrix-org/matrix-spec-proposals/pull/4155
- https://github.com/matrix-org/matrix-spec-proposals/pull/3847
- https://github.com/matrix-org/matrix-spec-proposals/pull/3659
- https://github.com/matrix-org/matrix-spec-proposals/pull/4264

All of these proposals are open and have been for more than 5 years. They are fairly complex, and handle various complex combinations of ignores and allowlists, for users, servers, and patterns, as well as some preauthorisation.

This current proposal is **deliberately very simple**, using the already established ignore machanism, which should be easy to implement using parts of the same code. It is hoped that it will not take years to be accepted, since the problem is acute.

cc: @Johennes

proposals/4294-ignore-invites.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

grinapo:
> @turt2live 
> Please consider switching the branch to a feature branch instead of `master`.

_(I hate git)_ 
What magical commands shall be casted upon the repo which would move it onto a feature branch while keeping this PR intact?

turt2live:
The internet is telling me it's not possible to change the source branch after the PR is opened, but I'm pretty sure we've done that before. For all the hassle, a new PR is probably easiest and causes no harm.

richvdh:
> Overall, I feel this MSC is in direct competition with #4155 which has a number of implementations and experimentation. If there's parts that 4155 doesn't cover, this MSC would be best positioned to propose extensions to 4155 so that it may protect more users.

Agreed, and I don't see any counterarguments forthcoming.

In the interest of reducing the number of active MSCs in this field:

@mscbot fcp close

PR #3955, opened 2023-01-13
MSC3955: Extensible Events - Automated event mixin (notices)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/notices/proposals/3955-extensible-events-bot-notice-mixin.md)

PR #4295, opened 2025-05-31
MSC4295: Bot bounce limit - a better loop prevention mechanism

m13253:
[Rendered](https://github.com/m13253/matrix-spec-proposals/blob/bot-bounce-limit/proposals/4295-bot-bounce-limit.md)

(**About me:** I develop E2EE-capable Matrix bots and bridges tailored for two communities. Recently, I open-sourced my [matrixbot-ezlogin](https://github.com/m13253/matrixbot-ezlogin) Rust library to help people build Matrix bots without worrying about the authentication and E2EE bootstrap process.)


proposals/4295-bot-bounce-limit.md
```diff
@@ -0,0 +1,235 @@
...
+3. For a room purposed for technical support, the operator can run an AI-powered bot to automatically answer common questions. Such AI bot is allowed to trigger other bots for certain helpful tasks.
+4. The room operator can run a "UTD notification bot" that notifies room members that their messages can't be decrypted by others. However, it is very important to prevent it from replying another bot's message.
+5. When bridging rooms across three or more platforms (e.g., Matrix ⇌ Telegram ⇌ IRC ⇌ Matrix), it is necessary to make sure each bridge doesn't pick up another bridge's messages.
+6. Bridges supporting double-puppeting needs to ignore messages sent by a reverse puppet. Although they already employ proprietary methods (e.g., vendor-prefixed tags like `fi.mau.double_puppet_source` or a list of ignored user IDs), it could be very useful to provide a standardized loop-preventing mechanism, allowing bridges from different vendors to work in harmony at the same room.
```
tulir, line 14: This proposal doesn't appear to solve the problem that `fi.mau.double_puppet_source` was made for. Specifically, when a bridge sends a message from a double puppet (not a bridge ghost user), it must have some flag to prevent echoing back the message to the remote network where it came from. The flag is not meant to stop any other bridge or bot from reacting to the message, it's only meant to be detected by the origin bridge.

proposals/4295-bot-bounce-limit.md
```diff
@@ -0,0 +1,235 @@
...
+
+There are three existing solutions to prevent bot loops:
+
+1. The `m.notice` message type:
+
+   By specification, `m.notice` messages must never be automatically responded to.
+
+   However, there are a few disadvantages of `m.notice`:
+
+   1. It is analogous to `m.text`, which doesn't support attached files or encrypted images.
```
tulir, line 42: Extensible events already have a solution for this https://github.com/matrix-org/matrix-spec-proposals/pull/3955

proposals/4295-bot-bounce-limit.md
```diff
@@ -0,0 +1,235 @@
...
+   2. It is designed for automated messages, not bridged messages sent originally by a human.
+   3. Similarly, `m.notice` won't be picked up by bridges to forward to a bridged platform.
```
tulir, line 44: Bridges can and do pick up `m.notice` if configured to do so

m13253: > Bridges can and do pick up `m.notice` if configured to do so

Thanks for your confirmation!

Indeed, I just checked the Matrix API Spec: The spec doesn’t say whether bridges should or shouldn’t pick up `m.notice`.

I was careless. I will rephrase this sentence.

proposals/4295-bot-bounce-limit.md
m13253: > This proposal doesn't appear to solve the problem that `fi.mau.double_puppet_source` was made for. Specifically, when a bridge sends a message from a double puppet (not a bridge ghost user), it must have some flag to prevent echoing back the message to the remote network where it came from. The flag is not meant to stop any other bridge or bot from reacting to the message, it's only meant to be detected by the origin bridge.

It’s an honor to get a feedback from Mautrix’s side!

After understanding your explanations, I admit this proposal doesn’t solve `fi.mau.double_puppet_source`’s problem. In fact, this proposal seems to solve **a completely different problem** orthogonal to `fi.mau.double_puppet_source`’s problem.

1. Although both mechanisms are designed to prevent infinite loop, `fi.mau.double_puppet_source` prevents looping between two platforms, `m.bounce_limit` prevents looping within Matrix.
2. `fi.mau.double_puppet_source` doesn’t prevent the message being interacted by other bots or bridges, while `m.bounce_limit` sets a bounce limit to do so.
3. On the other hand, `fi.mau.double_puppet_source` doesn't deal with the situation where a room has two independent bridge instances -- e.g., one relaybot maintained by the room operator, and one double-puppet maintained by a room member on a separate homeserver --, while `m.bounce_limit` tries to solve this problem.

(Please correct me if my understanding is still inaccurate.)

Probably I will need to rephrase or remove this sentence. I think `m.bounce_limit` won’t replace `fi.mau.double_puppet_source`. They will co-exist, because **two mechanisms solve two different problems**.

m13253: Update: I replaced this example in the Background section with another example.

proposals/4295-bot-bounce-limit.md
```diff
@@ -0,0 +1,242 @@
...
+
+A valid `m.bounce_limit` value can be in either of the following forms:
+
+1. Missing.
+2. The number 1.
+3. An integer between and including 2 and 2^53-1.
+
+These are invalid forms, and their normalization rules upon receiving:
+
+1. The number 0, which should be treated as missing. (This design is to simplify the development of bots in certain programming languages, such as Go.)
```
TheArcaneBrony: This doesn't make semantic sense. Logically a value of 0 would mean "do not forward".

m13253: > This doesn't make semantic sense. Logically a value of 0 would mean "do not forward".

Thanks for your comment.

Here are two explanations of this decision:

1. This is an analogy of Hop Limit in IP networks.

   An IP packet with Hop Limit of 1 means the packet is able to transmit to the recipient, but if the recipient is a router, it isn’t allowed to forward the packet to anywhere else.
   An IP packet with Hop Limit of 0, if I remember correctly, is invalid.
   Therefore, a developer who has experience with IP networking might be able to feel the current design more familiar to them, than making 0 a valid value.

2. If we make 0 an invalid value, some programming languages need some fewer steps to distinguish 0 and a missing value.

   One example of such programming languages, is Go.
   To distinguish a 0 value and a missing value, the Go struct needs to be written as:
   ```go
   type OriginalRoomMessageEventContent struct {
       Body        string `json:"body"`
       BounceLimit *int64 `json:"m.bounce_limit"` // *int64 instead of int64
       ...
   }
   ```
   which uses one layer of pointer to distinguish 0 and missing, meaning slower performance (although negligible), more memory fragments, and more work on the developer side to get the logic right.
   C++ may be similar — depending on which JSON library you use.
   In other programming languages that supports nullable data types, such as Rust’s `Option`, C#’s `Nullable`, or TypeScript’s `T | undefined`, at least one more check is required to distinguish the missing value and to extract the valid numbers out.

Therefore, making 0 an invalid value is just simpler, faster to develop and run, and more similar to other existing network protocols.

m13253: By the way, I have another question: Regarding the existing Matrix protocol, if any field is invalid, how should an implementation treat it?

Should an implementation treat it as 0, missing, or reject the message at all? Perhaps this new proposal needs to be consistent in this perspective…

m13253:
> This doesn't make semantic sense. Logically a value of 0 would mean "do not forward".

After I evaluated @TheArcaneBrony’s comment carefully again and talked with other people who have worked with the `ruma` library, along with my experience in trying to implement this logic, I now agree with @TheArcaneBrony.

I plan to change the logic **from “1 = stop” to “0 = stop.”** I will change the proposal Markdown file in a couple of days. But before I make the change, I want to announce here to make sure no one else is simultaneously working on an implementation.

P.S.: I am currently working on a Rust implementation. First, a proof of concept echo-bot is coming soon, then I will extend to other languages than Rust and incorporate it to some real-world bots.

proposals/4295-bot-bounce-limit.md
TheArcaneBrony: Sorry, seems your replies got lost in my mailbox...

> Regarding the existing Matrix protocol, if any field is invalid, how should an implementation treat it?

depends on the context, generally if this is event auth, then the event should be rejected, whereas for regular client-server operation, the spec explicitly mentions that event contents are untrusted.

In response to your top level message from 9 hours ago, I feel like logically, a missing TTL on event forwarding should be treated the same as having no hops left (if you wanted to stick to the IPv4 analogy), but also one could frame it as how any person would interpret it too: there's still a ticket left to pass on, so we should pass it on. hence, missing a value outright would logically be the same as treating it as zero. (Also see null pointers in C/C++ usually being a litteral 0x00000000 pointer)

m13253:
A new version has been pushed to the branch. The biggest modification is the behavioral change from “1 = stop” to “0 = stop.”

I also defined clear rules when dealing with multiple input messages rather than leaving it as an open question. I also removed the description about Telegram’s bot-to-bot restrictions, which is now outdated.

I added some more discussion about whether `m.bounce_limit` lying outside encryption envelope does or doesn’t affect security. I feel **I might be playing with fire**, so I really want to hear more ideas from other people.
(Or, should I just be safe, putting it inside the encryption envelope, and abandon my “UTD bot” idea?)

A working PoC called echo-bot is available at: https://codeberg.org/m13253/matrix-echo-bot or https://github.com/m13253/matrix-echo-bot (They are mirrors of each other.)
Please compile with `--features unstable-msc4295` to use bounce limit to prevent looping; with `--features allow-looping` to allow looping; with no feature flags to use `m.notice` to prevent looping.
Please add two instances of echo-bot into the same room, and use a human account to send an initial message as a seed. You will be able to observe their looping or loop prevention behavior.
I am currently working to implement it on real-world bots other than this simple echo-bot.

Finally, since I have been iterating on the branch for multiple commits, I am keeping all the historical commits so we don’t lose track of the comments tagged to those commits. Please instruct me whether and when and how I should squash the git history.

PR #4296, opened 2025-06-04
MSCXXXX: Mentions for device IDs

Johennes:


proposals/4296-device-id-mentions.md
turt2live, line 1: Implementation requirements:
* [ ] Client (sending)
* [ ] Client (receiving)
* [ ] Server

proposals/4296-device-id-mentions.md
```diff
@@ -0,0 +1,47 @@
+# MSC4296: Mentions for device IDs
+
+The [`m.mentions`] content property on events allows senders to include an indicator about whether
+the event should trigger a specialised notification for some or all room members. Among others,
+this can be used to direct events at a certain recipient group within the room. It is, however,
+not currently possible to direct events at a certain subset of devices within a room. This can
+be helpful, for instance, in cases where devices with different capabilities are participating
+in the room and the sender wants to provide a hint on which devices the recipient should pick up
+the message. This proposal makes it possible to mention specific devices via the existing
```
clokep, line 9: Like what? What features would this support?

proposals/4296-device-id-mentions.md
```diff
@@ -0,0 +1,47 @@
...
+[`m.mentions`] mechanism.
+
+## Proposal
+
+A new optional property `device_ids` is introduced in [`m.mentions`] to allow specifying an array
```
clokep: I think device IDs need to be namespaces by user?

proposals/4296-device-id-mentions.md
```diff
@@ -0,0 +1,47 @@
...
+"m.mentions": {
+  "device_ids": ["ABC1234"]
+}
+```
+
+It is legal for `room` and `user_ids` to be present within `m.mentions` simultaneously to `device_ids`.
+
+## Potential issues
+
+None.
```
Half-Shot: Would this be something more for bots to be using, rather than users? I'm trying to understand the client-side flow of how you'd mention a specific device. It certainly feels a bit odd for anything other than perhaps a security message from your server, to me.

proposals/4296-device-id-mentions.md
Johennes: I'm planning to use this as a dependency for another proposal that I'm currently working out. I think it won't get much clearer before that, sorry (and it could possibly still be a bad idea after all).

Maybe I should have delayed publishing this proposal but it seemed atomic enough and I wanted to get feedback on it as early as possible. Your other comment below shows me that this wasn't entirely futile at least.

proposals/4296-device-id-mentions.md
Johennes: Oh, I think you're right. So that would mean either you have to provide the MXID in `users` (and only one MXID) simultaneously with `device_ids` or `device_ids` would have to be a map of MXID to devices IDs.

proposals/4296-device-id-mentions.md
Johennes: Yes, in fact I'm planning to use this as a dependency for another proposal that involves bots. Sorry, this isn't clearer right now.

proposals/4296-device-id-mentions.md
Half-Shot: I'd personally do:

```json
{
  "m.mentions": {
    "device_ids": {
      "@alice:example.org": ["ABC1234"]
    }
  }
}
```

Half-Shot: Although that comes with the drawback of no fallbacks for unsupported clients, which probably depends on the usecase you are trying to solve.

Johennes: Yeah, you're right. I've made it a map now and recorded the lack of a fallback in the potential issues section.

proposals/4296-device-id-mentions.md
clokep: No problem, I'm just unable to come up with plausible ideas myself! Looking forward to seeing it.

Johennes: I was planning to use this in connection with [MSC4300](https://github.com/matrix-org/matrix-spec-proposals/pull/4300) and [MSC4301](https://github.com/matrix-org/matrix-spec-proposals/pull/4301) for device-to-device communication but it looks like the use case I had in mind may not materialize. I'll move this back into draft for the moment.

PR #4300, opened 2025-06-16
MSC4300: Processing status requests & responses

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/request-response/proposals/4300-request-response.md)

### DIFF
--- a/proposals/4300-request-response.md
+++ b/proposals/4300-request-response.md
@@ -19,6 +19,8 @@ clients when sending events. It has the following properties in `content`:
 
 - `from_device` (required, string): The sending device's device ID. Allows recipients to optionally
   submit their responses privately via to-device messages in the future.
+- `to_device`(optional, string): The receiving device's device ID. Should be set when the sender
+  wants to a adress a specific receiving device only.
 - `lifetime` (integer): The duration in milliseconds during which the sender will consider responses
   to this request. Prevents meaningless delayed responses when new or previously disconnected
   devices encounter requests on older events.
@@ -40,14 +42,13 @@ Clients MAY add `m.request.status` as a top-level property in `content` on any e
 }
 ```
 
-Clients that receive events containing `m.request.status` content blocks MAY respond to them with a
-new room event type `m.response.status`. The latter contains an `m.reference` relation pointing to
-the original event as well as an `m.response.status` content block with the following properties:
+Clients that receive events containing `m.request.status` content blocks where `to_device` is either
+missing or identical to their own device ID, MAY respond to them with a new room event type
+`m.response.status`. The latter contains an `m.reference` relation pointing to the original event as
+well as an `m.response.status` content block with the following properties:
 
 - `from_device` (required, string): The sending device's device ID. Helps clients identify the
   remote echo of their own responses.
-- `to_device`(optional, string): The receiving device's device ID. Should be set in case the sender
-  want to a adress a specific receiving client.
 - `status` (required, string, one of `success`, `error`): Whether the sending device has understood
   and successfully processed the event.
 - `messages` (array): An optional array of messages to help recipients understand the `status`.
--- a/proposals/4300-request-response.md
+++ b/proposals/4300-request-response.md
@@ -32,6 +32,7 @@ Clients MAY add `m.request.status` as a top-level property in `content` on any e
   "content": {
     "m.request.status": {
       "from_device": "RJYKSTBOIE",
+      "to_device": "EIOBTSKYJR", // optional the receiving device
       "lifetime": 90000, // 90s
     },
     // properties specific to m.pizza
@@ -45,6 +46,8 @@ the original event as well as an `m.response.status` content block with the foll
 
 - `from_device` (required, string): The sending device's device ID. Helps clients identify the
   remote echo of their own responses.
+- `to_device`(optional, string): The receiving device's device ID. Should be set in case the sender
+  want to a adress a specific receiving client.
 - `status` (required, string, one of `success`, `error`): Whether the sending device has understood
   and successfully processed the event.
 - `messages` (array): An optional array of messages to help recipients understand the `status`.
--- a/proposals/4300-request-response.md
+++ b/proposals/4300-request-response.md
@@ -125,9 +125,9 @@ scenarios that require auditability.
 
 ## Unstable prefix
 
-While this MSC is not considered stable, `m.request.status` and `m.response.status` (the event type)
-should be referred to as `de.gematik.msc4300.request.status` and
-`de.gematik.msc4300.response.status`, respectively.
+While this MSC is not considered stable, `m.request.status` and `m.response.status` should be
+referred to as `de.gematik.msc4300.request.status` and `de.gematik.msc4300.response.status`,
+respectively.
 
 ## Dependencies
 
--- a/proposals/4300-request-response.md
+++ b/proposals/4300-request-response.md
@@ -82,6 +82,10 @@ themselves and ignore any new `m.response` events received from other clients.
 
 ## Potential issues
 
+The mechanism introduced above doesn't enable clients to determine if other clients are able to
+understand an event without actually sending it. [MSC4301] builds upon the present proposal and
+introduces event capability queries to achieve this.
+
 This proposal doesn't strictly define what constitutes successful processing of an event. At a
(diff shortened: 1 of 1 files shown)


proposals/4300-request-response.md
turt2live, line 1: Implementation requirements:
* [ ] Sending client
* [ ] Receiving client (with support)
* [ ] Receiving client (without support)

turt2live, line 1: High level review: This sounds like [delivery receipts](https://github.com/matrix-org/matrix-spec-proposals/pull/4089) where the concept is meant to be extended to "actually processed by the client for the user".

Johennes: Thanks, I wasn't aware of this. Have included it in the alternatives as I think there's some overlap but cater to slightly different problems.

proposals/4300-request-response.md
```diff
@@ -0,0 +1,137 @@
...
+
+These problems are aggravated when one or both of the participating clients are an automated system
+(a.k.a. bot) which cannot easily issue or respond to generic "Did you see my message?" questions.
+
+The present proposal partially addresses this problem by defining a generic scheme for clients to
+request and receive an explicit processing status for events from other clients.
+
+## Proposal
+
+A new content block `m.request.status` is introduced to request a processing status from other
```
benkuly, line 17: Shouldn't it be `m.status.request` and `m.status.response` like the `m.key.verification.*` events?

benkuly, line 17: Do we really need a request for the described use cases? Couldn't the receiver proactively just decide to send a status, when it does not support event types instead of letting the sender always request the status?

Johennes: Hm, good question. My assumption was that there might be further specialized request types in future – as in requests that ask for something else than the processing status. Therefore, I made `request` the namespace. I have no concrete examples in mind, however, and don't feel strongly about the naming either.

Johennes: I think that could be helpful but it seems to solve a slightly different problem.

My assumption was that the sender would only add status requests on events that, for one reason or another, are important. Things such as prescriptions or sick notices where the content is not just plain text and where as the sender I need to make sure that you have actually understood and processed the event correctly. That means the majority of events would not include status requests which made adding it on the sender side appear acceptable.

Only sending the status response when an event was _not_ understood would mean that the only signal of successful processing is the absence of a response. But then you won't know if the recipient didn't respond because it understood the event or if it simply hasn't received and processed the event yet.

proposals/4300-request-response.md
```diff
@@ -0,0 +1,137 @@
...
+    "m.request.status": {
+      "from_device": "RJYKSTBOIE",
+      "lifetime": 90000, // 90s
+    },
+    // properties specific to m.pizza
+  }
+}
+```
+
+Clients that receive events containing `m.request.status` content blocks MAY respond to them with a
```
denis-loh: To understand the flow correctly:

1. client A wants to know, if the room members understand my message "m.pizza" by sending the pizza.
2. client B may or may not respond with a room event stating if the message was supported

My concern with this approach is, that it potentially causes unnecessary responses and or transfers data, which may not be needed if the counterpart does not understand it.

So to keep the metaphor: if I am sending the pizza without knowing if it will be eaten, I am likely wasting lots of food over the time. However, if I use m.request.status as its own room event like a knock-on event to ask if someone wants pizza, I first of all can reduce the initial overhead and can then also cover those clients, who did not respond that they implicitly don't want pizza.

I understand that the tradeoff of that is that the sending client must wait for the clients to confirm they understand that message first, but keeping the lifetime of the request much lower (e.g. 5 - 10 seconds) you can confirm two things: 1. the client supports the message and 2. if there's no response from a client, he either doesn't support it or isn't interested (e.g. polling or questionnaire messages). The latter is especially interesting, when no-one in the room actually understands the message: if no client supports it, there is no need to send the data at all, which is likely more data privacy compliant. More specifically, if a client responds to a request immediately, it may conflict with the users privacy settings, if he doesn't want to send read receipts (https://spec.matrix.org/v1.16/client-server-api/#receipts).

Also if client A already send a request and got a specific response from the other clients, he may cache the response if to is solely being used for identification of message type support.

Johennes: This proposal is meant to enable clients to understand if something they have sent was understood by the recipients. You are correct in that, on its own, this doesn't provide clients a way to determine if something will be understood by others before actually sending it. This part is covered in [MSC4301](https://github.com/matrix-org/matrix-spec-proposals/pull/4301). That proposal defines an ahead-of-time capability query which itself uses the request status mechanism from this proposal here.

denis-loh: But in that case, MSC4301 and MSC4300 are directly depending on each-other, which is not immediately clear to the readers. Without the MSC4301, I still have privacy concerns as data might get transferred without acknowledgement and it's to some extend unclear how it mitigates the issue "did you understood my message?" if the other client does not implement this feature. The response event will only be sent, if the client is online, which comes with two downsides:

1. if the client is offline, the sender will not get notified, which could mean, that either the client does not support the MSC or that the recipient hasn't read the message yet.
2. if the client is online, the recipient may still not support the message or doesn't want to propagate any "online" state.

So, in neither of these cases, the sending client will ever receive the requested response, which in turn might lead to the question again "have you understood my message?"

Johennes: > But in that case, [MSC4301](https://github.com/matrix-org/matrix-doc/pull/4301) and [MSC4300](https://github.com/matrix-org/matrix-doc/pull/4300) are directly depending on each-other, which is not immediately clear to the readers.

I have tried to make the relationship between these proposals clearer with cda5d2cbeabd3c9cc27d784bd8529328241cc376.

> Without the [MSC4301](https://github.com/matrix-org/matrix-doc/pull/4301), I still have privacy concerns as data might get transferred without acknowledgement [...]

Sending the `m.response.status` event is a MAY under this proposal. So clients and users with privacy concerns are free to not send it, only send it after explicit user confirmation or send it automatically but with some random jitter time[^1].

[^1]: In the gematik fork of the Matrix spec, we're planning to make responses [mandatory](https://gemspec.gematik.de/prereleases/Draft_gemF_TI-M_Strukturierte_Daten/gemF_TI-M_Strukturierte_Daten_V1.0.0_CC/). You're right in that this leaks metadata. I think that's a comment for https://github.com/gemSpecComment/Draft_gemF_TI-M_Strukturierte_Daten/issues rather than here though because that requirement doesn't exist in this proposal.

> [...] it's to some extend unclear how it mitigates the issue "did you understood my message?" if the other client does not implement this feature. The response event will only be sent, if the client is online, which comes with two downsides:
> 
>     1. if the client is offline, the sender will not get notified, which could mean, that either the client does not support the MSC or that the recipient hasn't read the message yet.
> 
>     2. if the client is online, the recipient may still not support the message or doesn't want to propagate any "online" state.
> 
> 
> So, in neither of these cases, the sending client will ever receive the requested response, which in turn might lead to the question again "have you understood my message?"

Yes, the mechanism in this proposal is best-effort only and won't work perfectly in all situations. A specific case where we expect it to work well is when two automated systems exchange data via Matrix. The online state should not be a concern in this situation.

I'm happy to hear about alternatives though if you can think of any.

proposals/4300-request-response.md
```diff
@@ -0,0 +1,146 @@
...
+    "m.relates_to": {
+      "event_id": "$1",
+      "rel_type": "m.reference",
+    },
+    // optional properties specific to m.pizza
+  }
+}
+```
+
+Clients can check whether a request has expired using `lifetime`, `origin_server_ts` and their local
```
marcusfey, line 83: Can I just ask: Why specify `lifetime` and not `valid_until_ts` (or something similar)?

This would make the process a bit more transparent on both sides. The sender has to calculate this timestamp anyway to check incoming responses.

Sidenote: My organizations show DieTechniker, that's due to historical collaboration.

Johennes: > Can I just ask: Why specify `lifetime` and not `valid_until_ts` (or something similar)?

Mostly because `lifetime` had some precedent in the spec with [VoIP](https://spec.matrix.org/v1.18/client-server-api/#mcallinvite) (from where this proposal essentially borrows).

PR #4301, opened 2025-06-16
MSC4301: Event capability negotiation between clients

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/event-capability-negotiation/proposals/4301-event-capability-negotiation.md)

### DIFF
--- a/proposals/4301-event-capability-negotiation.md
+++ b/proposals/4301-event-capability-negotiation.md
@@ -163,7 +163,7 @@ The concerns and remedies around leaking metadata from [MSC4300] apply to this p
 
 ## Unstable prefix
 
-While this MSC is not considered stable, `m.request.event_capability` (the event type) and
+While this MSC is not considered stable, `m.request.event_capability` and
 `m.response.event_capability` should be referred to as `de.gematik.msc4301.request.event_capability`
 and `de.gematik.msc4301.response.event_capability`, respectively. Properties inherited from
 [MSC4300] have their own prefixing requirements.


proposals/4301-event-capability-negotiation.md
turt2live, line 1: Implementation requirements:

* [ ] Sending client
* [ ] Receiving client (with support)
* [ ] Receiving client (without support)

proposals/4301-event-capability-negotiation.md
```diff
@@ -0,0 +1,115 @@
+# MSC4301: Event capability negotiation between clients
+
+Matrix allows clients to exchange both built-in and custom events with other clients in rooms. There
+is, however, no way for a client to understand what types of events the other clients in a room are
+able to understand. This is problematic as a compatibility mismatch means that the recipient user
+might only be able to see a fallback representation of an event or, in the worst case, nothing at
+all. At the same time, the sender is left wholly unaware of the recipient's experience.
```
turt2live: This problem area is discussed *a lot*, usually in the context of extensible events. I'm concerned that this MSC opens up pandora's box again by trying to add a negotiation layer on top of a compatibility layer - this proposal needs a strong rationale for *why* extensible events is insufficient beyond "it can sometimes result in fallback", which is by design.

Johennes: > this proposal needs a strong rationale for why extensible events is insufficient beyond "it can sometimes result in fallback", which is by design.

This is really it though.

We're using Matrix in the healthcare sector and in some cases it's simply not acceptable to rely on fallback representations. The events to be sent might contain complex medical data that cannot feasibly be condensed into a useful fallback and where the sender requires guarantees that the content will actually be understood because health depends on it.

benkuly: Imho a combination of extensible events and `m.status.response` would be sufficient. We could add an information like `allow_fallback: Boolean` or don't add a fallback at all in the extensible event. Then the receiver could answer with an `m.status.response` so the sender would get notified about displaying issues.

benkuly: For many cases a fallback would be sufficient. E. g. a DICOM event could have an image and file fallback, so the user would get a preview (image fallback) and could download the DICOM (file fallback) and open it in an external software.

Johennes: Those are good points. I suppose in essence the only thing this proposal adds over https://github.com/matrix-org/matrix-spec-proposals/pull/4300 is the shortcut of figuring out if an event type will be understood before actually sending the event.

Not having to send the event if the recipient is known to not support it is probably only a meaningful benefit if you're sending extremely large data.

However, being able to downgrade to another event type that the recipient understands feels harder to do with just https://github.com/matrix-org/matrix-spec-proposals/pull/4300. What I'm trying to hedge against are cases where there are multiple historic versions of an event type (similar to how some of the email message types [here](https://fachportal.gematik.de/toolkit/dienstkennung-kim-kom-le) had several versions).

Johennes: Have amended the motivation further to point at the FHIR use case. FHIR has a myriad of resources and profiles, both versioned, that can affect display and processing logic. It's hardly feasible to build fallback and compatibility layers in this case because the only apparent fallback is a raw JSON or XML file.

Johennes:
This has been superseded by https://github.com/matrix-org/matrix-spec-proposals/pull/4401. gematik are not currently pursuing the approach in this proposal anymore.

PR #4302, opened 2025-06-20
MSC4302: Exchanging FHIR resources via Matrix events

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/fhir-resources/proposals/4302-fhir-resources.md)

### DIFF
--- a/proposals/4302-fhir-resources.md
+++ b/proposals/4302-fhir-resources.md
@@ -101,7 +101,7 @@ new event type `m.fhir` is introduced. This type mandates the following properti
 - `m.text` (object, optional): Alternative textual information that can be displayed in case the
   client is not able to render the FHIR content.
 - `m.file` (object, optional): A downloadable alternative for the content of the FHIR structure for
-  example an editable PDF file represanting the same information
+  example an editable PDF file representing the same information.
 
 ``` json5
 {


proposals/4302-🔥-resources.md
turt2live, line 1: Implementation requirements:

* [ ] Demonstration of use

turt2live, line 1: Looks like github handles the emoji poorly - please stick to ASCII in file names.

Johennes: Sorry, didn't expect this to cause problems. 😕

proposals/4302-fhir-resources.md
```diff
@@ -0,0 +1,83 @@
+# MSC4302: Exchanging FHIR resources via Matrix events
+
+[FHIR] (pronounced 🔥) is a globally established standard for exchanging healthcare information
+electronically. The base building block of FHIR are so called resources, such as [`Patient`]. These
+resources can be serialised into JSON or XML which allows them to be transmitted via the [`m.file`]
+message type. This is insufficient, however, because a generic MIME type of `application/xml` or
+`application/json` doesn't let clients recognise the content as FHIR before downloading it. The
```
clokep: According to their [website](https://www.hl7.org/fhir/json.html) the mimetype should be `application/fhir+json` / `application/fhir+xml`? Is that not enough to know the format?

Johennes: You're right. I think my explanation could use improvement because I'm only covering this in the alternatives section. What I'm trying to provide with this proposal is details about what resource type is contained in the file and the possibility to inline the JSON where the size limit allows it.

Johennes: Have rearranged the text and slightly extended it. Hope it's a little clearer now.

denis-loh:
I am not sure, if it is a good idea to introduce a deep coupling to a specialized feature into a public standard. I know that there are also other use cases where custom room events are used (for instance in German BOS communication systems). If they would also contribute their custom types, wouldn't it dilute the matrix standard too much in future?

This MSC would conflict with the generalization (matrix)/ specialization (fhir) concept of the specification.

fnwbr:
When describing the alternative the risk of “leaking metadata to the home server” is being mentioned.

How does the proposed solution mitigate that risk?

proposals/4302-fhir-resources.md
```diff
@@ -0,0 +1,185 @@
...
+
+[RFC 2045] allows MIME types to include modifying parameters. The contents of
+`m.fhir.structure_definition` could, therefore, be included alongside the media type[^1].
+
+``` http
+Content-type: application/fhir+json; url="https://gematik.de/fhir/isik/StructureDefinition/ISiKFormularDefinition"; ...
+```
+
+This would allow reusing the `m.file` message type but leaks metadata to the home server in
+[`POST /_matrix/media/v3/upload`].
```
Johennes, line 174: (Copying the comment here because we're encouraged to keep discussions in threads on the diff so that they can more easily be split-tracked.)

@fnwbr [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4302#issuecomment-3783576657):

> When describing the alternative the risk of “leaking metadata to the home server” is being mentioned.
>
> How does the proposed solution mitigate that risk?

Johennes: The proposed solution encapsulates all information related to the FHIR resource in a room event. In encrypted rooms, this event will be end-to-end encrypted. So unless the event is used in public rooms, there should be no additional metadata leakage introduced by this proposal.

proposals/4302-fhir-resources.md
Johennes, line 1: (Copying the comment here because we're encouraged to keep discussions in threads on the diff so that they can more easily be split-tracked.)

@denis-loh [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4302#issuecomment-3640688212):

> I am not sure, if it is a good idea to introduce a deep coupling to a specialized feature into a public standard. I know that there are also other use cases where custom room events are used (for instance in German BOS communication systems). If they would also contribute their custom types, wouldn't it dilute the matrix standard too much in future?
>
> This MSC would conflict with the generalization (matrix)/ specialization (fhir) concept of the specification.

It's a fair question whether or not the use case is large enough to warrant inclusion into Matrix.

That being said, I don't think this proposal introduces a "deep coupling" between Matrix and FHIR. It's already possible to exchange FHIR via Matrix today. All this proposal does is add a handful of metadata properties to facility that. This doesn't feel much different from how [URL previews](https://spec.matrix.org/v1.16/client-server-api/#get_matrixclientv1mediapreview_url) include properties from the [Open Graph](https://ogp.me/) standard just that Open Graph is probably more widely known.

PR #4305, opened 2025-07-04
MSC4305: Pushed Authorization Requests (PARs) for OAuth authentication

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/par/proposals/4305-par.md)


proposals/4305-par.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Server

Johennes: I suspect we'll need an implementation in MAS, too?

PR #4306, opened 2025-07-08
MSC4306: Thread Subscriptions

reivilibre:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rei/msc_thread_subscriptions/proposals/4306-thread-subscriptions.md)

### DIFF
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -178,7 +178,7 @@ As motivation, we want threads to have the following notification semantics:
 
 To achieve this, we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between `content` and `room` but which can contain general-purpose rules:
 
-1. `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+1. `.m.rule.unsubscribed_thread`: This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
    The rule occurs after mention-specific rules and keyword mention rules, meaning that mentions continue to generate notifications.
    ```jsonc
    {
@@ -194,7 +194,7 @@ To achieve this, we propose the addition of two new push rules, both added to a 
        "actions": []
    }
    ```
-2. `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
+2. `.m.rule.subscribed_thread`: This rule causes events in subscribed threads to generate notifications.
    ```jsonc
    {
        "rule_id": ".m.rule.subscribed_thread",
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -120,7 +120,7 @@ In either case, servers must ensure that an automatic subscription cannot keep g
 ### New Client Behaviour: subscribe when sending in a thread
 
 When a user is sending a message in a thread, the user's client SHOULD simultaneously perform a manual subscription to that thread using `PUT /subscription`.
-The subscription is not considered automamtic, because it was initiated by a user action (sending a message). Further, it'd be best for latency to perform the subscription concurrently with sending the message, so the client doesn't yet have an event ID to attach a hypothetical automatic subscription to anyway.
+The subscription is not considered automatic, because it was initiated by a user action (sending a message). Further, it'd be best for latency to perform the subscription concurrently with sending the message, as the client doesn't yet have an event ID to attach a hypothetical automatic subscription to anyway.
 
 Clients MUST only perform these subscriptions for messages sent by the same client and at the time of the messages being sent. In other words, client MUST NOT apply them for existing messages or messages sent by other devices.
 
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -247,6 +247,7 @@ We need a new push rule kind ordered between `content` and `room` because:
   the ones underlying 'Mentions only' or 'Mentions & Keywords' settings available in common clients.
 
 For the time being, servers MUST ensure that clients MUST NOT be able to create custom rules with this push rule kind.
+If a client attempts to do this, a 400 / `M_INVALID_PARAM` error response MUST be produced by the server.
 
 ## Limitations and Potential Future Expansions
 
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -176,9 +176,9 @@ As motivation, we want threads to have the following notification semantics:
     - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
 - Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
 
-To achieve this, we propose the addition of two new push rules:
+To achieve this, we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between `content` and `room` but can contain general-purpose rules:
 
-1. an `underride` push rule, called `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+1. `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
    The rule occurs after mention-specific rules and keyword mention rules, meaning that mentions continue to generate notifications.
    ```jsonc
    {
@@ -194,7 +194,7 @@ To achieve this, we propose the addition of two new push rules:
        "actions": []
    }
    ```
-2. an `underride` push rule, called `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
+2. `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
    ```jsonc
    {
        "rule_id": ".m.rule.subscribed_thread",
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -186,7 +186,14 @@ The `thread_subscription` push condition is satisfied if and only if all the fol
 
 ## Unstable prefix
 
-TODO!
+Whilst this proposal is unstable, the following changes should be noted:
+
+- the endpoint is renamed to `/_matrix/client/unstable/io.element.msc4306.thread_subscriptions`
+- the push rules' IDs are renamed to:
+  - `.io.element.msc4306.rule.unsubscribed_thread`
+  - `.io.element.msc4306.rule.subscribed_thread`
+- the push rule condition `kind` is renamed to:
+  - `io.element.msc4306.thread_subscription`
 
 
 ## Dependencies
@@ -197,4 +204,4 @@ TODO!
 
 This proposal is known to be depended upon by the following MSCs:
 
-- <a name="ref-ss-ext"></a>\<proto-MSC for Thread Subscriptions extension to Sliding Sync\>
+- [MSC4308: Thread Subscriptions extension to Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/blob/rei/msc_ssext_threadsubs/proposals/4308-sliding-sync-ext-thread-subscriptions.md)
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -160,7 +160,7 @@ As motivation, we want threads to have the following notification semantics:
 
 To achieve this, we propose the addition of two new push rules:
 
-1. a `content` push rule, called `.m.rule.unsubscribed_thread`, at the end of the content rule list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+1. an `underride` push rule, called `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
    The rule occurs after mention-specific rules and keyword mention rules, meaning that mentions continue to generate notifications.
    ```jsonc
    {
@@ -176,7 +176,7 @@ To achieve this, we propose the addition of two new push rules:
        "actions": []
    }
    ```
-2. an `underride` push rule, called `.m.rule.subscribed_thread`, at the beginning of the underride list. This rule causes events in subscribed threads to generate notifications.
+2. an `underride` push rule, called `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
    ```jsonc
    {
        "rule_id": ".m.rule.subscribed_thread",
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -21,6 +21,13 @@ This proposal consists of three main parts:
 - a new push rule (including a new push rule condition) that prevents notifying about threads that the user has not subscribed to.
 
 
+### Terminology
+
+This proposal introduces the concept of being 'subscribed' to and 'unsubscribed' from a thread.
+
+The default state is unsubscribed.
+
+
 ### New Endpoints
 
 #### Subscribing to a thread
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -176,7 +176,7 @@ As motivation, we want threads to have the following notification semantics:
     - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
 - Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
 
-To achieve this, we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between `content` and `room` but can contain general-purpose rules:
+To achieve this, we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between `content` and `room` but which can contain general-purpose rules:
 
 1. `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
    The rule occurs after mention-specific rules and keyword mention rules, meaning that mentions continue to generate notifications.
@@ -220,6 +220,33 @@ These push rules use a new push condition `thread_subscription`, which takes an 
 The `thread_subscription` push condition is satisfied if and only if all the following hold:
 1. the event, which we are running push rules for, is part of a thread.
 2. the user is subscribed to the thread (`subscribed` = `true`) or is not subscribed to the thread (`subscribed` = `false`).
+
+### Explanation of the rationale behind the `postcontent` push rule kind
+
+We need a new push rule kind ordered between `content` and `room` because:
+
+**For `.m.rule.unsubscribed_thread`**
+- this rule must come after any keyword mentions the user has configured (which would be configured in `content`)
+  as those should still trigger mentions (and thus automatic thread subscriptions).
+- the rule must come before any room settings the user has configured, such as notifying on all messages in a room,
+  as that would cause automatic subscriptions to every thread in the room and it would cause inconsistencies
+  between the default 'notified for all messages' behaviour and 'notified for all messages in this room'.
+
+**For `.m.rule.subscribed_thread`**
+- it seems inappropriate for an `override` rule to produce notifications, given the `override` rules generally
+  are intended to suppress them.
+  - more concretely, we should allow rules in the `override` and `content` block the opportunity to produce
+    notifications with specific tweaks, e.g. `content` rules to match mentions might highlight the messages
+    rather than simply notifying.
+      - `.m.rule.contains_user_name` and user-specified keyword mention rules must therefore execute first
+        so they can make use of `"set_tweak": "highlight"`.
+- this rule should come after `content` in case any users perform negative filtering (i.e. suppressing
+  notifications for messages containing certain keywords), although the author is not aware of clients with
+  such functionality.
+- this rule must come before `room` to avoid thread subscriptions being defeated by user rules such as
+  the ones underlying 'Mentions only' or 'Mentions & Keywords' settings available in common clients.
+
+For the time being, servers MUST ensure that clients MUST NOT be able to create custom rules with this push rule kind.
 
 ## Limitations and Potential Future Expansions
 
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -120,9 +120,8 @@ As motivation, we want threads to have the following notification semantics:
 
 To achieve this, we propose the addition of two new push rules:
 
-1. an `override` push rule, called `.m.rule.unsubscribed_thread`, at the end of the override list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
-   The rule occurs after mention-specific rules, meaning that mentions continue to generate notifications.
-   **TODO how do we deal with keyword mentions**
+1. a `content` push rule, called `.m.rule.unsubscribed_thread`, at the end of the content rule list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+   The rule occurs after mention-specific rules and keyword mention rules, meaning that mentions continue to generate notifications.
    ```jsonc
    {
        "rule_id": ".m.rule.unsubscribed_thread",
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -137,6 +137,16 @@ If the client is already aware of the user being subscribed to the thread, then 
 
 If the mentioning user is banned in the room or ignored by the user, the automatic thread subscription should not occur.
 
+#### What counts as a mention
+
+The following things count as a mention for the sake of 'subscribe on mention' behaviour:
+
+- `@room` — both intentional and fallback;
+- intentional user mentions;
+- fallback plaintext mentions when intentional mentions are not in use; and
+- keyword mentions, if the user has configured some.
+
+
 #### Reversal of automatic subscriptions
 
 *This section is an optional aspect. Clients can choose not to implement it, particularly in scenarios where it doesn't make sense.*
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -55,7 +55,7 @@ In the general case, returns 200 with empty body `{}`.
 
 If an automatic subscription was requested but the provided cause event ID is not within the thread, returns 400/`M_NOT_IN_THREAD`.
 
-If an automatic subscription was requested but the automatic subscription is skipped by the server, returns 409 (Conflict) with `M_SKIPPED`. See the section below on *'Server-side automatic subscription ordering'*.
+If an automatic subscription was requested but the automatic subscription is skipped by the server, returns 409 (Conflict) with `M_CONFLICTING_UNSUBSCRIPTION`. See the section below on *'Server-side automatic subscription ordering'*.
 
 If the thread does not exist, or if the user does not have access to it, returns 404/`M_NOT_FOUND`.
 
@@ -252,6 +252,9 @@ Whilst this proposal is unstable, the following changes should be noted:
   - `.io.element.msc4306.rule.subscribed_thread`
 - the push rule condition `kind` is renamed to:
   - `io.element.msc4306.thread_subscription`
+- the error codes are renamed to:
+  - `IO.ELEMENT.MSC4306.M_CONFLICTING_UNSUBSCRIPTION`
+  - `IO.ELEMENT.MSC4306.M_NOT_IN_THREAD`
 
 For servers implementing this MSC, the `org.matrix.msc4306` feature should be advertised through `/_matrix/clients/versions`.
 
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -1,0 +1,200 @@
+MSC4306: Thread Subscriptions
+=====
+
+## Background and Summary
+
+Threads were introduced in [version 1.4 of the Matrix Specification](https://spec.matrix.org/v1.13/changelog/v1.4/) as a way to isolate conversations in a room, making it easier for users to track specific conversations that they care about (and ignore those that they do not).
+
+Thus far, there has been no good way for users to ensure that they see new messages in only the threads that they care about. The current rules only allow expressing one of two choices: users and their clients can either watch all threads, or ignore all threads.
+As a result, the user is either being presented with unwanted information (misusing their attention), or they risk missing important messages.
+
+This proposal introduces a new mechanism, Thread Subscriptions, as a first-class way for users to signal which threads they care about.
+
+We then add a new push rule so that users can receive push notifications only for threads that they care about.
+
+
+## Proposal
+
+This proposal consists of three main parts:
+- 3 simple endpoints for subscribing to and unsubscribing from threads;
+- new prescribed client behaviour when the user is mentioned in a thread; and
+- a new push rule (including a new push rule condition) that prevents notifying about threads that the user has not subscribed to.
+
+
+### New Endpoints
+
+#### Subscribing to a thread
+
+```
+PUT /_matrix/client/v1/rooms/{roomId}/thread/{eventId}/subscription
+```
+
+with request body:
+
+```jsonc
+{
+  // Whether the subscription was made automatically
+  // by a client, not by manual user choice.
+  "automatic": true
+}
+```
+
+Returns 200 with empty body `{}`.
+
+If the thread does not exist, or if the user does not have access to it, returns 404/`M_NOT_FOUND`.
+
+If the thread is already subscribed, then the subscription remains and:
+
+- if `automatic` is `false`, the thread is marked as manually-subscribed, even if the existing subscription has it marked as automatically subscribed. (In other words, manual subscriptions take precedence over automatic subscriptions.)
+
+
+#### Unsubscribing from a thread
+```
+DELETE /_matrix/client/v1/rooms/{roomId}/thread/{eventId}/subscription
+```
+
+Returns 200 with empty body `{}`.
+
+If the thread was not subscribed, returns 200 with empty body `{}` for idempotence.
+
+If the thread does not exist, or is inaccessible, returns 404/`M_NOT_FOUND`.
+
+
+#### Retrieving thread subscription
+
+```
+GET /_matrix/client/v1/rooms/{roomId}/thread/{eventId}/subscription
+
+{"automatic": true}
+```
+
+Retrieves the thread subscription for a given thread.
+
+On success (200), returns:
+
+```jsonc
+{
+  // Whether the subscription was automatically made or not
+  "automatic": true
+}
+```
+
+If there is no subscription to that thread, or the thread does not exist, or the thread is inaccessible, returns 404/`M_NOT_FOUND`.
+
+
+### New Client Behaviour: subscribe on mention
+
+When a user is mentioned in a thread (by another user — the *mentioning user*), the user's client should perform an automatic subscription to that thread using `PUT /subscription` with `{"automatic": true}`.
+
+The server does not perform this action on the client's behalf, principally because the server is not able to detect mentions in encrypted rooms.
+
+If the client is already aware of the user being subscribed to the thread, then making a `PUT /subscription` request is not necessary.
+
+If the mentioning user is banned, the automatic thread subscription should not occur.
+
+#### Reversal of automatic subscriptions
+
+If an automatic thread subscription occurs and the mentioning user is subsequently banned, then:
+
+- the thread subscription should be reversed,
+- provided that there aren't any other mentions by other, non-banned, users that would have caused the same automatic subscription.
+
+When a client becomes aware of a banned user in a room, it may need to backpaginate thread history to determine whether there are any threads whose automatic subscriptions should be reversed, or to determine if there are any other qualifying mentions that would obviate the need to reverse automatic subscriptions.
+
+For efficiency reasons, clients may limit the depth of this backpagination with an implementation-defined recency limit, owing to the observation that abuse is usually cleaned up shortly after it occurs.
+
+### New Push Rules
+
+As motivation, we want threads to have the following notification semantics:
+
+- Messages in unsubscribed threads should not count as activity at all; as a user, I do not want to see the room as unread because there are new messages in an unsubscribed thread.
+    - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
+- Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
+
+To achieve this, we propose the addition of two new push rules:
+
+1. an `override` push rule, called `.m.rule.unsubscribed_thread`, at the end of the override list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+   The rule occurs after mention-specific rules, meaning that mentions continue to generate notifications.
+   **TODO how do we deal with keyword mentions**
+   ```jsonc
+   {
+       "rule_id": ".m.rule.unsubscribed_thread",
+       "default": true,
+       "enabled": true,
+       "conditions": [
+           {
+               "kind": "thread_subscription",
+               "subscribed": false
+           }
+       ],
+       "actions": []
+   }
+   ```
+2. an `underride` push rule, called `.m.rule.subscribed_thread`, at the beginning of the underride list. This rule causes events in subscribed threads to generate notifications.
+   ```jsonc
+   {
+       "rule_id": ".m.rule.subscribed_thread",
+       "default": true,
+       "enabled": true,
+       "conditions": [
+           {
+               "kind": "thread_subscription",
+               "subscribed": true
+           }
+       ],
+       "actions": [
+           "notify",
+           {
+               "set_tweak": "sound",
+               "value": "default"
+           }
+       ]
+   }
+   ```
+
+These push rules use a new push condition `thread_subscription`, which takes an argument `subscribed` (boolean).
+The `thread_subscription` push condition is satisfied if and only if all the following hold:
+1. the event, which we are running push rules for, is part of a thread.
+2. the user is subscribed to the thread (`subscribed` = `true`) or is not subscribed to the thread (`subscribed` = `false`).
+
+## Limitations
+
+- Users will not have enough granularity to subscribe to threads in a way that lets them keep track of threads (being able to 'catch up' through some mechanism in their client) without also getting notifications for them, except by disabling ALL thread subscription notifications altogether.
+  - There is precedent for this granularity in the popular forum software *Discourse*, but the author is not aware of Instant Messaging software with this granularity.
+  - With that said, this could be feasibly extended by a later MSC with no apparent issues.
+
+## Alternatives
+
+- Clients could maintain thread subscription settings in Room Account Data, as a map from event ID to the subscription settings.
+    - This would be inefficient by requiring the entire subscription set for an entire room to be transferred at once, for example in Sliding Sync.
+    - This would make subscriptions vulnerable to Read-Modify-Write race conditions (though this could be addressed with extensions to the Room Account Data APIs).
+- Clients could maintain their thread subscription settings in their global Account Data, but this seems to be strictly worse than doing so in Room Account Data.
+
+## Security considerations
+
+- Abuse of 'subscribe on mention', particularly in public rooms
+    - Malicious users can be ignored by the user, to stop the subscribe on mention behaviour.
+    - Users can disable notifications entirely for rooms.
+    - Room moderators can take action against malicious users abusing the feature.
+    - If no room moderators are protecting the user, the user can of course also leave the room.
+
+## Notes to Client Implementors
+
+## Notes to Server Implementors
+
+- Since clients will be automatically updating the Thread Subscription Settings when their user is mentioned, server implementations should be ready to handle concurrent updates by multiple of the user's devices/clients when they are online at the same time.
+
+## Unstable prefix
+
+TODO!
+
+
+## Dependencies
+
+- no dependencies on other pending MSCs
+
+## Dependents
+
+This proposal is known to be depended upon by the following MSCs:
+
+- <a name="ref-ss-ext"></a>\<proto-MSC for Thread Subscriptions extension to Sliding Sync\>
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -188,7 +188,7 @@ The `thread_subscription` push condition is satisfied if and only if all the fol
 
 Whilst this proposal is unstable, the following changes should be noted:
 
-- the endpoint is renamed to `/_matrix/client/unstable/io.element.msc4306.thread_subscriptions`
+- the endpoint is renamed to `/_matrix/client/unstable/io.element.msc4306/rooms/{roomId}/thread/{eventId}/subscription`
 - the push rules' IDs are renamed to:
   - `.io.element.msc4306.rule.unsubscribed_thread`
   - `.io.element.msc4306.rule.subscribed_thread`
--- a/proposals/4306-thread-subscriptions.md
+++ b/proposals/4306-thread-subscriptions.md
@@ -40,13 +40,21 @@ with request body:
 
 ```jsonc
 {
-  // Whether the subscription was made automatically
-  // by a client, not by manual user choice.
-  "automatic": true
+  // If the subscription was made automatically
+  // by a client, this is the ID of the latest event that caused
+  // the automatic subscription.
+  //
+  // If the subscription was made manually,
+  // this field is omitted.
+  "automatic": "$abcdef"
 }
 ```
 
-Returns 200 with empty body `{}`.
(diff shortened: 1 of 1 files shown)


proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,200 @@
...
+
+- Users will not have enough granularity to subscribe to threads in a way that lets them keep track of threads (being able to 'catch up' through some mechanism in their client) without also getting notifications for them, except by disabling ALL thread subscription notifications altogether.
+  - There is precedent for this granularity in the popular forum software *Discourse*, but the author is not aware of Instant Messaging software with this granularity.
+  - With that said, this could be feasibly extended by a later MSC with no apparent issues.
+
+## Alternatives
+
+- Clients could maintain thread subscription settings in Room Account Data, as a map from event ID to the subscription settings.
+    - This would be inefficient by requiring the entire subscription set for an entire room to be transferred at once, for example in Sliding Sync.
+    - This would make subscriptions vulnerable to Read-Modify-Write race conditions (though this could be addressed with extensions to the Room Account Data APIs).
```
clokep, line 264: My hope was you could use (room) account data, but that doesn't have a "state key" or anything. I think it also limits you in size, which probably wouldn't work. 😢

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,200 @@
...
+
+### New Client Behaviour: subscribe on mention
+
+When a user is mentioned in a thread (by another user — the *mentioning user*), the user's client should perform an automatic subscription to that thread using `PUT /subscription` with `{"automatic": true}`.
+
+The server does not perform this action on the client's behalf, principally because the server is not able to detect mentions in encrypted rooms.
+
+If the client is already aware of the user being subscribed to the thread, then making a `PUT /subscription` request is not necessary.
+
+If the mentioning user is banned, the automatic thread subscription should not occur.
```
Johennes: Does "banned" mean the sender is in the recipient's `m.ignored_user_list`?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,200 @@
...
+#### Reversal of automatic subscriptions
+
+If an automatic thread subscription occurs and the mentioning user is subsequently banned, then:
+
+- the thread subscription should be reversed,
+- provided that there aren't any other mentions by other, non-banned, users that would have caused the same automatic subscription.
+
+When a client becomes aware of a banned user in a room, it may need to backpaginate thread history to determine whether there are any threads whose automatic subscriptions should be reversed, or to determine if there are any other qualifying mentions that would obviate the need to reverse automatic subscriptions.
+
+For efficiency reasons, clients may limit the depth of this backpagination with an implementation-defined recency limit, owing to the observation that abuse is usually cleaned up shortly after it occurs.
```
Johennes, line 169: This sounds like it might either be expensive or leave dangling subscriptions. If the server exposed an API to list all thread subscriptions in the room, clients could use `/relations` to only walk the subscribed threads which might be more efficient?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,200 @@
...
+
+### New Push Rules
+
+As motivation, we want threads to have the following notification semantics:
+
+- Messages in unsubscribed threads should not count as activity at all; as a user, I do not want to see the room as unread because there are new messages in an unsubscribed thread.
+    - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
+- Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
```
Johennes, line 177: An API [listing active subscriptions](https://github.com/matrix-org/matrix-spec-proposals/pull/4306/files#r2196734363) per room might also make unsubscribing easier?

proposals/4306-thread-subscriptions.md
Johennes: I was also thinking of account data and I believe it isn't subject to the 64KiB limit?

What were you planning to use a state key for with thread subscriptions?

proposals/4306-thread-subscriptions.md
reivilibre: The idea would be that through MSC4308 (or another mechanism) the client will already know its active subscriptions, so indeed it can just walk the (recently- and automatically-)subscribed threads.

However it's hard to give a general definition of 'time' to describe how much the client should consider, so it's left purposefully flexible here.

proposals/4306-thread-subscriptions.md
reivilibre: I meant banned from the room, but the mentioning user getting ignored by the mentioned user also sounds like a good case to apply this!

proposals/4306-thread-subscriptions.md
reivilibre: Using account data in this way may not have the 64 kiB limit, but ideally we should try to avoid using it in more places that will make it hard to add that limit. (Since the lack of a limit is a mistake, I believe.)

The two problems I noted above are the main reasons I don't like using account data here:

- The read-modify-write issue
- Clients and servers having to operate on whole lists when they only care about one tiny portion of it.

I think Patrick's 'state key' mention is along the lines of: 'we only have a single key for each piece of room account data: the type'.
You could imagine using having an extra key, so room account data could be keyed on (`m.thread.subscription`, `$abc123def456`).

Or we could stretch things further to have 'Event Account Data' or something.

Johennes: > I think Patrick's 'state key' mention is along the lines of: 'we only have a single key for each piece of room account data: the type'.
> You could imagine using having an extra key, so room account data could be keyed on (`m.thread.subscription`, `$abc123def456`).

Oh, I see.

clokep: > Since the lack of a limit is a mistake, I believe.

Correct.

And your example of a state keys is spot on for what I was thinking.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,200 @@
...
+1. the event, which we are running push rules for, is part of a thread.
+2. the user is subscribed to the thread (`subscribed` = `true`) or is not subscribed to the thread (`subscribed` = `false`).
+
+## Limitations
+
+- Users will not have enough granularity to subscribe to threads in a way that lets them keep track of threads (being able to 'catch up' through some mechanism in their client) without also getting notifications for them, except by disabling ALL thread subscription notifications altogether.
+  - There is precedent for this granularity in the popular forum software *Discourse*, but the author is not aware of Instant Messaging software with this granularity.
+  - With that said, this could be feasibly extended by a later MSC with no apparent issues.
+
+## Alternatives
```
clokep, line 260: I'm kind of surprised there's no changes to the `/threads` API?

proposals/4306-thread-subscriptions.md
giomfo: I would make optional this Reversal of automatic subscriptions. This add complexity client side for a low gain.

This is relevant to not subscribe if the mentioning user has been banned or ignored before the client takes into account the mention to set up the subscription.

I would let the user reverse manually the automatically subscribed thread if any

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,207 @@
...
+
+For efficiency reasons, clients may limit the depth of this backpagination with an implementation-defined recency limit, owing to the observation that abuse is usually cleaned up shortly after it occurs.
+
+### New Push Rules
+
+As motivation, we want threads to have the following notification semantics:
+
+- Messages in unsubscribed threads should not count as activity at all; as a user, I do not want to see the room as unread because there are new messages in an unsubscribed thread.
+    - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
+- Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
```
giomfo, line 177: When a user mutes a room, it puts an override rule in for that room. And this override rule is the highest priority

So this is not possible to receive notification from a thread in a muted room.
The manual subscription should be disabled in this room at the UI level until the end user unmutes the room.
When a room is muted, the client should pursue the automatic thread subscriptions. These subscriptions will be ignored until the room is unmuted Keep managing the automatic subscription eases the potential switch mute/unmute of a room. The client UI should let the end user know any potential thread subscription is disabled in a muted room

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Proposal
+
+This proposal consists of three main parts:
+- 3 simple endpoints for subscribing to and unsubscribing from threads;
+- new prescribed client behaviour when the user is mentioned in a thread; and
+- a new push rule (including a new push rule condition) that prevents notifying about threads that the user has not subscribed to.
+
+
+### New Endpoints
```
MadLittleMods, line 32: Why isn't there an endpoint to paginate all thread subscriptions in a room? (my expectation)

Similar to how clients back-paginate with messages in a room by starting with `/sync` and then using `/messages` to get the rest as necessary.

proposals/4306-thread-subscriptions.md
clokep: I wonder if expanding this to "relation subscriptions" would be helpful for any other use-cases. So a subscription is parent event+relation type?

Might not be super useful with the current relations though.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,207 @@
...
+  "automatic": true
+}
+```
+
+If there is no subscription to that thread, or the thread does not exist, or the thread is inaccessible, returns 404/`M_NOT_FOUND`.
+
+
+### New Client Behaviour: subscribe on mention
+
+When a user is mentioned in a thread (by another user — the *mentioning user*), the user's client should perform an automatic subscription to that thread using `PUT /subscription` with `{"automatic": true}`.
```
bnjbvr: Ok, I think some interactions can lead to "stuck" subscriptions now :upside_down_face: 

Imagine the following scenario.

- On device 1, I have a thread containing events E1, E2 (which mentions me), and E3.
- Because of E2, my client subscribed me automatically to the thread.
- Now, I decide that I don't want to be subscribed to the thread, so I manually unsubscribe from it.
- I open device 2, and back-paginate the thread there as well.

I think that, on device 2, because there's no data related to the thread subscription anymore, when device 2 will see E2 for the first time, it will automatically re-subscribe me to it.

Am I missing something that would prevent this stuck subscription behavior?

Ok, if I'm not mistaken, then. One feature that we'd like to allow, is re-subscription after a thread has been unsubscribed. That is, in my example, if E4 mentions me again, my client should be able to subscribe me again.

I thought we could use some kind of conceptual enum with 3 variants:

- subscribed
- unsubscribed
- (nothing)

Unsubscribing would lead to the `unsubscribed` state, instead of deleting the `subscribed` state and replacing it with nothing. Then, we'd need a rule like "if unsubscribed, don't subscribe me again".

But… that doesn't work either :melting_face: When E4 comes, which mentions me, my client might want to re-subscribe me to the thread. If we have a rule like "if unsubscribed, don't ever subscribe me ever again", then I can't be re-subscribed. The real information is "I'm unsubscribed until E3 included"; all events prior to it should not lead to a new automatic subscription, and all events _after_ it should automatically resubscribe me. As a matter of fact, I think including the event id in the `unsubscribe` variant should be sufficient…

…Now, if you're paying attention, I've said "events" and "after" in the same sentence. I'm talking about event ordering here, hence I'm afraid we might run into the "stuck notification" problem, because the ordering may be different if events have been received from sync ("stream" ordering) or from /relations (topological ordering). Here are a few ideas, how to solve it:

- maybe the "unsubscribe" event id could be… two event ids at most: one in topological ordering, one in sync ordering, and it's the client which determines which is last, in its local state. In this case, the server should accept multiple values at the same time, for the "unsubscribe" endpoint. That's an idea that was mentioned to solve the stuck read receipts dilemma in the past by @MadLittleMods if I recall correctly, so it could apply here too.
- maybe we should enforce one ordering over the other, and say that this one ordering is the correct one. In that case, it must be the topological ordering, since the sync ordering doesn't match from one device to another. I'm afraid this would lead to the stuck notifications issue, though, unless clients agreed to only use /relations for filling the content of a thread.
- MSC4033 gives some stable-ish event ordering, and it could be used for determining the order of subscriptions.
- We could actually use the user's own read receipts as a proxy of "unsubscribed until this event": only event *after* the read receipt could cause an automatic re-subscription. This is messy though, as it implies a processing ordering dependency in clients: the client would need to process read receipts *after* it's processed thread subscriptions.

MSC4033 would honestly be the most satisfying, but it's been stuck for a while, there's been valid objections in it, etc. I would be curious to hear other people's ideas or thoughts about the problem. Am I missing something here?

giomfo: I agree with the need to introduce a new variant `unsubscribed` to have the possibility to differentiate `undefined` with `unsubscribed`.

Then to keep thread subscriptions as simple as possible. I would suggest to disable automatic subscription in a thread which has been manually unsubscribed. I will receive the potential new mentions in this thread but this will not subscribe me again because I clearly unsubscribed myself from this thread

giomfo: Note: I would manage differently manual and automatic unsubscription (if automatic is planned)
- a manual unsubscription -> unsubscribed state (via a PUT)
- an automatic unsubscription -> undefined state (via a DELETE)

proposals/4306-thread-subscriptions.md
reivilibre: This would be pretty shoddy for performance if it was per-room (= 1 request needed per room), but maybe a global endpoint would be OK.

It just felt like sliding sync would do the job of getting the state down to the client without much fuss (like for room account data), but maybe that's wrong

bnjbvr: > It just felt like sliding sync would do the job of getting the state down to the client without much fuss (like for room account data), but maybe that's wrong

We'd need to increase the equivalent of "timeline_limit" to simulate back-pagination, and that complicates a few things:

- if we're using the same SSS connection, then we're penalizing other updates, because we might change the connection parameters again and again until we've "paginated" all the receipts of all rooms
- …or we have to use a different SSS connection just for that, but it sounds a bit overkill :smiling_face_with_tear: 

So agreed with @MadLittleMods that having the same "N last + pagination token if needs be" would be ideal.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,207 @@
...
+This proposal introduces a new mechanism, Thread Subscriptions, as a first-class way for users to signal which threads they care about.
+
+We then add a new push rule so that users can receive push notifications only for threads that they care about.
+
+
+## Proposal
+
+This proposal consists of three main parts:
+- 3 simple endpoints for subscribing to and unsubscribing from threads;
+- new prescribed client behaviour when the user is mentioned in a thread; and
```
andybalaam, line 21: Should we also subscribe when a user sends a message within a thread?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,207 @@
...
+=====
+
+## Background and Summary
+
+Threads were introduced in [version 1.4 of the Matrix Specification](https://spec.matrix.org/v1.13/changelog/v1.4/) as a way to isolate conversations in a room, making it easier for users to track specific conversations that they care about (and ignore those that they do not).
+
+Thus far, there has been no good way for users to ensure that they see new messages in only the threads that they care about. The current rules only allow expressing one of two choices: users and their clients can either watch all threads, or ignore all threads.
+As a result, the user is either being presented with unwanted information (misusing their attention), or they risk missing important messages.
+
+This proposal introduces a new mechanism, Thread Subscriptions, as a first-class way for users to signal which threads they care about.
```
andybalaam, line 11: My first expectation at how this should work was with 3 states for threads:

* Subscribed (either automatic or manual)
* Unsubscribed
* New - the user has not expressed a preference yet

This would allow clients to highlight "New" threads somewhere in the UI, offering them to the user so they can make a conscious choice about whether to subscribe.

As I understand it, this proposal assumes all threads default to Unsubscribed state (right?).

Maybe a product person could comment on whether we might want the option of showing "New" threads? (Or maybe I missed this discussion)

mxandreas: This question has been ask a few times. Shortly we do not want "new" threads, because there does not seem to be much purpose for that. It also does not exist in other apps that have threads, at least I am not aware of. You can see on the timeline if a thread has been started and if it is too much in the past, someone will mention it to you it it is important.

Johennes: This sounds sensible but I wonder if it needs to be an overarching product decision for all users and Matrix clients. In my experience, having missed important conversations that happened in a thread because one didn't notice the start of the thread is one of the main complaints I heard from people about threads on other platforms over the years.

dkasak: +1. Not only on other platforms but on Matrix (via Element) as well. I can attest to this being the case for my own use case.

proposals/4306-thread-subscriptions.md
clokep: Yes IMO

proposals/4306-thread-subscriptions.md
mxandreas: > having missed important conversations that happened in a thread because one didn't notice the start of the thread is one of the main complaints I heard from people about threads on other platforms over the years

The urgency and priority of this depends heavily on the context and on the use case.

1. The impact of this vs. the impact of not getting notifications when something happens in the thread you're subscribed to, is very different. Threads are close to useless if you can't subscribe and get notifications for replies.

2. In a workplace setting, people usually aren't reading all conversations proactively all the time. They want to focus on what's important - and usually there is somebody else who will mention them in the thread if it is important for them.

3. If threads actually work (you get notifications for what you're subscribed to) people start to use threads more, and the main timeline of the room gets way shorter - meaning that you're also more likely to notice if a thread has started somewhere.

Finally, GH, Jira, etc. all work in the same way - by default you're not subscribed to the issues, PRs, etc.  You can start watching manually or when somebody mentions or you comment. Not saying that this isn't a feature at all - but I think it is a nice to have and for specific use cases. And it does make notifications system overall more complex.

dkasak: In this envisioned system, do I have a way to list all threads, so that I don't have to rely on noticing them in the main timeline? I think having this is essential.

> Finally, GH, Jira, etc. all work in the same way - by default you're not subscribed to the issues, PRs, etc. 

At least in Github, you can still subscribe to all activity in a repository. However, I'd argue a Github repository has a larger scope than a Matrix room, at least in my mind. I would much more often want to be notified of all threads in a Matrix room than of all issues and PRs in a Github repo; this behaviour I would reserve mostly for repositories I own or maintain.

Given that here we are discussing a protocol mechanism rather than what a concrete implementation of this will look in a client, I really think we need to retain a bit more flexibility and allow for clients that might want to inform the user of all new threads.

Johennes: > Given that here we are discussing a protocol mechanism rather than what a concrete implementation of this will look in a client, I really think we need to retain a bit more flexibility and allow for clients that might want to inform the user of all new threads.

Yes, this is exactly what I was trying to hint at. What @mxandreas described sounds perfectly valid for a specific product. But I can equally see the use case of wanting to be notified about all threads. So ideally the protocol should enable both.

bnjbvr: While I completely agree the protocol should make it possible to be notified of new threads in a room / across rooms, as it would make it much easier to implement something akin to Element Web's "thread activity center", I would like people to note this is not the goal of the current MSC, as described in the first paragraph of the proposal.

For what it's worth, it's already possible to approximate getting a list of threads across all rooms, with extra effort that increases the server's load, since it wasn't really intended to support this use case:

- when receiving an event from sync, one can look at either the bundled thread summary in an event (indicating it's a thread root) or the `m.relates_to` property (if it's of type thread, then the thread root is also included in there). That's one way to notice new threads, across all rooms.
- it's also possible, in a given room, to use `/room/ROOM_ID/threads` to paginate all threads, compare that against a known list to know when to keep paginating or stop, and infer which threads are "new" in the room.
- repeating the above for all rooms would go heavy against the server (so I don't recommend implementing it like this), but if done in a round-robin way, it could give a list of all the new threads across all rooms.

It might be possible to use a plain `main`-thread read receipt to compute which threads are new or not, as well.

I would encourage people to write another MSC for this. We've talked about having some kind of new sliding sync extension, that would reuse something like the `timeline_limit` mechanism to give a list of the latest N threads across all rooms, and if needs be, a pagination token to be used in another new API endpoint, that would list all threads across all rooms. That's only one rough idea for implementing the feature, so there might be better suited mechanisms. But overall, it's not really the topic of this MSC, which is about "being notified of new messages in threads that one is interested in".

andybalaam: @bnjbvr I don't think @Johennes was talking about all threads across all rooms - just about all threads within one room.

bnjbvr: Independently of whether it's about getting new threads for a single room, or across rooms, I'd hold onto my stance with respect to not scope-creeping this MSC, and I would encourage folks to open other MSC(s) for being aware of new threads in a room (which is definitely something worth having in the protocol!).

proposals/4306-thread-subscriptions.md
bnjbvr, line 177: For what it's worth, I thought there was a little chicken-and-egg problem here, in the case of a new automatic subscription, but turns out there might not be.

Say I receive an event that would automatically subscribe me to a thread I haven't subscribed to (nor unsubscribed from). At this point, the thread subscription push rule wouldn't be triggered, because the subscription doesn't exist quite yet. But I'd expect the client to still notify about the thing! However, if an automatic subscription were to happen, that's likely because *other* push rules decided so: I've been mentioned in the thread, or there was a keyword mention, or something else. So there *should* be some notification created for that other event, which then causes the automatic subscription to the thread.

A notification will exist for the initial event that causes an automatic subscription; it is just not created by the new push rule yet. Future events in this thread will imply notifications because of the thread subscription push rule.

Am I getting this right? If so, it may be worth mentioning this around here, to help client implementations?

proposals/4306-thread-subscriptions.md
bnjbvr: Another precision question: should the wording say "when a user is mentioned" (in which case we're only interested in intentional mentions), or should the wording say "whenever a notification would be generated by push rules" (which then also includes keyword mentions, and would nicely generalize to any trigger of current and future push rules, making it future compatible as well)?

bnjbvr: Hmm, maybe not such a bright idea: in this case, in a room where I'd get a notification for any message, I would be automatically subscribed to every thread (since every thread message is a message, and would trigger a notification).

It would be nice to clarify the intent here:
- should we auto-subscribe any time a notification would be generated (per the push rules), as hinted in my previous comment?
- if not, what's a mention? (keyword mention? username mention in clear text? intentional mention in metadata?)

reivilibre: this was intended to be very explicitly mentions, but you're bang on that we should carefully say what we mean here.

Here's my thoughts:

- `@room` — both intentional and fallback
- intentional user mentions
- fallback plaintext mentions
- keyword mentions

proposals/4306-thread-subscriptions.md
reivilibre: I think this thread boils down to 2 things, tell me if I'm wrong:

- I might want to receive notifications for all threads
    - There's nothing stopping you in this proposal from editing the push rules (e.g. disabling the MSC4306 push rules and adding a custom one to push for every threaded event) to do this
    - I think this MSC doesn't harm this use case compared to today (but admittedly also doesn't do anything in support of it compared to today)

- I might want to *see* all threads if I hunt them down
    - I think the existing `/room/.../threads` and a future `/global_threads` endpoints would be the answer here. In fact, it's not really clear why subscriptions get involved at  all here — this becomes a UI problem for a totally different mechanism/endpoint (some of which already exist).

- Subscriptions should be tri-state — purposefully-subscribed, purposefully-unsubscribed and ambient
    - This is fair, an early draft of this MSC also did start out this way, but it got a bit unwieldy and there were UX concerns about it being overcomplicated compared to other instant messaging software.
    - But notably, if we suggest adding a 'muted' state to a later MSC, we actually end up with this solution, just different verbiage:
       - unsubscribed = the same as ambient/new
       - subscribed = the same
       - muted = the same as purposefully-unsubscribed
   - So on the whole I don't think this MSC is blocking us from such an expansion in the future.

reivilibre: Aside from that, a future MSC could also tweak the push condition a bit so that the user can set an ambient subscription, either per-room or globally, so threads default to effectively-subscribed, but users can also purposefully unsubscribe specific threads.
There seems to be no harm in making this a separable part from this MSC.

proposals/4306-thread-subscriptions.md
reivilibre: f0a7680c720febdb64a7c445e7ef9f562eed4b2d now includes this. Anything else to change here?

proposals/4306-thread-subscriptions.md
reivilibre: e1ad4b2e827854a53fe0a833e79c14b8a016392b makes this optional. Should I add anything else to that end?

proposals/4306-thread-subscriptions.md
reivilibre: This seems to boil down to: we should add an endpoint for backpaginating these. I'm tempted to push that into MSC4308 since it seems to complement the Sliding Sync API nicely

proposals/4306-thread-subscriptions.md
reivilibre: I don't want to make unsubscriptions permanent — if you accidentally subscribe to a thread (misclick, etc), I feel that it would be weird to not be able to properly undo that.
If unsubscriptions are permanent (insofar as preventing automatic subscriptions again) then it'd be some weird half-way house between unsubscribed and muted.

I don't like that, I think it makes strange UI.

So my proposal in 648fda4e9b4e33df91e425100edd838fc282b907 is that we expose enough information to the server, so the server can apply ordering rules. (Until MSC4033 or similar, we can't do this on the client regrettably)

proposals/4306-thread-subscriptions.md
reivilibre: Yeah, you have it right.
Notably, the push rule `unsubscribed_thread` is lower priority than the push rules for mentions, so it won't prevent you getting notifications for mentions

Though looking at it now, this `unsubscribed_thread` does have higher priority than the rule that makes the server always push encrypted events to clients, so that's an angle we might need to find a solution for.
Could well be 'this rule needs to get ignored on encrypted events' :/

proposals/4306-thread-subscriptions.md
reivilibre: Added the semantics for this in 84be8db9cbcd58e13855a5cecb789188107318d5, let me know if that works :)

proposals/4306-thread-subscriptions.md
zecakeh: Fallback mentions are going to be removed from the spec since MSC4210 was accepted.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,255 @@
...
+
+As motivation, we want threads to have the following notification semantics:
+
+- Messages in unsubscribed threads should not count as activity at all; as a user, I do not want to see the room as unread because there are new messages in an unsubscribed thread.
+    - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
+- Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
+
+To achieve this, we propose the addition of two new push rules:
+
+1. a `content` push rule, called `.m.rule.unsubscribed_thread`, at the end of the content rule list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
```
zecakeh: Although I can understand why `content` seems to be the appropriate order for processing those rules, this would change the format of rules in `content`: they don't have `conditions` according to the spec, only a `pattern`.

reivilibre: yep, moving to underride. c974c027462c8bb4830ada29c99b40eeb8ae19d3
Could not immediately spot anything wrong with doing so.
Praying this is OK, otherwise we have no good options :)

proposals/4306-thread-subscriptions.md
reivilibre: added in c3fecfa90a488289788af6a4b3f67c5e82d6adbc, let me know if this suits

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,266 @@
...
+- physical timestamps (comparing the time of the manual unsubscription to the time of receipt of the cause event); or
+- logical timestamps (for example, position of the events within an event stream and/or topological ordering position within the room event DAG)
+
+In either case, servers must ensure that an automatic subscription cannot keep getting retriggered by clients.
+
+
+### New Client Behaviour: subscribe when sending in a thread
+
+When a user is sending a message in a thread, the user's client SHOULD simultaneously perform a manual subscription to that thread using `PUT /subscription`.
+The subscription is not considered automamtic, because it was initiated by a user action (sending a message). Further, it'd be best for latency to perform the subscription concurrently with sending the message, so the client doesn't yet have an event ID to attach a hypothetical automatic subscription to anyway.
```
anoadragon453:
```suggestion
The subscription is not considered automatic, because it was initiated by a user action (sending a message). Further, it'd be best for latency to perform the subscription concurrently with sending the message, as the client doesn't yet have an event ID to attach a hypothetical automatic subscription to anyway.
```

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,266 @@
...
+
+When a user is mentioned in a thread (by another user — the *mentioning user*), the user's client should perform an automatic subscription to that thread using `PUT /subscription` with `{"automatic": "$eventId"}`.
+
+The server does not perform this action on the client's behalf, principally because the server is not able to detect mentions in encrypted rooms.
+
+If the client is already aware of the user being subscribed to the thread, then making a `PUT /subscription` request is not necessary.
+
+If the mentioning user is banned in the room or ignored by the user, the automatic thread subscription should not occur.
+
+#### What counts as a mention
```
anoadragon453, line 140: I'm curious how this portion of the MSC would be translated to spec text. I worry if we have a separate list of "what's a mention", then that may because out of date over time.

Was this included to minimise inconsistent behaviour between clients w.r.t subscribing to threads?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,266 @@
...
+
+
+#### Reversal of automatic subscriptions
+
+*This section is an optional aspect. Clients can choose not to implement it, particularly in scenarios where it doesn't make sense.*
+
+If an automatic thread subscription occurs and the mentioning user is subsequently banned (or ignored by the user, but without loss of generality we refer to the user as banned hereinafter), then:
+
+- the thread subscription should be reversed,
+- provided that there aren't any other mentions by other, non-banned, users that would have caused the same automatic subscription.
```
anoadragon453, line 165: Servers have access to ignored users and membership metadata (currently). Would it not make sense for the server to take this on?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,266 @@
...
+    - Room moderators can take action against malicious users abusing the feature.
+    - If no room moderators are protecting the user, the user can of course also leave the room.
+
+## Notes to Client Implementors
+
+## Notes to Server Implementors
+
+- Since clients will be automatically updating the Thread Subscription Settings when their user is mentioned, server implementations should be ready to handle concurrent updates by multiple of the user's devices/clients when they are online at the same time.
+
+## Unstable amendments
```
anoadragon453, line 281: `M_SKIPPED` and `M_NOT_IN_THREAD` should have unstable alternatives as well.

anoadragon453 (commented):
Overall I love how clearly written this MSC is; kudos.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,269 @@
...
+}
+```
+
+In the general case, returns 200 with empty body `{}`.
+
+If an automatic subscription was requested but the provided cause event ID is not within the thread, returns 400/`M_NOT_IN_THREAD`.
+
+If an automatic subscription was requested but the automatic subscription is skipped by the server, returns 409 (Conflict) with `M_CONFLICTING_UNSUBSCRIPTION`. See the section below on *'Server-side automatic subscription ordering'*.
+
+If the thread does not exist, or if the user does not have access to it, returns 404/`M_NOT_FOUND`.
```
MadLittleMods, line 60: Is it possible to pre-subscribe to a message so if people thread off it, you can catch it?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,269 @@
...
+
+It would be undesirable if a user was automatically re-subscribed to a thread, after having already unsubscribed, because they opened a dormant device which was not aware of the automatic subscription having already taken place.
+
+To this end, we must prevent automatic subscriptions being re-applied when an unsubscription occurred *after* receipt of the event causing a subscription.
+
+As clients are currently not privy to the ordering of events, we propose that the server is left in charge of making the ordering decision.
+
+For this to work, when a client requests an automatic subscription, it presents the event ID of the latest event it has seen that would give rise to this automatic subscription (**the 'cause event'**).
+
+The server will skip the subscription if it determines that the automatic subscription is for an event that was already acknowledged at the time a manual unsubscription was requested.
```
MadLittleMods, line 110: *Spawning from a [comment on a implementation PR](https://github.com/element-hq/synapse/pull/18756#discussion_r2252683498),*

Currently, in Synapse, this is implemented as when a person unsubscribes, it will store the latest `stream_ordering` and `topological_ordering` in the room to compare against any future subscription 'cause' events.

Looking up the max `stream_ordering` and `topological_ordering` on the server at the time of unsubscription has the flaw that if the client has an outdated view of the world and sends an unsubscribe, we will be unsubscribed even though other clients may see new mentions in the thread that they want to subscribe to.

proposals/4306-thread-subscriptions.md
MadLittleMods: Pagination endpoint added in [MSC4308](https://github.com/matrix-org/matrix-spec-proposals/pull/4308)

More discussion -> https://github.com/element-hq/synapse/pull/18695#discussion_r2223748883

proposals/4306-thread-subscriptions.md
```diff
@@ -176,9 +176,9 @@ As motivation, we want threads to have the following notification semantics:
...
 - Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.
 
-To achieve this, we propose the addition of two new push rules:
+To achieve this, we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between `content` and `room` but which can contain general-purpose rules:
 
-1. an `underride` push rule, called `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
+1. `.m.rule.unsubscribed_thread`, at the beginning of the underride list. This rule causes events in unsubscribed threads to skip notification processing without generating a notification.
```
bnjbvr: nit: this still says "at the beginning of the underride list"

proposals/4306-thread-subscriptions.md
```diff
@@ -194,7 +194,7 @@ To achieve this, we propose the addition of two new push rules:
        "actions": []
    }
    ```
-2. an `underride` push rule, called `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
+2. `.m.rule.subscribed_thread`, at the beginning of the underride list (following `.m.rule.unsubscribed_thread`). This rule causes events in subscribed threads to generate notifications.
```
bnjbvr: ditto here

proposals/4306-thread-subscriptions.md
giomfo: sgtm

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+    - Room moderators can take action against malicious users abusing the feature.
+    - If no room moderators are protecting the user, the user can of course also leave the room.
+
+## Notes to Client Implementors
+
+## Notes to Server Implementors
+
+- Since clients will be automatically updating the Thread Subscription Settings when their user is mentioned, server implementations should be ready to handle concurrent updates by multiple of the user's devices/clients when they are online at the same time.
+
+## Unstable amendments
```
zecakeh, line 281: `postcontent` should use an unstable prefix for now.

bnjbvr: I've used `io.element.msc4306.postcontent` at the time being, in the Ruma PR implementing the new push rules changes. Not sure it should be that or `post_content`, depending on how would one capitalize it: `Postcontent` or `PostContent`. (Let the doors of bikeshedding hell open :))

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+    rather than simply notifying.
+      - `.m.rule.contains_user_name` and user-specified keyword mention rules must therefore execute first
+        so they can make use of `"set_tweak": "highlight"`.
+- this rule should come after `content` in case any users perform negative filtering (i.e. suppressing
+  notifications for messages containing certain keywords), although the author is not aware of clients with
+  such functionality.
+- this rule must come before `room` to avoid thread subscriptions being defeated by user rules such as
+  the ones underlying 'Mentions only' or 'Mentions & Keywords' settings available in common clients.
+
+For the time being, servers MUST ensure that clients MUST NOT be able to create custom rules with this push rule kind.
```
anoadragon453, line 249: Could you provide the HTTP status code and errcode clients should expect in this case?

The [Synapse implementation](https://github.com/element-hq/synapse/pull/18819/commits/c7c398eef8910d388880e898ec1e8192aae1ae29) uses 400 / `M_INVALID_PARAM`.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+
+```jsonc
+{
+  // If the subscription was made automatically
+  // by a client, this is the ID of the latest event that caused
+  // the automatic subscription.
+  //
+  // If the subscription was made manually,
+  // this field is omitted.
+  "automatic": "$abcdef"
```
ajbura, line 50: What will be the benefits of describing this?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+Returns 200 with empty body `{}`.
+
+If the thread was not subscribed, returns 200 with empty body `{}` for idempotence.
+
+If the thread does not exist, or is inaccessible, returns 404/`M_NOT_FOUND`.
+
+
+#### Retrieving thread subscription
```
ajbura, line 79: should not there be a way of fetching a list of all the threads user has subscribed to instead of fetching the status one by one?

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+## Background and Summary
+
+Threads were introduced in [version 1.4 of the Matrix Specification](https://spec.matrix.org/v1.13/changelog/v1.4/) as a way to isolate conversations in a room, making it easier for users to track specific conversations that they care about (and ignore those that they do not).
+
+Thus far, there has been no good way for users to ensure that they see new messages in only the threads that they care about. The current rules only allow expressing one of two choices: users and their clients can either watch all threads, or ignore all threads.
+As a result, the user is either being presented with unwanted information (misusing their attention), or they risk missing important messages.
+
+This proposal introduces a new mechanism, Thread Subscriptions, as a first-class way for users to signal which threads they care about.
+
+We then add a new push rule so that users can receive push notifications only for threads that they care about.
```
ajbura, line 13: If the proposal is to receive notification for all messages from "subscribed" thread could not that be achieved with the existing push rules semantics? Example:
```json
{
  "conditions": [
    {
      "kind": "event_property_is",
      "value": "m.thread"
    },
    {
      "kind": "event_property_is",
      "key": "content.m\\.relates_to.event_id",
      "value": "$thread_root_message_id"
    }
  ],
  "actions": [
    "notify",
    {
      "set_tweak": "highlight"
    }
  ],
  "rule_id": "$thread_root_message_id",
  "default": true,
  "enabled": true
}
```

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,304 @@
...
+When a client becomes aware of a banned user in a room, it may need to backpaginate thread history to determine whether there are any threads whose automatic subscriptions should be reversed, or to determine if there are any other qualifying mentions that would obviate the need to reverse automatic subscriptions.
+
+For efficiency reasons, clients may limit the depth of this backpagination with an implementation-defined recency limit, owing to the observation that abuse is usually cleaned up shortly after it occurs.
+
+### New Push Rules
+
+As motivation, we want threads to have the following notification semantics:
+
+- Messages in unsubscribed threads should not count as activity at all; as a user, I do not want to see the room as unread because there are new messages in an unsubscribed thread.
+    - Exceptions: if the user is mentioned, this should generate a notification as usual. (The push notification thus generated is also useful for the client to realise it needs to create an automatic thread subscription.)
```
ajbura, line 176: Since the default thread subscription state is "unsubscribed", could not this confuse user as the propose of showing "activity" is to advertise user that there can be possible interesting conversation. This suggestion can make user missing them.

Also the exceptions mentions to generate notification for "mentions" which includes room mentions, user/fallback mention, user configured keywords. but the push rule semantic is free for other personal customization wondering how that will with work with them.

proposals/4306-thread-subscriptions.md
bnjbvr: In the set of conditions, where does the fact that the user is subscribed (or not) to a thread appear?

ajbura: Client can check for the `rule_id` generated with thread root `event_id` in `m.push_rules`. If it finds the rule that means user has opt for "subscribed" (if rule has actions) and "unsubscribed" (if actions are empty). If it does not find any rule, client will work how it works currently and suggest for "subscribe" / "mute" thread button in UI.

bnjbvr: Interesting idea, thanks for bringing it up. I think that it's a tradeoff: the approach you're suggesting here would likely make the size of the rules event blow up, especially thanks to the prescribed automatic subscription mechanisms which could add lots and lots of subscribed threads by default. The approach suggested in this MSC makes it a default rule, that requires a bit more work during evaluation of the push rules (now, a client might need to fetch via network, or load from disk, the subscription status of a thread). The MSC's approach doesn't make the push rules event grow linearly with the number of threads one's subscribed to/muted, which I find a strong benefit here.

proposals/4306-thread-subscriptions.md
reivilibre: The default 'unsubscription' state is intended to reduce noise and notification fatigue, so keeping it squirreled away in the UI is kind of intentional — not that you can't have other UI to surface missed activity there, just that it shouldn't (in principle) count as noisy room activity.

The semantics are written in https://github.com/matrix-org/matrix-spec-proposals/pull/4306/files#diff-f42a2b8bda41aad63a982965d5f5b7165e62ddc06484c0ee3806024646ab9fdaR140 and should generalise semi-well (about as well as one could hope, given the free-for-all nature of push rules) to custom rules.

proposals/4306-thread-subscriptions.md
reivilibre: Main problems I can see are that I would rather avoid speccing anything that depends on an unlimited account data size, ultimately.

Other than that, this doesn't support the automatic subscription ordering logic that we need.

proposals/4306-thread-subscriptions.md
reivilibre: MSC4308 provides this

proposals/4306-thread-subscriptions.md
reivilibre: For the automatic subscription ordering rules, please see the section later :)

proposals/4306-thread-subscriptions.md
reivilibre: Hmm, the secret is, we don't have a concept of mention for the subscription rules, that's just the *intent*. It's actually based on the existing push rules. (Kind of hard to read in between all the threads that interrupt, but if you look closely it's there, I think!)

So the spec text should probably try to make a stronger distinction with e.g. 'subscribe-on-notified', perhaps clarifying that the *intent* was for mentions to trigger this, but not only.

proposals/4306-thread-subscriptions.md
reivilibre: Hm, interesting point. Wasn't intending for the server to store the information about which events contain mentions (it's not super nice for metadata); was also ultimately hoping that sending the event IDs would be a temporary workaround until clients can determine ordering themselves.

Maybe we are just essentially divulging metadata to the server anyway, without making the most of the information server-side? Not sure.

proposals/4306-thread-subscriptions.md
reivilibre: My justification here is that, talking of user action, 'unsubscribe means unsubscribe' and 'subscribe means subscribe'. i.e. if the user wants to do something, we don't question them.
I think this is ultimately *defensible*, even if you could argue both ways about it.

It's a fairly slim window for a race, and even still, if you clicked unsubscribe at 18:09 but later scroll down and notice you are (still) auto-subscribed because of an event at 18:07 that you didn't read yet, would you think the software was being reasonable or unreasonable?

proposals/4306-thread-subscriptions.md
ajbura: The proposals says:
> I do not want to see the room as unread because there are new messages in an unsubscribed thread.

It says *unread* (generally a dot without count) compare to notification (noisy/not) (badge with count). Notifications as badge-with-count are suppressed purposefully but unread as dot-without-count indicate general room activity is what I have concern about.

proposals/4306-thread-subscriptions.md
ajbura: After reading again through the msc, I have following concerns which might be helpful for this alternative consideration:
1. Automatic subscription have exploitable behaviour and "Malicious users can be ignored by the user" is practically undoable as a system can not easily detect Malicious users, the same problem as we have with invites. And since the mention already create notification a user will notice the activity in thread and can decide to subscribe/ignore when they will open the thread.
2. Since the notification preference for "subscribe" and "unsubscribed" is applied on all rooms a user can not configure it per-room bases. Example: 1) I want to receive notification for all thread messages from my work related room and not for public community rooms. 2) I want sound notification from my subscribed thread from work room but without sound from public community rooms.
3. If a room is muted subscribed thread rule in `postContent` is suppressed as muted room rule will be in `override`.

We may can consider this kind of three rules:
1. In `underride` to configure default notification behaviour for all rooms and for all thread replies.
2. In `rooms` to configure notification behaviour for specific room for all thread replies.
3. In `override` to subscribe to specific thread

Although the event size will always be a limiting factor, but a client can provide a section for all subscribed threads to maintain this list but the experience of such a feature will depend on the quality of implementation.

bnjbvr: Re (2), is there anything preventing a client from reusing the `thread_subscription` condition kind in a `rooms` rule? I don't see any such restriction, so the behavior you're describing could be implemented by reusing this condition kind in this section.

proposals/4306-thread-subscriptions.md
reivilibre: Yes, suppressing the indicator of room-wide activity is considered a good thing, because it reduces the cognitive load on users. But for use cases where it's important to see all activity, I think it is easy to either disable the push rule or provide other UI if necessary

proposals/4306-thread-subscriptions.md
reivilibre: > (1) Automatic subscription have exploitable behaviour and "Malicious users can be ignored by the user" is practically undoable as a system can not easily detect Malicious users,

This is referring to the user's own 'ignored user' list, not any automated system.

(2) can  be configured with push rules, but we are cautious about suggesting exposing too many control knobs to users because it makes for confusing UI

(3) a room being muted when muted is indeed intentional. As described in the MSC, the mental model is that the 'Mentions & Keyword' setting becomes equivalent to 'Mentions, Keywords & Subscribed Thread' and this is the level we're working at. (Our hands are also tied whilst keeping the push rules as sane as possible, until push rules are replaced in the future)

---

The account data being of unlimited size is considered a mistake, fwiw, therefore I don't buy that we should rely on it for this.

ajbura: > Re (2), is there anything preventing a client from reusing the thread_subscription condition kind in a rooms rule? 

Yes, since the msc specify that:
> we propose the addition of two new push rules, both added to a new push rule `kind` called `postcontent`, which is ordered between content and room

any rule added in `rooms` with condition `kind`: `thread_subscription` will be overridden by the same rule in the `postcontent` since it has the high priority over `rooms`.

ajbura: > (3) a room being muted when muted is indeed intentional.

I am not sure if i get it correctly. What i mean was that msc says.
> Messages in subscribed threads should always count as a notification, and the (effective) room notification settings should not matter at all. E.g. the room can be muted, but if I, as a user, am subscribed to a thread, I still want to get a notification for new messages in that thread. If I do not want that, then I will unsubscribe.

user can not get notification for subscribed thread because the room mute rule is added to the `override` section and apply on every event and should suppress the subscribed thread notifications per the spec.

proposals/4306-thread-subscriptions.md
```diff
@@ -0,0 +1,305 @@
...
+If the mentioning user is banned in the room or ignored by the user, the automatic thread subscription should not occur.
+
+#### What counts as a mention
+
+To determine whether an event within a thread is considered as 'mentioning' the user,
+for the sake of the 'subscribe on mention' behaviour,
+clients evaluate the push rules against that event.
+
+If the evaluation of the push rules results in a `notify` action as the outcome, then
+that event is considered to be 'mentioning' the user.
```
Johennes, line 147: We've hit a problem with this in https://github.com/matrix-org/matrix-rust-sdk/issues/6735. The Rust SDK (like, I believe, many other SDKs / clients) defines an `AllMessages` notification mode where every message results in `notify`. This mode is incompatible with thread subscriptions because now you're also getting notifications for unsubscribed threads rendering the subscription feature essentially disabled.

I think what you actually want when using thread subscriptions is a notification mode that includes all messages in the main timeline plus messages in subscribed threads. Or in other words, `AllMessages` _without_ unsubscribed threads.

PR #4308, opened 2025-07-09
MSC4308: Thread Subscriptions extension to Sliding Sync

reivilibre:
Depends-on: #4186, #4306

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rei/msc_ssext_threadsubs/proposals/4308-sliding-sync-ext-thread-subscriptions.md)

### DIFF
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -128,7 +128,9 @@ URL parameters:
   If this token is not provided, then backpagination starts from the 'end'.
 
 - `to` (string, optional): a token used to limit the backpagination \
-  The token can be acquired from a Sliding Sync response.
+  The token, originally acquired from `pos` in a Sliding Sync response,
+  would be the same one used as the `pos` **request** parameter in the
+  Sliding Sync request that returned the `prev_batch`.
 
 - `limit` (int, optional; default `100`): a maximum number of thread subscriptions to fetch
   in one response. \
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -196,6 +196,8 @@ above they do not have any special meaning.
 Notably, `bump_stamp`s MUST NOT be compared between different threads, because servers MAY
 treat `bump_stamp`s as per-thread.
 
+`bump_stamp` MUST be an ECMAScript-compatible (Canonical JSON-compatible) integer.
+
 `bump_stamp`s are independent of the sliding sync session and remain valid for comparison
 even if the sliding sync connection is reset, or if the device is changed.
 Servers do not need to make `bump_stamp`s consistent across different users.
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -30,6 +30,8 @@ The Sliding Sync request format is extended to include the `thread_subscriptions
 
     // Used to opt-in to receiving changes to thread subscriptions.
     "thread_subscriptions": {
+      // Must be specified and true to enable the extension.
+      "enabled": true,
       // Maximum number of thread subscription changes to receive
       // in the response.
       // Defaults to 100.
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -196,6 +196,10 @@ above they do not have any special meaning.
 Notably, `bump_stamp`s MUST NOT be compared between different threads, because servers MAY
 treat `bump_stamp`s as per-thread.
 
+`bump_stamp`s are independent of the sliding sync session and remain valid for comparison
+even if the sliding sync connection is reset, or if the device is changed.
+Servers do not need to make `bump_stamp`s consistent across different users.
+
 ## Potential issues
 
 
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -142,7 +142,7 @@ Response body:
   },
   // If there are still more thread subscriptions to fetch,
   // a new `from` token the client can use to walk further
-  // backwards.
+  // backwards. (The `to` token, if used, should be reused.)
   "end": "OPAQUE_TOKEN"
 }
 ```
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -109,7 +109,7 @@ URL parameters:
 
 Response body:
 
-```
+```jsonc
 {
   // Required
   "chunk": {
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -10,23 +10,30 @@ Threads Subscriptions are proposed in [MSC4306](https://github.com/matrix-org/ma
 Sliding Sync is proposed in [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md) as a paginated replacement to `/_matrix/client/v3/sync` with smaller response bodies and lower latency.
 The `/_matrix/client/v3/sync` endpoint is notorious in real-world applications of Matrix for producing large response bodies (to the amplified detriment of mobile clients) and the high latency caused by waiting for the server to calculate that full payload.
 
-This MSC proposes an 'extension' to Sliding Sync that allows clients to opt-in to receiving updates to the user's thread subscriptions.
+This MSC proposes an 'extension' to Sliding Sync that allows clients to opt-in to receiving real-time updates to the user's thread subscriptions.
+
+To handle the case in which there have been many updates to the user's thread subscriptions and there are too many to return in
+Sliding Sync, a new companion endpoint is proposed to allow backpaginating thread subscriptions on the client's terms.
 
 ## Proposal
+
+### Sliding Sync extension
 
 The Sliding Sync request format is extended to include the `thread_subscriptions` extension as follows:
 
 ```jsonc
 {
   // ...
-  
+
   "extensions": {
     // ...
-    
-    // Used to opt-in to receiving updates to thread subscriptions.
+
+    // Used to opt-in to receiving changes to thread subscriptions.
     "thread_subscriptions": {
-      // Maximum number of thread subscription changes to receive.
+      // Maximum number of thread subscription changes to receive
+      // in the response.
       // Defaults to 100.
+      // Servers may impose a smaller limit than what is requested here.
       "limit": 100,
     }
   }
@@ -38,41 +45,105 @@ The response format is then extended to compensate:
 ```jsonc
 {
   // ...
-  
+
   "extensions": {
     // ...
-    
-    // Returns a limited window of updates to thread subscriptions
+
+    // Returns a limited window of changes to thread subscriptions
+    // Only the latest changes are returned in this window.
     "thread_subscriptions": {
-      "changed": [
-        {
-          "room_id": "!roomid:example.org",
-          "root_event_id": "$abc123",
-          
-          "subscribed": true,
-          
-          // must be present when subscribed is true,
-          // but must be absent when subscribed is false
-          "automatic": true
+      "changed": {
+        "!roomId1:example.org": {
+          // New subscription
+          "$threadId1": {
+            "automatic": true
+          },
+          // New subscription
+          // or subscription changed to manual
+          "$threadId2": {
+            "automatic": false
+          },
+          // Represents a removed subscription
+          "$threadId3": null
         },
-        {
-          "room_id": "!roomid:example.org",
-          "root_event_id": "$def456",
-          
-          "subscribed": false
-        },
-        ...
-      ]
+        "$roomId2:example.org": {
+          // ...
+        }
+        // ...
+      },
+
+      // A token that can be used to backpaginate other thread subscription
+      // changes that occurred since the last sync, but that were not
+      // included in this response.
+      //
+      // Only present when some thread subscriptions have been missed out
+      // from the response because there are too many of them.
+      "gap": "OPAQUE_TOKEN"
     }
   }
 }
 ```
 
+If two changes occur to the same subscription, only the latter change ever needs
+to be sent to the client. \
+Servers do not need to store intermediate subscription states.
+
+
+### Companion endpoint for backpaginating thread subscription changes
+
+```
+GET /_matrix/v1/thread_subscriptions
+```
+
+URL parameters:
+
+- `pos` (string, optional): a token used to continue backpaginating \
+  The token is either acquired from a previous `/thread_subscriptions` response,
+  or a Sliding Sync response. \
+  The token is opaque and has no client-discernible meaning. \
+  If this token is not provided, then backpagination starts from the 'end'.
+
+- `limit` (int, optional; default `100`): a maximum number of thread subscriptions to fetch
+  in one response. \
+  Must be greater than zero. Servers may impose a smaller limit than requested.
+
+Response body:
+
+```
+{
+  // Required
+  "chunk": {
+    "!roomId1:example.org": {
+      // New subscription
+      "$threadId1": {
+        "automatic": true
+      },
+      // New subscription
+      // or subscription changed to manual
+      "$threadId2": {
+        "automatic": false
+      },
+      // Represents a removed subscription
+      "$threadId3": null
+    },
+    "$roomId2:example.org": {
+      // ...
+    }
+  },
+  // If there are still more thread subscriptions to fetch,
+  // a new `pos` token the client can use to walk further
+  // backwards.
+  "pos": "OPAQUE_TOKEN"
+}
+```
+
+If two changes occur to the same subscription, only the latter change ever needs
+to be sent to the client. \
+Servers do not need to store intermediate subscription states.
+
+
 ## Potential issues
 
-When clients start a fresh sync with no initial state, it may be the case that there is a backlog of many thread_subscriptions to send down to the client.
-
-Servers MAY choose to return Thread Subscription Settings in an order that is more heuristically-useful to the client, such as 'most recently updated' or 'threads with most recent activity first', instead of 'oldest first'. This could be either for all Thread Subscriptions, or only the backlogged ones.
 
 ## Alternatives
 
@@ -86,7 +157,11 @@ Servers MAY choose to return Thread Subscription Settings in an order that is mo
 
 ## Unstable prefix
 
-TODO
+Whilst this proposal is unstable, a few unstable prefixes must be observed by experimental implementations:
+
+- the Sliding Sync extension is called `io.element.msc4306.thread_subscriptions` instead of `thread_subscriptions`
+- the companion endpoint is called `/_matrix/client/unstable/io.element.msc4308/thread_subscriptions` instead of `/_matrix/v1/thread_subscriptions`
+
 
 ## Dependencies
 
--- a/proposals/4308-sliding-sync-ext-thread-subscriptions.md
+++ b/proposals/4308-sliding-sync-ext-thread-subscriptions.md
@@ -72,16 +72,18 @@ The response format is then extended to compensate:
         // ...
       },
 
-      // A pair of tokens that can be used to backpaginate other thread subscription
+      // A token that can be used to backpaginate other thread subscription
       // changes that occurred since the last sync, but that were not
       // included in this response.
       //
-      // The tokens are to be used with the `/thread_subscriptions` endpoint
-      // as `from` and `to` parameters with `dir=b`.
+      // The token is to be used with the `/thread_subscriptions` endpoint
+      // as `from`, with `dir`=`b`.
+      // The `pos` parameter in the **request** would be used for the `to`
+      // parameter.
       //
       // Only present when some thread subscriptions have been missed out
       // from the response because there are too many of them.
-      "prev": {"from": "OPAQUE_TOKEN", "to": "OPAQUE_TOKEN"}
+      "prev_batch": "OPAQUE_TOKEN"
     }
   }
 }
@@ -105,7 +107,7 @@ URL parameters:
 
 - `from` (string, optional): a token used to continue backpaginating \
   The token is either acquired from a previous `/thread_subscriptions` response,
-  or a Sliding Sync response. \
+  or the `prev_batch` in a Sliding Sync response. \
   The token is opaque and has no client-discernible meaning. \
   If this token is not provided, then backpagination starts from the 'end'.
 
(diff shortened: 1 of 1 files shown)


proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,94 @@
...
+        ...
+      ]
+    }
+  }
+}
+```
+
+## Potential issues
+
+When clients start a fresh sync with no initial state, it may be the case that there is a backlog of many thread_subscriptions to send down to the client.
```
bnjbvr: As [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4306#discussion_r2223787760) in the other MSC, I think we'd need to a way to get *all* of them, one way or another. I'd rather not rely on increasing the `limit` parameter for this, otherwise:

1. we'd get duplicate data as we increase the limit,
2. we'd change the SSS connection parameters a lot, until we got all the subscriptions, and that may cause delays in processing of other things unrelated to the thread subscription (regular messages, call notifications, etc.)

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,94 @@
...
+        {
+          "room_id": "!roomid:example.org",
+          "root_event_id": "$abc123",
+          
+          "subscribed": true,
+          
+          // must be present when subscribed is true,
+          // but must be absent when subscribed is false
+          "automatic": true
+        },
```
bnjbvr: 1. Out of curiosity, why the `changed` key? I suppose it's for future backwards compat, might be worth saying it explicitly here?
2. Would it make sense to key the changes by `room_id`, in a hashmap above the array level? This would avoid a bit of duplication for rooms which have lots of changes in a single room.

```json
{
  "changed": {
    "!roomid:example.org": [{ "root_event_id": "$abc123", ... }]
  }
}
```

3. Maybe `root_event_id` could be only `root`, to spare a few bytes per new sub?
4. Maybe `subscribed` could be one of `false` (or `no`, or null), `automatic` (yes), `manual` (yes)? This would remove the need for the `automatic` field which presence depends on the `subscribed` field

reivilibre: Changed things around a bit here, to be denser and to more closely match the response of the individual `GET /thread_subscription`. Let me know what you think :)

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+      // Only present when some thread subscriptions have been missed out
+      // from the response because there are too many of them.
+      "gap": "OPAQUE_TOKEN"
+    }
+  }
+}
+```
+
+If two changes occur to the same subscription, only the latter change ever needs
+to be sent to the client. \
```
MadLittleMods, line 110: Are these on purpose? (applies elsewhere)

```suggestion
to be sent to the client.
```

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+        // ...
+      },
+
+      // A token that can be used to backpaginate other thread subscription
+      // changes that occurred since the last sync, but that were not
+      // included in this response.
+      //
+      // Only present when some thread subscriptions have been missed out
+      // from the response because there are too many of them.
+      "gap": "OPAQUE_TOKEN"
```
MadLittleMods: Based on prior art, feels like `prev_batch`/`next_batch` is the pattern to align with.

Ex.

 - `prev_batch` from `/sync` responses are supposed to be used with `/messages?from=xxx&to=xxx`
 - `next_batch` from `/sync` responses are supposed to be used with `/sync?since=xxx` and `/messages?from=xxx&to=xxx`
 - `pos` in [Sliding sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) responses are supposed to be used with Sliding Sync `/sync?since=xxx`
 - `start`/`end` from `/messages` responses are supposed to be used with `/messages?from=xxx&to=xxx`
 - `next_batch` from `/threads` responses is supposed to be used with `/threads?from=xxx`
 - `next_token` from `/notifications` responses is supposed to be used with `/notifications?from=xxx`
 - `next_batch` from `/hierarchy` responses is supposed to be used with `/hierarchy?from=xxx`


Overall, I find our API surface a bit confusing on how we've named fields.

We probably *do* want to include pagination tokens for both sides of the response (`prev`/`next`) so clients can fill in the gap from both directions as desired.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+
+If two changes occur to the same subscription, only the latter change ever needs
+to be sent to the client. \
+Servers do not need to store intermediate subscription states.
+
+
+### Companion endpoint for backpaginating thread subscription changes
+
+```
+GET /_matrix/v1/thread_subscriptions
```
MadLittleMods: I love the bulk endpoint :+1: but is there also a way to get `thread_subscriptions` for a single room?

I can definitely picture some UI which shows the subscribed threads in a room and it would suck to have to wait for the whole room list to paginate in before you could see the list for the single room you're staring at.

The dumb-simple solution is a filter parameter for this endpoint instead of a separate endpoint.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+  The token is opaque and has no client-discernible meaning. \
+  If this token is not provided, then backpagination starts from the 'end'.
+
+- `limit` (int, optional; default `100`): a maximum number of thread subscriptions to fetch
+  in one response. \
+  Must be greater than zero. Servers may impose a smaller limit than requested.
+
+Response body:
+
+```
```
MadLittleMods:
```suggestion
```jsonc
```

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+Servers do not need to store intermediate subscription states.
+
+
+### Companion endpoint for backpaginating thread subscription changes
+
+```
+GET /_matrix/v1/thread_subscriptions
+```
+
+URL parameters:
```
MadLittleMods, line 119: Per my [other comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4308#discussion_r2267950677),

> We probably *do* want to include pagination tokens for both sides of the response (`prev`/`next`) so clients can fill in the gap from both directions as desired.

This would also require a `dir` direction parameter. I'm not sure it's necessary to refactor to a `from`/`to` pattern like `/messages` but maybe more obvious if they behave the same.

We should also clarify the token bounds. Synapse reference for other places:

 - https://github.com/element-hq/synapse/blob/d27438bc253b772cadaa7cbd914bbd5631ee3b05/synapse/storage/databases/main/stream.py#L690-L691
 - https://github.com/element-hq/synapse/blob/d27438bc253b772cadaa7cbd914bbd5631ee3b05/synapse/handlers/sliding_sync/__init__.py#L202

proposals/4308-sliding-sync-ext-thread-subscriptions.md
bnjbvr: Out of curiosity, what's the use case for a `next` token? i.e., in which cases is one interested in getting a thread subscriptions and the ones *after* it? I feel that we're not going to need this in the short term, and, if we need it, we could also add it when it's actually justified by a use case. So, as a wish to keep the API as minimal as it can be, I'd be in favor of *not* adding support for `next` as well as a `dir`ection parameter, unless it's strongly motivated by a real-world use case.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,169 @@
...
+{
+  // ...
+
+  "extensions": {
+    // ...
+
+    // Returns a limited window of changes to thread subscriptions
+    // Only the latest changes are returned in this window.
+    "thread_subscriptions": {
+      "changed": {
```
bnjbvr: Out of curiosity, do you have ideas for other sibling fields at this level? Or is this for symmetry with other response subparts?

(The one idea that comes to mind is that the request/response parts could be a subset of a new `threads` field, something like:

```json
 "threads": {
    "subscriptions": {
      "!roomId1:example.org": { ... }
    }
  }
```

)

bnjbvr (commented):
Thanks, I like this updated version a lot!

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: There isn't, but we expect clients to be needing to cache this information locally anyway, so it'd cover a very tiny surface (i.e. during the time we haven't managed to fully sync yet) on a UI I'm not actually sure anyone would implement (rather than 'list of my subscriptions in this room in time order of the subscription', wouldn't you instead be interested in 'list of most recent threads in this room, filtered by whether I'm subscribed').

So I kinda felt it was not worth adding extra API surface here given I don't have a use case in mind for it. If someone disagrees this could easily be rectified in the future.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: I'd also rather not add a 'forward' direction. Maybe we can add the `dir=b` parameter now and require it to always point backwards.

I am struggling to line up the semantics I want here with the existing API patterns, which is namely tracking and filling a gap (i.e. there are bounds on both sides).

The clients should not really care about the direction of filling, because the time of a subscription is not really important, just the latest state of it.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: Other than `pos` which now exists (although might be renamed, c.f. the other thread), the idea was to not block off extensibility; i.e. if we just flattened this struct then we would have no avenue to extend this Sliding Sync extension.

`threads -> subscriptions` might work, but by the same token, the extension is called `threads_subscriptions`, so it is weird for the answer to not match the extension name (and I think the Sliding Sync MSC actually says they should match). It would still have the same extensibility problem :/

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: This is the most sane of the two ways of adding a line break in the more standard variants of Markdown.
(GFM is I think the only notable dialect that doesn't need this)

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: I have changed the `/thread_subscriptions` API to use `from=xxx&to=xxx&dir=b`, but wondering how we should pass back the pair of tokens from Sliding Sync? I've added a new proposal for that but I'm not sure it's the most satisfying (though couldn't see endpoints that do this already for double-ended bounds)

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: Thinking on this, this also means servers need to add indices for this and optimise a bit for it. It's not really for free. And if we don't think of a valid use for it, I guess that's actually harmful

MadLittleMods: If I come back to my desktop after a week of working on my phone on a holiday, my desktop client is going to need to catch up on the threads I'm subscribed to. If I click into a room and open a threads panel, I will be waiting until my clients catches up on everything whereas it could use the dedicated room endpoint to get everything directly ahead of the entire list on my account.

"filtered by whether I'm subscribed" is the key here. I need to know whether some threads are subscribed.

reivilibre: Hmm, I see what you'd mean, I'd kinda like a client dev who is implementing this to provide some kind of opinion though on what would work for them.

An ugly but maybe-valid workaround in that case could be to backpaginate the threads in general and check the subscriptions one-by-one manually. Given HTTP pipelining and multiplexing, there are probably worse crimes.
I think more effective would be to inline the subscription directly in the APIs where that makes sense. e.g. filtering the threads to only subscribed ones, directly in `/threads`, or at least returning the state there.

bnjbvr: > If I come back to my desktop after a week of working on my phone on a holiday, my desktop client is going to need to catch up on the threads I'm subscribed to. If I click into a room and open a threads panel, I will be waiting until my clients catches up on everything whereas it could use the dedicated room endpoint to get everything directly ahead of the entire list on my account.

Remember how thread subscriptions are entirely controlled by clients, and never created by the server? If you came back after a holiday and didn't open any client for a week, then you won't have any new subscriptions, by definition. Then, your client would catch up on the threads (not the thread subscriptions), using either automatic back-pagination for all the threads (bleh) or a new threads catchup mechanism (yay), and then only would it recompute new subscriptions for the threads that have been active during your holiday.

If other devices have sync'd and did subscribe you to threads: you'd receive the updates from the sliding sync extension, and then if there were too many of them, you'd use the global endpoints.

Having a room-specific endpoint might likely make the second case above slightly faster (if the other device subscribed to many other threads on many other rooms). But if this endpoint's processing speed is Good Enough™ (viz. not blocked on network/federation), I imagine that it would be sufficient to have only the global feed, for the time being. Does it make sense?

> An ugly but maybe-valid workaround in that case could be to backpaginate the threads in general and check the subscriptions one-by-one manually. Given HTTP pipelining and multiplexing, there are probably worse crimes.

For what it's worth, I did experiment with backpaginating all of a room's threads, when implementing a dummy activity center in the SDK, and this was way too slow for a room like MatrixHQ (of course this is the extreme edge case).

> I think more effective would be to inline the subscription directly in the APIs where that makes sense. e.g. filtering the threads to only subscribed ones, directly in /threads, or at least returning the state there.

Maybe this could be part of the [bundled thread summary](https://spec.matrix.org/latest/client-server-api/#server-side-aggregation-of-mthread-relationships), for thread roots, next to the `current_user_participated` which really thread subscriptions are a generalization of?

Also: note that filtering to only subscribed ones wouldn't give you the information that you may have unsubscribed from another device in the meanwhile.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+  endpoints. The forward direction is not yet specified to be implemented.
+
+- `from` (string, optional): a token used to continue backpaginating \
+  The token is either acquired from a previous `/thread_subscriptions` response,
+  or the `prev_batch` in a Sliding Sync response. \
+  The token is opaque and has no client-discernible meaning. \
+  If this token is not provided, then backpagination starts from the 'end'.
+
+- `to` (string, optional): a token used to limit the backpagination \
+  The token can be acquired from a Sliding Sync response.
```
bnjbvr: Hmm, the wording is confusing here as it's talking about the sliding sync response. We might want to precise here again that it's the `pos` value from the latest *request*?

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+      // included in this response.
+      //
+      // The token is to be used with the `/thread_subscriptions` endpoint
+      // as `from`, with `dir`=`b`.
+      // The `pos` parameter in the **request** would be used for the `to`
+      // parameter.
+      //
+      // Optional. Only present when some thread subscriptions have been
+      // missed out from the response because there are too many of them.
+      "prev_batch": "OPAQUE_TOKEN"
```
bnjbvr, line 103: Should we add a `client behavior` section defining that ideally the thread subscriptions would all be persisted on the client's side, to avoid having to call into the `GET` endpoint for individual thread subscriptions when computing notifications locally (notably for e2ee rooms)? The "ideally" in my sentence suggests this could be worded as a "MAY".

Then I guess, to implement this optimization correctly, a client would have to remember if it's synchronized all threads subscriptions yet, with the following assumptions:

- when a SSS connection has been reset (via a new `pos`), a client should assume it doesn't know about all the thread subscriptions yet (set an hypothetical boolean `known_thread_subscriptions` to false) 
- after receiving a response *and* paginating all the previous unknown thread subscriptions, the client should assume it does know with certainty about all the thread subscriptions (set `known_thread_subscriptions` to true)
- when getting a thread subscription: if `!known_thread_subscriptions`, a GET request should be sent to figure whether a thread has a subscription or not; otherwise, the state synchronized locally can be used.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+and `/sync` request both return information about the same thread subscription.
+
+The semantics of the `bump_stamp` are that for two updates about the same thread,
+the update with the higher `bump_stamp` is later and renders the update with the lower
+`bump_stamp` obsolete.
+
+Clients MUST NOT interpret any other semantics of the `bump_stamp`; other than the semantics
+above they do not have any special meaning.
+Notably, `bump_stamp`s MUST NOT be compared between different threads, because servers MAY
+treat `bump_stamp`s as per-thread.
```
bnjbvr, line 199: Is the `bump_stamp` guaranteed to be an integer, or can it be a floating-point value? Is it guaranteed to be non-negative, or can it be negative?

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+
+The pagination structure of this endpoint matches that of the `/messages` endpoint, but fixed
+to the backward direction (`dir=b`).
+For simplicity, the `start` response field is removed as it is entirely redundant.
+
+### Use of `bump_stamp`
+
+The `bump_stamp`s within each thread subscription can be used for determining which
+state is latest, for example when a concurrent `/thread_subscriptions` backpagination request
+and `/sync` request both return information about the same thread subscription.
```
bnjbvr, line 190: For what it's worth, my implementation will likely now store all the bumpstamps in the database (so as to be able to order the latest sub vs unsub, notably). This creates a small discrepancy, when using the endpoints introduced from MSC4306:

- when we use the PUT or DELETE endpoints, we don't receive a bump_stamp. In this case, it's expected that the bumpstamp will come with the remote echo from the next sync. If we want to implement "local echo" behvaior when using the `PUT` or `DELETE` endpoints, then we'd need to store the latest status (subscribed/unsubscribed) in the database, and we wouldn't have a new bumpstamp. I think that in this case, it might be sufficient to *not* update the bumpstamp.
- When using the GET endpoint, we don't receive a bumpstamp, but that's fine: it's supposed to be the latest version anyways. Ideally, we'd cache the result in database too, here. And maybe we can perform the same hack as indicated above (keep the previous bumpstamp stored in DB, and only update the subscription status, if it's changed).

TL;DR: adding `bump_stamp` to the MSC4306 endpoints might be better for API consistency, but not sure it's achieving anything useful, considering the hack described above.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+
+The semantics of the `bump_stamp` are that for two updates about the same thread,
+the update with the higher `bump_stamp` is later and renders the update with the lower
+`bump_stamp` obsolete.
```
bnjbvr, line 194: Does this guarantee (greater bump stamp means more recent info) hold across sliding sync sessions? (i.e., if i restart my app / the `pos` is cleared / the sliding sync connection is marked expired)

(I think it must, otherwise it wouldn't work correctly across restarts, but just wanted to clarify this.)

proposals/4308-sliding-sync-ext-thread-subscriptions.md
```diff
@@ -0,0 +1,221 @@
...
+  "extensions": {
+    // ...
+
+    // Used to opt-in to receiving changes to thread subscriptions.
+    "thread_subscriptions": {
+      // Maximum number of thread subscription changes to receive
+      // in the response.
+      // Defaults to 100.
+      // Servers may impose a smaller limit than what is requested here.
+      "limit": 100,
```
bnjbvr, line 39: - it'd be nice to add the `enabled`: boolean here too, for consistency with other extensions (and because the synapse impl requires it, and I was confused as to why my request wasn't taken into account xD)
- also i see in the impl that `limit` is required, but I wonder if it should? I didn't put it first, assuming the server would give me a nice default value.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: let's define it to be ECMAScript-compatible int: b3ac5ee7fed8fca5ec40313073140614d5e66403

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: *Tried* (and maybe failed) to improve with 4d88858787dd73ff948368e3cd58b124fcfed9bd

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: semantics spelled out in cef3446ed3db0e49a56b617cc83ced0ac4a7a522, nice thinking

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: Added the enabled flag, sorry for forgetting it!

The `limit` is now opened as a bug in Synapse.

proposals/4308-sliding-sync-ext-thread-subscriptions.md
MadLittleMods: Related conversation https://github.com/matrix-org/matrix-spec-proposals/pull/4186#discussion_r2104882482

proposals/4308-sliding-sync-ext-thread-subscriptions.md
MadLittleMods: Perhaps the bulk endpoint should work like this:

> We could implement a bulk `/messages` endpoint, where the client would specify multiple rooms and `prev_batch` tokens.
>
> *-- [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)*

And this also gets around the problem where you can't get to the thread subscriptions that you care about without paginating through the bulk. This way, you can specify the rooms you want to get more thread subscriptions in.

Not sure how fetching thread subscriptions across *all* rooms would look like yet (many options).

Also being explored in https://github.com/element-hq/synapse/pull/19005#discussion_r2436863944

proposals/4308-sliding-sync-ext-thread-subscriptions.md
reivilibre: Just testing this now, the limit flag does exist and it does default to 100. So not sure what you were seeing or if you were perhaps on an old draft? :S

proposals/4308-sliding-sync-ext-thread-subscriptions.md
MadLittleMods: Matrix spec on pagination conventions: `prev_batch` and `next_batch`, see https://spec.matrix.org/v1.18/appendices/#pagination

PR #4310, opened 2025-07-10
MSC4310: MatrixRTC decline `m.rtc.decline`

toger5:
[rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC-call-decline/proposals/4310-MatrixRTC-call-decline.md)

---

name: Proposal ready for review
about: A proposal that is ready for review by the core team and community.
title: ''
labels: proposal, proposal-in-review
assignees: ''

---

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/master/CONTRIBUTING.md#sign-off)
* [ ] Update the title and file name of your proposal to match this PR's number (after opening).
* [ ] Pull request includes a ['Rendered' link](https://matrix.org/docs/spec/proposals#process) above.
* [ ] Your MSC adheres to each point in the [MSC Checklist](MSC_CHECKLIST.md).
* [ ] Ask in
      [#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) to
      get feedback on this PR.


proposals/4310-matrixRTC-call-decline.md
```diff
@@ -0,0 +1,150 @@
...
+### Proposal (Changes)
+
+This MSC proposes the `m.rtc.decline` event type, with a standard `m.reference`
+relation and a `reason` field.
+
+- relation: `"m.relates_to": {"rel_type": "m.reference", "event_id": "$call_notification_event_id"}`
+- reason:
+
+  ```json
+  "reason"?: "decline description"
```
dbkr: I'd fairly strongly vote against a string reason field: we have a collection of these string fields in the protocol, mostly not used for anything, and when we do they i18n poorly and we're trying to replace them. I'd suggest removing it unless we actually have a concrete use case for it, in which case spec some standard codes please.

proposals/4310-matrixRTC-call-decline.md
```diff
@@ -0,0 +1,150 @@
...
+  }
+}
+```
+
+Client handling guidance:
+
+- For 1:1 rings, on receipt of a decline, stop ringing and show a declined state.
+- For group rings, on receipt of a decline from a participant, update that
+  participant’s state. The ring should continue until at least one participant
+  accepts or all have declined/timeout.
```
dbkr: This implies "hunt group" like semantics where you're trying to establish a 1:1 call with any one member of a group. This surprises me since MatrixRTC calls are first & foremost group calls, so I'd expect the default behaviour to be that it rings for everyone until they answer/decline/time out maybe?

dbkr (commented):
Just a couple of comments as I saw this whilst reviewing the js-sdk PR.

proposals/4310-matrixRTC-call-decline.md
toger5: That of course also works. The codes sound like a good idea but probably also worth waiting until we have a concrete idea what we want here.

proposals/4310-matrixRTC-call-decline.md
toger5: This is phrased a bit unfortunatly. It was referencing the ring feedback on the sender side. But I am not sure if this is even relevant in the MSC or can be left as a client decision.

proposals/4310-MatrixRTC-call-decline-relations.svg
zecakeh, line 1: All bubbles should have a background color, this is barely legible in dark mode.

zecakeh, line 1: The `reason` field should be removed.

proposals/4310-MatrixRTC-call-decline.md
```diff
@@ -0,0 +1,148 @@
...
+
+### Proposal (Changes)
+
+This MSC proposes the `m.rtc.decline` event type, with a standard `m.reference`
+relation.
+
+- relation: `"m.relates_to": {"rel_type": "m.reference", "event_id": "$call_notification_event_id"}`
+
+The `m.relates_to` field allows referencing the original `m.rtc.notification` event.
+The optional reason can provide a message to the user receiving the decline.
```
zecakeh, line 51: The `reason` field was removed.

proposals/4310-MatrixRTC-call-decline.md
```diff
@@ -0,0 +1,148 @@
...
+which introduces the `m.rtc.notification` event.
+This event lives in parallel with the MatrixRTC session and uses intentional mentions to make the receiving clients ring.
+The receiving clients will then monitor the room state and stop ringing once the user joined with one of their devices.
+
+## Proposal
+
+Conceptually this MSC will introduce an event, sent as a relation to the
+`m.rtc.notification` event, which communicates a decline from one or more
+parties. It can be used on clients to provide a good UX around a call decline
+(stop ringing, play a decline sound, prompt the user with: "the call has been
```
BillCarsonFr, line 22: This feels wrong to me. I cannot see any other app that would play a specific song or a promp when the other person declines your call. That would be very rude.
A good UX would be to just stop ringing and not shame me that the other side rejected my call :D

proposals/4310-matrixRTC-call-decline.md
BillCarsonFr: I think the reason is very important.
Declining a call is very rude, usually people would just silence their phone and let it ring until it times out.
If it is someone you know you will decline and immedialty send a text to explain why you cannot answer.

The decline icoming call usual experience is:
- Silence the call
- Decline
- Decline with a message , pre-defined`Can I call you later?` `Sorry can't talk now` or a custom one.

So there is a clear use case for reason, it is common to the decline experience.

What we miss is the `Silence call` use-case (silence on my other devices if any)

proposals/4310-matrixRTC-call-decline.md
```diff
@@ -0,0 +1,148 @@
...
+- Clients have two possible ways of displaying **historic calls in the
+  timeline**. They can compute the exact call history using the `m.rtc.member`
+  events which contain the information at what time, which device joined.
+  Alternatively they can use the notify and decline events only to render when a
+  person tried to start a call and if that got declined. This is much simpler
+  and therefore desired for very simple client implementations.
+
+### Proposal (Changes)
+
+This MSC proposes the `m.rtc.decline` event type, with a standard `m.reference`
```
BillCarsonFr, line 45: Sould the `m.rtc.decline` also uses intentional mention, to be symetrical to the `m.rtc.notification` event?
Like that Apps that woke up to ring, would woke up to stop ringing?

proposals/4310-matrixRTC-call-decline.md
BillCarsonFr: As discussed somewere else, maybe a regular follow-up reply/message can be used for the reason?

proposals/4310-MatrixRTC-call-decline.md
Johennes: These are just examples, not normative.

Johennes:
This is obsoleted and now included in https://github.com/matrix-org/matrix-spec-proposals/pull/4075

PR #4311, opened 2025-07-11
MSC4311: Ensuring the create event is available on invites

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4311-stripped-state-create-event.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published with my role as a member of the SCT.

Dependencies:
* https://github.com/matrix-org/matrix-spec-proposals/pull/4291

----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#issuecomment-3168942186)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#issuecomment-3168934604)

### DIFF
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -2,7 +2,7 @@
 
 Historically, when processing an incoming invite or outgoing knock, safety tooling would parse the room ID despite
 [being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids), to determine the server which
-originally created the room. If that server was considered abusive, the invite or 
+originally created the room. If that server was considered abusive, the invite or
 knock may be rejected or blocked early by the tooling. Note that checking the domain of the
 sender of an invite is inadequate, because the sender may not be on the same server as the
 user who created the room.
@@ -43,8 +43,9 @@ ID by hashing the event in MSC4291+ room versions. For other room versions (1 th
 can at most compare the `room_id` field of the create event with the invite/knock membership event.
 
 If any of the events are not a PDU, not for the room ID specified, or fail [signature checks](https://spec.matrix.org/v1.15/server-server-api/#validating-hashes-and-signatures-on-received-events),
-or the `m.room.create` event is missing, the receiving server MUST respond to invites with a `400 M_MISSING_PARAM`
-standard Matrix error (new to the endpoint). For knocks, the server SHOULD remove any events from
+or the `m.room.create` event is missing, the receiving server MAY respond to invites with a `400 M_MISSING_PARAM`
+standard Matrix error (new to the endpoint). For invites to room version 12+ rooms, servers SHOULD
+rather than MAY respond to such requests with `400 M_MISSING_PARAM`. For knocks, the server SHOULD remove any events from
 `knock_room_state` which fail the same validation check before passing the details along to clients.
 Ideally, the server would be able to prevent the knock from happening, though by the time the server
 can see the `knock_room_state`, the knock has already happened.
@@ -93,11 +94,12 @@ affecting client or server implementations.
 
 Mentioned above, existing server implementations SHOULD warn rather than fail on invites which don't
 have complete PDUs inside `invite_room_state` until their local ecosystem adoption allows for the
-full set of validation to be applied. If PDUs are complete, but for a different room, the invite MUST
-still fail per the validation above.
+full set of validation to be applied. If PDUs are complete, but for a different room, the invite SHOULD
+still fail in v12 rooms per the validation above.
 
 This proposal suggests that servers wait no longer than 3 months (or about 1 full spec release cycle)
-after this proposal is released to enforce the full validation.
+after this proposal is released to enforce the full validation, though servers may extend this as
+needed for their ecosystems to gain support.
 
 
 ## Dependencies
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -1,11 +1,11 @@
-# MSC4311: Ensuring the create event is available on invites and knocks
+# MSC4311: Ensuring the create event is available on invites
 
-Historically, when processing an invite or knock, safety tooling would parse the room ID despite
-[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
-originally created the room. If that server was considered abusive, the incoming invite or outbound
-knock may be rejected or blocked early by the tooling. This approach is preferred because the user
-sending the invite may not be on the same server as the user who created the room, though both sender
-and creator are checked by safety tooling.
+Historically, when processing an incoming invite or outgoing knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids), to determine the server which
+originally created the room. If that server was considered abusive, the invite or 
+knock may be rejected or blocked early by the tooling. Note that checking the domain of the
+sender of an invite is inadequate, because the sender may not be on the same server as the
+user who created the room.
 
 With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
 domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
@@ -13,7 +13,8 @@ recommending rather than requiring the `m.room.create` event, makes the above ch
 impossible when the create event is missing or incomplete, as the room ID cannot be confirmed in
 MSC4291+ room versions.
 
-This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
+To mitigate the problem in the case of invites,
+this MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
 to ensure the event matches the room. To support the new validation, the `m.room.create` event must
 be formatted as a full PDU in the stripped state of [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
 over federation. Similar treatment is applied to other stripped state events for uniformity.
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -29,7 +29,7 @@ have already sent the knock before stripped state is received.
 On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
 where available. No other changes are proposed to the Client-Server API. For clarity, this means clients
 continue to receive events which only have `content`, `sender`, `state_key` (optional), and `type` in
-the `invite_room_state`, `knock_room_state`, and whereever else stripped state is used.
+the `invite_room_state`, `knock_room_state`, and wherever else stripped state is used.
 
 Over federation, servers MUST include the `m.room.create` event in the [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
 and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid).
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -31,17 +31,18 @@ where available. No other changes are proposed to the Client-Server API.
 
 Over federation, for room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
 the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
-and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid)
 and MUST be a properly-formatted PDU according to that room version's event format specification. The
 full PDU format is used to ensure that receiving applications can independently verify the room ID
 by calculating the reference hash of the create event themselves.
 
 If the `m.room.create` event is not present, not a PDU, or not for the room ID specified, the server
-MUST fail to continue processing the invite or knock. For invites, this is a `400 M_MISSING_PARAM`
-standard Matrix error (new to the endpoint). For knocks, this means the server drops the `make_knock`
-response and never completes a `send_knock`. For both operations, the associated Client-Server API
-request is failed with `500 M_BAD_STATE`. A 4xx error isn't used for the Client-Server API because
-there's nothing the client can materially do differently to fix that request.
+MUST respond to invites with a `400 M_MISSING_PARAM` standard Matrix error (new to the endpoint). For
+knocks, the server SHOULD remove the `m.room.create` event from `knock_room_state` before passing the
+information along to clients. Ideally, the server would be able to prevent the knock from happening,
+though by the time the server can see the `knock_room_state`, the knock has already happened. A 4xx
+error isn't used for the Client-Server API because there's nothing the client can materially do
+differently to fix that request.
 
 For room versions *not* affected by MSC4291, servers SHOULD include the properly-formatted `m.room.create`
 PDU. This is not made mandatory to avoid a situation where servers trust data that shouldn't be trusted
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -116,6 +116,25 @@ This proposal does not require an unstable prefix as the behaviour can be accomp
 affecting client or server implementations.
 
 
+## Migration
+
+*This section is non-normative for spec writing purposes. It only affects implementations which have
+implemented room version 12 upon its release.*
+
+Room version 12 contains MSC4291 and is expected to be used in production prior to this proposal
+becoming stable itself. To account for this, servers SHOULD treat "MUST" as "MAY" throughout this
+proposal until 1 full spec release cycle has passed since this MSC's own release in the specification.
+
+This translates to a timeline anywhere between 2 and 6 months, depending on ecosystem rollout. An
+example *possible* release schedule is:
+
+1. August 2025 - Matrix 1.16 is released with Room Version 12.
+2. October 2025 - Matrix 1.17 is released with this proposal; servers use "MAY" keywords.
+3. January 2026 - Matrix 1.18 is released; servers switch to "MUST" keywords.
+
+Servers MAY switch to "MUST" keywords early if their local ecosystems are prepared for the change.
+
+
 ## Dependencies
 
 This proposal requires [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291) in
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -50,7 +50,8 @@ while one without a domain component (and happens to be `!<43 unpadded urlsafe b
 affected. This is done to ensure the server is not potentially confused by a malicious server providing
 a create event for a different, unaffected, room.
 
-When a room is affected, the server MUST validate the `m.room.create` event as follows:
+When a room is affected, the server MUST validate the `m.room.create` event as follows for the purposes
+of the above:
 
 1. If the event has a `room_id`, reject.
 2. If the event does not otherwise comply with the event format for its self-described room version,
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -66,23 +66,17 @@ of the above:
 
 ### Client-Server API
 
-Using the same "affected" calculation above, when serving `m.room.create` in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
-to clients in affected room versions, servers MUST format that event in the same way it would for
-[`GET /rooms/:roomId/event/:eventId`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3roomsroomideventeventid).
+A prior iteration of this proposal required that servers format the `m.room.create` event as a full
+client event in affected rooms, however that format did not have a strong use case for deviating from
+the regular [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state) format.
+Instead, this proposal makes *no* changes to the format of `m.room.create` in stripped state over the
+Client-Server API, meaning the `m.room.create` event is represented as a regular stripped state event.
+For history on this change, see [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4311/files#r2232855570).
 
-For unaffected room versions, servers SHOULD NOT format the create event, if known, any different than
-what stripped state already requires. This is to ensure clients do not accidentally rely on information
-which may be confusing to them, such as the "creator server name". This is especially true because the
-server is required to treat those events as untrusted for the reasons described in MSC4291.
-
-**Note**: "serving to clients" includes [pushing to Application Services](https://spec.matrix.org/v1.15/application-service-api/#pushing-events).
-
+**Note**: Servers can technically return more than specified in stripped state if they like, though
+this is not usually recommended as clients may unintentionally rely on that behaviour.
 
 ## Potential issues
-
-* Not using the same format for both affected and unaffected room versions may lead to parser complexity.
-  This is considered tolerable by this proposal to ensure that clients (and servers) have limited
-  ability to confuse themselves.
 
 * This technique is not applied to other state events present in stripped state. A future MSC or
   series of MSCs is expected to address this particular concern.
@@ -91,11 +85,6 @@ server is required to treat those events as untrusted for the reasons described 
   the Federation API and Client-Server API. Such implementations are encouraged to make the create
   event reasonably available in its full form to those applications. Typically, this will be an internal
   representation of the event which still has the capability to serialize down to a PDU.
-
-* For unaffected room versions, it's a little awkward that the Federation API SHOULD use PDU format,
-  but the Client-Server API SHOULD NOT format the event as a real event. This proposal makes this
-  choice to give *some* amount of information to intermediary tooling within the server, but can
-  easily be modified to be a "SHOULD NOT" for both cases.
 
 
 ## Alternatives
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -48,6 +48,9 @@ standard Matrix error (new to the endpoint). For knocks, the server SHOULD remov
 Ideally, the server would be able to prevent the knock from happening, though by the time the server
 can see the `knock_room_state`, the knock has already happened.
 
+**Note**: Servers SHOULD consider their local ecosystems before imposing this validation completely,
+per the "Migration" section later in this document.
+
 The `400 M_MISSING_PARAM` error SHOULD be translated to a 5xx error by the sending server over the
 Client-Server API. This is done because there's nothing the client can materially do differently to
 make the request succeed.
@@ -87,25 +90,13 @@ affecting client or server implementations.
 
 ## Migration
 
-*This section is for server implementations to follow upon release of room version 12. It is not
-intended to enter the spec in any way.*
+Mentioned above, existing server implementations SHOULD warn rather than fail on invites which don't
+have complete PDUs inside `invite_room_state` until their local ecosystem adoption allows for the
+full set of validation to be applied. If PDUs are complete, but for a different room, the invite MUST
+still fail per the validation above.
 
-Room version 12 contains MSC4291 and is expected to be used in production prior to this proposal
-becoming stable itself. To account for this, for 1 spec release cycle, servers are encouraged to
-warn rather than fail on invites which don't have a complete `m.room.create` PDU in the `invite_room_state`.
-If the PDU is complete, but for a different room, the invite should still fail per the proposal text
-above.
-
-This translates to a timeline anywhere between 2 and 6 months, depending on ecosystem rollout. An
-example *possible* release schedule is:
-
-1. August 2025 - Matrix 1.16 is released with Room Version 12 and this proposal; servers log warnings
-   about invites missing complete `m.room.create` PDUs.
-2. October 2025 - Matrix 1.17 is released; servers stop using warnings and instead fully apply the
-   validation logic of this proposal, causing invites missing full create event PDUs to fail.
-
-Servers are encouraged to make the switch to full validation early if their ecosystem conditions
-allow. For example, if logged warnings are sufficiently low or of insignicant consequence.
+This proposal suggests that servers wait no longer than 3 months (or about 1 full spec release cycle)
+after this proposal is released to enforce the full validation.
 
 
 ## Dependencies
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -1,0 +1,116 @@
+# MSC4311: Ensuring the create event is available on invites and knocks
+
+Historically, when processing an invite or knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
+originally created the room. If that server was considered abusive, the invite/knock may be rejected
+or blocked early by the tooling. This approach is preferred because the user sending the invite may
+not be on the same server as the user who created the room, but is often still checked by safety
+tooling.
+
+With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
+domain component, which can make the first check described above harder if not impossible. When
+speaking over federation however, [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [knocks](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+contain a concept of "stripped state events" which give (unsigned) context about the room for the
+consuming clients. The Client-Server API [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+section has more information about what these 'events' are and how they work.
+
+This MSC aims to reintroduce the accessibility of the creator's server name to clients (and safety
+tooling) by requiring the `m.room.create` event *at least* be a full and proper PDU in affected room
+versions when transiting the Federation API via stripped state, allowing intermediary servers and
+clients to inspect the creator's domain name.
+
+
+## Proposal
+
+This proposal is split into the Federation API changes and Client-Server API changes for clarity.
+
+
+### Federation API
+
+For room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
+the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+and MUST be a properly-formatted PDU according to that room version's event format specification.
+
+If the `m.room.create` event is not present, not a PDU, or not for the room ID specified, the server
+MUST fail to continue processing the invite or knock. For invites, this is a `400 M_MISSING_PARAM`
+standard Matrix error (new to the endpoint). For knocks, this means the server drops the `make_knock`
+response and never completes a `send_knock`. For both operations, the associated Client-Server API
+request is failed with `500 M_BAD_STATE`. A 4xx error isn't used for the Client-Server API because
+there's nothing the client can materially do differently to fix that request.
+
+For room versions *not* affected by MSC4291, servers SHOULD include the properly-formatted `m.room.create`
+PDU. This is not made mandatory to avoid a situation where servers trust data that shouldn't be trusted
+for the reasons described by MSC4291.
+
+To determine whether a room is "affected" by MSC4291, servers MUST inspect the `room_id` rather than
+the create event's `room_version`. Specifically, a room ID with a domain component is *not* affected
+while one without a domain component (and happens to be `!<43 unpadded urlsafe base64 chars>`) *is*
+affected. This is done to ensure the server is not potentially confused by a malicious server providing
+a create event for a different, unaffected, room.
+
+When a room is affected, the server MUST validate the `m.room.create` event as follows:
+
+1. If the event has a `room_id`, reject.
+2. If the event does not otherwise comply with the event format for its self-described room version,
+   reject.
+3. If the event fails [signature checks](https://spec.matrix.org/v1.15/server-server-api/#validating-hashes-and-signatures-on-received-events),
+   reject. The content hash check MAY be skipped as the event can safely be redacted prior to all of
+   these checks.
+4. If the event's [reference hash](https://spec.matrix.org/v1.15/server-server-api/#calculating-the-reference-hash-for-an-event)
+   does not match the event ID contained in the room ID, reject.
+5. Otherwise, allow.
+
+
+### Client-Server API
+
+Using the same "affected" calculation above, when serving `m.room.create` in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+to clients in affected room versions, servers MUST format that event in the same way it would for
+[`GET /rooms/:roomId/event/:eventId`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3roomsroomideventeventid).
+
+For unaffected room versions, servers SHOULD NOT format the create event, if known, any different than
+what stripped state already requires. This is to ensure clients do not accidentally rely on information
+which may be confusing to them, such as the "creator server name". This is especially true because the
+server is required to treat those events as untrusted for the reasons described in MSC4291.
+
+**Note**: "serving to clients" includes [pushing to Application Services](https://spec.matrix.org/v1.15/application-service-api/#pushing-events).
+
+
+## Potential issues
+
+* Not using the same format for both affected and unaffected room versions may lead to parser complexity.
+  This is considered tolerable by this proposal to ensure that clients (and servers) have limited
+  ability to confuse themselves.
+
+* This technique is not applied to other state events present in stripped state. A future MSC or
+  series of MSCs is expected to address this particular concern.
+
+* Some server implementations allow safety tooling and other applications to hook into them between
+  the Federation API and Client-Server API. Such implementations are encouraged to make the create
+  event reasonably available in its full form to those applications. Typically, this will be an internal
+  representation of the event which still has the capability to serialize down to a PDU.
+
+
+## Alternatives
+
+This proposal fills a potential gap in information created by MSC4291, making the alternatives roughly
+equivalent to "don't do this".
+
+
+## Security considerations
+
+Security considerations are made throughout, especially in areas where a server may accidentally trust
+data it shouldn't.
+
+
+## Unstable prefix
+
+This proposal does not require an unstable prefix as the behaviour can be accomplished without overly
+affecting client or server implementations.
+
+
+## Dependencies
+
+This proposal requires [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291) in
+order to make any amount of sense.
--- a/proposals/4311-stripped-state-create-event.md
+++ b/proposals/4311-stripped-state-create-event.md
@@ -4,20 +4,24 @@ Historically, when processing an invite or knock, safety tooling would parse the
 [being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
 originally created the room. If that server was considered abusive, the incoming invite or outbound
 knock may be rejected or blocked early by the tooling. This approach is preferred because the user
-sending the invite may not be on the same server as the user who created the room, but is often still
-checked by safety tooling.
+sending the invite may not be on the same server as the user who created the room, though both sender
+and creator are checked by safety tooling.
 
 With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
 domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
 recommending rather than requiring the `m.room.create` event, makes the above check harder if not
-impossible in some cases.
+impossible when the create event is missing or incomplete, as the room ID cannot be confirmed in
+MSC4291+ room versions.
 
 This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
 to ensure the event matches the room. To support the new validation, the `m.room.create` event must
 be formatted as a full PDU in the stripped state of [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
-and [knocks](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
-over federation. Together, these changes allow safety tooling (and servers) to better validate invites
-in particular.
+over federation. This is not applied to other events for reasons described later in the proposal.
+
+[Knocks](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid)
+additionally include the full PDU format, though only to ensure symmetry between the two instances of
+stripped state. It's not possible to prevent a knock based on stripped state because the server will
+have already sent the knock before stripped state is received.
 
 
 ## Proposal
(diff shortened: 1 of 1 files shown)


proposals/4311-stripped-state-create-event.md
turt2live, line 1: Implementation requirements:

* [x] Server (returning stripped state)
* [x] Server (receiving stripped state) - See [comment below](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2263728366)
* [x] Client (receiving stripped state)

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,141 @@
...
+  event reasonably available in its full form to those applications. Typically, this will be an internal
+  representation of the event which still has the capability to serialize down to a PDU.
+
+* For unaffected room versions, it's a little awkward that the Federation API SHOULD use PDU format,
+  but the Client-Server API SHOULD NOT format the event as a real event. This proposal makes this
+  choice to give *some* amount of information to intermediary tooling within the server, but can
+  easily be modified to be a "SHOULD NOT" for both cases.
+
+
+## Alternatives
```
clokep, line 75: Is an alternative to the migration approach to rev the API version? I'm unsure it'll help but feels like changing an API is not great.

turt2live: I'm not sure it's an alternative in practice: a version bump on the endpoints would mean the information gap still exists, so servers/safety tooling would just reject all old-style invites instead of trying to let them through. It'd functionally be the same as this MSC's "must validate" clauses.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,141 @@
...
+4. If the event's [reference hash](https://spec.matrix.org/v1.15/server-server-api/#calculating-the-reference-hash-for-an-event)
+   does not match the event ID contained in the room ID, reject.
+5. Otherwise, allow.
+
+
+### Client-Server API
+
+Using the same "affected" calculation above, when serving `m.room.create` in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+to clients in affected room versions, servers MUST format that event in the same way it would for
+[`GET /rooms/:roomId/event/:eventId`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3roomsroomideventeventid).
```
Kladki: Reading the MSC, it's not clear to me _why_ the create event should use the sync event format, instead of the stripped format. What benefit does it provide?

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,141 @@
+# MSC4311: Ensuring the create event is available on invites and knocks
+
+Historically, when processing an invite or knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
+originally created the room. If that server was considered abusive, the invite/knock may be rejected
+or blocked early by the tooling. This approach is preferred because the user sending the invite may
+not be on the same server as the user who created the room, but is often still checked by safety
+tooling.
```
Kladki: Why would this be needed for knocks (i.e. what's the benefit)? The server receiving the knock already should have the full create event, so it's not needed for them. It's also not clear why a server would reject it's own knock if it didn't receive a correct create event.

proposals/4311-stripped-state-create-event.md
TheArcaneBrony: The entire point of the MSC is to make it be a full event, especially in order to have some assurance that the information within (most importantly creator/s) is valid, ie for moderation tooling which may block invites based on room origin servers.

Kladki: Yes, but why does that need the format of the event in `/sync` to change? From what I can tell, this only needs the federation API changes.

proposals/4311-stripped-state-create-event.md
TheArcaneBrony: I'm not 100% sure on how knocking works, but I think it's more useful to the server that's sending the knock?

proposals/4311-stripped-state-create-event.md
tulir: The room creation time might be nice to have, although I can't immediately think of use cases like I can for the invite event itself (https://github.com/matrix-org/matrix-spec/issues/2138).

The only other benefit I can think of is that it could be used as an indicator for whether the server validated the create event or not, but it might not work reliably for that use case (in case some servers pass through too much information from the remote server currently)

turt2live: This MSC is primarily intended for moderation tooling. Some tooling may only exist on the client side after a sync has completed, meaning that tooling needs more information about the event than normal (especially in v12 rooms).

It's also just reciprocal: if it's required over federation, then the server can largely pass it straight through instead of stripping it way down.

Kladki: > meaning that tooling needs more information about the event than normal (especially in v12 rooms).

Not really sure how `origin_server_ts` or `unsigned` would be any more helpful in v12 than other room versions, but I guess the former could be useful if, for example, you wanted to check if many rooms were created in a short burst around the same time.

I think this should be explicitly stated in the MSC text.

> It's also just reciprocal: if it's required over federation, then the server can largely pass it straight through instead of stripping it way down.

I think this is a pretty solid reason, and should be mentioned in the MSC text.

proposals/4311-stripped-state-create-event.md
Kladki: https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2233053282

> It's also just reciprocal: if it's required over federation, then the server can largely pass it straight through instead of stripping it way down.

Is also relevant here.

proposals/4311-stripped-state-create-event.md
erikjohnston: I'd also like some expansion on what extra data clients would gain from this? So long as the create event is verified by the server, the client can trust the creator, etc.

Generally, I'm not a fan of having different formats for different items in a list as it can lead to a confusing and muddled API.

proposals/4311-stripped-state-create-event.md
clokep: Right, I guess it would have to be paired with "version 12 rooms require v3 of the endpoint" or something, which seems odd.

proposals/4311-stripped-state-create-event.md
tulir: > Generally, I'm not a fan of having different formats for different items in a list as it can lead to a confusing and muddled API.

There already needs to be a different format for the user's member event (which is in the same list as stripped state in /sync) so that clients can tell when an invite was sent (https://github.com/matrix-org/matrix-spec/pull/2181)

Kladki: Is there a reason the `/state` format is used, instead of the regular sync event format (event without room ID)?

turt2live: Because it's possible for the events to appear outside of sync.

proposals/4311-stripped-state-create-event.md
turt2live: Ruma (server; sending): 
* https://github.com/ruma/ruma/pull/2171
* https://github.com/ruma/ruma/pull/2173

tulir: Meowlnir (client/bot replacing room ID parsing with creator parsing in invite antispam): https://github.com/maunium/meowlnir/commit/eeccd4aef99d53ceed767f29c3bd00c65211183d

proposals/4311-stripped-state-create-event.md
turt2live: so, I don't actually have a good reason to use a full event format here. Reciprocating the format is nice from a courtesy point of view, but that's hardly a strong technical justification. Similarly, including the `origin_server_ts` may just be something we do to all stripped state rather than on a case-by-case basis to achieve uniformity on the format of events in the list (this is something I'll leave on the new-since-this-thread MSC4319). Finally, as mentioned above, this isn't a good indicator for server support if the server passes through the data. As it turns out, servers already do pass through the data, so it's *definitely* not effective as an indicator.

I'm going to change the MSC to use the normal stripped state format on the Client-Server API - if someone would like to advocate for using the full format, please leave a comment on the new diff once it's landed :)

turt2live: (implicitly resolved per https://github.com/matrix-org/matrix-spec-proposals/pull/4311/files#r2232855570 )

proposals/4311-stripped-state-create-event.md
turt2live: clarified in the MSC that this is the outbound knock, not inbound knock.

proposals/4311-stripped-state-create-event.md
turt2live: Though servers haven't actually added the validation on the stripped state yet, per the migration section of the MSC, no one has indicated that it's a nightmare to do so. I'm inclined to consider that "implementation" for the purposes of starting FCP, but welcome (particularly SCT) opinions on whether that's valid.

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
Though I've just pushed a refactor of this MSC, the vast majority has had healthy discussion prior and appears to have settled and should therefore be ready for FCP. There's an open question on the implementation requirements, but I hope to at least collect tickyboxes until implementation can be drafted (possibly next week).

@mscbot fcp merge

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+recommending rather than requiring the `m.room.create` event, makes the above check harder if not
+impossible in some cases.
+
+This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
+to ensure the event matches the room. To support the new validation, the `m.room.create` event must
+be formatted as a full PDU in the stripped state of [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [knocks](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+over federation. Together, these changes allow safety tooling (and servers) to better validate invites
+in particular.
```
TheArcaneBrony: This wording adds a couple points of confusion:
- Will the full PDU be available over the CS-API? The wording kind of implies that it won't be ("must be formatted as a full PDU ... over federation")
- Over the CS-API, does this imply the regular client event format or the federation event format?

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+
+For room versions *not* affected by MSC4291, servers SHOULD include the properly-formatted `m.room.create`
+PDU. This is not made mandatory to avoid a situation where servers trust data that shouldn't be trusted
+for the reasons described by MSC4291.
+
+To determine whether a room is "affected" by MSC4291, servers MUST inspect the `room_id` rather than
+the create event's `room_version`. Specifically, a room ID with a domain component is *not* affected
+while one without a domain component (and happens to be `!<43 unpadded urlsafe base64 chars>`) *is*
+affected. This is done to ensure the server is not potentially confused by a malicious server providing
+a create event for a different, unaffected, room.
```
TheArcaneBrony: This sounds pretty unsafe to me. Couldn't you validate the room ID by calculating it on the create event?

proposals/4311-stripped-state-create-event.md
turt2live: The intro is generally less precise than the proposal text - it's a preview of the changes, not duplicate.

I believe this is addressed in the proposal.

proposals/4311-stripped-state-create-event.md
turt2live: That's what this is saying. In v12 rooms there is no `room_id` on the events, so the server needs to calculate it. The risk is if the server believes the `room_version` field and assumes it's an affected room.

proposals/4311-stripped-state-create-event.md
Kladki: So wait, servers are supposed to ignore their own knock if the resident server doesn't return a valid create event? The sender user _wants_ to knock on the room, so I don't see why it should ever reject it's own knock.

turt2live: If the server knocked on a room and got a different create event, it's not knocking on the room it thought it was.

Kladki: Sure, but I don't see why it would need a full create event PDU to be sent to the knocking server, since the T&S reasoning doesn't really apply here.

turt2live: It applies less, but it's easier to specify changes to both stripped state areas than just one.

proposals/4311-stripped-state-create-event.md
Kladki: Server returning and receiving: https://gitlab.com/famedly/conduit/-/commit/532b17ade86b4373db9f3bcca6af6815dd9b5b10

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
+# MSC4311: Ensuring the create event is available on invites and knocks
+
+Historically, when processing an invite or knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
+originally created the room. If that server was considered abusive, the incoming invite or outbound
+knock may be rejected or blocked early by the tooling. This approach is preferred because the user
+sending the invite may not be on the same server as the user who created the room, but is often still
+checked by safety tooling.
```
dbkr: I don't think this is clear about what is "still checked by safety tooling".

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+
+With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
+domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+recommending rather than requiring the `m.room.create` event, makes the above check harder if not
+impossible in some cases.
```
dbkr: There's clearly a lot unsaid behind these words around exactly why it might be possible in some cases and I appreciate trying to make the MSC shorter but I think in this case I'm finding the omission of it quite confusing.

proposals/4311-stripped-state-create-event.md
clokep: That doesn't really match what this paragraph says -- "inspect" is the wrong word if you're calculating the room ID from the crate event for comparison to what's sent over the APIs. Is that even what this is saying? I'm unsure.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Proposal
+
+On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
+where available. No other changes are proposed to the Client-Server API.
+
+Over federation, for room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
+the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+and MUST be a properly-formatted PDU according to that room version's event format specification.
```
clokep: I'm failing to understand why it must be the full PDU now? Is that so the room ID can be calculated? Isn't a stripped create event essentially the full event anyway since nothing gets redacted?

(I don't really have an issue with it being the full, event just the rationale is not clear in here.)

richvdh: +1. It feels really confusing if we now have some "Stripped state events" which are stripped, and some which are not. If we want to change things so that stripped state events are not stripped, then let's do it to *all* the events, and find a different name for them.

(https://spec.matrix.org/v1.15/client-server-api/#stripped-state defines "Stripped state event" to be "A stripped down state event, with only the `type`, `state_key`, `sender`, and `content` keys")

tulir: Yes, the full PDU is required to be able to calculate the room ID. [Stripped state](https://spec.matrix.org/v1.15/client-server-api/#stripped-state) is not the same as redacted events. Stripped state removes all keys except `type`, `sender`, `state_key` and `content` regardless of any redaction rules, but it doesn't remove anything in content.

richvdh: I guess the reason we need it to be a full PDU is so that the receiving server can correctly calculate the event hash and make sure it matches the `room_id`, to check that the sending server isn't playing shenanigans. But again, it feels like some of those checks should apply to all the events in `invite_room_state` and `knock_room_state`, so we should probably make this change to all of them. (Do we know why it was specced as "stripped" in the first place?)

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+becoming stable itself. To account for this, servers SHOULD treat "MUST" as "MAY" throughout this
+proposal, with the exception of the Client-Server API changes, until 1 full spec release cycle has
+passed since this MSC's own release in the specification.
+
+This translates to a timeline anywhere between 2 and 6 months, depending on ecosystem rollout. An
+example *possible* release schedule is:
+
+1. August 2025 - Matrix 1.16 is released with Room Version 12.
+2. October 2025 - Matrix 1.17 is released with this proposal; servers use "MAY" keywords.
+3. January 2026 - Matrix 1.18 is released; servers switch to "MUST" keywords.
```
richvdh: seems like you are hoping to get this into 1.16, so a more likely schedule is maybe:

```suggestion
1. August 2025 - Matrix 1.16 is released with Room Version 12; servers use "MAY" keywords.
2. October 2025 - Matrix 1.17 is released; servers switch to "MUST" keywords.
```

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+
+Servers MAY switch to "MUST" keywords early if their local ecosystems are prepared for the change.
```
richvdh: This feels like a hugely confusing way of presenting this. 

> 'SHOULD treat "MUST" as "MAY"'
> 'MAY switch to "MUST"

:exploding_head: 

(also, I think you want some of the MUSTs to be SHOULDs rather than MAYs for the transition period)

Also also, I don't think we've ever had a single MSC that requires changes in two spec versions, and I don't think we have great tooling to manage it.

How about changing the MUSTs to SHOULDs or MAYs, and then following up with a second MSC which changes them all to MUSTs?

richvdh (changes requested):

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+
+With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
+domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+recommending rather than requiring the `m.room.create` event, makes the above check harder if not
+impossible in some cases.
+
+This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
+to ensure the event matches the room. To support the new validation, the `m.room.create` event must
+be formatted as a full PDU in the stripped state of [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [knocks](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
```
dbkr: Did you mean to link to make_knock - looks like only send_knock returns stripped state, although even then I'm not sure how much use it would be since the user would already have knocked on the room by the time it arrived on their HS to check it. Best to be consistent though, I suppose.

proposals/4311-stripped-state-create-event.md
tulir: > But again, it feels like some of those checks should apply to all the events in invite_room_state and knock_room_state, so we should probably make this change to all of them

Making all the events verifiable would be nice, but is much more complicated: you'd need to ship the entire auth chain to be able to prove that the events are allowed, and even then there's no way to prove you didn't omit events. I think the idea is that since proper verification is too complicated, it's better to just explicitly mark them as untrusted

The create event alone is relatively trivial though, plus it restores the verifiability of the room creator server which was possible pre-v12. In old rooms, to fake the creator server you'd need to fake the room ID, but if you faked the room ID, the invite would be broken/unjoinable. In new rooms the creator server isn't in the room ID anymore, so all you need to do is fake the create event. This MSC prevents faking the create event to bring v12+ back on par with pre-v12 in that specific context.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
+where available. No other changes are proposed to the Client-Server API.
+
+Over federation, for room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
+the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#get_matrixfederationv1make_knockroomiduserid)
+and MUST be a properly-formatted PDU according to that room version's event format specification.
+
+If the `m.room.create` event is not present, not a PDU, or not for the room ID specified, the server
+MUST fail to continue processing the invite or knock. For invites, this is a `400 M_MISSING_PARAM`
```
dbkr: Grammar nit, but "fail to continue" sounds weird to me. Maybe say the invite/send_knock request must fail?

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,116 @@
...
+standard Matrix error (new to the endpoint). For knocks, this means the server drops the `make_knock`
```
dbkr: As above, I don't see how this affects the make_knock endpoint as it doesn't involve stripped state, as far as I can see.

proposals/4311-stripped-state-create-event.md
turt2live: what tulir said^

I'm less confident that we can ship "all events are no longer stripped" quickly, but recognize that is the direction we want to go in. There's also other (undisclosed) work in the area which will make it easier to provide stripped state as full PDUs in the future.

proposals/4311-stripped-state-create-event.md
turt2live: This section will not be added to the spec in any capacity, per the first sentence. It's an unstable->stable migration, where "unstable" is a bit longer than usual.

proposals/4311-stripped-state-create-event.md
turt2live: edit: copy/paste fail.

hmm, true. For some reason I ended up with the wrong link. The send_knock indeed makes the knocking changes questionable.

turt2live: 5ac4656775f357fd5ceb4fb92e16aa19f0baaab9

proposals/4311-stripped-state-create-event.md
turt2live: 5ac4656775f357fd5ceb4fb92e16aa19f0baaab9

proposals/4311-stripped-state-create-event.md
turt2live: 5ac4656775f357fd5ceb4fb92e16aa19f0baaab9

proposals/4311-stripped-state-create-event.md
turt2live: hopefully addressed in e651b619ddaeebccf366b82e456f0f55dacd4205

proposals/4311-stripped-state-create-event.md
turt2live: afff25ed72eb7b71ee04cbdc22de86c5d4f21114

proposals/4311-stripped-state-create-event.md
turt2live: afff25ed72eb7b71ee04cbdc22de86c5d4f21114

proposals/4311-stripped-state-create-event.md
turt2live: 72f6a3a00b85b843dbf8cb20d71c9d36da938c2b

proposals/4311-stripped-state-create-event.md
turt2live: bef65403ca2db18160334652a8ec81034737ebc0

proposals/4311-stripped-state-create-event.md
turt2live: I've reworked the section to hopefully be a bit clearer, and avoid uppercase spec-like keywords: bef65403ca2db18160334652a8ec81034737ebc0

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,131 @@
...
+
+On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
+where available. No other changes are proposed to the Client-Server API.
+
+Over federation, for room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
+the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid)
+and MUST be a properly-formatted PDU according to that room version's event format specification. The
+full PDU format is used to ensure that receiving applications can independently verify the room ID
+by calculating the reference hash of the create event themselves.
```
richvdh: Following on from the [previous thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2269526101):

Whether or not recipients can make use of the full PDU for other event types, I don't really understand why we don't require that senders send it today? As well as being more consistent, it seems like it would reduce migration pain for the future.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+## Proposal
+
+On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
+where available. No other changes are proposed to the Client-Server API.
+
+Over federation, for room versions affected by [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291),
+the `m.room.create` event MUST be included in [`invite_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid)
```
richvdh: Why are we only applying this to MSC4291 room versions? Wouldn't it be simpler all round to require it for all rooms? Again, the fact it is there doesn't mean that the recipient has to use it.

richvdh (changes requested):

richvdh:
@mscbot concern inconsistency between event types feels like an unnecessary complication

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,131 @@
...
+[Knocks](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid)
+additionally include the full PDU format, though only to ensure symmetry between the two instances of
+stripped state. It's not possible to prevent a knock based on stripped state because the server will
+have already sent the knock before stripped state is received.
+
+
+## Proposal
+
+On the Client-Server API, `m.room.create` MUST be provided in [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state),
+where available. No other changes are proposed to the Client-Server API.
```
richvdh: so, to be clear, under the proposal as it stands, `m.room.create`, would continue to be in the "stripped" format in the CS-API?

proposals/4311-stripped-state-create-event.md
erikjohnston: +1 to not special casing some events and not other. Such special casing/exceptions makes the API cognitively harder to reason about and implement correctly. I don't see any reason why it'd be easier to just do this for create events rather than all events.

proposals/4311-stripped-state-create-event.md
turt2live: yes.

proposals/4311-stripped-state-create-event.md
turt2live: I really don't think we should block the MSC on this. If we send the full PDU, we're saying that it's possible for the receiver to actually validate the data, or worse, implying that it's trusted and requires no verification. To fix this we'd need to send the entire `auth_events` chain for all events, and even then there's *zero* guarantee that the auth chain is complete or current to the invite. This will lead to implementations assuming that the data can be relied upon when in fact it's still highly untrusted. By having "stripped state" be a subset of fields, impossible to verify, we're at least making it clearer to implementations that the data is untrusted and can be falsified.

Solving the above concerns is very possible, but well beyond the scope of this MSC. This MSC deliberately represents the minimum changes required to patch a security issue introduced by room version 12 before v12 ships in the spec. 

Would folks be comfortable letting this MSC continue through the process if a second MSC was published aiming to address the consistency and auth chain concerns?

proposals/4311-stripped-state-create-event.md
turt2live: answer implied by https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2274302179 - the create event cannot be trusted in pre-4291 rooms and therefore should not be presented as trustworthy.

proposals/4311-stripped-state-create-event.md
turt2live: For clarity: I also want to make the format consistent for all events in the list. It's indeed far from great to have two formats, but I also *really* don't want to delay the spec release behind a technical design which needs significant thought (or worse, is rushed through and hard to fix later).

proposals/4311-stripped-state-create-event.md
richvdh: I don't really agree. Most of the stuff in `invite_room_state` is "not trustworthy". We're not proposing removing it though.

proposals/4311-stripped-state-create-event.md
richvdh: I had a long conversation with Travis out of band about this. As I understand it, his concerns boil down to: by requiring homeservers to provide full PDUs in `invite_room_state`, then implementations may start to consider the state in `invite_room_state` to have been validated, despite admonitions in the spec against that.

Affected implementations could include:
 * Homeservers that have received `invite_room_state`
 * Clients that receive an invite
 * Tooling that interacts with homeservers via implementation-specific interfaces. (For example, some tools look at `unsigned.invite_room_state` in events in Synapse)

To guard against that, Travis believes that, if the full PDUs are there, then the receiving homeserver must validate them. That would require the auth chains to be provided, as well as requiring more substantial changes on the implementation side.

I believe the concern is moot for each of those three classes of implementations, for different reasons:

 * Somebody implementing a homeserver should be in no doubt as to which events have been validated and which have not. 	Synapse, for example, deliberately keeps `invite_room_state` in a very different place to normal events, because it hasn't been validated.

 * Clients are unaffected by this change, since (as clarified in the other thread), events sent over the C-S API will continue to be stripped. There is a "slippery slope" argument here, in which we might forget this discussion, and make the same change in the C-S API, and then clients *will* be affected; but particularly given how strongly this change has been argued I find it very unlikely such a change will slip in under the radar.

 * Tooling that uses implementation-specific interfaces seems to me to be somewhat out of scope for design of the Matrix protocol: rather, it is up to the designers of those implementation-specific interfaces to provide appropriate guardrails. As a concrete example: Synapse could *strip* incoming PDUs in `invite_room_state` *before* storing it in `unsigned`, so that it would be clear to downstream tooling that it has not been validated. I don't think we need to design the federation protocol around this.

   Further: the risk of tooling misinterpreting events in `unsigned.invite_room_state` as verified exists today. Synapse provides very few guarantees about what is there, and it would not be hard for a malicious or mischievous homeserver operator to make an `/invite` request with `invite_room_state` that *looks* like full PDUs (thus tricking tooling), but in fact being bogus. Admittedly that concern grows once it becomes the norm rather than the exception for `invite_room_state` to contain full PDUs, but still, if "tools might misinterpret `unsigned.invite_room_state`" is a genuine concern, then it seems like the solution is better guardrails to Synapse irrespective of this MSC.

richvdh: To conclude: I still feel like Travis's concerns about events "looking trusted" are outweighed by my concerns around (a) inconsistency between event types and (b) [inconsistency between room versions](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2273340163).

turt2live: I spent some time since that discussion drafting up an alternative which didn't use stripped state, primarily to see what the changes might look like. In the end, the changes would have similar misinterpretation concerns (if not worse due to data duplication). Because of this, I'm more inclined to agree that we should just make the jump to using full PDUs for all stripped state, though I still hold the misinterpretation concerns. 

For those interested, I published the alternative at https://github.com/matrix-org/matrix-spec-proposals/pull/4329

proposals/4311-stripped-state-create-event.md
turt2live: 2104c97e1c1f5f7241c265ec9171bf762a0e67ec makes it apply to all room versions

proposals/4311-stripped-state-create-event.md
turt2live: 2104c97e1c1f5f7241c265ec9171bf762a0e67ec uses full PDU formats everywhere now

proposals/4311-stripped-state-create-event.md
turt2live: clarified in 2104c97e1c1f5f7241c265ec9171bf762a0e67ec

proposals/4311-stripped-state-create-event.md
turt2live: This MSC now requires all stripped state events to be PDUs over federation. I believe https://github.com/element-hq/synapse/pull/18822 and https://github.com/matrix-org/complement/pull/796 demonstrate it's not impossible, even if the code itself is bad.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,114 @@
...
+With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
+domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+recommending rather than requiring the `m.room.create` event, makes the above check harder if not
+impossible when the create event is missing or incomplete, as the room ID cannot be confirmed in
+MSC4291+ room versions.
+
+This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
+to ensure the event matches the room. To support the new validation, the `m.room.create` event must
+be formatted as a full PDU in the stripped state of [invites](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+over federation. Similar treatment is applied to other stripped state events for uniformity.
```
kegsay, line 20: > Similar treatment is applied to other stripped state events for uniformity.

But why, aside from vdh's general ick feeling of having 2 data types in the same array? By doing it for all events:
 - we add more bandwidth to invites
 - we don't gain any more security as the events can't be verified via their auth chains
 - we risk giving the impression that we are authorising all events in `invite_room_state`, which we ain't.

To fix the ick, I'd rather we moved the create event to be a new key outside of `invite_room_state` and leave `invite_room_state` as-is, although that would either be backwards incompatible (the create event would be missing from `invite_room_state`) or duplicate the create event fields if we keep the create event in `invite_room_state` for a while.

proposals/4311-stripped-state-create-event.md
```diff
@@ -27,73 +27,56 @@ have already sent the knock before stripped state is received.
...
-   reject.
-3. If the event fails [signature checks](https://spec.matrix.org/v1.15/server-server-api/#validating-hashes-and-signatures-on-received-events),
-   reject. The content hash check MAY be skipped as the event can safely be redacted prior to all of
-   these checks.
-4. If the event's [reference hash](https://spec.matrix.org/v1.15/server-server-api/#calculating-the-reference-hash-for-an-event)
-   does not match the event ID contained in the room ID, reject.
-5. Otherwise, allow.
+where available. No other changes are proposed to the Client-Server API. For clarity, this means clients
+continue to receive events which only have `content`, `sender`, `state_key` (optional), and `type` in
+the `invite_room_state`, `knock_room_state`, and whereever else stripped state is used.
```
richvdh: It might be worth emphasising that server implementations will now have to actively remove any other keys on `invite_room_state` received over federation (whereas Synapse currently just passes it through.)

proposals/4311-stripped-state-create-event.md
richvdh: I need to give this a full re-review, but this looks better now, thanks.

proposals/4311-stripped-state-create-event.md
richvdh: For reference: this follows on from two previous threads ([1](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2269526101), [2](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2273294243)) where this was discussed in detail.

> we add more bandwidth to invites

This feels negligible though.

> we don't gain any more security as the events can't be verified via their auth chains

It is not proposed as a security improvement; rather as one that makes the API more consistent. 

> we risk giving the impression that we are authorising all events in invite_room_state, which we ain't.

I really don't buy this argument. The fact that an event is passed over the federation API doesn't say anything to me about whether it is authorised or not.

I'll also re-emphasise that if you stick to the "full PDUs make people think that they are authorised" argument, then we have to do something different here for pre-v12 rooms (see [previous thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2273340163)) which is also gnarly.

kegsay: > This [additional bandwidth] feels negligible though.

To you perhaps, but this is the main reason why we have the concept of stripped state events at all vs sending full-fat PDUs.

I don't feel strongly on this. Having stripped state always be the full-fat PDU seems fine.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,114 @@
+# MSC4311: Ensuring the create event is available on invites and knocks
+
+Historically, when processing an invite or knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
+originally created the room. If that server was considered abusive, the incoming invite or outbound
+knock may be rejected or blocked early by the tooling. This approach is preferred because the user
```
richvdh: Given we're not fixing knocks here, I wonder if it would be clearer to remove the references to knocks in the introductory paragraph and instead update the later paragraph to say "Note that the problem also exists for knocks but we can't fix it".

Very light suggestion, feel free to ignore.

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,114 @@
...
+and [`knock_room_state`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1send_knockroomideventid).
+Servers MUST additionally format events in `invite_room_state` and `knock_room_state` as PDUs according
+to that room version's event format specification. Together, these changes allow servers to validate
+the room ID matches the invite (or knock, though it's already sent by the time validation would happen).
+
+Specifically, including the `m.room.create` event as a full PDU allows servers to calculate the room
+ID by hashing the event in MSC4291+ room versions. For other room versions (1 through 11), the server
+can at most compare the `room_id` field of the create event with the invite/knock membership event.
+
+If any of the events are not a PDU, not for the room ID specified, or fail [signature checks](https://spec.matrix.org/v1.15/server-server-api/#validating-hashes-and-signatures-on-received-events),
```
richvdh, line 45: I remain rather uncomfortable that, taken as written, this breaks compatibility with all existing servers. I know you have some words in "Migration" about this, but I think it's insufficient given "It is not intended to enter the spec"; I think if you want to mandate that servers enforce full PDUs, we need something to end up in the spec about this.

Can we at least soften the language here to say something along the lines of "servers must consider their ecosystem and consider softening this check, especially for older room versions"?

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,114 @@
...
+This translates to a timeline anywhere between 2 and 6 months, depending on ecosystem rollout. An
+example *possible* release schedule is:
+
+1. August 2025 - Matrix 1.16 is released with Room Version 12 and this proposal; servers log warnings
+   about invites missing complete `m.room.create` PDUs.
+2. October 2025 - Matrix 1.17 is released; servers stop using warnings and instead fully apply the
+   validation logic of this proposal, causing invites missing full create event PDUs to fail.
+
+Servers are encouraged to make the switch to full validation early if their ecosystem conditions
+allow. For example, if logged warnings are sufficiently low or of insignicant consequence.
```
richvdh: Given most of the MSC no longer talks about room v12, the heavy emphasis in v12 here feels out of place

richvdh (changes requested):

richvdh:
@mscbot resolve inconsistency between event types feels like an unnecessary complication

proposals/4311-stripped-state-create-event.md
turt2live: I'd prefer to keep it in because a few folks have left comments in the past about being surprised by the knocks mention later in the proposal. Suggestions to make it less prominent in the intro are welcome.

proposals/4311-stripped-state-create-event.md
turt2live: 249dec303c14e3bb1ddc4046d2c2a5d409f2603d

proposals/4311-stripped-state-create-event.md
turt2live: 249dec303c14e3bb1ddc4046d2c2a5d409f2603d

proposals/4311-stripped-state-create-event.md
turt2live: I think this was done in one of the commits.

richvdh (approved):
thanks

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,105 @@
+# MSC4311: Ensuring the create event is available on invites and knocks
```
richvdh:
```suggestion
# MSC4311: Ensuring the create event is available on invites
```

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,105 @@
...
+
+Historically, when processing an invite or knock, safety tooling would parse the room ID despite
+[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids) to determine the server which
+originally created the room. If that server was considered abusive, the incoming invite or outbound
+knock may be rejected or blocked early by the tooling. This approach is preferred because the user
+sending the invite may not be on the same server as the user who created the room, though both sender
+and creator are checked by safety tooling.
```
richvdh:
```suggestion
Historically, when processing an incoming invite or outgoing knock, safety tooling would parse the room ID despite
[being opaque](https://spec.matrix.org/v1.15/appendices/#room-ids), to determine the server which
originally created the room. If that server was considered abusive, the invite or 
knock may be rejected or blocked early by the tooling. Note that checking the domain of the
sender of an invite is inadequate, because the sender may not be on the same server as the
user who created the room.
```

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,105 @@
...
+
+With [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291), room IDs lose their
+domain component. This, combined with [Stripped State](https://spec.matrix.org/v1.15/client-server-api/#stripped-state)
+recommending rather than requiring the `m.room.create` event, makes the above check harder if not
+impossible when the create event is missing or incomplete, as the room ID cannot be confirmed in
+MSC4291+ room versions.
+
+This MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
```
richvdh:
```suggestion
To mitigate the problem in the case of invites,
this MSC shifts the `m.room.create` event to a *required* stripped state event, and imposes validation
```

proposals/4311-stripped-state-create-event.md
richvdh: made a few light suggestions

richvdh (commented):
some light editing of the intro

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Migration
+
+Mentioned above, existing server implementations SHOULD warn rather than fail on invites which don't
+have complete PDUs inside `invite_room_state` until their local ecosystem adoption allows for the
+full set of validation to be applied. If PDUs are complete, but for a different room, the invite MUST
+still fail per the validation above.
+
+This proposal suggests that servers wait no longer than 3 months (or about 1 full spec release cycle)
+after this proposal is released to enforce the full validation.
```
clokep: This seems really aggressive. We know people don't update their servers quickly. Should we rev a v3 version to send the full PDU and require using v3 for room version 12?

clokep:
@mscbot concern This changes an API in a backwards incompatible manner

proposals/4311-stripped-state-create-event.md
turt2live: I assume this thread is for the `This changes an API in a backwards incompatible manner` concern. 

I'm failing to see the "incompatible" part of the changes. Servers already do apply validation to invites leading to rejections, so it should be fine to ask that servers add more validation. It's also acceptable to send *more* data than requested on any endpoint.

I don't think bumping the endpoint version will help. https://github.com/matrix-org/matrix-spec-proposals/pull/4329 is an example of doing that, and it's no less awful than the changes here.

clokep: > I'm failing to see the "incompatible" part of the changes. Servers already do apply validation to invites leading to rejections, so it should be fine to ask that servers add more validation. It's also acceptable to send _more_ data than requested on any endpoint.

It is requiring a change for the client (note the client here is still a homeserver since this is a federation API) without changing the endpoint version, so it is incompatible.

turt2live:
FCP is blocked while concerns are raised.

proposals/4311-stripped-state-create-event.md
turt2live: We've spent the last week or so discussing this internally amongst the SCT. There are a variety of options available to us, including bumping endpoint versions, though the ultimately suggested change is:

Servers MAY validate for room versions 1 through 11 and SHOULD validate for 12 and higher. A future MSC may introduce a new endpoint which changes it to MUST (potentially regardless of room version).

clokep (approved):

clokep:
@mscbot resolve This changes an API in a backwards incompatible manner

turt2live:
Concluding FCP now that it's unblocked. Given this MSC has hit a few edge cases in the process, if there are strong objections in the next few days we'll discuss and incorporate into the MSC as needed.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2207

turt2live:
Merged 🎉

proposals/4311-stripped-state-create-event.md
```diff
@@ -0,0 +1,108 @@
...
+ID by hashing the event in MSC4291+ room versions. For other room versions (1 through 11), the server
+can at most compare the `room_id` field of the create event with the invite/knock membership event.
+
+If any of the events are not a PDU, not for the room ID specified, or fail [signature checks](https://spec.matrix.org/v1.15/server-server-api/#validating-hashes-and-signatures-on-received-events),
+or the `m.room.create` event is missing, the receiving server MAY respond to invites with a `400 M_MISSING_PARAM`
+standard Matrix error (new to the endpoint). For invites to room version 12+ rooms, servers SHOULD
+rather than MAY respond to such requests with `400 M_MISSING_PARAM`. For knocks, the server SHOULD remove any events from
+`knock_room_state` which fail the same validation check before passing the details along to clients.
+Ideally, the server would be able to prevent the knock from happening, though by the time the server
+can see the `knock_room_state`, the knock has already happened.
```
MadLittleMods, line 51: Why are knocks handled differently than invites?

Specifically: "For knocks, the server SHOULD remove any events from `knock_room_state` [...] before passing the details along to clients."

Versus invites: "If any of the events are [found invalid] [reject]" (all or nothing)

turt2live: (opening spec clarification issues is best after an MSC is merged - comments can and do go missing, and rarely lead to the spec itself being improved)

---

The desirable property is preventing the action from ever happening, which is only possible on invites. Invites can have their events prevented by refusing to sign the event. Knocks have a different series of endpoints which leads to the knock event already being sent by the time `knock_room_state` has happened, so the MSC forfeits the ability to prevent the event and instead forces removal of invalid data to get as close as possible to the desirable property.

PR #4312, opened 2025-07-15
MSC4312: Resetting cross-signing keys in the OAuth world

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/x-signing-reset-with-nextgen-auth/proposals/4312-x-signing-reset-with-nextgen-auth.md)

Implementations:

- Server
  - [x] Synapse: https://github.com/element-hq/synapse/pull/17509 (July 2024)
  - [x] MAS:
    - https://github.com/matrix-org/matrix-authentication-service/pull/2142
    - https://github.com/matrix-org/matrix-authentication-service/pull/2364
    - https://github.com/matrix-org/matrix-authentication-service/pull/3102
- Client:
  - [x] Element Web (supporting the new UIA stage): https://github.com/element-hq/matrix-react-sdk/pull/34
  - [x] Example of client that doesn't support the UIA stage working using the fallback:

https://github.com/user-attachments/assets/b7b42b9b-9963-46c1-9b7d-6fef4b7218c1

----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4312#issuecomment-3330366883)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4312#issuecomment-3330352715)

### DIFF
--- a/proposals/4312-x-signing-reset-with-nextgen-auth.md
+++ b/proposals/4312-x-signing-reset-with-nextgen-auth.md
@@ -20,8 +20,8 @@ documents these workarounds as a low-effort interim workaround until better solu
 
 Clients that have authenticated via the new [OAuth APIs] continue to use
 [`/_matrix/client/v3/keys/device_signing/upload`] to replace cross-signing keys. Homeservers
-continue to enforce UIA on the endpoint but MUST only use a single stage `m.oauth`[^1] together with
-a URL that points to the authorization server's account management UI.
+continue to enforce UIA on the endpoint with a flow containing a single stage `m.oauth`[^1] together
+with a URL that points to the authorization server's account management UI.
 
 ``` json5
 {


proposals/4312-x-signing-reset-with-nextgen-auth.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Unsupporting client
* [ ] Supporting client

Johennes: Server: https://github.com/element-hq/synapse/pull/17509 (July 2024)
Supporting client: https://github.com/element-hq/matrix-react-sdk/pull/34 (September 2024)

I'm not sure how an unsupporting client would be implemented. They'd simply use the UIA fallback mechanism without any changes needed?

hughns:
MAS lists the `org.matrix.cross_signing_reset` action in the `account_management_actions_supported` from MSC4191. e.g. in https://account.matrix.org/.well-known/openid-configuration

Does this need to be added to this MSC?

Johennes:
> MAS lists the `org.matrix.cross_signing_reset` action in the `account_management_actions_supported` from [MSC4191](https://github.com/matrix-org/matrix-doc/pull/4191). e.g. in https://account.matrix.org/.well-known/openid-configuration
> 
> Does this need to be added to this MSC?

Good point. I hadn't noticed that [MSC4191](https://github.com/matrix-org/matrix-doc/pull/4191) doesn't list this action. Have added it here with a781be5bfdf62f5750ce0a3c265bd59bf71f85ac.

proposals/4312-x-signing-reset-with-nextgen-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+Semantically, resetting cross-signing keys doesn't fall into the authorization server's domain. The
+scheme outlined above increases coupling between the authorization server and the homeserver and
+makes it more difficult to use off-the-shelve OAuth authorization servers.
+
+## Alternatives
+
+Rather than approving cross-signing reset specifically, the authorization server could provide
+mechanisms for temporary scope elevation.
+
```
hughns, line 74: Following a discussion with @erikjohnston earlier it might be helpful to elaborate on this:

```suggestion

An example of a potential mechanism that could help achieve this is the
[RFC 9470 OAuth 2.0 Step Up Authentication Challenge Protocol](https://datatracker.ietf.org/doc/rfc9470/).

Theoretically such a mechanism could act as full replacement for UIA in the CS API
where protection is needed for sensitive actions.

However, the is no proposal on how this would be applied to the CS API and therefore
it is proposed to codify this present mechanism that does allow for the specific
cross-signing reset action.
```

proposals/4312-x-signing-reset-with-nextgen-auth.md
turt2live: The unsupporting client would indeed be the fallback support - ideally there's evidence of that working in such a client.

proposals/4312-x-signing-reset-with-nextgen-auth.md
```diff
@@ -0,0 +1,93 @@
...
+This is obviously not practical and unofficial workarounds have been invented to enable resetting
+one's cross-signing keys in the client / homeserver / authorization server triangle. This proposal
+documents these workarounds.
+
+## Proposal
+
+Clients that have authenticated via the new [OAuth APIs] continue to use
+[`/_matrix/client/v3/keys/device_signing/upload`] to replace cross-signing keys. Homeservers
+continue to enforce UIA on the endpoint but MUST only use a single stage `m.cross_signing_reset`
```
turt2live:
```suggestion
continue to enforce UIA on the endpoint but MUST only use a single stage `m.oauth`
```

because we'll probably have other endpoints which need a UIA shim too in future (assuming we don't remove the shim/legacy auth before we add those endpoints).

turt2live (approved):
otherwise lgtm

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4312-x-signing-reset-with-nextgen-auth.md
turt2live: This is non-blocking for FCP.

turt2live:
@mscbot fcp merge
@mscbot concern Requires evidence that fallback UIA support kinda mostly works well enough
@mscbot concern Include Step Up alternative in proposal

proposals/4312-x-signing-reset-with-nextgen-auth.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Alternatives
+
+Rather than approving cross-signing reset specifically, the authorization server could provide
+mechanisms for temporary scope elevation.
+
+## Security considerations
+
+Lifting UIA temporarily creates a time window in which an attacker with an access token could take
+over the account.
```
clokep: This seems to imply that UIA is lifted for a period of time but I don't see other references to this in the MSC? Is this leaking of an implementation in the MSC?

proposals/4312-x-signing-reset-with-nextgen-auth.md
clokep: This feels like the correct solution?

Johennes: Yeah, it probably is. This proposal was originally only intended to document the status quo that I found on matrix.org so that it could be reused by other implementations in the interim. It doesn't feel great to add this into the spec. If the good solution is far away though, it might be better than the current situation where the spec renders the OAuth login APIs and cross-signing key reset incompatible.

I applied @hughns suggestion separately with d8649fbdc8f991be5a568c826ebc710537a11d7c but will leave this thread open in case there's further input.

proposals/4312-x-signing-reset-with-nextgen-auth.md
Johennes: Oh, you're right. This is what Synapse & MAS do. Since there is no other implementation it might still be worth calling out? Have clarified it in 3de64d76448565d1627706769398b10aa624093b.

proposals/4312-x-signing-reset-with-nextgen-auth.md
```diff
@@ -0,0 +1,102 @@
...
+
+This is obviously not practical and unofficial workarounds have been invented to enable resetting
+one's cross-signing keys in the client / homeserver / authorization server triangle. This proposal
+documents these workarounds.
+
+## Proposal
+
+Clients that have authenticated via the new [OAuth APIs] continue to use
+[`/_matrix/client/v3/keys/device_signing/upload`] to replace cross-signing keys. Homeservers
+continue to enforce UIA on the endpoint but MUST only use a single stage `m.oauth`
```
dbkr: Is this what's being specced here? The intro has just said this MSC is just documenting workarounds. Also, looks like this bit didn't get changed when it was renamed.

Johennes: > Is this what's being specced here? The intro has just said this MSC is just documenting workarounds.

The scope of this proposal has changed. I originally intended to only document the workarounds from matrix.org for other implementations. Apparently there is now some interest to add it into the spec, however. I have amended the introduction slightly with e43282ee0af85a2b1e3684802aae77533348777f to make this clearer.

There is [another thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4312#discussion_r2376510250) open for discussing whether merging this proposal is a good idea or not.

> Also, looks like this bit didn't get changed when it was renamed.

Actually this was _the only_ bit that was changed in the rename. 🤦 Have renamed the others with 33011178a3eb8d833f545ea33961dbfecbc89c18.

proposals/4312-x-signing-reset-with-nextgen-auth.md
hughns: I've added these details and a video of Element Classic iOS using the fallback to the PR description.

proposals/4312-x-signing-reset-with-nextgen-auth.md
Johennes: It may not change the plan for this proposal but I have tried to adapt RFC9470 to Matrix in https://github.com/matrix-org/matrix-spec-proposals/pull/4363.

proposals/4312-x-signing-reset-with-nextgen-auth.md
```diff
@@ -0,0 +1,102 @@
...
+
+This is obviously not practical and unofficial workarounds have been invented to enable resetting
+one's cross-signing keys in the client / homeserver / authorization server triangle. This proposal
+documents these workarounds as a low-effort interim workaround until better solutions are available.
+
+## Proposal
+
+Clients that have authenticated via the new [OAuth APIs] continue to use
+[`/_matrix/client/v3/keys/device_signing/upload`] to replace cross-signing keys. Homeservers
+continue to enforce UIA on the endpoint but MUST only use a single stage `m.oauth` together with a
```
dbkr: I'm still a bit confused with this proposal, specifically Travis's change, which is all well and logical, but means this MSC is no longer documenting the status quo which is what it claims to be doing, so anyone implementing what is says is not going to get a working client/server, which seems to defeat the point of speccing it.

Less importantly, the fact that this said "documenting the current workaround" and then using a MUST felt incongruent to me and I think it probably just needs to be rephrased that the server offers a flow where the only stage is `m.oauth` (or whatever it ends up as): as this makes it sound like the server must offer one flow and only one flow.

dbkr:
@mscbot concern Is not documenting the current in-the-wild implementation

proposals/4312-x-signing-reset-with-nextgen-auth.md
Johennes: > I'm still a bit confused with this proposal, specifically Travis's change, which is all well and logical, but means this MSC is no longer documenting the status quo which is what it claims to be doing, so anyone implementing what is says is not going to get a working client/server, which seems to defeat the point of speccing it.

I think Travis' renaming from `m.cross_signing_reset` to `m.oauth` hasn't really changed this aspect. The implementations in Synapse, MAS, etc. are using `org.matrix.cross_signing_reset` which differs from both. If compatibility with that choice is our goal, we'll need to fix the stage name to `org.matrix.cross_signing_reset`.

> Less importantly, the fact that this said "documenting the current workaround" and then using a MUST felt incongruent to me and I think it probably just needs to be rephrased that the server offers a flow where the only stage is `m.oauth` (or whatever it ends up as): as this makes it sound like the server must offer one flow and only one flow.

How does this sound?

```suggestion
continue to enforce UIA on the endpoint with a flow containing a single stage `m.oauth` together with a
```

proposals/4312-x-signing-reset-with-nextgen-auth.md
richvdh: Given that this is already deployed on matrix.org ( :cry: ), and Element-Web implements support for it (cf https://github.com/element-hq/element-meta/issues/2956), I think we had better spec what we have asap and leave the Correct Solution for another time.

proposals/4312-x-signing-reset-with-nextgen-auth.md
richvdh: > I think Travis' renaming from `m.cross_signing_reset` to `m.oauth` hasn't really changed this aspect. The implementations in Synapse, MAS, etc. are using `org.matrix.cross_signing_reset` which differs from both. If compatibility with that choice is our goal, we'll need to fix the stage name to `org.matrix.cross_signing_reset`.

Agreed, on all fronts.

I like the idea of switching away from the unstable prefix (and, if we're going to do that, it may as well be to `m.oauth`), but we may need to consider the mechanics of doing so. I think probably servers will need to offer *two* flows (one with each of `m.oauth` and `org.matrix.cross_signing_reset`), so that clients which only support the latter will use it.

turt2live:
@mscbot resolve Requires evidence that fallback UIA support kinda mostly works well enough
@mscbot resolve Include Step Up alternative in proposal

proposals/4312-x-signing-reset-with-nextgen-auth.md
Johennes: > I think probably servers will need to offer _two_ flows (one with each of `m.oauth` and `org.matrix.cross_signing_reset`), so that clients which only support the latter will use it.

Good point. I have added this as a recommendation with dc4881793027fcf709735e32ecc0373cd0014438.

dbkr: Ah yep, excellent point - I'd failed to realise it was (thankfully) using an org.matrix prefix in practice. It might be worth just putting in a note to say that the stable flow was renamed in the MSC process just to help clarify to future readers.

And yep, updated wording above sounds perfect, thank you!

dbkr:
@mscbot resolve Is not documenting the current in-the-wild implementation

dbkr:
@mscbot resolve Is not documenting the current in-the-wild implementation

proposals/4312-x-signing-reset-with-nextgen-auth.md
Johennes: Excellent. 👍 Have added a footnote about the `m.oauth` renaming with 3f49d152259938910649f5b62cfb301856af111d and applied the suggestion from above in 40e1687f1afb943988d793b9639a9d047e254b26.

clokep:
To write it out explicitly -- I'm pretty meh on this change (probably a -0.3), but not enough to block it.

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2234

turt2live:
Merged 🎉

PR #4313, opened 2025-07-15
MSC4313: Require HTML `<ol>` `start` Attribute support

HarHarLinks:
[Rendered](https://github.com/HarHarLinks/matrix-spec-proposals/blob/require-ol-start-attribute/proposals/4313-require_html_ol_start_attribute.md)

Signed-off-by: HarHarLinks <2803622+HarHarLinks@users.noreply.github.com>

I am employed at [Nordeck](https://nordeck.net), Matrix community member, and Member of the Matrix.org Foundation Governing Board. This proposal was written and published a community member.

----

Known implementations:
- Element-Web
- Element X iOS
- Element classic iOS
- Fractal
- FluffyChat 2.0

Partial implementations:
- Nheko 0.12.0
- NeoChat 25.04.02
- Element classic Android 1.6.42

----

SCT stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4313#issuecomment-3545007262)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4313#issuecomment-3155659335)

### DIFF
--- a/proposals/4313-require_html_ol_start_attribute.md
+++ b/proposals/4313-require_html_ol_start_attribute.md
@@ -78,15 +78,22 @@ which could mean that fewer clients could choose to implement `ol` at all.
 
 ## Alternatives
 
-- Define a list of all HTML tags whose displaying must be supported if `formatted_body` is used to display
-  messages at all, based on whether tags can replace characters such as in the demonstrated example.
-  This could apply recursively also for all attributes.
-- Find a way for clients to determine whether the `body` matches its supported interpretation of the
-  `formatted_body`.
-  This could end up very similar to the previous alternative and additionally lead to inconsistent
-  behavior on clients where `formatted_body` is only sometimes used for display as a result.
-- Remove HTML from the spec entirely. Possibly replace it with another markup language that prevents
-  this issue.
+1. Define a list of all HTML tags whose displaying must be supported if `formatted_body` is used to display
+   messages at all, based on whether tags can replace characters such as in the demonstrated example.
+   This could apply recursively also for all attributes.
+2. Find a way for clients to determine whether the `body` matches its supported interpretation of the
+   `formatted_body`.
+   This could end up very similar to the previous alternative and additionally lead to inconsistent
+   behavior on clients where `formatted_body` is only sometimes used for display as a result.
+3. Remove support for `start` and possibly other attributes and tags that can lead to this type of issue.
+   This is obviously not backwards compatible, though perhaps closer to common expectations about messenger UX.
+   (Note that the spec usually leaves this type of decision as an implementation detail.
+   Feedback discussing this MSC has shown that even with the current proposal implemented, many wish for the UX of
+   currently commonly used clients to adapt to different expectations of their users regarding numbered lists.)
+   In keeping to the spirit of this MSC, this would change the spec to introduce some "MUST NOT use" attributes
+   on top of the current list of "strongly suggested" ones.
+4. Remove HTML from the spec entirely. Possibly replace it with another markup language that prevents
+   this issue.
 
 
 ## Security considerations


proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+Rendering this, Alices screen shows:
+
+Bob said:  
+1.
+
+This is a clear break in communication, since this message has an entirely different meaning not only
+from Bob's intended meaning, but also as it is viewed from different client implementations.
+
+This MSC proposes to alter the spec such that a client implementing rendering of the `ol` HTML tag
+in `formatted_body`s is also required ("MUST") to implement its `start` attribute, in oder to prevent
```
HarHarLinks:
```suggestion
in `formatted_body`s is also required ("MUST") to implement its `start` attribute, in order to prevent
```

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+This proposal increases the load on client developers, though presumable only a tiny bit,
+which could mean that fewer clients could choose to implement `ol` at all.
+
+
+## Alternatives
+
+- Define a list of all HTML tags whose displaying must be supported if `formatted_body` is used to display
+  messages at all, based on whether tags can replace characters such as in the demonstrated example.
+  This could apply recursively also for all attributes.
+- Find a way for clients to dermine whether the `body` matches its supported interpretation of the
```
HarHarLinks:
```suggestion
- Find a way for clients to determine whether the `body` matches its supported interpretation of the
```

proposals/4313-require_html_ol_start_attribute.md
turt2live, line 1: Implementation requirements:

* [x] Client (ideally several)

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+
+Alice asks:  
+1. \<a very long option\>
+2. \<another very long description\>
+3. \<a huge third description\>
+
+Bob replies:  
+2.
+
+Let's assume Bob's client takes the option to translate the plain text `2.` to HTML.
```
turt2live, line 48: arguably this is the bug, not the lack of `start`

turt2live: (also, why aren't they using polls?)

HarHarLinks: The spec says I may use `start` in my HTML body. If I use it and my conversational partner cannot interpret it correctly, then the spec does not adequately prevent communication from breaking there. How we get into this situation is irrelevant imo, unless we want to start speccing how the `formatted_body` must be constructed, which I hope is widely out of scope of Matrix.

(Whether to use polls or not is also not the question.)

spaetz: err, all that you need to do to invoke this issue is typing in EXA that you like the
4. of July.

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: - I know this is already implemented in Element-Web, Element X iOS, Element classic iOS, Fractal, FluffyChat 2.0. There are likely more. What do you need as confirmation?

- Implementations meeting this proposal but with limited issues (such as not rendering the number when the item has no content) exist, e.g. Nheko 0.12.0-9571d811 [#1369](https://github.com/Nheko-Reborn/nheko/issues/1396), NeoChat, Element/SchildiChat classic, FluffyChat 1.25.

- A counter-example would be Element X Android, though the issue [has been fixed in some places](https://github.com/element-hq/element-x-android/pull/5044) 3 days after opening this MSC, while timeline previews, replies, etc remain an issue.

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+loss of meaning of a message.
+
+
+## Potential issues
+
+This proposal increases the load on client developers, though presumable only a tiny bit,
+which could mean that fewer clients could choose to implement `ol` at all.
+
+
+## Alternatives
```
Johennes, line 79: Could you prevent the sending client from turning `2.` into an `<ol>` in the first place?

toger5: Independent of where this MSC goes I feel like this ("better logic for when to convert `X.` (numbers) into `<ol>`) is an issue report that should be opened against clients that currently do not do that in a sane way ?
I even remember seeing such issues iirc.

HarHarLinks: Should I change my example, because apparently many reviewers are missing the issue I am trying to point out for another issue.

It is currently legal and intended within the Matrix spec for Alice to type:
```
3. A
4. B
5. C
```
and Alice's client to send this as `"formatted_body": "<ol start=\"3\"><li>A</li><li>B</li><li>C</li>"`.

The above is a legal message, and there are valid use cases to get there. This `formatted_body` legally existing is the basis for this MSC. It is not concerned with how to get there, as long as you are allowed to get there.

Then the issue is that Bob's client missing support for `start` might render it as
```
1. A
2. B
3. C
```

---

What I at this point regard as out of scope of this MSC (convince me otherwise?) is prescribing to clients how to and how not to generate their HTML. Possible ways for clients to construct `formatted_body`s could be:
- interpret "plain text" as some markdown dialect (yes this is inconsistent across different clients), as most clients do today
- users enter custom HTML straight in the input box, e.g. using Element Web's `/html` command
- parse Org mode
- a bot doing STT and heuristically applying formatting
- any custom input language
- AI generating `formatted_body` from a user signing sign language into their webcam
- anything you can think of

So to get rid of erratic behavior, the spec could try to come up with rules trying to cover a million edge and corner cases that prevent a simple answer of `2.` (current MSC example) or `4. of July` (from spaetz) from being converted to an `ol`. But still the problem is not solved, see example above.

An easy alternative would be to disallow `start` from the spec, i.e. only allow ordered lists starting at 1, and leaving it up to clients to not mistakenly convert numbered lists starting not at 1 (incl misinterpretations like `4. of July`) to `ol`s.

Can anyone find a reference as to when `start` support was added to the spec, ideally the MSC and reasoning for it?

Johennes: I think fixing the `start` issue as per this proposal definitely makes sense in any case.

I was just thinking that when I type "4. Juli", I actually wouldn't want it to become an ordered list starting at 4. Even if it's displayed similarly, semantically it's something very different from what I intended. Agree that this is a different problem though and should be solved in another proposal.

HarHarLinks: > I was just thinking that when I type "4. Juli", I actually wouldn't want it to become an ordered list starting at 4. Even if it's displayed similarly, semantically it's something very different from what I intended. Agree that this is a different problem though and should be solved in another proposal.

To be clear, I agree that this is annoying and suboptimal, and clients should do better. I am not sure though if this even falls into the area of concern of the spec.

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+empty entry, that starts at an index of two.
+
+Let's assume Alice's client also implements HTML markup in a configuration allowed by the spec:
+Her client supports `ol` tags, but not the `start` attribute.
+A common implementation is to parse the HTML and simply remove any tags not implemented by the client.
+After safely ingesting the message, Alice's client ends up with `"formatted_body": "<ol></ol>"`.
+Rendering this, Alices screen shows:
+
+Bob said:  
+1.
```
HarHarLinks: For evidence, this is the case in Element X as of 25.07.0.

Further evidence: nheko 0.12.0-9571d811 and SchildiChat ("classic") will show only whitespace here.

proposals/4313-require_html_ol_start_attribute.md
spaetz: Agreed that this is beyond this spec proposal and should be part of client implementations. I just want to highlight some of the [commonmark spec text](https://spec.commonmark.org/0.31.2/#lists), stating:
> In CommonMark, a list can interrupt a paragraph. That is, no blank line is needed to separate a paragraph from a following list:
> ```
> Foo
> 1. bar
> ``` 
> Markdown.pl [the original markdown linter] does not allow this, through fear of triggering a list via a numeral in a hard-wrapped line:
> 
> The number of windows in my house is
> 14.  The number of doors is 6.

So, requiring an empty line before starting an ordered list would at least solve some of this problem. It would not solve the problem when I write `4. Juli` at the beginning of a paragraph, of course.

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: I am resolving this thread in the understanding question/comment is sufficiently clarified. If not @turt2live please elaborate.

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: I am resolving this thread in the understanding question/comment is sufficiently clarified. If not please elaborate.

turt2live:
I'm still a bit unsure about whether this needs to be fixed in the spec, or on the sending client's side, but not enough to block progress.

@mscbot fcp merge

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,87 @@
...
+
+This is a clear break in communication, since this message has an entirely different meaning not only
+from Bob's intended meaning, but also as it is viewed from different client implementations.
+
+This MSC proposes to alter the spec such that a client implementing rendering of the `ol` HTML tag
+in `formatted_body`s is also required ("MUST") to implement its `start` attribute, in order to prevent
```
anoadragon453:
```suggestion
in `formatted_body`s MUST implement its `start` attribute, in order to prevent
```

HarHarLinks: Thanks, done in [49fc648](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/49fc6482d300f4a62360a6ac7436b46417117471)!

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+Bob said:  
+1.
+
+This is a clear break in communication, since this message has an entirely different meaning not only
+from Bob's intended meaning, but also as it is viewed from different client implementations.
+
+This MSC proposes to alter the spec such that a client implementing rendering of the `ol` HTML tag
+in `formatted_body`s MUST also implement its `start` attribute, in order to prevent
+loss of meaning of a message.
+
```
HarHarLinks, line 71: When I investigated the current status of implementation in https://github.com/matrix-org/matrix-spec-proposals/pull/4313#discussion_r2208496530, I believe I found the information that HTML already specifies that a empty `li` should still render the enumeration mark, i.e. drawing on the example from the MSC, `<ol start="2"><li></li></ol>` MUST be rendered as `2.`. As I applied anoa's wording suggestion, it crossed my mind whether this would be worth mentioning here in the MSC. Not to be included as normative text, but to resolve ambiguity since some clients have this issue which is part of the goal to resolve with this MSC. However I can't find that info again right now - maybe someone happens to know?

In the area are https://html.spec.whatwg.org/#the-ol-element or https://www.w3.org/TR/css-lists-3/#marker-pseudo but neither have the definition I'm looking for.

HarHarLinks: All I could find again only heavily implies the above behavior between the lines, so I don't have anything reliable to reference, hence resolving.

proposals/4313-require_html_ol_start_attribute.md
anoadragon453: Considering multiple clients have implemented this successfully, I believe the implementation requirements have been satisfied.

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+The Matrix specification allows text messages to optionally contain a HTML-formatted version over the plain text
+body.
+A set of "safe" tags is recommended, along with a set of "safe" attributes for some of the tags that
+support them. 
+Additional Matrix-specific attributes are also introduced. However, all of this is optional on any level:
+Clients may choose for example to
+- not implement sending or showing or HTML-formatting at all
+- only implement some tags
+- implement additional tags outside of the existing recommendation
```
richvdh, line 14: can we have some links to existing bits of spec where all this is defined, please?

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+
+Imagine the following conversation to illustrate:
+
+Alice asks:  
+1. \<a very long option\>
+2. \<another very long description\>
+3. \<a huge third description\>
+
+Bob replies:  
+2.
```
richvdh: this is hard to read in the rendered output, partly because Github's markdown renderer turns Alice's message into a list, but not Bob's. Maybe wrap both messages in triple-backticks?

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: Pretty much this subsection https://spec.matrix.org/latest/client-server-api/#mroommessage-msgtypes

I assume you mean putting it in the MSC text; I'll do that in a bit.

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+[RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Proposal
+
+Imagine the following conversation to illustrate:
+
+Alice asks:  
+1. \<a very long option\>
+2. \<another very long description\>
+3. \<a huge third description\>
```
richvdh: maybe add some context about this?


```suggestion
1. \<a very long option\>
2. \<another very long description\>
3. \<a huge third description\>

(Alice's client will likely turn this markdown into a numbered list; however that is not important for this discussion: what is important is the presence of a set of numbered options that can be read by Bob.)
```

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+Let's assume Bob's client takes the option to translate the plain text `2.` to HTML.
+Assuming further that Bob's client has full support to the extent recommended by the spec, then Bob's
+message becomes `"formatted_body": "<ol start=\"2\"><li></li></ol>"`, i.e. an ordered ("numbered") list with a single,
+empty entry, that starts at an index of two.
+
+Let's assume Alice's client also implements HTML markup in a configuration allowed by the spec:
+Her client supports `ol` tags, but not the `start` attribute.
+A common implementation is to parse the HTML and simply remove any tags not implemented by the client.
+After safely ingesting the message, Alice's client ends up with `"formatted_body": "<ol><li></li></ol>"`.
+Rendering this, Alices screen shows:
```
richvdh:
```suggestion
Rendering this, Alice's screen shows:
```

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+
+This MSC proposes to alter the spec such that a client implementing rendering of the `ol` HTML tag
+in `formatted_body`s MUST also implement its `start` attribute, in order to prevent
+loss of meaning of a message.
+
+
+## Potential issues
+
+This proposal increases the load on client developers, though presumable only a tiny bit,
+which could mean that fewer clients could choose to implement `ol` at all.
```
richvdh: This doesn't mean what you want it to mean.

```suggestion
This proposal increases the load on client developers (though presumably only a tiny bit),
which could mean that fewer clients could choose to implement `ol` at all.
```

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+
+
+## Alternatives
```
richvdh, line 79: remove support for `start`?

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -0,0 +1,90 @@
...
+  This could end up very similar to the previous alternative and additionally lead to inconsistent
+  behavior on clients where `formatted_body` is only sometimes used for display as a result.
+- Remove HTML from the spec entirely. Possibly replace it with another markup language that prevents
+  this issue.
+
+
+## Security considerations
+
+No potential security issues are known to the author.
+Only options already allowed are being defined more precisely.
```
richvdh: This doesn't mean quite what you want it to mean.

```suggestion
We are only defining more precisely options that are already allowed.
```

proposals/4313-require_html_ol_start_attribute.md
richvdh: thanks. Recommend linking to a specific spec version (v1.17) rather than `latest`, to give a stable link.

HarHarLinks: done in [29842fe](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/29842fe2b58cf936af0810f8492f35d81897dfdd)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: [1f15cfc](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/1f15cfc0e4452a9da9f5a9c26e28aeaf1cde44fd)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: [1f15cfc](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/1f15cfc0e4452a9da9f5a9c26e28aeaf1cde44fd)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: [1f15cfc](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/1f15cfc0e4452a9da9f5a9c26e28aeaf1cde44fd)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: I gave it a shot in [9a6c59a](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/9a6c59a16c1b48d1d6c38f9c3594023943b26cf9)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: [1f15cfc](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/1f15cfc0e4452a9da9f5a9c26e28aeaf1cde44fd)

proposals/4313-require_html_ol_start_attribute.md
HarHarLinks: added in [565c9e0](https://github.com/matrix-org/matrix-spec-proposals/pull/4313/commits/565c9e0a5f07caa9a0364034ded4fc61aeb443bd)

proposals/4313-require_html_ol_start_attribute.md
```diff
@@ -21,17 +24,26 @@ The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL
...
+2. <another very long description>
+3. <a huge third description>
+```
 
-Bob replies:  
+Bob replies very minimally, using the number indices Alice conveniently provided:
+
+```
 2.
+```
```
richvdh, line 46: much clearer, thanks!

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2299

turt2live:
Merged 🎉

PR #4315, opened 2025-07-28
Clarify `/upgrade` behaviour in MSC4289 to match implementation

turt2live:
Discovered while writing the spec PR.

**Process note**: MSCs can receive clarification changes post-FCP, but anything materially different requires a new MSC. I believe this qualifies as clarification because it isn't core to the MSC and the implementation used for FCP does the behaviour described. SCT members are able to disagree and require a new MSC instead.

### DIFF
--- a/proposals/4289-privilege-creators.md
+++ b/proposals/4289-privilege-creators.md
@@ -154,6 +154,13 @@ the _entire set_ of additional creators are specified up-front. For example, con
 
 This allows the creator status to move between users.
 
+**Note** (added July 28, 2025 post-acceptance for clarity): If `additional_creators` is specified on
+`/upgrade` but the new room version doesn't support `additional_creators`, the field is not used and
+does nothing. For example, if a room was being upgraded to `11` with `additional_creators: [@alice:example.org]`,
+the request would (probably) 200 OK but the new room's `m.room.create` event would *not* have an
+`additional_creators` field. If the room was instead being upgraded to `12` (which contains this MSC),
+then `additional_creators` would show up in the create event's `content`.
+
 ### Potential Issues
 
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):

proposals/4289-privilege-creators.md
```diff
@@ -154,6 +154,13 @@ the _entire set_ of additional creators are specified up-front. For example, con
 
 This allows the creator status to move between users.
 
+**Note** (added July 28, 2025 post-acceptance for clarity): If `additional_creators` is specified on
+`/upgrade` but the new room version doesn't support `additional_creators`, the field is not used and
+does nothing. For example, if a room was being upgraded to `11` with `additional_creators: [@alice:example.org]`,
+the request would (probably) 200 OK but the new room's `m.room.create` event would *not* have an
+`additional_creators` field. If the room was instead being upgraded to `12` (which contains this MSC),
+then `additional_creators` would show up in the create event's `content`.
```
clokep, line 162: Why wouldn't it have it? I would expect it to have and it and just do nothing with it, like today if you add any other random field to a create event?

I.e. I'd expect either:

1. An error
2. The field to be used as normal

turt2live: This is just what the implementation did. I hold no opinion on whether it's correct.

tulir: I think it matches what an old (non-v12-aware) server would do? `/upgrade` doesn't have `creation_content`, so there's no expectation of fields being passed through. Unknown fields are also generally ignored rather than throwing errors

clokep: That's fair enough. 👍

PR #2882, opened 2020-11-27
MSCxxxx: Tempered Transitive Trust

uhoreg:

PR #4316, opened 2025-07-29
MSC4316: External cross-signing signatures with X.509 certificates and (semi-)automated verification

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/x509-cross-signing/proposals/4316-x509-cross-signing.md)

### DIFF
--- a/proposals/4316-x509-cross-signing.md
+++ b/proposals/4316-x509-cross-signing.md
@@ -96,6 +96,9 @@ Including certificates in [`/_matrix/client/v3/keys/query`] will increase the si
 
 ## Alternatives
 
+[MSC2882] solves a similar problem by introducing transitive trust and letting users vouch for other
+users. This only works when a first level of trusted users exists though.
+
 External signatures and certificates could be combined by using PKCS \#7 detached signatures.
 
 ``` json5
@@ -130,4 +133,5 @@ This proposal doesn't depend on but plays together with [MSC4317].
   [format]: https://spec.matrix.org/v1.15/appendices/#signing-details
   [unpadded base64]: https://spec.matrix.org/v1.15/appendices/#unpadded-base64
   [avoid leaking social graphs]: https://spec.matrix.org/v1.15/client-server-api/#key-and-signature-security
+  [MSC2882]: https://github.com/matrix-org/matrix-spec-proposals/pull/2882
   [MSC4317]: https://github.com/matrix-org/matrix-spec-proposals/pull/4317
--- a/proposals/4316-x509-cross-signing.md
+++ b/proposals/4316-x509-cross-signing.md
@@ -121,7 +121,7 @@ While this MSC is not considered stable, `external` should be referred to as
 
 ## Dependencies
 
-None.
+This proposal doesn't depend on but plays together with [MSC4317].
 
   [cross-signing]: https://spec.matrix.org/v1.15/client-server-api/#cross-signing
   [device verification]: https://spec.matrix.org/v1.15/client-server-api/#device-verification
@@ -130,3 +130,4 @@ None.
   [format]: https://spec.matrix.org/v1.15/appendices/#signing-details
   [unpadded base64]: https://spec.matrix.org/v1.15/appendices/#unpadded-base64
   [avoid leaking social graphs]: https://spec.matrix.org/v1.15/client-server-api/#key-and-signature-security
+  [MSC4317]: https://github.com/matrix-org/matrix-spec-proposals/pull/4317
--- a/proposals/4316-x509-cross-signing.md
+++ b/proposals/4316-x509-cross-signing.md
@@ -1,0 +1,132 @@
+# MSC4316: External cross-signing signatures with X.509 certificates and (semi-)automated verification
+
+Matrix allows users to establish explicit trust and prevent MITM attacks through [cross-signing] and
+[device verification]. The current verification options require an out-of-band channel such as a
+video call or meeting in person and comparing authentication strings or scanning QR codes. This
+process is somewhat cumbersome and less helpful for people that haven't yet met in the physical
+world.
+
+As an example, say you've found the MXID of Dr. Bob, General Physician, online and invite him to
+chat about your latest health issues. You may now verify him in a video call but you haven't
+actually met Dr. Bob before and the person on the other end might as well be anybody. You do know
+that Dr. Bob works in a hospital and the hospital has an X.509 certificate from a national health
+CA. The hospital accrediting that Dr. Bob is in fact Dr. Bob would, therefore, help you trust him.
+However, Matrix doesn't currently support binding Dr. Bob's cryptographic identity to his
+organization's certificate.
+
+To address this shortcoming, the present proposal introduces a mechanism to attach and share X.509
+based signatures on cross-signing keys.
+
+## Proposal
+
+The `signatures` properties in the response of [`/_matrix/client/v3/keys/query`] are extended to
+allow a special key `external`. As this is not a valid [MXID], it cannot clash with other keys
+inside `signatures`. `external` is meant to store signatures that are tied to X.509 certificates in
+the same [format] already used for other signatures. Since in this case the certificate encapsulates
+both the signing algorithm and the public key, the generic string `x509` is used as the algorithm
+identifier while the certificate's SHA-256 fingerprint encoded in [unpadded base64] is used as the
+key identifier.
+
+As a result, an externally signed device key looks like this:
+
+``` json5
+{
+  "device_keys": {
+    "@bob:hospital.org": {
+      "ABCDEFGHIJ": {
+        // Device identity & signing keys
+        "keys": {
+          "curve25519:ABCDEFGHIJ": "...",
+          "ed25519:ABCDEFGHIJ": "..."
+        },
+        ...,
+        "user_id": "@bob:hospital.org",
+        "signatures": {
+          "@bob:hospital.org": {
+            // Device self-signature
+            "ed25519:ABCDEFGHIJ": "...",
+            // Self-signing key signature
+            "ed25519:$SSK_BASE64": "..."
+          },
+          "external": {
+            // External X.509 signature
+            "x509:$FINGERPRINT_BASE64": "$SIGNATURE"
+          }
+        }
+      }
+    }
+  },
+  ...
+}
+```
+
+Home servers MAY additionally include the certificate itself via a new top-level property `external`
+that stores the DER encoding of the certificate encoded in [unpadded base64] under the same signing
+key identifier.
+
+``` json5
+{
+  "device_keys": { ... },
+  ...
+  "external": {
+    "x509:$FINGERPRINT_BASE64": "$CERT.DER_BASE64"
+  }
+}
+```
+
+In closed federations, it is expected that clients are aware of a central PKI which obviates the
+need to expose the certificate in [`/_matrix/client/v3/keys/query`].
+
+Clients MAY check external signatures using the certificate and its chain. If they do, they SHOULD
+also verify certificate validity, for instance, using OCSP. Clients MAY additionally present the
+signing certificate to the user and prompt them for review and confirmation. Once a signature has
+been checked successfully, clients SHOULD mark the signed key as verified.
+
+Servers currently limit the visibility of signatures to [avoid leaking social graphs]. Since
+external signatures are not tied to a signing account, servers MUST make any external signatures and
+certificates in [`/_matrix/client/v3/keys/query`] visible to all users.
+
+This proposal doesn't prescribe how an external signature is added to a user's account as this is
+assumed to be an administrator task. It is expected that home servers will offer suitable admin APIs
+to facilitate this task.
+
+## Potential issues
+
+Including certificates in [`/_matrix/client/v3/keys/query`] will increase the size of the response.
+
+## Alternatives
+
+External signatures and certificates could be combined by using PKCS \#7 detached signatures.
+
+``` json5
+"signatures": {
+  "external": {
+    "pkcs7": "$FILE.P7S_BASE64"
+  }
+}
+```
+
+This would remove the need to identify and look-up certificates. It would, however, fit less well
+with the current signature format in Matrix.
+
+## Security considerations
+
+A client that was tricked into trusting a malign certificate may be tricked into considering
+arbitrary users as verified.
+
+## Unstable prefix
+
+While this MSC is not considered stable, `external` should be referred to as
+`de.gematik.msc4316.external`.
+
+## Dependencies
+
+None.
+
+  [cross-signing]: https://spec.matrix.org/v1.15/client-server-api/#cross-signing
+  [device verification]: https://spec.matrix.org/v1.15/client-server-api/#device-verification
+  [`/_matrix/client/v3/keys/query`]: https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3keysquery
+  [MXID]: https://spec.matrix.org/v1.15/appendices/#user-identifiers
+  [format]: https://spec.matrix.org/v1.15/appendices/#signing-details
+  [unpadded base64]: https://spec.matrix.org/v1.15/appendices/#unpadded-base64
+  [avoid leaking social graphs]: https://spec.matrix.org/v1.15/client-server-api/#key-and-signature-security
--- a/proposals/4316-x509-cross-signing.md
+++ b/proposals/4316-x509-cross-signing.md
@@ -27,7 +27,7 @@ both the signing algorithm and the public key, the generic string `x509` is used
 identifier while the certificate's SHA-256 fingerprint encoded in [unpadded base64] is used as the
 key identifier.
 
-As a result, an externally signed device key looks like this:
+As an example, an externally signed device key looks like this:
 
 ``` json5
 {


proposals/4316-x509-cross-signing.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+need to expose the certificate in [`/_matrix/client/v3/keys/query`].
+
+Clients MAY check external signatures using the certificate and its chain. If they do, they SHOULD
+also verify certificate validity, for instance, using OCSP. Clients MAY additionally present the
+signing certificate to the user and prompt them for review and confirmation. Once a signature has
+been checked successfully, clients SHOULD mark the signed key as verified.
+
+Servers currently limit the visibility of signatures to [avoid leaking social graphs]. Since
+external signatures are not tied to a signing account, servers MUST make any external signatures and
+certificates in [`/_matrix/client/v3/keys/query`] visible to all users.
```
codedust, line 87: Wouldn't it be sufficient to reveal certificates (EDIT: and signatures) to room members after accepting an invite, i.e. after joining the room.

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Proposal
+
+The `signatures` properties in the response of [`/_matrix/client/v3/keys/query`] are extended to
+allow a special key `external`. As this is not a valid [MXID], it cannot clash with other keys
+inside `signatures`. `external` is meant to store signatures that are tied to X.509 certificates in
+the same [format] already used for other signatures. Since in this case the certificate encapsulates
+both the signing algorithm and the public key, the generic string `x509` is used as the algorithm
+identifier while the certificate's SHA-256 fingerprint encoded in [unpadded base64] is used as the
+key identifier.
```
codedust, line 28: The format already used for other signatures only supports the `ed25519` signing algorithm. This would restrict X509 certificates to Curve25519 keys.

Alternatively, JSON Web Signatures ([RFC 7515](https://www.rfc-editor.org/rfc/rfc7515.html)) could be used in combination with X.509 certificates.

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+As an example, say you've found the MXID of Dr. Bob, General Physician, online and invite him to
+chat about your latest health issues. You may now verify him in a video call but you haven't
+actually met Dr. Bob before and the person on the other end might as well be anybody. You do know
+that Dr. Bob works in a hospital and the hospital has an X.509 certificate from a national health
+CA. The hospital accrediting that Dr. Bob is in fact Dr. Bob would, therefore, help you trust him.
+However, Matrix doesn't currently support binding Dr. Bob's cryptographic identity to his
+organization's certificate.
+
+To address this shortcoming, the present proposal introduces a mechanism to attach and share X.509
+based signatures on cross-signing keys.
```
codedust, line 18: Below, it says:

> "As a result, an externally signed device key looks like this:"

```suggestion
based signatures on device keys.
```

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Proposal
+
```
codedust, line 21: Is it in scope of this MSC to display the x509 subject and issuer to the user (the introduction suggests that this is required: "the person on the other end might as well be anybody").

A short section describing the use of the x509 certificate would be super helpful, i.e. what exactly is signed by the x509 cert and why and how many certificates are used (one per user or one per device?). Do we sign the master key (MSK), the self-signing key (SSK) or the device keys using the x509 cert?

1. Signing the master key validates the users' identity (and leaves it up to the user which devices to validate via the self-signing key).
2. Signing the self-signing key does not explicitly validate the users' identity but is otherwise very similar to 1.
3. Signing one or all device keys allows for per-device x509 certificates at the expense that x509 certificates need to be rolled out to every device (as opposed to the possibility to sign the master key / the self-signing key on the server-side (?)).
4. For ideal privacy properties, we **do not want to use x509 signatures on any of these keys at all since this is a privacy risk**: If the x509 cert contains personally identifiable information, the publication of the x509 certificate and the x509 signature (or a data leak including these) would leak and publicly confirm the ownership of the respecting matrix account. Instead, x509 signatures could be used to establish an authenticated channel (Olm handshake / device verification) *after* the user accepted a room invite. For this reason, TLS 1.3 uses the x509 private key to sign a hash of the TLS handshake transcript (see RFC8446, [section 4.4.3](https://www.rfc-editor.org/rfc/rfc8446#section-4.4.3) and [appendix E.1 | grep "Informally, "](https://www.rfc-editor.org/rfc/rfc8446#appendix-E.1)). The x509 signature would then not need to be published nor revealed to the administrator of the users' home server. This is especially relevant if the PKI and the home server are operated by two different entities.

Also the implications of this and expected client behavior needs to be described. When Alice' device keys are signed by one or multiple x509 certificates, do clients still accept device key signatures by the users' self-signing key?

I created a visualization using [Asciiflow](https://asciiflow.com/) based on the visualization in the [cross-signing section of the Client-Server API](https://spec.matrix.org/v1.15/client-server-api/#cross-signing):

```
+------------------------------+                ..................   +----------------+
| +--------------------------+ |   ..................            :   | +------------+ |
| |                          v v   v            :   :            v   v v            | |
| |                       +-----------+         :   :         +-----------+         | |
| |                       | Alice MSK |         :   :         |  Bob MSK  |         | |
| |                       +-----------+         :   :         +-----------+         | |
| |                         |       :           :   :           :       |           | |
| |                      +--+       :...        :   :        ...:       +--+        | |
| |                      v             v        :   :        v             v        | |
| |  +-----------+ +-----------+ .............  :   :  ............. +-----------+  | |
| |  | X.509 crt | | Alice SSK | : Alice USK :  :   :  :  Bob USK  : |  Bob SSK  |  | |
| |  +-----------+ +-----------+ :...........:  :   :  :...........: +-----------+  | |
| |             |    |  ...  |         :        :   :        :         |  ...  |    | |
| |             V    V       V         :........:   :........:         V       V    | |
| |            +---------+   -+                                   +---------+   -+  | |
| |            | Devices | ...|                                   | Devices | ...|  | |
| |            +---------+   -+                                   +---------+   -+  | |
| |                 |  ...  |                                          |  ...  |    | |
| +-----------------+       |                                          |       +----+ |
+---------------------------+                                          +--------------+
```

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+"signatures": {
+  "external": {
+    "pkcs7": "$FILE.P7S_BASE64"
+  }
+}
+```
+
+This would remove the need to identify and look-up certificates. It would, however, fit less well
+with the current signature format in Matrix.
+
```
uhoreg, line 114: https://github.com/matrix-org/matrix-spec-proposals/pull/2882 should probably be listed as an alternative, since it has a similar objective of trusting some keys to vouch for other identities.

proposals/4316-x509-cross-signing.md
Johennes: I actually deliberately used the current phrasing because I wanted to support both signing a device key and signing the MSK. So maybe rather than "As a result" the sentence below should start with "As an example".

proposals/4316-x509-cross-signing.md
Johennes: Hm, good question. This restriction does not seem to exist for the other contents of `/keys/query`. I assume this is because the current verification process can happen either in a room or via to-device messages.

For the verification as per this proposal, it might be nice if you could verify somebody even before opening a room with them?

proposals/4316-x509-cross-signing.md
Johennes: My intention was to add `x509` to the list of supported signature algorithms as a meta value where `x509` would mean infer the actual algorithm from the certificate. I think that inference could be ambiguous though.

Using JWS is a really interesting idea. It does have a number of different ways to identify the key though which makes me wonder how we'd best combine it with the existing response structure. Maybe `external` could just be an array of base64-encoded JWSs?

``` json5
"external": ["$JWS_BASE64", ...]
```

proposals/4316-x509-cross-signing.md
Johennes: We might be thinking about the certificate in two different ways. I had intended the certificate to be tied to an organization that accredits users / devices. So it wouldn't contain any personal information. It just holds the public part of the key used for making that accreditation.

I believe you're envisioning a scheme where users / devices get their own certificates?

> Is it in scope of this MSC to display the x509 subject and issuer to the user (the introduction suggests that this is required: "the person on the other end might as well be anybody").

I had included it as an [optional step](https://github.com/matrix-org/matrix-spec-proposals/pull/4316/files#diff-ce4864babe2df68380fae0b8319c06c89d87bbc8cdc2b595a3b1f1809e01af8aR81-R82), yes. My thinking was that your client might be pre-configured to automatically trust certain certificates.

> * Signing the master key validates the users' identity (and leaves it up to the user which devices to validate via the self-signing key).
> 
> * Signing the self-signing key does not explicitly validate the users' identity but is otherwise very similar to 1.
> 
> * Signing one or all device keys allows for per-device x509 certificates at the expense that x509 certificates need to be rolled out to every device (as opposed to the possibility to sign the master key / the self-signing key on the server-side (?)).

Signing either the MSK or a device key seems sensible to me. I wonder if we should exclude the SSK? Did you have a specific use case in mind there?

proposals/4316-x509-cross-signing.md
codedust: Verification before accepting an invite would mean that user-identifying information (name, role, etc.) of Alice needs to be accessible to other users who want to verify the identity of Alice.

There is a tradeoff we need to decide on: The ability to verify a users' identity prior to accepting an invite would allow a sender to verify and send messages to that user even if that user is offline at that time. If verification happens only after the invite (which is more privacy-friendly) one a sender needs to be online after the receiver accepted the invite and send messages in a second step.

This is why many messenger (including Signal, e.g.) only reveals profile information (like the display name and the avatar image) after accepting invitations.

Existing keys (master key, user-signing key, self-singing key) do not contain any identifying information (as opposed to an x509 certificate). This is why the master key and the self-singing key can be public but the **signatures** created by the user-signing key (revealing social graph) and the x509 cert (revealing user-identifying information) should not.

proposals/4316-x509-cross-signing.md
codedust: No, I didn't have any specific arguments for signing the SSK (option 2). I just wanted to present a complete list of possible options.

Personally, I'd prefer option 4 (the TLS way) since this option supports deniability. Signing the master key or device keys cryptographically binds the certificate to a user identity / the users' devices which seems unnecessary.

codedust: I dont quite get which problem exactly you want to solve.

> I had intended the certificate to be tied to an organization that accredits users / devices. So it wouldn't contain any personal information.

If we share the same definition of "accrediting", this would mean that one can only verify that a specific matrix account belongs to an organization but would not reveal any information about the specific user. But at the same time, you wrote, that it is in scope of this MSC to display the x509 subject (containing user-identifying information?) and issuer to other users.

Johennes: So our use case is the following: We have doctors working in health care institutions. The institutions operate Matrix home servers in a private federation. Each institution has a X.509 certificate issued by a national health agency. These certificates are de-facto public within the federation because they are already being used for all sorts of other things. Since the certificates are issued to the organization, they don't contain any personally identifiable information themselves.

As a patient communicating with a doctor, verifying them via the normal process isn't very helpful because I may have never met them in person before. Instead, the idea is to let the organization verify the doctor and then transitively extend that trust based on the certificate chain up to the national agency. In essence, this is similar to https://github.com/matrix-org/matrix-spec-proposals/pull/2882 (mentioned by Hubert below) but instead of expanding the trust through another user, we expand it through an external certificate.

Integrating the certificate into the Olm setup would be nice but doesn't work in our case because the private key is held by the institution rather than the user. 

Now, you're right that some amount of personal information still needs to be revealed because we were planning to use this proposal together with https://github.com/matrix-org/matrix-spec-proposals/pull/4317. The reason that this data was split off from the key signature is simply that `/keys/query` didn't seem like the right place to store profile information. We could devise a scheme to not store either and expose both only during verification. I acknowledge this would be better for privacy but it feels a little impractical because it requires both parties to be online.

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+        "signatures": {
+          "@bob:hospital.org": {
+            // Device self-signature
+            "ed25519:ABCDEFGHIJ": "...",
+            // Self-signing key signature
+            "ed25519:$SSK_BASE64": "..."
+          },
+          "external": {
+            // External X.509 signature
+            "x509:$FINGERPRINT_BASE64": "$SIGNATURE"
```
richvdh, line 53: we'd need to ensure that the format for the signature is clearly specified. My knowledge of the X.509 suite isn't great: would this be an (unpadded?) base64-encoded PKCS#1 signature?

proposals/4316-x509-cross-signing.md
richvdh: We used JWTs (obviously not the same as JWSes, but related) for [encrypted media](https://spec.matrix.org/v1.15/client-server-api/#extensions-to-mroommessage-msgtypes). Turns out that you need to be very precise about what each field has to be, and I'm not sure that there was much advantage to it. So I'd be looking for a specific reason to use a JWS.

Is it not the case that a JWS includes the payload (i.e. the signed object)? That seems inappropriate here, unless we want to replace the entire key object with a JWS.

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,133 @@
...
+  "device_keys": { ... },
+  ...
+  "external": {
+    "x509:$FINGERPRINT_BASE64": "$CERT.DER_BASE64"
+  }
+}
+```
+
+In closed federations, it is expected that clients are aware of a central PKI which obviates the
+need to expose the certificate in [`/_matrix/client/v3/keys/query`].
```
richvdh, line 78: I don't think it needs calling out explicitly, but I find it an interesting thought experiment to consider that, if the client already knows which keys are trusted, then we don't need the X.509 infrastructure at all: it's literally "just" a hardcoded public key which the client trusts to make signatures.

I suspect we want the hierarchical support that X.509 supplies anyway, as well as CRLs/OCSP and so on. But still.

proposals/4316-x509-cross-signing.md
Johennes: > We used JWTs (obviously not the same as JWSes, but related) for [encrypted media](https://spec.matrix.org/v1.15/client-server-api/#extensions-to-mroommessage-msgtypes). Turns out that you need to be very precise about what each field has to be, and I'm not sure that there was much advantage to it. So I'd be looking for a specific reason to use a JWS.

I think the problem is that (contrary to what the proposal currently says) the X.509 certificate alone doesn't fix the signature algorithm unless we add some additional restrictions. We could, for instance, say that only certs with Ed25519 keys are allowed. Or we could say that certs with RSA keys must use RS256 for signatures. JWSes would avoid that because they encode the concrete algorithm.

If we don't want JWSes and we also don't want to limit signature algorithms, maybe we could use a triplet structure to fix both the certificate and the algorithm?

```
"x509:$CERT_FINGERPRINT:$ALGORITHM": "$SIGNATURE"
```

> Is it not the case that a JWS includes the payload (i.e. the signed object)?

Normally yes but it looks like the [RFC](https://www.rfc-editor.org/rfc/rfc7515#appendix-F) also allows for a "detached" variant where the payload is simply erased after signing.

proposals/4316-x509-cross-signing.md
Johennes: Added in 5018d2ac6452972f39fcfa5b7424cb4a52f94c99.

proposals/4316-x509-cross-signing.md
Johennes: Fixed in 95e1c5477fc493d6aa0e89570068fea1c63d6790.

proposals/4316-x509-cross-signing.md
Johennes: This relates to the [thread above](https://github.com/matrix-org/matrix-spec-proposals/pull/4316/files#r2272622052) in so far as the certificate alone doesn't fully fix the signing algorithm. If we limited the algorithms, we could be more precise here. Otherwise, I think `$SIGNATURE` would have to be the unpadded base64 encoding of whatever bytes the signing algorithm produced.

proposals/4316-x509-cross-signing.md
richvdh: > JWSes would avoid that because they encode the concrete algorithm.

I think, even if we were using JWSes, we'd still need to specify exactly which key types we'd expect implementations to support, for compatibility reasons. And once we've done that, we kinda might as well specify the signature format.

> Normally yes but it looks like the [RFC](https://www.rfc-editor.org/rfc/rfc7515#appendix-F) also allows for a "detached" variant where the payload is simply erased after signing.

Interesting.

JWSes certainly do seem worth considering here. It might be worth trying to draft the words about what we'd actually need to specify here: in particular, which header parameters would be there and what they would look like,.

Johennes: I think for this proposal, we would need to use the JOSE header parameters in the following way:

**Required parameters**

- [`alg`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.1): The signing algorithm. IANA maintains a public [registry](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) of REQUIRED / RECOMMENDED / etc. algorithms.
- At least one of:
  - [`x5u`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.5): A URL to retrieve the signing certificate (chain).
  - [`x5c`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.6): The signing certificate chain as base64-encoded DER serializations.
  - [`x5t`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.7): The SHA-1 thumbprint of the signing certificate. The client needs an implementation-specific way to retrieve the certificate based on the thumbprint in this case.
  - [`x5t#S256`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.8): The SHA-256 thumbprint of the signing certificate. Ditto.

**Prohibited parameters**

- [`jku`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.2), [`jwk`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.3), [`kid`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.4): These are used to identify the signing key via JSON Web Keys. We won't need these because we'll identify the key via an X.509 certificate instead.
- [`cty`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.10): The content type of the signed data. We won't need this because the placement within the `/keys/query` response makes it clear what's being signed.
- [`crit`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.11): The headers that use critical extensions. We won't need this because we don't use any extensions at the moment.
- [`typ`](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.9): The JWS media type. We could use the values `JOSE` / `JOSE+JSON` here to indicate [JWS Compact Serialization](https://www.rfc-editor.org/rfc/rfc7515#section-7.1) / [JWS JSON Serialization](https://www.rfc-editor.org/rfc/rfc7515#section-7.2). It might be sufficient to just mandate a particular form though.

What I like about this option is that it allows for a lot of flexibility in distributing the certificate. For instance, the certificate could be inlined in the JWS which spares us from having to provide it via a top-level `external` object in `/keys/query`.

Since JWSes themselves don't carry IDs, we could just list them as an array under `signatures.external`:

``` json5
{
  "device_keys": {
    "@bob:hospital.org": {
      "ABCDEFGHIJ": {
        // Device identity & signing keys
        "keys": {
          "curve25519:ABCDEFGHIJ": "...",
          "ed25519:ABCDEFGHIJ": "..."
        },
        ...,
        "user_id": "@bob:hospital.org",
        "signatures": {
          "@bob:hospital.org": {
            // Device self-signature
            "ed25519:ABCDEFGHIJ": "...",
            // Self-signing key signature
            "ed25519:$SSK_BASE64": "..."
          },
          // External X.509 signatures
          "external": [$JWS1, $JWS2, ...]
```

The only alternative to this that I can currently think of is the triplet structure mentioned in https://github.com/matrix-org/matrix-spec-proposals/pull/4316#discussion_r2272622052.

```
"x509:$CERT_FINGERPRINT:$ALGORITHM": "$SIGNATURE"
```

This breaks with the existing `"alg:key": "signature"` pattern, too, though and allows for less flexibility.

uhoreg: FWIW, if you want to use a triplet, I would put the algorithm before the fingerprint:

```
"x509:$ALGORITHM:$CERT_FINGERPRINT": "$SIGNATURE"
```

proposals/4316-x509-cross-signing.md
```diff
@@ -0,0 +1,137 @@
...
+based signatures on cross-signing keys.
+
+## Proposal
+
+The `signatures` properties in the response of [`/_matrix/client/v3/keys/query`] are extended to
+allow a special key `external`. As this is not a valid [MXID], it cannot clash with other keys
+inside `signatures`. `external` is meant to store signatures that are tied to X.509 certificates in
+the same [format] already used for other signatures. Since in this case the certificate encapsulates
+both the signing algorithm and the public key, the generic string `x509` is used as the algorithm
+identifier while the certificate's SHA-256 fingerprint encoded in [unpadded base64] is used as the
```
uhoreg, line 27: Is the certificate's SHA-256 fingerprint enough for a user to find the certificate used to sign the identity?  Is it expected that the user already has a copy of the certificate that they can just look up by hash?  Or do they have to go through some process to download it?

proposals/4316-x509-cross-signing.md
codedust: I see. I assumed that the x509 cert was issued to the user.

Johennes:
It looks like gematik is planning to go ahead with a variant where cross-signing keys are included in signed profile data in a scheme akin to https://github.com/matrix-org/matrix-spec-proposals/pull/4317. I'll close this as I'm not aware of other planned use cases for this proposal.

PR #4317, opened 2025-07-29
MSC4317: Signed profile data

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/signed-profiles/proposals/4317-signed-profile-fields.md)

### DIFF
--- a/proposals/4317-signed-profile-fields.md
+++ b/proposals/4317-signed-profile-fields.md
@@ -106,7 +106,8 @@ still store it locally to be able to present the data to their user again upon r
 ## Security considerations
 
 If signed profile data is leaked during a security breach, the authenticity of the data can be
-verified by anyone which prevents deniability.
+verified by anyone which prevents deniability. Servers SHOULD mitigate this risk by limiting access
+to profiles using the mechanism from [MSC4170].
 
 ## Unstable prefix
 
@@ -123,3 +124,4 @@ This proposal depends on [MSC4316].
   [signing specification]: https://spec.matrix.org/v1.15/appendices/#signing-json
   [`/_matrix/client/v3/keys/query`]: https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3keysquery
   [QR-code verification]: https://spec.matrix.org/v1.15/client-server-api/#qr-codes
+  [MSC4170]: https://github.com/matrix-org/matrix-spec-proposals/pull/4170


proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+        "x509:$FINGERPRINT_BASE64": "$SIGNATURE"
+      }
+    }
+  }
+}
+```
+
+Clients SHOULD prefer the signed data if they are able to validate the signature. If they cannot
+validate the signature, they cannot establish any proof of the signed data's authenticity and SHOULD
+prefer the top-lvel properties regardless of whether there is any collision.
```
clokep, line 88: What if they differ?

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+
+Additionally, current clients won't know how to interpret signed profile data. For backwards
+compatibility home servers MAY, therefore, duplicate properties from `m.signed` to the top level.
+
+## Alternatives
+
+Rather than signing the key and profile data separately, they could be combined in an X.509
+certificate (or more generally a signature document). There is, however, no canonical place in
+Matrix to store something that is both a key and a profile.
```
clokep, line 97: It could be uploaded as media?

proposals/4317-signed-profile-fields.md
Johennes: My thinking was that if you cannot validate the signature, you have no reason to believe the signed data is legit and should rather ignore it entirely.

proposals/4317-signed-profile-fields.md
Johennes: Hm, that's true though I guess you'll then have to store the MXC URI somewhere? You could put it into the profile making it sort of an attachment. That would still mean storing keys in the profile though. I guess that wouldn't be a huge issue. It just seemed a little cleaner to me to use `/profile` and `/keys/query` given that they already exist.

proposals/4317-signed-profile-fields.md
clokep: And if it has a valid signature, but the info in the signed bit is different than the top-level fields... you just ignore the top level fields? Or do you warn or something about spoofing?

Johennes: Good question. A warning might not be a bad idea.

Alternatively, given that `m.signed` is meant to be managed by the home server, maybe the server could prevent conflicts to begin with? It could reject any modifications of top-level properties that already exist within `m.signed`.

proposals/4317-signed-profile-fields.md
turt2live, line 1: Implementation requirements:

* [ ] Client demonstrating use case (ideally multiple)

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+
+[MSC4133] introduced the ability to store custom data beyond `avatar_url` and `displayname` in user
+profiles. In some cases home servers might want to manage certain profile data. In a corporate
+setup, for instance, some employee data such as a person's role, location or telephone number might
+be sourced from a central HR system. The spec doesn't forbid managed profile properties and
+[MSC4133] allows users to understand which properties they can modify through the `m.profile_fields`
+capability. There is, however, no way to tell which properties are managed in *another* user's
+profile, especially when they're on a different home server. In environments where a user's trust
+into home servers is larger than their trust into other users this prevents users from identifying
+impersonation attacks.
```
codedust: What is the threat model here? If users trust in home servers, no signature would be required at all since the profile data is already authenticated via TLS already.

If neither users nor home servers are trusted, an external entity (PKI + third party receiving the signature certificate from that PKI) needs to be trusted.

Who holds the private key to the certificate in this scenario?

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+during device verification via a new verification method akin to [QR-code verification]. This,
+however, requires both clients to be online simultaneously, making (semi-)automated verification as
+introduced in [MSC4316] impossible. Additionally, it only marginally improves the security concern
+of leaking signed data because any client that received the certificate during verification would
+still store it locally to be able to present the data to their user again upon request.
+
+## Security considerations
+
+If signed profile data is leaked during a security breach, the authenticity of the data can be
+verified by anyone which prevents deniability.
```
codedust: What prevents users from collecting signed profile data from all users of a home server and leak/sell these information?

Selling personal data than is provably authentic is way more attractive than selling non-signed personal data.

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+    "user_id": "@alice:matrix.org",
+    "signatures": {
+      "external": {
+        // External X.509 signature
+        "x509:$FINGERPRINT_BASE64": "$SIGNATURE"
+      }
+    }
+  }
+}
+```
```
codedust, line 49: What's your use-case here? What's the reasoning behind signing roles and telephone numbers in this example?

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+
+Clients SHOULD prefer the signed data if they are able to validate the signature. If they cannot
+validate the signature, they cannot establish any proof of the signed data's authenticity and SHOULD
+prefer the top-lvel properties regardless of whether there is any collision.
+
+Additionally, current clients won't know how to interpret signed profile data. For backwards
+compatibility home servers MAY, therefore, duplicate properties from `m.signed` to the top level.
+
+## Alternatives
```
codedust, line 93: Other alternatives could include not signing the data ("`m.validated_by_home_server`" instead of `m.signed`) or using zero-knowledge proofs to validate user attributes without loosing deniability.

Cryptographers explicitly demand that they EU digital wallet (EUDI Wallet) should never exposes any signed data to the relying party but instead should use zero-knowledge proofs for this purpose (see https://github.com/eu-digital-identity-wallet/eudi-doc-architecture-and-reference-framework/discussions/211).

proposals/4317-signed-profile-fields.md
Johennes: My phrasing in the proposal is possibly fuzzy, apologies.

The scenario here is the same as in https://github.com/matrix-org/matrix-spec-proposals/pull/4316#discussion_r2256902764. The server admin holds the private key. The certificate that carries the public key was issued by a national agency.

I think TLS isn't enough here because it doesn't protect against a rogue user claiming to be somebody else by modifying their profile.

proposals/4317-signed-profile-fields.md
Johennes: Phone number is maybe a stretch here but, as mentioned in https://github.com/matrix-org/matrix-spec-proposals/pull/4317/files#r2256919100, the threat is somebody modifying their profile to look as if they were somebody else.

proposals/4317-signed-profile-fields.md
Johennes: This proposal can be combined with https://github.com/matrix-org/matrix-spec-proposals/pull/4170 to limit access to profiles. This should probably be added to the proposal.

proposals/4317-signed-profile-fields.md
Johennes: > Other alternatives could include not signing the data ("m.validated_by_home_server" instead of m.signed) [...]

How would you know that `m.validated_by_home_server` was inserted by the server rather than the user themselves?

> [...] or using zero-knowledge proofs to validate user attributes without loosing deniability.

This sounds interesting though I'm struggling a bit to see how it would apply here. We want to actually expose the information, not just prove possession or do some calculation on it. Is that possible with zero-knowledge proofs?

proposals/4317-signed-profile-fields.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Proposal
+
+In conjunction with [MSC4316], a new profile property `m.signed` is introduced which is an object
+with the following mandatory properties:
+
+- `user_id`: The account's user identifier.
+- `signatures`: Signatures of the `m.signed` value as per the current [signing specification].
+
+Any other properties in `m.signed` are optional and extend the data under signature.
```
uhoreg, line 30: Just thinking out loud: I wonder if there would ever be a use case for having some profile data signed by one entity, and some other profile data signed by another entity.  We probably don't want to overcomplicate this, if such a situation is unlikely.

proposals/4317-signed-profile-fields.md
Johennes: I have rephrased the introductory section with 5a053a9448513079888d73560be144131703cfd6 to make the threat model more clear.

proposals/4317-signed-profile-fields.md
Johennes: I updated the example to match the medical use case described in the introduction with 25c2b03a5315ec3a719ef1cf91a8ae7ab4ace3b7.

proposals/4317-signed-profile-fields.md
Johennes: I have added this into the proposal with 98dc850093fc13f721466cd645a9c97f9ee921df.

proposals/4317-signed-profile-fields.md
Johennes: I don't have a concrete use case on my mind right now but this does indeed sound like something we may want to enable if it's easy. Turning `m.signed` into an array seems straightforward. The only possible problem I can think of is if the same key occurs in multiple signed sub-profiles. For instance, you could have `displayname` in two separately signed chunks and with different values. This is basically an extension of the problem @clokep raised in https://github.com/matrix-org/matrix-spec-proposals/pull/4317/files#r2239842049.

proposals/4317-signed-profile-fields.md
codedust: The idea of `m.validated_by_home_server` would be that this attribute can only be set by the server.

In case you want to expose the information, a zkp is not the ideal tool here.

From what I understood about you use-case in https://github.com/matrix-org/matrix-spec-proposals/pull/4316#discussion_r2256902764, it seems sufficient to provide authenticated profile data by the healthcare institution (authenticated == MAC as opposed to signed == asymmetric signature). One of the easiest ways would be to use the X.509 certificate issued by a national health agency as a TLS cert in an external matrix user directory of the healthcare institution (e.g. just a simple nginx setup) and just let the matrix clients ask the server via TLS about user identities. This way, the clients could validate user identities against the healthcare institutions' X509 cert without any availability requirements for other clients.

Johennes:
It appears that the intricacies of the gematik ecosystem will make it virtually impossible to motivate and architect this proposal in a way that is generic enough for the wider Matrix world yet specific enough to be reused in gematik's own spec. Therefore, I'm closing this for the time being and gematik can go ahead with their own custom extension of the Matrix spec in this case.

PR #4318, opened 2025-07-30
MatrixRTC proposal | Update to-device key event format + cleaning

BillCarsonFr:


Johennes:
Most of this appears to already be reflected in #4143 by now. I have carried over the bits around encryption and forward secrecy that were missing in 5bcd4f0bcbcaf9a2602164ea732d08b43066085f.

PR #3943, opened 2022-12-02
MSC3943: Partial joins to nameless rooms should include heroes' memberships

DMRobertson:
Per https://github.com/matrix-org/matrix-spec-proposals/pull/3706/files#r1023076655 and https://github.com/matrix-org/synapse/issues/12995#issuecomment-1290607134.

[Rendered.](https://github.com/DMRobertson/matrix-spec-proposals/blob/dmr/heroes-in-faster-joins/proposals/3943-heroes-in-partial-send-join-response.md#user-content-fnref-1-a7b72ecdddc9e77d54f2dd1f378ca8f2)

Implemented in https://github.com/matrix-org/synapse/pull/14442

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3943#issuecomment-1381766519)

PR #4319, opened 2025-08-01
MSC4319: Room member events for invite and knock rooms in the `/sync` response

zecakeh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4319-room-member-events-for-invite-knock-sync-response.md)

This was spawned by the discussion in https://github.com/matrix-org/matrix-spec/pull/2181.

Server Implementations:

- Synapse:
    - Own non-stripped `m.room.member` in `invite_state`:  https://github.com/element-hq/synapse/blob/458e6410e8388fbc6b60b5575850405939bc53eb/synapse/rest/client/sync.py#L450-L462
    - Stripped inviter's `m.room.member` in `invite_state`:  https://github.com/element-hq/synapse/blob/a6e326582f735e03e420c3b86475388fb5f2656c/synapse/handlers/message.py#L1835-L1839
    - Own non-stripped `m.room.member` in `knock_state`: https://github.com/element-hq/synapse/blob/458e6410e8388fbc6b60b5575850405939bc53eb/synapse/rest/client/sync.py#L486-L514
    - Add the `state` key to invite and knock rooms: https://github.com/element-hq/synapse/pull/19330
- Dendrite:
    - Own non-stripped `m.room.member` in `invite_state`:  https://github.com/element-hq/dendrite/blob/8d2da78744387e55c28bb8925ae0cc70dd3e02e3/syncapi/types/types.go#L530-L574
    - Does not seem to add the inviter in `invite_state`.  
    - Does not have support for `knock_state`.
- Conduit (and forks):
    - Own stripped `m.room.member` in `invite_state` (common to all forks):  https://gitlab.com/famedly/conduit/-/blob/ed5b0514f52f848e53785dee5fe8de63780cf092/src/api/server_server.rs#L2160-2172
    - Stripped inviter's `m.room.member` in `invite_state` (Conduit): https://gitlab.com/famedly/conduit/-/merge_requests/766  
    - Own non-stripped `m.room.member` in `knock_state` (Conduit): https://gitlab.com/famedly/conduit/-/blob/ed5b0514f52f848e53785dee5fe8de63780cf092/src/api/server_server.rs#L2160-2172

Most clients already rely on being able to access the `m.room.member` events of the knock or invite process, and of the inviter.

The following client implementations also rely on being able to access the `origin_server_ts`:

- Cinny: https://github.com/cinnyapp/cinny/blob/67b05eeb09c8019fc67564f65ebca3d4b37653c6/src/app/pages/client/inbox/Invites.tsx#L88
- Neochat: https://invent.kde.org/network/neochat/-/blob/c5caffcdf9b250eb5a6c57a40ab951fa188563d1/src/libneochat/neochatroom.cpp?page=2#L1661

This client implementation supports the new `state` key:
- Ruma: https://github.com/ruma/ruma/pull/2322

----

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4319#issuecomment-4113786727)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4319#issuecomment-4113794876)


proposals/4319-non-stripped-member-in-stripped-state.md
```diff
@@ -0,0 +1,69 @@
...
+implementations and that clients already rely on._
+
+## Proposal
+
+The `m.room.member` event that was created during the invite or knock process MUST be present in
+the `events` of the `invite_state` or `knock_state`. Since servers have access to the whole event,
+and for clients to be able to access the time of the event via the `origin_server_ts` field, it
+MUST use the same format as can be found in the `events` of `timeline` and `state` for rooms under
+`join`. This is the format currently defined as
+[`ClientEventWithoutRoomID`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync_response-200_clienteventwithoutroomid).
```
turt2live: Alternatively, because clients want access to the `origin_server_ts` and generally speaking we don't want different formats for data in a list, this MSC could do two things:

1. Add `origin_server_ts` to *all* stripped state events
2. Add the inviting `m.room.member` event to the list of suggested state

zecakeh: I also found a use case for accessing `unsigned` of an invite `m.room.member` event, for clients to be able to know if the `invite` is after a `knock` by looking into the `prev_content`.

turt2live: I'm not seeing why that would be important to them, or why that needs to come from `unsigned` instead of somewhere else

zecakeh: To auto-accept an invite following a knock for example. If it's a new device there is no other place where this information can be found.

zecakeh: I changed the stripped state event format to include `origin_server_ts` and `unsigned` (since this one is already optional in other event formats anyway).

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+```
+
+Finally, the list of events that should be included in the stripped state is extended with the
+stripped `m.room.member` event of the `sender` of the invite. This allows clients to be able to
+display information about the sender of an invite, like their display name or avatar.
+
+
+## Potential issues
```
Gnuxie, line 173: Clients have depended on synapse's behaviour of including the full client formatted event for the `m.room.member` invite event.  

https://github.com/the-draupnir-project/Draupnir/issues/945

In draupnir this dependency is more than just a parsing issue. The matrix-protection-suite provides a handle called `handleExternalMembership` which invitations are fed into from the matrix-protection-suite's conception of the timeline, which is closer to appservice push than `/sync`. As appservice push also pushed invitations this way.

Deduplication and buffering is then coordinated using the event id as it is identical to any other client formatted event and not special cased. So this is a deep dependency.  

https://github.com/Gnuxie/matrix-protection-suite/blob/a85102c0aaa6ba73beae8fdb35600eec0999eada/src/Protection/Protection.ts#L176
https://github.com/the-draupnir-project/Draupnir/blob/main/src/protections/invitation/JoinRoomsOnInviteProtection.tsx#L86-L96
https://github.com/the-draupnir-project/Draupnir/blob/f1dad52288ff32a096cbd3ff5be110648574c51f/src/Draupnir.ts#L355
https://github.com/Gnuxie/matrix-protection-suite/blob/453e55b182282aad59db430f53549d89541a8e21/src/ClientManagement/StandardClientRooms.ts#L95-L126
https://github.com/the-draupnir-project/Draupnir/blob/f1dad52288ff32a096cbd3ff5be110648574c51f/src/DraupnirBotMode.ts#L97-L99
https://github.com/the-draupnir-project/Draupnir/blob/f1dad52288ff32a096cbd3ff5be110648574c51f/src/appservice/AppService.ts#L355

Gnuxie, line 173: The change complicates clients that allow either `/sync` or appservice push to be their event source. Since now invitations sourced from `/sync` will be a different format to those provided in appservice push. See https://github.com/matrix-org/matrix-spec-proposals/pull/4319#discussion_r2330206182.

tulir: Additionally, the appservice event can't be stripped, because then you'd have two different data formats in the `events` array of `/transactions`

In this specific context, the AS API is closer to S-S than C-S /sync, as both AS and S-S send full invite events with stripped state in `unsigned` -> `invite_room_state`, while /sync flattens invite event + stripped state into an `invite_state` array

proposals/4319-room-member-events-in-stripped-state.md
```diff
@@ -0,0 +1,114 @@
...
+- `m.room.create`
+- `m.room.name`
+- `m.room.avatar`
+- `m.room.topic`
+- `m.room.join_rules`
+- `m.room.canonical_alias`
+- `m.room.encryption`
+
+Although these events are useful to be able to present information about the room, they don't
+contain information about the event itself.
```
MadLittleMods: Can be punted to another MSC but here is another piece of missing info with `invites` necessary to calculate the room name:

> One weird corner with the current spec: If the room doesn't have a name set and the user is invited/knock to the room, we don't have any `heroes` to provide here. This is because `heroes` membership isn't included in the stripped state that the server receives when they are invited/knocked over federation.
> 
> Related to https://github.com/matrix-org/matrix-spec/issues/380
> 
> This was solved for partial-joins via [MSC3943](https://github.com/matrix-org/matrix-spec-proposals/pull/3943)
> 
> Not necessarily a blocker but just calling out something that won't work completely until that spec issue is solved.
>
> *-- @MadLittleMods, https://github.com/matrix-org/matrix-spec-proposals/pull/4186#discussion_r1745991652*

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: I have changed the proposal in https://github.com/matrix-org/matrix-spec-proposals/pull/4319/commits/1b81df4d945bdf92b903fd78326a7a6d3dba1708 to send the full event in a different place to attempt to take care of two concerns at once: the event formats should not be mixed in `invite_state`, and implementations rely on all fields of the full event format.

proposals/4319-room-member-events-in-stripped-state.md
zecakeh: I updated the proposal in https://github.com/matrix-org/matrix-spec-proposals/pull/4319/commits/1b81df4d945bdf92b903fd78326a7a6d3dba1708 to mention this, but this is left as out-of-scope of this MSC.

proposals/4319-room-member-events-in-stripped-state.md
```diff
@@ -0,0 +1,181 @@
...
```
richvdh:
```suggestion
contain information about the invite/knock event itself.
```

proposals/4319-room-member-events-for-invite-knock-sync-response.md
richvdh: It feels like including the event in stripped format in `invite_room_state` doesn't satisfy anyone, if clients are relying on having full events there? It neither provides backwards-compatibility, nor gives us a clean API. We're just going to end up continuing to be in a mess.

I think I'd suggest a fairly aggressive timeline like
 * have server implementations continue to include the full event in `{invite,knock}_room_state` as a temporary, *strictly time-limited* measure
 * clients that are currently relying on the unspecified behaviour having the invite or knock event in `{invite,knock}_room_state` switch to using `state` (with fallback to `{invite,knock}_room_state`, if they want compat with older servers)
 * after a few months, servers remove the event from `{invite,knock}_room_state`. Clients still relying on the unspecced behaviour are out of luck.

Interested to hear thoughts from others on this, though

proposals/4319-room-member-events-in-stripped-state.md
```diff
@@ -0,0 +1,181 @@
...
+> [!NOTE]
+> Making it mandatory makes sense because this event is the reason why the room appears in `invite`
+> or `knock` in the first place. Providing the full event format allows clients to access details
+> like the `origin_server_ts`, the `event_id` or the `unsigned` object.
+
+For compatibility with the current client implementations, homeservers SHOULD also include this
+event in the `events` array of the `invite_state` or `knock_state` in stripped format.
+
+Clients SHOULD expect the `state` key to be missing and SHOULD look for the `m.room.member` event in
+`invite_state` or `knock_state` as a fallback.
```
richvdh: per my comment [on the other thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4319#discussion_r2330222136), this feels a bit weak, and is just propagating the status quo.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,181 @@
...
+Clients SHOULD expect the `state` key to be missing and SHOULD look for the `m.room.member` event in
+`invite_state` or `knock_state` as a fallback.
+
+> [!NOTE]
+> The example for the response of `GET /sync` already includes the stripped `m.room.member` event
+> although it is not specified.
+
+Finally, the list of events that should be included in the stripped state is extended with the
+stripped `m.room.member` event of the `sender` of the invite. This allows clients to be able to
+display information about the sender of an invite, like their display name or avatar.
```
richvdh: In federation contexts, I don't think the server has the `sender` of the invite? We shouldn't include it unless it's going to work reliably.

proposals/4319-room-member-events-in-stripped-state.md
```diff
@@ -0,0 +1,181 @@
+# MSC4139: Room member events in stripped state
```
richvdh: It strikes me that the current proposal is not really "Room member events in stripped state"; rather, it is "add new fields [which are independent of stripped state] to the `/sync` response".

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: The list of events that should be included in the stripped state in the spec applies to the CS and SS APIs alike. This is asking for the `m.room.member` event of the sender to be sent in the `invite_room_state` and `knock_room_state` over federation and forwarded in stripped state to the client.

richvdh: oh, I see, right. It would probably be good to clarify that in the MSC text.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: This proposal has been updated to include the full event again in 1b81df4d945bdf92b903fd78326a7a6d3dba1708.

zecakeh: I agree with this suggestion, 3bb2ab37db4c3e2226339fc9751311261af05d23 adds a time limit of 1 spec release to the compatibility behavior.

proposals/4319-room-member-events-in-stripped-state.md
zecakeh: Applied in e2791191a774e0714ee46f578fc0dd182d786a1c.

proposals/4319-room-member-events-in-stripped-state.md
zecakeh: This has been relaxed as a MAY for compatibility with current server implementations rather than a SHOULD in https://github.com/matrix-org/matrix-spec-proposals/pull/4319/commits/3bb2ab37db4c3e2226339fc9751311261af05d23

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: Clarified in 3fb1fc924586cfa64fa8dd9a111daa847a99a685

proposals/4319-room-member-events-in-stripped-state.md
zecakeh: The core of the proposal is still to have access to the `m.room.member` events in invite and knock rooms, but I updated the title to not mention the stripped state in 98a7a3e3fddf8d4893175f1f9bc79f802424892c

zecakeh:
There are implementations for the latest state of the MSC in:

- Synapse: https://github.com/element-hq/synapse/pull/19330
- Ruma: https://github.com/ruma/ruma/pull/2322

turt2live (approved):

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
This proposal looks ready for broader SCT review, at priority 2:

@mscbot fcp merge

proposals/4319-room-member-events-in-stripped-state.md
MadLittleMods: Created https://github.com/matrix-org/matrix-spec/issues/2353 so this idea isn't lost

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,204 @@
...
+For compatibility with the current server implementations, clients MAY look for the `m.room.member`
+event in `invite_state` or `knock_state` as a fallback if it is not found under the `state` key.
+
+> [!NOTE]
+> The example for the response of `GET /sync` includes the stripped `m.room.member` event although
+> it is not specified.
+
+Finally, the [list of events that should be included in the stripped state][stripped-state] over the
+Client-Server and Server-Server APIs is extended with the `m.room.member` event of the `sender` of
+the invite. This event has the same format as other events in the stripped state, i.e. the full
```
anoadragon453: > extended with the `m.room.member` event of the `sender` of the invite.

And... the `sender` of the knock as well?

It's odd that this section only talks about invites, whereas the rest of the proposal mentions invites and knocks in tandem.

anoadragon453 (commented):
LGTM other than one small clarification.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: The sender of a knock can only be the target user (according to the authorisation rules) so it would be the same event as in `state`. That's why this part doesn't mention it.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,204 @@
...
+> This MSC doesn't change the format of this event over the Server-Server API. This only proposes to
+> forward more information to clients that has always been available to homeservers.
+
+For compatibility with the current client implementations, homeservers SHOULD also continue to
+include this event in the `events` array of the `invite_state` or `knock_state` in the same format
+as in `State` for a time limited to 1 spec release after this proposal is released in a new spec
+version.
+
+For compatibility with the current server implementations, clients MAY look for the `m.room.member`
+event in `invite_state` or `knock_state` as a fallback if it is not found under the `state` key.
```
richvdh: Minor clarification, since there are multiple `m.room.member` events being discussed in this MSC

```suggestion
For compatibility with the current server implementations, clients MAY look for the knock/invite `m.room.member`
event in `invite_state` or `knock_state` as a fallback if it is not found under the `state` key.
```

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+## Proposal
+
+For clients to be able to get all the details about an invite or knock, a `state` key is added to
+the [`InvitedRoom`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync_response-200_invited-room)
+and [`KnockedRoom`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync_response-200_knocked-room)
+objects. It uses the same format as the [`State`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync_response-200_state)
+object in [`JoinedRoom`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync_response-200_joined-room)
+and MUST include the `m.room.member` event that was created during the invite or knock process, in
```
richvdh, line 57: Are we expecting it to contain *only* that event? Would be good to say so.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
richvdh: suggest adding "for rooms that the user has been invited to" somewhere in this sentence.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,204 @@
...
+version.
+
+For compatibility with the current server implementations, clients MAY look for the `m.room.member`
+event in `invite_state` or `knock_state` as a fallback if it is not found under the `state` key.
+
+> [!NOTE]
+> The example for the response of `GET /sync` includes the stripped `m.room.member` event although
+> it is not specified.
+
+Finally, the [list of events that should be included in the stripped state][stripped-state] over the
```
dbkr: I'm a bit confused I'm afraid. Am I right in saying that this part is added because it's what Synapse currently does, but the proposal is that having the format with the event ID is useful and so it's proposing to add it to a new field called `state` where it has the event id? If so, I think it would be beneficial to just say this in the MSC.

But if so, this also proposes that server only send this for 1 further release after this, so why add it to the spec?

zecakeh: This paragraph is about a different `m.room.member` event than before. Up to this point the changes were about including the `m.room.member` with an `invite` or `knock` membership in `state`.

However for a membership of `invite` there are 2 users involved: the target user (the **invitee**) and the sender (the **inviter**). We already have the `m.room.member` event of the target user in `state`, but this is about adding _also_ the `m.room.member` event of the _sender_ of the invite event.

The local homeserver might not have this event if it's not already in the room so it must be provided by the remote homeserver in the stripped state.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: Added in 8d67bda70f4e7dae067b8bb77bd1fafaf2bd3197.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: Applied in 8d67bda70f4e7dae067b8bb77bd1fafaf2bd3197.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: I added a sentence for this in 8d67bda70f4e7dae067b8bb77bd1fafaf2bd3197.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: I added subsections in 8c44d6abefbbdfd42b9e9ab35e7a727a87f03bae to make this clearer hopefully.

dbkr: ah okay, thanks for clarifying!

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,220 @@
...
+      },
+    ]
+  }
+}
+```
+
+### Dependencies
+
+Although it is not considered a dependency of this MSC, if this is accepted the new `/sync`
+endpoint introduced in [MSC4286: Simplified Sliding Sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4186)
```
tulir, line 167: Is this dependencies section intentionally separate from the other dependencies section at the bottom? Seems a bit weird to have 2

zecakeh: I believe that it's more the title of this section that could be improved, because the text says that it's not a dependency. Maybe it should be renamed to "Changes to other MSCs" or something?

anoadragon453: "Changes to other MSCs" sounds sensible to me.

zecakeh: Done in 63aebd553260ce37ccc0f5707e625a9d764b8043.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,220 @@
...
+We could add more fields to the stripped state format, but given all the fields that are needed for
+the different use cases, it would mean that the stripped state has the same format as normal state.
+Using the full event format might give the wrong idea that this state has been validated by the
+homeserver, which is currently not possible
+(see [this discussion in MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2274781824)).
+Besides, those fields are really only necessary for the `invite` or `knock` `m.room.member` event.
+
+This doesn't solve the case where a room doesn't have an `m.room.name` or `m.room.canonical_alias`
+state event, so [its display name should be computed using the room summary](https://spec.matrix.org/v1.15/client-server-api/#calculating-the-display-name-for-a-room).
+This is left to another MSC.
```
MadLittleMods, line 203: This can be cross-linked to https://github.com/matrix-org/matrix-spec/issues/2353

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,220 @@
...
+
+We could put the full `m.room.member` event under another key in the `invite_state` or
+`knock_state`, the `state` key was chosen for its similarity with other room objects. It will also
+allow to add more events using their full format in the future if needed.
+
+We could add more fields to the stripped state format, but given all the fields that are needed for
+the different use cases, it would mean that the stripped state has the same format as normal state.
+Using the full event format might give the wrong idea that this state has been validated by the
+homeserver, which is currently not possible
+(see [this discussion in MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2274781824)).
```
MadLittleMods, line 197: This equally applies to the new `state` with full events as well. The `invite` membership event is not "validated" (checked to pass auth rules).

Related reading on [`outliers`](https://github.com/element-hq/synapse/blob/3622579e7bc9fe70d3682bc20bb4955fbb0c9e82/docs/development/room-dag-concepts.md#outliers) and [`out-of-band membership events`](https://github.com/element-hq/synapse/blob/3622579e7bc9fe70d3682bc20bb4955fbb0c9e82/docs/development/room-dag-concepts.md#out-of-band-membership-events) in the Synapse docs.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,220 @@
...
+
+For compatibility with the current server implementations, clients MAY look for the invite/knock
+`m.room.member` event in `invite_state` or `knock_state` as a fallback if it is not found under the
+`state` key.
+
+> [!NOTE]
+> The example for the response of `GET /sync` includes the stripped `m.room.member` event although
+> it is not specified.
+
+### `m.room.member` event of the `sender` of the invite
```
MadLittleMods, line 83: I feel like this MSC should solely propose this change on its own: Include the `m.room.member` for the invite/knock event itself AND `m.room.member` event for the `sender` of the invite in stripped state.

All of the extra complexity around clients wanting access to `event_id`, `origin_server_ts`, or even full state events is a separate topic. I don't believe the complexity around `state` vs `invite_state` is a great approach.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
```diff
@@ -0,0 +1,220 @@
...
+> like the `origin_server_ts`, the `event_id` or the `unsigned` object.
+
+> [!IMPORTANT]
+> This MSC doesn't change the format of this event over the Server-Server API. This only proposes to
+> forward more information to clients that has always been available to homeservers.
+
+For compatibility with the current client implementations, homeservers SHOULD also continue to
+include this event in the `events` array of the `invite_state` or `knock_state` in the same format
+as in `State` for a time limited to 1 spec release after this proposal is released in a new spec
+version.
```
MadLittleMods, line 73: What other clients have relied on the full state of the `invite`/`knock` being available?

The only [example given](https://github.com/matrix-org/matrix-spec-proposals/pull/4319#discussion_r2330206182) so far has been [Draupnir](https://github.com/the-draupnir-project/Draupnir) wanting `event_id`.

zecakeh: In the description of the PR there are a couple of clients accessing the `origin_server_ts`.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: The purpose of this MSC is to specify behavior that has already been implemented for a long time in some homeservers and clients while not being part of the spec.

It turns out that the current implementations rely on an undesirable API (like [mixing event formats](https://github.com/matrix-org/matrix-spec-proposals/pull/4319#discussion_r2263655381)) so the extra complexity is needed to have a more consistent API at the cost of requiring changes in implementations.

I don't believe that wanting to access the extra fields is a separate topic because it is all about clients having enough data to present/handle invites and knocks. Without the extra fields, the `m.room.member` for the invite/knock event itself isn't very useful because it doesn't really provide more data (we can already infer the `membership`).

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: This doesn't say that the event that we place in `state` was validated though. It just explains why we should avoid _as much as possible_ to provide the full event format.

Maybe this should be clarified in the spec PR for this MSC?

proposals/4319-room-member-events-for-invite-knock-sync-response.md
zecakeh: Done in f67380514f8564cb564e22d23749ffbd6742aa83.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
MadLittleMods: Using full state event in `state` has the same drawbacks of using the full state events in `invite_room_state`.

I'm not seeing the benefit of the complexity around having both `state` and `invite_room_state` (which will be hard to understand from a newcomers perspective) when we could just use full events in `invite_room_state` with some better notes in the spec.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
deepbluev7: This seems like much too short of a timeline. I think quite a few clients rely on this event being in the invite_state, even if it is stripped. Removing that after 3 months, when until then clients couldn't rely on it being elsewhere and breaking every client someone might have gotten from a distro repo like Ubuntu, is a bit problematic. But I guess servers will have to support it in both places for longer, if they support an old spec version?

proposals/4319-room-member-events-for-invite-knock-sync-response.md
deepbluev7: Yeah, this introduces additional complexity, that doesn't seem that beneficial. It would be much easier and more backwards compatible to simply have those events in the invite_state as before.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
MadLittleMods: The stripped state version does provide value: The invite is useful to see the invite `reason` and `sender` and including the `m.room.member` of the `sender` gives you access to the display name/avatar of the person that invited you.

I would consider the current Synapse behavior flawed and including the *full* state event probably was a simple oversight (https://github.com/element-hq/synapse/commit/ab9cf732585244781ba67f4bb4c235ded3d4661a).

The only examples of clients that use this extra information are progressive enhancement examples. We can equally instead fix the spec and the Synapse implementation to do the right thing without the clients losing much at all for spec-compliant homeservers. The desire for more info is valid and can be separately advanced in its own MSC that adds the necessary metadata (whatever solution that may be).

proposals/4319-room-member-events-for-invite-knock-sync-response.md
turt2live: The version of the spec matters less here. The timeline is probably a bit short, but as a `SHOULD` there's room for servers to make the best decision for their local ecosystems. Some servers may be able to move faster, some slower - this MSC sets a guideline for what to aim for.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
turt2live: The SCT has previously expressed interest in consistency: either return full state events, or don't. There should not be a mix of formats. The SCT has also expressed that continuing to call it "stripped state" when it contains fully formatted events is wrong. There's also mild opinions around naming it specifically "[invite|knock]_state" when it's the same thing regardless of membership type - it doesn't need to be overly qualified. 

These considerations are mentioned somewhere in the depths of https://github.com/matrix-org/matrix-spec-proposals/pull/4311, alongside other arguments not relevant to this proposal.

Whether the information is valuable or not is somewhat irrelevant - the format should be consistent, and getting the information needed from an `m.room.member` event means it's best formatted as a real event. Therefore, all previously-stripped state events should be formatted as real events too.

proposals/4319-room-member-events-for-invite-knock-sync-response.md
turt2live: The original concern of this thread regarding validation of events has a long and storied history. The conclusion is validation isn't needed right now because the data remains untrusted by nature. If/when a future MSC includes the full recursive auth chain, then servers can verify the invite.

For the concern of having two fields doing different things: another MSC to deprecate `[invite|knock]_state` is welcome. This MSC would not be the one to make that change. If that future MSC were to move quickly enough through the process, I'd suggest that the spec PR for this MSC and that future MSC incorporate both changes (potentially delaying this MSC's merge date).

proposals/4319-room-member-events-for-invite-knock-sync-response.md
MadLittleMods: It is valuable to confirm that this half-baked state that this MSC ships is not great.

I still think this MSC should have simply defined the extra `m.room.member` events in stripped state and then another MSC would propose returning full state events for the further use cases.

It sounds like the way forward now that this MSC is merged is to just define that all stripped state events should be presented as full state events in `state`. Based on https://github.com/matrix-org/matrix-spec-proposals/pull/4319#discussion_r3662355833, perhaps even deprecating `invite_state`/`knock_state`

turt2live: That would be my preferred direction, as an individual. 

Introducing `m.room.member` as stripped state provides none of the useful information to clients while the full event does. We are currently opposed to mixing formats within the same structure, so a new field is required.

MadLittleMods: > Introducing `m.room.member` as stripped state provides none of the useful information to clients

(it does, see "The stripped state version does provide value:" above)

There are other use cases it doesn't cover but those are the things to be covered by the further MSC.

PR #4320, opened 2025-08-02
MSC4320: Rich Presence

88572:
[Rendered](https://github.com/88572/matrix-spec-proposals/blob/rich-presence/proposals/4320-rich-presence.md)

### DIFF
--- a/proposals/4320-rich-presence.md
+++ b/proposals/4320-rich-presence.md
@@ -13,8 +13,10 @@ Required fields if `type` is `m.rpc.media`:
 - `artist` containing the artist.
 - `album` containing the album.
 - `track` containing the track name.
+
 Required fields if `type` is `m.rpc.activity`:
 - `name` containing the activity name.
+
 Optional fields if `type` is `m.rpc.media`:
 - `progress` containing `length` with the length of the track in seconds.
 	- `time_complete` is needed for clients to show the media progress. It should only be updated in the event of the timestamp not being able to be guessed by a client.
@@ -22,6 +24,7 @@ Optional fields if `type` is `m.rpc.media`:
 - `cover_art` containing a MXC with the album art. Clients should hash the album art and re-use the MXC for album art matching that hash to avoid duplicating uploads.
 - `player` containing the media player.
 - `streaming_link` containing a link to the song on streaming services (i.e. Spotify, YouTube). Clients should warn the user before opening the link.
+
 Optional fields if `type` is `m.rpc.activity`:
 - `image` containing an MXC to show as an activity image. Clients should hash the image and re-use the MXC for album art matching that hash to avoid duplicating uploads.
 - `details` containing details (i.e. current map, location, etc).
@@ -31,11 +34,11 @@ Media:
 ```json
 {
 	"avatar_url": "…", "displayname": "…",
-    "m.rpc": {
+	"m.rpc": {
 	    "type": "m.rpc.media",
 	    "progress"{
-		    "length": 204,
-		    "complete": 102,
+		"length": 204,
+		"complete": 102,
 	    },
 	    "artist": "ari melody",
 	    "album": "free2play",
@@ -43,18 +46,18 @@ Media:
 	    "cover_art": "mxc://ip-logger.com/YS5tqBewwZ4HFF3hU9KT8OskREUmlPfM",
 	    "player": "Spotify",
 	    "streaming_link": "https://open.spotify.com/track/1zJWGrUJ7dy7wQuMSVbvCn"
-    }
+	}
 }
 ```
 Games:
 ```json
 {
 	"avatar_url": "…", "displayname": "…",
-    "m.rpc": {
-	    "type": "m.rpc.activity",
+	"m.rpc": {
+		"type": "m.rpc.activity",
 		"name": "SuperTuxKart",
 		"details": "Playing Oliver's Math Class as Xenia on Intermediate"
-    }
+	}
 }
 ```
 ### Potential implementations
--- a/proposals/4320-rich-presence.md
+++ b/proposals/4320-rich-presence.md
@@ -1,0 +1,71 @@
+# MSC4320: Rich Presence
+
+# Context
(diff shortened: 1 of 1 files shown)


proposals/4320-rich-presence.md
tulir, line 1: Implementation requirements:

* [ ] Matrix client receiving activity data and sending it to the homeserver
* [ ] External app sending activity data to the local Matrix client

Jordan-b1027:
could be turned on or off from people who dont want it

ryotairi:
Implemented rendering rich presences in [Extera Next](https://github.com/ExteraApp/Extera) latest commit

proposals/4320-rich-presence.md
```diff
@@ -0,0 +1,74 @@
...
+
+Optional fields if `type` is `m.rpc.media`:
+- `progress` containing `length` with the length of the track in seconds.
+	- `time_complete` is needed for clients to show the media progress. It should only be updated in the event of the timestamp not being able to be guessed by a client.
+	- Both of these are highly recommended to be added.
+- `cover_art` containing a MXC with the album art. Clients should hash the album art and re-use the MXC for album art matching that hash to avoid duplicating uploads.
+- `player` containing the media player.
+- `streaming_link` containing a link to the song on streaming services (i.e. Spotify, YouTube). Clients should warn the user before opening the link.
+
+Optional fields if `type` is `m.rpc.activity`:
```
Airyzz, line 28: Maybe including a timestamp `start_time` and `expire_time`, if a user goes offline without clearing their status, the `expire_time` could be used to not render the activity

IchiiDev:
I would like to suggest the possibility of enabling third party services to stream rich presence data without the need to go through the client, this is something I thought lacked from Discord's RPC feature. 

For instance, I'm listening to my Jellyfin server on my phone using a client that does not support Rich Presence on phone, or a Desktop client that doesn't support matrix: I make a small program that runs somewhere, authenticated to my account and that syncs data from my Jellyfin server to the matrix RPC directly, without the need to have my computer running with matrix open. 

There is probably no need to change the proposal much as OAuth2 flow already exists in the matrix documentation and it could be used to authenticate on the `PUT/DELETE /_matrix/client/v3/profile/{userId}/m.rpc` route on the server. This would fix the need to have a client installed and provide a way for compatible apps to implement Rich Presence anyway.

Another change would be to allow presence concurrency, so that an app running on the client can decide to not override the existing presence and simply add another one. This could be handled by allowing providing a presence source header in the request, or in the body. (Although I think using the header would be cleaner)

Discord does implement having multiple RPCs running on the client at the same time too.
image

ryotairi:
maybe communication with webclients can be done using iframes? like element does with widgets...

ryotairi:
image

proposals/4320-rich-presence.md
TheArcaneBrony: That is what Discord does

proposals/4320-rich-presence.md
```diff
@@ -0,0 +1,74 @@
+# MSC4320: Rich Presence
+
+# Context
+Discord, a popular instant messenger, has a feature where alongside normal presence (ie online/away/busy/offline, status messages) there can be current activities. For example, music shows album art, song name, artist and current song progress, games show the map and such, etc.
+# Proposal 
+### Data transport
+JSON data will be sent over a websocket running on `http://localhost:{TBD}/rpc`. This will be easiest for native clients, web clients will require something running on the host. Clients will send data to the server via a `PUT` to `/_matrix/client/v3/profile/{userId}/m.rpc` with content explained in "data content". Servers should preform basic validation on `m.rpc` to ensure only the needed fields are included. After an activity is closed or a song is paused/stopped, a`DELETE` should be sent to `_matrix/client/v3/profile/{userId}/m.rpc`. Note that skipping a track should not delete the RPC, as long as there's music playing. In that event, the `m.rpc` field should just be updated.
```
TheArcaneBrony, line 7: We should avoid allocating a port for this, as there is no guarantee that a given port isn't taken - Discord for example uses `/run/user/$UID/discord-ipc-N` (where N is a number from 0 to 9, taking the lowest available value)

proposals/4320-rich-presence.md
```diff
@@ -0,0 +1,74 @@
...
+	"avatar_url": "…", "displayname": "…",
+	"m.rpc": {
+		"type": "m.rpc.activity",
+		"name": "SuperTuxKart",
+		"details": "Playing Oliver's Math Class as Xenia on Intermediate"
+	}
+}
+```
+### Potential implementations
+- Something using [arRPC](https://arrpc.openasar.dev/) to pretend to be Discord to get info.
```
TheArcaneBrony, line 64: This relies on automating a Discord user account to obtain application info (ie. resolving an application ID to it's name and other info) - this isn't available to bots as far as I'm aware.

proposals/4320-rich-presence.md
```diff
@@ -0,0 +1,74 @@
...
+Discord, a popular instant messenger, has a feature where alongside normal presence (ie online/away/busy/offline, status messages) there can be current activities. For example, music shows album art, song name, artist and current song progress, games show the map and such, etc.
+# Proposal 
+### Data transport
+JSON data will be sent over a websocket running on `http://localhost:{TBD}/rpc`. This will be easiest for native clients, web clients will require something running on the host. Clients will send data to the server via a `PUT` to `/_matrix/client/v3/profile/{userId}/m.rpc` with content explained in "data content". Servers should preform basic validation on `m.rpc` to ensure only the needed fields are included. After an activity is closed or a song is paused/stopped, a`DELETE` should be sent to `_matrix/client/v3/profile/{userId}/m.rpc`. Note that skipping a track should not delete the RPC, as long as there's music playing. In that event, the `m.rpc` field should just be updated.
+### Data federation
+`m.rpc` data will be federated over a `GET` to `/_matrix/client/v3/profile/{userId}/m.rpc`, as per [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
```
TheArcaneBrony, line 9: Concern: what about homeservers/clients that enforce long TTLs on profile caching? (ie. 15 minutes doesn't seem unreasonable whatsoever, which would be eg. about 5x the length of an average song)

PR #4324, opened 2025-08-06
MSC4324: Fixing MSC4289's power level for tombstones

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/fix4289-pl150-tie/proposals/4324-fix-msc4289-tombstone-power-level.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published with my role as a member of the SCT.

*See MSC for spec process notes.*

---

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4324#issuecomment-3161163892)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4324#issuecomment-3161160426)

### DIFF
--- a/proposals/4324-fix-msc4289-tombstone-power-level.md
+++ b/proposals/4324-fix-msc4289-tombstone-power-level.md
@@ -1,0 +1,74 @@
+# MSC4324: Fixing MSC4289's power level for tombstones
+
+This MSC modifies [MSC4289 (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4289-privilege-creators.md).
+
+*Spec process note*: After an MSC is accepted, some changes may require a further
+MSC at the discretion of the Spec Core Team (SCT). This can be done up until a
+proposal becomes merged, at which point the regular MSC process applies. When an
+MSC modifies another accepted-but-not-merged MSC, that MSC gets its own dedicated
+file and changes the target MSC directly in the same changeset.
+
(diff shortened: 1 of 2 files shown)


proposals/4324-fix-msc4289-tombstone-power-level.md
turt2live, line 1: Implementation requirements:

* Technically, a server which does this, which is true as part of the room version 12 rollout: Synapse uses 150, which matches the "above all other PLs" requirement.

turt2live:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
In the interest of getting opinions on the approach here, and to encourage this MSC along for the spec release:

@mscbot fcp merge

turt2live:
Out of band discussion in a few places concludes that this MSC is not really required as a bug fix. The spec's (now approved) wording is as follows, which sufficiently covers the case this MSC is concerned about:

> If the room's version is 12 or higher, the power level for sending `m.room.tombstone` events MUST explicitly be higher than `state_default`. **For example, set to 150 instead of 100.**

(emphasis mine)

A future MSC may still wish to suggest default power levels, or possibly even require them, to avoid awkward situations like this in the future. 

@mscbot fcp cancel

PR #4327, opened 2025-08-13
Clarify capability behaviour with respect to avatar_url and displayname in MSC4133

Johennes:
Discovered while writing the spec PR. See https://github.com/matrix-org/matrix-spec-proposals/pull/4133#discussion_r2254938551 for further details.

Process note: MSCs can receive clarification changes post-FCP, but anything materially different requires a new MSC. I believe this qualifies as clarification because it isn't core to the MSC and the implementation used for FCP does the behaviour described. SCT members are able to disagree and require a new MSC instead.


richvdh (approved):
Thanks!

PR #4329, opened 2025-08-15
MSC4329: Inviting with authorization

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/invite-with-auth/proposals/4329-invite-with-auth.md)

Used to investigate an alternative to https://github.com/matrix-org/matrix-spec-proposals/pull/4311


turt2live:
For the reasons at https://github.com/matrix-org/matrix-spec-proposals/pull/4311#discussion_r2278002009 I don't believe this MSC is materially better (or worse) than 4311 as proposed, so I'm closing this immediately. There may be parts of this MSC that future MSCs are interested in, such as dropping `room_version` and `{eventId}` from the invite endpoint, but there's no driving factor in this MSC to keep those changes open.

PR #4331, opened 2025-08-26
MSC4331: Device Account Data

jevolk:
[rendered](https://github.com/matrix-construct/matrix-refreshed/blob/device-account-data/proposals/4331-device-account-data.md)

### DIFF
--- a/proposals/4331-device-account-data.md
+++ b/proposals/4331-device-account-data.md
@@ -1,0 +1,56 @@
+# MSC4331: Device Account Data
+
+Specification version 1.15 § 
+[10.21.2](https://spec.matrix.org/v1.15/client-server-api/#client-behaviour-17) 
+features key+value storage allocated on the server for a user's
+"Account Data." It also features similar storage for a user's "Room 
+Account Data" like Account Data but with a `room_id` division. These 
+form the two components of Account Data which have existed prior to the 
+1.0 specification through the present release.
+
+## Proposal
+
+We specify a third Account Data component: "Device Account Data" with 
+similar characteristics as the other two but with a `device_id` 
+division. Device Account Data acts as an arbitrary server-side 
+key+value store for each of a user's devices. Writes to this store are 
+only permitted by the device with a matching `device_id`. Reads from 
+this store can occur by any device and the server.
+
+It is in fact the server's read-access which is the motivator for this 
+proposal. Without it there is little reason to burden the server with 
+storing anything on behalf of a device with its own local storage.
+
+Future-specified event types and contents communicate per-device 
+information to the server. Examples of such information for further 
+specifications include:
+- Supported room versions (replaces 
+[MSC4292](https://github.com/matrix-org/matrix-spec-proposals/pull/4292)).
+- Client software version identifier.
+- Encryption metadatas.
+
+## Client-server endpoints
+
+- `GET /_matrix/client/v3/user/{userId}/devices/{deviceId}/account_data/{type}`
+
+- `PUT /_matrix/client/v3/user/{userId}/devices/{deviceId}/account_data/{type}`
+
+## Stabilization
+
+Such communication to the server via Account Data can certainly occur 
+via other components with well-specified types and contents but no 
+other use case for Device Account Data has been proffered. This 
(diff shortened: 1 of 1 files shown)


proposals/4331-device-account-data.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4331-device-account-data.md
```diff
@@ -0,0 +1,56 @@
...
+proposal is therefore to remain dormant (but not draft) until another
+proposal invokes it as a dependency. At that time this proposal is to 
+be stabilized concurrently with its first dependent.
+
+## Enhancements
+
+It is possible to add an `account_data` field to the v1.15 § 
+[10.11.1](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3devices) 
+response's `Device` object as a convenience, but that will not be
+mandated by this proposal without feedback.
```
tulir: Devices can already have metadata https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3devicesdeviceid, why not add fields to that mechanism rather than adding a new one? The example use cases seem like they'd fit in device metadata just fine

proposals/4331-device-account-data.md
tulir, line 1: This appears to be missing the potential issues, alternatives and security considerations sections

proposals/4331-device-account-data.md
jevolk: If the semantics of the metadata are satisfactory (and appear so) we could just pivot this MSC to minimally replace #4292 by specifying `supported_room_versions` as another metadata field, and maybe `client_version` for fun.

Metadata isn't a granular key-store though, but it doesn't seem hard to extend those endpoints with another path component e.g. `GET /_matrix/client/v3/devices/{deviceId}/display_name` which would basically supersede the full functionality of the Account Data approach. This might be mission-creep though.

jevolk: Another alternative is a little more radical: deprecate the device metadata suite itself in favor of Device Account Data which has resemblance to other Account Data. One thing missing from the existing proposal is the fact that `account_data` should come down through `/sync` which other devices can react to. AFAIK not possible with device metadata?

proposals/4331-device-account-data.md
```diff
@@ -0,0 +1,56 @@
...
+
+Future-specified event types and contents communicate per-device 
+information to the server. Examples of such information for further 
+specifications include:
+- Supported room versions (replaces 
+[MSC4292](https://github.com/matrix-org/matrix-spec-proposals/pull/4292)).
+- Client software version identifier.
+- Encryption metadatas.
+
+## Client-server endpoints
```
uhoreg: Does the device account data get deleted when the device is logged out?

proposals/4331-device-account-data.md
```diff
@@ -0,0 +1,56 @@
...
+form the two components of Account Data which have existed prior to the 
+1.0 specification through the present release.
+
+## Proposal
+
+We specify a third Account Data component: "Device Account Data" with 
+similar characteristics as the other two but with a `device_id` 
+division. Device Account Data acts as an arbitrary server-side 
+key+value store for each of a user's devices. Writes to this store are 
+only permitted by the device with a matching `device_id`. Reads from
```
uhoreg: If any of the user's devices is allowed to write to any other device's account data, it could be used for things like [MSC3890](https://github.com/matrix-org/matrix-spec-proposals/pull/3890) (and, in fact, that MSC calls out [per-device account data](https://github.com/matrix-org/matrix-spec-proposals/pull/3890/files#diff-b32a4af9b74efc2c7a6af62a1ce0b53ec4f12f9d92349b75f5682f4ebb16cce7R71-R76) as an alternative).

jevolk: The same-device-writer clause is intended to prevent ABA issues out-of-the-box. The cited case is certainly compelling though. I think a solution would be to allow type-specific semantics, but defaulting to the proposed same-device-writer if the type doesn't specify otherwise. In many cases a single-writer can be specified to mitigate ABA's just as well, even if that single-writer isn't the same device.

PR #4332, opened 2025-08-27
MSC4332: In-room bot commands

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/in-room-bot-commands/proposals/4332-in-room-bot-commands.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published in my capacity as Tech Lead for T&S, though does not achieve any particular T&S project/goal.

### DIFF
--- a/proposals/4332-in-room-bot-commands.md
+++ b/proposals/4332-in-room-bot-commands.md
@@ -112,6 +112,8 @@ When the user sends the command, the client creates either an `m.room.message` e
   "m.mentions": {
     "user_ids": ["@bot:example.org"] // should be a single element array, containing the bot's user ID
                                      // from the `m.bot.commands` state event's `state_key` (or `sender`).
+                                     // Note: doesn't include other users which may be referenced by the
+                                     // command being sent, such as via `user_id` arguments.
   },
 
   // This is a new content block so bots don't *need* to do string unpacking when


proposals/4332-in-room-bot-commands.md
turt2live, line 1: Implementation requirements:
* [ ] Client
* [ ] Bot

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,200 @@
...
+  the room version.
+* `roomId` - Must be a valid [room ID](https://spec.matrix.org/v1.15/appendices/#room-ids).
+* `roomAlias` - Must be a valid [room alias](https://spec.matrix.org/v1.15/appendices/#room-aliases).
+* `eventId` - Must be a valid [event ID](https://spec.matrix.org/v1.15/appendices/#event-ids).
+* `serverName` - Must be a valid [server name](https://spec.matrix.org/v1.15/appendices/#server-name).
+* `permalink` - Must be a valid [permalink URI](https://spec.matrix.org/v1.15/appendices/#uris)
+  (either `matrix.to` or `matrix:`).
+
+**Note**: For clarity, the above variables do not have to point at a room/user/server/etc that is known
+to the client. They just need to *look* valid per the grammar.
```
tulir: Why are there hardcoded variable names instead of a variable type field? What if you want 2 user IDs? (not to mention it'd be unfortunate to be forced to use that ugly capitalization instead of `user_id`, `userID` or `UserID`)

turt2live: With this proposal's high bikeshed potential, I didn't want to add complexity that isn't actually needed right away. Variable types does sound like a thing we probably want, but without an in-practice use case it's hard to justify that complexity.

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,200 @@
...
+variables:
+
+* `userId` - Must be a valid [user ID](https://spec.matrix.org/v1.15/appendices/#user-identifiers) for
+  the room version.
+* `roomId` - Must be a valid [room ID](https://spec.matrix.org/v1.15/appendices/#room-ids).
+* `roomAlias` - Must be a valid [room alias](https://spec.matrix.org/v1.15/appendices/#room-aliases).
+* `eventId` - Must be a valid [event ID](https://spec.matrix.org/v1.15/appendices/#event-ids).
+* `serverName` - Must be a valid [server name](https://spec.matrix.org/v1.15/appendices/#server-name).
+* `permalink` - Must be a valid [permalink URI](https://spec.matrix.org/v1.15/appendices/#uris)
+  (either `matrix.to` or `matrix:`).
```
Gnuxie: Could we instead declare types for variables (or probably more accurately schema) rather than special cased properties?

Gnuxie: Some of these seem to be conflicting too, why can't a permalink be provided as a *user ID*, a *room ID*, a *room alias*, or an *event ID*?

Feel free to draw inspiration from how Draupnir creates presentation types: https://github.com/the-draupnir-project/interface-manager/blob/main/src/TextReader/TextPresentationTypes.ts (it may also help to peek into https://github.com/the-draupnir-project/interface-manager/blob/main/src/TextReader/TextCommandReader.ts)

Draupnir's command system reads both permalinks with reference to a *room ID* and a bare *room ID* as both `MatrixRoomID`'s. And aliases to `MatrixRoomAlias`. And the union of both is a `MatrixRoomReference`. https://github.com/the-draupnir-project/matrix-basic-types/blob/main/src/MatrixReference/MatrixRoomReference.ts

(All code linked is Apache-2.0)

Gnuxie: To see an example of this feel free to look at Draupnir's ban command (AFL-3.0 https://github.com/the-draupnir-project/Draupnir/blob/main/src/commands/Ban.tsx#L64-L92)

proposals/4332-in-room-bot-commands.md
turt2live: From https://github.com/matrix-org/matrix-spec-proposals/pull/4332#discussion_r2304858451 :

> To see an example of this feel free to look at Draupnir's ban command (AFL-3.0 https://github.com/the-draupnir-project/Draupnir/blob/main/src/commands/Ban.tsx#L64-L92)

proposals/4332-in-room-bot-commands.md
turt2live: moving to https://github.com/matrix-org/matrix-spec-proposals/pull/4332#discussion_r2304897396

turt2live: The theory is the bot would publish self-conflicting commands which the user can choose from, so there'd be `draupnir ban {userId}` and `draupnir ban {permalink}`.

Though, having actually written that down: the bot is surely going to get confused when the user themselves become confused by "ban [event ID link]".

proposals/4332-in-room-bot-commands.md
turt2live: https://github.com/the-draupnir-project/Draupnir/pull/944 is an incomplete bot implementation. It would still need to react properly to the mixin/mention, and ideally not be hardcoded.

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,200 @@
...
+        // See https://spec.matrix.org/v1.15/client-server-api/#mroomtopic_topiccontentblock
+        // See https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md
+        // See https://github.com/matrix-org/matrix-spec-proposals/pull/3554
+        "roomId": {
+          "m.text": [{"body": "The room ID"}]
+        },
+        "userId": {
+          "m.text": [{"body": "The user ID"}]
+        }
+      },
```
Gnuxie: In order for Draupnir to automatically generate the supported commands from [interface-manager's](https://github.com/the-draupnir-project/interface-manager) command descriptions, variadic arguments would need to be represented.

*options* would also be nice to consider but wouldn't be necessary,

proposals/4332-in-room-bot-commands.md
turt2live: Incorporated into the proposal as a client tip

turt2live:
For the overall changelog: this had a bunch of discussion amongst T&S developers from the community, resulting in the above changes. Namely: variables are now called arguments. Arguments have types now. Variadic arguments are possible.

proposals/4332-in-room-bot-commands.md
tulir: https://github.com/gomuks/gomuks/pull/655 is a very ugly client implementation, but I think in theory it might actually work

tulir: The gomuks implementation technically also includes a "bot" now, because the backend will intercept commands sent by the frontend targeted at `@gomuks` https://github.com/gomuks/gomuks/blob/429e014eb28e99291a27a676e0a4ca64ac79b710/pkg/hicli/send.go#L87-L89 (there was no command system in the past, so this new system is used for both built-in and external bot commands)

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,243 @@
...
+
+```jsonc
+{
+  "sigil": "!", // Defaults to `!` if not specified. Clients can use this to show the user a consistent
+                // experience in the form of slash commands, but ultimately send the command as a
+                // sigil-prefixed string to the room. Eg: the UI might say `/botname`, but the command
+                // becomes `!botname` upon sending.
+  "commands": [
+    {
+      "syntax": "botname {action} {roomId} {timeoutSeconds} {applyToPolicy} {userId...}", // `{words}` are positional arguments.
```
Gnuxie, line 38: why are we using camel case for the argument names?

Even with, specifically `roomId` and `userId` are a bit opinionated.

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,243 @@
...
+
+![](./images/4332-discord-example.png)
+
+When the user sends the command, the client creates either an `m.room.message` event with the following
+`content` shape:
+
+```jsonc
+{
+  // These fields would be replaced by MSC1767 Extensible Events in future.
+  "body": "!botname ban_and_suspend !room:example.org 42 true @alice:example.org @bob:example.org", // note that the syntax template is populated
```
Gnuxie, line 107: I think we are over estimating the value in `body` as a fallback for bots that haven't adapted their command parsers to MSC4332 yet. And we risk restricting the benefit we will get from client support for commands by allowing too much freedom here for the purpose of backwards compatibility with existing bots. Additionally the MSC already imposes decisions on how bots have to parse commands in the `body` representation that is incompatible with existing bots that already have complex command parsing.

Gnuxie: See https://github.com/matrix-org/matrix-spec-proposals/pull/4338

Gnuxie: See also https://github.com/matrix-org/matrix-spec-proposals/pull/4340

proposals/4332-in-room-bot-commands.md
turt2live: this is just the example - I prefer that format, but ultimately it's just a string.

proposals/4332-in-room-bot-commands.md
turt2live: I suspect that bots with suitably complex commands will need to update their parsers at the same time they advertise support for the MSC. Simple bots (gif searching, TWIM, etc) may be able to avoid it.

proposals/4332-in-room-bot-commands.md
TheArcaneBrony, line 1: I couldn't find a relevant part to comment on, but does this MSC consider accessibility (ie. screen readers) for clients that implement in-composer templating UI? Discord has this problem where slash commands aren't available at all if you enable the accessible text input under Accessibility settings.

timedoutuk: My bot [sensebot-go](https://git.nexy7574.co.uk/nex/sensebot-go) ([`@sensebot:synapse.nexy7574.co.uk`](matrix:u/sensebot:synapse.nexy7574.co.uk?action=chat)) has been [updated](https://git.nexy7574.co.uk/nex/sensebot-go/compare/97acdee5a3033a7b4a2ce6cee947ec8c7a68ac46..568d08c778749cba437df04617a7c67b54fab396) to use MSC4332 commands, with a UX that looks like below:

https://github.com/user-attachments/assets/02020a07-753a-4f0e-ae11-86d4053bf525

Built using mautrix-go, like Gomuks does above. It supports registering commands (only manually for now) and receiving them, differentiating between non-fallback and fallback parsing. Pretty sure this meets the requirements for the bot checkbox, and Gomuks Web should be able to check the client checkbox.

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,243 @@
+# MSC4332: In-room bot commands
+
+Many bots on Matrix have a command interface consisting of `!botname <command>`, and have a pretty
+long help menus which can make it difficult to find the right command. Many clients already have a
+concept of "slash commands" which are [desirable to reuse](https://github.com/matrix-org/matrix-spec/issues/93)
+and [come up occasionally](https://github.com/matrix-org/matrix-spec/issues/2170) - finding a way to
+populate this feature with bot-specific details is beneficial.
+
+This proposal suggests that bots maintain a state event in the rooms it joins to advertise available
```
timedoutuk, line 9: I don't have a solution to this, but it's worth noting down that this requires room admin intervention since your typical user won't be able to grant bots the power level required to send this state event, and setting the event's power level to world accessible opens the door to abuse. The MSC does elaborate on this but doesn't seem to highlight the severity of this issue, or at least I don't feel it does.

Perhaps stuffing the command registration into memberships retains the exclusivity, but doesn't require a power level modification?

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,243 @@
...
+When presenting command options to users, clients SHOULD use this event to suggest per-bot commands
+too, unless the user ID implied by the `state_key` is not joined to the room. (Note that this means
+"invalid" state keys get treated as unjoined users: an empty string, "not_a_user_id", etc can't
+join rooms, but `@bot:example.org` can.)
+
+The `content` for such an event fits the following implied schema:
+
+```jsonc
+{
+  "sigil": "!", // Defaults to `!` if not specified. Clients can use this to show the user a consistent
```
timedoutuk, line 32: nitpicking, `prefix` might be better that `sigil` - it's not uncommon for a multi-character prefix, and `!botname` itself might even be considered the prefix with the word after it being the command to invoke (e.g. `!botname` `help` instead of `!` `botname` `help`)

proposals/4332-in-room-bot-commands.md
```diff
@@ -0,0 +1,243 @@
...
+          "type": "user_id",
+          "description": { "m.text": [{"body": "The user ID(s)"}] },
+          "variadic": true // can only apply to the last argument. Default false when not supplied.
+        }
+      ],
+      "description": {
+        // We also use m.text here for the same reason as the argument descriptions above.
+        "m.text": [{"body": "An example command with arguments"}]
+      }
+    }
```
timedoutuk, line 82: The ability to register aliases or shortcuts without having to define them as their own separate commands (which itself will then inherently clog up the command list displayed to users) would be nice

slavb18:
Hi! 

Is it possible to bridge buttons to telegram bot?

proposals/4332-in-room-bot-commands.md
tulir, line 1: @slavb18 said outside of a thread:

> Hi!
>
> Is it possible to bridge buttons to telegram bot?

tulir: This MSC (and the newer replacement https://github.com/matrix-org/matrix-spec-proposals/pull/4391) are only for commands, not buttons. A future MSC could build buttons in a similar way, but it's out of scope here.

(also, all comments on MSCs must be in threads so they can be appropriately replied to and resolved)

PR #4333, opened 2025-08-27
MSC: Room state API for moderation bots

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/modbot-api/proposals/4333-moderation-bot-state-api.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published in my capacity as Tech Lead for T&S, though does not achieve any particular T&S project/goal.

### DIFF
--- a/proposals/4333-moderation-bot-state-api.md
+++ b/proposals/4333-moderation-bot-state-api.md
@@ -1,0 +1,146 @@
+# MSC4333: Room state API for moderation bots
+
+Moderation bots such as [Draupnir](https://github.com/the-draupnir-project/Draupnir), [Meowlnir](https://github.com/maunium/meowlnir),
+and [Mjolnir](https://github.com/matrix-org/mjolnir) typically have the same basic functions relevant
+to this MSC's interests:
+
+* When a ban recommendation is published to a watched [policy room](https://spec.matrix.org/v1.15/client-server-api/#moderation-policy-lists),
+  the bot bans that user/server.
+* Publishing to policy rooms, hiding the original moderator which added the recommendation (a feature).
+* Redacting events sent by users either as part of a ban or as a once-off action.
+* Kicking users from rooms.
+
+Other functionality is common among these bots, such as an ability to add "protections" which monitor
+for certain classifications of spam and taking appropriate action, but are generally out of scope for
+the first cut of this proposal. A future proposal may build upon this MSC to add more functionality
+as needed.
+
+To exercise the above functionality, a moderator typically needs to encounter problematic content,
+record/copy some details of the user or permalink, find their bot's "management room", formulate the
+command (complete with typos), and send it. Instead, a desirable workflow would be to have the moderator's
+ban, redact, and kick buttons already present in their client's UI actually route to the bot's management
+room, reducing the need for context switching and typos in commands.
+
+This proposal achieves that button rewiring by leveraging [MSC4332](https://github.com/matrix-org/matrix-spec-proposals/pull/4332)
+bot commands and a new state event published in the moderation bot's management room.
+
+
+## Proposal
+
+A new state event is added: `m.bot.moderation_config`. Like and for the same reasons in MSC4332, the
+`state_key` is the bot's user ID. If the user ID referenced by the state key is not present in the
+room where the moderation config event is sent, that event does not apply.
+
+The `content` for the new event fits the following implied schema:
+
+```jsonc
+{
+  // The room IDs the moderation bot "protects". Typically, these are the rooms where policy recommendations
+  // are automatically applied.
+  "protected_room_ids": [
+    "!room:example.org"
+  ],
+  "commands": {
+    // These are the MSC4332 commands the client UI can use instead of the normal Client-Server API,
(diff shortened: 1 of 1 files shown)


proposals/4333-moderation-bot-state-api.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Moderation bot

proposals/4333-moderation-bot-state-api.md
```diff
@@ -0,0 +1,146 @@
...
+record/copy some details of the user or permalink, find their bot's "management room", formulate the
+command (complete with typos), and send it. Instead, a desirable workflow would be to have the moderator's
+ban, redact, and kick buttons already present in their client's UI actually route to the bot's management
+room, reducing the need for context switching and typos in commands.
+
+This proposal achieves that button rewiring by leveraging [MSC4332](https://github.com/matrix-org/matrix-spec-proposals/pull/4332)
+bot commands and a new state event published in the moderation bot's management room.
+
+
+## Proposal
```
turt2live, line 28: TODO: Describe *why* we go with the commands approach instead of something else. The short version is we could support HTTP APIs, toDevice, custom event types (this one is mentioned briefly later), toWidget, etc, but commands are universally easy to support and hook up per MSC4332.

proposals/4333-moderation-bot-state-api.md
```diff
@@ -0,0 +1,146 @@
...
+user chooses to use the management room, the client sends the command specified (if it can) to the
+management room, using the existing context it already has.
+
+For example, if the user is trying to ban another user from a protected room, the client might show
+a checkbox saying "Use 'Moderation Internal' to ban this user". If checked when the user submits the
+dialog, the `reason` and user ID the user is trying to ban will be converted to a command which is
+then sent to the management room (instead of calling the `/ban` API, for example).
+
+
+## Potential issues
```
Gnuxie, line 95: Draupnir in the future is going to require a `--no-confirm` flag for the ban command. I think if we spot that the `org.matrix.msc4332.command`mixin is used in an event that we will have to pretend the `--no-confirm` flag is present.

proposals/4333-moderation-bot-state-api.md
turt2live: WIP as of writing, but moderation bot exists here: https://github.com/the-draupnir-project/Draupnir/pull/944

Also WIP as of writing, client is here: https://github.com/element-hq/element-web/pull/30648

proposals/4333-moderation-bot-state-api.md
```diff
@@ -0,0 +1,146 @@
...
+* The Community Moderation Effort (CME) folks have access to moderation bots and policy lists which
+  the human operators know ultimately protect a given room, but the bots/list can't reasonably say
+  they do in fact protect the room ID. Though not formally part of this proposal (**TODO**: Yet?),
+  it's suggested that clients maintain an `m.bot.moderation_config` account data event with `content`
+  of `{"management_room": "!manage:example.org"}`. The choice of management room would then instruct
+  the client to use that room's `m.bot.moderation_config` state event, ignoring the `protected_rooms`
+  array (instead, treating *all* rooms the client sees as "protected" by that management room).
+
+
+## Alternatives
```
Gnuxie, line 114: So seemingly this MSC tries to create an "adaptor" for bot commands with similar semantics, so that a client can present these as common commands to their users. When in reality the bots have different options and subtly different semantics for each common command.

I think an alternative that should be explored is just defining the common commands and semantics. With careful consideration for what they mean from the client perspective. And letting bots provide dedicated support to these commands and the spec's specific semantics for the commands. Rather than the client trying to adapt to different bot commands with different semantics.

Gnuxie, line 114: Continuing on from https://github.com/matrix-org/matrix-spec-proposals/pull/4333/files#r2454823574 in another line of thought.

Another alternative is to try break down the MSC into other primitives that would allow bot commands to present themselves in the client UI generically. Such as in the user pane from within a room.

A separate line of exploration that could make this a lot easier to do, is if Draupnir interface-manager style prompt support is added to the MSC series. As any command that accepts a user as an argument can then be shown in the user pane. And any other arguments can be prompted for.

Gnuxie: Note: When i say prompt support, i really mean partial commands. Sending a partial command to the bot, so that it can reply dynamically with a selection for the user to make.

Gnuxie: This becomes much easier to design if https://github.com/matrix-org/matrix-spec-proposals/pull/4332 is reworked away from a perspective of providing an adaptation layer for "fallback" text based commands[^1]. And instead focus on the command matrix event mixin. As partial commands can then be sent to prompt for individual arguments in any order, regardless of whether they are positional arguments. And this would open the door for the kind of generalisation that would allow commands to be placed in the user pane or other parts of the client UI. 

[^1]: Which has already been shown not to work, as commands not specifically adapted to the MSC have lots of differences in syntax and parameter parsing.

Issue #1504, opened 2018-08-13
Homeserver resource limiting error codes


neilisfragile:
Documentation: https://docs.google.com/document/d/1_CgwkfLznU56fwLUFZXFnKivzVUPjjNsTO0DpCB9ncM/edit?usp=sharing

Encompasses #1470 (since M_SOFT_LIMIT_EXCEEDED covers M_MAU_LIMIT_EXCEEDED)

turt2live:
How does `M_SOFT_LIMIT_EXCEEDED` cover `M_MAU_LIMIT_EXCEEDED`? Was the MAU error replaced?

It'd be nice to have the use case for having two different errors documented. A client doesn't appear to have instructions on what to do when a subset of users are disallowed versus all of them - it seems like it should only care if the requesting user is limited for some reason.

neilisfragile:
How does M_SOFT_LIMIT_EXCEEDED cover M_MAU_LIMIT_EXCEEDED?  Was the MAU error replaced?Yes

It seems like it should only care if the requesting user is limited for some reason. I'd see it as more for the users benefit - they know *they* can't access the service but that the rest of the service is not down.

If it is not much use from a client perspective - then I'm not overly attached to it

turt2live:
I think a single error code would be okay, given the user would (hopefully) be made aware of what's going on through other channels, such as the server notices room. Would love to hear from some client authors though, like @KitsuneRal and @mujx, as well as the wider client dev community.

richvdh:
I think this has been implemented without ever having completed review, or indeed ever being written up into a coherent document :angry: 

I guess it needs speccing now

turt2live:
Merged via https://github.com/matrix-org/matrix-doc/pull/1874

PR #4335, opened 2025-08-29
MSC4335: M_USER_LIMIT_EXCEEDED error code

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/user-limit-exceeded/proposals/4335-user-limit-exceeded.md)

---
I am employed by Element to write this MSC on behalf of the Matrix Foundation for use on the [matrix.org homeserver](https://matrix.org/homeserver/about/). Hence the use of the `org.matrix` namespace on the unstable values.

## Implementations

- Server:
  - [x] Synapse: https://github.com/element-hq/synapse/pull/18876
- Client:
  - [x] Element Web: https://github.com/element-hq/element-web/pull/30738

In the prototype Element Web implementation when this error code is encountered for a file upload, it is rendered as follows for a soft limit:

image

And like this for a hard limit:

image

## Customisation of clients for matrix.org homeserver

The Matrix.org Foundation has discussed with [Element Creations Limited](https://element.io) the possibility of having Element clients show a custom message when the `M_USER_LIMIT_EXCEEDED` error is received from the matrix.org homeserver. Element have agreed to this in principle.

If this MSC is accepted then the Foundation intends to request that other client developers consider implementing a similar customisation for the matrix.org homeserver. The purpose of this is to support the [funding of the matrix.org homeserver](https://matrix.org/blog/2025/06/funding-homeserver-premium/).

Whilst the exact design/form of this in Element is to be decided, the following illustrates the level of customisation that is anticipated. (For example, it might appear as feedback within the timeline instead of a modal)

In the case of a Free user encountering a limit that can be increased by upgrading to Premium:

image

For a hard limit:

image

---

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issuecomment-3391828661)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issuecomment-3391550751)

### DIFF
--- a/proposals/4335-user-limit-exceeded.md
+++ b/proposals/4335-user-limit-exceeded.md
@@ -21,7 +21,7 @@ This proposal adds a new [common error code]
 `M_USER_LIMIT_EXCEEDED` to the Matrix specification. This error code should be returned when a user has exceeded
 limits that are specifically associated with their account, such as:
 
-* **Storage quotas**: When a user has exceeded their allocated storage space for smedia uploads,
+* **Storage quotas**: When a user has exceeded their allocated storage space for media uploads,
   message history, or other persistent data.
 * **Resource limits**: When a user has reached their maximum number of allowed rooms, devices,
   or other account-scoped resources.
@@ -270,7 +270,7 @@ take start the action of upgrading.
 However, this led to [uncertainty](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#discussion_r2582153412)
 about what exactly would happen when the button is pressed.
 
-To remove this uncertainty `increase_uri` was removed and the homserver gets to provide a single `info_uri` and it is
+To remove this uncertainty `increase_uri` was removed and the homeserver gets to provide a single `info_uri` and it is
 the responsibility of the homeserver to provide a UX that makes sense and is clear about what is happening.
 
 ## Dependencies


proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,126 @@
...
+* `M_FORBIDDEN`: Used for authorization failures or policy violations not related to usage limits
+* `M_RESOURCE_LIMIT_EXCEEDED`: Used for server-wide resource constraints affecting all users
+* `M_TOO_LARGE`: Used when a request is too large (file size, message length, etc.) regardless of user limits
+
+## Potential issues
+
+This error code does not specify the exact nature of the limit that was exceeded, which could
+potentially lead to ambiguity. However, this is consistent with other Matrix error codes that
+rely on the human-readable `error` field to provide specific details. Implementers should ensure
+they provide clear, actionable error messages.
```
t3chguy: Which is problematic as it lacks i18n and not everyone understands English especially more technical things like quotas. So clients will end up ignoring the error field, as they tend to, and show an opaque error which will feel unhelpful. We have specific customer asks to not leak server errors thru to ui due to lack of i18n.

hughns: I am hesitant to suggest attempting to enumerate possible reasons as I suspect they will end up being quite specific to the homeserver.

Might providing a link field similar to `admin_contact` on `M_RESOURCE_LIMIT_EXCEEDED` be a suitable alternative approach?

So, the client could render `M_USER_LIMIT_EXCEEDED` with i18n as something like:

> The request failed as you have exceeded a usage limit imposed by the homeserver. More information on this is available \<a href="link from error">here\</a>

or as buttons along with other actions:

> The request failed as you have exceeded a usage limit imposed by the homeserver
> \[Try again] \[Cancel] \[More information]

t3chguy: No qualms with including a link for more information, whether its shown inline vs buttons would be up to the individual apps' designers

hughns: I've updated the proposal in [`fbd628a` (#4335)](https://github.com/matrix-org/matrix-spec-proposals/pull/4335/commits/fbd628a5cd39e7e23973a226c8348a7098c42f23) to include an <del>`info_link`</del>`info_url` field.

t3chguy: Your commit says `info_link` but the markdown says `info_url` :D

hughns: Thank you 👍

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,125 @@
...
+While this proposal is being developed and refined, implementations should use the prefix the
+`errcode` and `info_url` fields with `org.matrix.msc4335.` and use a value of `M_UNKNOWN` for
+`errcode`.
+
+For example:
+
+```json
+{
+  "errcode": "M_UNKNOWN",
+  "org.matrix.msc4335.errcode": "M_USER_LIMIT_EXCEEDED",
```
Johennes: No need to change anything since it appears to already be implemented like this but I was slightly surprised that the prefix is on the key rather than the value. Would this not hypothetically allow for odd situations where you have `errcode`s from several MSCs, possibly leaving clients confused about which one to choose? I couldn't find any guidance on where to put the prefix in https://spec.matrix.org/proposals/ FWIW.

hughns: It would be great to get some clarification on what is the "right" way to do this.

Re-reading https://spec.matrix.org/v1.16/client-server-api/#standard-error-response I see it does state how custom error codes should be written.

I think this implicitly means that clients *should* expect non-standard error codes and do something sensible with them (which was what I was assuming would not be the case and hence prefixing the key).

hughns: I've updated the unstable values in 79518f08b8b75ae0615f30eb5aa14de87d25d764

Johennes: I have opened https://github.com/matrix-org/matrix-spec/pull/2222 to try and clarify the spec accordingly.

ara4n:
@mscbot fcp merge

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,150 @@
...
+* **Feature limits**: When a user has exceeded usage limits for specific features (e.g., number
+  of public rooms they can create, number of invites they can send).
+* **Account tier restrictions**: When a user's account type (free, premium, etc.) prevents them
+  from performing certain operations.
+
+The error response must also contain additional fields:
+
+* `info_uri` string (required) - a URI that the client can link the user to in order to get more context on the error
+* `soft_limit` boolean (required) - `true` means that the specific limit encountered can be increased. `false` means
+  that it is a hard limit that cannot be increased.
```
clokep: Should `soft_limit` be optional and default to false? I don't feel too strongly about it.

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,150 @@
...
+* `M_RESOURCE_LIMIT_EXCEEDED`: Used for server-wide resource constraints affecting all users
+* `M_TOO_LARGE`: Used when a request is too large (file size, message length, etc.) regardless of user limits
+
+## Potential issues
+
+This error code does not specify the exact nature of the limit that was exceeded, which could
+potentially lead to ambiguity. However, this is consistent with other Matrix error codes that
+rely on the human-readable `error` field to provide specific details. Instead the `info_uri`
+provides a way for the homeserver to apply arbitrary limits without the client having to understand
+every type in advance.
```
clokep, line 95: Should we include a `type` field that gets passed to `info_uri` as a query parameter or something?

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,150 @@
...
+}
+```
+
+### Applicable Endpoints
+
+This error code can be returned by any Matrix Client-Server API endpoint where user-specific limits
+might be enforced. Examples could include:
+
+* [`POST /_matrix/media/v3/upload`] - When storage quota is exceeded
+* [`POST /_matrix/client/v3/rooms/{roomId}/invite`] - When invite limits (like maximum participant count) are exceeded
```
clokep: Is there a full list of endpoints this might apply to?

tulir:

Checklist:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4335-user-limit-exceeded.md
KitsuneRal: Why not just encode the necessary information in `info_uri` straight away on the server side?

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,150 @@
...
+MSC if deemed necessary.
+
+## Alternatives
+
+Several alternatives were considered for this proposal:
+
+**Use M_RESOURCE_LIMIT_EXCEEDED**: The existing [`M_RESOURCE_LIMIT_EXCEEDED`] error code could be
+expanded to cover user-specific limits. However, this code is currently used for server-wide
+resource constraints, and overloading it could create confusion about whether the limit applies
+to the user specifically or the server generally.
```
Johennes: This is true but I wonder if it the distinction actually matters to users. Why does a client need to know whether they ran into a server-wide quote or a user-specific quota? It seems sufficient to know _that_ a quota was exceeded and whether or not the limit can be increased. As such, `M_RESOURCE_LIMIT_EXCEEDED` seems equivalent to `M_USER_LIMIT_EXCEEDED` with `soft_limit = false` and we might as well add the new fields on the existing error code?

proposals/4335-user-limit-exceeded.md
hughns: > Why not just encode the necessary information in `info_uri` straight away on the server side?

Exactly. The intention is that the URIs are opaque and the homeserver may choose to put a type or similar field in there.

I've attempted to clarify this in 4dd453bf459971dfeda0138fe7250566a17081f9.

proposals/4335-user-limit-exceeded.md
hughns: No, because this is a common error.

I've attempted to clarify this in 48713cbaeee2b051b951d91e06bfdd41a70e6d6d including add an alternative about using an "other" error code instead.

proposals/4335-user-limit-exceeded.md
hughns: It would be more concise. I have made this change in a64558b504870a5a3970308fd18c3c4714813c2a

proposals/4335-user-limit-exceeded.md
hughns: One thing to note, on re-reading `M_RESOURCE_LIMIT_EXCEEDED` I see that it is listed as an "other" error code but it sounds like it could actually apply to any API endpoint:

> Typically, this error will appear on routes which attempt to modify state (e.g.: sending messages, account data, etc) and not routes which only read state

Furthermore it is not listed against any specific endpoints.

My assessment is that `M_RESOURCE_LIMIT_EXCEEDED` is actually a common error code but is misclassified.

proposals/4335-user-limit-exceeded.md
hughns: I agree that from a user perspective the distinction doesn't matter. They need to know that a limit has been reached and what actions they can take.

However, the spec has always been quite clear about `M_RESOURCE_LIMIT_EXCEEDED` being a limit applied to the *server* not to the *user*.

Hence the default for an unrecognised `limit_type` in an [example client](https://github.com/element-hq/element-web/blob/c96da5dbf8e20ced4a512a03a75c91f8680e8d40/src/i18n/strings/en_EN.json#L1112) is as follows:

> This homeserver has exceeded one of its resource limits

Which I think would be confusing to users. This could of course be updated in the clients.

As I noted [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#discussion_r2425623107) I think the `M_RESOURCE_LIMIT_EXCEEDED` is in practise a common error code rather than "other".

It looks like Synapse is [using](https://github.com/element-hq/synapse/blob/d399d7649aedaeb3cd9edbc258917c407893c172/synapse/api/errors.py#L512) the `limit_type` field which isn't given in the error definition.

The format of `admin_contact` is also not well specified. Based on some of the breadcrumbs I think it is reasonable to assume that it has the same definition as [here](https://spec.matrix.org/v1.16/client-server-api/#mroommessagemserver_notice).

For those doing archeology:

- MSC in Google Doc form: https://docs.google.com/document/d/1_CgwkfLznU56fwLUFZXFnKivzVUPjjNsTO0DpCB9ncM/edit?tab=t.0#heading=h.eaxawgsbu6t2
- Tracking issue for MSC: https://github.com/matrix-org/matrix-spec-proposals/issues/1504
- Spec change: https://github.com/matrix-org/matrix-spec-proposals/pull/1874

hughns: So, instead we could do something like:

- `M_USER_LIMIT_EXCEEDED` -> `M_RESOURCE_LIMIT_EXCEEDED`
- `limit_type` of something like `user_quota`
- `info_uri` -> `admin_contact`
- `soft_limit` -> new field on error
- `increase_uri` new field on error

Given the flaws in the spec for `M_RESOURCE_LIMIT_EXCEEDED` I would want to sort these out at the same time.

As to whether this better than doing a "clean" new common code as originally described? I'm not sure.

Johennes: > Hence the default for an unrecognised `limit_type` in an [example client](https://github.com/element-hq/element-web/blob/c96da5dbf8e20ced4a512a03a75c91f8680e8d40/src/i18n/strings/en_EN.json#L1112) is as follows:
> 
> > This homeserver has exceeded one of its resource limits
> 
> Which I think would be confusing to users.

Good point (and fabulous research!). This might be good to have in the proposal text.

> As to whether this better than doing a "clean" new common code as originally described? I'm not sure.

It might not after all. I hadn't considered that extending `M_RESOURCE_LIMIT_EXCEEDED` doesn't fall back nicely due to existing client-side copy. I still think the differentiation might be unnecessary but given that we already have `M_RESOURCE_LIMIT_EXCEEDED` with specific semantics, it sort of fits the bill.

turt2live:
@mscbot concern Checklist not complete/started

(any SCT member can resolve this)

proposals/4335-user-limit-exceeded.md
clokep: > Why not just encode the necessary information in `info_uri` straight away on the server side?

Well I assume you want to tell people _what_ limit they've passed, would each of those be a separate `info_uri`? Feels like it ties your homeserver implementation tightly to whatever is at those URIs.

proposals/4335-user-limit-exceeded.md
hughns: Right. What I've done:

- incorporated relevant parts of above into 29d1e50cd3903111e772a13e01cf43d279040fa4
- made https://github.com/matrix-org/matrix-spec/pull/2232 to propose correcting the location of `M_RESOURCE_LIMIT_EXCEEDED` in the spec doc

I haven't done anything about adding `limit_type` to the `M_RESOURCE_LIMIT_EXCEEDED` as I think this would need to be a new MSC as it was never mentioned in https://docs.google.com/document/d/1_CgwkfLznU56fwLUFZXFnKivzVUPjjNsTO0DpCB9ncM/edit?tab=t.0#heading=h.eaxawgsbu6t2

hughns: I've now made https://github.com/matrix-org/matrix-spec-proposals/pull/4368 to add `limit_type` to `M_RESOURCE_LIMIT_EXCEEDED`.

Johennes: All excellent. Thank you. 👍

proposals/4335-user-limit-exceeded.md
clokep: I'm not understanding how just saying they're "opaque" resolves this. This makes it very tied to particular homeserver implementations.

hughns: That’s the point. It is tied to the homeserver.

Why is that a concern?

clokep: How is a client supposed to show any useful information to the end user? "Go to this webpage to read about one of 10 resources you might have overflowed" is not useful.

Maybe I'm misunderstanding how this is meant to be used though. Is it implemented anywhere? Are there screenshots?

hughns: Screenshots are at the top of this PR. https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issue-3367446512

hughns: The client knows what operation failed so can give some context. Like, in the screenshots, it can say that "upload failed" which is useful to provide the context.

clokep: Looking at https://github.com/element-hq/synapse/pull/18876, I don't see how the user will know what limit they went over though.

hughns: In the context of Synapse and the media upload limits feature, I imagined it could be configured with context like this:

```yaml
media_upload_limits:
- time_period: 24h
  max_size: 100M
  msc4335_info_uri: https://example.com/quota?type=daily
  msc4335_soft_limit: true
  msc4335_increase_uri: https://example.com/increase-quota?type=daily
- time_period: 28d
  max_size: 500M
  msc4335_info_uri: https://example.com/quota?type=monthly
  msc4335_soft_limit: true
  msc4335_increase_uri: https://example.com/increase-quota?type=monthly
```

Which would result in error response such as:

```json
{
  "errcode": "M_USER_LIMIT_EXCEEDED",
  "error": "User has exceeded their upload limit",
  "info_uri": "https://example.com/quota?type=daily",
  "soft_limit": true,
  "increase_uri": "https://example.com/increase-quota?type=daily"
}
```

The web page at the other end can then serve up content that is specific to the limit exceeded.

Generalising this, the homeserver implementation and/or operator could also do things like:

- Make use of path: https://example.com/quota-monthly
- Fragment: https://example.com/quota#monthly
- Multiple params: https://example.com/plan?action=limit_exceeded&type=upload_limit&period=monthly

In order to implement the soft limit and option to upgrade, a Synapse operator would need to make use the [get_media_upload_limits_for_user](https://element-hq.github.io/synapse/latest/modules/media_repository_callbacks.html#get_media_upload_limits_for_user) from the module API to allow certain users to have different media upload limits.

@clokep is it that you prefer the approach described at https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/user-limit-exceeded/proposals/4335-user-limit-exceeded.md#add-structured-error-information? If so then could you add a "vote" or preference to that alternative instead?

If not, then I feel like I am missing something.

clokep: OK, that sounds fine. I think that's not clear from the text that a _different_ URI per resource is expected. And the matching implementation doesn't do this.

hughns: I've attempted to clarify this in 721928be1fe1bc727212a6e31b2dd359de8b3251.

I've also added an alternative of using server side translations in e695894e43c8529fb729451f6d55bb4f192ec32b.

hughns:
I have added an alternative of using server side translations [here](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/user-limit-exceeded/proposals/4335-user-limit-exceeded.md#use-server-side-translations) instead of `info_uri`, `soft_limit` and `increase_uri`.

I think this would actually provide a better UX overall, albeit at the expense of complexity for the homeserver operator and error payload size.

I would welcome input on this.

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+    "en": "You have reached your daily upload limit. More information on the usage limits that apply to your account is available <a href=\"https://matrix.org/homeserver/pricing/#usage-limits\">here</a>.",
+    "fr": "Vous avez atteint votre limite de téléchargement quotidienne. Pour plus d'informations sur les limites d'utilisation applicables à votre compte, cliquez <a href=\"https://matrix.org/homeserver/pricing/#usage-limits\">ici</a>.",
+  }
+}
+```
+
+Comparison:
+
+* An advantage is that the homeserver can be as specific as it wishes with the messages. For example, the matrix.org
+  homeserver can specifically refer to the usage plans that are available rather than having to be generic.
```
Johennes, line 208: Could the server not use specific wording in `error` field of the current standard error response, too?

E.g.:

```json5
{
  "errcode": "M_USER_LIMIT_EXCEEDED",
  "error": "You have reached your daily upload limit. More information on the usage limits that apply to your account is available <a href=\"https://matrix.org/homeserver/pricing/#usage-limits\">here</a>. You can upgrade to a Premium account to increase the limits <a href=\"https://account.matrix.org/account?action=org.matrix.plan_management\">here</a>.",
  ..
}
```

From a quick check the only restriction on `error` that I could find in the spec is that it's human readable.

> The `error` string will be a human-readable error message, usually a sentence explaining what went wrong.

t3chguy: Clients shouldn't show a non-i18n'd error field though because it may not match the user's language requirements, so clients end up having to use only `errcode` and mapping those to translations themselves

Johennes: Ah. Ok, that makes sense, thanks.

turt2live: (considering resolved - please open a new thread if untrue)

turt2live:
@mscbot resolve Checklist not complete/started
@mscbot concern TravisR - Implementation not sufficient - doesn't appear to be tested in practice, or validates that existing clients handle the new error code appropriately

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+  from performing certain operations.
+
+The error response must also contain additional fields:
+
+* `info_uri` string (required) - an opaque URI that the client can link the user to in order to get more context on the
+  encountered limit.
+* `soft_limit` boolean (optional, default `false`) - `true` means that the specific limit encountered can be increased.
+  Otherwise it is a hard limit that cannot be increased.
+* `increase_uri` (required if `soft_limit` is `true`) - an opaque URI where the user can undertake actions to increase
+  the encountered limit.
```
anoadragon453: Should we just include `increase_uri` if the limit can be increased, otherwise omit it?

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+> memory or disk space.
+
+As such this error code is currently rendered by
+[existing](https://github.com/element-hq/element-web/blob/c96da5dbf8e20ced4a512a03a75c91f8680e8d40/src/i18n/strings/en_EN.json#L1112)
+[clients](https://github.com/element-hq/element-ios/blob/2dc7b76c44545b3d027cdf0196c6af6eba8932f4/Riot/Assets/en.lproj/Vector.strings#L615)
+as something similar to:
+
+> This homeserver has exceeded one of its resource limits
+
+As such, I think this message would be confusing to users the interim whilst clients updated their implementations and
```
anoadragon453:
```suggestion
As such, I think this message would be confusing to users  in the interim whilst clients updated their implementations and
```

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+could include machine-readable fields for limit types, current usage, and maximum limits. While
+this would provide more information, it would require a more significant change to the error
+response format and could be added in a future MSC if needed.
+
+### Multiple specific error codes
+
+Separate error codes could be introduced for different types
+of limits (e.g., `M_STORAGE_LIMIT_EXCEEDED`, `M_ROOM_LIMIT_EXCEEDED`). However, this approach
+would require many new error codes and doesn't provide significant benefits over a single code
+with descriptive error messages.
```
anoadragon453, line 155: It also prevents homeservers from implementing new limits without writing an MSC (the next alternative also speaks to this).

However, it would solve the translation problem without the homeserver needing to define every translation.

proposals/4335-user-limit-exceeded.md
clokep: Do you mean not have the `soft_limit` flag?

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+
+The HTTP response code should be chosen based on the specification for the individual endpoint. For
+example, the most appropriate code for [`POST /_matrix/media/v3/upload`] would be `403 Forbidden`.
+
+An example response body for the error might look as follows:
+
+```json
+{
+  "errcode": "M_USER_LIMIT_EXCEEDED",
+  "error": "User has exceeded their storage quota of 10GB",
```
uhoreg: nit: since clients sometimes show the `error` to the user, I think

```suggestion
  "error": "You have exceeded your storage quota of 10GB",
```

would look better. (same below)

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,262 @@
...
+The error response must also contain additional fields:
+
+* `info_uri` string (required) - an opaque URI that the client can link the user to in order to get more context on the
+  encountered limit.
+* `soft_limit` boolean (optional, default `false`) - `true` means that the specific limit encountered can be increased.
+  Otherwise it is a hard limit that cannot be increased.
+* `increase_uri` (required if `soft_limit` is `true`) - an opaque URI where the user can undertake actions to increase
+  the encountered limit.
+
+The `info_uri` and `increase_uri` are "opaque" in the sense that the homeserver implementation may choose to encode
```
dbkr: I do think the example UI given might be a bit unclear, specifically that its not really clear to me as a user why I'd ever directly click the 'increase limit' button without knowing more about what the actual limit is. It also looks like it's a button that's going to immediately do something and start charging me more. So, I feel a bit like having both `info_uri` and `increase_uri` might be encouraging this and maybe we should just have the info URI which will give you the option to upgrade if you can (in this situation we would need to keep the `soft_limit` flag to indicate whether it's possible to upgrade (although I might suggest something like `can_upgrade` to be a bit more does-what-it-says-on-the-tin).

proposals/4335-user-limit-exceeded.md
richvdh: I came here with the same question. What would it mean if `increase_uri` is set but `soft_limit` is false? (and if the answer is "that doesn't make sense", then yes let's get rid of `soft_limit`.)

anoadragon453:
@hughns could you clean up the remaining threads and address the remaining concern? This MSC looks nearly ready for FCP otherwise!

turt2live:
The implementation here I think is fine, given its age.

@mscbot resolve TravisR - Implementation not sufficient - doesn't appear to be tested in practice, or validates that existing clients handle the new error code appropriately

hughns:
> The implementation here I think is fine, given its age.

n.b. the implementations were never merged into Synapse or EW, so your concern might stand?

I am also going to update the implementations based on latest feedback.

proposals/4335-user-limit-exceeded.md
hughns: I agree that this is unclear. In 737f53bfac518a920df44cd05c3c560e7db507a0 I have removed `increase_uri` entirely and also renamed the `soft_limit` field to `can_upgrade`.

proposals/4335-user-limit-exceeded.md
hughns: I concur about determining `soft_limit` based on whether `increase_uri` is set.

However, in 737f53bfac518a920df44cd05c3c560e7db507a0 I have removed `increase_uri` entirely and so `soft_limit` (now renamed to `can_upgrade`) is still needed (but no longer redundant).

hughns:
## Existing clients

Behaviour of current clients when encountering this new error code:

Cinny v4.10.2:
Cinny

Element Web v1.12.8:
Element Web

FluffyChat v2.3.0:
FluffyChat

Element X iOS:
Element X iOS

Element Classic iOS:
Element Classic iOS

SchildiChat:
image

Tammy:
image

Commet:
image

< polycule >:
I failed to get it to send an image attachment with or without the error code enabled. 🤷

proposals/4335-user-limit-exceeded.md
hughns: Added in 8b1522ded5cdd6e5dfc9f3437543b991c0bf1129.

hughns:
I have:

- cleaned up the remaining threads
- updated the implementations to match latest version of the spec
- added [screenshots from existing clients](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issuecomment-3769636127)
- added a section to the [PR description](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issue-3367446512) about customisation for the matrix.org homeserver

proposals/4335-user-limit-exceeded.md
```diff
@@ -21,7 +21,7 @@ This proposal adds a new [common error code]
 `M_USER_LIMIT_EXCEEDED` to the Matrix specification. This error code should be returned when a user has exceeded
 limits that are specifically associated with their account, such as:
 
-* **Storage quotas**: When a user has exceeded their allocated storage space for media uploads,
+* **Storage quotas**: When a user has exceeded their allocated storage space for smedia uploads,
```
anoadragon453: small stypo:

```suggestion
* **Storage quotas**: When a user has exceeded their allocated storage space for media uploads,
```

proposals/4335-user-limit-exceeded.md
```diff
@@ -244,11 +258,21 @@ For example:
...
 
+### Include a URI to complete the upgrade action
+
+An earlier version of the proposal included an `increase_uri` that could be attached as a UI button for the user to
+take start the action of upgrading.
+
+However, this led to [uncertainty](https://github.com/matrix-org/matrix-spec-proposals/pull/4335#discussion_r2582153412)
+about what exactly would happen when the button is pressed.
+
+To remove this uncertainty `increase_uri` was removed and the homserver gets to provide a single `info_uri` and it is
```
anoadragon453:
```suggestion
To remove this uncertainty `increase_uri` was removed and the homeserver gets to provide a single `info_uri` and it is
```

anoadragon453 (approved):

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,286 @@
...
+  Otherwise it is a hard limit that cannot be increased.
+
+The `info_uri` is "opaque" in the sense that the homeserver implementation may choose to encode
+information, such as the type of limit encountered, within the URI but it may do so using an encoding of its choosing.
+
+The homeserver may return *different* values for `info_uri` depending on what type of limit was
+reached.
+
+The HTTP response code should be chosen based on the specification for the individual endpoint. For
+example, the most appropriate code for [`POST /_matrix/media/v3/upload`] would be `403 Forbidden`.
```
KitsuneRal, line 47: Should we encourage also using `402 Payment required` where appropriate?

proposals/4335-user-limit-exceeded.md
```diff
@@ -0,0 +1,286 @@
...
+["other" error code](https://spec.matrix.org/v1.16/client-server-api/#other-error-codes) and have the specific endpoints
+listed.
+
+A downside of this is that if a homeserver wished to introduce a new type of limit or quota that was not foreseen, then
+another MSC would be required to introduce it.
+
+Instead, by making it a [common error code] the homeserver operator has flexibility over what types of limit they
+choose without requiring further coordination with clients.
+
+### Use server side translations
```
hughns, line 174: n.b. the was some discussion about this alternative in the [Matrix Client Developers](https://matrix.to/#/!J24qqSs_MrchJWzuVhjH9JGHuMyQiWJO-HzxflwLhdI/$jJFJAzX5gMGJSKAn0H_nLZVRyKqC2YBoHPZc_o9O-yM?via=element.io&via=matrix.org&via=4d2.org) room today, with some folks expressing a preference for this approach.

KitsuneRal:
@mscbot concern how can clients programmatically discern between a soft limit and a payment required limit if both are setup on a homeserver?

proposals/4335-user-limit-exceeded.md
KitsuneRal: I actually think it should be a SHOULD in the MSC text because it may become a real UX issue if both a soft limit and a payment required limit are set up on, say, matrix.org.

turt2live:
The bot isn't aware that FCP is paused when a concern is raised during the countdown. This still has 4 days of FCP to go once the concern is cleared.

proposals/4335-user-limit-exceeded.md
KitsuneRal: Upon discussion in the room - personally I'm not against this. Translatable strings increasingly get introduced to the server side already, indeed.

richvdh:
Ahem:

https://spec.matrix.org/proposals/#process:

> a member of the Spec Core Team can raise a concern and block FCP from completing. This will not reset **or pause** the 5 day FCP timer

turt2live:
The rest of that quote feels relevant:

> ... but FCP will not conclude until all concerns have been resolved.

so there's 0 days left, but it's still in FCP.

See also: https://github.com/matrix-org/mscbot-python/issues/3 (also note that the process on spec.matrix.org is full of lies in many places, but here it looks accurate enough).

proposals/4335-user-limit-exceeded.md
KitsuneRal: After further discussion in the Matrix Client Developers room - okay, not that I want to introduce the 402 code at all costs, I rather want to avoid special-casing matrix.org. I cannot come up with an example that wouldn't be too contrived, of both limits on the same homeserver; that said, I'd rather provide a programmatic way to convey the reason behind the soft limit more clearly. If we want to specify only the "monetary" case - fine, but then we could stick to 402 and leave other options (403, in particular) for a separate MSC. I'm against having a branch in each client to the effect of "if it's matrix.org, it wants money; otherwise, no idea, here's the link to find out".

KitsuneRal:
@mscbot resolve how can clients programmatically discern between a soft limit and a payment required limit if both are setup on a homeserver?

turt2live:
we'll fix the spell check later

hughns:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2315

turt2live:
Merged 🎉

PR #4339, opened 2025-09-02
MSC4339: Allow the user directory to return full profiles

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/user-search-profiles/proposals/4339-user-directory-profiles.md)

### DIFF
--- a/proposals/4339-user-directory-profiles.md
+++ b/proposals/4339-user-directory-profiles.md
@@ -64,7 +64,7 @@ fields. Otherwise, this doesn't increase the surface of available information.
 
 ## Unstable prefix
 
-While this proposal is not considered stable, implementations should use `/_matrix/client/unstable/org.matrix.mscXXXX/user_directory/search` 
+While this proposal is not considered stable, implementations should use `/_matrix/client/unstable/org.matrix.msc4339/user_directory/search` 
 instead. Clients should note the [`M_UNRECOGNIZED` behaviour](https://spec.matrix.org/v1.10/client-server-api/#common-error-codes)
 for servers which do not support the (un)stable endpoint.
 


proposals/4339-user-directory-profiles.md
tulir, line 1: Implementation requirements:

* [ ] Server
* [ ] Client making use of the data?

proposals/4339-user-directory-profiles.md
```diff
@@ -0,0 +1,73 @@
...
+}
+```
+
+The `profile_fields` field controls which fields SHOULD be in `m.profile`. If no `profile_fields` value is given, the
+implementation MAY choose which fields are returned by default. The intention of this field is to reduce the amount of
+data needed to be transmitted to render the client's UI for user search, without prescribing the fields that every
+client should use.
+
+Because this changes the response format of the endpoint, the new endpoint should use `v4`. The full
+endpoint would be `POST /_matrix/client/v4/user_directory/search`.
```
clokep, line 45: Not really related to this, but was considering given to whether the given profile fields should be used to control which fields are searched?

Half-Shot: The spec says:

> The search is performed case-insensitively on user IDs and display names preferably using a collation determined based upon the Accept-Language header provided in the request, if present.

I wonder if we should just relax this to "the server searches relevant profile fields" and leaves those fields up to the implementation rather than be overly specific? I'd rather not extend the spec to specifying specific fields to search, but I think the server should be allowed to search any field it wishes

clokep: I think that makes sense, but... not sure it would fit all use-cases. I'm OK with it being out of scope, but might require another API rev in the future.

proposals/4339-user-directory-profiles.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Potential issues
+
+This will increase the response size of `user_directory`, but since the client has control over the
+fields returned this can be carefully managed.
+
+
+## Alternatives
```
Johennes, line 53: Could you reasonably shoehorn other profile fields into the existing `User` object (thereby not breaking the response format)? The `user_id` field could cause clashes but it seems fairly unlikely that somebody would have a `user_id` field in their profile?

Half-Shot: I'm a bit skeptical that someone hasn't tried it. But even if, I feel like you might one day put non-profile fields in the response anyway like "mutual_rooms", or some other non-profile criteria ("is_bot"?)

I feel for extensibility purposes it's good to get a clean slate. 

But the point is valid, this would be a breaking change. I think that's why it's probably good to have a version bump for this endpoint. If a client tries to use a `v3` endpoint then servers just need to do a little wrapper code to return a reduced profile set.

ASR-ASU:
Does it make sense to extend this MSC to the `/_matrix/client/v3/profile/{userId}` endpoint?

Half-Shot:
> Does it make sense to extend this MSC to the `/_matrix/client/v3/profile/{userId}` endpoint?

That already returns a full profile.

ASR-ASU:
The following request body could be helpful for clients to choose which elements of the profile they would like
to fetch:
```
{
  "m.profile_fields": ["display_name", "avatar_url", "org.example.job_title"]
}
```


That will avoid fetching all the fields from `/_matrix/client/v3/profile/{userId}` or sending several requests to
fetch only the needed fields.

PR #4342, opened 2025-09-05
MSC4342: Limiting the number of devices per user ID

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/device-limit/proposals/4342-device-limit.md)

### DIFF
--- a/proposals/4342-device-limit.md
+++ b/proposals/4342-device-limit.md
@@ -80,7 +80,7 @@ or else they will eventually be unable to login.
 
 ### Alternatives
 
-The number can be higher or lower than 10. 10 was chosen to limit the impact on existing users using matrix.org
+The number can be higher or lower than 30. 30 was chosen to limit the impact on existing users using matrix.org
 as their homeserver.
 
 We could continue to have no limit. However, this will eventually become untenable due to the unbounded
--- a/proposals/4342-device-limit.md
+++ b/proposals/4342-device-limit.md
@@ -27,6 +27,7 @@ In comparison to other encrypted apps:
 ### Proposal
 
 The maximum number of devices a user can have at any one time is reduced to 10.
+Servers MAY have an even lower limit than this. Servers MUST NOT have a higher limit than this.
 
 >[!NOTE]
 > 10 was chosen based on a statistical analysis of the matrix.org database:
--- a/proposals/4342-device-limit.md
+++ b/proposals/4342-device-limit.md
@@ -1,0 +1,80 @@
+## MSC4342: Limiting the number of devices per user ID
+
+Matrix currently does not specify any limit to the number of devices a user can have. This has
+negative impacts on homeservers:
+ - They need to persist an unbounded set of devices. If those devices use E2EE, they also need to persist one-time keys (OTKs).
+ - The way device lists are synced over federation (`prev_id` and `stream_id` in `m.device_list_update` EDUs) is complex and error-prone, primarily to support this unbounded workload.
+ - Any wildcard queries (`/sendToDevice` with a device ID of `*`) causes massive traffic amplification.
+
+..and on clients:
+ - Sending encrypted messages encrypts for all devices in the room. A single user with thousands of devices
+   can significantly impact performance when sending encrypted messages.
+ - Size limits on HTTP requests means a user with thousands of devices risks causing unable to decrypt errors as the response
+   to `/keys/claim` exceeds the size permitted by reverse proxies. As `/keys/claim` is server-scoped, this can impact the ability
+   to claim OTKs for innocent users.
+
+What's worse, often users with many devices are unaware that they have so many e.g because they have logged in via their browser repeatedly.
+This ultimately reduces security of E2EE as messages are encrypted for an unnecessarily large number of sessions, any one of which could be
+compromised to then decrypt the messages.
+
+Because the number of devices affects not just the user with those devices but everyone they communicate with,
+the limit needs to be enforced in the specification, hence this proposal.
+
+In comparison to other encrypted apps:
+- WhatsApp restricts to [5 devices](https://faq.whatsapp.com/378279804439436/?cms_platform=android) (one primary, 4 linked).
+- Signal restricts to [6 devices](https://support.signal.org/hc/en-us/articles/360007320551-Linked-Devices) (one primary, 5 linked).
+
+### Proposal
+
+The maximum number of devices a user can have at any one time is reduced to 10.
+
+>[!NOTE]
+> 10 was chosen based on a statistical analysis of the matrix.org database:
+> - 99.312% of users have <= 5 devices. A limit of 5 will affect 1 in every 145 users.
+> - 99.839% of users have <= 10 devices. A limit of 10 will affect 1 in every 621 users.
+> - 99.931% of users have <= 15 devices. A limit of 15 will affect 1 in every 1449 users.
+
+Attempts to login and exceed this limit returns the error code `M_TOO_MANY_DEVICES`. A client receiving this
+error code should instruct the user to logout an existing device and try again.
+
+>[!NOTE]
+> There's two main options here: prevent the limit being exceed or logout the longest inactive device.
+> Logging out devices causes data loss because it drops to-device events which contain encryption keys.
+> Therefore, this proposal instead prevents the limit being exceeded by returning an error code.
+
+Application service users will be unaffected by this restriction.
+
+>[!NOTE]
+> Application services may have workflows outside expected use cases. To reduce the risk of this proposal
+> having unintended consequences, application service users (and their exclusive namespaced users) are exempt
+> from this limit.
+
+### Potential Issues
+
+Clients which do not log out cleanly by failing to talk to the server when logging out will still be tracked
+as a device. Certain workflows which repeatedly destroy clients in this way (e.g the client runs on a VM
+which is destroyed at the end of every day) will be impacted by this change as they will slowly accumulate
+more and more devices. Clients which behave in this manner need to manage the existing sessions for the user
+or else they will eventually be unable to login.
+
+### Alternatives
+
+The number can be higher or lower than 10. 10 was chosen to limit the impact on existing users using matrix.org
+as their homeserver.
+
+We could continue to have no limit. However, this will eventually become untenable due to the unbounded
+growth of devices. 
+
+### Security Considerations
+
+An attacker who has the password for a user may login many devices to stop the real user from logging in
+new devices. However, if an attacker has your password they can do worse and logout your devices causing
+data loss. The additional denial of service risk seems small in comparison.
+
+### Unstable prefix
+
+The error code is `ORG_MATRIX_MSC4342_M_TOO_MANY_DEVICES` whilst in development.
+
+### Dependencies
+
+None.
--- a/proposals/4342-device-limit.md
+++ b/proposals/4342-device-limit.md
@@ -24,16 +24,33 @@ In comparison to other encrypted apps:
 - WhatsApp restricts to [5 devices](https://faq.whatsapp.com/378279804439436/?cms_platform=android) (one primary, 4 linked).
 - Signal restricts to [6 devices](https://support.signal.org/hc/en-us/articles/360007320551-Linked-Devices) (one primary, 5 linked).
 
+Unlike these apps however, Matrix is a decentralised protocol with many different types of clients/apps available. This biases towards
+users running multiple apps for the same account.
+
 ### Proposal
 
-The maximum number of devices a user can have at any one time is reduced to 10.
+The maximum number of devices a user can have at any one time is reduced to 30.
 Servers MAY have an even lower limit than this. Servers MUST NOT have a higher limit than this.
 
 >[!NOTE]
-> 10 was chosen based on a statistical analysis of the matrix.org database:
+> This proposal has to balance the needs and desires of users with the needs and desires of server admins
+> who cannot allow unbounded amounts of data. This proposal aims to balance this by affecting <1% of "power users"
+> where "power users" is defined as having >= 2 devices logged into matrix.org as of Sept 2025.
+>
+> 30 was chosen based on a statistical analysis of the matrix.org database:
 > - 99.312% of users have <= 5 devices. A limit of 5 will affect 1 in every 145 users.
 > - 99.839% of users have <= 10 devices. A limit of 10 will affect 1 in every 621 users.
 > - 99.931% of users have <= 15 devices. A limit of 15 will affect 1 in every 1449 users.
+> - 99.978% of users have <= 30 devices. A limit of 30 will affect 1 in every 4545 users.
+>
+> Excluding users with only 1 device (e.g throwaway accounts or users who try Matrix once and never again):
+> - 79.809% of users with >1 device have <= 5 devices. A limit of 5 will affect 1 in every 5 users with >1 device.
+> - 95.277% of users with >1 device have <= 10 devices. A limit of 10 will affect 1 in every 21 users with >1 device.
+> - 98.818% of users with >1 device have <= 20 devices. A limit of 20 will affect 1 in every 85 users with >1 device.
+> - 99.367% of users with >1 device have <= 30 devices. A limit of 30 will affect 1 in every 158 users with >1 device.
+> - 99.558% of users with >1 device have <= 40 devices. A limit of 40 will affect 1 in every 226 users with >1 device.
+> - 99.654% of users with >1 device have <= 50 devices. A limit of 50 will affect 1 in every 289 users with >1 device.
+> - 99.712% of users with >1 device have <= 60 devices. A limit of 60 will affect 1 in every 347 users with >1 device.
 
 Attempts to login and exceed this limit returns the error code `M_TOO_MANY_DEVICES`. A client receiving this
 error code should instruct the user to logout an existing device and try again.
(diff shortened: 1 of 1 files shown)


proposals/4342-device-limit.md
turt2live, line 1: Implementation requirements:
* [ ] Server
* [ ] Client (error code)

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,81 @@
...
+The maximum number of devices a user can have at any one time is reduced to 10.
+Servers MAY have an even lower limit than this. Servers MUST NOT have a higher limit than this.
+
+>[!NOTE]
+> 10 was chosen based on a statistical analysis of the matrix.org database:
+> - 99.312% of users have <= 5 devices. A limit of 5 will affect 1 in every 145 users.
+> - 99.839% of users have <= 10 devices. A limit of 10 will affect 1 in every 621 users.
+> - 99.931% of users have <= 15 devices. A limit of 15 will affect 1 in every 1449 users.
+
+Attempts to login and exceed this limit returns the error code `M_TOO_MANY_DEVICES`. A client receiving this
```
MTRNord, line 55: What about users with existing accounts that exceed this limit? Will they only get this on login or will it randomly choose existing exceeding devices?

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,81 @@
...
+> having unintended consequences, application service users (and their exclusive namespaced users) are exempt
+> from this limit.
+
+### Potential Issues
+
+Clients which do not log out cleanly by failing to talk to the server when logging out will still be tracked
+as a device. Certain workflows which repeatedly destroy clients in this way (e.g the client runs on a VM
+which is destroyed at the end of every day) will be impacted by this change as they will slowly accumulate
+more and more devices. Clients which behave in this manner need to manage the existing sessions for the user
+or else they will eventually be unable to login.
```
MTRNord, line 79: What about a safety net in case you lose access to all existing devices, but are at the top of this limit? How would you recover your account in that case, considering most clients probably won't let you log in without a device these days so you likely never get access to the UI to delete them, even if at API level this may be possible.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Alternatives
+
+The number can be higher or lower than 10. 10 was chosen to limit the impact on existing users using matrix.org
```
MTRNord: From experience, while this might not impact 99% of the normal users, this may impact devs which work on various projects and therefore end up with multiple clients they have to work with. Of course this is a subgroup of people, but I wonder how to deal with this. My “high score” in that scenario was around 30 devices due to even simple things like loss of the database and therefore device due to a code bug while developing. To me, this change could be rather annoying and slow development if I have to constantly deal with this. 

(This thread is somewhat more of a conversation starter. I don't really have a good idea how to deal with this edge case. In fact, a limit sounds like a sensible idea. I am just worrying that 10 is too low since the statistics skew to a certain group of users while minorities are left behind despite them having significant impact on the ecosystem.)

clokep: Yeah, seems a bit odd to spec the number of allowed devices instead of just that homeservers may specify a limit, the errors that happen when going over it, and it is recommended to be <= X.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,81 @@
...
+the limit needs to be enforced in the specification, hence this proposal.
+
+In comparison to other encrypted apps:
+- WhatsApp restricts to [5 devices](https://faq.whatsapp.com/378279804439436/?cms_platform=android) (one primary, 4 linked).
+- Signal restricts to [6 devices](https://support.signal.org/hc/en-us/articles/360007320551-Linked-Devices) (one primary, 5 linked).
+
+### Proposal
+
+The maximum number of devices a user can have at any one time is reduced to 10.
+Servers MAY have an even lower limit than this. Servers MUST NOT have a higher limit than this.
```
timedoutuk, line 33: I don't like that servers cannot set higher limits. I frequently have around 10-20 sessions on my main accounts (I use a lot of different clients across a lot of different physical devices), and this restriction would seriously impact how much I can use matrix.

To summarise sessions I currently have on my primary account on my single-user homeserver:

- two cinny sessions on my Linux desktop boot (stable + develop versions)
- two element web sessions on my windows desktop boot (again, stable + nightly)
- Gomuks web
- my invite managing selfbot (needs a crypto session for encrypted log rooms)
- my utility selfbot (needs a crypto session to read and write to encrypted rooms)
- an element x session on my phone
- a Fluffychat session on my phone
- a cinny session on my phone
- an element session on my phone
- a cinny session on my windows desktop boot
- a cinny session on my laptop
- an element web session on my laptop

This brings me to fourteen devices that I use almost daily. And that's excluding ephemeral sessions that I use in virtual machines or on machines I do not own, like university computers or another loaned device, which can bring me to over 20 sessions in the worst case.
I exercise extremely conservative session management as I used to have 60+ sessions, which made E2EE almost unusable, and have since kept watch on my inactive devices and cleaned up any I haven't used in a while or don't plan to.

I cannot resolve my issue by liberally using log out and log in when needed - my initial sync is now over the 500MiB uncompressed mark and takes 5+ minutes for my server to aggregate. This is a similar situation on my envs.net and matrix.org accounts despite being nowhere near as big.

I really hope that this is reworded to discourage increasing the limit rather than forbidding it - I understand that most people won't be affected by this at all, but I also know that it's incredibly easy to hit this limit if you know that matrix is more than the element app.

proposals/4342-device-limit.md
kegsay: I would like to, but as the proposal already says:

> Because the number of devices affects not just the user with those devices but everyone they communicate with, the limit needs to be enforced in the specification

How do you propose resolving this if not a hard limit?

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,81 @@
...
+
+>[!NOTE]
+> 10 was chosen based on a statistical analysis of the matrix.org database:
+> - 99.312% of users have <= 5 devices. A limit of 5 will affect 1 in every 145 users.
+> - 99.839% of users have <= 10 devices. A limit of 10 will affect 1 in every 621 users.
```
jevolk, line 42: This is a superb empirical basis for proposing the value; I will remember this MSC as a benchmark to be cited in other situations. Bravo.

proposals/4342-device-limit.md
kegsay: I've increased the limit to 30 after running an analysis on matrix.org with users who have >1 device.

timedoutuk: 30 is more reasonable :+1:

clokep: @MTRNord OK to resolve this?

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+> - 99.654% of users with >1 device have <= 50 devices. A limit of 50 will affect 1 in every 289 users with >1 device.
+> - 99.712% of users with >1 device have <= 60 devices. A limit of 60 will affect 1 in every 347 users with >1 device.
+
+Attempts to login and exceed this limit returns the error code `M_TOO_MANY_DEVICES`. A client receiving this
+error code should instruct the user to logout an existing device and try again.
+
+>[!NOTE]
+> There's two main options here: prevent the limit being exceed or logout the longest inactive device.
+> Logging out devices causes data loss because it drops to-device events which contain encryption keys.
+> Therefore, this proposal instead prevents the limit being exceeded by returning an error code.
```
Johennes, line 61: (It might be good to move this paragraph to the "Alternatives" section.)

The auto-logout option actually seems a little more user friendly to me than the login limit. The limit introduces the severe risk of locking yourself out of your own account by losing access to all logged in devices. Also, if you're at the limit and want to log in, you now first need to find another logged-in device and log it out which feels cumbersome.

It's a valid point that auto-logout would cause data loss but I think with a reasonable threshold it would only impact power users (much like the current proposal) or sessions that are obsolete anyway.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Potential Issues
+
+Clients which do not log out cleanly by failing to talk to the server when logging out will still be tracked
+as a device. Certain workflows which repeatedly destroy clients in this way (e.g the client runs on a VM
+which is destroyed at the end of every day) will be impacted by this change as they will slowly accumulate
+more and more devices. Clients which behave in this manner need to manage the existing sessions for the user
+or else they will eventually be unable to login.
+
+### Alternatives
```
Johennes, line 81: As another alternative, if you're at the limit and try to log in a new device, maybe the server could allow the login but force you to choose another device to log out (e.g. by returning a new error code `M_TOO_MANY_DEVICES` on all requests except for `/devices` and `/logout`)?

proposals/4342-device-limit.md
kegsay: Clarified.

proposals/4342-device-limit.md
kegsay: I agree with your sentiment. I did another analysis, specifically focusing on users with >1 device and have increased the proposed limit to 30 based on that data. I have also added:

> Matrix is a decentralised protocol with many different types of clients/apps available. This biases towards users running multiple apps for the same account.

But you must understand that we cannot make this a "SHOULD" due to the fact that we don't control servers on the public federation. We _have_ to have a hard limit so federated servers that have users with 1000 devices can be restricted on the receiving side as non-compliant. Without a hard limit, we would have no choice but to accept the devices, making the entire proposal pointless.

proposals/4342-device-limit.md
kegsay: Hmm, maybe that is the best way then. A compromised password would already allow the attacker to log out all your devices, so by cycling out old devices on login it doesn't introduce any new attack vector.

We would need to have clarity on how the definition of "oldest used device" works. For example, if we tie it to `/sync` then this would negatively impact bots which may never call `/sync`.

proposals/4342-device-limit.md
kegsay: I worry that this would complicate the client-server API. It's not clear that it would end with only `/devices` and `/logout` (e.g you may want to know your profile info or account info, how does this integrate with OIDC?). I'd rather have a binary "full access" or "no access" approach personally.

Johennes: Yeah, fair point. Probably not worth the overhead.

proposals/4342-device-limit.md
timedoutuk: Yeah, that makes sense. My request for a SHOULD was more of a compromise since the original ceiling was set to 10 and to allow servers to set higher ones, but you've now raised that to 30 and as mentioned in https://github.com/matrix-org/matrix-spec-proposals/pull/4342#discussion_r2331077149 I see no problem with that (and as you said, a should would just make the proposal merely a suggestion and consequently ineffective). Thanks!

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+- WhatsApp restricts to [5 devices](https://faq.whatsapp.com/378279804439436/?cms_platform=android) (one primary, 4 linked).
+- Signal restricts to [6 devices](https://support.signal.org/hc/en-us/articles/360007320551-Linked-Devices) (one primary, 5 linked).
+
+Unlike these apps however, Matrix is a decentralised protocol with many different types of clients/apps available. This biases towards
+users running multiple apps for the same account.
+
+### Proposal
+
+The maximum number of devices a user can have at any one time is reduced to 30.
+Servers MAY have an even lower limit than this. Servers MUST NOT have a higher limit than this.
```
tonkku107, line 33: Should homeservers do anything to reject other servers over federation that exceed this limit?
If so, how would the transitionary period work while the proposal is being adopted? How would we deal with raising or lowering the limit in the future?
But if nothing is done over federation a server could just ignore the limit and some of the problems persist.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+> Excluding users with only 1 device (e.g throwaway accounts or users who try Matrix once and never again):
+> - 79.809% of users with >1 device have <= 5 devices. A limit of 5 will affect 1 in every 5 users with >1 device.
+> - 95.277% of users with >1 device have <= 10 devices. A limit of 10 will affect 1 in every 21 users with >1 device.
+> - 98.818% of users with >1 device have <= 20 devices. A limit of 20 will affect 1 in every 85 users with >1 device.
+> - 99.367% of users with >1 device have <= 30 devices. A limit of 30 will affect 1 in every 158 users with >1 device.
+> - 99.558% of users with >1 device have <= 40 devices. A limit of 40 will affect 1 in every 226 users with >1 device.
+> - 99.654% of users with >1 device have <= 50 devices. A limit of 50 will affect 1 in every 289 users with >1 device.
+> - 99.712% of users with >1 device have <= 60 devices. A limit of 60 will affect 1 in every 347 users with >1 device.
+
+Attempts to login and exceed this limit returns the error code `M_TOO_MANY_DEVICES`. A client receiving this
```
tonkku107, line 55: There is no mention of next gen auth here. It could be as simple as displaying an equivalent error page to the user.

Since users can log in to their homeserver's management interface without creating a device, they could also be instructed to remove an inactive device right there before seamlessly continuing with the login flow.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+> Therefore, this proposal instead prevents the limit being exceeded by returning an error code.
+
+Application service users will be unaffected by this restriction.
+
+>[!NOTE]
+> Application services may have workflows outside expected use cases. To reduce the risk of this proposal
+> having unintended consequences, application service users (and their exclusive namespaced users) are exempt
+> from this limit.
+
+Servers which retrospectively apply this MSC MAY arbitrarily logout the oldest devices to reclaim resources from those users.
```
timedoutuk, line 70: I don't think "oldest" device is the right idea here. Perhaps the *last recently used* device? It's possible that the user's "oldest" device (going by creation time) is their most recently used device, which may in turn log them out of a session they expect to remain active.

proposals/4342-device-limit.md
MTRNord: > @MTRNord OK to resolve this?

@clokep 

I am very late to this but yes ok to resolve :)

proposals/4342-device-limit.md
timedoutuk: I've implemented this (server) behind an unstable flag in Hammerhead: https://codeberg.org/timedout/hammerhead/pulls/57

proposals/4342-device-limit.md
FSG-Cat: > I would like to, but as the proposal already says:
> 
> > Because the number of devices affects not just the user with those devices but everyone they communicate with, the limit needs to be enforced in the specification
> 
> How do you propose resolving this if not a hard limit?

How are we supposed to enforce this hard limit though is a important question? Servers not aware of this spec change wont care and are we actually going to break E2EE for people ranomly?

proposals/4342-device-limit.md
TheArcaneBrony, line 33: I cannot help but feel like 30 is very low - after logging out all of my unused sesions I still have ~20 of them left (not including non-e2ee sessions, which would add another 60-70 sessions) - if this MSC included non-e2ee devices, I feel like I would be pushed into having separate *accounts* per some amount of devices, which I feel would be counter to this MSC, by now causing N_DEVICES powerlevel updates and membership events (aka PDUs rather than a couple extra EDUs) whenever room lists and powerlevels are synchronised.

N.B. The choice to disambiguate devices from sessions was intentional - as I highly depend on having multiple *sessions* per *device* (different clients, particularly)

TheArcaneBrony (changes requested):

proposals/4342-device-limit.md
TheArcaneBrony: Additionally, the trigger method for this MSC would need to account for the fact that it isn't known at login time whether a client will use E2EE at all - and that's an unreasonable assumption

proposals/4342-device-limit.md
kegsay: Yes, there would have to be some breakage for some people at some point. We can reduce the scope of the breakage, reduce the number of people affected and push the point at which things break out into the future, but ultimately yes, things will have to break. This is generally why we should be designing APIs which are restricted/conservative: it's much easier to open it up later than it is to close a hole later.

proposals/4342-device-limit.md
```diff
@@ -0,0 +1,101 @@
...
+> There's two main options here: prevent the limit being exceed or logout the longest inactive device.
+> Logging out devices causes data loss because it drops to-device events which contain encryption keys.
+> Therefore, this proposal instead prevents the limit being exceeded by returning an error code.
+
+Application service users will be unaffected by this restriction.
+
+>[!NOTE]
+> Application services may have workflows outside expected use cases. To reduce the risk of this proposal
+> having unintended consequences, application service users (and their exclusive namespaced users) are exempt
+> from this limit.
```
tulir, line 68: I think appservices should also be affected, otherwise the "MUST NOT have a higher limit" clause is basically meaningless. I've never heard of any appservice use cases that would require multiple devices per user. End-to-bridge encryption only requires one device globally for the entire appservice rather than one device per user.

If the limit is enforced for all users consistently for all users, we'll be able to apply it over federation in the future (after a lengthy migration period to let servers drop extra devices locally first)

PR #4346, opened 2025-09-09
Fix links and remove linebreaks in checklist

tulir:
Both changes are needed to make it easy to copy the checklist into a comment on a MSC, because GitHub uses hard linebreaks in comments and will not translate absolute paths into URLs pointing at the repo files.

### DIFF
--- a/MSC_CHECKLIST.md
+++ b/MSC_CHECKLIST.md
@@ -11,8 +11,7 @@ MSC authors, feel free to ask in a thread on your PR or in the
 [#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) room for
 clarification of any of these points.
 
-- [ ] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal)
-  specified in the MSC’s PR description?
+- [ ] Are [appropriate implementation(s)](https://spec.matrix.org/proposals/#implementing-a-proposal) specified in the MSC’s PR description?
 - [ ] Are all MSCs that this MSC depends on already accepted?
 - [ ] For each new endpoint that is introduced:
   - [ ] Have authentication requirements been specified?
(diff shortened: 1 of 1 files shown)


MSC_CHECKLIST.md
```diff
@@ -22,31 +21,23 @@ clarification of any of these points.
...
+  - [ ] Is the reason for a new room version clearly stated? For example, modifying the set of redacted fields changes how event IDs are calculated, thus requiring a new room version.
 - [ ] Are backwards-compatibility concerns appropriately addressed?
 - [ ] Are the [endpoint conventions](https://spec.matrix.org/latest/appendices/#conventions-for-matrix-apis) honoured?
   - [ ] Do HTTP endpoints `use_underscores_like_this`?
   - [ ] Will the endpoint return unbounded data? If so, has pagination been considered?
-  - [ ] If the endpoint utilises pagination, is it consistent with
-    [the appendices](https://spec.matrix.org/v1.8/appendices/#pagination)?
-- [ ] An introduction exists and clearly outlines the problem being solved.
-      Ideally, the first paragraph should be understandable by a non-technical audience.
+  - [ ] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/v1.8/appendices/#pagination)?
```
anoadragon453: May as well update this while we're here.

```suggestion
  - [ ] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/latest/appendices/#pagination)?
```

anoadragon453 (approved):
Thank you for doing this!

MSC_CHECKLIST.md
tulir: Hmm, @turt2live usually doesn't like /latest links as they can technically break with any future release

turt2live: usually I don't, but for cases like this it's probably fine/best.

It causes problems when folks link to `/latest` in context of issues/questions, especially if the expected outcome is that the spec changes - the link becomes unhelpful because the problem is no longer visible.

PR #4347, opened 2025-09-09
MSC4346: Emoji verification images

andybalaam:


richvdh:
Related: https://github.com/matrix-org/matrix-spec/issues/13

proposals/4347-emoji-verification-images.md
turt2live, line 1: Implementation requirements:

* [ ] Usage of emoji images in 1 (or ideally more) clients

richvdh: The element clients already use twemoji (with the exception of EXI (https://github.com/element-hq/element-x-ios/issues/2242)), so maybe there is nothing to do here?

turt2live: it'd still be nice to check the "can in fact pull images from a JSON file" box if we can, I think. Especially to understand if client teams expect the spec to be the CDN for those images, or if they'll cache them locally (if they cache them locally, we may want to inline the images rather than URL them)

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,157 @@
...
+  further improve consistency, but might make life difficult for client
+  developers who are not able to render certain image types.
+
+* We could use a different emoji set for the images, for example:
+
+    * [Emojitwo](https://emojitwo.github.io)
+    * [noto-emoji](https://github.com/googlefonts/noto-emoji)
+
+    but twemoji is already in widespread use in Matrix clients, so seems the
+    least disruptive alternative.
```
tulir, line 175: Does anyone other than Element even use twemoji? It's been dead for years

For reference, Beeper is standardizing on Noto color emoji for verification and some other areas

richvdh: @ara4n as the person who wanted to mandate twemoji in https://github.com/matrix-org/matrix-spec/issues/13, any insights here?

kfiven: Cinny also use Tweomji.

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,157 @@
...
+These images are attractive and clear.
+
+These images are provided under the CC BY 4.0 license so may be used with
+proper attribution, which should be added to the spec text and client
+applications.
+
+## Potential issues
+
+* If existing clients use Unicode code points to aid accessibility (e.g. for
+  blind users), the use of an image might reduce accessibility. However, the
```
Johennes: Maybe these clients could continue using the Unicode code points in `alt`, `aria-label` or other platform specific accessibility labels?

andybalaam: Yes, definitely.

andybalaam: Done in ae6a9f30

proposals/4347-emoji-verification-images.md
Huskydog9988: While the main twemojj repo is dead, [jacked’s fork](https://github.com/jdecked/twemoji) is still active, and used by Discord iirc

andybalaam: Since the proposal is to fix on a static set of images, the fact that the source is dead is not very relevant.

It would be nice to choose a set of images that is clear, attractive and suitably licensed.

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,159 @@
...
+into the spec, and freezing the set of images at that time. (Until/unless some
+other MSC requires them to be changed.)
+
+These images are attractive and clear.
+
+These images are provided under the CC BY 4.0 license so may be used with
+proper attribution, which should be added to the spec text and client
+applications.
+
+## Potential issues
```
anoadragon453, line 132: A potential edge case: images at HTTP URLs may not load in a p2p matrix environment. Perhaps such a client should come with those images pre-loaded from the latest spec version it supports.

andybalaam: I am actually expecting all or most clients to come with the images pre-loaded. Maybe I should clarify this somewhere in the proposal?

andybalaam: Further clarified in cbaf8a9997156822cb017d39159eccdb309b31fd , along with wording about how there is no uptime guarantee for these URLs, so client should not depend on them being available, and wording about how we don't expect to change these images in future.

proposals/4347-emoji-verification-images.md
andybalaam: As written (and recently clarified), we expect clients to include the images within their own packaging. It seems like we should require an implementation of this for at least one client.

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,198 @@
...
+None required.
+
+## Dependencies
+
+None.
+
+## Implementations
+
+There is a draft Element Web implementation in
+[element-web#31571](https://github.com/element-hq/element-web/pull/31571), which
```
tulir: Implementations are supposed to be listed in the PR description rather than the document

turt2live: (or on the thread saying "Implementation Requirements" - we need to update the process steps)

andybalaam: Thanks, removed in d3d9ae1e and added to PR description.

proposals/4347-emoji-verification-images.md
andybalaam: As described in the PR description, there is a draft implementation in Element Web, based on a draft change to the spec repo to package the images in the spec NPM package.

turt2live:
This should be ready for review: https://matrix.to/#/!0KNSXYXB_2xtEUkQ9MGBRy5oNIOfAKoq2uIqPZCJbI8/$M9DOtE2TeWrS2qvThi3nJpS1a5SnUur_opWvnT8iPjk?via=matrix.org&via=element.io&via=envs.net

anoadragon453:
This looks good from a server-side perspective, and seems fine otherwise. I think this is ready for wider review.

@mscbot fcp merge

anoadragon453:
Spec Core Team (SCT) members, please ensure that all of the following checks pass before accepting the Matrix Spec Change (MSC).

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,196 @@
...
+emojis below match those shown on your other session".
+
+The spec provides words to describe each character in various languages, and
+also a Unicode code point for each character. Different devices and applications
+render these characters differently, based on the fonts they have available.
+
+This causes confusion: especially for a user who has never seen this before, it
+is not clear what exactly is being asked. (For all they know the list of symbols
+is always the same and the important point is whether the font/rendering styles
+are identical!)
```
clokep, line 23: Is there data to back this up? Is the confusion what the glyphs used to represent the symbols are or the words to use or why they're even being shown these in the first place?

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,196 @@
...
+### Changes to the spec wording
+
+I propose updating the wording of the spec to say graphical clients SHOULD
+include copies of these images in their distributions and display them instead
+of the corresponding Unicode code points rendered using a font. They can choose
+freely whether to use SVG or PNG images, but prefer SVG if both are possible.
+
+I also propose that graphical clients SHOULD display these images in two rows,
+centred vertically, with 4 images on the first row and 3 images on the second,
+with the local translation of the corresponding word below each image.
```
clokep, line 81: This is really specific and might not be possible depending on screen size, etc. I understand it is a SHOULD, but is this really causing confusion?

proposals/4347-emoji-verification-images.md
TheArcaneBrony, line 23: I feel like "if unsure, assume we're talking about font rendering and exact bitmaps" is a bit of a stretch for most users?

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,196 @@
...
+Non-graphical clients or other clients that are unable to use the supplied
+images may continue to use Unicode code points as before.
+
+### Sourcing the images
+
+I propose taking a snapshot of the relevant images from
+[twemoji](https://github.com/twitter/twemoji) at the time of incorporating this proposal
+into the spec, and freezing the set of images at that time.
+
+These images are attractive and clear.
```
TheArcaneBrony, line 101: This line feels very subjective - ie. I dislike colorised unicode emoji fonts.

I feel like this could be a detractor for folks that go out of their way to choose an emoji font (or switch out/override the system one, if one is present)

anoadragon453: @TheArcaneBrony these images would only be shown during the verification process. This wouldn't affect the emoji shown throughout the rest of the app, or in messages.

The purpose of the proposal is to prevent apps from choosing different emoji font, and non-technical users being confused that the following doesn't match, i.e.:

image

proposals/4347-emoji-verification-images.md
turt2live: Users do start to question whether the emoji are "the same" when the layout changes, so in the spirit of making everything consistent I think this is reasonable to suggest to clients.

clokep: Any source for this? (Issue links or something?)

turt2live: I'm not sure how much of it is written down, but we regularly see people at conferences trip over these kinds of things. Usually it's the emoji being slightly different, but I have had to tell folks that the layout of the emoji doesn't matter as much as the order before (back when Element Web didn't strictly use the layout in this MSC).

proposals/4347-emoji-verification-images.md
andybalaam: I have a piece of anecdata on this: a new colleague hit this problem a couple of weeks ago, within days of joining our company. He was trying to verify a device and saw the verification dialog on his phone and his PC. The "bike" emoji was a different color on both devices and he had no idea whether to say they matched or didn't match.

I have spoken to several other people who encountered similar confusion. So I don't have data, but I have heard it directly from several people.

If you combine it with examples like @anoadragon453 's [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4347#discussion_r2642944350) I find it pretty compelling: we have an opportunity to prevent confusion for people very early in their experience of Matrix, with few downsides.

proposals/4347-emoji-verification-images.md
andybalaam: I am not going to die on a hill about this, but I do think it will be helpful if we do it. Many apps already use this layout (e.g. Element, Fractal) and when I recently used an app with a different layout (Nheko) I found it significantly harder to scan the emoji and determine whether they matched. As a SHOULD, it won't cause anyone problems if it is not a suitable layout for their environment.

proposals/4347-emoji-verification-images.md
andybalaam: The point is that the user is being asked whether two things match. They don't know what we mean by that, so it helps if we make them match as exactly as we can (when we want them to answer "yes").

One of the most common times people are asked to say whether two things match like this is in a "spot the difference" puzzle, where subtle differences are precisely what they are looking for.

Bear in mind that many new users have never seen anything like this before, and have zero context about the purpose of it or the underlying mechanisms. It is quite likely closest in their minds to a captcha, which again requires users to spot subtle issues.

proposals/4347-emoji-verification-images.md
andybalaam: The line certainly is subjective, but if we are going to choose a standard set of images for verification, we are going to need to make a subjective judgement about which images are attractive and clear, and usable in as many contexts as possible.

proposals/4347-emoji-verification-images.md
richvdh: I am very supportive of the initiative in this MSC -- I think that presenting the same emoji in the same way will, in general, make it easier to spot differences and improve people's first time experience.

I think it's worth saying though, that, *in addition* to the suggestions in this MSC, it might be a good idea to give a user some guidelines about what they are looking for. In other words: do they really need to spend 30 minutes playing "spot the difference" just in case there is a pixel out of place, or is a quick scan sufficient? Indeed, what is the point of them even doing this exercise? A "learn more" link to some documentation about what we are doing here and (in layman's terms) how it works might go a long way to reducing confusion.

clokep: > I think it's worth saying though, that, in addition to the suggestions in this MSC, it might be a good idea to give a user some guidelines about what they are looking for. In other words: do they really need to spend 30 minutes playing "spot the difference" just in case there is a pixel out of place, or is a quick scan sufficient? Indeed, what is the point of them even doing this exercise? A "learn more" link to some documentation about what we are doing here and (in layman's terms) how it works might go a long way to reducing confusion.

This feels more important than the images being exact matches. It also won't solve regional differences where American speakers will say wrench and British speakers will say spanner (or a thousand other examples I can give just in the US).

Maybe images, while cute, are the wrong way to do this?

andybalaam: > This feels more important than the images being exact matches

I'd disagree with this on the basis that "users don't read things" so providing explanation is not likely to help.

> or a thousand other examples

Note that there are only [64 images/words](https://spec.matrix.org/v1.15/client-server-api/#sas-method-emojio) so hopefully the number of examples of this kind of problem is relatively small.

> Maybe images, while cute, are the wrong way to do this?

That may well be the right way to go, but the intention of this MSC is to make a small, targetted improvement that reduces a point of friction for new users without making anything worse, and certainly does not prevent us from choosing a totally different way to do this in future.

Johennes: On the premise that we already have image-based verification, standardizing it on static images, like this proposal does, definitely improves the status quo.

Beyond that, @clokep has a point IMHO. It feels like images might not be the best way to determine "equality" to begin with. Compared to numbers or strings, images carry intrinsic detail (the pixels) and don't have canonical names that are easily understood across different languages.

richvdh: > > This feels more important than the images being exact matches
> 
> I'd disagree with this on the basis that "users don't read things" so providing explanation is not likely to help.

I think *both* approaches (unifying emoji presentation, improving documentation) will help *a bit*. I don't think we need to pick a single approach, and should proceed with both. This MSC handles the first approach; improving documentation can be handled elsewhere.

> > Maybe images, while cute, are the wrong way to do this?
> 
> That may well be the right way to go, but the intention of this MSC is to make a small, targetted improvement that reduces a point of friction for new users without making anything worse, and certainly does not prevent us from choosing a totally different way to do this in future.

+1.

proposals/4347-emoji-verification-images.md
```diff
@@ -0,0 +1,196 @@
+# MSC4347: Emoji verification images
```
ara4n, line 1: I'm afraid to say that I'm dubious about this (with about 7/10 strength of dubiousness). As a client implementor, i don't want the aesthetics of my app messed up by predefined imagery, whether that's due to disliking twemoji's visuals or wanting a proper dark theme, or not wanting circles around the images, or wanting to show 3D emoji because my UI is 3D, or my UI can't render images (terminal) or whatever.

Rather than complicating the spec with a bunch of images like this, might it be easier to instead recommend that users always compare the words rather than the emoji (and treat the emoji just as cute illustrations of the words)?  This obviously still suffers from the <del>wrench/spanner</del> spanner/schraubenschlüssel problem, but perhaps we could solve that by showing the words in the languages of both users if the i18n differs between them (at the risk of having to leak locale or at least language to the other user)?  So you might have something like:

# 🔧
Spanner
(Schraubenschlüssel)

I agree with the assessment elsewhere that it may just be that emoji are a bad verification mechanism in general (personally I think we should aim to verify mainly based on key transparency and transitive trust in future, minimising SAS verification as an edge case, whereupon we could just use numbers or something).  But for a quick fix for this problem, i'm really not sure that forcing the same imagery on everyone is a good idea when we could instead fix the instructions to make it clear that users shouldn't compare the images.

richvdh: (aside: the wrench/spanner problem doesn't really exist, because we always use "Spanner", even for en-US users. Possibly to their confusion/entertainment. But the problem still arises between users of different languages. So this doesn't really change the point here, other than to note it should maybe referred to as the "spanner/schraubenschlüssel" problem, not least because everything sounds more impressive with a good German compound thrown in.)

andybalaam: > aesthetics of my app messed up by predefined imagery

So I guess we're balancing potentially messed up aesthetics with user confusion. That's always going to be a judgement call, but given that I have personally witnessed the user confusion I still lean towards trying to reduce it.

> complicating the spec with a bunch of images

This does increase the size of the spec, but I'm not sure it meaningfully complicates implementing SAS - in some circumstances it might make it easier because it reduces the burden of guaranteeing the client has a suitable font that can render all the relevant emoji.

>  fix the instructions to make it clear that users shouldn't compare the images

I still maintain that users often don't read, so instructions may be missed, and even if they eventually figure it out by reading the instructions, it's still a jarring experience to be confronted with very early in your relationship with Matrix.

> I agree with the assessment elsewhere that it may just be that emoji are a bad verification mechanism in general

Very possible but a) this is a small fix we could apply now, without huge effort and b) I think we are likely to keep some kind of verification like this as a last resort, so why not make it better?

Anyway, as I said above, I think the key decision point here is a judgement call, so let's see what the SCT members' judgements are.

richvdh:
MSC4405 proposes to deprecate emoji verification rather than trying to improve it.

andybalaam:
This proved much more controversial than I expected, and it seems like rather than trying to improve emoji verification people would prefer to remove it, which is fine by me. Actually, I think removing it is better.

So I am closing this and hoping that [MSC4405: Deprecate the emoji method for SAS verification](https://github.com/matrix-org/matrix-spec-proposals/issues/4405) will be merged instead.

proposals/4347-emoji-verification-images.md
t3chguy: Worth noting also that Element Web is moving away from the strict 4-3 layout like this comment asks for on smaller viewports as part of accessibility work where the text was too cramped in such a case

turt2live:
@mscbot fcp cancel

PR #4348, opened 2025-09-09
MSC4348: Portable and serverless accounts in rooms

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/portable-accounts-in-rooms/proposals/4348-portable-accounts-in-rooms.md)

### DIFF
--- a/proposals/4348-portable-accounts-in-rooms.md
+++ b/proposals/4348-portable-accounts-in-rooms.md
@@ -106,12 +106,33 @@ localparts as Account
 Keys](https://github.com/matrix-org/matrix-spec-proposals/pull/4243) provides
 the prior work for this MSC and much of the inspiration.
 
-The difference is MSC4243 takes an iterative approach towards
-portable identity without exploring its implications or making the
-changes necessary to facilitate it. MSC4243 also requires use of a
-policy server in order to keep the room secure, whereas this proposal
-does not, as we depend upon MSC4345 server participation to provide us
-security from malicious keys and malicious backfill.
+The differences between this and is MSC4243 are
+
+- MSC4243 takes an iterative approach towards
+  portable identity without exploring its implications or making the
+  changes necessary to facilitate it.
+
+- MSC4243 requires use of a policy server in order to keep the room
+  secure, whereas this proposal does not, as we depend upon MSC4345
+  server participation to provide us security from malicious keys and
+  malicious backfill.
+
+- Because MSC4243 takes an iterative approach, the User ID format does
+  not need to change. This provides the argument that the MSC is
+  non-breaking. But the reality is different because of the security
+  implications that the MSC imposes on clients and servers. We believe
+  that the MSC still imposes breaking changes on clients that interact
+  with power levels and also moderation.  It also mandates changes in
+  clients to ensure that unverified users are shown in a safe way.
+
+- MSC4243 embeds the domain into the account key user ID, which always
+  allows for that information to be faked at the DAG level. Clients
+  must be protected by servers from this with consideration to replace
+  User ID's in client formatted events. This likely will be a
+  complicating factor for that MSC when events interact with the power
+  level and access control mechanism of the room.
+
+
 
 ## Security considerations
 


proposals/4348-portable-accounts-in-rooms.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Complement tests

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+  Just like MSC4243, network requests are not required to verify the
+  signature of inbound events as all public keys are available in the
+  DAG.
+
+- Unlike
+  [MSC4243](https://github.com/matrix-org/matrix-spec-proposals/4243),
+  no endpoint or network requests are needed for servers or clients to
+  verify server ownership over an account, as all events are co-signed
+  by the room scoped server key. This is both a security and
+  consistency enhancement.
```
Gnuxie, line 51: We need to be clear about whether servers do own accounts or whether participating in the room from multiple servers at the same time is possible.

Gnuxie: It seems like there would be no way to enforce this other than through a change in membership otherwise

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+of the queried account key.
+
+### Obtaining account and profile information as a client
+
+The account name is annotated in `unsigned.account`. This is the
+account name only and not the account name user ID.
+
+The domain can be found from MSC4345's `unsigned.server_domain`.
+
+## Potential issues
```
Gnuxie, line 75: I'd like someone to talk to me about E2EE in the context of this MSC because I actually have no idea how tf E2EE already works on Matrix.

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+This is desirable because:
+
+- Portable snd serverless accounts accounts are now native to this new room model.
+- Pseudo identity is supported in the room model for servers and room
+  admins that wish to immediately implement it.
+- Servers can immediately be implemented as relays for P2P clients.
+- Within the context of MSC4345, the server key can be rotated without
+  effecting users or the room.
+
+## Proposal
```
Gnuxie, line 30: We need to specify how precisely the server co-signs the event that has first been signed by the account key. And how this interacts with authorization rules, event creation, and room creation.

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+>    do not remove the user ID so these messages persist in the room.
+
+In addition to the above, we go further and remove the server from user ID's completely:
+
+This is desirable because:
+
+- Portable snd serverless accounts accounts are now native to this new room model.
+- Pseudo identity is supported in the room model for servers and room
+  admins that wish to immediately implement it.
+- Servers can immediately be implemented as relays for P2P clients.
```
Gnuxie, line 26: This isn't necessarily true. There are situations where MSC4345 would require cooperation from an account key.

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+to servers that are participating in the room. We assess this is
+acceptable because we expect completing the same registration
+requirements on the new server as they would with a brand new user
+account will be required to migrate an account.
+
+We expect servers with weak registration requirements or setup for the
+use of ban evasion will be identified and removed as they are
+currently. However, with the security enhancements of MSC4345, such
+servers can no longer join rooms unanticipated. So we have an
+advantage.
```
Gnuxie, line 151: This implies that the membership event of a user needs explicitly tying to the co-sigining server's participation, and if either the co-singing server key or the account key are revoked, then the membership event cannot be used to authorize events.

I think that it is ok for a new membership event to be created with the same account key co-signed by a new server.

proposals/4348-portable-accounts-in-rooms.md
```diff
@@ -0,0 +1,162 @@
...
+### Changes to room membership
+
+Room membership must use the account key user ID as the `state_key`.
+
+### Obtaining account and profile information as a server
+
+We use the same `POST /_matrix/federation/v1/query/accounts` endpoint
+from MSC4243, with the exception that the requested server must verify
+that the requesting server is participating in common rooms to those
+of the queried account key.
```
timedoutuk, line 66: Forgive me if I've simply missed where this is expanded upon, but in MSC4243, the server is meant to query the origin that is in the user ID. 
Given this proposal then removes that server name, where is the remote server supposed to query? I don't see a source of origin in events anymore, since the room ID no longer has an origin, the event ID no longer has an origin, and now the sender no longer has an origin.
Should the signing servers be queried instead? What if the event is only signed by the origin, and the origin is no longer responding to requests?

Gnuxie: That comes from what is currently called the `unverified_domain` in the parent MSC4345 which is annotated on the `m.server.participation` event for the co-signing server key (that can be found by looking for the public key as the state key).

If the server is no longer responding then profile information cannot be obtained

PR #4349, opened 2025-09-10
MSC4349: Causal barriers and enforcement

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/causal-barriers-and-enforcement/proposals/4349-causal-barriers-and-enforcement.md)

### DIFF
--- a/proposals/4349-causal-barriers-and-enforcement.md
+++ b/proposals/4349-causal-barriers-and-enforcement.md
@@ -55,7 +55,7 @@ authorization_.
 
 ### Deferred authorization as causal enforcement
 
-When a server recieves an event which is concurrent with a _causal
+When a server receives an event which is concurrent with a _causal
 barrier_ the receiving server should determine whether it has complete
 knowledge of all of the _causal barrier_ event's causal predecessors.
 
@@ -64,7 +64,7 @@ soft fail the concurrent event (which is what happens currently in
 Matrix).
 
 However once the server has all of the _causal predecessors_ for a
-_causal barrier_ it can infact revist the decision, and make a final
+_causal barrier_ it can infact revisit the decision, and make a final
 decision.
 
 To do this, the server consults the predecessors of the causal barrier
@@ -122,14 +122,14 @@ solutions that can be explored:
 
 - Policy servers as causal authorities: Policy servers can act as
   causal authorities. They could do this by performing a
-  full-synchronisation on each event they recieve. And then only
+  full-synchronisation on each event they receive. And then only
   signing events that are consistent with the policy server's vector
   clock frontier.
 
 - Deferred authorization as causal enforcement (described in the MSC
   introduction), combined with a system that allows distrust to be
   expressed in senders in order to provide a manual resolvation to
-  forks that would constitute a politcal problem.
+  forks that would constitute a political problem.
 
 This would allow us to immediately begin to use consistent terminology
 within the context of [MSC4345: Server key identity and room


Gnuxie:
Context: This MSC was made because MSC4345 described an informal version of the _deferred authorization causal enforcement_ model discussed in this MSC. Which uncovered the problem of causal barrier conflicts.

proposals/4349-causal-barriers-and-enforcement.md
turt2live, line 1: Implementation requirements:

* [ ] General sense of agreement among protocol developers towards adopting definitions
  * Metric: FCP starts successfully (likely ignoring the checkbox for "has implementation").

proposals/4349-causal-barriers-and-enforcement.md
```diff
@@ -0,0 +1,158 @@
...
+
+Such a _causal barrier enforcement model_ uncovers a side-effect of
+soft-failure which is less obvious.
+
+If the sender of a ban event was to deliberately ignore the history of
+the target, they now have the power to erase the target's history for
+all room participants. Unfortunately, this already can happen today,
+it's just room participants with a more complete synchronisation are
+unable to notice. And anyone with a partial synchronisation, such as a
+new joiner, can.
```
Gnuxie, line 85: To be honest I get confused every time i revisit this. It's not actually clear to me what does currently happen here.

proposals/4349-causal-barriers-and-enforcement.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Deferred authorization as causal enforcement
+
+When a server receives an event which is concurrent with a _causal
+barrier_ the receiving server should determine whether it has complete
+knowledge of all of the _causal barrier_ event's causal predecessors.
+
+If knowledge is incomplete, the server makes an immediate decision to
+soft fail the concurrent event (which is what happens currently in
+Matrix).
```
Gnuxie, line 64: There's more nuance to this, in that the way this decision gets made is by determining if the event is authorizable with respect to current authorization state rather than the cited state from `auth_events`.

proposals/4349-causal-barriers-and-enforcement.md
```diff
@@ -0,0 +1,158 @@
...
+- Deferred authorization as causal enforcement (described in the MSC
+  introduction), combined with a system that allows distrust to be
+  expressed in senders in order to provide a manual resolvation to
+  forks that would constitute a political problem.
+
+This would allow us to immediately begin to use consistent terminology
+within the context of [MSC4345: Server key identity and room
+membership](https://github.com/matrix-org/matrix-spec-proposals/pull/4345#discussion_r2336627872)
+and other MSC's which may have security concerns that cannot be
+expressed without this understanding.
```
Gnuxie, line 138: This should come before the suggested solutions. Also the link is broken and the apostrophe shouldn't be there

proposals/4349-causal-barriers-and-enforcement.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Potential issues
+
+- None considered
+
+## Alternatives
+
+- None considered
+
+## Security considerations
```
Gnuxie, line 148: It might not be safe to use `auth_events` to create a vector clock frontier for a sender.

This is because if the sender joined via a partial synchronisation and only have a snapshot of the current auth state from `/make_join` then they can be forced to indirectly reference unauthorized events. Implementations may also be lazy about recursively verifying auth_state from `prev_events` when merging branches too. Which absolutly removes the ability to use `prev_events` to form a vector clock frontier.

Gnuxie: Effectively this makes it impossible for someone rebuilding the entire auth_chain to form a vector clock frontier from the auth_events and prev_events to reason about the causal knowledge of a sender. Because the frontier can be full of random events that the sender hasn't bothered to either verify or authorize.

Gnuxie: And that stops ANYONE from being able to reliably reason about causal barrier violations.

Gnuxie: image

Gnuxie: Blue is trying to ban yellow in order to erase yellow's events. But new joiners can't tell wtf is going on here and can't even tell that there is a causal barrier violation because Blue has been tricked by Green into tying in the events Blue tried to erase.

Essentially, Yellow's membership is a causal predecessor of Blue's ban on yellow. The events are not conflicting. But blue has been tricked into referencing a yellow event which would normally be soft failed because of the causal barrier from Blue's ban on yellow. A ban that doesn't include yellow's event as a casual predecessor.

Gnuxie: probably easier to understand if it looks like this 

image

Gnuxie: There's no reliable way to prevent this. It's impossible.

Gnuxie: image

image

Gnuxie: Problem discussion moved to https://github.com/matrix-org/matrix-spec-proposals/pull/4353

kathleenmotley49-maker (approved):

PR #4350, opened 2025-09-10
MSC4350: Permitting encryption impersonation for appservices

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/appservice-encryption-impersonation/proposals/4350-appservice-encryption-impersonation.md)

### DIFF
--- a/proposals/4350-appservice-encryption-impersonation.md
+++ b/proposals/4350-appservice-encryption-impersonation.md
@@ -183,4 +183,6 @@ devices), so it should be easily noticeable.
 device metadata.
 
 ## Dependencies
-This MSC enhances [MSC4153], but does not formally depend on it.
+This MSC builds on [MSC4153]. While this proposal would technically work without
+4153, some security properties assume that 4153 is accepted either before or at
+the same time as this MSC.


proposals/4350-appservice-encryption-impersonation.md
tulir, line 1: Implementation requirements:

* [ ] Bridge or other appservice making use of impersonation
* [ ] Client that validates impersonation requirements
* [ ] Server that exposes `signatures` properly

proposals/4350-appservice-encryption-impersonation.md
```diff
@@ -0,0 +1,180 @@
...
+to receive keys the standard way.
+
+### Client behavior
+Following the rules of [MSC4153], an impersonated message is considered to be
+from a "non-cross-signed device" if:
+
+* the user specified in `impersonator` is not a member of the room
+* the device specified in `impersonator` is not cross-signed itself
+* the user who owns the impersonatable device has cross-signing keys and the
+  impersonatable device is not signed by the user's self-signing key
```
dkasak, line 51: What happens if there are no cross-signing keys?

proposals/4350-appservice-encryption-impersonation.md
```diff
@@ -0,0 +1,180 @@
...
+`/keys/query`, gets a new optional `impersonator` field. The field contains a
+full device keys object, excluding `signatures`.
+
+The device owned by the ghost user (or real Matrix user in case of double
+puppeting) with the `impersonator` field will be referred to as the
+"impersonatable" device, while the bridge bot's device is the "impersonator"
+device.
+
+The `algorithms` and `keys` fields are still present and required, but MUST be
+empty when an impersonator is defined. Because there are no keys, `signatures`
```
dkasak: What does this mean? Both fields need to be set to an empty list?

If so, I think it would be better to say it like that explicitly. And second, why require them to be defined but be an empty list?

proposals/4350-appservice-encryption-impersonation.md
```diff
@@ -0,0 +1,180 @@
...
+another user is allowed to cryptographically impersonate the owner of the
+device.
+
+The device keys object, which is written by `/keys/upload` and read by
+`/keys/query`, gets a new optional `impersonator` field. The field contains a
+full device keys object, excluding `signatures`.
+
+The device owned by the ghost user (or real Matrix user in case of double
+puppeting) with the `impersonator` field will be referred to as the
+"impersonatable" device, while the bridge bot's device is the "impersonator"
```
dkasak, line 31: Given that the "impersonatable device" has no keys, in what way is it a device being impersonated? Rather, I believe "impersonatable device" is a bit of a misnomer; it is simply a way for a user to specify a foreign device that is allowed to impersonate them. It is the user being impersonated, not any specific device of theirs.

proposals/4350-appservice-encryption-impersonation.md
```diff
@@ -0,0 +1,180 @@
...
+all rooms (see alternative below).
+
+## Alternatives
+### Send room keys to impersonator devices
+To remove the requirement for the bridge bot being a member of all rooms, we
+could define that clients must send keys to any impersonator devices in the
+room. However, we prefer an explicit bot membership over a hidden key sharing
+mechanism. Additionally, there would need to be a way to turn off sending keys
+to impersonator devices to avoid the bridge receiving keys in all encrypted
+rooms when using double puppeting.
```
dkasak:
```suggestion
mechanism. Additionally, there would need to be a way to turn off sending
message keys to impersonator devices to avoid the bridge receiving message keys
in all encrypted rooms when using double puppeting.
```

Let's be explicit about which keys we mean please.

proposals/4350-appservice-encryption-impersonation.md
```diff
@@ -0,0 +1,180 @@
...
+[MSC4261](https://github.com/matrix-org/matrix-spec-proposals/pull/4261) could
+be used to avoid receiving keys for the ghost users, but it would still require
+the bridge to create Olm sessions between each ghost and recipient device to
+share keys to users.
+
+## Security considerations
+The requirement to cross-sign the impersonatable device could be bypassed by a
+malicious server that hides the user's cross-signing keys. However, hiding
+cross-signing will also prevent other devices of that user from working, so it
+should be easily noticeable.
```
dkasak: I think this is concerning and should likely be fleshed out more. Can you spell out why hiding the cross-signing keys would stop other devices from working?

tulir: If you require cross-signing for real devices with keys (as per MSC4153), then suppressing cross-signing keys server-side will make clients stop encrypting messages to those devices and hide messages from those devices.

proposals/4350-appservice-encryption-impersonation.md
tulir: Then the device is trusted with just the impersonator's signature

proposals/4350-appservice-encryption-impersonation.md
tulir: Yeah need to figure out some better terms. The impersonator is the bridge bot's device which is obvious enough, but I'm not sure what to call the device entry that points at the impersonator without it being confusable with the impersonator itself.

proposals/4350-appservice-encryption-impersonation.md
tulir: The fields are currently required, so the main reason is to avoid changing that. They could totally be removed too if that's preferred, I was mostly thinking about what would cause the least unexpected behavior in old clients.

proposals/4350-appservice-encryption-impersonation.md
tulir: Have added a paragraph below the bullet to call it out explicitly

proposals/4350-appservice-encryption-impersonation.md
tulir: The security consideration now specifically mentions MSC4153 as the reason why it should be safe

PR #4352, opened 2025-09-15
MSC4352: Customizable HTTPS permalink base URLs via server discovery

DonPrus:
**Summary**

This MSC proposes to generalize HTTPS permalinks by introducing a generic *Matrix link resolver* (drop-in replacement for `matrix.to`) and a standardized way for clients to discover a resolver base URL via `/.well-known/matrix/client`. This enables organizations to use branded resolver domains while keeping the existing matrix.to path/query grammar intact.

**Motivation**

- Spec currently hardcodes `matrix.to` in the "matrix.to navigation" section, making custom resolvers non-standard.
- The `matrix:` URI scheme is standard, but HTTPS permalinks are still required for email/web contexts.
- Fragmentation already exists (Element `permalinkPrefix`, Synapse `client_base_url`) without a protocol-level knob.

**What this MSC introduces**

- A resolver-agnostic concept for HTTPS permalinks (same grammar as matrix.to).
- Discovery precedence: user setting → `/.well-known/matrix/client` → fallback to `https://matrix.to`.
- Unstable key: `org.matrix.msc4352.permalink_base_url` → Stable key: `m.permalink_base_url`.

**File**

- `proposals/4352-customizable-permalink-base-url.md` (this PR assigns the MSC number)

**References**

- Spec process & how MSC IDs / filenames work; proposal template & checklist.
- Spec: `matrix:` URI scheme and matrix.to navigation (Appendices/URIs).
- Prior art in the ecosystem:
  - Element Web `permalinkPrefix` (docs/historical config.md)
  - Synapse `client_base_url` for emails (notification links defaulting to `https://matrix.to`)

**Rendered Markdown**

[Rendered: proposals/4352-customizable-permalink-base-url.md](https://github.com/DonPrus/matrix-spec-proposals/blob/msc-custom-link-resolver/proposals/4352-customizable-permalink-base-url.md)

- **Client (qualifying)**: iamb — discovery of `org.matrix.msc4352.permalink_base_url`,
  `:permalink` for outside-of-Matrix sharing, and automatic linkify to `matrix:` URIs on send.
  *Tracking PR*:  https://github.com/ulyssa/iamb/pull/512

**Author**

Author: @DonPrus 
idea : https://github.com/element-hq/element-meta/discussions/2399

### DIFF
--- a/proposals/4352-customizable-permalink-base-url.md
+++ b/proposals/4352-customizable-permalink-base-url.md
@@ -60,15 +60,36 @@ Clients MUST validate that the discovered value is a valid HTTPS origin and MUST
 
 ### 3. Generation algorithm (normative)
 
-When a client needs an HTTPS permalink (e.g., for emails, social platforms, or contexts without a `matrix:` handler):
+**Scope:** This algorithm applies **only** to generating HTTPS permalinks for **outside‑Matrix** contexts (e.g. email,
+social networks, the web). Clients **MUST NOT** use resolver‑based HTTPS links for in‑app mentions/links/navigation.
 
-1.  Determine `permalink_base_url` by the precedence above.
-2.  Construct the permalink path using the **matrix.to navigation grammar** from the spec (identifier encoding, optional
-    event segment, `via` routing parameters, etc.).
-3.  Produce `https://permalink_base_url/#/<encoded identifier>[/<encoded event>][?<args>]`.
+1.  Determine `permalink_base_url` by the precedence in Section 2 (user override → well‑known → client default).
+2.  Construct the path using the matrix.to navigation grammar defined by the spec (identifier encoding, optional event
+    segment, routing parameters like `via`, etc.).
+3.  Produce `https://<permalink_base_url>/#/<encoded identifier>[/<encoded event>][?<args>]`.
 
-> Note: Clients SHOULD continue to prefer the `matrix:` URI scheme where appropriate (in-app deep linking, OS-level
-> handlers). HTTPS permalinks remain the web-friendly representation.
+Clients **SHOULD** continue to prefer the `matrix:` URI scheme for deep‑linking whenever appropriate.
+
+#### In‑app usage (normative)
+
+Clients **MUST NOT** use resolver‑based HTTPS permalinks for in‑app links (mentions, room/event links, or any links
+displayed to Matrix users). Clients **MUST** prefer the `matrix:` URI scheme (or native navigation) for all in‑app UX.
+
+#### Sending links in Matrix (normative)
+
+When a user types/pastes a resolver‑style HTTPS permalink in a message composer, the client **SHOULD** preserve the
+visible text but **MUST** emit, in `formatted_body`, an anchor whose `href` is the equivalent `matrix:` URI, to ensure
+cross‑client interoperability. The plain‑text `body` may keep the original text.
+
+#### Receiving (parsing)
+
+Clients **MUST** handle `matrix:` URIs. Clients **MAY** additionally recognise HTTPS permalinks that follow the
+matrix.to navigation grammar regardless of hostname, but this is **not required** by this MSC.
+
+#### Non‑goals
+
+This MSC does **not** standardise applying resolver prefixes to in‑app mentions or in‑app permalinks. Clients **MUST
+NOT** rewrite mentions or in‑app links to use resolver domains.
 
 ### 4. Resolution behavior
 


proposals/4352-customizable-permalink-base-url.md
tulir, line 1: Implementation requirements:

* [ ] Client

proposals/4352-customizable-permalink-base-url.md
```diff
@@ -0,0 +1,121 @@
+# MSC4352: Customizable HTTPS permalink base URLs via server discovery
+
+## Problem
+
+Matrix currently supports two major ways to reference resources: the `matrix:` URI scheme and HTTPS-based permalinks commonly using `matrix.to`. The specification explicitly documents the `matrix.to` form and even states that such URIs "must always start with `https://matrix.to/#/`" for the matrix.to navigation format. This tight coupling to a single domain prevents organizations and deployments from using their own branded resolver domains for invites and permalinks, despite valid branding, trust and policy needs. It also fragments behavior across clients which already implement ad-hoc overrides. [Spec references: Appendices/URIs and matrix.to navigation; `matrix:` scheme]
```
tulir: Please wrap lines to ~120 characters

proposals/4352-customizable-permalink-base-url.md
```diff
@@ -0,0 +1,152 @@
...
+
+> Note: Clients SHOULD continue to prefer the `matrix:` URI scheme where appropriate (in-app deep linking, OS-level
+> handlers). HTTPS permalinks remain the web-friendly representation.
+
+### 4. Resolution behavior
+
+This MSC does **not** require clients to perform network requests to the resolver domain. Like `matrix.to`, clients
+SHOULD treat HTTPS permalinks primarily as **hints** for in-app navigation: parse the identifier and act locally if
+possible (join, open DM, navigate to event). The resolver service is optional infrastructure for environments without a
+Matrix-aware client.
```
tulir, line 99: This part sounds suspicious. Does the MSC suggest that custom permalink prefixes should be used in-app too instead of only outside the app? If so, how would a client on another homeserver detect that it's a matrix.to-equivalent link?

Element web used to have (or still has? not sure, I hope it was fixed) a bug/anti-feature where `permalinkPrefix` was applied in @mentions and such in messages too. The result was that mentions, room links and other permalinks were broken for everyone on other clients. I'd be strongly opposed to codifying such a broken feature in the spec.

Even if it's not intended to be used in-app, if a "share" button gives out such links, there's a risk users will send them in-app, leading to worse UX for everyone on other servers (because the app wouldn't be able to route the links internally). If the spec wants to standardize allowing custom links, it might also need to mandate some kind of automatic conversion to `matrix:` URIs when sending such links in-app. For example, when the user sends `https://custom.permalink.to/#/#room:example.com`, the client could linkify it before sending and produce `<a href="matrix:r/room:example.com">https://custom.permalink.to/#/#room:example.com</a>`, which would work in all clients

DonPrus: Thanks for the detailed feedback — I fully agree with the risks. MSC4352 does not intend to use a custom HTTPS
resolver for in‑app navigation. The goal is only to generate shareable links for outside‑Matrix contexts
(email/social/other messengers/etc.), while `matrix:` URIs remain the in‑app priority.

To remove ambiguity, here are normative clarifications I propose to add:
1. In‑app usage (normative):
Clients MUST NOT use a resolver-based HTTPS permalink for in‑app links (mentions, room/event links,
or any navigation shown to Matrix users).
Clients MUST prefer the matrix: URI scheme (or native navigation) for all in‑app UX.
2. Sending links in Matrix (normative):
If a user pastes a https://<resolver>/#/<id>[/<event>][?...] link, the client SHOULD keep the visible text but
MUST emit a link whose href is the equivalent matrix: URI in formatted_body, so it works in all clients.
The plain‑text body may keep the original text; clickability is provided by formatted_body.
No network calls to the resolver are needed, and no domain is imposed on other clients.
3. Receiving (non‑breaking):
Clients MUST support matrix: URIs.
Clients MAY additionally recognise HTTPS permalinks that match the matrix.to navigation grammar regardless of
hostname, but this is not required by MSC4352.
This means MSC4352 does not require clients to parse arbitrary domains; matrix: URIs provide the compatibility path.
4. Editorial tweak to “Generation algorithm”:
State that the algorithm applies to share/copy for outside‑Matrix contexts, not in‑app navigation.
Add an explicit preference for matrix: URIs for any in‑app scenarios.

If the MUST/SHOULD/MAY wording looks reasonable, I can update the MSC

famfo: > The goal is only to generate shareable links for outside‑Matrix contexts
(email/social/other messengers/etc.), while matrix: URIs remain the in‑app priority.

What are the requirements for a web service? Should it be something like the matrix.to server hosted alongside a homeserver?

Why another URI scheme (which will inevitably suffer from linkrot when homeservers disappear again)? The Matrix URI is both a registered URI scheme with the [iana](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml#well-known-uris-1) and is a [safelisted schema](https://html.spec.whatwg.org/multipage/system-state.html#safelisted-scheme) for HTML5, so it can be used outside of Matrix clients already. Wouldn't it preferable to improve client support for the matrix URI scheme to not have the centralization around resolver servers while still having the trust of "this is an official URI scheme my client understands"?

PR #4345, opened 2025-09-08
MSC4345: Server key identity and room membership

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/server-key-identity-and-room-membership/proposals/4345-server-key-identity-and-room-membership.md)

Signed-off-by: Gnuxie [Gnuxie@protonmail.com](mailto:Gnuxie@protonmail.com)

PR #4353, opened 2025-09-15
MSC0000: Per-origin linear chain

Gnuxie:


proposals/4353-per-origin-linear-chain.md
tulir, line 1: Implementation requirements:

* [ ] Server

proposals/4353-per-origin-linear-chain.md
```diff
@@ -0,0 +1,164 @@
...
+through sequential use of `/send`. It's not understood by me
+whether this is a bug or an intentional design decision(?)
+
+### Recovery from downtime/db failure
+
+Extra care must be taken to ensure that implementations maintain
+knowledge of their forward_extremity. Currently it doesn't matter if a
+server sends events with only partial knowledge of their prior
+history. Implementations may need to query other servers and ensure
+they have complete knowledge of their own forward extremity.
```
Gnuxie, line 133: I believe that if servers persist events before sending them, then the risk is limited to db recovery. While querying other servers for the extremity may be necessary in that scenario, equally the server key could be rotated within the context of https://github.com/matrix-org/matrix-spec-proposals/pull/4345 instead. Which removes the risk entirely.

proposals/4353-per-origin-linear-chain.md
```diff
@@ -0,0 +1,164 @@
...
+
+### An example of erasure with linear origin chain
+
+![](./images/4353-linear-origin-chain-equivocator.png)
+
+If Blue does attempt to retroactively erase Yellow, it must explicitly
+create an event that is concurrent with its own origin chain. And thus
+it is observable to everyone that Blue is an equivocator/byzantine
+traitor. As Blue's own events will have multiple causal successors
+sent by Blue.
```
Gnuxie, line 101: It's still possible to erase history, it's just that doing so without creating a concurrent event requires you to explicitly acknowledge yellow's history as a part of the ban. And if bans are mandated to be scoped to a target's origin predecessor, it is then actually impossible to abuse to erase history. Since a condition for authorising the ban would be not to contradict your own causal frontier in auth rules.

proposals/4353-per-origin-linear-chain.md
Gnuxie: Unfortunatly it is entirely possible through a naive db restore that a server sends authorization events that conflict with its prior history that is now unknown to itself. So even if a definition of a traitor is liberalised it's still going to be possible to become one by accident. We don't actually propose that equivicators or inconsistent servers are noted or even acted upon in any way but it's still going to be important that we allow ways out (which key rotation seems to be the best of them).

proposals/4353-per-origin-linear-chain.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Example using today's room model
+
+![](./images/4353-current-matrix-ban-and-key.png)
+
+Consider A scenario where there are three servers in a room: Blue,
+Green, and Yellow. Consider that Blue immediately bans Yellow after
+Yellow joins the room and the ban event directly references Yellow's
+join in `prev_events`. Consider that concurrent to Blue's ban,
+Yellow sends a message. So that Yellow's join event has to known
```
Johennes, line 22:
```suggestion
Yellow sends a message. So that Yellow's join event has two known
```

proposals/4353-per-origin-linear-chain.md
```diff
@@ -0,0 +1,164 @@
...
+This means that every event will reference either one or no
+predecessors from the same origin.
+
+This has some major advantages:
+
+- Intentional forks are explicit and detectable. Servers cannot create
+  events concurrent to their own causal predecessors without
+  explicitly ignoring their maintained forward extremity. The proposal
+  does not prevent this, but it does mean that such an equivocation is
+  detectable and provable. And so the sender's repuatation can become
```
Johennes, line 59:
```suggestion
  detectable and provable. And so the sender's reputation can become
```

PR #4356, opened 2025-09-16
MSCXXXX: Recently used emoji

Johennes:


proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,65 @@
...
+clients.
+
+This proposal introduces a way for clients to maintain a shared storage of recently used emoji to
+enable emoji suggestions across clients.
+
+## Proposal
+
+A new global account data event `m.recent_emoji` is introduced. In `content`, it contains a single
+property `recent_emoji` that is an array where each element is itself an array. The first element in
+this nested array is the emoji, the second element is a counter for how often it was used. The outer
```
clokep: Do counters reset?

Johennes: Hm, they don't currently in any of the implementations. The only use case for that I can think of is avoiding overflowing integer limits on certain platforms?

clokep: You'll want to avoid integer limits of JSON, although that's quite large so not really a "real" concern. Another option is just capping it at that.

tulir: A few paragraphs below it defines only 100 items are kept, which effectively resets counters and caps the size. I don't think it's physically possible to use an emoji 2^53-1 times, so there's probably no need to limit counters themselves

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,65 @@
...
+    "recent_emoji": [
+      [ "😅",  7 ], // Most recently used, 7 times overall
+      [ "👍", 84 ], // Second most recently used, 84 times overall
+      ...
+  }
+}
+```
+
+When an emoji is used in a message or an annotation, the sending client moves (or adds) it to the
+beginning of the `recent_emoji` array and increments (or initializes) its counter.
```
tulir: Can we put custom emojis here? Since reactions just use the raw `mxc://` URI as key (#4027), the same key here would probably make sense

(for reference, gomuks web already abuses element's recent emoji event and puts mxc uris there. Element doesn't support custom emojis in the emoji picker, so they're ignored, but it doesn't otherwise mind their presence)

proposals/4356-recent-emoji.md
Johennes: Ah, right. That's effectively a reset, too. I have added a note with 6e1cf39a015d025a51ae95a987606ba0118a85aa explaining the absence of other reset mechanisms.

proposals/4356-recent-emoji.md
Johennes: Oh, interesting. That sounds reasonable to me, yeah.

Given that both are MSCs, do we just leave this open as a comment for now or do we need a note in the proposal text?

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,69 @@
...
+Some emoji picker libraries support this feature by locally tracking emoji usage. This doesn't work
+well in a multi-device environment, however, because such history cannot easily be shared between
+clients.
+
+This proposal introduces a way for clients to maintain a shared storage of recently used emoji to
+enable emoji suggestions across clients.
+
+## Proposal
+
+A new global account data event `m.recent_emoji` is introduced. In `content`, it contains a single
```
bmarty, line 15: I know that the plural form of `emoji` can be `emoji` but to avoid confusion, maybe using `m.recent_emojis` help to clarify that we have a list of emojis?

proposals/4356-recent-emoji.md
bmarty: > so not really a "real" concern

Any client could set Integer.MAX to the counter and let other clients have trouble when incrementing it, no?

clokep: I think just acknowledging that it should be capped according to the JSON rules of Matrix is fine.

proposals/4356-recent-emoji.md
tulir: Hmm, it would probably be better to be in the proposal so clients are aware they need to at least tolerate the existence of mxc uris in the list. Inline images are also already in the spec. No need to introduce a dependency on 4027 though, just something along the lines of

> When an image is sent as an inline image or in a reaction (using MSC4027), the `mxc://` URI of the image MAY be used as the "emoji" in this event. Clients which do not support such use of images MUST tolerate the existence of `mxc://` entries, e.g. by ignoring the entries when deciding what to display to the user, while still preserving them when modifying the list.

proposals/4356-recent-emoji.md
tulir: Technically the JSON rules of Matrix don't apply as this is an account data event, but probably fine to define that as the limit anyway if such a definition is wanted.

> Any client could set Integer.MAX to the counter and let other clients have trouble when incrementing it, no?

Clients can do lots of nasty things to cause trouble for the user. For example, other clients will probably have trouble incrementing it if some evil client sets the value to an object instead of an integer. It won't affect any other users, so it's not really a concern.

proposals/4356-recent-emoji.md
Johennes: Makes sense. Have added this with 1ff30d4b24cc1cac8a4f2ee1d0c5fdc32ecf6c46.

proposals/4356-recent-emoji.md
Johennes: I have made 2^53-1 the explicit upper limit now with 58d45ad44a7f8e6f4d1e63b7b03b314014ce5b56.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,74 @@
+# MSC4356: Recently used emoji
```
anoadragon453: I think it should be a duct instead. So clients can quickly get the count instead of needing to iterate over the whole thing. You can't and shouldn't trust the order anyways

Also we should include the last-used timestamp so you can tie-break better.

This also allows us to prune older entries, if a client desires.

Clients should be aware of the counter being incremented to just below int_max by another client. Maybe a client should be normalise events in that case? Eh, probably not needed. Just don't overflow.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,74 @@
...
+leading to significantly different UX across clients.
+
+## Alternatives
+
+Further metadata such as the concrete access time or the room could be tracked together with emoji.
+It is unclear, however, if this would lead to materially better suggestions, however.
+
+## Security considerations
+
+This proposal doesn't mandate encrypting the `m.recent_emoji` account data event. Since emoji are
```
anoadragon453: Why not encrypt this? Is it possible to put this in SSSS easily or no?

proposals/4356-recent-emoji.md
anoadragon453: Client developers should be able to determine the type of a field from the spec, rather than relying on the field name.

As a native English speaker, `recent_emoji` sounds best to me.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Potential issues
+
+Clients could choose wildly different ways to generate recommendations from the shared storage
+leading to significantly different UX across clients.
+
+## Alternatives
+
+Further metadata such as the concrete access time or the room could be tracked together with emoji.
+It is unclear, however, if this would lead to materially better suggestions, however.
```
anoadragon453: One alternative would be to structure `m.recent_emoji` as the following:

```json5
"recent_emoji": {
  "😅": {
    "last_used": 10000000,
    "count": 7
  },
  "👍": {
    "last_used": 20000000,
    "count": 84
  }
}
```

where e.g. `10000000` is a unix timestamp of the last time the emoji was used.

* A dict would allow a client to access the attributes of a given emoji more efficiently than an array.
* Documenting when the emoji was last used can be helpful. If it wasn't used for a long time, a client could drop the emoji from the list; even if it had been used many, many times in the past.

To save bandwidth, this could instead be represented as:

```
"recent_emoji": {
  "😅": [10000000, 7],
  "👍": [20000000, 84]
}
```

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Security considerations
+
+This proposal doesn't mandate encrypting the `m.recent_emoji` account data event. Since emoji are
+most commonly used in annotations which are not encrypted, servers could already track and abuse
+this information today, however.
+
```
anoadragon453, line 86: Could you add a couple sentences clarifying that clients SHOULD properly handle the case where they need to add 1 to a count where a count is already the maximum integer allowed by canonical JSON.

Given enough time, every emojis count could naturally reach this state. In that case, it would make sense to "normalise" all the numbers (divide all counts by 2, don't go below 1).

In addition, clients SHOULD "fix" any negative counts by simply removing them from the list.

**Edit:** I read [the discussion above](https://github.com/matrix-org/matrix-spec-proposals/pull/4356#discussion_r2352369255) after submitting this comment. But feel free to still adopt any of these suggestions to make to clients if they happen to come across data from a malicious (or more likely buggy) client :slightly_smiling_face:

proposals/4356-recent-emoji.md
anoadragon453:
```suggestion
It is unclear, however, if this would lead to materially better suggestions.
```

anoadragon453 (changes requested):
Thanks for writing this up! 😁

proposals/4356-recent-emoji.md
Johennes: Have tried to capture it in 594df055385ad132fd4d3305fda828ac14bca00d.

proposals/4356-recent-emoji.md
Johennes: Are you suggesting to add this to the (discarded) alternatives or to use this instead of the current scheme? At the moment, this proposal simply captures what some of Element's clients had implemented previously.

I personally like using objects for the emoji properties, as in your variant, because that makes the semantics of the different values more explicit. I suspect that with request compression, it should have a negligible impact on bandwidth.

For the outer structure, an array, as in the current proposal text, would have the benefit that it automatically stays ordered by last usage time. Then again, the set of tracked emoji will probably be relatively small. So this optimization might be superfluous.

Johennes: To illustrate what this change would mean for the current implementations: Element Web currently generates suggestions by taking the first 24 array elements (meaning the 24 most recently used emoji), then stably sorts that list by usage count.

With the alternative suggested above, this logic would require a sort of the entire set by `last_used`, truncating the result at a length of 24, then stably sorting the remainder by `count`.

Also for illustration: My own `io.element.recent_emoji` event currently contains 52 elements only (Element Web allows for up to 100). If this is representative, any performance considerations might be moot.

anoadragon453: I originally intended this as a schema to replace the one in the proposal, but I'm less convinced of that now.

Part of the reason for tracking `last_used` time was to be able to cull entries which hadn't been used in a very long time. But thinking about it, I don't think there's really a need to do that, especially if clients already limit the overall length of the list.

Let's leave it as is. If you like, you could add a small section to the alternatives saying "we could track `last_used` to cull emoji that haven't been used in a very long time, but this was not considered necessary given the list already has a maximum length".

anoadragon453 (approved):
I think this is practically ready to go.

anoadragon453:
@mscbot fcp merge

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot concern checklist incomplete

(any SCT member can resolve this)

proposals/4356-recent-emoji.md
Johennes: Have added this in 7855bdd51b4a8a3c890557060427c63b5e9c0a46.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,84 @@
...
+{
+  "type": "m.recent_emoji",
+  "content": {
+    "recent_emoji": [
+      [ "😅",  7 ], // Most recently used, 7 times overall
+      [ "👍", 84 ], // Second most recently used, 84 times overall
+      ...
+  }
+}
+```
```
Johennes, line 41: This proposal was written by reverse-engineering existing implementations in the Element clients which use this format. One glaring question here is if the nested-array storage is actually a good idea or not?

This scheme is very compact but given request compression and the typical amount of emoji, any performance benefits resulting from it are probably negligible.

Furthermore, the nested array doesn't lend itself well to future extension. We could just append more array elements and assign special semantics to certain array indexes. This feels somewhat brittle and unnecessarily implicit, however.

I cannot currently think of a possible future extension of the tracked data. So maybe this concern is moot to begin with. I'd still be interested to hear other thoughts on this though.

KitsuneRal: I agree; as for possible extensions, I can imagine adding dates, or required custom fonts/emoji packs, so the inner array certainly better be a dictionary. As for the outer one, I first thought of a dictionary keyed with the emoji but then custom fonts may encode different emojis with the same code point; so probably an array is appropriate here. Another benefit of the array on the outer level is that a naïve client implementation can just take it as is while a dictionary would have to then be sorted by date or [frecency](https://en.wikipedia.org/wiki/Frecency) or whatever.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,84 @@
...
+large for all practical purposes.
+
+Clients MAY freely customise the logic for generating recommendations from the stored emoji. As an
+example, they could select the 24 first (= most recently used) emoji and stably sort them by their
+counters (so that more recently used emoji are ordered first on ties).
+
+## Potential issues
+
+Clients could choose wildly different ways to generate recommendations from the shared storage
+leading to significantly different UX across clients.
```
KitsuneRal: Nitpick: another sentence could be added here on why the author believes it's not worth it to specify client behaviour further (my take is that this is an area where UX exploration should still take place as we don't know the "best" way to sort recent emojis).

proposals/4356-recent-emoji.md
anoadragon453: Tracking custom emoji (MSC2545) is a good shout. Having an emoji act as the keys to a dictionary doesn't quite work in that case.

If we wanted to keep the nested array structure, one could say:

* If the first array element is a string, then assume that this is a unicode emoji :slightly_smiling_face: 
* If the first array element is an object, then assume a non-unicode emoji and the following structure:

    ```json
    {
        "image_url": "mxc://example.org/blah",
        "image_pack_room": "!mDgU8OWQ6CAgEK-Fb2Dhv0fR0nHs_tEM_Na93SSIhu0",
        "image_pack_event": "$OMn1hZchIFSrWDNgBQj6MQkojvnGPR6tL7su8IkHW_Y"
    }
    ```

Example:

```json
{
  "type": "m.recent_emoji",
  "content": {
    "recent_emoji": [
      [ "😅",  7 ],
      [
        {
          "image_url": "mxc://example.org/blah",
          "image_pack_room": "!mDgU8OWQ6CAgEK-Fb2Dhv0fR0nHs_tEM_Na93SSIhu0",
          "image_pack_event": "$OMn1hZchIFSrWDNgBQj6MQkojvnGPR6tL7su8IkHW_Y"
        },
        84
      ]
    ]
  }
}
```

or similar. With a dictionary inside an array:

```json
{
  "type": "m.recent_emoji",
  "content": {
    "recent_emoji": [
      {
        "emoji": "😅",
        "total": 7
      },
      {
        "image_url": "mxc://example.org/blah",
        "image_pack_room": "!mDgU8OWQ6CAgEK-Fb2Dhv0fR0nHs_tEM_Na93SSIhu0",
        "image_pack_event": "$OMn1hZchIFSrWDNgBQj6MQkojvnGPR6tL7su8IkHW_Y",
        "total": 84
      }
    ]
  }
}
```

---

I took a look at what the byte count difference after gzip has been applied to the two above data structures. Turns out it's exactly the same!

<details>

<summary>Click to view byte sizes</summary>

```
➜ cat <<'EOF' | gzip -cn9 | wc -c
➜ {
  "type": "m.recent_emoji",
  "content": {
    "recent_emoji": [
      {
        "emoji": "😅",
        "total": 7
      },
      {
        "image_url": "mxc://example.org/blah",
        "image_pack_room": "!mDgU8OWQ6CAgEK-Fb2Dhv0fR0nHs_tEM_Na93SSIhu0",
        "image_pack_event": "$OMn1hZchIFSrWDNgBQj6MQkojvnGPR6tL7su8IkHW_Y",
        "total": 84
      }
    ]
  }
}
➜ EOF
257

➜ cat <<'EOF' | gzip -cn9 | wc -c
➜ {
  "type": "m.recent_emoji",
  "content": {
    "recent_emoji": [
      [ "😅",  7 ],
      [
        {
          "image_url": "mxc://example.org/blah",
          "image_pack_room": "!mDgU8OWQ6CAgEK-Fb2Dhv0fR0nHs_tEM_Na93SSIhu0",
          "image_pack_event": "$OMn1hZchIFSrWDNgBQj6MQkojvnGPR6tL7su8IkHW_Y"
        },
        84
      ]
    ]
  }
}
➜ EOF
257
```

</details>

Given the negligible difference in size, I'm in favour of the easier to work with dictionary-inside-an-array approach.

(None of that needs to go in the current proposal, of course. What's important is that it can be extended.)

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,84 @@
...
+    "recent_emoji": [
+      [ "😅",  7 ], // Most recently used, 7 times overall
+      [ "👍", 84 ], // Second most recently used, 84 times overall
+      ...
+  }
+}
+```
+
+When an emoji is used in a message or an annotation, the sending client moves (or adds) it to the
+beginning of the `recent_emoji` array and increments (or initializes) its counter.
```
turt2live: This should be an implementation detail rather than a requirement of a client. Not all clients will track an emoji in a message as "used".

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,84 @@
...
+
+When an image is sent as an inline image or in a reaction (using [MSC4027]), the `mxc://` URI of the
+image MAY be used as the "emoji" in this event. Clients which do not support such use of images MUST
+tolerate the existence of `mxc://` entries, e.g. by ignoring the entries when deciding what to
+display to the user, while still preserving them when modifying the list.
+
+As new emoji are being used, clients SHOULD limit the length of the `recent_emoji` array by dropping
+elements from the end. A RECOMMENDED maximum length is 100 emoji. Apart from this, no other
```
turt2live, line 53: `RECOMMENDED` is not a recognized keyword: does this mean to say MAY, SHOULD, or "*strongly recommended*"?

turt2live:
@mscbot resolve checklist incomplete

proposals/4356-recent-emoji.md
tulir: Note that custom emojis are already supported in this MSC. The solution is as trivial as MSC4027 for reactions: just use the `mxc://` URI as the key instead of an unicode emoji. gomuks web already has that implemented in the `io.element.recent_emoji` event (and element doesn't mind the funny entries)

proposals/4356-recent-emoji.md
KitsuneRal: It is? Both RFC 2119 and [our own text](https://spec.matrix.org/v1.16/#requirement-levels) recognise it.

proposals/4356-recent-emoji.md
KitsuneRal: What I had in mind, though, is not having custom emojis stored as images but as a part of an emoji font. In that case, the same code point in different fonts could look different; you also have to specify which font to use. Anyway, as already mentioned above, all this is just meaning to support turning the inner array to a dictionary.

anoadragon453: > Note that custom emojis are already supported in this MSC. The solution is as trivial as [MSC4027](https://github.com/matrix-org/matrix-doc/pull/4027) for reactions: just use the `mxc://` URI as the key instead of an unicode emoji. gomuks web already has that implemented in the `io.element.recent_emoji` event (and element doesn't mind the funny entries)

Good to know, and the existing support is noted. 

The solution I presented above would allow you to link back to the image pack containing the custom emoji however, which could be valuable for a client to display.

proposals/4356-recent-emoji.md
turt2live: I don't recall our text recognizing it :( RFC 2119 does, but my understanding was that we used a smaller subset for clarity.

turt2live: I guess that understanding isn't written down though, so resolving this thread.

proposals/4356-recent-emoji.md
Johennes: Thanks everyone for the input. I have switched to an array of objects with 82a600940008d4353ed0b1a3052659575990c475. Since the previous layout is already in use in various clients, I have also changed the prefix to avoid any conflicts.

I'm unsure if the change of schema invalidates the implementation requirement. I'd say the change is minor and only structural so the old implementations should still suffice?

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,101 @@
...
+to negative values or the maximum allowed value. To prevent this, clients SHOULD drop any emojis
+with a count below 0 from the list. When observing a count of 2^53-1 for an emoji, clients SHOULD
+normalise the entire set of emoji by dividing all counts by two and rounding up. To prevent race
+conditions, both of these changes SHOULD only be applied when the client is updating the account
+data event due to local emoji usage.
+
+## Unstable prefix
+
+While this MSC is not considered stable, `m.recent_emoji` should be referred to as
+`io.github.Johennes.msc4356.recent_emoji`.
```
Johennes: This feels a bit odd to me and I've never seen it elsewhere. I understand from the discussion on https://github.com/matrix-org/matrix-spec/pull/2222 that this is the desired format, however, for people not directly affiliated with the Foundation.

proposals/4356-recent-emoji.md
Johennes: Good call. Have extended the text with cb46b7197883cb6cd88df5ba92cb655d81f130c2.

proposals/4356-recent-emoji.md
Johennes: Have relaxed things with 403d4d01ccd7b9083b8b42ebd46f40bbde6f4383.

proposals/4356-recent-emoji.md
Johennes: Given lack of further concerns, I'll assume the current wording is acceptable.

proposals/4356-recent-emoji.md
tulir: GitHub usernames are case insensitive, but event types are case sensitive, so they should probably always be lowercase. Otherwise the prefix is fine

```suggestion
While this MSC is not considered stable, `m.recent_emoji` should be referred to as
`io.github.johennes.msc4356.recent_emoji`.
```

Johennes: TIL they're case insensitive. 🤦‍♂️

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,103 @@
...
+## Security considerations
+
+This proposal doesn't mandate encrypting the `m.recent_emoji` account data event. Since emoji are
+most commonly used in annotations which are not encrypted, servers could already track and abuse
+this information today, however.
+
+Malicious or buggy clients could cause undefined behavior on other clients by setting emoji counters
+to negative values or the maximum allowed value. To prevent this, clients SHOULD drop any emojis
+with a count below 0 from the list. When observing a count of 2^53-1 for an emoji, clients SHOULD
+normalise the entire set of emoji by dividing all counts by two and rounding up. To prevent race
```
dbkr: Given you could send 1 emoji a second for 250 million years and still have headroom left, this feels unnecessary.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Security considerations
+
+This proposal doesn't mandate encrypting the `m.recent_emoji` account data event. Since emoji are
+most commonly used in annotations which are not encrypted, servers could already track and abuse
+this information today, however.
+
+Malicious or buggy clients could cause undefined behavior on other clients by setting emoji counters
+to negative values or the maximum allowed value. To prevent this, clients SHOULD drop any emojis
+with a count below 0 from the list. When observing a count of 2^53-1 for an emoji, clients SHOULD
```
dbkr: You could also just say "invalid entries" which includes any where the usage count is not a number, etc.

proposals/4356-recent-emoji.md
dbkr: Curveball suggestion: we could also just not adopt this part into the spec (the 'total' field) - is it actually used? It doesn't seem massively useful?

tulir: The emoji picker sorts by the total, so it's necessary. Recently used ordering of the array is just there so you can delete the least recently used entries when the list is full (and for tie-breaking if totals are the same)

Johennes: > Given you could send 1 emoji a second for 250 million years and still have headroom left, this feels unnecessary.

I think part of the reasoning in the suggestion that led to this was that a client could set this maliciously or due to a bug and thereby cause undefined behavior on other clients.

proposals/4356-recent-emoji.md
Johennes: Good point.

```suggestion
with invalid entries from the list. When observing a count of 2^53-1 for an emoji, clients SHOULD
```

proposals/4356-recent-emoji.md
clokep: Yes -- that was the thinking.

clokep: That being said -- I don't know if we need to explicitly define client behavior for that situation, but it seems like the spec should say _something_ about it.

Johennes: Maybe we could just highlight the overflow problem and add the "divide by two" logic as an example or recommendation only?

dbkr: Sure, the validation is absolutely necessary, I was just talking about the divide-by-two algorithm. Feels like clients could just fix it appropriately if they felt the need to. At most you could say it might represent relative usage frequency rather than a literal number of times used, but, again… 250 million years: if any of the numbers reach this, the data is already broken.

uhoreg: It feels like just capping the value at 2^53-1 would be sufficient.  It seems like a very low-value target and unlikely to happen, so it probably isn't worth adding too much complication.

Johennes: Alright, I have removed the divide-by-two logic in c26eebe514f79dabd362e20122f9a5619040c3ec and only kept the `<2^53` part.

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+When an emoji is used, the sending client moves (or adds) it to the beginning of the `recent_emoji`
+array and increments (or initializes) its counter. What exactly constitutes emoji usage is left as
+an implementation detail for clients. Obvious choices are sending emoji in messages or annotations.
+
+When an image is sent as an inline image or in a reaction (using [MSC4027]), the `mxc://` URI of the
+image MAY be used as the "emoji" in this event. Clients which do not support such use of images MUST
+tolerate the existence of `mxc://` entries, e.g. by ignoring the entries when deciding what to
+display to the user, while still preserving them when modifying the list.
```
richvdh: I guess the question arises of what should happen to values of `emoji` that are neither an emoji nor a (valid) mxc URI.

my inclination is to say that clients should ignore `emoji` that they don't understand in general.

Johennes: Yes, sounds good. Does this capture it?

```suggestion
When an image is sent as an inline image or in a reaction (using [MSC4027]), the `mxc://` URI of the
image MAY be used as the "emoji" in this event. Clients MUST tolerate any `emoji` values that they
don't understand or support, by ignoring the entries when deciding what to display to the user,
while still preserving them when modifying the list.
```

KitsuneRal:
@mscbot concern a couple threads still open

proposals/4356-recent-emoji.md
Johennes: Given nobody expressed further concerns, I assume this is alright.

anoadragon453:
@mscbot resolve a couple threads still open

proposals/4356-recent-emoji.md
```diff
@@ -0,0 +1,100 @@
...
+leading to significantly different UX across clients. A "correct" algorithm is not currently known,
+however, and will require future UI exploration in client implementations.
+
+## Alternatives
+
+Further metadata such as the concrete access time or the room could be tracked together with emoji.
+It is unclear, however, if this would lead to materially better suggestions. A last-used timestamp
+could also be used to cull emoji that haven't been used in a very long time. Given that implementations
+are already encouraged to limit the maximum number of tracked emoji, this doesn't appear necessary,
+however.
```
TheArcaneBrony, line 73: Some other platforms (ie. Discord) use some kind of weighting system between both count and a limited list of recent use timestamps (not withstanding the weird behavior across devices there due to only synchronising it with the server every couple of hours...)
They also store it as a Map<string Emoji, object FrequencyData>.

This has the benefit of showing emojis that you use more often in a higher position rather than strictly ordering them by last used. It's actually a pretty nice system :)

tulir: This MSC already allows ordering emojis by use count rather than time. That's how Element works too which is what this MSC is based on (iirc element uses count as primary order and last used as tiebreaker)

Johennes:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2291

PR #3489, opened 2021-11-14
MSC3489: Sharing streams of location data with history

ara4n:
In which we introduce `m.beacon_info.*` and `m.beacon.*` events for streaming location data, building on [MSC3488](https://github.com/matrix-org/matrix-doc/pull/3488).

[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/location-streaming/proposals/3489-location-streaming.md)

**Matrix React SDK implementation**

https://github.com/matrix-org/matrix-react-sdk/tree/develop/src/components/views/location

**Element Android implementation**

https://github.com/vector-im/element-android/tree/develop/vector/src/main/java/im/vector/app/features/location

**Matrix iOS SDK implementation**

https://github.com/matrix-org/matrix-ios-sdk/tree/develop/MatrixSDK/LocationSharing

Preview: https://pr3489--matrix-org-previews.netlify.app

PR #4357, opened 2025-09-16
MSC4357: Live Messages via Event Replacement

DonPrus:
**Summary**

This MSC proposes "Live Messages": recipients see a message evolve in real time while it is being typed.
It leverages existing **m.replace** edits on **m.room.message** (no new event type), marked by an unstable
flag in content to denote a "live session". Non-supporting clients gracefully degrade to a normal edited
message; supporting clients render a smooth, streaming-like UX.

**What this adds**

- Identification: initial message carries `org.matrix.mscXXXX.live: {}`; subsequent updates use `m.replace`
  with full `m.new_content`; the final update omits the live flag (session complete).
- Control: optional state event `m.room.live_messaging` (`enabled: boolean`) to allow room-level opt-in/out.
- Server: no new endpoints; existing relations/aggregation apply; admins may rate-limit high-frequency edits.

**Backwards compatibility**

Old clients see "a message with edits" and always end up with the final content.

**Rendered Markdown**
[Rendered: proposals/4357-live-messages.md](https://github.com/matrix-org/matrix-spec-proposals/blob/2518c94e1aa37dcf4a916c521792fe53cb54c06e/proposals/4357-live-messages.md)

**Author**

Author: @donprus

### DIFF
--- a/proposals/4357-live-messages.md
+++ b/proposals/4357-live-messages.md
@@ -1,0 +1,123 @@
+# MSC4357: Live Messages via Event Replacement
+
+## Introduction
+
+This proposal introduces **Live Messages** for Matrix: recipients can see a message evolve in near real time while it is
+being typed. The user experience is similar to streaming answers from LLMs (token-by-token) and to Simplex.chat's
+mutable "chat items", but without introducing a new event type. Instead, we rely on Matrix's existing edit mechanism
+(`m.replace` on `m.room.message`). Non-supporting clients degrade gracefully: they see a normal message which was edited
+a few times and always end up with the final content.
+
+Motivation includes more natural conversation flow, better perceived responsiveness, and resilience in unstable
+environments (e.g., live reporting): every chunk that was sent remains stored, so the last known state is preserved even
+if the sender disconnects prematurely.
+
+## Proposal
+
+A Live Message is a session of high-frequency edits applied to a single persistent `m.room.message`. The session starts
+when the user enables "live mode" and ends when the user finalizes the message. We introduce an **unstable marker** in
+event content to identify a live session and specify client/server behavior that reuses existing primitives.
+
+### Event identification
+
+- **Initial event**: a normal `m.room.message` whose `content` **MUST** include the unstable key
+  `"org.matrix.msc4357.live": {}` (empty JSON object). This marks the start of a live message session.
+- **Update events**: each update **MUST** be a `m.room.message` with `m.relates_to.rel_type = "m.replace"` and
+  `m.relates_to.event_id` referencing the initial event; its `m.new_content` **MUST** contain the *full* updated body
+  (and formatted body if applicable).
+- **Final update**: when the user completes the message, the last update **SHOULD NOT** include the live marker. The
+  absence of the live marker in the aggregated content signals session completion.
+
+### Client behavior (sending)
+
+- **Activation**: when the user toggles live mode, clients provide a clear UI affordance (e.g., lightning icon).
+- **First send**: upon a natural boundary (e.g., finishing a word) or a short delay, send the initial message with the
+  live marker and retain its `event_id` for subsequent `m.replace`.
+- **Periodic updates**: send updates every ~2–3 seconds and/or on natural boundaries. Avoid per-keystroke updates. Each
+  update replaces the entire content (`m.new_content` reflects the complete current text).
+- **Completion**: on explicit send or mode exit, send a final update without the live marker. After this, do not send
+  further updates for this session.
+- **Rate limiting**: clients **MUST** avoid flooding and respect server guidance; batching is encouraged.
+
+### Client behavior (receiving)
+
+- When the initial event contains the live marker, supporting clients **SHOULD** render it in a "live" state (e.g.,
+  subtle animation / icon). Updates targeting the same message via `m.replace` **SHOULD** update the displayed text in
+  place, without showing a separate "(edited)" marker or spamming notifications. When a final update arrives without the
+  live marker, the client **SHOULD** transition the message to a normal finalized state.
+- Non-supporting clients treat the flow as "message followed by edits", ending with the correct final content.
+
+### Server behavior
+
+- No new endpoints or types are required. Homeservers store and federate the initial message and its `m.replace` updates
+  as usual. Existing aggregation behavior applies. Homeservers **MAY** apply rate limits to high-frequency edits and/or
+  offer retention policies for rooms with heavy live usage.
+
+## Room-level control: `m.room.live_messaging` (optional)
+
+To allow administrators and room moderators to control the feature, we introduce a room state event:
+
+- **Type**: `m.room.live_messaging` (unstable: `org.matrix.msc4357.live_messaging` until accepted)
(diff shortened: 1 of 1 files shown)


proposals/4357-live-messages.md
```diff
@@ -0,0 +1,123 @@
...
+
+- **Spam / flooding**: A malicious client could emit updates too frequently. Mitigations include server rate-limits,
+  client batching, and room-level disabling via `m.room.live_messaging`. This is not a new class of risk compared to
+  rapid normal edits; existing controls apply.
+- **Privacy**: Live updates reveal intermediate drafts. This is a user-experience choice; clients should clearly label
+  the mode and allow users to opt out. The content remains subject to the same E2EE/federation properties as standard
+  messages.
+- **Storage**: Live sessions produce more events. Deployments can rely on retention policies and rate-limits.
+
+## Alternatives
```
Johennes, line 90: Conceptually this proposal sounds similar to https://github.com/matrix-org/matrix-spec-proposals/pull/3489 (live location sharing). That one uses a state event and reference relations though. Replacements strike me as a more intuitive solution though.

proposals/4357-live-messages.md
```diff
@@ -0,0 +1,123 @@
...
+- **Content**:
+  ``` json
+  {
+    "enabled": true
+  }
+  ```
+
+### Semantics:
+
+- If absent, default is enabled (unless server policy states otherwise).
```
Johennes, line 71: IIUC the enforcement can only happen on the client side. This means clients would need a way to discover such global server policy?

proposals/4357-live-messages.md
```diff
@@ -0,0 +1,123 @@
...
+- If `enabled: false`, clients **MUST NOT** offer live mode in this room.
+- Servers **MAY** enforce rejection of events that carry live markers in rooms where it is disabled.
```
Johennes, line 73: This is only possible in unencrypted rooms because `org.matrix.msc4357.live` is in `content` which is encrypted.

proposals/4357-live-messages.md
```diff
@@ -0,0 +1,123 @@
...
+### Client behavior (sending)
+
+- **Activation**: when the user toggles live mode, clients provide a clear UI affordance (e.g., lightning icon).
+- **First send**: upon a natural boundary (e.g., finishing a word) or a short delay, send the initial message with the
+  live marker and retain its `event_id` for subsequent `m.replace`.
+- **Periodic updates**: send updates every ~2–3 seconds and/or on natural boundaries. Avoid per-keystroke updates. Each
+  update replaces the entire content (`m.new_content` reflects the complete current text).
+- **Completion**: on explicit send or mode exit, send a final update without the live marker. After this, do not send
+  further updates for this session.
+- **Rate limiting**: clients **MUST** avoid flooding and respect server guidance; batching is encouraged.
```
Johennes, line 40: Does batching mean combining several updates into a single event replacement?

proposals/4357-live-messages.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Client behavior (sending)
+
+- **Activation**: when the user toggles live mode, clients provide a clear UI affordance (e.g., lightning icon).
+- **First send**: upon a natural boundary (e.g., finishing a word) or a short delay, send the initial message with the
+  live marker and retain its `event_id` for subsequent `m.replace`.
+- **Periodic updates**: send updates every ~2–3 seconds and/or on natural boundaries. Avoid per-keystroke updates. Each
+  update replaces the entire content (`m.new_content` reflects the complete current text).
+- **Completion**: on explicit send or mode exit, send a final update without the live marker. After this, do not send
+  further updates for this session.
```
Johennes, line 39: Is it allowed to transition back from completion into live mode?

proposals/4357-live-messages.md
turt2live, line 1: Implementation requirements:

* [ ] Client (sending)
* [ ] Client (receiving)

proposals/4357-live-messages.md
DonPrus: I agree the concept is similar to MSC3489 in that both stream incremental state. For live text, I believe `m.replace` on `m.room.message` is the most intuitive and compatibility‑friendly primitive: older clients already understand “edits” and will converge to the final content; supporting clients can render the intermediate states seamlessly. It also avoids adding a new event type purely for live typing and reuses the well‑understood edit aggregation path (per MSC2676 / spec).

proposals/4357-live-messages.md
DonPrus: In the draft I wrote “default is enabled (unless server policy states otherwise)”, which is ambiguous without a standard way to advertise that policy. I propose to clarify that the normative control here is room‑level via m.room.live_messaging; clients MUST honor it. Servers MAY reject live‑marked events in unencrypted rooms as a local policy, but there is no standardized discovery for a global ban, so it’s out of scope for this MSC. I’ll remove/adjust the “unless server policy states otherwise” wording to avoid implying a discovery mechanism that doesn’t exist

proposals/4357-live-messages.md
DonPrus: yep, feature still works end‑to‑end (supporting clients decrypt, detect the marker, and render progressively), but enforcement and special aggregation are client‑side only in E2EE. I’ll add an explicit “E2EE considerations” subsection stating that server‑side enforcement is only feasible in unencrypted rooms.

proposals/4357-live-messages.md
DonPrus: By batching I mean coalescing local interim changes and emitting a single `m.replace` update at a bounded cadence (e.g. every ~2–3s or at natural boundaries like end‑of‑word), rather than sending an event per keystroke. Each update still contains the full current content in `m.new_content`; there’s no special “multi‑update container”. I’ll tighten the wording to make this explicit.

proposals/4357-live-messages.md
DonPrus: I intend completion to be terminal for that message. Once the final update (without the live marker) is sent, the live session is over. Further changes are normal edits; resuming live mode should be done by starting a new live message. Allowing a message to re‑enter live mode risks confusing clients (especially non‑supporting ones) that have already treated it as finalized. I’ll add a normative MUST‑NOT to make this explicit.

kathleenmotley49-maker (approved):

Saiv46:
Shouldn't we use Sticky Events from MSC4354? It seems a perfect fit for live messages. Then on completion just send ordinary message without the edits.

Otherwise users would hit into existing rate limits by constantly updating a single message.

toger5:
@Saiv46 Sticky events solve a different problem:
 - On sync I want to guarantee users download a specific event.
 
 This MSC seems to be targeted so that If I am already looking at an event i can see updates.
 Which does not need to be "force synced" if I dont look at the tile containing this "update stack" I am not interested in those events, so they can be fetched via a pull approach.
 
Sticky events are useful if you want a timeline item to (for at most 1h) have an impact on the room list **even for gappy syncs** where you are not yet actively viewing a room.
Like for call participations.

PR #4358, opened 2025-09-17
MSC0000: Out of room server discovery

Gnuxie:


proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
...
+## Alternatives
+
+### OPRF key in the create event
+
+An OPRF[^oprf] key could be placed into the create event which is used
+instead via `OPRF(room_key, server_name)`. However, it's unclear
+if there are any situations where the `room_id` is known and not
+the create event.
+
+[^oprf]: https://datatracker.ietf.org/doc/rfc9497/
```
Gnuxie, line 79: Technically people sometimes share a link to a private room in a public room to share context with some people who are in both rooms (but not all people are in both of them)

proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
...
+
```
Gnuxie, line 80: You should mention explicitly that the information used in the tag has to be intrinsic to the room. So that the room creators can't erase it or misrepresent it in any possible branch of the room.

turt2live:
It appears this MSC is still a work in progress (or at least, doesn't explain how/when to use the new functionality), so I'm converting it to a draft instead of listing implementation requirements.

proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
...
+### Useless if a victim server hasn't built up knowledge of common peers through other rooms
+
+A server will not know which servers to query if it hasn't interacted
+with any other server before through membership in other rooms.
+
+### Not clear when it is appropriate to query common rooms
+
+How often should rooms check for unexpected results?
+
+When joining a new room is a good start. But how frequently otherwise?
```
Gnuxie, line 68: If epochs were going to be considered, then would be a good time too. It's not clear which servers you would select to try pry information...

proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Response body
+
+The responding server simply responds with the token tags for
+which it has a match stored:
+
+- The `recognised_tags` are any of the queried tags which the
+  requested server acknowledges. This property is an array of strings.
+
+## Potential issues
```
Gnuxie, line 51: The protocol doesn't scale and requires more data for more rooms. So a passive scan doesn't really work for a large homeserver.

Gnuxie: The way to make a passive mechanism that does scale is to incrementally collect tags from other servers and store them. And possibly even allow them to be passed on, but this does mean just storing random noise in most cases.

proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
...
+### OPRF key in the create event
+
+An OPRF[^oprf] key could be placed into the create event which is used
+instead via `OPRF(room_key, server_name)`. However, it's unclear
+if there are any situations where the `room_id` is known and not
+the create event.
+
+[^oprf]: https://datatracker.ietf.org/doc/rfc9497/
+
+## Security considerations
```
CobaltCause, line 81: This proposal would allow HS1 to check whether HS2 is participating in a room, so long as HS1 knows HS2's server name and the room ID they want to check for. (The only benefit `hash(server_name, room_id)` brings is that HS1 needs to know the room IDs they want to check for ahead of time.) Thus, this proposal can leak an HS's participation in a room to HSs not participating in said room, which is not great.

proposals/4358-out-of-room-server-discovery.md
```diff
@@ -0,0 +1,91 @@
+# MSC4358: Out-of-room server discovery
```
Gnuxie, line 1: To be blunt: The MSC needs a gossip protocol. And no doubt other MSCs need one too.

PR #4360, opened 2025-09-17
Sliding Sync threads extension

devonh:
Depends-on: #4186

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/devon/ssext_threads/proposals/xxxx-sliding-sync-extension-threads.md)

### DIFF
--- a/proposals/xxxx-sliding-sync-extension-threads.md
+++ b/proposals/xxxx-sliding-sync-extension-threads.md
@@ -1,0 +1,244 @@
+MSCXXXX: Threads extension to Sliding Sync
+===
+
+## Background and Summary
+
+Threads were introduced in [version 1.4 of the Matrix Specification](https://spec.matrix.org/v1.13/changelog/v1.4/) as a way to isolate conversations in a room, making it easier for users to track specific conversations that they care about (and ignore those that they do not).
+
+Sliding Sync is proposed in [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md) as a paginated replacement to `/_matrix/client/v3/sync` with smaller response bodies and lower latency.
+
+It is currently a hassle, or nearly impossible, to be able to determine which threads have been updated in a user's
+joined rooms when a client returns after having been offline for a period of time. The full set of events in the
+timeline has to be paginated through to attempt to discover any threaded events that may have been missed. This is both
+cumbersome and slow.
+
+This MSC proposes an 'extension' to Sliding Sync that allows clients to opt-in to receiving real-time updates to threads in the user's joined rooms.
+
+To handle the case in which there have been many thread updates and there are too many to return in
+Sliding Sync, a new companion endpoint is proposed to allow backpaginating thread updates across all of the user's joined rooms on the client's terms.
+
+
+## Proposal
+
+The `threads` Sliding Sync extension alleviates the issue presented above of a client returning after a period of time
+by having the homeserver track which threads have had updates since the last sync and present a list of those threads to
+the client during sync. The presented list of threads containing updates can cover a larger scope of the timeline than
+just the normal sync response would be able to handle, thus allowing the client to quickly gather information on which
+threads may need to be specifically targeted for further inquery, ie. by using the `/relations` endpoint.
+
+In addition to the scenario of a user returning after a period of time, the proposed solution also enables a client to
+monitor the latest events across all threads via the `include_roots` option. If `include_roots` is set to `true` then
+the `threads` extension of Sliding Sync will always include any thread updates, even if the latest event/s of those
+threads are included in the normal sync response. This is to ensure things such as edits to the thread root are captured
+and passed down in a useful way to the client. The thread root events also include a copy of the latest event in that
+thread to make it extremely easy for a client to present a view of threads, whether there are updates, and a preview of
+the latest content in each thread.
+
+### Sliding Sync extension
+
+The Sliding Sync request format is extended to include the `threads` extension as follows:
+
+
+```jsonc
+{
+  // ...
+
+  "extensions": {
+    // ...
+    
+    // Used to opt-in to receiving changes to threads in joined rooms.
+    "threads": {
+      // Whether to enable this extension.
+      "enabled": true,
+
+      // Whether to include thread root events in the extension response.
+      "include_roots": true | false,
+
+      // Optional. Maximum number of thread updates to receive
+      // in the response.
+      // Defaults to 100.
+      // Servers may impose a smaller limit than what is requested here.
+      "limit": 100,
+    }
+  }
+}
+```
+
+The response format is then extended to compensate:
+
+```jsonc
+{
+  // ...
+
+  "extensions": {
+    // ...
+
+    // Returns a limited window of changes to updated threads.
+    // Only the latest changes are returned in this window.
+    // If the client hasn't missed anything, then this whole
+    // section of the response is omitted.
+    // If a client receives the event/s in the normal response section of `/sync` that would result
+    // in a thread being considered updated, then that thread is omitted from the extension response.
+    // This is true unless `include_roots` is set to `true` in which case all thread updates will be
+    // reported in the extension response.
+    // ie. When `include_roots` is `false, only threads with updates that haven't otherwise been
+    // presented to the client via the normal `/sync` response are included in the extension response.
+    "threads": {
+      "updates": {
+        "!roomid:example.org": {
+          "$threadrootid:example.org": {
+            // Only included if the request contains `include_roots: true`.
+            // A `BundledThreadEvent` (as outlined in https://spec.matrix.org/v1.15/client-server-api/#server-side-aggregation-of-mthread-relationships)
+            // is a thread root event which contains the `m.thread` aggregation included under the
+            // `m.relations` property in the `unsigned` field of the event.
+            "thread_root": BundledThreadEvent,
+
+            // A token that can be used to backpaginate other thread updates,
+            // in this thread, that occurred since the last sync but that were not
+            // included in this response.
+            //
+            // The token is to be used with the `/relations` endpoint
+            // as `from`, with `dir`=`b`.
+            //
+            // Optional. Only present in the response if the client missed some events, i.e. there
+            // was at least one other event in the thread, in addition to the latest event.
+            // In other words, the `prev_batch` points to the prior-to-latest event.
+            "prev_batch": "OPAQUE_TOKEN",
+          },
+
+          // ...
+        }
+      },
+
+      // A token that can be used to backpaginate other thread updates
+      // that occurred in any thread, in any room, since the last sync but that were not
+      // included in this response.
+      //
+      // The token is to be used with the new `/threads` endpoint
+      // as `from`, with `dir`=`b`.
+      // The `pos` parameter in the **request** would be used for the `to`
+      // parameter.
+      //
+      // Optional. Only present when some thread updates have been
+      // missed out from the response because there are too many of them.
+      "prev_batch": "OPAQUE_TOKEN"
+    }
+  }
+}
+```
+
+### Companion endpoint for backpaginating thread updates across all rooms
+
+```
+GET /_matrix/client/v1/thread_updates
+```
+
+URL parameters:
+
+- `dir` (string, required): always `b` (backward), to mirror other pagination
+  endpoints. The forward direction is not yet specified to be implemented.
+
+- `from` (string, optional): a token used to continue backpaginating \
+  The token is either acquired from a previous `/thread_updates` response,
+  or the `prev_batch` in a Sliding Sync response. \
+  The token is opaque and has no client-discernible meaning. \
+  If this token is not provided, then backpagination starts from the 'end'.
+
+- `to` (string, optional): a token used to limit the backpagination \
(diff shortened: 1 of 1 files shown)


proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+
+```jsonc
+{
+  "chunk": {
+    "!roomid:example.org": {
+      "$threadrootid:example.org": {
+        // A `BundledThreadEvent` (as outlined in https://spec.matrix.org/v1.15/client-server-api/#server-side-aggregation-of-mthread-relationships)
+        // is a thread root event which contains the `m.thread` aggregation included under the
+        // `m.relations` property in the `unsigned` field of the event.
+        "thread_root": BundledThreadEvent,
```
devonh, line 208: Should this be optional similar to the new sliding sync extension response? Or even removed from this response altogether?

kathleenmotley49-maker (approved):

kathleenmotley49-maker (approved):

proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+URL parameters:
+
+- `dir` (string, required): always `b` (backward), to mirror other pagination
+  endpoints. The forward direction is not yet specified to be implemented.
+
+- `from` (string, optional): a token used to continue backpaginating \
+  The token is either acquired from a previous `/thread_updates` response,
+  or the `prev_batch` in a Sliding Sync response. \
```
MadLittleMods, line 186: Clarify part of the thread extension in the Sliding Sync response

proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+        // in this thread, that occurred since the last sync but that were not
+        // included in this response.
+        //
+        // The token is to be used with the `/relations` endpoint
+        // as `from`, with `dir`=`b`.
+        //
+        // Optional. Only present in the response if the client missed some events, i.e. there
+        // was at least one other event in the thread, in addition to the latest event.
+        // In other words, the `prev_batch` points to the prior-to-latest event.
+        "prev_batch": "OPAQUE_TOKEN",
```
MadLittleMods, line 220: Feels like this should be present no matter what. A server could come online and have some messages in between which we then later want to paginate.

proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+The `threads` Sliding Sync extension adds additional output to the `/sync` response as well as a new `/thread_updates` companion endpoint
+that can be used to paginate thread updates across all of a user's joined rooms. A client's homeserver will track which threads
+have had updates since the last time `/sync` was called and present a list of threads containing updates to
+the client the next time `/sync` is called. 
+The presented list of threads containing updates can cover a larger portion of the timeline than
+just the normal sync response would be able to handle, thus allowing clients to quickly gather information on which
+threads have updates without having to paginate through the entire timeline of events. 
+
+
+### Sliding Sync extension
```
MadLittleMods, line 35: I'm failing to see the value in the Sliding Sync extension.

It seems like someone could just paginate `GET /_matrix/client/v1/thread_updates` directly to get the same value.

proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+      // missed out from the response because there are too many of them.
+      "prev_batch": "OPAQUE_TOKEN"
+    }
+  }
+}
+```
+
+### Companion endpoint for backpaginating thread updates across all rooms
+
+A new `/thread_updates` endpoint is added to allow a client to obtain missing thread updates for a client.
```
MadLittleMods, line 168: What is a thread update?

Does it include edits to messages in the thread? Edits to the thread root (and maybe the latest message in thread) are mentioned in the MSC but it's unclear about messages in the middle. Especially in the case where your client has already paginated the thread, a period of time passes leaving a gap with some edits to the thread messages and you `/sync` or use this endpoint again.

proposals/4360-sliding-sync-extension-threads.md
```diff
@@ -0,0 +1,286 @@
...
+Sliding Sync is proposed in [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md) as a paginated replacement to `/_matrix/client/v3/sync` with smaller response bodies and lower latency.
+
+It is currently a hassle, or nearly impossible, to be able to determine which threads in a user's joined rooms contain
+updates. This is especially true when a client returns after having been offline for a period of time. The full set of
+events in the timeline has to be paginated through to attempt discovery of any thread events that may have been missed.
+This is both cumbersome for the client and slow.
+
+This MSC proposes an 'extension' to Sliding Sync that allows clients to opt-in to receiving real-time updates to threads
+in the user's joined rooms. The new `extension` provides a mechanism for clients to quickly and easily "catch up" to any
+missed thread updates.
```
MadLittleMods, line 17: In a similar regard to ["what is a thread update"](https://github.com/matrix-org/matrix-spec-proposals/pull/4360#discussion_r2429782409), how do we expect people to get thread activity for historical rooms (they were previously joined and are no longer joined)? They should be able to just as easily spider out their interested threads.

This includes the case where they are `newly_left` within the batch.

proposals/4360-sliding-sync-extension-threads.md
devonh: They would need to continually query `/thread_updates` in that case to get the same behaviour of not missing new thread updates, effectively running an additional but separate sync-style loop on `/thread_updates` that is specific to obtaining thread updates.

proposals/4360-sliding-sync-extension-threads.md
devonh: Hmm I'm not sure I see the use case you are talking about. Can you spell out the specifics of when this might occur?

proposals/4360-sliding-sync-extension-threads.md
MadLittleMods: Here is how I'm thinking about it right now: `/sync` provides all messages as they come in. If there is a gap (`limited: true`), then people can use `/thread_updates`.

Am I missing something?

proposals/4360-sliding-sync-extension-threads.md
devonh: A thread update is simply any event which has a `m.thread` relation. This doesn't currently include edits to any of the thread events. I'm not sure if edits would be particularly useful to clients, whereas new events in a given thread is a clear and meaningful change to a thread that a user would want to be notified about.

proposals/4360-sliding-sync-extension-threads.md
devonh: in the scenario of `limited: true` from the sliding sync response, this MSC would allow for any additional thread events that weren't included in the regular sync response to be included in the threads extension portion.
So you save yourself needing to go and do the extra query to `/thread_updates` (unless you are also limited on thread updates as well in the extension response)

ie.
Sync response covers events A-Z
Sync response only returns events A-M
The threads extension will include any threads that have updates caused by events N-Z.

devonh: So yes - you could just go and hit the new endpoint.
But this allows you not to need to do that, and to just receive those updates directly in the sync response you already have.

proposals/4360-sliding-sync-extension-threads.md
MadLittleMods: How do we expect a client to show the most recent view of a thread (with edits and other bundled aggregations)?

Imagine the scenario where you have kept up to date with all of the messages in a thread, then you go offline, some edits and reactions occur to the thread messages, more unrelated activity occurs to fill up the `/sync` response which leaves the edits/reactions in the gap and not included in the `timeline` when you come back online and `/sync` again.

proposals/4360-sliding-sync-extension-threads.md
MadLittleMods: I think was thinking of the backfill scenario. But I guess backfilled messages wouldn't be considered updates since they are historical :fast_forward: 

And for any new messages since a server joined, the server should be doing `/send` transactions so those messages will show up as thread updates.

---

The following is a general existing problem as far as I can tell but perhaps would be good more my understanding if I had an explanation here (segregated networks). I'm a bit confused on what would happen if you had servers A, B, C where B and C can't talk to each other directly. A new message from B will flow to A normally via `/send` transactions. And then the only way I could see the new message flowing from B to C is if C backfilled from A. But C has no reason to backfill so it probably won't see any of those messages :thinking:. Is that expected?

And even if it did backfill, they are backfilled messages so they would never been seen as thread updates. And I don't see how the client would know to re-paginate anything to show the full view of the thread. There weren't even any gaps when the client originally paginated the thread so [indicating gaps in the timeline](https://github.com/matrix-org/matrix-spec-proposals/pull/3871) wouldn't help either.

```mermaid
flowchart TD
    A <--> B
    A <--> C
```

Perhaps an even simpler scenario is if a bunch of activity occurred on one server and then it tries to `/send` it to the other server but it's is offline (or receives a 429 rate-limited response by the other server) and backs-off and gives up. In this case, we have a mechanism in Synapse to try to `/send` again later (wake-up destinations that need catchup).

proposals/4360-sliding-sync-extension-threads.md
MadLittleMods, line 35: Clarify what this includes in the case of initial sync

MadLittleMods: This reminds me of my [suggestion for sticky events](https://github.com/element-hq/synapse/pull/18968#discussion_r2383536342). In both of these cases, we are trying to tackle the same problem of gathering a subset of the events in the `timeline`:

> I think the better way to go about is to apply the same pattern that we just worked through with [thread subscriptions](https://github.com/matrix-org/matrix-spec-proposals/pull/4308). We would need a few things:
> 
> - A way to return the new sticky events in `/sync`
> - Dedicated `sticky_events_limited` flag when there are other sticky events that we didn't return
> - A pagination endpoint for the sticky events
> 
> To be more detailed, these could be normal `timeline` events. If there are more sticky events in between the given sync token and the current position, we set the `sticky_events_limited` flag. For the pagination endpoint, we could overload `/messages` with a new filter.
>
> *-- @MadLittleMods, [internal room](https://matrix.to/#/!vvWZSNiEDRDfsyFZfa:matrix.org/$OBWJFRGwD0ykF_xOGyGZBgbeQvGYXkUumdi-fj2V5-k?via=matrix.org&via=t2l.io&via=element.io)*

With some more thinking on the subject, the dedicated Sliding Sync extension worked well for `thread_subscriptions` because `thread_subscriptions` aren't part of the response already.

Whereas with sticky events and thread updates, they are already part of the `timeline`.

To break down the list of things needed (as listed above) and my current thinking:

#### "A way to return the new sticky events in `/sync`"

In the case of threads and sticky events, these events are already included as part of the `timeline` and `/sync` already keeps you up to date with all of the events.

#### "A pagination endpoint for the sticky events"

The dedicated `/threads_update` makes sense to back-paginate the gaps (whenever the `timeline` is `limited: true`)

#### "Dedicated `sticky_events_limited` flag when there are other sticky events that we didn't return"

With more thinking, I think this one is optional.

We could have an extension that indicates whether `thread_updates_limited` but this only saves a few extra requests in the cases where the `timeline` is `limited: true` but there weren't actually any new thread updates in the gap.


---


Having another field in the Sliding Sync extension to include more thread updates in initial and gappy syncs isn't that useful (and makes things more complicated). You can just call `/thread_updates`

devonh: @bnjbvr Would you be able to comment here on your experience with client-side thread syncing and whether what @MadLittleMods is suggesting would be better or worse?

proposals/4360-sliding-sync-extension-threads.md
devonh: I'm not sure what the best approach should be. I can see an argument where clients may or may not want to include events and/or annotations as "thread updates".
We could extend the MSC to include edits & annotations as "thread updates", but if we did that we may want to extend the extension config to make them optional, or provide a better way for clients to define what they want to be considered as a "thread update".

@bnjbvr Can you weigh in here on what clients would want?

proposals/4360-sliding-sync-extension-threads.md
devonh: I don't know if clients would care about getting "updates" for rooms they are no longer following.
The intent of the new API is to allow clients to follow along with new updates that are relevant. This is especially evident with the API design of only returning `(room_id, thread_id)` pairs in the response to indicate that a particular thread had at least one update in that window. The APIs are not meant to be able to query all thread events.

If a client wants historical thread events, they should use the `/relations` endpoint to retrieve them.

MadLittleMods: Once you're left, using `/threads` makes sense to find threads in the room and `GET /relations/{threadRootId}/m.thread` to find all events in the thread.

I think this just needs to be clarified that you should still get updates if you have a Sliding Sync `pos` token where you are joined to the room, some thread updates occur, you leave, and then you do another incremental sync (`newly_left`). You should get thread updates until you left.

PR #4354, opened 2025-09-16
MSC4354: Sticky Events

kegsay:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/persist-edu/proposals/4354-sticky-events.md)

- [x] Client (receive/handle) https://github.com/matrix-org/matrix-js-sdk/pull/5017
- [x] Client (usage) https://github.com/element-hq/element-call/pull/3513
- [x] Server
    - Original PoC Synapse PR: https://github.com/element-hq/synapse/pull/18968
    - Production-ready Synapse PRs:
        - [Support sending and receiving MSC4354 Sticky Event metadata](https://github.com/element-hq/synapse/pull/19591)
        - [Expose MSC4354 Sticky Events over the legacy (v3) /sync API](https://github.com/element-hq/synapse/pull/19487)
        - [Expose MSC4354 Sticky Events over MSC4186 (Simplified) Sliding Sync](https://github.com/element-hq/synapse/pull/19591)
        - [Catch up MSC4354 Sticky Events to remote homeservers that have missed them](https://github.com/element-hq/synapse/pull/20165) (not necessary to prove implementation, but useful for homeserver authors to reference)
- [x] Complement Tests https://github.com/matrix-org/complement/pull/806 

----

SCT Stuff:

~~[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4354#issuecomment-3353839132)~~

[FCP tickyboxes (round 2)](https://github.com/matrix-org/matrix-spec-proposals/pull/4354#issuecomment-5577222177)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4354#issuecomment-3353835809)

PR #4361, opened 2025-09-23
MSC4361: Non-federating Membership Authorization Rule Amendments

jevolk:
[rendered](https://github.com/matrix-construct/matrix-refreshed/blob/jevolk/non-federating-member-auth/proposals/4361-non-federating-member-auth.md)

### DIFF
--- a/proposals/4361-non-federating-member-auth.md
+++ b/proposals/4361-non-federating-member-auth.md
@@ -1,20 +1,38 @@
 # MSC4361: Non-federating Membership Authorization Rules
 
-Specification version 1.16. Room Version 12[^1] § 2.2 (4)[^2] is the sole clause within the Authorization Rules conditioned on the `m.federate` property. The `m.federate` property is an immutable declaration within the `m.room.create` event which isolates a room to a specific origin[^3]. The sole condition to achieve a non-federating room is to deny events with a `sender` from a remote server.
+Specification version 1.16. Room Version 12[^1] § 2.2 (4)[^2] is the sole clause within the
+Authorization Rules conditioned on the `m.federate` property. The `m.federate` property is
+an immutable declaration within the `m.room.create` event which isolates a room to a specific
+origin[^3]. The sole condition to achieve a non-federating room is to deny events with a
+`sender` from a remote server.
 
-This method has several leakages. The "target" of an `m.room.member` event declared by its `state_key` is allowed to be a remote user. Compliant servers will both create, transmit[^4] and accept such an event, only to find later it cannot be further acted upon. This specification change aims to correct this leakage and subsequent ⊥ by amending Rule `5. If type is m.room.member` prepending a new sub-rule `1. If the content of the m.room.create event in the room state has the property m.federate set to false, and the state_key domain of the event does not match the sender domain of the create event, reject.` shifting-up existing rules.
+This method has several leakages. The "target" of an `m.room.member` event declared by its
+`state_key` is allowed to be a remote user. Compliant servers will both create, transmit[^4]
+and accept such an event, only to find later it cannot be further acted upon. This
+specification change aims to correct this leakage and subsequent ⊥ by amending Rule `5. If
+type is m.room.member` prepending a new sub-rule `1. If the content of the m.room.create event
+in the room state has the property m.federate set to false, and the state_key domain of the
+event does not match the sender domain of the create event, reject.` shifting-up existing rules.
 
 ## Alternatives
 
-Synapse has been observed in at least some cases to preclude the transmission and/or reception of events for non-federating rooms "manually." These restrictions are specified external to the authorization rules in some other places. This approach denudes the significance of authorization rules. One could ask: why specify _anything_ as authorization rules? This is the question a proponent of the status quo must bear.
+Synapse has been observed in at least some cases to preclude the transmission and/or reception
+of events for non-federating rooms "manually." These restrictions are specified external to
+the authorization rules in some other places. This approach denudes the significance of
+authorization rules. One could ask: why specify _anything_ as authorization rules? This is
+the question a proponent of the status quo must bear.
 
 ## Security Considerations
 
-While this proposal clearly increases security by reducing cross-origin leakage of events, the addition of any new restrictions is a theoretical Denial of Service vector which must be scrutinized during review. No such possibility is known to the author.
+While this proposal clearly increases security by reducing cross-origin leakage of events,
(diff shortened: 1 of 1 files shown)


proposals/4361-non-federating-member-auth.md
```diff
@@ -0,0 +1,29 @@
...
+
+[1]: https://spec.matrix.org/v1.16/rooms/v12/#authorization-rules
+
+<sup><b>Special thanks to Dasha for assisting in the research for this proposal.</b></sup>
+
+<sup>* In addition to all earlier stable room versions.</sup>
+
+<sup>† Server or cluster sharing the same domain.</sup>
+
+<sup>‡ Synapse has been observed to impose certain restrictions external to the auth-rules. See: Alternatives for discussion.</sup>
```
Johennes: Not super important on this short proposal but I think the built-in footnotes in GitHub's markdown would make for better readability. They're numbered and navigateable in both directions. See e.g. [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4354/files#diff-d76bc1a1d612c6da37d024f5b57f7b8352939b8db8a7ee9c6b71c1a848359afdR348) for an example.

proposals/4361-non-federating-member-auth.md
tulir, line 1: Implementation requirements:

* [ ] Server

proposals/4361-non-federating-member-auth.md
```diff
@@ -0,0 +1,29 @@
+# MSC4361: Non-federating Membership Authentication Rules
+
+Specification version 1.16. Room Version 12<sup>*</sup> § 2.2 (4)<sup>[1]</sup> is the sole clause within the Authentication Rules conditioned on the `m.federate` property. The `m.federate` property is an immutable declaration within the `m.room.create` event which isolates a room to a specific origin<sup>†</sup>. The sole condition to achieve a non-federating room is to deny events with a `sender` from a remote server.
+
+This method has several leakages. The "target" of an `m.room.member` event declared by its `state_key` is allowed to be a remote user. Compliant servers will both create, transmit<sup>‡</sup> and accept such an event, only to find later it cannot be further acted upon. This specification change aims to correct this leakage and subsequent ⊥ by amending Rule `5. If type is m.room.member` prepending a new sub-rule `1. If the content of the m.room.create event in the room state has the property m.federate set to false, and the state_key domain of the event does not match the sender domain of the create event, reject.` shifting-up existing rules.
```
tulir: Please wrap lines to ~120 characters

proposals/4361-non-federating-member-auth.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Alternatives
+
+Synapse has been observed in at least some cases to preclude the transmission and/or reception of events for non-federating rooms "manually." These restrictions are specified external to the authentication rules in some other places. This approach denudes the significance of authentication rules. One could ask: why specify _anything_ as authentication rules? This is the question a proponent of the status quo must bear.
```
kegsay: My understanding of this proposal is that, when `m.federate` is `false`. 

> The "target" of an `m.room.member` event declared by its `state_key` is allowed to be a remote user.

This will pass auth checks locally on compliant HSes, since the auth rules only check the `sender`:

> If the content of the m.room.create event in the room state has the property m.federate set to false, and the sender domain of the event does not match the sender domain of the create event, reject.

..but will be completely useless and inactionable. Either servers won't send it to the target server, or if they do, they can't convert it into a valid membership.

It seems sensible to just catch this at event creation time on compliant servers.

jevolk: The naive compliant server currently:
1. Creates an invite which passes auth checks.
2. Transmits the invite to the remote server where it's rightfully targeted.
3. The remote server acknowledges the invite (as it either doesn't know not to and it still passes auth-checks anyway).
4. Presents the invite to the remote user.
5. The remote user accepts the invite.

_Only then_ does their join event fail. This is the behavior of condu*.

Auth checks already fetch and scrutinize the create event for several reasons (`m.federate` being one). Are you suggesting it is better to fetch the create event to examine some other event outside the context of auth-rules instead, but while auth-rules already has that exact purpose?

kegsay: Nope, I'm not making any suggestions here. I'm wording the proposal in a way that makes sense to me, and I think it's a fine addition to make. I do wonder if this should be more restricted specifically for invites which has this problem, as opposed to member events in general?

jevolk: What I'm looking for is a reason why remote-user state keys shouldn't be summarily blocked. @anoadragon453 was chasing this exact line of thought too and I feel like we all need to meditate on this for a bit until that reason reveals itself. Conversely if none exists, and we only block invites now out of caution, we might regret when some foul play occurs in the future requiring yet another revisit to this.

kathleenmotley49-maker (approved):

proposals/4361-non-federating-member-auth.md
dasha-uwu: implementation:
ruma https://github.com/matrix-construct/ruma/tree/msc4361
tuwunel https://github.com/matrix-construct/tuwunel/tree/msc4361

proposals/4361-non-federating-member-auth.md
anoadragon453: Looking at the [other available membership types](https://spec.matrix.org/v1.16/client-server-api/#mroommember):

* `knock` events would be sent by a remote user, so would be blocked by the `sender` rule already. Likewise with `join`.
* `ban`ning a remote user is pointless, since they won't be able to join the room anyhow. But it's not necessarily problematic.
* `leave` would similarly be pointless, as remote users cannot join the room.

And we may not want to limit future membership types. I suggest we just focus on `invite` in this MSC.

jevolk: > And we may not want to limit future membership types. 

Can you think of any reason a future membership type should circumvent the non-federating restriction? That's really the issue here. I can't think of a good one. The natural forward compatibility given by *not* focusing on invites is an asset: we won't have to remember this MSC and case them going forward. This MSC only exists because the non-federating case was already forgotten.

If there is a reason to involve federation for a new membership type, it can always be special cased. Otherwise the default should be to not involve federation when no federation is optioned for the room.

jevolk: Also ftr I don't feel strongly about this issue or that it's worth holding up the MSC for. If the consensus is to move forward with just invites covered it's great and I'm happy to amend.

PR #4362, opened 2025-09-25
MSC4362: Simplified Encrypted State Events

kaylendog:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/kaylendog/msc/simplified-encrypted-state/proposals/4362-simplified-encrypted-state.md)

Implementations:

- `matrix-js-sdk`: https://github.com/matrix-org/matrix-js-sdk/pull/4994
- `matrix-rust-sdk`: There are quite a lot of PRs involved in this implementation, you can view the list [here](https://github.com/matrix-org/matrix-rust-sdk/issues?q=assignee%3A%40me%20state%3Amerged%20state%20events).
- `element-web`: https://github.com/element-hq/element-web/pull/30877, [Netlify deployment](https://pr30877--matrix-react-sdk.netlify.app/)

### DIFF
--- a/proposals/4362-simplified-encrypted-state.md
+++ b/proposals/4362-simplified-encrypted-state.md
@@ -61,7 +61,9 @@ An encrypted state event looks very similar to a regular encrypted room message:
 `m.room.encrypted` and the `content` is the same shape as a regular `m.room.encrypted` event. The
 `state_key` for encrypted state events is constructed from the plaintext `type` and `state_key`
 fields, formatted as `{type}:{state_key}`, preserving the uniqueness of the `type`-`state_key`
-mapping required for the server to perform state resolution.
+mapping required for the server to perform state resolution. In rooms where both encrypted and
+unencrypted versions of the same state event exist (that is, for the same `(type, state_key)` pair),
+clients **must** use the encrypted version and ignore the unencrypted one.
 
 To track whether a room has state encryption enabled, and to preserve compatibility with older
 clients that cannot work with encrypted state events, a new boolean field `encrypt_state_events` is
@@ -104,8 +106,8 @@ and interpret state events unencrypted.
 
 To encrypt an `m.room.name` state event, the client first constructs the "packed state key" by
 concatenating the event type and the state key, separated by a colon (`:`), following the template
-`<event_type>:<state_key>`. For `m.room.name` events, the `state_key` is typically an empty string,
-so the packed state key becomes `m.room.name:`.
+`<type>:<state_key>`. For `m.room.name` events, the `state_key` is typically an empty string, so the
+packed state key becomes `m.room.name:`.
 
 Next, the client prepares the plaintext payload to be encrypted. This payload contains the original
 event content and state key:
--- a/proposals/4362-simplified-encrypted-state.md
+++ b/proposals/4362-simplified-encrypted-state.md
@@ -14,13 +14,27 @@ would be a good thing to list here._
 _If you're having troubles coming up with a description, a good question to ask is "how does this
 proposal improve Matrix?" - the answer could reveal a small impact, and that is okay._-->
 
-This proposal builds upon the earlier MSC3414, aiming to provide a simplified approach to encrypted
-state events in Matrix. Currently, all room state is unencrypted and accessible to everyone in the
-room, and occasionally people outside the room (such as via the public room directory, invite state,
-or peekable rooms). The server also has access to these state events in order to perform state
-resolution, and so is visible to server administrators. Most events in room state could be encrypted
-to provide confidentiality, which is what this MSC seeks to achieve more straightforwardly. Some
-parts, however, cannot be encrypted to maintain a functioning protocol.
+Currently, all state events are unencrypted. This allows the homeserver to read state event content
+in order to do its job in implementing the Matrix protocol: processing room membership and power
+levels, and performing state resolution. A side effect of homeservers being able to read state event
+content is that anyone with access to the homeserver's data (such as an administrator or a
+successful attacker) can also read these events.
+
+The set of events that are actually needed by the homeserver is quite small, so we propose
+encrypting everything else. This provides a significant reduction in the amount of visible metadata,
+at the cost of some user inconvenience (because users need decryption keys to see state information
+like room names).
+
+[MSC3414](https://github.com/matrix-org/matrix-spec-proposals/pull/3414) has similar goals to this
+proposal, but it specifies a concrete mechanism for hiding encrypted event types, and resolving
+state where it cannot be fully resolved by the server. We think this approach could be problematic,
+and may effectively require us to implement full state resolution on the client. Here, we simply
+propose the "easy" part: encrypting state events without hiding their types from the server.
+
+The intent is to allow real-world usage of encrypted state, accepting the limitations imposed
+because state is hidden from users in situations where they might want it, without requiring us to
+draw conclusions on the trickiest parts (sharing historical state, resolving state the server can't
+identify, and exposing room names and topics).
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,172 @@
...
+_This is where alternative solutions could be listed. There's almost always another way to do things
+and this section gives you the opportunity to highlight why those ways are not as desirable. The
+argument made in this example is that all of the text provided by the template could be integrated
+into the proposals introduction, although with some risk of losing clarity._-->
+
+A number of alternatives to string-packing the plaintext `type` and `state_key` are possible:
+
+- Preserving the values of `type` and `state_key`;
+- Introducing an adjacent `true_type` field;
+- Hashing `type` and `state_key` with HMAC.
```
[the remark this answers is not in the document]
Gnuxie: Would it be possible to instead have an extensible events mixin from encrypted content?

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Preserved Fields
+
+Rather than string-packing the `type` and `state_key` together, we could preserve these values on
+the encrypted event, but still encrypt the event content. This provides the same (lack of)
+confidentiality as the approach laid out in this MSC while avoiding string packing. However, this
+approach would introduce a difference between the encryption of message events and state events,
+which may be undesirable.
```
Gnuxie, line 244: Would it be possible to instead have an extensible events mixin for the encrypted content? Why is this undesirable?

kaylendog: Like in [MSC3956](https://github.com/matrix-org/matrix-spec-proposals/pull/3956)? I think the consensus is that an `m.encrypted` mixin would allow partial encryption of event types, which, per the author's note on that MSC, is undesirable. I would have to ask around to get a better idea, however, since progress on that MSC seems to have stalled. I cant think of any immediate issues with partial encryption, but it would require some work to get it integrated with the existing `matrix-sdk-crypto` and `ruma` crates.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,173 @@
...
+- `m.room.member`
+- `m.room.join_rules`
+- `m.room.power_levels`
+- `m.room.third_party_invite`
+- `m.room.history_visibility`
+- `m.room.guest_access`
+- `m.room.encryption`
+
+An encrypted state event looks very similar to a regular encrypted room message: the `type` becomes
+`m.room.encrypted` and the `content` is the same shape as a regular `m.room.encrypted` event. The
```
nico-famedly, line 61: This doesn't seem to address the core problem with encrypted state events, where state events have a different lifecycle from timeline events. But I also don't see the problem mentioned on the MSC at all.

Say you have this timeline in a room with history visibility invite:

```
m.room.name <- A
m.room.message <-B
invite to X
X joins
```

To make the room name sent at A visible to X, you need to share the megolm key for it, which might include the key for decrypting message B as well.

To prevent that, you basically need a different megolm session for every (event_type, state_key) tuple. Otherwise you might leak an arbitrary set of past messages or states.

kaylendog: Yes! I do not intend to address the issue of key-sharing in this MSC, except for what I've said about history sharing. I briefly covered my thoughts on how this would be done for the HMAC key distribution below, and I imagine we could re-use this infrastructure to resolve this..

kathleenmotley49-maker (approved):

proposals/4362-simplified-encrypted-state.md
```diff
@@ -62,33 +76,29 @@ This MSC relies on the room key sharing mechanism outlined in
 [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268), which enables clients to
 decrypt historical state events.
 
+## Limitations
+
+### Room names and topics are not visible from outside
+
+The name and topic of a room with encrypted state will not be visible without access to the keys
+used to encrypt them. Without additional proposals, this will make it impossible to provide a room
+directory entry, list the room inside a space, or display room details when invited.
```
kaylendog, line 203: Would it be worth mentioning potential fixes for these limitations in the content of this MSC, or is that generally handled elsewhere? We previously discussed the potential of a `m.space.child_info` event to mitigate this, which I included in my earlier draft.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,301 @@
...
+The client must validate the following:
+
+- The outer event's `state_key` must be present if and only if the inner (decrypted) event's
+  `state_key` is present. That is, both should either be present or both absent.
+- The outer event's `state_key` must be able to be split into a `(type, state_key)` pair using the
+  `type:state_key` format (for example, `("m.room.topic", "")` in this case).
+- The `(type, state_key)` pair obtained from unpacking the outer event's `state_key` must exactly
+  match the `type` and `state_key` fields found in the decrypted (inner) event.
+
+If any of these checks fail, the event should be considered invalid and ignored.
```
kaylendog, line 195: This might belong in the previous section since it's the only content currently outlining what it means to verify a packed state key.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,301 @@
...
+directory entry, list the room inside a space, or display room details when invited.
+
+### State sent before joining the room is inaccessible
+
+Upon joining a room with encrypted state, new users will not be able to decrypt room state, making
+the room name, topic and other information (e.g. ongoing whiteboard sessions or call) inaccessible.
+
+This limitation does not apply if
+[MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268) is available and the room
+settings allow sharing the relevant events.
```
kaylendog, line 212: Do you think it's worth clarifying that we allow state event encryption to be enabled at any time, like message encryption? We don't state that anywhere in the proposal at the moment, although it can be assumed since we're using the existing `m.room.encryption`.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -76,6 +76,122 @@ This MSC relies on the room key sharing mechanism outlined in
...
+    "device_id": "<sender_device_id>",
+    "sender_key": "<sender_curve25519_key>",
+    "session_id": "<outbound_group_session_id>"
+  }
+}
+```
+
+Clients receiving this event will use the packed state key to determine which state event it
+represents, decrypt the payload, and verify that the decrypted `type` and `state_key` match the
+packed state key.
```
andybalaam, line 152: Also, because the packed state key is unique for a given type, state_key pair, with no server changes, server provided state will contain the latest state event for each pair (subject to state resolution as normal).

andybalaam: In rooms with encrypted state, clients MUST prefer encrypted state over unencrypted state if both are present for a given type, state_key pair.

andybalaam: ^ Maybe this should go elsewhere, but these worked examples are a nice way to explain what's happening.

proposals/4362-simplified-encrypted-state.md
andybalaam: I think it's OK here - we don't have to be super-formal in MSCs.

proposals/4362-simplified-encrypted-state.md
kaylendog: > Also, because the packed state key is unique for a given type, state_key pair, with no server changes, server provided state will contain the latest state event for each pair (subject to state resolution as normal).

I think we cover this in the second paragraph of the proposal

> The `state_key` for encrypted state events is constructed from the plaintext `type` and `state_key`
fields, formatted as `{type}:{state_key}`, preserving the uniqueness of the `type`-`state_key`
mapping required for the server to perform state resolution.

kaylendog: > In rooms with encrypted state, clients MUST prefer encrypted state over unencrypted state if both are present for a given type, state_key pair.

To me, this seems relevant to include in the core proposal content? I'll make a change.

proposals/4362-simplified-encrypted-state.md
andybalaam: I feel like it's still quite open how we could deal with this, which is why I removed that part - it's just one example of what we could do. But if you want to put it in, that's fine by me.

proposals/4362-simplified-encrypted-state.md
andybalaam: > I think we cover this in the second paragraph of the proposal

Yeah, but I thought it was worth re-iterating here since we're talking through what happens. (Feel free to leave out)

proposals/4362-simplified-encrypted-state.md
andybalaam: Yes, that is a good idea.

Also, I feel like I heard that you can't send more than one `m.room.encryption` event in the entire history of a room (later ones would be ignored) but I just checked the spec and I can't see that anywhere. Maybe I'm making it up.

It would be a good idea to decide whether it is possible to turn on state encryption later in the history of a room, after encryption was already turned on, and whether it's possible to turn it off again by sending another `m.room.encryption` later.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,303 @@
...
+- `m.room.member`
+- `m.room.join_rules`
+- `m.room.power_levels`
+- `m.room.third_party_invite`
+- `m.room.history_visibility`
+- `m.room.guest_access`
+- `m.room.encryption`
+
+An encrypted state event looks very similar to a regular encrypted room message: the `type` becomes
+`m.room.encrypted` and the `content` is the same shape as a regular `m.room.encrypted` event. The
```
krille-chan, line 61: When so many state events then have the type `m.room.encrypted` the server behavior also needs to change. Right now every homeserver checks the type to determine if a user is allowed to push a state event but here the real type is now in the first half of the state_key, so the server needs to support a logic like this:

- Is event.type == `m.room.encrypted`?
- No: Use event.type to calculate permission to change the event
- Yes: Use `event.state_key.split(':').first` to calculate permission to change the event

krille-chan: A possible solution would be to switch out type and state key here. So that we have a normal type of for example `m.room.name` and the state key becomes `m.room.encrypted:`

andybalaam: Yes, I think that might work, but it would require more changes on the client side: with the type being `m.room.encrypted` it was relatively easy to persuade clients to decrypt these events.

But that's not an in-principle objection, and might not actually be a big issue either.

proposals/4362-simplified-encrypted-state.md
```diff
@@ -0,0 +1,303 @@
...
+encrypting eligible state events according to this proposal, and will also attempt to decrypt any
+encrypted state events they receive. Clients that do not support this feature will continue to send
+and interpret state events unencrypted.
+
+### Sending an encrypted state event
+
+To encrypt an `m.room.name` state event, the client first constructs the "packed state key" by
+concatenating the event type and the state key, separated by a colon (`:`), following the template
+`<type>:<state_key>`. For `m.room.name` events, the `state_key` is typically an empty string, so the
+packed state key becomes `m.room.name:`.
```
codedust, line 110: This obviously leaks the event type and the state key. If I'm not wrong (I might be!), the only important property of the packed state key is that it is unique. So, couldn't we just hash and pepper it as follows

```
packed state key = hash("<type>:<state_key>:<random_uuid>")
```

The unhashed event type and state key is already included in the plaintext payload (see example below). We only need to include the random uuid from the example above in the plaintext payload of the event:

```json
{
  "room_id": "!room:example.org",
  "type": "m.room.name",
  "state_key": "",
  "packed_state_key_hash_entropy": "<random_uuid_from_packed_state_key_hash>"
  "content": {
    "name": "Example"
  }
}
```

Clients would then need to verify that `packed state key == hash("<type>:<state_key>:<random_uuid_from_packed_state_key_hash>")`

Subsequent updates of the room state need to use the same `<random_uuid_from_packed_state_key_hash>` for this combination of event type and state key.

PR #4363, opened 2025-09-30
MSC4363: OAuth step up authentication

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/oauth-step-up-authentication/proposals/4363-oauth-step-up-authentication.md)

### DIFF
--- a/proposals/4363-oauth-step-up-authentication.md
+++ b/proposals/4363-oauth-step-up-authentication.md
@@ -36,7 +36,7 @@ authentication of 5 minutes is required. The ACR values were taken from [Okta's 
   "errcode": "M_INSUFFICIENT_USER_AUTHENTICATION",
   "error": "Additional authentication required to complete request",
   "acr_values": "urn:okta:loa:2fa:any urn:okta:loa:1fa:pwd",
-  "max_age": 300,
+  "max_age": 300
 }
 ```
 
@@ -69,9 +69,30 @@ None apparent.
 
 ## Alternatives
 
-The scheme proposed in [MSC4312] (ab)uses UIA to deeplink into the authorization server's account
-management UI. This is less flexible and powerful than OAuth step up authentication, however. It is
-also contrary to the idea of standardizing authentication in Matrix on OAuth mechanisms.
+Rather than extending the existing [standard error response], the error and parameters could be
+communicated via the `WWW-Authenticate` header as suggested in [RFC9470]. Reusing the example from
+above, the header would look like this:
+
+``` http
+HTTP/1.1 401 Unauthorized
+WWW-Authenticate: Bearer error="insufficient_user_authentication",
+  error_description="Additional authentication required to complete request",
+  acr_values="urn:okta:loa:2fa:any urn:okta:loa:1fa:pwd",
+  max_age="300"
+```
+
+Switching to the header would break with the established error mechanism in Matrix, however. As a
+middleground, the header could also be used *on* standard error responses. It's unclear though what
+advantages it would bring to have two different mechanisms that convey the same information.
+
+A [previous version] of [MSC2967] used the `insufficient_scope` error code from [RFC6750] (The OAuth
+2.0 Authorization Framework: Bearer Token Usage) to communicate missing scopes back to the client.
+This is comparable to but less flexible than the present proposal as it lacks ACRs and token age.
+
+Finally, the scheme proposed in [MSC4312] (ab)uses UIA to deeplink into the authorization server's
+account management UI. This is less flexible and powerful than OAuth step up authentication,
+however. It is also contrary to the idea of standardizing authentication in Matrix on OAuth
+mechanisms.
 
 ## Security considerations
 
@@ -107,4 +128,8 @@ The new authorization request parameters should not use prefixes since they foll
   [RFC9470]: https://datatracker.ietf.org/doc/rfc9470/
   [Okta's reference]: https://developer.okta.com/docs/guides/step-up-authentication/main/
   [current list of request parameters]: https://spec.matrix.org/v1.15/client-server-api/#login-flow
+  [standard error response]: https://spec.matrix.org/v1.16/client-server-api/#standard-error-response
+  [previous version]: https://github.com/matrix-org/matrix-spec-proposals/pull/2967/commits/544d75b413277f14393a635ea17be3d0d49529c5
+  [MSC2967]: github.com/matrix-org/matrix-doc/pull/2967
+  [RFC6750]: https://datatracker.ietf.org/doc/html/rfc6750
   [OIDC core specification]: https://openid.net/specs/openid-connect-core-1_0.html#Terminology
--- a/proposals/4363-oauth-step-up-authentication.md
+++ b/proposals/4363-oauth-step-up-authentication.md
@@ -1,0 +1,110 @@
+# MSC4363: OAuth step up authentication
+
+Homeservers may want to enforce re-authentication to guard access to potentially severe user actions
+such as deactivating an account or logging out all devices. Prior to the [OAuth APIs] introduced in
+Matrix 1.15, this was handled by an API called [User-Interactive Authentication (UIA)]. UIA is
+incompatible with the new [OAuth APIs], however, leading to suboptimal workarounds such as
+[MSC4312].
+
+Fortunately, OAuth itself provides a scheme for elevating access rights with [RFC9470] (OAuth 2.0
+Step Up Authentication Challenge Protocol). The present proposal generalizes [RFC9470] for use with
+Matrix.
+
+## Proposal
+
+The following description assumes that the client has authenticated via the [OAuth APIs].
+
+When a homeserver determines that the authentication event associated with the access token
+presented on a request is insufficient, it MAY respond with HTTP 401 and a new error code
+`M_INSUFFICIENT_USER_AUTHENTICATION`. Three optional top-level properties are allowed in the
+response to convey the authentication requirements challenge back to the client:
+
+- `acr_values`: A space-separated string listing the authentication context class reference (ACR)
+  values in order of preference.[^1] The protected resource requires *one* of these values for the
+  authentication event associated with the access token.
+- `max_age`: A non-negative integer specifying the allowable elapsed time in seconds since the last
+  active authentication event associated with the access token.
+- `scope`: A space-separated string listing the *full* set of scopes that are required to access the
+  protected resource.
+
+The following is an example of an error response that requests authentication using either any 2FA
+method, or, failing that, password-based authentication. Additionally, a maximum elapsed time since
+authentication of 5 minutes is required. The ACR values were taken from [Okta's reference].
+
+``` json5
+{
+  "errcode": "M_INSUFFICIENT_USER_AUTHENTICATION",
+  "error": "Additional authentication required to complete request",
+  "acr_values": "urn:okta:loa:2fa:any urn:okta:loa:1fa:pwd",
+  "max_age": 300,
+}
+```
+
+A client receiving a challenge SHOULD use the `acr_values`, `max_age` and `scope` properties to
+construct a new authorization request. For this purpose, `acr_values` and `max_age` are added to the
+[current list of request parameters] using the same definition given above.
+
+Following successful authentication, the client can use the new access token to repeat the original
+request.
+
+### Aspects that are out of scope
+
+This proposal consciously leaves a number of things undefined.
+
+Firstly, the act of evaluating whether or not an access token meets a resource's requirements is
+left as an implementation detail between the homeserver and the authorization server. As per
+[RFC9470], this can, for instance, be achieved by encoding the access token with additional
+properties in JWT format or through a dedicated token introspection endpoint.
+
+Furthermore, no concrete ACR values are specified. Again, this is a detail to arrange between the
+homeserver and the authorization server as ACR values are not intended for introspection by clients.
+
+Finally, it is left up to the authorization server to determine how exactly the `acr_values`,
+`max_age` and `scope` parameters are used during authentication. Implementations should, however,
+note the recommendations in [RFC9470].
+
+## Potential issues
+
+None apparent.
+
+## Alternatives
+
+The scheme proposed in [MSC4312] (ab)uses UIA to deeplink into the authorization server's account
+management UI. This is less flexible and powerful than OAuth step up authentication, however. It is
+also contrary to the idea of standardizing authentication in Matrix on OAuth mechanisms.
+
+## Security considerations
+
+The security impact of this proposal largely follows [RFC9470].
+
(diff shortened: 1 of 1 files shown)


proposals/4363-oauth-step-up-authentication.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,110 @@
...
+
+Finally, it is left up to the authorization server to determine how exactly the `acr_values`,
+`max_age` and `scope` parameters are used during authentication. Implementations should, however,
+note the recommendations in [RFC9470].
+
+## Potential issues
+
+None apparent.
+
+## Alternatives
```
zecakeh, line 82: Two other alternatives:

- The method that was initially in MSC2967 was to define more scopes, and use the `insufficient_scope` error code from [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750).
- Use the `WWW-Authenticate` header as defined in the RFC. The alternative in MSC2967 used a mix for the error: the `WWW-Authenticate` header as defined in the RFC, and a Matrix error in the body with an `M_FORBIDDEN` error code.

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,110 @@
...
+  "max_age": 300,
+}
+```
+
+A client receiving a challenge SHOULD use the `acr_values`, `max_age` and `scope` properties to
+construct a new authorization request. For this purpose, `acr_values` and `max_age` are added to the
+[current list of request parameters] using the same definition given above.
+
+Following successful authentication, the client can use the new access token to repeat the original
+request.
```
zecakeh: We should clarify what the client should do with the access token, because the RFC says that it is application-dependent.

Possible options:

1. **The privileged access token can only be used once for the original request but clients keep using the old access token for other requests**.

    This would mean that there is no `refresh_token` when the access token is returned from the token endpoint.
    
    It is simple, but it requires a new authorization every time that a privileged request is made. A common behavior nowadays is to be able to do several privileged requests in a row for a limited time, without requiring further authorization.

2. **The previous access token is superseded by the privileged access token, which should be used until it expires. When it expires, refreshing the token should provide an unprivileged access token.**

    Instead of being valid for a single use, the privileged access token is short-lived.

    If we decide to request the same scope as during login, including the device ID, this option would also match the [current text from the spec](https://spec.matrix.org/v1.16/client-server-api/#relationship-between-access-tokens-and-devices):

    > If the client sets the `device_id`, the server will invalidate any access and refresh tokens previously assigned to that device.

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,110 @@
...
+  "errcode": "M_INSUFFICIENT_USER_AUTHENTICATION",
+  "error": "Additional authentication required to complete request",
+  "acr_values": "urn:okta:loa:2fa:any urn:okta:loa:1fa:pwd",
+  "max_age": 300,
+}
+```
+
+A client receiving a challenge SHOULD use the `acr_values`, `max_age` and `scope` properties to
+construct a new authorization request. For this purpose, `acr_values` and `max_age` are added to the
+[current list of request parameters] using the same definition given above.
```
zecakeh: I believe that we should clarify which `scope` needs to be sent in the request, if it is not provided in the error.

Since an access token [is usually associated to a device ID](https://spec.matrix.org/v1.16/client-server-api/#relationship-between-access-tokens-and-devices), I guess that we can just request the exact same scope as in the login flow, using the same device ID as during login?

proposals/4363-oauth-step-up-authentication.md
Johennes: Thanks for digging these out! ❤️ 

Have added both in 4c716b68c3a11aff518f00bd2faa14fa3be9cddc.

proposals/4363-oauth-step-up-authentication.md
Johennes: I wonder if we could leave this as an implementation detail between the home server and the authorization server? That would require clients to be able to reasonably infer how to use the token. I guess that could be tricky. 🤔

proposals/4363-oauth-step-up-authentication.md
Johennes: Good point. Have clarified this with d48b01d87f55ee013fccd514db5dab6077445d3c.

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,137 @@
...
+
+Following successful authentication, the client can use the new access token to repeat the original
+request.
+
+### Aspects that are out of scope
+
+This proposal consciously leaves a number of things undefined.
+
+Firstly, the act of evaluating whether or not an access token meets a resource's requirements is
+left as an implementation detail between the homeserver and the authorization server. As per
```
zecakeh: In the Matrix specification there is no distinction between the homeserver and the authorization server, i.e. the homeserver is considered to be the authorization server when clients make OAuth 2.0 requests (even if it might not be true).

proposals/4363-oauth-step-up-authentication.md
zecakeh: The reason that I brought it up is that whatever we decide to be the correct solution for **clients**, it needs to be specced. No assumption is possible because it is intentionally out of scope of the RFC, as per the last paragraph of [section 2](https://datatracker.ietf.org/doc/html/rfc9470#section-2), specifically:

> This document does not recommend any specific token-caching strategy: that choice will be dependent on the characteristics of every particular scenario and remains application-dependent as in the core OAuth cases.

proposals/4363-oauth-step-up-authentication.md
Johennes: Right, yes. It's slightly confusing because homeservers can also delegate authorization to e. g. MAS. Have tried to clarify it with bbae0bb542e027f05387baec9ed4bcaabd64f684.

proposals/4363-oauth-step-up-authentication.md
Johennes: The RFC deliberately leaves it unspecified, yes. Matrix is also just a protocol with many different use cases, though. That's why I wasn't sure if we need to prescriptive here.

I think from an implementation's perspective, option 2 seems preferable because clients and servers can continue to maintain a single refreshable token per device. Homeservers could still enforce one-shot semantics by just reapplying the step up mechanism if they find the token to have been used on a protected resource before. Does that make sense?

zecakeh: > Matrix is also just a protocol with many different use cases, though. That's why I wasn't sure if we need to prescriptive here.

Of course, and it's always possible to use different solutions in private deployments/other specs, but we need to specify what public homeservers should do, so that Matrix clients can interact with them.

> I think from an implementation's perspective, option 2 seems preferable because clients and servers can continue to maintain a single refreshable token per device. Homeservers could still enforce one-shot semantics by just reapplying the step up mechanism if they find the token to have been used on a protected resource before. Does that make sense?

Completely, this is why I proposed this solution, it seems to be the best of both worlds: minimal changes in the client and the criteria for the validity of the step-up are left as an implementation detail for homeservers.

Johennes: > Of course, and it's always possible to use different solutions in private deployments/other specs, but we need to specify what public homeservers should do, so that Matrix clients can interact with them.

Ok yeah, I think you're right.

> Completely, this is why I proposed this solution, it seems to be the best of both worlds: minimal changes in the client and the criteria for the validity of the step-up are left as an implementation detail for homeservers.

Great. 👍 Have tried to add it into the proposal text with 8cdb9b63ac74d5d43218497f7d8ce8661eafb2b7.

zecakeh: It looks great, thanks!

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,147 @@
...
+}
+```
+
+A client receiving a challenge SHOULD use the `acr_values`, `max_age` and `scope` properties to
+construct a new authorization request. For this purpose, `acr_values` and `max_age` are added to the
+[current list of request parameters] using the same definition given above. If the challenge doesn't
+include any scopes, the client SHOULD use the same scopes it used during login on the authorization
+request.
+
+Following successful authentication, the client's previous access token SHOULD be invalidated and a
```
zecakeh:
```suggestion
Following successful authorization, the client's previous access token SHOULD be invalidated and a
```

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,147 @@
...
+new, more privileged, token be issued. The new token SHOULD have a short lifetime. The client can
+then use the new token to repeat the original request but may also use it for other API requests.
+Renewing the token SHOULD produce a less privileged token again. This approach ensures that
```
zecakeh:
```suggestion
Refreshing the token SHOULD produce a less privileged token again. This approach ensures that
```

zecakeh (commented):
Just a couple of nits on the wording.

proposals/4363-oauth-step-up-authentication.md
Johennes: I always mix these up. 🤦‍♂️

kathleenmotley49-maker (approved):

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,147 @@
...
+Fortunately, OAuth itself provides a scheme for elevating access rights with [RFC9470] (OAuth 2.0
+Step Up Authentication Challenge Protocol). The present proposal generalizes [RFC9470] for use with
+Matrix.
+
+## Proposal
+
+The following description assumes that the client has authenticated via the [OAuth APIs].
+
+When a homeserver determines that the authentication event associated with the access token
+presented on a request is insufficient, it MAY respond with HTTP 401 and a new error code
```
hughns:
```suggestion
presented on a request is insufficient, it MAY respond with HTTP 401 and a new common error code
```

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+
+Finally, it is left up to the homeserver to determine how exactly the `acr_values`, `max_age` and
+`scope` parameters are used during authentication. Implementations should, however, note the
+recommendations in [RFC9470].
+
+### Cross-signing key reset
+
+Implementations MAY use this proposal to replace UIA when the client has obtained its access token
+via the OAuth APIs. The only endpoint currently using UIA for which this is relevant is
+[/_matrix/client/v3/keys/device_signing/upload].
```
Johennes, line 76: We could consider introducing a dedicated scope to support this. Though I think it is not strictly needed. The only advantage I can think of is that clients could request the scope ahead of hitting the endpoint.

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+Furthermore, no concrete ACR values are specified. Again, this is an implementation detail of the
+homeserver as ACR values are not intended for introspection by clients.
+
+Finally, it is left up to the homeserver to determine how exactly the `acr_values`, `max_age` and
+`scope` parameters are used during authentication. Implementations should, however, note the
+recommendations in [RFC9470].
+
+### Cross-signing key reset
+
+Implementations MAY use this proposal to replace UIA when the client has obtained its access token
```
hughns, line 74: Problem: when the homeserver receives a request to `POST /_matrix/client/v3/keys/device_signing/upload`, how does it know whether it should respond with the `m.oauth` flow stage from MSC4312, or the new `M_INSUFFICIENT_USER_AUTHENTICATION` errcode?

We would want the server to be backwards compatible with older clients, so I think the server needs to know what the client supports.

So, basically some kind of versioning:

- query param?
- body param?
- `/_matrix/client/v3/keys/device_signing/upload` -> `/_matrix/client/v4/keys/device_signing/upload`?

zecakeh: There is also the possibility of returning both MSCs fields in the body of the response, since there are no conflicts with the keys.

A client that doesn't support this MSC should be able to handle the UIA fields, and a client that supports this MSC will prefer to handle these new fields.

proposals/4363-oauth-step-up-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+A client receiving a challenge SHOULD use the `acr_values`, `max_age` and `scope` properties to
+construct a new authorization request. For this purpose, `acr_values` and `max_age` are added to the
+[current list of request parameters] using the same definition given above. If the challenge doesn't
+include any scopes, the client SHOULD use the same scopes it used during login on the authorization
+request.
+
+Following successful authorization, the client's previous access token SHOULD be invalidated and a
```
gingershaped, line 49: I can't figure out how the homeserver is supposed to tell that an existing client is authenticating again, and not a new client. Is the client supposed to supply its access token in the OAuth request somehow?

Johennes: Hm, I think it technically doesn't need to because the `client_id` should identify the client? The AS should, however, be able to identify a possibly existing session based on cookies, I assume.

gingershaped: > Hm, I think it technically doesn't need to because the `client_id` should identify the client?


The `client_id` may not be unique for each device. Continuwuity derives the `client_id` from a hash of the client registration to avoid filling the database up with thousands of identical registrations.

> The AS should, however, be able to identify a possibly existing session based on cookies, I assume.


Continuwuity's session cookie only uniquely identifies a single login to the web dashboard; it doesn't have any relation to Matrix devices. The only thing that identifies a Matrix device is its access token.

gingershaped: I'm asking about this because MSC4484 needs some sort of temporary privilege escalation mechanism to be ergonomically usable by clients. Since registering a device is a hard requirement under the current OAuth scheme, a client that wants to use the server administration OAuth scope can't obtain a temporary privileged token without creating a useless new device. Therefore some way for the client to obtain more privileges on its existing token is necessary.

zecakeh: If it's necessary to track the device ID, then the homeserver could just add the `urn:matrix:client:device:<device_id>` scope token to the `scope` field in the error response. If there is no `scope` field in the error the MSC says that the client should use the same as during login so it should also include the device ID of the session in theory?

gingershaped: That would probably work for identifying the device, yes.

gingershaped, line 49: Should a new refresh token be issued as well, or should the old one be reused?

PR #4366, opened 2025-10-14
MSC4366: Resident servers in and around the room directory

Johennes:
[Rendererd](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/resident-public-rooms/proposals/4366-resident-public-rooms.md)

Fixes: https://github.com/matrix-org/matrix-spec/issues/1375

### DIFF
--- a/proposals/4366-resident-public-rooms.md
+++ b/proposals/4366-resident-public-rooms.md
@@ -50,8 +50,7 @@ proposal.
 ## Alternatives
 
 Non-resident rooms could be allowed in room directory responses if the server would accompany them
-with suitable `via` values. It is unclear, however, how the server would reliably determine other
-resident servers to provide these values. A future proposal might investigate this idea further.
+with suitable `via` values. This approach is covered in [MSC4367].
 
 ## Security considerations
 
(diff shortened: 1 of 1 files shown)


proposals/4366-resident-public-rooms.md
tulir, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4366-resident-public-rooms.md
```diff
@@ -0,0 +1,76 @@
...
+is called with an empty or missing `server` parameter, the server MUST only return rooms listed on
+its directory that it has at least one joined member in.
+
+Similarly, the server-server endpoints
+
+- [`GET  /_matrix/federation/v1/publicRooms`]
+- [`POST /_matrix/federation/v1/publicRooms`]
+
+MUST only include rooms from the server's local directory if the server has at least one joined
+member in them.
```
tulir, line 32: This looks like it will completely forbid directory-only servers like https://matrixrooms.info/

I think it'd be better to define a `via` field in the response entries instead of this proposal. Basically, let the origin server figure out if it can find a route rather than mandating a single allowed route in the spec. The spec could still mandate that `via`s must be returned

Johennes: I was struggling with how to best express this but I had intended it to only apply to rooms from the local directory, meaning the server MAY continue to additionally include rooms from other server's directories like today. That only half-solves the problem of course, as stated under potential issues.

Maybe we should combine both? For any room in the response the server must either be resident or include a `via` value in the response?

tulir: > For any room in the response the server must either be resident or include a `via` value in the response?

Is that any different than just requiring all entries include a `via` value, which may simply be the server itself?

Johennes: I guess not. 😄

Johennes: Have captured this in https://github.com/matrix-org/matrix-spec-proposals/pull/4367.

PR #4367, opened 2025-10-14
MSC4367: via routes in the published room directory

Johennes:
[Rendered](https://github.com/Johennes/matrix-spec-proposals/blob/johannes/public-rooms-via/proposals/4367-public-rooms-via.md)

Fixes: https://github.com/matrix-org/matrix-spec/issues/1375

### DIFF
--- a/proposals/4367-public-rooms-via.md
+++ b/proposals/4367-public-rooms-via.md
@@ -6,9 +6,12 @@ resident server to facilitate the join. The spec doesn't make it clear how to fi
 however.
 
 To begin with, it is unclear whether the endpoints for listing a server's directory may return rooms
-that the server is not joined to. Additionally, there is no recommendation on what `via` parameter
-clients should use when attempting to join rooms found in the directory. As a result, clients can
-end up browsing published rooms without being able to join them.[^1]
+that the server is not joined to. One example where including such rooms is desirable are
+directory-only servers like [matrixrooms.info].
+
+Additionally, there is no recommendation on what `via` parameter clients should use when attempting
+to join rooms found in the directory. As a result, clients can end up browsing published rooms
+without being able to join them.[^1]
 
 This proposal addresses the situation by adding `via` values in the room directory responses.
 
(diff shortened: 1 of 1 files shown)


proposals/4367-public-rooms-via.md
tulir, line 1: Implementation requirements:

* [ ] Server that returns `via` in both APIs, and passes it through in the C-S API when querying a remote server
* [ ] Client that uses the returned `via` for `/join` calls

velikopter: Implemented into Vona: https://foundry.fsky.io/vel/matrix-vona/src/branch/master/vona/globals/__init__.py#L258-L272

proposals/4367-public-rooms-via.md
```diff
@@ -0,0 +1,75 @@
+# MSC4367: `via` routes in the published room directory
+
+The [published room directory] allows clients to discover rooms on their own or other homeservers.
+Naturally, clients will want to join rooms found in the directory which requires knowledge of a
+resident server to facilitate the join. The spec doesn't make it clear how to find such a server,
+however.
+
+To begin with, it is unclear whether the endpoints for listing a server's directory may return rooms
+that the server is not joined to. Additionally, there is no recommendation on what `via` parameter
```
zecakeh: This first sentence of this paragraph is not addressed by the proposal either, maybe it should? Like stating that the spec should be updated to say that the room directory can include rooms that the server is not joined to. Otherwise it's hard to see a reason for adding the `via` parameter.

Also I couldn't see a use case for this at first, until I read [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4366/files#r2428543345), so maybe it should be added as an example as a justification for this proposal?

Johennes: > This first sentence of this paragraph is not addressed by the proposal either, maybe it should? Like stating that the spec should be updated to say that the room directory can include rooms that the server is not joined to. Otherwise it's hard to see a reason for adding the `via` parameter.

You're right. Have tried to make it more explicit with 6987a9d2f8d3f5217972f8ca96ad9bf2636dd05a.

> Also I couldn't see a use case for this at first, until I read [this thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4366/files#r2428543345), so maybe it should be added as an example as a justification for this proposal?

Good idea. Have added this with d4c5696855439004dc259e470affcec8559d62b3.

PR #4368, opened 2025-10-15
MSC4368: Combine definitions of M_RESOURCE_LIMIT_EXCEEDED error code and m.server_notice.usage_limit_reached server notice type

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/resource-limit-exceeded-extra-fields/proposals/4368-resource-limit-exceeded-extra-fields.md)

Implementations:

- Server: https://github.com/element-hq/synapse/blob/develop/synapse/api/errors.py#L489-L513
- Client: https://github.com/element-hq/element-web/commit/9bde468e303bbf7cebbb25fa83b76ad59e8b4e83

---

I am employed by Element to write this MSC. Hence the use of the `io.element` namespace on the unstable values.

### DIFF
--- a/proposals/4368-resource-limit-exceeded-extra-fields.md
+++ b/proposals/4368-resource-limit-exceeded-extra-fields.md
@@ -1,9 +1,22 @@
-# MSC4368: Add limit_type to M_RESOURCE_LIMIT_EXCEEDED error code
+# MSC4368: Combine definitions of M_RESOURCE_LIMIT_EXCEEDED error code and m.server_notice.usage_limit_reached server notice type
+
+When a homeserver has exceeded some kind of resource limit it can currently communicate the fact to the client and user
+using two distinct mechanism:
+
+* the [`M_RESOURCE_LIMIT_EXCEEDED`] error code returned on the affected API endpoint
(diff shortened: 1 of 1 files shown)


proposals/4368-resource-limit-exceeded-extra-fields.md
```diff
@@ -0,0 +1,50 @@
+# MSC4368: Add limit_type to M_RESOURCE_LIMIT_EXCEEDED error code
+
+## Proposal
+
+1. Add an optional `limit_type` field to the body of the [`M_RESOURCE_LIMIT_EXCEEDED`] error response that has the same
```
turt2live: I was confused at first because this MSC appears to be describing what's already in the spec (https://spec.matrix.org/v1.16/client-server-api/#events-20), but it's actually trying to copy existing behaviour to another area of the spec - words to clarify what is going on would be good.

hughns: I've tried to make this clearer in ff78bc5dea64590485b7bb5e00c3f225e2473915.

PR #4369, opened 2025-10-17
MSC4369: M_CAPABILITY_NOT_ENABLED error code for when capability is not enabled on an API endpoint

hughns:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hughns/capability-error-code/proposals/4369-capability-error-code.md)

---

I am employed by Element to write this MSC. Hence the use of the `io.element` namespace in the unstable values.

### DIFF
--- a/proposals/4369-capability-error-code.md
+++ b/proposals/4369-capability-error-code.md
@@ -60,14 +60,14 @@ section isn't used in the wild.
 
 ## Alternatives
 
-### Codify existing Synapse behaviour
+### Codify existing Synapse behaviour
 
 We could simply codify what is in the existing Synapse implementation (see above). This is what was originally proposed
 in https://github.com/matrix-org/matrix-spec/pull/2212.
 
 However, that doesn't consider non-Synapse implementations.
 
-### Increment version on affected endpoints
+### Increment version on affected endpoints
 
 Deprecate the old version and introduce a new version with the updated error responses.
 


proposals/4369-capability-error-code.md
```diff
@@ -0,0 +1,132 @@
...
+
+I believe this is an omission in the current spec that should be addressed for existing capabilities
+and be checked for future MSCs that use capabilities.
+
+For background: this MSC originates from a
+[spec PR](https://github.com/matrix-org/matrix-spec/pull/2212) that has since been closed.
+
+## Proposal
+
+Define a new [other error code] of `M_CAPABILITY_NOT_ENABLED`.
```
zecakeh, line 17: Is there any value in having a specific error code for this? Clients should already know which endpoint is behind a capability, so `M_FORBIDDEN` should work just as well.

Actually I am wondering if instead of returning `403`, it wouldn't be simpler to return `404` with `M_UNRECOGNIZED`, as if the endpoint is unimplemented?

This means that the endpoint doesn't actually have to be implemented by the homeserver if it doesn't support this capability (like `POST /_matrix/client/v3/account/password` which is useless if the homeserver only supports the OAuth 2.0 API for authentication), rather than forcing the homeserver to support this endpoint, only to return a specific error.

anoadragon453:
@hughns apologies for having this sit on my queue for so long.

Is this still something you'd like to pursue?

hughns:
> Is this still something you'd like to pursue?

I am motivated to resolve this lack of clarity - around what the server should return if the capability is not available but is attempted anyway - in the spec one way or another.

I'm not opposed to @zecakeh's [comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4369#discussion_r2498901682) about re-using existing codes if that feels like a better fit.

PR #4370, opened 2025-10-17
MSC4370: Federation endpoint for retrieving current extremities

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/prev-events-fed/proposals/4370-prev-events-federation-endpoint.md)

### DIFF
--- a/proposals/4370-prev-events-federation-endpoint.md
+++ b/proposals/4370-prev-events-federation-endpoint.md
@@ -68,7 +68,7 @@ Though not primary drivers for this proposal, this MSC's new endpoint also enabl
 ## Proposal
 
 A new endpoint is added to the [Server-Server API](https://spec.matrix.org/v1.16/server-server-api/)
-to retreive the current extremities (as event IDs) the server has in the given room. The endpoint
+to retrieve the current extremities (as event IDs) the server has in the given room. The endpoint
 requires normal authentication for federation endpoints and MAY be rate limited. "Current" is at the
 time of the request and has no guarantee of being accurate after the request is completed.
 


proposals/4370-prev-events-federation-endpoint.md
turt2live, line 1: Implementation requirements:

* [ ] Server (calling)
* [x] Server (serving)

turt2live: I'd recommend waiting to implement this MSC until it has a reasonably healthy amount of review given the security considerations.

1vel:
This doesn't seem to consider v1 or v2 rooms, which have `prev_events` laid out differently.

proposals/4370-prev-events-federation-endpoint.md
```diff
@@ -0,0 +1,141 @@
...
+* 403 / `M_FORBIDDEN` - The requesting server is not in the room, or is excluded from the room via
+  `m.room.server_acl`.
+* 404 / `M_NOT_FOUND` - The server is not aware of the room.
+
+Servers SHOULD NOT attempt to fully resolve the DAG before returning a response - the calling server
+is expected to handle the extremities as needed. For minimal/lightweight policy server implementations,
+this may mean requesting state at each event and merging the results for a good enough representation
+of current state - specific handling is left as an implementation detail.
+
+## Potential issues
```
turt2live, line 110: @velikopter [says](https://github.com/matrix-org/matrix-spec-proposals/pull/4370#issuecomment-3417564934):

> This doesn't seem to consider v1 or v2 rooms, which have `prev_events` laid out differently.

(please use threads to ensure comments are visible and can be replied to)

turt2live: in all honesty, I'm not concerned about v1 or v2 rooms at this point in our history. If a server *really* wants to support those room versions, it can call this endpoint then the [`GET /event/:eventId`](https://spec.matrix.org/v1.16/server-server-api/#get_matrixfederationv1eventeventid) endpoint to get the hash, then wedge that tuple into whatever it needs to.

proposals/4370-prev-events-federation-endpoint.md
turt2live: This MSC has not had the healthy amount of review recommended, but I've started an implementation anyway: https://github.com/element-hq/synapse/pull/19314

I intend to keep that PR in a draft state until MSC review can happen, disqualifying it as an implementation for now.

proposals/4370-prev-events-federation-endpoint.md
```diff
@@ -0,0 +1,141 @@
+# MSC4370: Federation endpoint for retrieving current extremities
```
kegsay, line 1: Overall, I approve of this MSC. We need something like this in order to ensure eventual delivery in Matrix. [MSC4242](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/placeholder-1/proposals/4242-state-dags.md#soft-failure) actually talks about this:

> In current Matrix, eventual delivery of events is only ensured under the assumption that any server continuously sends new events. The additional assumption is needed because homeservers do not eagerly share known forward extremities with each other: it relies on servers sending events. As such, you can have two servers who disagree on the current state because one server is aware of events that affect the current state and the other is not. This is less of a problem than the soft-failure case because it will correct itself when an event is sent, but the system requires some external input to provide eventual delivery, in order to eventually converge. This will be addressed in a future MSC.

Having an endpoint which is compatible with those aims as well as policy server's aims would be helpful.

That being said.. what exactly does the policy server _do_ with the information it gets from this new endpoint? The proposal as written is unclear:

> so it may then retrieve the current state for the room

> that server can request (and cache) current state whenever it feels it needs to rather than waiting for a remote server to become active.

There is no such endpoint to do this today. The closest is [`GET /_matrix/federation/v1/state/{roomId}`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1stateroomid) but this only accepts a single event ID. It is not secure to just pick a random prev event and use that state, nor is it secure to just merge individual states together without doing state resolution.

turt2live: In policyserv, there's a concept of "trusted" servers that it's willing to learn state from. Currently, that state learning happens when policyserv receives an event from one of those trusted servers. It uses that event's ID to make a `/state` request if it hasn't updated its room state in a while. This can cause its state information to be old if the trusted server doesn't send an event for a while, which isn't great. 

The hope is to keep that same "call `/state` with an event ID" logic, but using the trusted server's extremities. If there's multiple extremities, the policy server can send an event, wait a bit, then learn state off its own event ID. This does mean that a server becomes trusted to not only supply resolved state, but also trusted to be moderately up to date on the room to provide recent extremities. 

This isn't perfectly secure because yes, the state can and does change depending on how you look at the DAG, but `/state` is supposed to do that state resolution to some degree, and the policy server in this case isn't actually that concerned with it being a perfect representation of state anyway. If it gets "wrong" state, it will mean the cache of some state events is a little broken for a bit. In policyserv, that might mean not recognizing a recently-promoted admin as an admin or missing a policy rule from a moderation bot - both of these are easily correctable the next time the cron job runs.

Aside from that, the policy server can also use the extremities from a (trusted) server to send its own redactions and other events without having to bear the burden of the DAG itself.

proposals/4370-prev-events-federation-endpoint.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+(there are no query parameters or request body)
+
+Response shape:
+
+```
+Content-Type: application/json
+{
+  "prev_events": ["$event1", "$event2"]
```
reivilibre, line 92: Must this contain ALL prev_events, even if there are thousands of them? Millions? There is unfortunately no limit to how many forward extremities you can have.

proposals/4370-prev-events-federation-endpoint.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+The response is contained in an object for future expansion. `prev_events` is required, and MUST have
```
reivilibre, line 96: (Minor: Is `prev_events` a good name for this, or is it a Synapsism? They aren't really 'previous' to anything yet. I'm not fussy though.)

turt2live: The theory is someone could just copy the object on top of an event template, though that's not exactly a thing we *need* to support.

proposals/4370-prev-events-federation-endpoint.md
turt2live: it should be "whatever you'd need to send an event", which I think is limited in some way. Will clarify in the MSC if it's not already there.

proposals/4370-prev-events-federation-endpoint.md
turt2live: There's been some review on this, so the implementation was put forward for review and has now landed. Next steps are to get it implemented in another server, probably policyserv.

PR #4373, opened 2025-10-22
MSC4373: Server opt-out of specific EDU types

1vel:
[Rendered](https://github.com/velikopter/matrix-spec-proposals/blob/edutypes/proposals/4373-edu-types.md)

Implementations:
- Server (setting value):
  - Continuwuity: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1137
  - Calyx: https://foundry.fsky.io/vel/calyx/src/commit/5ef3ca34986202931b4b73a53d026f524008947e/synapse/federation/transport/server/federation.py#L879
- Server (reading value):

### DIFF
--- a/proposals/4373-edu-types.md
+++ b/proposals/4373-edu-types.md
@@ -34,7 +34,7 @@ If the EDU type is listed, and is set to `false`, any EDUs of that type
 SHOULD NOT be sent to the target homeserver. The target server MUST ignore
 EDUs of these types.
 
-If the EDU type is set to `true` or absent, the EDU SHOULD be sent to the target 
+If the EDU type is set to `true` or absent, the EDU MAY be sent to the target 
 homeserver, unless other factors disallow it (such as room ACLs, where that is
 relevant).
 
--- a/proposals/4373-edu-types.md
+++ b/proposals/4373-edu-types.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Server opt-out of specific EDU types
+# MSC4373: Server opt-out of specific EDU types
 
 Some servers may wish to not receive specific types of EDUs, such as presence, to
 cut down on the amount of bandwidth used (as an example).
--- a/proposals/4373-edu-types.md
+++ b/proposals/4373-edu-types.md
@@ -36,6 +36,8 @@ SHOULD NOT be sent to the target homeserver.
 If the EDU type is set to `true`, the EDU MAY be sent to the target homeserver,
 unless other factors disallow it (such as room ACLs, where that is relevant).
 
+The endpoint response SHOULD be cached, for a week at maximum.
+
 ## Potential issues
 
 Older homeservers (or simply non-compliant ones) will still send unwanted EDUs,
--- a/proposals/4373-edu-types.md
+++ b/proposals/4373-edu-types.md
@@ -14,15 +14,15 @@ The response for this endpoint is shaped like the following:
 ```json
 {
     "m.presence": false,
-    "m.typing": true,
-    "m.read": true
+    "m.receipt": true,
+    "m.typing": true
 }
 ```
 
 The allowed types are:
 * `m.presence`
+* `m.receipt`
 * `m.typing`
-* `m.read`
 
 Other types of EDUs (signing key updates, device lists, to-device messaging, etc)
 are likely unsafe to opt-out of.
--- a/proposals/4373-edu-types.md
+++ b/proposals/4373-edu-types.md
@@ -27,7 +27,7 @@ The allowed types are:
 Other types of EDUs (signing key updates, device lists, to-device messaging, etc)
 are likely unsafe to opt-out of.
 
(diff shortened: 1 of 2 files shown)


proposals/4373-edu-types.md
turt2live, line 1: Implementation requirements:

* [ ] Server (setting value)
* [ ] Server (reading value)

proposals/4373-edu-types.md
```diff
@@ -0,0 +1,70 @@
...
+This endpoint SHOULD NOT require authentication, but if provided, follow the normal
+verification process.
+
+If the EDU type is listed, and is set to `false`, any EDUs of that type
+SHOULD NOT be sent to the target homeserver. The target server MUST ignore
+EDUs of these types.
+
+If the EDU type is set to `true` or absent, the EDU SHOULD be sent to the target 
+homeserver, unless other factors disallow it (such as room ACLs, where that is
+relevant).
```
Johennes: If the receiving server can decide what to receive, could the sending server also have an interest to decide what to send? E.g. a server might want to opt out of sending presence for its users over federation. I wonder if this would warrant a MAY rather than a SHOULD?

Johennes: A MAY would also make the new endpoint sort of equivalent to the `/capabilities` endpoint in the CS API.

1vel: It does, I'll change it to a MAY

proposals/4373-edu-types.md
```diff
@@ -0,0 +1,70 @@
+# MSC4373: Server opt-out of specific EDU types
+
+Some servers may wish to not receive specific types of EDUs, such as presence, to
+cut down on the amount of bandwidth used (as an example).
+
+## Proposal
+
+### `GET /_matrix/federation/v1/edutypes`
```
jevolk: Would this be better organized under `/_matrix/federation/query/` `/edutypes` instead of a top-level ep?

proposals/4373-edu-types.md
```diff
@@ -0,0 +1,70 @@
...
+defaulting to allowing filtering of all EDU types, which is bad for the reasons
+described above.
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+`/_matrix/federation/unstable/io.fsky.vel/edutypes`
```
1vel, line 66: I am well aware this makes the proposal unready for FCP based on the "does not pollute any global namespace", however implementations already exist, so I do not want to make any breaking changes.

proposals/4373-edu-types.md
```diff
@@ -0,0 +1,70 @@
+# MSC4373: Server opt-out of specific EDU types
+
+Some servers may wish to not receive specific types of EDUs, e.g. presence, to cut down
+on bandwidth usage.
+
+## Proposal
+
+### `GET /_matrix/federation/v1/query/edutypes`
```
timedoutuk, line 8: After having this sit for some time, I think a better approach might be modifying the response body of `/_matrix/federation/v1/send`. Something like adding a key of `"ignored_edu_types": ["m.presence", "..."]`, although not sure how easy that change would be considering the current body is expected to be a mapping of event IDs to error objects.

The benefit of this is that servers are already constantly exchanging transactions, so this would avoid that additional roundtrip and could even ensure caches remain fresh more accurately. The only downside (other than the potential key conflict) I can think of is that this means servers may end up sending unwanted EDUs in a transaction if it's their first or their cache is stale, but considering the proposal also states that servers MUST ignore unwanted EDUs, I don't think this is a problem.

My other thought was putting this info in well-known, as it would be a rather apt place to put that info, but unfortunately not everyone uses well-known, and requiring well-known to advertise preferences probably wouldn't go down well.

proposals/4373-edu-types.md
1vel: (seeing other MSCs get merged while polluting a global namespace makes me think this doesn't really matter.)

proposals/4373-edu-types.md
turt2live, line 1: This looks a bit too close to a capabilities structure, which starts to invoke fear that the spec becomes too optional. The cited reason of reducing bandwidth is a noble one, but the traffic itself is largely insignificant compared to other features of Matrix. We should also prefer to fix presence specifically if we're concerned about the bandwidth of it.

Generally speaking, dropping EDUs on the floor is acceptable (and what some servers already do).

There's also the issue of how often should the capability be checked by remote servers? For some use cases it's likely a reasonably permanent solution, but others turn them off for a day or so to reduce load during abnormally high traffic periods. Finding an interval which respects both use cases might very well cause close to the same amount of traffic to check if an EDU should be sent as just sending the EDU and letting it be ignored.

As written, I would be looking for a stronger use case before accepting this proposal myself. Other SCT members may have different opinions as individuals.

turt2live (commented):
Sorry this has sat in the review queue for so long. I've emitted some thoughts as an individual on the SCT.

PR #4199, opened 2024-09-24
Mandate a 'Security Considerations' section on MSCs

kegsay:
And link to lists of possible problems to think about. This is part of an effort to improve the overal security of Matrix during the design process.

PR #4378, opened 2025-11-11
Proposal template: explicitly mention RFC 3552 in security section

codedust:
Just a small addition to https://github.com/matrix-org/matrix-spec-proposals/pull/4199: Let's not only mention [RFC3552](https://datatracker.ietf.org/doc/html/rfc3552) in `MSC_CHECKLIST.md` but also in `proposals/0000-proposal-template.md`

### DIFF
--- a/proposals/0000-proposal-template.md
+++ b/proposals/0000-proposal-template.md
@@ -87,7 +87,7 @@ idea.
 
 **All proposals must now have this section, even if it is to say there are no security issues.**
 
-*Think about how to attack your proposal. See [RFC3552](https://datatracker.ietf.org/doc/html/rfc3552)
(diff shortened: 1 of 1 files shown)


proposals/0000-proposal-template.md
```diff
@@ -87,7 +87,8 @@ idea.
 
 **All proposals must now have this section, even if it is to say there are no security issues.**
 
-*Think about how to attack your proposal, using lists from sources like
+*Think about how to attack your proposal. See [RFC3552](https://datatracker.ietf.org/doc/html/rfc3552)
```
turt2live:
```suggestion
*Think about how to attack your proposal. See [RFC 3552](https://datatracker.ietf.org/doc/html/rfc3552)
```

turt2live (approved):

PR #4380, opened 2025-11-18
MSC4380: Invite blocking

richvdh:
An alternative to MSC4155.

Written wearing my Spec Core Team hat, in an attempt to unblock the situation that `matrix.org` and Element-Web have implemented an unspecced MSC which we can't turn off.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/invite_blocking/proposals/4380-invite-blocking.md)

Server implementation: https://github.com/element-hq/synapse/pull/19203
Client implementation: https://github.com/element-hq/element-web/pull/31268

---- 

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4380#issuecomment-3603050579)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4380#issuecomment-3549308609)

### DIFF
--- a/proposals/4380-invite-blocking.md
+++ b/proposals/4380-invite-blocking.md
@@ -39,6 +39,13 @@ is invited:
 Any rejected requests should result in an HTTP 403 status code, with the Matrix
 error code `M_INVITE_BLOCKED`. This is a new error code.
 
+Note: in the case of `POST /createRoom`, it is possible for one of the invited users
+to be rejected whilst the room creation as a whole succeeds. This is an
+[existing problem](https://github.com/matrix-org/matrix-spec/issues/1951) in
+Matrix which we do not attempt to resolve here; we specify only that
+implementations should handle such rejections similarly to other failed
+invites.
+
 In addition, existing events already in the database MUST NOT be served over client synchronisation endpoints such as
 [`GET
 /_matrix/client/v3/sync`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync)


proposals/4380-invite-blocking.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4380-invite-blocking.md
```diff
@@ -0,0 +1,103 @@
...
+ * [`PUT /_matrix/federation/v1/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+ * [`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv2inviteroomideventid)
+ * [`POST /_matrix/client/v3/rooms/{roomId}/invite`](https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3roomsroomidinvite)
+ * [`POST /_matrix/client/v3/createRoom`](https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3createroom)
+   (checking the invite list)
+ * [`PUT /_matrix/client/v3/rooms/{roomId}/state/m.room.member/{stateKey}`](https://spec.matrix.org/v1.15/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+   (for invite membership)
+
+Any rejected requests should result in an HTTP 403 status code, with the Matrix
+error code `M_INVITE_BLOCKED`.
```
turt2live:
```suggestion
error code `M_INVITE_BLOCKED`. This is a new error code.
```

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4380-invite-blocking.md
richvdh: Server implementation: https://github.com/element-hq/synapse/pull/19203
Client implementation: https://github.com/element-hq/element-web/pull/31268

turt2live: implementation looks sufficient to me

proposals/4380-invite-blocking.md
```diff
@@ -0,0 +1,106 @@
...
+   (for invite membership)
+
+Any rejected requests should result in an HTTP 403 status code, with the Matrix
+error code `M_INVITE_BLOCKED`. This is a new error code.
+
+In addition, existing events already in the database MUST NOT be served over client synchronisation endpoints such as
+[`GET
+/_matrix/client/v3/sync`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync)
+or `/v4/sync` from
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4158). Invite
```
marwing:
```suggestion
[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186). Invite
```

richvdh: thanks

richvdh:
Per fbdeb55, I decided to switch from `block_all: true|false` to `default_action: "block"|"allow"`, mostly to allow easier extension in future to `"block"|"allow"|"ignore"`.

richvdh:
I claim this is ready for FCP.

@mscbot fcp merge

proposals/4380-invite-blocking.md
```diff
@@ -0,0 +1,112 @@
...
+error code `M_INVITE_BLOCKED`. This is a new error code.
+
+In addition, existing events already in the database MUST NOT be served over client synchronisation endpoints such as
+[`GET
+/_matrix/client/v3/sync`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3sync)
+or `/v4/sync` from
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186). Invite
+events received over federation should likewise not be served over `/sync` requests.
+
+Servers MAY return any suppressed invite events over `/sync` if invite
```
uhoreg, line 56: I'm a bit confused.  If `PUT /_matrix/federation/v?/invite/{roomId}/{eventId}` returns an error, then the invite event won't be signed by the invited server, so won't be sent out to other servers, because the `/invite/{roomId}/{eventId}` endpoints say that the inviting server doesn't send out the invite until it's signed by the invited server.  Which ... I would expect would mean that it can't yet be part of the room state, but looking through the auth rules I don't actually see a rule saying that invites will be rejected if they aren't signed by the invited server.

So I'm a bit confused about what the expected behaviour is.  What I would expect is:

- the inviting user calls `POST /_matrix/client/v3/rooms/{roomId}/invite` or similar
- the inviting server calls `PUT /_matrix/federation/v1/invite/{roomId}/{eventId}` on the invited server
- the invited server returns `M_INVITE_BLOCKED`
- the inviting server passes that `M_INVITE_BLOCKED` error to the inviting user
- the inviting and invited server both drop the blocked invite, and treat the room as if the invite never happened

But from this sentence it sounds like this isn't the case?  It sounds like what actually happens is:

- the inviting user calls `POST /_matrix/client/v3/rooms/{roomId}/invite` or similar
- the inviting server calls `PUT /_matrix/federation/v1/invite/{roomId}/{eventId}` on the invited server
- the invited server returns `M_INVITE_BLOCKED`
- (not sure what the inviting server returns to the inviting user)
- the inviting server still adds the event to its copy of the room?
- the invited server server can remember the invite and serve it to the user later on if blocking gets disabled?
- not sure if/when other servers find out about the invite event?

It could be that I'm missing something, but this sentence (and the next paragraph) suggests that invite blocking works differently from what I expected it to do.

tulir: I think you could hypothetically receive an invite event that you (or someone with your federation keys) signed previously?

Other than that, invite events that are already in the database are suppressed according to the MSC, so the second sentence probably lets servers resend such suppressed events when invite blocking is disabled.

proposals/4380-invite-blocking.md
mehdiamiri9902-spec: spaces

proposals/4380-invite-blocking.md
```diff
@@ -0,0 +1,112 @@
...
+[`m.ignored_user_list`](https://spec.matrix.org/v1.15/client-server-api/#mignored_user_list)).
+
+When invites to a given user are blocked, the user's
+homeserver MUST respond to the following endpoints with an error, if the user
+is invited:
+
+ * [`PUT /_matrix/federation/v1/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv1inviteroomideventid)
+ * [`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.15/server-server-api/#put_matrixfederationv2inviteroomideventid)
+ * [`POST /_matrix/client/v3/rooms/{roomId}/invite`](https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3roomsroomidinvite)
+ * [`POST /_matrix/client/v3/createRoom`](https://spec.matrix.org/v1.15/client-server-api/#post_matrixclientv3createroom)
```
dbkr, line 34: Presumably in this case, the room would also not be created? Also, I think this is the only endpoint that. allows inviting multiple users at once, so does it have a way to signal which matrix ID was blocked?

tulir: Invites can only be sent after the room is created, so currently any invite blocking will result in the room being created, but some or all invites not being sent

richvdh: yup, there are a number of bugs around this, such as https://github.com/element-hq/synapse/issues/8895 and https://github.com/matrix-org/matrix-spec/issues/1951, and I don't plan to fix that today.

proposals/4380-invite-blocking.md
richvdh: yup, this is all about events that were created (and signed) before the invite block was in place.

proposals/4380-invite-blocking.md
dbkr: OK, but the spec says the endpoint must respond with an error, which it sounds like it doesn't if the room gets created? I'm not suggesting fixing the impls but I think the spec should be clear in what it requires.

dbkr:
@mscbot concern behaviour of room create endpoint

proposals/4380-invite-blocking.md
tulir: Current implementations will create the room, then respond with an error. Creation is not atomic, so the endpoint returning an error doesn't mean creation failed. The client will discover the empty room in the next /sync response.

richvdh: I'm *somewhat* reluctant to spec exactly the current behaviour, because it's clearly such a mess.

dbkr: Yeah, I agree it makes no sense to spec that, but this doesn't really feel up to scratch to go in the spec as it is? If anything it feels like the endpoint should *not* return an error if it did create the room? Then a later enhancement can be some way of indicating that some users weren't invited?

richvdh: >  If anything it feels like the endpoint should not return an error if it did create the room?

Hrm, arguably. That doesn't sound great either: at least, if there is an error, you get an indication that not all was happy.

My inclination is to keep this consistent with the existing behaviour when an invite fails (e.g.: because the receiving server is offline); I'd be happy to add those words to the MSC. It still sucks, but at least it sucks consistently.

dbkr: Fair enough, main concern here is that the client will see error and tell the user the room creation didn't work, but I'm favour of consistency so that seems like a good compromise.

turt2live:
The bot isn't aware that FCP is paused when a concern is raised during the countdown. This still has 5 days of FCP to go once the concern is cleared.

richvdh:
> This still has 5 days of FCP to go once the concern is cleared.

As per https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issuecomment-3800789985, I don't think that's correct.

turt2live:
*0 days remaining. https://github.com/matrix-org/matrix-spec-proposals/pull/4335#issuecomment-3800814676

proposals/4380-invite-blocking.md
richvdh: @dbkr I've added a paragraph about this in 3b38d48. Does it address your concern?

dbkr: Thanks, I think that does help significantly.

dbkr:
@mscbot resolve concern behaviour of room create endpoint

dbkr:
@mscbot resolve behaviour of room create endpoint

zecakeh:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2305

turt2live:
Merged 🎉

PR #4382, opened 2025-11-20
MSC4382: Peppered hash verification for E2EE content moderation

seuros:
[Rendered](https://github.com/seuros/matrix-spec-proposals/blob/peppered-hash-verification/proposals/4382-peppered-hash-verification.md)

This proposal adds a `verification_hash` field to encrypted events to enable cryptographic verification of reported content without breaking E2EE.

Uses SHA-256(plaintext || ciphertext_hash) construction where the ciphertext hash acts as a unique per-message pepper, preventing rainbow tables and mass surveillance while enabling moderation of reported messages.

### DIFF
--- a/proposals/4382-peppered-hash-verification.md
+++ b/proposals/4382-peppered-hash-verification.md
@@ -67,20 +67,31 @@ receiving encrypted events.
 When a user reports encrypted content, the server verifies:
 
 ```python
-plaintext_event = canonical_json(report['plaintext'])
-ciphertext = event['content']['ciphertext']
+# Server receives report with event_id and claimed plaintext
+report = receive_report(room_id, event_id)
 
+# Fetch stored encrypted event from database (local operation)
+event = database.get_event(event_id)
+
+# Extract verification data (all already stored locally)
+ciphertext = event['content']['ciphertext']
+verification_hash = event['content']['verification_hash']
+
+# Compute hash of reported plaintext with stored ciphertext
+plaintext_event = canonical_json(report['plaintext'])
 computed = base64(sha256(plaintext_event + ciphertext))
 
-if computed == event['content']['verification_hash']:
+# Single comparison - no decryption, no network requests
+if computed == verification_hash:
     # Report verified - plaintext is authentic
 else:
     # Report is false - reporter is lying
 ```
 
 The server never needs decryption keys or access to the encryption
-session. It only verifies that the reported plaintext matches the
-verification hash.
+session. All verification data (ciphertext and verification_hash) is
+already stored in the database. Verification requires only a local
+database fetch and a single SHA-256 computation.
 
 ### Security Properties
 
@@ -147,6 +158,7 @@ capabilities.
 
 During development, implementations should use:
 - Field name: `org.matrix.msc4382.verification_hash`
+- Report field: `org.matrix.msc4382.plaintext`
 - Client capability: `org.matrix.msc4382.peppered_hash_verification`
 
 ## Dependencies


proposals/4382-peppered-hash-verification.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+
+Add a `verification_hash` field to `m.room.encrypted` events:
+
+```
+verification_hash = SHA-256(plaintext || ciphertext_hash)
+```
+
+Where:
+- `plaintext` is the unencrypted message body
+- `ciphertext_hash` is the existing `hashes.sha256` field value
```
uhoreg: `hashes.sha256` is probably not what you want.  `hashes.sha256` is calculated by taking the redacted content, which means that it won't take into account things `content.ciphertext`.  Also, it is not something that the client can calculate, since the client does not have access to some of the information that is needed to calculate it.

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+## Proposal
+
+Add a `verification_hash` field to `m.room.encrypted` events:
+
+```
+verification_hash = SHA-256(plaintext || ciphertext_hash)
+```
+
+Where:
+- `plaintext` is the unencrypted message body
```
uhoreg: Presumably the message body here is encoded using canonical JSON

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+  "sender": "@alice:example.org",
+  "content": {
+    "algorithm": "m.megolm.v1.aes-sha2",
+    "ciphertext": "AwgAEtAB...",
+    "session_id": "SESSION_ID"
+  },
+  "hashes": {
+    "sha256": "k13N8DRLpSIaP9l8..."
+  },
+  "verification_hash": "f9d8c7b6a5e4d3c2...",
```
uhoreg: Since this is something set by the client, it should be within `content`.  Clients do not have a way to set properties outside `content`.

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+When users report problematic content in encrypted rooms, server administrators have no way to verify that the reported plaintext matches what was actually sent. Reporters can claim an encrypted message contains policy violations when it does not, enabling false reporting and coordinated brigading attacks.
+
+This proposal adds a `verification_hash` field to encrypted events that allows servers to cryptographically verify reported content without requiring decryption keys or breaking E2EE privacy guarantees.
+
+## Proposal
+
+Add a `verification_hash` field to `m.room.encrypted` events:
+
+```
+verification_hash = SHA-256(plaintext || ciphertext_hash)
```
uhoreg: presumably the sha-256 gets base64-encoded, as is done elsewhere with binary data

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Allows targeted verification**: Servers can verify suspected content in specific messages when they have reason to investigate (the intended moderation use case).
+
+**Does not enable mass surveillance**: Checking if a keyword exists requires testing it against every individual message's unique pepper, which does not scale.
+
+## Potential Issues
+
+**Backwards compatibility**: Existing clients don't send `verification_hash`. The field should be optional during transition, with servers falling back to social trust when absent. During the unstable period, use `org.matrix.msc4382.verification_hash`.
+
+**Event size increase**: Adds 32 bytes per encrypted event (~10% overhead for typical messages). This is considered acceptable for the moderation benefit provided.
```
uhoreg: If you're going to give actual numbers: 43 bytes for base64-encoded data, plus the field name and delimiters.

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
+# MSC: Peppered Hash Verification for End-to-End Encrypted Content Moderation
+
+When users report problematic content in encrypted rooms, server administrators have no way to verify that the reported plaintext matches what was actually sent. Reporters can claim an encrypted message contains policy violations when it does not, enabling false reporting and coordinated brigading attacks.
```
uhoreg: editorial comment: please wrap your lines to around 80-120 characters

proposals/4382-peppered-hash-verification.md
uhoreg: I'm also assuming that when you say "unencrypted message body", you mean the plaintext event that gets fed into the encryption algorithm, i.e. the

```
{
  "type": "<event_type>",
  "content": "<event_content>",
  "room_id": "<the room_id>"
}
```

as given in https://spec.matrix.org/unstable/client-server-api/#mmegolmv1aes-sha2

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,91 @@
...
+    "sha256": "k13N8DRLpSIaP9l8..."
+  },
+  "verification_hash": "f9d8c7b6a5e4d3c2...",
+  "signatures": { ... }
+}
+```
+
+### Verification Process
+
+When a user reports encrypted content, they provide the event ID and claimed plaintext. The server verifies:
```
uhoreg: Presumably this is in reference to https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3roomsroomidreporteventid in which case you need to update the endpoint to add the reported plaintext

proposals/4382-peppered-hash-verification.md
```diff
@@ -0,0 +1,157 @@
...
+plaintext_event = canonical_json(report['plaintext'])
+ciphertext = event['content']['ciphertext']
+
+computed = base64(sha256(plaintext_event + ciphertext))
+
+if computed == event['content']['verification_hash']:
+    # Report verified - plaintext is authentic
+else:
+    # Report is false - reporter is lying
+```
```
Gnuxie, line 89: At the moment this ciphertext "pepper" is not helping and is salt because it is public and we can reproduce the hash and start making guesses.

Instead a secret pepper should be embedded in the plaintext event content, and that way it can only be accessed when you forward the entire plaintext event in a report (or decrypt the event).

seuros: The Problem I'm Solving:

I run servers where users get brigaded,  99 reports against one person based on their identity or something they said in a private room. As a server admin, I face an impossible choice:
  - Believe the mass (potentially unjust)
  - Ask the reported user for their private keys (unacceptable)
  - Ignore reports entirely (enables actual abuse)


Simple `hash(plaintext)` enables rainbow tables.

With unique ciphertext per message, checking if "buy crypto" appears in 1 million messages requires 1 million fresh hash computations. This doesn't scale for surveillance but works perfectly for verification when I receive a report with claimed plaintext.

When I receive a report, I have:
  - event_id (which message is being reported)
  - claimed_plaintext (what reporter says it says)

I fetch from my database:
  - ciphertext (already stored in the encrypted event)
  - verification_hash (already stored in the encrypted event)

Then I compute: hash(claimed_plaintext || ciphertext)

If it matches the stored verification_hash, the report is authentic.

Alternative i considered we could pepper with user_id: hash(plaintext || user_id)

Your secret pepper approach is more robust cryptographically, but we're not protecting against post-quantum super computers. 

We are solving a practical moderation problem: verifying report authenticity without breaking E2EE.

I'm open to changing the design if there is a use case or security concern that the current approach doesn't address. 

What specific attack scenario are you protecting against that per-message public pepper doesn't prevent?

PR #4383, opened 2025-11-21
MSC4383: Client-Server Discovery of Server Version

jevolk:
[rendered](https://github.com/matrix-construct/matrix-refreshed/blob/jevolk/client-server-info/proposals/4383-client-server-info.md)

## Implementations

- **Server**: [Tuwunel](https://github.com/matrix-construct/tuwunel) (shipped in v1.6.2), populating the `server` object on `GET /_matrix/client/versions` and advertising `net.zemos.msc4383` in `unstable_features`.
- **Protocol types**: [ruma/ruma#2495](https://github.com/ruma/ruma/pull/2495) adds the `Server` struct and the `Response.server` field behind the `unstable-msc4383` Cargo feature.
- **Client SDK**: [matrix-org/matrix-rust-sdk#6622](https://github.com/matrix-org/matrix-rust-sdk/pull/6622) rewrites `Client::server_vendor_info` to prefer the client-server source, falling back to `GET /_matrix/federation/v1/version` only when the object is absent. The feature is enabled by default in `matrix-sdk-ffi`, so applications consuming the FFI binding acquire the client-server path on rebuild without an app-side flag.

## Pre-acceptance checklist

- [x] Are appropriate implementation(s) specified in the MSC's PR description? (See above.)
- [x] Are all MSCs that this MSC depends on already accepted? (No dependencies.)
  - [x] Have authentication requirements been specified? (`GET /_matrix/client/versions` remains unauthenticated.)
  - [x] Have rate-limiting requirements been specified? (Not rate-limited.)
  - [x] Have guest access requirements been specified? (Guest-access restrictions are not applicable.)
  - [x] Are error responses specified? (No new error responses; the field is purely additive on the existing response body.)
    - [x] Does each error case have a specified `errcode` and HTTP status code? (N/A.)
      - [x] If a new `errcode` is introduced, is it clear that it is new? (N/A.)
- [x] Will the MSC require a new room version, and if so, has that been made clear? (No new room version.)
  - [x] Is the reason for a new room version clearly stated? (N/A.)
- [x] Are backwards-compatibility concerns appropriately addressed? (See §Backwards Compatibility: additive wire shape, `unstable_features` advertise gate, and SDK migration path preserving the existing `Client::server_vendor_info` contract.)
- [x] Are the endpoint conventions honoured?
  - [x] Do HTTP endpoints `use_underscores_like_this`? (Existing endpoint; convention preserved.)
  - [x] Will the endpoint return unbounded data? If so, has pagination been considered? (Fixed-shape object; pagination not applicable.)
  - [x] If the endpoint utilises pagination, is it consistent with the appendices? (N/A.)
- [x] An introduction exists and clearly outlines the problem being solved. (§Introduction.)
- [ ] All outstanding threads are resolved.
  - [x] All feedback is incorporated into the proposal text itself, either as a fix or noted as an alternative.
- [x] While the exact sections do not need to be present, the details implied by the proposal template are covered:
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies (none; omitted)
- [x] Stable identifiers are used throughout the proposal, except for the unstable prefix section.
  - [x] Unstable prefixes consider the awkward accepted-but-not-merged state (`net.zemos.msc4383.stable` companion flag listed in §Unstable prefix).
  - [x] Chosen unstable prefixes do not pollute any global namespace (vendor prefix is the controlled reverse-DNS `net.zemos.*`).
- [x] Changes have applicable Sign Off from all authors/editors/contributors. Per-commit Signed-off-by on every new commit.
- [x] There is a dedicated "Security Considerations" section.

### DIFF
--- a/proposals/4383-client-server-info.md
+++ b/proposals/4383-client-server-info.md
@@ -4,8 +4,10 @@
 
 Client applications utilizing the Matrix SDK[^6], a Client-Server[^7] library, have been observed
 making requests to `GET /_matrix/federation/v1/version`[^2]. It is the only known example of a
-cross-interface request to have any possible[^5] use to any implementation[^3], motivating this
-proposal to reestablish a properly clean partition between client and federation interfaces.
+cross-interface request made by an implementation[^3]. The purpose is valid: informing logs
+used for crash reports of the server specifics; often essential for efficient triage and robust
+patching of client software. This proposal reestablises the proper partition between client and
+federation interfaces while retaining the value added by the de facto misuse.
 
 Such cross-interface requests are problematic: the partition in the `/_matrix` URL hierarchy is
 understood by site-administrators to allow for distinct middleware configurations. Sites commonly
@@ -83,9 +85,6 @@ proxy) will have to note their implementation's new exposure of it when upgradin
 [^3]: https://github.com/matrix-org/matrix-rust-sdk/blob/d228bde8ef51a98da10a6b7d4b9f3e5b8f49ad3c/crates/matrix-sdk/src/client/mod.rs#L581-L616
 
 [^4]: https://github.com/matrix-org/matrix-spec-proposals/pull/2301
-
-[^5]: Server version information can provide useful context in logs and may be essential for
-effective crash reports.
 
 [^6]: https://github.com/matrix-org/matrix-rust-sdk
 
--- a/proposals/4383-client-server-info.md
+++ b/proposals/4383-client-server-info.md
@@ -1,0 +1,92 @@
+# MSC4383: Client-Server Discovery of Server Version
+
+### Motivation
+
+Client applications utilizing the Matrix SDK[^6], a Client-Server[^7] library, have been observed
+making requests to `GET /_matrix/federation/v1/version`[^2]. It is the only known example of a
+cross-interface request to have any possible[^5] use to any implementation[^3], motivating this
+proposal to reestablish a properly clean partition between client and federation interfaces.
+
+Such cross-interface requests are problematic: the partition in the `/_matrix` URL hierarchy is
+understood by site-administrators to allow for distinct middleware configurations. Sites commonly
+employ WAF rules covering the client and federation hierarchies with source and destination
+assumptions. Some sites disable the entire federation portion, rendering the status quo as unreliable.
+
+This proposal reestablishes the partition between client-server and server-server by simply offering
+the same version data over both. With the adoption of this proposal, no need for cross-interface
+requests will be known to remain.
+
+### Proposal
+
+An object is added to the top-level of `GET /_matrix/client/versions`[^1] named `server` with
+parity to the eponymous object returned by `GET /_matrix/federation/v1/version`[^2].
+
+Example of the `server` object:
+
+```
+  "server": {
+    "name": "My_Homeserver_Implementation",
+    "version": "ArbitraryVersionNumber"
+  }
+```
+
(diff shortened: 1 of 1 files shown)


proposals/4383-client-server-info.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4383-client-server-info.md
```diff
@@ -0,0 +1,80 @@
+# MSC4383: Client-Server Discovery of Server Version
+
```
turt2live, line 2: This proposal appears to be lacking a problem statement and/or introduction, making it difficult to review with an aim to include it in the spec.

jevolk: The problem is stated in the [Alternatives](https://github.com/matrix-construct/matrix-refreshed/blob/jevolk/client-server-info/proposals/4383-client-server-info.md#alternatives) section. I didn't feel the need to make a redundant introduction given the status quo alternative is prominent and it's a small proposal otherwise. Should the first paragraph be moved to the top?

turt2live: Relevant information and background should always be first. I didn't feel the paragraph in the Alternatives was suited as an introduction because it lacks background and context.

turt2live:
(please avoid force pushing after review is given - it makes review more difficult. We do not require clean commit history)

proposals/4383-client-server-info.md
```diff
@@ -0,0 +1,92 @@
...
+### Motivation
+
+Client applications utilizing the Matrix SDK[^6], a Client-Server[^7] library, have been observed
+making requests to `GET /_matrix/federation/v1/version`[^2]. It is the only known example of a
+cross-interface request to have any possible[^5] use to any implementation[^3], motivating this
+proposal to reestablish a properly clean partition between client and federation interfaces.
```
Johennes: Why do these applications query the SS API in the first place? Also, do they query the SS API on their own home server or on remote servers?

jevolk: In the found case, the information is logged and not used to determine application-logic. The logs are then included in rageshakes.

The query is only to the local server, as remote servers would require server name resolution which the client is not capable of—arguably the local server also requires the resolution for the federation interface which is itself another concern for why the status quo is untenable.

Johennes: Thanks! I think both of these would be good to include in the main text as they help understand the proposed solution. For instance, if querying `/versions` on a remote server was a use case, you would need a different API shape that allows to pass in the server name.

jevolk: Are you suggesting morphing this proposal into a client API where the homeserver can run version queries on any other server on behalf of the client?

Johennes: Sorry, no. I just meant to use that as an example to illustrate that without describing the use case it's not clear why the proposed solution is the best one.

jevolk:
Implemented in Tuwunel (shipped in v1.6.2). `GET /_matrix/client/versions` includes the unstable `net.zemos.msc4383.server` object populated with the homeserver name and version string, alongside the `net.zemos.msc4383` flag in `unstable_features`.

Source: [src/api/client/versions.rs#L32-L38](https://github.com/matrix-construct/tuwunel/blob/51c19f3ffd21d2e02d25ce11ef9b035f0372d01d/src/api/client/versions.rs#L32-L38).

PR #4386, opened 2025-12-08
MSC4386: Automatically sharing secrets after device verification

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/sharing_secrets_after_verification/proposals/4386-sharing-secrets-after-verification.md)

Disclosure: I am a member of the Element crypto team, and of the Spec Core Team. This proposal is written as a member of the Element crypto team.

### DIFF
--- a/proposals/4386-sharing-secrets-after-verification.md
+++ b/proposals/4386-sharing-secrets-after-verification.md
@@ -1,4 +1,4 @@
-# MSCxxxx: Automatically sharing secrets after device verification
+# MSC4386: Automatically sharing secrets after device verification
 
 Currently, after a user verifies a new device, in many clients, the new device
 [requests secrets](https://spec.matrix.org/v1.16/client-server-api/#sharing)
@@ -196,9 +196,9 @@ verification, this is not a problem.
 
 Before this MSC is accepted, the following names are used:
 
-- `io.element.mscxxxx.secrets` instead of `secrets` for the property name in
+- `io.element.msc4386.secrets` instead of `secrets` for the property name in
   `m.key.verification.request`/`m.key.verification.ready` events, and
-- `io.element.mscxxxx.key.verification.secrets` instead of `m.key.verification.secrets`
+- `io.element.msc4386.key.verification.secrets` instead of `m.key.verification.secrets`
   for the event type.
 
 ## Dependencies


proposals/4386-sharing-secrets-after-verification.md
turt2live, line 1: Implementation requirements:

* [ ] Sending client
* [ ] Receiving client (ideally different from the sending one)

proposals/4386-sharing-secrets-after-verification.md
```diff
@@ -0,0 +1,206 @@
...
+    "from_device": "DEVICEB",
+    "methods": ["m.sas.v1"],
+    "transaction_id": "a_unique_string",
+    "secrets": {
+      "available": [
+        "m.cross_signing.master",
+        "m.cross_signing.self_signing",
+        "m.cross_signing.user_signing"
+      ],
+      "requested": ["org.example.custom"]
```
Johennes, line 143: Wouldn't DEVICEB already know that DEVICEA cannot provide this secret based on the earlier verification request?

proposals/4386-sharing-secrets-after-verification.md
```diff
@@ -0,0 +1,206 @@
...
+DEVICEB, and so doesn't send one.  But notices that it expects to receive one,
+so it waits until it receives the `m.key.verification.secrets` event from
+DEVICEB, and then sends an `m.key.verification.done` event.
+
+FIXME: add some examples of how clients could use the `secrets` property and
+`m.missing_secrets` code to provide feedback to user.
+
+## Potential issues
+
+None?
```
Johennes, line 183: When using in-room verification, I think, you could theoretically run into the 64KiB event size limit? Secrets stored in account data won't be subject to this limit. Even if they were, you could exceed it because multiple secrets are combined in a single `m.key.verification.secrets` event.

This is probably not a practical problem because secrets should normally be relatively small. So I'm not sure if we really need to do anything. Maybe a dedicated error code at most?

uhoreg: We don't share secrets with other users, and in-room verification is only used for verifying other users.

But yes, we should be aware of event sizes.  It shouldn't be a problem for now because we only have a small number of secrets, and each secret is small, but maybe in the future.  I'll add a comment..

proposals/4386-sharing-secrets-after-verification.md
uhoreg: Yes, it if wanted to, DEVICEB could tailor its response based on what it saw from DEVICEA's event.

PR #4387, opened 2025-12-09
MSC4387: `M_SAFETY` error code

turt2live:
> [!WARNING]
> **Content Warning**: This proposal discuses and identifies harmful content, but does not attempt to describe the harm posed in detail. This includes identifiers for child safety, sexual abuse, self-harm, and other types of harm a user may encounter on the open internet.

----

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/m_safety/proposals/4387-safety-error-code.md)

----


proposals/4387-safety-error-code.md
turt2live, line 1: Implementation requirements:
* [ ] Server returning new error code on (doesn't have to be the same server for each):
  * [ ] Room directory
  * [ ] Sending events (ideally via a policy server)
  * [ ] Media uploads and downloads
* [ ] Client reacting to `harms` on (doesn't have to be the same client for each either):
  * [ ] Room directory
  * [ ] Sending events
  * [ ] Media uploads and downloads
* [ ] Client unaware of `M_SAFETY` responding reasonably well to the new error code (ie: doesn't crash, even if the error message given to the user is a bit ugly)

TheArcaneBrony: Anything based on Rory&::LibMatrix would probably not crash, assuming the client implementer actually does error handling and doesn't either re-throw or fall through unknown errors. Not sure if this counts as "Client unaware of M_SAFETY doesn't crash"?

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+* `m.spam.fraud` - Might include scams and phishing too.
+* `m.spam.impersonation`
+* `m.spam.election_interference`
+* `m.spam.flooding`
+
+**Adult Content & Safety**
+
+* `m.adult`
+* `m.adult.sexual_abuse`
+* `m.adult.ncii` - "Non-Consensual Intimate Imagery"
```
TheArcaneBrony: When would this be returned, and depending on the answer, how does one define consent (in terms of matrix spec, not the meaning of consent itself)?

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+to some safety legislation around the world. They are split up first into major categories then
+subtyped for slightly more specific use. Where a subtyped harm doesn't apply, the general category
+(the first one in each list) should be used instead.
+
+**Spam**
+
+* `m.spam`
+* `m.spam.fraud` - Might include scams and phishing too.
+* `m.spam.impersonation`
+* `m.spam.election_interference`
```
TheArcaneBrony: Isn't election interference the same as fraud? This seems oddly specific.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+* `m.spam.flooding`
+
+**Adult Content & Safety**
+
+* `m.adult`
+* `m.adult.sexual_abuse`
+* `m.adult.ncii` - "Non-Consensual Intimate Imagery"
+* `m.adult.deepfake`
+* `m.adult.animal_sexual_abuse`
+* `m.adult.sexual_violence`
```
TheArcaneBrony: I feel like both of these could be merged into sexual abuse? I'm not sure I see a case for separating these in particular. I could be missing context here or just not understand the reasoning behind why these are separate, though.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+* `m.tos.hacking`
+* `m.tos.prohibited` - Might include drugs, weapons, etc.
+* `m.tos.ban_evasion`
+
+## Future scope
+
+The harms list is intended to be reused in other areas of Matrix, such as ban reasons, account
+suspensions, and room takedowns. These areas need significant thought before becoming standalone
+proposals however, so are excluded here in favour of future MSCs.
+
```
TheArcaneBrony: I think I prefer reasons remaining as plaintext, as it allows us to communicate better, but at the same time I've also been wishing for a tag-like system in ie. policy events, which would be very similar to what this MSC is proposing in a different context.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+  capability issue whereas `M_SAFETY` is intended to be more dynamic and less deterministic.
+
+* Some variation of sub-errors, where there's a top level `errcode` with one or more "sub error codes".
+  This could theoretically replace the `harms` array and be more generic outside of safety use cases,
+  though practical examples aren't as easily found.
+
+* Instead of pushing the rendering to the client, the server could provide pre-translated errors with
+  [MSC4176](https://github.com/matrix-org/matrix-spec-proposals/pull/4176). `M_SAFETY` is compatible
+  with MSC4176, but decided against using it directly due to concerns about server projects needing
+  to sprout a translations layer and the complexities involved in doing that.
```
TheArcaneBrony, line 157: There will still be reasonable English `error` fields, I presume (also given the previous examples)?

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+```jsonc
+{
+  "errcode": "M_SAFETY",
+  "error": "That message appears to be trying to mention 3000 people, which is too many.",
+  "harms": [
+    "m.spam",
+  ]
+  // permanent because `expiry` is missing: resending the message with its 3000 mentions won't work
+}
+```
```
TheArcaneBrony, line 83: Is there even any semantic difference here between `expiry` being null and it being missing? I *feel* like the answer is no, given the above paragraph, but I just wanted to make sure - I much prefer the spec way of just calling a property optional rather than saying "null or missing".

proposals/4387-safety-error-code.md
turt2live: I don't recall if the spec requires `error` to be English specifically, but this MSC makes it a required field when the `M_SAFETY` error code is present.

proposals/4387-safety-error-code.md
turt2live: missing is the same as elided, just a matter of perspective: "missing" is from the client's point of view while "elided" is from the server's.

proposals/4387-safety-error-code.md
TheArcaneBrony: Well, generally it tends to be in English due to programming languages and most public communities being English-centric, which I suppose is what I was getting at :)

proposals/4387-safety-error-code.md
turt2live: It certainly helps. With my SCT hat rather than author hat on though, I'd want to see the end client not crashing rather than the library being safe. Usually the requirement is met around the same time as error code implementation because the "before" state can usually be tested pretty quickly.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+* `m.tos.ban_evasion`
+
+## Future scope
+
+The harms list is intended to be reused in other areas of Matrix, such as ban reasons, account
+suspensions, and room takedowns. These areas need significant thought before becoming standalone
+proposals however, so are excluded here in favour of future MSCs.
+
+The harms list is also expected to be used in submitting event/user/room reports during the "reporting
+v2" project being undertaken by the Foundation's T&S team.
```
Gnuxie: Labels can be useful for report triaging and for keeping internal context on moderation decisions.

However, labels for policies carries significant risks and challenges, and it is my assessment that they are neither necessary or worth the risk.

The first challenge is ironically preventing the classification of content via policies. While at first this might seem like the exact opposite of what is desirable, this is actually the mechanism by which the incentive to use a dictionary attack against policies (in order to use them as a directory of abuse) is removed (See https://github.com/matrix-org/matrix-spec-proposals/pull/4204 & https://github.com/matrix-org/matrix-spec-proposals/pull/4205). Because now the only thing that is classified is that someone wanted it removed.

The second is that the most reliable way of determining whether a policy is appropriate to apply is your familiarity and trust of both the subject of and curator of the policy. 

On social networks with decentralised moderation decisions, it is important for the mechanisms providing safety to retain trust. This can be managed in a number of ways, including keeping permanent and irreversible consequences out of automated tooling. The culture, processes, and practices of the community is also important here. The use of labels within blocklist is an example of a consequence with high risk because it can enable for the automatic classification of a user's activity. Which not only has privacy implications. But it also allows for policies to become a means by which the details of allegations are propagated, which is inappropriate, and undermines trust in the system.

proposals/4387-safety-error-code.md
Half-Shot: I plan to implement this in matrix-js-sdk explicitly, but even so I'm sure Element Web for ex will be fine with unexpected errcodes and formats.

SDK-PR: https://github.com/matrix-org/matrix-js-sdk/pull/5107

proposals/4387-safety-error-code.md
KitsuneRal: It very much is not? One can interfere with elections without committing fraud.

proposals/4387-safety-error-code.md
KitsuneRal: Plain text reasons are not localisable.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+**Violence**
+
+* `m.violence`
+* `m.violence.animal_welfare`
+* `m.violence.threats`
+* `m.violence.graphic`
+* `m.violence.glorification`
+* `m.violence.extremist`
+* `m.violence.human_trafficking`
+
```
jevolk: Speaking as a proud donor to my local shelter I am certain this list falls tragically short without any mention of Intimate-Partner or Domestic Violence which is one of if not the most prevalent form of societal harm and should not be forgotten here.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,268 @@
...
+impact of any action taken against their account.
+
+This MSC introduces a new error code to explain to users that safety tools acted on a request they
+made. Rather than a generic `M_FORBIDDEN`, tools can use `M_SAFETY` to explain the reasons for the
+action, and if the impact is permanent.
+
+## Proposal
+
+A new error code, `M_SAFETY` with HTTP 400 status code, is introduced on the following endpoints:
+
```
Half-Shot, line 16: I think profile endpoints should also appear here

turt2live: A future MSC is best for adding more endpoints - the list is already pretty ambitious, and making it longer will delay the MSC through the process.

Most of the Client-Server API qualifies for this new error code.

proposals/4387-safety-error-code.md
turt2live: considering resolved - they are distinct concepts.

proposals/4387-safety-error-code.md
turt2live: NCII is a term used in industry to generally mean "shared with more people than communicated".

Noting that these websites deal with sexual abuse topics, [StopNCII](https://stopncii.org/), [INHOPE](https://inhope.org/EN/articles/what-is-ncii), and [Meta's Safety Center](https://www.meta.com/safety/topics/bullying-harassment/ncii/) all describe what this means.

proposals/4387-safety-error-code.md
turt2live: Several jurisdictions distinguish between the two, and can often carry different consequences for the offences.

proposals/4387-safety-error-code.md
turt2live: Would adding `m.violence.intimate` cover those cases in your opinion? Another option may be `m.adult.ipv` to match definitions from the [RCMP](https://rcmp.ca/en/relationship-violence/intimate-partner-violence-and-abuse)/[Women and Gender Equality Canada](https://www.canada.ca/en/women-gender-equality/gender-based-violence/intimate-partner-violence.html), [CDC](https://www.cdc.gov/intimate-partner-violence/about/index.html), [WHO](https://apps.who.int/violence-info/intimate-partner-violence/), and [UN](https://www.un.org/en/coronavirus/what-is-domestic-abuse).

proposals/4387-safety-error-code.md
turt2live: "ban reasons" is intentionally broad in this MSC to not necessarily mean "the `reason` field". I'll aim to clarify this in the MSC.

proposals/4387-safety-error-code.md
turt2live: The reporting system is expecting to take a harm as an input to help guide initial ticket routing and prioritization. It'll unfortunately be an implementation detail whether the system makes automated decisions based upon the user-supplied data, though the Foundation does not intend to support code that does so. In the design we're thinking of, once in the system, the harms associated with a reported entity (user, server, room, community, etc) will have no technical reason to leave the system. We are not planning to endorse or write MSCs which add harms to policy room rules for the exact reasons you've already highlighted.

I'll aim to clarify this in the MSC to help deter future MSCs from heading down that path.

proposals/4387-safety-error-code.md
turt2live: It may be, but also the `error` is hopefully overridden by clients via the `harms` list.

proposals/4387-safety-error-code.md
jevolk: I think `m.violence.domestic` adequately covers all of the above without being too broad on the subject, and fits well lexically with the code category; it would be my first choice all things considered.

proposals/4387-safety-error-code.md
TheArcaneBrony: Wouldn't it be easier to just spec it as a generic code then? (ie. may be thrown from any CS endpoint, rather than being specific to certain endpoints)

turt2live: Not all endpoints will need it, but most of the ones that do are also non-trivial implementations, so we'd have to implement each and every endpoint before this MSC gets merged.

From a process perspective, more MSCs is no bad thing here.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,273 @@
...
+  "harms": [
+    // there might not be any articulable harm, so leave the array empty or elide `harms` entirely.
+  ],
+  "expiry": 1765312137979 // "5ish minutes from now"
+}
+```
+
+Note that clients can (and SHOULD, where possible) render more detailed error messages than those
+provided as `error`. For example, if `harms` contains `m.child_safety.csam`, then the client might
+include links to Lucy Faithfull Foundation's [Stop It Now](https://www.stopitnow.org.uk/) support
```
Half-Shot: Element Web is unlikely to start allowing you to click on links in errors, but I think a learn more URL in the error would be useful (also for linking to the tos). This would enable us to replace something like the retry button with a link to help resources in the client.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,273 @@
...
+is deliberately left as an implementation detail. The harms listed below do not include explicit
+definitions for a similar reason: the implied definition attached to a given harm may differ between
+environments, but otherwise carry the same communicated intent to the user.
+
+The initial set of specified harms are intended to mirror those used by other services or are common
+to some safety legislation around the world. They are split up first into major categories then
+subtyped for slightly more specific use. Where a subtyped harm doesn't apply, the general category
+(the first one in each list) should be used instead.
+
+**Spam**
```
Half-Shot: When implementing this, I found it challenging to find the right wording for all of these. I will probably lean on the m.org T&S team for help, but we could do with a guide on how to phrase each of these harms to users.

proposals/4387-safety-error-code.md
Half-Shot: Element Web now has an impl in draft https://github.com/element-hq/element-web/pull/31558. It currently only handles message failures for the moment, due to the complexity to get this off the ground.

proposals/4387-safety-error-code.md
```diff
@@ -0,0 +1,273 @@
...
+    // there might not be any articulable harm, so leave the array empty or elide `harms` entirely.
+  ],
+  "expiry": 1765312137979 // "5ish minutes from now"
+}
+```
+
+Note that clients can (and SHOULD, where possible) render more detailed error messages than those
+provided as `error`. For example, if `harms` contains `m.child_safety.csam`, then the client might
+include links to Lucy Faithfull Foundation's [Stop It Now](https://www.stopitnow.org.uk/) support
+website. The `error` is provided as fallback and should be understandable to a human user.
```
Half-Shot: Noting that Element Web will not allow us to use the server response as it will violate our rules to translate strings (https://github.com/element-hq/element-web/pull/31558#discussion_r2686596776), so for this we'd have to instead introduce a generic harms statement.

turt2live: That's expected - will incorporate it into the proposal on the next editing pass.

proposals/4387-safety-error-code.md
turt2live: This would be the error the client renders *instead of* the server-provided error, so should be possible (hopefully). Rendering links from `error` is not expected of clients.

proposals/4387-safety-error-code.md
turt2live: this might be more challenging - the MSC avoids assigning definitions to these. Adding phrasing can imply more of a definition than intended.

Something to consider as the MSC evolves, though.

turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4456 touches on this a little bit. TLDR: "Broad descriptions are better than specific ones for compatibility".

proposals/4387-safety-error-code.md
turt2live: resolving - list has moved to https://github.com/matrix-org/matrix-spec-proposals/pull/4456

proposals/4387-safety-error-code.md
turt2live: resolving - list has moved to https://github.com/matrix-org/matrix-spec-proposals/pull/4456

proposals/4387-safety-error-code.md
turt2live: considering resolved - I'm adding some language to the MSC to clarify that `error` might not have markup.

PR #4391, opened 2025-12-24
MSC4391: Simplified in-room bot commands

Gnuxie:
[Rendered](https://github.com/Gnuxie/matrix-doc/blob/gnuxie/simplified-in-room-bot-commands/proposals/4391-simplified-in-room-bot-commands.md)

---

This MSC incorporates work from [MSC4332](https://github.com/matrix-org/matrix-spec-proposals/pull/4332)


proposals/4391-simplified-in-room-bot-commands.md
tulir, line 1: Implementation requirements:

* [x] Client
* [x] Bot

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
+# MSC4391: Simplified in-room bot commands
```
Gnuxie, line 1: At this stage the review i am looking for is whether this is an approach that we like. From there we need to:

- Give a more concrete description of the various schema for command descriptions and commands, particularly the command parameter schema type.
- Descriptions and schema as well as the existing examples.

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+The `content` for such an event fits the following implied schema:
+
+```json
+{
+  "type": "m.bot.command_description",
+  "sender": "@draupnir:draupnir.space",
+  // derived from `sha256(designator.join('') + mxid)`
+  "state_key": "JBDLR6YMe+72yqsEMi/MVdTmjN3ynPThMz+M7QLATZQ=",
+  "content": {
+    "designator": ["ban"],
```
tulir: Why is this an array? If it's for subcommands, why not just have a space-separated string? Also, aliases are important

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+- While a response format is not specified in this MSC, the MSC is designed to
+  be forward compatible with a follow up MSC for _partial commands_ and
+  _prompts_.
+
+The proposal makes no attempt to specify the syntax of commands within an
+`m.room.message` body or describe the syntax of those commands. This is
+specifically because when
+[MSC4332 tried to do this](https://github.com/matrix-org/matrix-spec-proposals/pull/4332/files#r2313755345)
+it was not found to be easy to do in a backwards compatible way and would have
+required some existing bots to make changes to their parsers. Additionally the
```
tulir: Not sure if "doesn't work with some bots" is a valid argument for removing fallbacks entirely, since the fallbacks would work with many simple bots. I think the main issue with MSC4332 was that it made the fallback syntax the source of truth, rather than having a template string that's exclusively used for generating the fallback.

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+
+### Command description
+
+A new state event type is introduced: `m.bot.command_description`. When
+presenting command options to users, clients SHOULD use this event to suggest
+commands, scoped by the `sender` of the description.
+
+The `content` for such an event fits the following implied schema:
+
+```json
```
tulir:
```suggestion
```jsonc
```

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: Yeah we should make this a single string that may be space separated.

The difficulty with aliases is whether a room admin is supposed to be able to set those or just the bot. Because one of the security invariants is that the `sender` of the command description describes which bot it is for, and this is in order to ensure bots don't sabotage each other or something.

tulir: `designator` is a weird word, should probably just call it `command`

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+- `room_id` - An object with the property `room_id` which must be a valid
+  [room ID](https://spec.matrix.org/v1.15/appendices/#room-ids). The object
+  SHOULD also contain a `via` property that contains an array of server names
+  that the room can be joined via.
+- `event_id` - An object with the same properties as `room_id` with the addition
+  of an `event_id` property that must be a valid
+  [event ID](https://spec.matrix.org/v1.15/appendices/#event-ids).
+
+When an object is provided as the type schema, the type is a derived from a
+schema described by the property `schema_type`:
```
tulir, line 280: `type` being either a string or an object is annoying to parse in some languages. It would be preferable if it was always an object with varying keys. e.g. `{"schema_type": "primitive", "type": "room_id"}`

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+
+- The `array` schema type specifies the type of the items with the `items`
+  property.
+- The `union` schema type specifies the types of the variants with the
+  `variants` property. Which is an array of type schema.
```
tulir: There should probably be some rules, like

* an array can only appear at the top level
  * let's not do matrices as command parameters
* an union can only appear at the top level or directly inside a top-level array
  * nested unions are never needed as they can be flattened
  * an array of unions is necessary, e.g. for a `/join` command that takes a list of things to join (union of `room_id` and `room_alias`)
  * an union of arrays could theoretically be useful for a command that e.g. takes either a list of room ids or a list of user ids (but not a mix), but that sounds complicated to implement for clients and I can't immediately think of use cases, so probably best to forbid it too

tl;dr: `items` only allows the `union`, `primitive` and `literal` schema types. `variants` must only consist of `primitive` and `literal` schema types.

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+- The `literal` schema type specifies a literal value with the `value` property
+  and the type of the literal value with the `literal_type` property.
```
tulir, line 295: I don't think there's any reason to specify a reference type here, so is `literal_type` needed if it can only be a string, integer or boolean? Could just be the `value` field that contains one of the three types (never an object/array/null)

proposals/4391-simplified-in-room-bot-commands.md
tulir: Probably enough for the bot to be able to set them

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+      },
+
+      {
+        "key": "apply_to_policy",
+        "type": "boolean",
+        "description": {
+          "m.text": [{ "body": "Whether to apply this to the policy" }]
+        },
+        // This argument is not required
+        "required": false
```
tulir: As an alternative/extension to optional parameters, a default value might be useful. The `/raw` command in gomuks has `{}` as the default value for event JSON for example

Also, `"optional": true` might be nicer than `"required": false` to allow treating undefined as false

proposals/4391-simplified-in-room-bot-commands.md
tulir: Also, should arrays be allowed only as the final parameter (i.e. as a variadic parameter)? Multiple arrays or other parameters after arrays could get complicated

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,332 @@
...
+commands, scoped by the `sender` of the description.
+
+The `content` for such an event fits the following implied schema:
+
+```json
+{
+  "type": "m.bot.command_description",
+  "sender": "@draupnir:draupnir.space",
+  // derived from `sha256(designator.join('') + mxid)`
+  "state_key": "JBDLR6YMe+72yqsEMi/MVdTmjN3ynPThMz+M7QLATZQ=",
```
tulir, line 103: Is the state key format mandatory or just a suggestion? If it's mandatory, clients could validate it as a form of deduplication. If not, clients will probably need to ignore state keys and implement some deduplication on the command itself

proposals/4391-simplified-in-room-bot-commands.md
tulir: https://github.com/gomuks/gomuks/pull/675 is mostly complete implementation of client and technically also bot, converted from MSC4332. It doesn't have full support for room/event IDs or unions yet, but mostly it should be there

It assumes most of my suggestions in other comments are applied. Specifically:
* `command` string instead of `designator` array
* `"optional": true` instead of `"required": false`
* primitive schemas nested in an object with `"schema_type": "primitive"`
* arrays only at top level and unions only at top level or in arrays)
* optionally: list of `aliases`, `default_value` for default values

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: Should the type property also just be changed to `schema` then on the parameter description?

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: Could you tell me what you are anxious about if reference types and arrays are supported literals?

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: So i guess it's the same situation/precedent as the state key for moderation policies. Would you like to make it a SHOULD to avoid client work or do you have something specific in mind as an alternative?

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: I didn't include defaults at this stage because they can also be provided by partial commands and interactive prompts ie the way suggestions work in draupnir. I guess it might be ok if it's treated like a literal default.

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: Well i did that

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: So the original thinking is that arrays should be supported on optional arguments because in syntax or client ux they can appear anywhere and have many representations. Ie a lot of the trouble with providing a ubiquitous optional argument in the prior MSC was that there were so many weird syntaxes for multiple options. Arrays here are meant to eliminate that.

I was also thinking that clients would ideally be providing ui for command composing rather than naive slash commands with syntax. Ie an argument for an array of rooms would open up a filterable room list to select multiple. 

The concern is that it's a pita for clients to parse or provide weird syntax or ui for arrays in the middle of commands.

We can go forward and say that bots that do this could expect weird behaviour from clients when doing this but that doesn't really seem acceptable.

Do you have further thoughts here?

proposals/4391-simplified-in-room-bot-commands.md
tulir: MUST would be nice since that allows clients to discard any duplicate commands simply by enforcing the state key format, but just needs some keyword. If it's not MUST, clients will ignore the state key entirely and implement deduplication elsewhere (deduplication of some kind is probably necessary, duplicate command definitions don't make much sense)

proposals/4391-simplified-in-room-bot-commands.md
tulir: Text command parsing handles literals by checking if the input text matches the literal value. It's easiest to do for string/integer/boolean and I don't see any valid reason to have literal objects or arrays

proposals/4391-simplified-in-room-bot-commands.md
tulir: Yeah that makes sense

proposals/4391-simplified-in-room-bot-commands.md
tulir: I ended up implementing arrays in the middle by requiring that they're surrounded in `<>`, which will probably work fine. It's not a problem in UI, but I suspect many clients will start with text parsing rather than a full command UI, plus terminal clients and such may always use text parsing.

Gnuxie: I think we will write a note about it then

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: It shouldn't be difficult to provide a definition of literal equality for reference types and arrays though? The objects aren't arbitrary with their own structure

tulir: Room and event IDs have multiple textual representations, but those might be fine-ish. I don't see any reason to allow arrays though. Parsing those would be even more difficult, plus they'd cause more schema nesting issues (union of literals is allowed, but an union of literals of arrays would be bad)

Lack of arrays won't prevent anything, you can always have a literal comma-separated string or whatever. At worst it'll have slightly worse UI, but not sure if literal arrays would ever have good UI either. You can't put an array into a `<select>` UI without stringifying it first, at which point the command definition could've just stringified it ahead of time to save everyone the trouble.

Gnuxie: ok

Gnuxie: > Lack of arrays won't prevent anything, you can always have a literal comma-separated string or whatever. At

This would be unacceptable. The entire point of the proposal is to provide structure, and this would be the one place where we decide to torpedo ourselves. I'm ok with taking them out for simplicity and following up, but if someone has to resort to this it will be crap.

Gnuxie: The purpose of the literal schema is so that an argument provided for that parameter is structurally equivalent to the literal. I don't think I understand the problem you're raising with serializing for `<select>`, it sounds like the purpose has been confused with a default value or suggestions list.

tulir: Literals primarily exist for creating enums (via unions), an union of literals would likely show up as a `<select>`. The only non-enum use I can think of is flags where a literal is an optional parameter, although a boolean could do that too. Neither of those use cases need arrays.

Even in the unlikely event that an array is needed, the structure is only really relevant for the bot, because the client isn't allowed to touch the value. Structure is important when multiple parties (client/UI & bot) need to agree on how values are formed, which is not the case for literals.

Gnuxie: We had a discussion out of band about this and idk what the summary was but essentially i was coming at literals from an angle where i was only expecting a literal value, and tulir was expecting a union of literals to provide enums. In any case, taking arrays out of literals and probably also references to matrix entities would make things simpler for now so we'll just do that.

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: I'm a little bit hesitant to do so right now and am thinking of making a JSON RPC style response follow up MSC that can explore the prompts and partial commands before settling on how to do defaults.

proposals/4391-simplified-in-room-bot-commands.md
Gnuxie: (all of these suggestions were up-streamed into the MSC)

tulir: Proper bot implementation: https://github.com/maunium/meowlnir/pull/50

Both meowlnir and parts of gomuks use logic from https://github.com/mautrix/go/pull/446 (gomuks web has its own typescript implementation, gomuks terminal uses mautrix-go)

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,374 @@
...
+
+- Some predefined validation on arguments, like a range for integers or
+  maximum/minimum length of strings.
+
+## Potential issues
+
+- Using state events limits a bot's ability to advertise commands if it isn't
+  given power to do so.
+
+## Alternatives
```
Gnuxie, line 336: I'm in the process of adding a "response" mixin in a follow up MSC. It would be really useful to also be able to proxy the commands into a mutual room (like draupnir's command room). See https://github.com/the-draupnir-project/Draupnir/issues/1041#issuecomment-4071332194

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,374 @@
...
+  may be a better option. In either case, bots should not be able to cause users
+  to send state events to prevent bots from tricking users into changing power
+  levels, join rules, etc.
+
+- Support for non-text-like arguments like images, files, etc.
+
+- Some predefined validation on arguments, like a range for integers or
+  maximum/minimum length of strings.
+
+## Potential issues
```
Gnuxie, line 331: When commands are available across an entire space it is a pita to have to resend the state events everywhere. So maybe it would be nice to be able to only specify the commands in a space or take inspiration from sticker pack MSCs if they haven't encountered the same problem.

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,374 @@
...
+
+```jsonc
+{
+  "type": "m.bot.command_description",
+  "sender": "@draupnir:draupnir.space",
+  // derived from `sha256(command + mxid)`
+  "state_key": "JBDLR6YMe+72yqsEMi/MVdTmjN3ynPThMz+M7QLATZQ=",
+  "content": {
+    // space separated string for commands such as "rooms add"
+    "command": "ban",
```
TheArcaneBrony, line 106: Is there a way to define command aliases?

proposals/4391-simplified-in-room-bot-commands.md
```diff
@@ -0,0 +1,374 @@
...
+    "parameters": [
```
TheArcaneBrony, line 107: this seems to be missing support for positional parameters?

tulir: required parameters are positional

> * The position of parameter descriptions in the parameters property that are required is significant and clients SHOULD use the same order when prompting for arguments.
> * The position of parameter descriptions that are not required is not significant.

PR #4392, opened 2025-12-26
MSC4392: Encrypted reactions and replies

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/encrypted-reactions-and-replies/proposals/4392-encrypted-reactions-and-replies.md)

Ref https://github.com/matrix-org/matrix-spec/issues/660

### DIFF
--- a/proposals/4392-encrypted-reactions-and-replies.md
+++ b/proposals/4392-encrypted-reactions-and-replies.md
@@ -58,6 +58,14 @@ level of support for the encrypted reply metadata has been reached.
 Reactions could use a similar migration period, but it is uncertain whether
 current clients will accept encrypted `m.reaction` events at all.
 
+### Push notifications
+The proposed reaction format is indistinguishable from normal encrypted messages,
+which means servers would have to send wakeup pushes to mobile clients for all
+of them. This could be remediated with a separate flag in the unencrypted payload
+that indicates it should not trigger a push, e.g. `"m.priority": "low"`.
+[MSC3996](https://github.com/matrix-org/matrix-spec-proposals/pull/3996) has
+explored a similar problem for encrypted mentions-only rooms.
+
 ## Alternatives
 Some sort of breaking change is required to fix metadata leakage. It could be
 done as a part of a bigger breaking change, such as switching to extensible


proposals/4392-encrypted-reactions-and-replies.md
turt2live, line 1: Implementation requirements:

* [ ] Client, ideally multiple

krille-chan:
But this also means that every reaction triggers a push notification, which then has to be ignored on the client side, right? Any idea how we could avoid this? Especially for Apple devices, triggering push notifications which then do not result into a visible notification on the device this could lead to problems iirc and it is in general not very energy efficient if every reaction wakes up a device, decrypts a message only for the info that it can be ignored.

Maybe adding a flag to the unencrypted content, that this message does not necessary trigger a notification could be useful here? This could also then be used for future use cases like encrypted state events

proposals/4392-encrypted-reactions-and-replies.md
```diff
@@ -0,0 +1,75 @@
...
+that anyone needs server-side aggregations for replies, and therefore there is
+no reason to keep them unencrypted.
+
+When this document talks about the "unencrypted payload", it is referring to the
+cleartext/wire content that the homeserver sees. The "encrypted payload" is the
+content that was encrypted, which the homeserver can't see.
+
+## Proposal
+Clients SHOULD include the entire `m.relates_to` object in the encrypted payload
+of all events that include it.
```
anoadragon453, line 29: @krille-chan originally [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4392#issuecomment-3707057869) (PR threads should be used for discussion):

> But this also means that every reaction triggers a push notification, which then has to be ignored on the client side, right? Any idea how we could avoid this? Especially for Apple devices, triggering push notifications which then do not result into a visible notification on the device this could lead to problems iirc and it is in general not very energy efficient if every reaction wakes up a device, decrypts a message only for the info that it can be ignored.
> 
> Maybe adding a flag to the unencrypted content, that this message does not necessary trigger a notification could be useful here? This could also then be used for future use cases like encrypted state events

anoadragon453: This refers to the [`.m.rule.suppress_edits` default override rule](https://spec.matrix.org/v1.17/client-server-api/#predefined-rules:~:text=Suppresses%20notifications%20related%20to%20event%20replacements%2E) which suppresses notifications related to event replacements:

```
{
    "rule_id": ".m.rule.suppress_edits",
    "default": true,
    "enabled": true,
    "conditions": [
        {
            "kind": "event_property_is",
            "value": "m.replace"
        }
    ],
    "actions": []
}
```

This proposal would result in this push rule being ineffective in encrypted rooms. A push notification would always be sent to the client for edits (using some resources), but clients running the push rules locally would then not actually pop a notification as they'd be able to see the `m.relates_to` field in the decrypted event source.

Battery life vs. privacy of communications has come up before [during the Intentional Mentions MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/3952#discussion_r1112154200), and ultimately the SCT decided to prioritise privacy over battery life. MSC3996 is a proposal which aims to mitigate the problem somewhat by adding a boolean to the unencrypted portion that hints to the homeserver whether they should send a push or not, though that MSC has not gotten traction.

I would like to see this MSC at least mention the implications for push notifications.

tulir: Added a push notifications section to potential issues

krille-chan: @tulir thank you so much. I very like the `"m.priority": "low"` field. What do you think about make it even be used for mention only rooms with `"m.priority": "high"` which does the same as `"m.has_mentions": true` from MSC3996 with the intention that maybe other events as mentions could also make it a priority high or other things like edits could also become priority low, if we ever want to encrypt them as well

PR #3916, opened 2022-10-23
MSC3916: Authentication for media

richvdh:
A rewrite and update of MSC1902, and a precursor to MSC3911.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/authentication-for-media/proposals/3916-authentication-for-media.md)

Implementations:
* Server: https://github.com/turt2live/matrix-media-repo/pull/509
* Another server: https://github.com/element-hq/synapse/pull/17172
* Web client: https://github.com/element-hq/element-web/pull/27326

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/3916#issuecomment-2110908554)

PR #4396, opened 2026-01-06
[WIP] MSC4396: Inline linked media

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/inline-linked-media/proposals/4396-inlined-linked-media.md)

### DIFF
--- a/proposals/4396-inlined-linked-media.md
+++ b/proposals/4396-inlined-linked-media.md
@@ -1,0 +1,344 @@
+# MSC4396: Inline linked media
+
+**This proposal is a work in progress and may change dramatically.**
+
+Linking media to events is desirable to allow users to have their media deleted when their events
+are no longer accessible and to ensure that only those with visibility on an event can download the
+media for it.
+
+Like [MSC3911](https://github.com/matrix-org/matrix-spec-proposals/pull/3911), this proposal builds
+on the [authentication requirements](https://spec.matrix.org/v1.17/client-server-api/#content-repository:~:text=%5BChanged%20in%20v1.11%5D%20The,later%20version%20of%20the%20specification)
+added to the download endpoints in Matrix 1.11 by requiring that users also be *authorized* to access
+the media.
+
+The general concept of MSC3911 is preserved by this proposal: media *must* be linked to an event. The
+difference is how that linking happens. In MSC3911, media is "attached" to an event at send-time after
+being uploaded. In this proposal, media is uploaded at send-time and attached in the same operation.
+
+Combining the operations ensures there's a guaranteed link between media and event, but does also
+mean that media must be available to be combined in the operation. This also may lead to client
+compatibility issues because the sender will not be able to populate the `url` on an event itself.
+
+**Note**: Backwards compatibility is explored later in this proposal, but it's not perfect. More
+review and thought needs to happen here.
+
+The fact that an event contains media does not necessarily mean that event needs to be linked to
+that media, however. For example, in the case of an [MSC4027-style](https://github.com/matrix-org/matrix-spec-proposals/pull/4027)
+reaction, the reaction event is copying media that was already uploaded (and ideally, linked) to an
+event elsewhere. If that reaction event were to be redacted, the underlying media would not be affected,
+so it is not required to be linked to the event. If the event containing the original media were to
+be redacted however, most (if not all) reactions which used that media ID would stop working.
+
+## Adjacent use cases
+
+There are some use cases which are addressed by this MSC, but not big enough to be drivers for this
+specific design:
+
+* Combining the media upload and event sending allows safety tooling to more efficiently scan the
+  media. Currently, a safety tool will need to extract the media ID out of the event, download it,
+  scan it, then take relevant action. With the combined operation, the safety tool can skip the
+  extraction and download steps, improving response time.
+
+**TODO**: More?
+
+## Proposal
+
+[`POST /_matrix/media/v3/upload`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixmediav3upload)
+and [`POST /_matrix/media/v1/create`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixmediav1create)
+are **deprecated** with no direct replacement. [`PUT /_matrix/media/v3/upload/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixmediav3uploadservernamemediaid)
+is deprecated and moved to `PUT /_matrix/client/v1/media/upload/:serverName/:mediaId` with no other
+changes to its schema. This completes the transition started in [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916).
+
+The following endpoints receive an endpoint version bump (`v3` -> `v4` in most cases) to support a
+new request format, described below.
+
+* [`PUT /_matrix/client/v3/rooms/:roomId/send/:type/:txnId`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
+* [`PUT /_matrix/client/v3/rooms/:roomId/state/:type/(:stateKey)`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+* [`PUT /_matrix/client/v3/sendToDevice/:type/:txnId`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid)
+* [`PUT /_matrix/client/v3/profile/:userId/:keyName`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3profileuseridkeyname)
+
+**TODO**: More?
+
+In addition to supporting their already-specified schemas, these endpoints now support an additional
+(and optional) `multipart/mixed` body. This is the same approach as the Federation API's
+[media download response body](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1mediadownloadmediaid).
+
+The first part is *always* the JSON request body (as defined by the endpoint's normal schema). There
+MAY be zero or more parts denoting different media objects to assign to the event. Servers SHOULD
+limit how many media objects can be assigned to an event, and MUST NOT limit the caller to less than
+5 objects. This is to support media with thumbnails in rooms.
+
+An example request might be:
+
+```
+PUT /_matrix/client/v3/rooms/!x/send/m.room.message/1
+Authorization: Bearer TOKEN
+Content-Type: multipart/mixed; boundary=a1b2c3
+
+--a1b2c3
+Content-Type: application/json
+
+{
+  "msgtype": "m.image",
+  "body": "image.png",
+  "url_index": 0,
+  "info": {
+    "h": 1600,
+    "w": 1600,
+    "size": 206895,
+    "mimetype": "image/png",
+    "thumbnail_url_index": 1
+    "thumbnail_info": {
+      "h": 32,
+      "w": 32,
+      "size": 4950,
+      "mimetype": "image/png"
+    }
+  }
+}
+
+--a1b2c3
+Content-Type: image/png
+
+[png data]
+
+--a1b2c3
+Content-Type: image/png
+X-Matrix-Request-Async-File: true
+
+--a1b2c3
+```
+
+There are a few notable details in this example:
+
+1. `url` and `thumbnail_url` are replaced by `[thumbnail_]url_index`. This is the index of the media
+   URI in the `m.media` list (described later in this proposal).
+2. `X-Matrix-Request-Async-File` is a header on the third part. By setting this to `true`, the caller
+   is saying that it doesn't yet have the media and would like to use the (newly defined)
+   `PUT /_matrix/client/v1/media/upload/:serverName/:mediaId` endpoint instead. The caller discovers
+   the media ID to upload to via the endpoint's reponse.
+
+   The async media has the same time-to-upload and similar restrictions from the now-deprecated
+   `/create` endpoint.
+
+For this example, the response may be:
+
+```jsonc
+{
+  "event_id": "$x",
+  "media_uris": [
+    "mxc://example.org/one", // the first media part
+    "mxc://example.org/two"  // the second media part (which is an async upload)
+  ],
+  "unused_expires_at": 1647257217083 // for the async media parts
+}
+```
+
+Internally, the server mutates the resulting event's `content` to include an
+[MSC1767-style](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+`m.media` mixin.
+
+**Note**: The server *always* overwrites `m.media`, even if there's no media to assign to the event.
+If the `m.media` array is empty, the field SHOULD be deleted from the `content` instead.
+
+`m.media` is an array of linked (or assigned, or attached, or $synonym) media URIs, like so:
+
+```jsonc
+{
+  // other event fields omitted for brevity
+  "type": "m.room.message",
+  "content": {
+    "msgtype": "m.image",
+    "body": "image.png",
+    "url_index": 0,
+    "info": {
+      "thumbnail_url_index": 1
+      // and etc
+    },
+    "m.media": [
+      "mxc://example.org/one",
+      "mxc://example.org/two"
+    ]
+  }
+}
+```
+
+If the client does *not* need to link media to the event, they can use the normal `url` field instead.
+This will result in an event with no (or empty) `m.media` array, making that media immune to removal
+through this event.
+
+### Media removal
+
+Later, when an event with `m.media` becomes redacted (or effectively redacted via GDPR or similar),
+the media IDs in that array also become redacted/removed. It's left as an implementation detail whether
+to immediately delete the underlying file on the server, but the effect MUST be that attempts to
+download the media fail with a `410 Gone` HTTP status code and `M_GONE` error code.
+
+**Note**: `410 M_GONE` is new to the following endpoints:
+
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaid)
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId/:fileName`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaidfilename)
+* [`GET /_matrix/client/v1/media/thumbnail/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid)
+* [`GET /_matrix/federation/v1/media/download/:mediaId`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1mediadownloadmediaid)
+* [`GET /_matrix/federation/v1/media/thumbnail/:mediaId`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1mediathumbnailmediaid)
+* Deprecated variants of the above (if still in the spec when this MSC lands).
+
+Servers MAY also respond with a generic `404 M_NOT_FOUND` to hide the previous existence of media.
+
+Servers MAY allow server admins (or similar concept) to download redacted media for investigative or
+safety purposes.
+
+Clients SHOULD purge local caches of media upon redaction where possible.
+
+### Authorized downloads
+
+A user MUST NOT be able to download media from events they do not have visibility of. This affects:
+
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaid)
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId/:fileName`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaidfilename)
+* [`GET /_matrix/client/v1/media/thumbnail/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid)
+* Deprecated variants of the above (if still in the spec when this MSC lands).
+
+Users receive an `403 M_FORBIDDEN` error when downloading media associated with an event they cannot
+see.
+
+A user can "see" an event if the event is in a [`world_readable`](https://spec.matrix.org/v1.17/client-server-api/#room-history-visibility)
+room or the history visibility otherwise permits them to see the event.
+
+The visibility constraints are carried over federation using the currently-empty JSON object part
+on the download endpoints below:
+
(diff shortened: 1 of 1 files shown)


proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+on the [authentication requirements](https://spec.matrix.org/v1.17/client-server-api/#content-repository:~:text=%5BChanged%20in%20v1.11%5D%20The,later%20version%20of%20the%20specification)
+added to the download endpoints in Matrix 1.11 by requiring that users also be *authorized* to access
+the media.
+
+The general concept of MSC3911 is preserved by this proposal: media *must* be linked to an event. The
+difference is how that linking happens. In MSC3911, media is "attached" to an event at send-time after
+being uploaded. In this proposal, media is uploaded at send-time and attached in the same operation.
+
+Combining the operations ensures there's a guaranteed link between media and event, but does also
+mean that media must be available to be combined in the operation. This also may lead to client
```
deepbluev7, line 19: How do you intend to handle forwarding of events? Is media just re-uploaded?

Edits pose a similar problem, but that can be solved by media being attached to the original event at least. State events are slightly more problematic.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+**TODO**: Incorporate into the MSC properly.
+
+* to-device is supported for [MSC4268](https://github.com/matrix-org/matrix-spec-proposals/pull/4268)
+  use cases. It's unclear what the precise endpoint shape would be for this (hopefully similar to
+  what's already above).
+* more examples of how all of this works, including sequence diagrams
+
+## Potential issues
+
+**TODO**
```
deepbluev7, line 328: I'd be interested in a comparison to #3911, since editing json server side seems quite a bit more complicated and now clients need to use a 2 level scheme to get to the actual media file in some cases.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+Internally, the server mutates the resulting event's `content` to include an
+[MSC1767-style](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+`m.media` mixin.
+
+**Note**: The server *always* overwrites `m.media`, even if there's no media to assign to the event.
+If the `m.media` array is empty, the field SHOULD be deleted from the `content` instead.
+
+`m.media` is an array of linked (or assigned, or attached, or $synonym) media URIs, like so:
```
deepbluev7, line 144: Just to understand this correctly, how would I handle the case when I want to add an additional file to a state event? Do I have to reupload all media files or does the server merge the list of media files. It reads like the former, but that seems super wasteful?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+    "mxc://example.org/one", // the first media part
+    "mxc://example.org/two"  // the second media part (which is an async upload)
+  ],
+  "unused_expires_at": 1647257217083 // for the async media parts
+}
+```
+
+Internally, the server mutates the resulting event's `content` to include an
+[MSC1767-style](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+`m.media` mixin.
```
deepbluev7, line 139: What about text messages with img tags? Are those edited?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+is deprecated and moved to `PUT /_matrix/client/v1/media/upload/:serverName/:mediaId` with no other
+changes to its schema. This completes the transition started in [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916).
+
+The following endpoints receive an endpoint version bump (`v3` -> `v4` in most cases) to support a
+new request format, described below.
+
+* [`PUT /_matrix/client/v3/rooms/:roomId/send/:type/:txnId`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
+* [`PUT /_matrix/client/v3/rooms/:roomId/state/:type/(:stateKey)`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)
+* [`PUT /_matrix/client/v3/sendToDevice/:type/:txnId`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid)
+* [`PUT /_matrix/client/v3/profile/:userId/:keyName`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3profileuseridkeyname)
```
TheArcaneBrony, line 58: What about account data? Does it now become impossible to reference a file from account data?

proposals/4396-inlined-linked-media.md
turt2live: The intro currently serves this purpose.

proposals/4396-inlined-linked-media.md
turt2live: See later in proposal

proposals/4396-inlined-linked-media.md
turt2live: If a use case arises, it can use the room method.

proposals/4396-inlined-linked-media.md
turt2live: Events should only contain a single logical datum, so there shouldn't be multiple files in a single event (state or otherwise).

cloudrac3r:
Can you add a section for Motivation/why you want this? It looks to me like it would badly break backwards compatibility, and I'm not sure why this is needed (if the positives outweigh the drawbacks). So that I'm not missing anything, it would help if you directly explained the advantages that this MSC would provide.

turt2live:
@cloudrac3r top level comments on MSCs are often ignored/missed - please use the diff instead for future comments.

As this is a work in progress MSC, it is not expected to have all of the sections or details. The motivation is partially mentioned in the introduction already, and MSC3911 has a bit more direct detail.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+There are a few notable details in this example:
+
+1. `url` and `thumbnail_url` are replaced by `[thumbnail_]url_index`. This is the index of the media
+   URI in the `m.media` list (described later in this proposal).
+2. `X-Matrix-Request-Async-File` is a header on the third part. By setting this to `true`, the caller
+   is saying that it doesn't yet have the media and would like to use the (newly defined)
+   `PUT /_matrix/client/v1/media/upload/:serverName/:mediaId` endpoint instead. The caller discovers
+   the media ID to upload to via the endpoint's response.
```
Johennes, line 119: I think it's not necessarily that the caller doesn't yet have the media but more that it wants to upload it asynchronously so that the event itself can be sent and read by recipients without blocking on the upload?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+media for it.
+
+Like [MSC3911](https://github.com/matrix-org/matrix-spec-proposals/pull/3911), this proposal builds
+on the [authentication requirements](https://spec.matrix.org/v1.17/client-server-api/#content-repository:~:text=%5BChanged%20in%20v1.11%5D%20The,later%20version%20of%20the%20specification)
+added to the download endpoints in Matrix 1.11 by requiring that users also be *authorized* to access
+the media.
+
+The general concept of MSC3911 is preserved by this proposal: media *must* be linked to an event. The
+difference is how that linking happens. In MSC3911, media is "attached" to an event at send-time after
+being uploaded. In this proposal, media is uploaded at send-time and attached in the same operation.
```
Johennes, line 16: I found "media is uploaded at send-time" slightly misleading because the proposal still allows async uploads.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+
+A user MUST NOT be able to download media from events they do not have visibility of. This affects:
+
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaid)
+* [`GET /_matrix/client/v1/media/download/:serverName/:mediaId/:fileName`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediadownloadservernamemediaidfilename)
+* [`GET /_matrix/client/v1/media/thumbnail/:serverName/:mediaId`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid)
+* Deprecated variants of the above (if still in the spec when this MSC lands).
+
+Users receive an `403 M_FORBIDDEN` error when downloading media associated with an event they cannot
+see.
```
Johennes, line 203: Should this also optionally allow 404 as in the removal case above?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+
+There are some use cases which are addressed by this MSC, but not big enough to be drivers for this
+specific design:
+
+* Combining the media upload and event sending allows safety tooling to more efficiently scan the
+  media. Currently, a safety tool will need to extract the media ID out of the event, download it,
+  scan it, then take relevant action. With the combined operation, the safety tool can skip the
+  extraction and download steps, improving response time.
+
+**TODO**: More?
```
Johennes, line 42: It seems like the proposal only partially addresses the single listed use case because async uploads still exist. So more use cases would really be helpful here, in my opinion.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Profiles
+
+Profile changes often result in `m.room.member` events being sent. This proposal changes the
+[`PUT /_matrix/client/v3/profile/:userId/:keyName`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3profileuseridkeyname)
+endpoint to support the new `m.media` mixin, but requires special handling because the data is not
+in a traditional event shape. When `keyName` is `avatar_url`, the multipart request body schema still
+applies, though the JSON object part will always be an empty object. The first (and hopefully, only)
+media part will become the user's avatar, populating both the `avatar_url` and `m.media` fields on
+the resulting `m.room.member` event.
```
Johennes, line 279: I think a request example might help to better grasp this. It would look like this, if I understood correctly?

```
PUT /_matrix/client/v3/profile/{userId}/avatar_url
Authorization: Bearer TOKEN
Content-Type: multipart/mixed; boundary=a1b2c3

--a1b2c3
Content-Type: application/json

{}

--a1b2c3
Content-Type: image/png

[png data]
```

I assume async uploads are not allowed here?

Does the server echo back `media_uris` in the response?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+MAY be zero or more parts denoting different media objects to assign to the event. Servers SHOULD
+limit how many media objects can be assigned to an event, and MUST NOT limit the caller to less than
+5 objects. This is to support media with thumbnails in rooms.
+
+An example request might be:
+
+```
+PUT /_matrix/client/v3/rooms/!x/send/m.room.message/1
+Authorization: Bearer TOKEN
+Content-Type: multipart/mixed; boundary=a1b2c3
```
kegsay, line 76: Using `multipart/mixed` has all the same downsides I [mentioned on MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916#discussion_r1600690488), except now we're asking all _clients_ to do this not just servers. From discussion on that thread, it seems to indicate that Range headers aren't possible, which would be very sad as it's common to want range headers to resume file uploads on lossy networks. That being said, [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests#multipart_ranges) seems to think it is possible so I'm confused.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+  "content": {
+    "msgtype": "m.image",
+    "body": "image.png",
+    "url_index": 0,
+    "info": {
+      "thumbnail_url_index": 1
+      // and etc
+    },
+    "m.media": [
+      "mxc://example.org/one",
```
kegsay, line 159: Is this protected from redaction? Limits on the length of the array?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+  "event_id": "$x",
+  "media_uris": [
+    "mxc://example.org/one", // the first media part
+    "mxc://example.org/two"  // the second media part (which is an async upload)
+  ],
+  "unused_expires_at": 1647257217083 // for the async media parts
+}
+```
+
+Internally, the server mutates the resulting event's `content` to include an
```
kegsay, line 137: This materially makes metadata protection worse as media URIs have to be in the clear for the server to do this.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+```jsonc
+{
+  // other event fields omitted for brevity
+  "type": "m.room.message",
+  "content": {
+    "msgtype": "m.image",
+    "body": "image.png",
+    "url_index": 0,
+    "info": {
+      "thumbnail_url_index": 1
```
kegsay, line 155: Unclear how this is set, CTRL+F for `thumbnail_url_index` only use it in examples..?

proposals/4396-inlined-linked-media.md
HarHarLinks: We use multiple media per event in NeoBoard https://github.com/nordeck/matrix-neoboard/blob/d43423112fbeae23525307a61fa6213dbd2f3f53/docs/model/crdt-documents.md#imageelement

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+For the purposes of this proposal, "(custom) emoji" encompasses inline emoji via
+[mxc-`src`'d `img` tags](https://spec.matrix.org/v1.17/client-server-api/#mroommessage-msgtypes),
+[MSC4027-style reactions](https://github.com/matrix-org/matrix-spec-proposals/pull/4027),
+and [stickers](https://spec.matrix.org/v1.17/client-server-api/#sticker-messages).
+
+With the current [MSC2545-led](https://github.com/matrix-org/matrix-spec-proposals/pull/2545) direction,
+emoji is compiled into "packs" which may be reused by other users. In this setup, media uploads would
+be linked to the events in the packs rather than when those emoji are used.
+
+When a pack event is redacted, the associated emoji would stop working wherever it was used.
```
HarHarLinks, line 266: This is very unfortunate. Is this expected UX on other platforms?

For unencrypted rooms, deduplicating should be possible by content hash. In encrypted rooms, perhaps there could be a compromise between reuploading each sticker each time it is sent and full metadata leak?

proposals/4396-inlined-linked-media.md
networkException: Could [MSC3911's](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/linking_media_to_events/proposals/3911-linking-media-to-events.md#detailed-spec-changes) 'New "media copy" API' simply be reused here?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+If the client does *not* need to link media to the event, they can use the normal `url` field instead.
+This will result in an event with no (or empty) `m.media` array, making that media immune to removal
+through this event.
+
+### Media removal
+
+Later, when an event with `m.media` becomes redacted (or effectively redacted via GDPR or similar),
+the media IDs in that array also become redacted/removed. It's left as an implementation detail whether
+to immediately delete the underlying file on the server, but the effect MUST be that attempts to
+download the media fail with a `410 Gone` HTTP status code and `M_GONE` error code.
```
networkException, line 175: I believe this MSC should include explicit handling for all local users leaving a room, given it's mentioned as a problem in [MSC3911](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/linking_media_to_events/proposals/3911-linking-media-to-events.md#msc3911-linking-media-to-events) and by that would be an incremental improvement over [MSC4322](https://github.com/timedoutuk/matrix-spec-proposals/blob/nexy7574/simple-media-self-redaction/proposals/4322-simple-media-self-redaction.md)

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+### Custom emoji
+
+For the purposes of this proposal, "(custom) emoji" encompasses inline emoji via
+[mxc-`src`'d `img` tags](https://spec.matrix.org/v1.17/client-server-api/#mroommessage-msgtypes),
+[MSC4027-style reactions](https://github.com/matrix-org/matrix-spec-proposals/pull/4027),
+and [stickers](https://spec.matrix.org/v1.17/client-server-api/#sticker-messages).
+
+With the current [MSC2545-led](https://github.com/matrix-org/matrix-spec-proposals/pull/2545) direction,
+emoji is compiled into "packs" which may be reused by other users. In this setup, media uploads would
+be linked to the events in the packs rather than when those emoji are used.
```
reivilibre, line 264: It feels odd/wrong to link the events to the packs, as this means if I have an emoji pack then I need to invite all potential receivers of the emoji to my room so that they would be able to see the emoji when I send them.

Current usecases of emoji packs don't require the emoji pack to be joined by a potential receiver; I suspect (but am not sure) that there is a tension with personal packs too.

proposals/4396-inlined-linked-media.md
reivilibre: Would this restriction not apply to any linked media scheme?

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+X-Matrix-Request-Async-File: true
+
+--a1b2c3
+```
+
+There are a few notable details in this example:
+
+1. `url` and `thumbnail_url` are replaced by `[thumbnail_]url_index`. This is the index of the media
+   URI in the `m.media` list (described later in this proposal).
+2. `X-Matrix-Request-Async-File` is a header on the third part. By setting this to `true`, the caller
```
reivilibre, line 116: I believe we have pointed out before that the `X-` header prefix is obsolete and this should just be `Matrix-..`

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+
+If the calling server doesn't have any users which would be able to see the event, the federation
+download instead fails with `403 M_FORBIDDEN`.
+
+### Encryption
+
+In encrypted rooms, media is first encrypted before being referenced in an event. This capability is
+retained, though the `m.media` mixin will be on the unencrypted outer portion of the event. To reduce
+(though not eliminate) metadata risk of observers guessing at which of the URIs is the original and
+which is a thumbnail, the event contents use `(thumbnail_)url_index` fields which reference a specific
```
reivilibre, line 248: seems like theatre, given the body length should give it away

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+If the client does *not* need to link media to the event, they can use the normal `url` field instead.
+This will result in an event with no (or empty) `m.media` array, making that media immune to removal
+through this event.
+
+### Media removal
+
+Later, when an event with `m.media` becomes redacted (or effectively redacted via GDPR or similar),
+the media IDs in that array also become redacted/removed. It's left as an implementation detail whether
```
reivilibre, line 173: I might be missing something, but I think this is not quite right.

It's the media's own association that we need to trust, not the `m.media` content of an event.

Nothing stops me from putting your media URI in *my* event's `m.media` section and then redacting it, but that doesn't mean it should delete your media.

proposals/4396-inlined-linked-media.md
reivilibre: I suspect 'media copy' (maybe in bulk) like from MSC3911 would be beneficial for this case. (Linking to another comment already suggesting this MSC regain that: https://github.com/matrix-org/matrix-spec-proposals/pull/4396/files#r3531487947)

proposals/4396-inlined-linked-media.md
reivilibre: - leaving → forgetting (as left users can still read history and should still be able to see media therefore)
- I think this is basically best implied as part of the cleanup that homeservers do when removing rooms forgotten by everyone locally. (But I suppose I'm not against stating it explicitly)

proposals/4396-inlined-linked-media.md
reivilibre: I guess 'media copy' (maybe in bulk) like from MSC3911 would **also** be beneficial for this case. (Linking to another comment already suggesting this MSC regain that: https://github.com/matrix-org/matrix-spec-proposals/pull/4396/files#r3531487947)

That way the emoji media lifetime would be bound to the event that it contains, including the visibility.

proposals/4396-inlined-linked-media.md
reivilibre: I guess 'media copy' (maybe in bulk) like from MSC3911 would **also** be beneficial for this case. (Linking to another comment already suggesting this MSC regain that: https://github.com/matrix-org/matrix-spec-proposals/pull/4396/files#r3531487947)

That way the emoji media lifetime would be bound to the event that it contains, including the visibility.

proposals/4396-inlined-linked-media.md
reivilibre: It feels like it'd be pretty clean to allow, as an alternative to uploading a piece of media into the 'slot' for an event, specifying a media URI to 'copy' (like MSC3911 but still inline in the request sending an event).

That appears to solve most of the practical problems raised on other threads of this MSC.

proposals/4396-inlined-linked-media.md
```diff
@@ -0,0 +1,344 @@
...
+
+The first part is *always* the JSON request body (as defined by the endpoint's normal schema). There
+MAY be zero or more parts denoting different media objects to assign to the event. Servers SHOULD
+limit how many media objects can be assigned to an event, and MUST NOT limit the caller to less than
+5 objects. This is to support media with thumbnails in rooms.
+
+An example request might be:
+
+```
+PUT /_matrix/client/v3/rooms/!x/send/m.room.message/1
```
reivilibre, line 74: I think it might be worth changing this to `/v4` here and moving the event content one level deeper into the request body.

We could then put in parameters for doing a MSC3911-style 'media copy' ([see thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4396/files#r3531487947))

Or alternatively we could allow uploading temporary media and then linking it to the event on send, which might be an easier alternative for some client devs, addressing [this comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4396/files#r2757979966)

proposals/4396-inlined-linked-media.md
HarHarLinks: On the other hand, media copy would not be practical for custom reaction emoji, if everyone reacting with the same emoji sends a new mxc.

proposals/4396-inlined-linked-media.md
cloudrac3r: > Events should only contain a single logical datum, so there shouldn't be multiple files in a single event (state or otherwise).

There are multiple files in state events for custom emoji packs (MSC now merged).

PR #4400, opened 2026-01-14
MSC4400: Remove the depth field from PDUs

1vel:
[Rendered](https://github.com/velikopter/matrix-spec-proposals/blob/klabammo_depth/proposals/4400-remove-depth.md)

Implementations:
- Unstable room version:
- Complement tests:

### DIFF
--- a/proposals/4400-remove-depth.md
+++ b/proposals/4400-remove-depth.md
@@ -8,6 +8,7 @@ timeline ordering.
 ## Proposal
 
 Remove the depth field from PDUs.
+`depth` is also no longer included after an event has been redacted.
 
 This change would result in a PDU now looking like this:
 
--- a/proposals/4400-remove-depth.md
+++ b/proposals/4400-remove-depth.md
@@ -1,4 +1,4 @@
-# MSC4398: Remove the depth field from PDUs
+# MSC4400: Remove the depth field from PDUs
 
 The depth field is useless ever since room v3 (possibly
 v2?) and seems to only cause problems, especially on
@@ -57,7 +57,7 @@ None.
 
 ## Unstable prefix
 
-The room version for this MSC is `com.nhjkl.msc4398.opt1`, and it is
+The room version for this MSC is `com.nhjkl.msc4400.opt1`, and it is
 based on room version `12`.
 
 ## Dependencies
--- a/proposals/4400-remove-depth.md
+++ b/proposals/4400-remove-depth.md
@@ -9,6 +9,7 @@ timeline ordering.
 
 Remove the depth field from PDUs.
 `depth` is also no longer included after an event has been redacted.
+This will require a new room version.
 
 This change would result in a PDU now looking like this:
 
--- a/proposals/4398-remove-depth.md
+++ b/proposals/4398-remove-depth.md
@@ -1,0 +1,65 @@
+# MSC4398: Remove the depth field from PDUs
+
+The depth field is useless ever since room v3 (possibly
+v2?) and seems to only cause problems, especially on
+Synapse servers due to Synapse using the depth field for
+timeline ordering.
+
+## Proposal
+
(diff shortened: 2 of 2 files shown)


proposals/4400-remove-depth.md
turt2live, line 1: Implementation requirements:

* [ ] Unstable room version
* [ ] Complement tests

proposals/4400-remove-depth.md
```diff
@@ -0,0 +1,66 @@
...
+
+The depth field is useless ever since room v3 (possibly
+v2?) and seems to only cause problems, especially on
+Synapse servers due to Synapse using the depth field for
+timeline ordering.
+
+## Proposal
+
+Remove the depth field from PDUs.
+`depth` is also no longer included after an event has been redacted.
```
turt2live, line 11:
```suggestion
Remove the depth field from PDUs.
`depth` is also no longer included after an event has been redacted.
This will require a new room version.
```

proposals/4400-remove-depth.md
turt2live, line 1: [Sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md) will be required for this to pass into the merge FCP stages.

turt2live (changes requested):
I have not read the MSC in detail - this is just an early checklist check.

proposals/4400-remove-depth.md
```diff
@@ -0,0 +1,67 @@
...
+  "unsigned": {
+    "age": 4612
+  }
+}
+```
+
+## Potential issues
+
+Synapse would need to change how it handles timeline ordering.
+Other than that, none.
```
Kladki, line 50: What kept me from making an MSC like this is that depth is used on [`/get_missing_events`](https://spec.matrix.org/v1.17/server-server-api/#post_matrixfederationv1get_missing_eventsroomid), but I'm not sure if any server actually makes use of the parameter here.

timedoutuk: Only synapse even implements it. Last I checked, dendrite and the conduit family tree do not support the depth parameter at all, and synapse doesn't trust it in the first place if I recall

turt2live: This is still a reason that we may not be able to remove it fully. The endpoint will need cleaning up before we can remove `depth` from PDUs. Especially if servers aren't using it or aren't actually relying on the `depth` the spec (supposedly) wants them to use.

If the changes are small, they can be included in this very MSC.

turt2live (commented):
Sorry for the long delay in review. I've added some thoughts as an individual on the SCT - other SCT members may have other opinions.

In general: I'm in favour of this proposal, but it looks like we have some cleanup to do.

proposals/4400-remove-depth.md
gamesguru: Is reported/intended `depth` easily reconstructable if omitted? Or is it "lost" metadata?

For the potential benefit of forensic analysis alone, I am hesitant to omit something unreconstructable (or only sloppily reconstructable).

If the intention is `depth`'s exclusion from canonical PDU JSONs and post-processing (whether it be sorting, state res, hashing, signing, redacting, and/or something else), perhaps this is better accomplished with single underscore or "dunder" prefix syntax, e.g., `__depth: 1337`. Or if there is another equivalent way to do this in the codebases? If `depth` is sometimes causing reported issues, it may be useful to perform a combined code audit of common homeservers for endpoint and algorithm consistency & correctness (either C2S or S2S implementations).

I'm unaware of the any other motivations behind this proposal (appservice/bot? housekeeping? something else?). But the amount of extra text/data consumed in my opinion is not an issue. If current versions of Synapse use `depth`, it's possible even more optimized endpoints will rely on it, too, and that we should keep it around in some convenient way that compromises with the author's intentions and with those in favor of the change.

PR #4401, opened 2026-01-15
MSC4401: Publishing client capabilities via profiles

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/profile-capabilities/proposals/4401-profile-capabilities.md)

### DIFF
--- a/proposals/4401-profile-capabilities.md
+++ b/proposals/4401-profile-capabilities.md
@@ -28,9 +28,12 @@ is an object with the following properties:
 - `events` (array): A list of objects containing details about the events this device supports.
   - `type` (string): The type of the event.
   - `content` (array): An optional list of objects describing additional requirements for properties
-    inside the `content` of the event.
+    inside the `content` of the event. This is helpful for "container events" where a secondary
+    type-like property beneath `content` determines the full schema of the event content. An example
+    for this is `msgtype` in `m.room.message` or the properties beneath
+    `m.fhir.structure_definition` in the `m.fhir` event from [MSC4302].
     - `key` (string, required): The dot-separated path of the property (analogous to `key` in
-      `event_match` [push rule conditions]).
+      `event_match` [push rule conditions] but starting at `content` rather than the event root).
     - `value` (string, required): The exact value of the property.
 
 As an example, here is a profile that advertises support for processing medication statements using
@@ -88,6 +91,6 @@ While this MSC is not considered stable, `m.client_capability.*` should be refer
 None.
 
   [MSC1767]: https://github.com/matrix-org/matrix-spec-proposals/pull/1767
-  [push rule conditions]: https://spec.matrix.org/v1.16/client-server-api/#conditions-1
   [MSC4302]: https://github.com/matrix-org/matrix-spec-proposals/pull/4302
+  [push rule conditions]: https://spec.matrix.org/v1.16/client-server-api/#conditions-1
   [MSC4301]: https://github.com/matrix-org/matrix-spec-proposals/pull/4301


proposals/4401-profile-capabilities.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Server

proposals/4401-profile-capabilities.md
```diff
@@ -0,0 +1,91 @@
...
+      }
+    ]
+  }
+}
+```
+
+## Potential issues
+
+Profiles are currently limited to 64 KiB in size which may be reached through excessive usage of
+this feature.
```
tulir, line 67: Are client capability entries cleaned up when a device is logged out? If not, the limit seems more like "will be reached" than "may be reached"

Johennes: Ah, good point. I've added server-side purging on logout with 2fac16c0b7661be853febeb01493cb61ee4cb7f1.

proposals/4401-profile-capabilities.md
```diff
@@ -0,0 +1,93 @@
...
+      }
+    ]
+  }
+}
+```
+
+To avoid stale capabilities filling up the profile, servers SHOULD remove a device's
+`m.client_capability.{DEVICE_ID}` key when the device is logged out.
+
+## Potential issues
```
clokep, line 64: Is there a risk of overflowing profile data?

proposals/4401-profile-capabilities.md
```diff
@@ -0,0 +1,93 @@
...
+for events via profiles.
+
+## Proposal
+
+A new set of profile keys `m.client_capability.{DEVICE_ID}` is introduced. The value of these keys
+is an object with the following properties:
+
+- `events` (array): A list of objects containing details about the events this device supports.
+  - `type` (string): The type of the event.
+  - `content` (array): An optional list of objects describing additional requirements for properties
```
clokep, line 30: I'm not following what this is used for, what does the key/value mean?

Johennes: It's similar to the `event_match` push rule condition and lets you match on fields in the event other than the `type`.

For instance, when you have a "container event" like `m.fhir` from https://github.com/matrix-org/matrix-spec-proposals/pull/4302 you could match on `content.m\.fhir\.structure_definition.url` and `content.m\.fhir\.structure_definition.version` to express that you only support FHIR resources of a specific type and version.

Or, as a simpler but more contrived example, you could express that you support `m.room.message` but only with certain values of `msgtype`.

proposals/4401-profile-capabilities.md
Johennes: Do you mean something different than the 64 KiB limit on the overall profile size?

clokep: I have no idea how I just skipped the sentence below this. 🤦‍♂️

proposals/4401-profile-capabilities.md
Johennes: Have copied part of this into the proposal with b2790ab0fbbd84b0d55db5d88277abb8163b126a.

proposals/4401-profile-capabilities.md
```diff
@@ -0,0 +1,96 @@
+# MSC4401: Publishing client capabilities via profiles
```
kegsay, line 1: I like the concept of client capabilities and wish we had it for MSC4428 / MSC4430 so clients could advertise "I support room versions 1-12" such that servers could filter out unsupported rooms in `/sync`.

I'm not convinced using profiles is the right place for this data though. I see your use case with needing other clients to have visibility on the capabilities, but profiles are per-user data structures (so you've got book-keeping in this MSC to namespace by device ID, which feels inelegant).

Johennes: Yes, proper per-device profiles would be better, agreed. It's the same problem with device-specific features that build on account data such as https://github.com/matrix-org/matrix-spec-proposals/pull/3890.

PR #4403, opened 2026-01-15
MSC4403: Forbid `event_id` on PDUs received over federation

1vel:
[Rendered](https://github.com/velikopter/matrix-spec-proposals/blob/rmeventid/proposals/4403-forbid-event_id.md)

Implementations:
- Unstable room version:
  - Calyx: [room version](https://foundry.fsky.io/vel/calyx/src/commit/5ef3ca34986202931b4b73a53d026f524008947e/synapse/api/room_versions.py#L541), [event checks](https://foundry.fsky.io/vel/calyx/src/commit/5ef3ca34986202931b4b73a53d026f524008947e/synapse/event_auth.py#L122) 
- Complement tests:

### DIFF
--- a/proposals/4403-forbid-event_id.md
+++ b/proposals/4403-forbid-event_id.md
@@ -1,16 +1,16 @@
-# MSC4403: Forbid `event_id` on events
+# MSC4403: Forbid `event_id` on PDUs received over federation
 
 The top-level `event_id` key has no use ever since room
 version 3, and there is also no defined behaviour in spec
-on how to handle the key if it is present on a v3+ event.
+on how to handle the key if it is present on a v3+ PDU.
 This has led to some implementations outright rejecting
-these events, even though they are perfectly valid.
+these PDUs, even though they are perfectly valid.
 
 ## Proposal
 
-Forbid `event_id` from appearing on the top-level of events. 
-This also removes it from the protected keys list in the
-redaction algorithm.
+Forbid `event_id` from appearing on the top-level of PDUs
+received over federation. This also removes it from the
+protected keys list in the redaction algorithm.
 
 This will require a new room version.
 
--- a/proposals/4403-forbid-event_id.md
+++ b/proposals/4403-forbid-event_id.md
@@ -12,6 +12,9 @@ Forbid `event_id` from appearing on the top-level of PDUs
 received over federation. This also removes it from the
 protected keys list in the redaction algorithm.
 
+A new rule is added to the top of the auth rules:
+> If the event has an `event_id`, reject.
+
 This will require a new room version.
 
 ## Potential issues
--- a/proposals/4403-forbid-event_id.md
+++ b/proposals/4403-forbid-event_id.md
@@ -1,4 +1,4 @@
-# MSC4403: Remove `event_id` from redacted events
+# MSC4403: Forbid `event_id` on events
 
 The top-level `event_id` key has no use ever since room
 version 3, and there is also no defined behaviour in spec
@@ -8,7 +8,11 @@ these events, even though they are perfectly valid.
 
 ## Proposal
 
-Remove the `event_id` key from redacted events. This will require a new room version.
+Forbid `event_id` from appearing on the top-level of events. 
+This also removes it from the protected keys list in the
+redaction algorithm.
+
+This will require a new room version.
 
 ## Potential issues
 
@@ -24,7 +28,7 @@ None.
 
 ## Unstable prefix
 
-The room version for this proposal is `com.nhjkl.msc4403.opt1`,
+The room version for this proposal is `com.nhjkl.msc4403.opt2`,
 and it is based on room version 12.
 
 ## Dependencies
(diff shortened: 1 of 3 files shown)


proposals/4403-forbid-event_id.md
turt2live, line 1: Implementation requirements:

* [ ] Unstable room version
* [ ] Complement tests

proposals/4403-remove-event_id.md
```diff
@@ -0,0 +1,32 @@
...
+
+The top-level `event_id` key has no use ever since room
+version 3, and there is also no defined behaviour in spec
+on how to handle the key if it is present on a v3+ event.
+This has led to some implementations outright rejecting
+these events, even though they are perfectly valid.
+
+## Proposal
+
+Remove the `event_id` key from redacted events.
```
turt2live:
```suggestion
Remove the `event_id` key from redacted events. This will require a new room version.
```

proposals/4403-forbid-event_id.md
turt2live, line 1: [Sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md) will be required for this to pass into the merge FCP stages.

turt2live (changes requested):
I have not read the MSC in detail - this is just an early checklist check.

proposals/4403-forbid-event_id.md
```diff
@@ -0,0 +1,32 @@
...
+
+Remove the `event_id` key from redacted events. This will require a new room version.
+
+## Potential issues
+
+None?
+
+## Alternatives
+
+None.
```
tulir, line 26: Forbidding `event_id` entirely in federation events might make more sense. v12 `m.room.create` already forbids `room_id`

1vel: IMO, not really. Keys that were removed from the redaction algorithm before weren't outright forbidden from ever appearing in an event again, I don't see why `event_id` should be special here.

tulir: > I don't see why `event_id` should be special here.

perhaps because event and room IDs are very special cases and not at all similar to other redaction-protected fields?

Event IDs weren't removed from the spec, the event still has an ID, it's just not transmitted over federation. Servers have to overwrite that field before serving it to clients, so there's no reason to allow the field in the first place.

1vel: Fair enough point. Should I rescope this proposal or make a new one?

tulir: Up to you, authors can make basically any changes they want until proposed FCP

proposals/4403-forbid-event_id.md
```diff
@@ -0,0 +1,36 @@
...
+
+The top-level `event_id` key has no use ever since room
+version 3, and there is also no defined behaviour in spec
+on how to handle the key if it is present on a v3+ event.
+This has led to some implementations outright rejecting
+these events, even though they are perfectly valid.
+
+## Proposal
+
+Forbid `event_id` from appearing on the top-level of events.
```
anoadragon453: Which status code and `errcode` should homeservers return in this case?

tulir: I guess it should be an auth rule like rejecting `room_id` is? rule 1.2 in https://spec.matrix.org/v1.17/rooms/v12/#authorization-rules

1vel: It would be an auth rule, yes. I'm unsure where else it _could_ go.

proposals/4403-forbid-event_id.md
```diff
@@ -0,0 +1,36 @@
+# MSC4403: Forbid `event_id` on events
```
zecakeh: It believe this proposal would be clearer if it referred to "PDUs" rather than "events".

"event" is a general term which has several different formats across the various APIs, while "PDU" is the specific event format that is targeted here.

Besides, several event formats in the Client-Server API have an `event_id`, so it is confusing.

Half-Shot: +1 I'm afraid this proposal made no sense to me because as a client developer we do use the event_id field :). Can it be made clear which bits of the spec this refers to.

proposals/4403-forbid-event_id.md
turt2live, line 1: This MSC is confusing (and needs links to the existing spec, but that's a secondary point). Modern room versions already don't have `event_id` as a field on events, but they *do* protect the field from redaction. This MSC doesn't address redaction, but does add the field as protected by the auth rules. In either case, the field is not supposed to be on the event in the first place.

Altering the redaction algorithm still requires a new room version, but would be a direction that is more understandable to me. If there's a reason to also update the auth rules, that needs to be described in more detail in the MSC - auth rules are typically not used for this purpose.

turt2live (commented):
Sorry for the long review time in the queue. I've left some thoughts as an individual on the SCT - other SCT members might have different opinions. 

Overall: the idea of dropping `event_id` feels correct, but the approach feels wrong.

PR #4404, opened 2026-01-20
MSC4404: Compare emoji by name rather than image

ara4n:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/msc4404/proposals/4404-compare-emoji-by-name.md)

Alternative to [MSC4347](https://github.com/matrix-org/matrix-spec-proposals/pull/4347)

### DIFF
--- a/proposals/4404-compare-emoji-by-name.md
+++ b/proposals/4404-compare-emoji-by-name.md
@@ -1,4 +1,4 @@
-x# MSC4404: Compare emoji by name rather than image
+# MSC4404: Compare emoji by name rather than image
 
 ## Problem
 


proposals/4404-compare-emoji-by-name.md
turt2live, line 1: Implementation requirements:
* [ ] Client (sending)
* [ ] Client (receiving)

proposals/4404-compare-emoji-by-name.md
```diff
@@ -0,0 +1,104 @@
...
+      🍎      ⌛       ⚓
+    Apple  Hourglass Anchor
+   (Apfel) (Sanduhr) (Anker)
+
+```
+
+We can also recommend that emoji are typically presented in a row of 4 and a row of 3 for
+consistency and ease of comparison, but this would be a MAY rather than a MUST.
+
+## Potential issues
```
Johennes, line 64: One downside of including the translations is that we're increasing the amount of information on the comparison screen and thereby possibly the user confusion about what to actually check for. As an extreme example, you might now think that you're required to compare the images and both sets of strings, one of them possibly being in a language you don't understand.

proposals/4404-compare-emoji-by-name.md
```diff
@@ -0,0 +1,104 @@
...
+   dictionary for support.
+ * This (optionally) leaks the language of the user to the peer being verified (and the servers
+   in between), which could be quite sensitive in terms of outing people's nationalities and
+   implied geolocale.
+
+## Alternatives
+
+ * Emoji comparison is basically flawed, and while fun and quirky, has awkward edge cases like this.
+   We could/should deprecate it in favour of simple decimal SAS comparison, but we're stuck with
+   emoji for compatibility, hence this incremental fix.
```
Johennes, line 82: We cannot directly remove it from the spec but why not deprecate it now? The [`decimal`](https://spec.matrix.org/v1.17/client-server-api/#sas-method-decimal) method requires comparing 3 numbers of 4 digits each. That should be trivial to explain to users and has far fewer localization problems.

Ideally shorter numbers would suffice but this still feels less ambiguous and less mentally overloading than emoji + string comparison.

kegsay: I agree. The developer effort to implement and land this MSC (`accept_languages` and translations) vs the developer effort to use `decimal` seem roughly equivalent. Both require wire-format changes. Both require UI changes.

Therefore, we should deprecate emoji verification and encourage client developers to implement `decimal`, rather than putting 💄 on a 🐷,

richvdh:
https://github.com/matrix-org/matrix-spec-proposals/pull/4405 proposes to deprecate emoji verification rather than trying to improve it.

ara4n:
closing in favour of #4405

PR #4405, opened 2026-01-20
MSC4405: Deprecate the emoji method for SAS verification

uhoreg:
[Rendered](https://github.com/uhoreg/matrix-doc/blob/deprecate_emoji_verification/proposals/4405-deprecate-emoji-verification.md)

Disclosure: I am a member of the Spec Core Team, and the Element crypto team.  This proposal was written with my Spec Core Team hat on.

----

SCT Stuff:

[MSC Checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4405#issuecomment-4500662453)

[FCP Tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4405#issuecomment-4500672023)

Designated reviewers:

* [x] Crypto team
* [ ] @ara4n with a Product hat

### DIFF
--- a/proposals/4405-deprecate-emoji-verification.md
+++ b/proposals/4405-deprecate-emoji-verification.md
@@ -105,16 +105,19 @@ perceived usability of using emojis.
 ## Alternatives
 
 [MSC4347](https://github.com/matrix-org/matrix-spec-proposals/issues/4347)
-proposes to standardize the visual representation of the emojis.
+proposes to standardize the visual representation of the emojis.  This MSC tries
+to fix the issue of differences in emoji representation, but does not fix the
+translation issue.
+
 
 [MSC4404](https://github.com/matrix-org/matrix-spec-proposals/pull/4404)
 proposes to specify that the emoji names should be used rather than the visual
-representations.  As a historical note, this would in some sense, make emoji
-verification more similar to the original ZRTP, which SAS verification was based
-on, and which uses a sequence of words as the Short Authentication String.
+representations.  This MSC tries to fix the both the issue of differences in
+emoji representation, and the translation issue.  As a historical note, this
+would in some sense, make emoji verification more similar to the original ZRTP,
+which SAS verification was based on, and which uses a sequence of words as the
+Short Authentication String.
 
-These two MSCs try to fix the issue of differences in emoji representation, but
-do not fix the translation issue.
 
 ## Security considerations
 
--- a/proposals/4405-deprecate-emoji-verification.md
+++ b/proposals/4405-deprecate-emoji-verification.md
@@ -1,4 +1,4 @@
-# MSCxxxx: Deprecate the emoji method for SAS verification
+# MSC4405: Deprecate the emoji method for SAS verification
 
 Emojis are cute and fun.  They can be used to make things seem more friendly,
 and to make a very technical process feel more approachable.  That was the hope
--- a/proposals/4405-deprecate-emoji-verification.md
+++ b/proposals/4405-deprecate-emoji-verification.md
@@ -45,7 +45,7 @@ In addition, we made at least one error in choosing emojis: 🔧 is "spanner" in
 en_GB, but "wrench" in en_US, so there may even be confusion between users
 speaking the same language.  The spec picks "spanner" as the name, and if a
 British and American user are verifying each other, the British user can explain
-it to the American user.  But if two American users verifying each other may
+it to the American user.  But if two American users are verifying each other, they may
 need to figure out on their own what a "spanner" is.  (✈️ suffers a similar
 problem, but less severely, since it is likely that users can figure out that
 "aeroplane" is the same as "airplane".)
@@ -57,6 +57,15 @@ terminal-based clients) may not have a choice in which font is used.  Thus users
 may not see the exact same images on the clients, causing them to wonder if the
 difference is significant.  The standardised names were intended to help with
 this, but in practice, it turns out that people still get confused.
+
+#### Complexity of user instructions
+
+It can be difficult to formulate instructions for what to actually compare during
+emoji verification for non-technical users. For instance, users have to understand
+that they don't need to compare the emoji graphics pixel by pixel. They also need
+to understand that they don't need to compare the emoji names letter by letter
+because they might not be in the same language. This complexity undermines the
+anticipated simplicity that emoji verification was meant to provide.
 
 ### Decimal numbers do not have the same issues
 
@@ -112,7 +121,7 @@ translation issue.
 
 [MSC4404](https://github.com/matrix-org/matrix-spec-proposals/pull/4404)
 proposes to specify that the emoji names should be used rather than the visual
-representations.  This MSC tries to fix the both the issue of differences in
+representations.  This MSC tries to fix both the issue of differences in
 emoji representation, and the translation issue.  As a historical note, this
 would in some sense, make emoji verification more similar to the original ZRTP,
 which SAS verification was based on, and which uses a sequence of words as the
--- a/proposals/xxxx-deprecate-emoji-verification.md
+++ b/proposals/xxxx-deprecate-emoji-verification.md
@@ -1,0 +1,129 @@
+# MSCxxxx: Deprecate the emoji method for SAS verification
+
+Emojis are cute and fun.  They can be used to make things seem more friendly,
+and to make a very technical process feel more approachable.  That was the hope
+when the emoji method was introduced for [Short Authentication String (SAS)
+verification](https://spec.matrix.org/v1.17/client-server-api/#short-authentication-string-sas-verification).
+
+In practice, emojis present some usability problems.
+
+## Background
+
+When SAS authentication was first introduced, two methods were defined: emoji
+and decimal.  The intention was that emoji would be the primary method, but for
+clients who were unable to present emojis, the decimal method was added.  In
+order to ensure that any pair of clients could verify, the decimal method was
+made mandatory since it is the lowest common denominator.
+
+### Problems with emojis
+
+#### Communicating the emojis
+
+When two people verify each other in person, they can look at each others'
+devices and check that the emojis match directly.  However, if the verification
+is not done in person, the two people need to communicate the emojis with each
+other.  In most cases, the people would need to look at the emoji, and try to
+determine how to describe the emoji to the other person.
+
+To help with this, the spec defines standardised English names for the emojis,
+and encourages client authors to collaborate on
+[translations](https://github.com/matrix-org/matrix-doc/blob/master/data-definitions/).
+However, this is not always sufficient.
+
+Suppose that Alice's first language is German, and Bob's first language is
+French.  They both speak English, but neither is fluent in it.  So Alice's
+client shows Alice the German names for the emojis, and Bob's client shows Bob
+the French names for the emojis.  Thus Alice and Bob need to translate the
+German or French words into English.  While the emojis were chosen to be easily
+describable, they are not all common words that would be known by non-native
+speakers of a language, for example, 🌵 or ⌛.
+
+A client could offer to show the names in multiple languages, but that would
+complicate the user interface.
+
+In addition, we made at least one error in choosing emojis: 🔧 is "spanner" in
+en_GB, but "wrench" in en_US, so there may even be confusion between users
+speaking the same language.  The spec picks "spanner" as the name, and if a
+British and American user are verifying each other, the British user can explain
+it to the American user.  But if two American users verifying each other may
+need to figure out on their own what a "spanner" is.  (✈️ suffers a similar
+problem, but less severely, since it is likely that users can figure out that
+"aeroplane" is the same as "airplane".)
+
+#### Differences in emoji representations
+
+Clients may use different fonts for displaying emojis.  Some clients (such as
+terminal-based clients) may not have a choice in which font is used.  Thus users
+may not see the exact same images on the clients, causing them to wonder if the
+difference is significant.  The standardised names were intended to help with
+this, but in practice, it turns out that people still get confused.
+
+### Decimal numbers do not have the same issues
+
+Decimal numbers, the other method defined for SAS verification, do not have the
+issues mentioned above.  Decimal numbers are widely understood (even in
+languages that do not natively use Arabic numerals), so people will not struggle
+to describe them to each other.  People who know enough of a language to have a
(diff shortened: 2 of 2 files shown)


proposals/4405-deprecate-emoji-verification.md
turt2live, line 1: Implementation requirements:

* [ ] Some notion of approval from product-focused teams

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,132 @@
...
+
+A client could offer to show the names in multiple languages, but that would
+complicate the user interface.
+
+In addition, we made at least one error in choosing emojis: 🔧 is "spanner" in
+en_GB, but "wrench" in en_US, so there may even be confusion between users
+speaking the same language.  The spec picks "spanner" as the name, and if a
+British and American user are verifying each other, the British user can explain
+it to the American user.  But if two American users verifying each other may
+need to figure out on their own what a "spanner" is.  (✈️ suffers a similar
```
Johennes:
```suggestion
it to the American user.  But if two American users are verifying each other, they may
need to figure out on their own what a "spanner" is.  (✈️ suffers a similar
```

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,132 @@
...
+[MSC4347](https://github.com/matrix-org/matrix-spec-proposals/issues/4347)
+proposes to standardize the visual representation of the emojis.  This MSC tries
+to fix the issue of differences in emoji representation, but does not fix the
+translation issue.
+
+
+[MSC4404](https://github.com/matrix-org/matrix-spec-proposals/pull/4404)
+proposes to specify that the emoji names should be used rather than the visual
+representations.  This MSC tries to fix the both the issue of differences in
+emoji representation, and the translation issue.  As a historical note, this
```
Johennes:
```suggestion
representations.  This MSC tries to fix both the issue of differences in
emoji representation, and the translation issue.  As a historical note, this
```

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,132 @@
...
+"aeroplane" is the same as "airplane".)
+
+#### Differences in emoji representations
+
+Clients may use different fonts for displaying emojis.  Some clients (such as
+terminal-based clients) may not have a choice in which font is used.  Thus users
+may not see the exact same images on the clients, causing them to wonder if the
+difference is significant.  The standardised names were intended to help with
+this, but in practice, it turns out that people still get confused.
+
```
Johennes, line 60: One thing that might help further cement the motivation is that the complexity of user instructions differs significantly between the two verification methods. For decimal verification, something like "Compare these numbers with the other device" is sufficient and unambiguous. For emoji verification, you need to explain that the images might look differently on the other device and might also have different names.

I'm not sure if this captures it particularly well but here is a suggestion:

```suggestion

#### Complexity of user instructions

It can be difficult to formulate instructions for what to actually compare during
emoji verification for non-technical users. For instance, users have to understand
that they don't need to compare the emoji graphics pixel by pixel. They also need
to understand that they don't need to compare the emoji names letter by letter
because they might not be in the same language. This complexity undermines the
anticipated simplicity that emoji verification was meant to provide.

```

proposals/4405-deprecate-emoji-verification.md
richvdh, line 1: @uhoreg I think the main thing that will help drive this is actually trying it out: notably we can unilaterally turn off emoji verification in Element clients without the MSC landing or any sort of ecosystem approval. Could you maybe take the lead on pitching in a Crypto team meeting?

richvdh: FWIW Element-web makes decimal comparison look like this:

Image

andybalaam: This looks easy and clear to me, and arguably less intimidating than emojis.

uhoreg: As a data point, Nico says that Famedly is already only using decimal verification, due to requests from customers.

soupermkc: This seems like more stuff to check between the two devices compared to the emoji approach.
Is there a reason something like OpenMoji or the like can't be used if consistency is the issue here?
(Plus, there's more room for error with this approach.)

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,141 @@
+# MSC4405: Deprecate the emoji method for SAS verification
```
TheArcaneBrony, line 1: I struggle really badly at handling numbers (to the point where 2FA is a nightmare and I have quit doing 2FA on a separate device, or taking 5-10 minutes to type over a bank account number).

This also would be majorly problematic for anyone with dyscalculia or otherwise poor/bad vision.

I would prefer if Emoji verification remained, because it's very visual and intuitive, even if the rendering can be inconsistent between apps, devices and platforms.

proposals/4405-deprecate-emoji-verification.md
ollietheotter983: > This looks easy and clear to me, and arguably less intimidating than emojis.

it should be considered that this is less accessible to those with struggles such as ADHD, dyslexia and dyscalculia than the emojis

Johennes: > This seems like more stuff to check between the two devices compared to the emoji approach.

Personally, I don't think it is. For comparison, here is how Element Web / Desktop presents emoji verification.

Screenshot 2026-01-26 at 08 14 57

The emoji alone might feel easier to compare. However, that alone isn't enough as evidenced by attempts to fix it such as https://github.com/matrix-org/matrix-spec-proposals/pull/4404 or https://github.com/matrix-org/matrix-spec-proposals/pull/4347. When you factor in emoji images, names and translations this actually feels more mentally burdening and potentially ambiguous than comparing plain numbers.

> it should be considered that this is less accessible to those with struggles such as ADHD, dyslexia and dyscalculia than the emojis

Would those people have the same issues in emoji verification when factoring in emoji names and possibly translations?

proposals/4405-deprecate-emoji-verification.md
Johennes: > I struggle really badly at handling numbers (to the point where 2FA is a nightmare and I have quit doing 2FA on a separate device, or taking 5-10 minutes to type over a bank account number).

I'm not sure how much of a practical difference it makes but bank account numbers in Europe are almost double the length of the verification decimals (20 vs. 12 symbols) and you shouldn't have to type them during verification.

> This also would be majorly problematic for anyone with dyscalculia or otherwise poor/bad vision.

This is true. Though I suppose a dyslexic person would have similar issues when having to also read emoji names and possibly translations. I wonder if voiceover / screenreaders could assist in both of these cases?

proposals/4405-deprecate-emoji-verification.md
ara4n: i suspect concerns could be mediated somewhat here by using a dyslexia-friendly font for the numbers and improving the layout in EW (e.g. boosting their size).

Another option: given these are just ways of representing the same data, could we solve this by letting the user select different representations? e.g. default to numbers, but have a 'compare emoij instead' if they prefer (which in turn could prompt you to switch language that the emoji names are shown in)?  So the humans would then converge on the right representation to compare, according to what's best for them (numbers, emoji, words-in-english, words-in-french etc)?

ara4n: > @uhoreg I think the main thing that will help drive this is actually trying it out: notably we can unilaterally turn off emoji verification in Element clients without the MSC landing or any sort of ecosystem approval. Could you maybe take the lead on pitching in a Crypto team meeting?

i filed https://github.com/element-hq/element-meta/issues/3100 fwiw

richvdh: Resolving this thread in favour of https://github.com/element-hq/element-meta/issues/3100. Let's use the other thread to discuss accessibility concerns.

proposals/4405-deprecate-emoji-verification.md
richvdh: @ara4n suggested:
 * change the font/layout in EW to make it more dyslexia-friendly.
 * Allow users to click "compare emoji instead".

TheArcaneBrony: Providing the text descriptions of emojis may be a fair solution to avoid the confusion problem, as it can be ignored outright if the emojis match up.

As for the 20 vs. 12 symbols thing, I mentioned 2FA as another annoying case specifically because it's 6 symbols - it still takes a considerable amount of time and effort to verify, whereas with the 8 emojis you can rely on visual memory at a glance.

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,141 @@
...
+
+In addition, we made at least one error in choosing emojis: 🔧 is "spanner" in
+en_GB, but "wrench" in en_US, so there may even be confusion between users
+speaking the same language.  The spec picks "spanner" as the name, and if a
+British and American user are verifying each other, the British user can explain
+it to the American user.  But if two American users are verifying each other, they may
+need to figure out on their own what a "spanner" is.  (✈️ suffers a similar
+problem, but less severely, since it is likely that users can figure out that
+"aeroplane" is the same as "airplane".)
+
```
NiXTheDev, line 52: This extends well beyond _just_ emojis, any sort of pictogram(even universal, like recycling: ♻️) will fail in a cross-language scenario presented(Alice speaks French, Bob speaks German, Both speak English, but neither is fluent enough)

Here's an example:

- Shapes
  <details>
  <summary>Translation Examples</summary>

  - **Triangle** (English)
    - *good (Similar to English)*
      - French - Triangle
      - Spanish - Triángulo
      - Italian - Triangolo
      - Portuguese - Triângulo
      - Romanian - Triunghi
    - *bad (Different from English)*
      - German - Dreieck
      - Finnish - Kolmio
      - Hungarian - Háromszög
      - Turkish - Üçgen
      - Dutch - Driehoek

  - **Square** (English)
    - *good (matches English root logic)*
      - German - Quadrat
      - Spanish - Cuadrado
      - Italian - Quadrato
      - Portuguese - Quadrado
      - Russian - Квадрат (Kvadrat)
    - *bad (no visual match to "Square"/"Quad")*
      - French - Carré
      - Finnish - Neliö
      - Hungarian - Négyzet
      - Turkish - Kare
      - Romanian - Pătrat

  - **Pentagon**
    - *good*
      - Spanish - Pentágono
      - Italian - Pentagono
      - French - Pentagone
      - Portuguese - Pentágono
      - German - Pentagon
    - *bad*
      - Finnish - Viisikulmio (Five-corner)
      - Hungarian - Ötszög
      - Turkish - Beşgen
      - Russian - Пятиугольник (Pyatiugolnik)
      - Polish - Pięciokąt

  - **Hexagon**
    - *good*
      - Spanish - Hexágono
      - French - Hexagone
      - Portuguese - Hexágono
      - Romanian - Hexagon
      - German - Hexagon
    - *bad*
      - Italian - Esagono (Drops the 'H' sound)
      - Dutch - Zeshoek (Six-corner)
      - Finnish - Kuusikulmio
      - Turkish - Altıgen
      - Russian - Шестиугольник (Shestiugolnik)

  - **Cross**
    - *good*
      - Spanish - Cruz
      - Italian - Croce
      - Portuguese - Cruz
      - French - Croix
      - Romanian - Cruce
    - *bad*
      - German - Kreuz
      - Finnish - Risti
      - Hungarian - Kereszt
      - Turkish - Haç
      - Russian - Крест (Krest)

  - **Line**
    - *good*
      - Spanish - Línea
      - Italian - Linea
      - French - Ligne
      - Portuguese - Linha
      - Russian - Линия (Liniya)
    - *bad*
      - Finnish - Viiva
      - Hungarian - Vonal
      - Turkish - Çizgi
      - Indonesian - Garis
      - Czech - Čára
  </details>

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,141 @@
...
+to describe them to each other.  People who know enough of a language to have a
+conversation in that language are likely to know enough about numbers in that
+language be able to compare them.  And since the set of numbers is commonly
+known, people do not need to worry about what differences in rendering are
+significant.
+
+## Proposal
+
+The emoji method of SAS verification is deprecated.  Clients SHOULD only offer
+to use the decimal method.
```
turt2live, line 84: to future proof this a bit if/when we add a different mechanism:


```suggestion
The emoji method of SAS verification is deprecated.  Clients SHOULD NOT offer
to use the emoji method.
```

turt2live (approved):
Overall this LGTM. I'm not sure why this got stuck in process for the last few months, but let's fix that :)

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge

Designated reviewers are @ara4n with his Product hat on. Given this is written by a member of the crypto team, assigning crypto review feels unnecessary.

proposals/4405-deprecate-emoji-verification.md
```diff
@@ -0,0 +1,141 @@
...
+and decimal.  The intention was that emoji would be the primary method, but for
+clients who were unable to present emojis, the decimal method was added.  In
+order to ensure that any pair of clients could verify, the decimal method was
+made mandatory since it is the lowest common denominator.
+
+### Problems with emojis
+
+#### Communicating the emojis
+
+When two people verify each other in person, they can look at each others'
```
albertotirla, line 22: this is about people verifying each other in an end to end encryption context, but what about a user bootstrapping a new device by verifying with one of their existing, verified ones? do we get the decimal verification method instead of the emoji one there as well, or is this applicable only to verification between the devices of different users?

NiXTheDev: > do we get the decimal verification method instead of the emoji one there as well, or is this applicable only to verification between the devices of different users?

My understanding is that this proposal encompasses both the E2EE user and device verification

While it would make little sense to deprecate emoji method for cross-device verification(as a quick glance on both devices' screens is more than enough to check whether they're the same or not, and it is faster than comparing numbers), _not_ deprecating it just for CDV makes even less sense if it's being deprecated for cross-user verification

PR #4409, opened 2026-01-23
MSC4409: Clarify thumbnailing behavior in E2EE

stefanceriu:
This spec change proposal aims to clarify how clients should behave in regards to thumbnails in an E2EE world, or at least start a conversation around it that will eventually lead to clarifications.

This stemmed from recent internal discussions around expected behaviors for thumbnails of thumbnails and how that should be applied for local echoes.

[Rendered](https://github.com/stefanceriu/matrix-spec-proposals/blob/main/proposals/4409-carify-thumbnailing.md)

Implementation

EX iOS:
- Video thumbnail generation - https://github.com/element-hq/element-x-ios/blob/cf79cb9b013311f37911b9ed195ae75ab381f67d/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift#L398
- Image thumbnail generation - https://github.com/element-hq/element-x-ios/blob/cf79cb9b013311f37911b9ed195ae75ab381f67d/ElementX/Sources/Services/Media/MediaUploadingPreprocessor.swift#L342C18-L342C43
- Timeline view handling - https://github.com/element-hq/element-x-ios/blob/cf79cb9b013311f37911b9ed195ae75ab381f67d/ElementX/Sources/Screens/Timeline/View/TimelineItemViews/ImageRoomTimelineView.swift#L45-L65
- Actual loading - https://github.com/element-hq/element-x-ios/blob/cf79cb9b013311f37911b9ed195ae75ab381f67d/ElementX/Sources/Services/Media/Provider/MediaProvider.swift#L50-L54

RustSDK 
- Local echos - https://github.com/matrix-org/matrix-rust-sdk/blob/3be8726afb4a8cfd41899c7569f579e2509ca673/crates/matrix-sdk/src/send_queue/upload.rs#L607

---

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4409#issuecomment-4500596132)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4409#issuecomment-4500599232)

No designated reviewers assigned due to MSC size and scope.

### DIFF
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -1,0 +1,46 @@
+# MSC4409: Clarify thumbnailing behavior
+
+The purpose of this spec change is to clarify how clients should behave in regards to thumbnails in an E2EE world.
+
+Matrix clients capable of rendering media can choose to render the original source or a thumbnail representing it. For 
+example one would not want to render a full video player within a scrollable timeline view but would a thumbnail image,
+for performance reasons. The clients are still able to choose the original source if the situation warrants it e.g. animating
+GIFs.
+
+Historically homeservers would come to the clients' help by providing an [endpoint to do so](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid). 
+The [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails) section of the Spec's client behavior 
+states that homeserver provided thumbnails are available for certain media formats, in a fixed number of resolutions and
+following rules like no upscaling and returning the original if the thumbnail is smaller than the original (for supported
+formats).
+
+With the advent of E2EE aware clients server support is no longer possible. Clients are instead expected to 
+upload their own thumbnail resources and attach them to newly introduced (at the time) `thumbnail_url`/`thumbnail_info` 
+fields on various events.
+
+The spec does not however clarify how the 2 should be handled in mixed encrypted and non-encrypted environments.
+
+## Proposal
+
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails)
+section that does address this as follows:
+
+E2EE aware clients are expected to provide thumbnail resources next to the original media they're uploading as close 
+in resolution to what the maximum backend endpoints would provide (800x600 at the time of writing).
+
+For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
+800x600 for non-image media.
+
+Providing thumbnails should happen for both encrypted and non-encrypted rooms in order to provide maximum flexibility in 
+thumbnail choice, reduce backend strain and improve format compatibility.
+
+Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
+Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
+return the original resource in encrypted rooms, where backend support isn't available.
+
+Client SDKs are encouraged to apply post network request processing to original thumbnails to have them fit 
+the final client's request size if necessary but that's an optimization left to the implementer.
+
+## Security concerns
+
+Malicious clients can of course upload thumbnails that do not represent the original content but that does not fall 
+under the umbrella of this MSC.
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -34,11 +34,13 @@ The sending client should provide thumbnails for both encrypted and non-encrypte
 flexibility in thumbnail choice, reduce backend strain and improve format compatibility.
 
 Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
-Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
-return the original resource for encrypted media, where backend support isn't available.
+Calling the thumbnailing homeserver endpoint on an availlable encrypted thumbnail resource is allowed, similar to any 
+other resource, and will result in the original media, with backend support not being available to generate other 
+thumbnail variants.
 
-Client SDKs are encouraged to apply post network request processing to original thumbnails to have them fit 
-the final client's request size if necessary but that's an optimization left to the implementer.
+Client SDKs are encouraged to apply post download checks and processing to the thumbnails received from the homeserver as
+resolution limits cannot be enforced on encrypted media. The result should fit the caller requested size but that's 
+an optimization left to the implementer.
 
 ## Security concerns
 
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -7,8 +7,8 @@ example one would not want to render a full video player within a scrollable tim
 for performance reasons. The clients are still able to choose the original source if the situation warrants it e.g. animating
 GIFs.
 
-Historically homeservers would come to the clients' help by providing an [endpoint to do so](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid). 
-The [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails) section of the Spec's client behavior 
+Historically homeservers would come to the clients' help by providing an [endpoint to do so](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid). 
+The [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails) section of the Spec's client behavior 
 states that homeserver provided thumbnails are available for certain media formats, in a fixed number of resolutions and
 following rules like no upscaling and returning the original if the thumbnail is smaller than the original (for supported
 formats).
@@ -21,7 +21,7 @@ The spec does not however clarify how the 2 should be handled in mixed encrypted
 
 ## Proposal
 
-The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails)
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
 section that does address this as follows:
 
 E2EE aware clients are expected to provide thumbnail resources next to the original media they're uploading as close 
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -34,7 +34,7 @@ The sending client should provide thumbnails for both encrypted and non-encrypte
 flexibility in thumbnail choice, reduce backend strain and improve format compatibility.
 
 Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
-Calling the thumbnailing homeserver endpoint on an availlable encrypted thumbnail resource is allowed, similar to any 
+Calling the thumbnailing homeserver endpoint on an available encrypted thumbnail resource is allowed, similar to any 
 other resource, and will result in the original media, with backend support not being available to generate other 
 thumbnail variants.
 
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -30,12 +30,12 @@ in resolution to what the maximum backend endpoints would provide (800x600 at th
 For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
 800x600 for non-image media.
 
-The sending client should provide thumbnails for both encrypted and non-encrypted rooms in order to provide maximum 
+The sending client should provide thumbnails for both encrypted and non-encrypted media in order to provide maximum 
 flexibility in thumbnail choice, reduce backend strain and improve format compatibility.
 
 Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
 Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
-return the original resource in encrypted rooms, where backend support isn't available.
+return the original resource for encrypted media, where backend support isn't available.
 
 Client SDKs are encouraged to apply post network request processing to original thumbnails to have them fit 
 the final client's request size if necessary but that's an optimization left to the implementer.
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -35,12 +35,10 @@ flexibility in thumbnail choice, reduce backend strain and improve format compat
 
 Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
 Calling the thumbnailing homeserver endpoint on an available encrypted thumbnail resource is allowed, similar to any 
-other resource, and will result in the original media, with backend support not being available to generate other 
-thumbnail variants.
+other resource, but will result in bad requests (400) as encrypted files cannot be thumbnailed.
 
-Client SDKs are encouraged to apply post download checks and processing to the thumbnails received from the homeserver as
-resolution limits cannot be enforced on encrypted media. The result should fit the caller requested size but that's 
-an optimization left to the implementer.
+Clients should not assume that images from thumbnail endpoints are the exact size requested, potentially downscaling on 
+the client if appropriate.
 
 ## Security concerns
 
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -17,7 +17,7 @@ With the advent of E2EE aware clients server support is no longer possible. Clie
 upload their own thumbnail resources and attach them to newly introduced (at the time) `thumbnail_url`/`thumbnail_info` 
 fields on various events.
 
-The spec does not however clarify how the 2 should be handled in mixed encrypted and non-encrypted environments.
+The spec does not however clarify how the two should be handled in mixed encrypted and non-encrypted environments.
 
 ## Proposal
 
@@ -30,8 +30,8 @@ in resolution to what the maximum backend endpoints would provide (800x600 at th
 For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
 800x600 for non-image media.
 
-Providing thumbnails should happen for both encrypted and non-encrypted rooms in order to provide maximum flexibility in 
-thumbnail choice, reduce backend strain and improve format compatibility.
+The sending client should provide thumbnails for both encrypted and non-encrypted rooms in order to provide maximum 
+flexibility in thumbnail choice, reduce backend strain and improve format compatibility.
 
 Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
 Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
--- a/proposals/4409-carify-thumbnailing.md
+++ b/proposals/4409-carify-thumbnailing.md
@@ -24,7 +24,7 @@ The spec does not however clarify how the two should be handled in mixed encrypt
 The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
 section that does address this as follows:
 
-E2EE aware clients are expected to provide thumbnail resources next to the original media they're uploading as close 
+Clients are expected to provide thumbnail resources next to the original media they're uploading as close 
 in resolution to what the maximum backend endpoints would provide (800x600 at the time of writing).
 
 For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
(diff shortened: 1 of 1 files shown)


proposals/4409-carify-thumbnailing.md
turt2live, line 1: Implementation requirements:

* [x] Client (sending)
* [x] Client (receiving)

stefanceriu: Not entirely sure why a spec clarification needs implementations but so be it, I've updated the description.

turt2live: All MSCs require implementation to show that they're actually possible to implement, or represent desirable behaviour. Clarifications usually fall into the second part: showing that it's already present-day behaviour. 

(there's also a path where clarifications don't need MSCs at all, but this looks to be on the fence for whether it needs an MSC)

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+
+The purpose of this spec change is to clarify how clients should behave in regards to thumbnails in an E2EE world.
+
+Matrix clients capable of rendering media can choose to render the original source or a thumbnail representing it. For 
+example one would not want to render a full video player within a scrollable timeline view but would a thumbnail image,
+for performance reasons. The clients are still able to choose the original source if the situation warrants it e.g. animating
+GIFs.
+
+Historically homeservers would come to the clients' help by providing an [endpoint to do so](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid). 
+The [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails) section of the Spec's client behavior
```
turt2live: `latest` can and does change, breaking historical links. We prefer to pin to a specific spec version instead:


```suggestion
Historically homeservers would come to the clients' help by providing an [endpoint to do so](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid). 
The [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails) section of the Spec's client behavior 
```

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+
+With the advent of E2EE aware clients server support is no longer possible. Clients are instead expected to 
+upload their own thumbnail resources and attach them to newly introduced (at the time) `thumbnail_url`/`thumbnail_info` 
+fields on various events.
+
+The spec does not however clarify how the 2 should be handled in mixed encrypted and non-encrypted environments.
+
+## Proposal
+
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails)
```
turt2live:
```suggestion
The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
```

turt2live (approved):
This arguably doesn't need to be a proposal, but it's here so we should talk about it in this context.

I'm also waiving requirements for some sections to exist in the proposal because they just don't feel necessary - this MSC is describing in-practice behaviour, so the alternatives and potential issues amount to "don't do that", which would have already happened if we were unhappy with the outcome.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge

No designated reviewers assigned due to MSC size and scope.

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+section that does address this as follows:
+
+E2EE aware clients are expected to provide thumbnail resources next to the original media they're uploading as close 
+in resolution to what the maximum backend endpoints would provide (800x600 at the time of writing).
+
+For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
+800x600 for non-image media.
+
+Providing thumbnails should happen for both encrypted and non-encrypted rooms in order to provide maximum flexibility in
```
uhoreg:
```suggestion
Providing thumbnails should happen for both encrypted and non-encrypted media in order to provide maximum flexibility in 
```

This should be stated in terms of the media, rather than the room, since encrypted media can in theory be sent to an unencrypted room, and unencrypted media can be sent in an encrypted room.

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+
+For image media smaller than that no thumbnail should be provided. Uploading clients can provide thumbnails larger than 
+800x600 for non-image media.
+
+Providing thumbnails should happen for both encrypted and non-encrypted rooms in order to provide maximum flexibility in 
+thumbnail choice, reduce backend strain and improve format compatibility.
+
+Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
+Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
+return the original resource in encrypted rooms, where backend support isn't available.
```
uhoreg:
```suggestion
return the original resource for encrypted media, where backend support isn't available.
```

proposals/4409-carify-thumbnailing.md
uhoreg: If this is supposed to apply to both encrypted and unencrypted media, then why is this recommendation only applied to E2EE-aware clients?

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+The [Thumbnails](https://spec.matrix.org/latest/client-server-api/#thumbnails) section of the Spec's client behavior 
+states that homeserver provided thumbnails are available for certain media formats, in a fixed number of resolutions and
+following rules like no upscaling and returning the original if the thumbnail is smaller than the original (for supported
+formats).
+
+With the advent of E2EE aware clients server support is no longer possible. Clients are instead expected to 
+upload their own thumbnail resources and attach them to newly introduced (at the time) `thumbnail_url`/`thumbnail_info` 
+fields on various events.
+
+The spec does not however clarify how the 2 should be handled in mixed encrypted and non-encrypted environments.
```
dbkr:
```suggestion
The spec does not however clarify how the two should be handled in mixed encrypted and non-encrypted environments.
```
(prose readability nit)

proposals/4409-carify-thumbnailing.md
dbkr: I think active voice would be clearer here, ie specify what should provide thumbnails, "The sending client should provide thumbnails for both encrypted and non-encrypted rooms".

proposals/4409-carify-thumbnailing.md
dbkr: This feels weird, ie. changing it so HSs must return the original when asked to thumbnail something they don't understand.

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+
+Providing thumbnails should happen for both encrypted and non-encrypted rooms in order to provide maximum flexibility in 
+thumbnail choice, reduce backend strain and improve format compatibility.
+
+Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.
+Calling the thumbnailing homeserver endpoint on a thumbnail resource is allowed, similar to any other resource, but will 
+return the original resource in encrypted rooms, where backend support isn't available.
+
+Client SDKs are encouraged to apply post network request processing to original thumbnails to have them fit 
+the final client's request size if necessary but that's an optimization left to the implementer.
```
dbkr: 1. "original thumbnail" == "thumbnails from the HS"?

2. Also not sure of the justification for this: why go to extra effort to throw information away when the effort of transferring it over the wire has already been expended? I can see the argument that rendering a timeline full of 4k profile pictures at postage stamp size will slow things down, this does feel like the problem might be that the HS/sender should be providing thumbnail?

proposals/4409-carify-thumbnailing.md
uhoreg: Yup.  The spec currently says that the server can return a `400` HTTP code if you try to tell it to thumbnail a media type that it can't thumbnail.  (Though I'm not sure why the example gives an `M_UNKNOWN` error code for that.)

uhoreg:
@mscbot concern unresolved comments

proposals/4409-carify-thumbnailing.md
stefanceriu: Oh, I just worded that poorly. I meant that for encrypted media the homesever won't generate any of the thumbnail variants and will return the originally specified thumbnail, if any, otherwise ~404~ 400 as usual.

proposals/4409-carify-thumbnailing.md
stefanceriu: 1) Yes

2) Well we can't really enforce an encrypted media thumbnail to be a certain size so this acts more as a warning than anything else: "you might receive big images, do act accordingly". I've made some changes that hopefuly make it clearer.

proposals/4409-carify-thumbnailing.md
stefanceriu: I envisioned this going under the generic /#thumbnails section of the spec so I reckon that's generic enough. 
Perhaps we should make it the first paragraph instead?

uhoreg: Sorry, I might not have been clear enough.  I was referring to the second paragraph in the "Proposal" section:

> E2EE aware clients are expected to provide thumbnail resources next to the original media they're uploading as close 
> in resolution to what the maximum backend endpoints would provide (800x600 at the time of writing).

If we want the clients to provide thumbnails for both encrypted and unencrypted media, then it's not clear why that sentence is is speaking specifically about E2EE-aware clients.  Are only E2EE-aware clients expected to generate thumbnails for unencrypted media, and non-E2EE-aware clients expected to not generate thumbnails?

proposals/4409-carify-thumbnailing.md
dbkr: Sorry, I'm still (or possibly more) lost. What do you mean by, "the originally specified thumbnail"?

proposals/4409-carify-thumbnailing.md
dbkr: Okay, your sentence is talking about SDKs though, which I read as SDKs being encouraged to postprocess received thumbnails to squash them down further before passing to the caller if they turn up larger than the caller requested, rather than the caller just being prepared that it might get a larger image and so don't rely on it being an exact size.

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,48 @@
...
+thumbnail variants.
+
+Client SDKs are encouraged to apply post download checks and processing to the thumbnails received from the homeserver as
+resolution limits cannot be enforced on encrypted media. The result should fit the caller requested size but that's 
+an optimization left to the implementer.
+
+## Security concerns
+
+Malicious clients can of course upload thumbnails that do not represent the original content but that does not fall 
+under the umbrella of this MSC.
```
uhoreg, line 48: This seems like a T&S concern.  For E2EE media, we obviously don't have much of a choice than to trust the uploader.  But if we're extending it so that unencrypted media gets sender-defined thumbnails, then that's more of a concern. @jimmackenzie do you have any opinions on this?

tulir: Note that unencrypted media already has sender-defined thumbnails and clients generally use them for video covers at least (for images it varies more). This MSC isn't extending the spec, it's just clarifying what clients should send

proposals/4409-carify-thumbnailing.md
stefanceriu: Perhaps I'm missing something myself, I was hoping the updated wording will clear it up.
> will return the original resource in encrypted rooms, where backend support isn't available

with original resource meaning the media file as it has been first uploaded, no processing, no changes, no nothing, just give that back.

proposals/4409-carify-thumbnailing.md
stefanceriu: > SDKs being encouraged to postprocess

Yup, that's exactly what I mean. I see this this as somewhat of a footgun that final implementers shouldn't care about and the perfect thing to put within an SDK.

proposals/4409-carify-thumbnailing.md
stefanceriu: Ah, I get it now. My thinking was that clients sending non-encrypted media can rely on the backend to generate thumbnails and it's not as important as encrypted ones where there's no choice. 
Should I perhaps merge the second and forth paragraph? 🤔

proposals/4409-carify-thumbnailing.md
dbkr: Yep, that's what we're questioning here: currently the HS will fail the request with a 400 if asked to thumbnail something it can't thumbnail. This is changing it so that it will instead succeed, returning the original binary blob that was uploaded and I don't really understand the reason for this change and it seems like it could be confusing if you asked for a thumbnail and got the original media instead.

stefanceriu: Okay, _now_ I finally understand what you meant, I didn't realise that's how it actually behaves. Sorry about all the back and forth 🤦‍♂️ 

I went and tried and actual request resulting in this: `{"errcode":"M_UNKNOWN","error":"Cannot find any thumbnails for the requested media ('...'). This might mean the media is not a supported_media_format=(image/jpeg, image/jpg, image/webp, image/gif, image/png) or that thumbnailing failed for some other reason. (Dynamic thumbnails are disabled on this server.)"}`

Should I change that paragraph to something along these lines:

> Calling the thumbnailing homeserver endpoint on an available encrypted thumbnail resource is allowed, similar to any  other resource, but will result in bad requests (400) as encrypted files cannot be thumbnailed.

proposals/4409-carify-thumbnailing.md
dbkr: Okay, I think the point here is that this is adding a blanket recommendation that SDKs always process images to be the size requested, but the reality is more nuanced: if the original is a high res photo then this might be appropriate but if it's 800x600 and the app asked for 640x480 then it's probably just a waste of processing power.

proposals/4409-carify-thumbnailing.md
dbkr: No problem! Yep, that sounds fine.

proposals/4409-carify-thumbnailing.md
stefanceriu: Would you prefer I just delete the whole paragraph then instead of saying that it's an "optimization left to the implementer" i.e. do whatever you feel's right?

dbkr: Or maybe something like, "Clients should not assume that images from thumbnail endpoints are the exact size requested, potentially downscaling on the client if appropriate."?

stefanceriu: Works for me, yeah 👍

proposals/4409-carify-thumbnailing.md
uhoreg: In my opinion, it should either limit the suggestion to encrypted media, or else drop "E2EE aware" from the 2nd paragraph.

stefanceriu: Okay, I see no reason why we should explicitely limit it so let's go with dropping the E2EE aware part. Thanks 👍

proposals/4409-carify-thumbnailing.md
uhoreg: That's a fair point.  It would be nice to have a recommendation for how to handle sender-defined thumbnails, but that could be out of scope for this MSC.  On the other hand, it would be weird if we recommend that senders include thumbnails, and then we also later on recommend that clients ignore sender-defined thumbnails.  So it would still be nice to get some T&S input here, but that's not blocking.

uhoreg:
The remaining comment threads are non-blocking, so

@mscbot resolve unresolved comments

proposals/4409-carify-thumbnailing.md
jimmackenzie: I'm keen to see [MSC4278](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/media-preview-switch/proposals/4278-media-preview-controls.md) make it through FCP, as a useful mitigation for harmful thumbnails. So my current perspective is that is fine to recommend that senders include thumbnails, and users can choose whether to display them.

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,46 @@
...
+
+The spec does not however clarify how the two should be handled in mixed encrypted and non-encrypted environments.
+
+## Proposal
+
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
+section that does address this as follows:
+
+Clients are expected to provide thumbnail resources next to the original media they're uploading as close 
+in resolution to what the maximum backend endpoints would provide (800x600 at the time of writing).
```
richvdh: Took me a while to grok what a "maximum backend" was.

```suggestion
Clients are expected to provide, next to the original media they're uploading, thumbnail images
as close as possible in resolution to the maximum resolution the 
[/thumbnail](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid) endpoint 
would provide (800x600 at the time of writing).
```

richvdh: How is the client supposed to know what resolution the /thumbnail endpoint would provide, given that is configurable by the server admin?

Are we really just saying "provide an 800x600" image?

Also, what if the original image is square? Should the thumbnail be 800x800? 600x600?

stefanceriu: Way better, thanks 🙏

stefanceriu: I was thinking the original instructions are enough:

> Servers SHOULD produce thumbnails with the following dimensions and methods
> ...
> 800x600, scale
> “scale” maintains the original aspect ratio of the image
> Servers MUST NOT upscale thumbnails under any circumstance. Servers MUST NOT return a smaller thumbnail than requested, unless the original content makes that impossible.

Which I read as 800x600 being the best quality a backend can provide so it should be the quality e2ee clients should as well BUT respecting the same aspect ratio and scaling rules.

richvdh: Wow, that's unclear. What *is* a server supposed to do with a square original? Or, for that matter, a very rectangular one?

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Proposal
+
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
+section that does address this as follows:
+
+Clients are expected to provide, next to the original media they're uploading, thumbnail images
+as close as possible in resolution to the maximum resolution the 
+[/thumbnail](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediathumbnailservernamemediaid) endpoint 
+would provide (800x600 at the time of writing).
```
richvdh: Thanks for adding the link. I still don't think this is very clear though: there's no way for clients to know, in general, what the maximum resolution the `/thumbnail` endpoint would provide.

```suggestion
Clients are expected provide, next to the original media they're uploading, a thumbnail image.
Clients should use the maximum resolution among those 
[recommended for server-side thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
(i.e. 800x600 at the time of writing) as the target thumbnail resolution).
```

Possibly?

proposals/4409-carify-thumbnailing.md
stefanceriu: Yeah, that's not exactly clear but if you were to have a 1:1 (square) image and scale it down to the 600 component of 800x600 (max) resolution then you would enter the "crop" method, not the "scale" one as defined by the rules?

proposals/4409-carify-thumbnailing.md
stefanceriu: Sounds good to me, yes 👍

richvdh (approved):

proposals/4409-carify-thumbnailing.md
```diff
@@ -0,0 +1,48 @@
...
+fields on various events.
+
+The spec does not however clarify how the two should be handled in mixed encrypted and non-encrypted environments.
+
+## Proposal
+
+The proposal is to introduce new language in the [Thumbnails](https://spec.matrix.org/v1.18/client-server-api/#thumbnails)
+section that does address this as follows:
+
+Clients are expected provide, next to the original media they're uploading, a thumbnail image.
```
tulir, line 27: It's probably worth noting that receiving clients should still render media to the best of their ability even if no thumbnail is set. Wouldn't want a situation like element classic where clients completely refused to render media that didn't have a thumbnail set, which lead to other implementations copying the original url into the thumbnail field

tulir: Actually maybe this line a few paragraphs down is sufficient, though it could be more explicit too

> Rendering clients are expected to choose the thumbnail resources (if available) when rendering media previews.

PR #4411, opened 2026-01-30
MSC4411: Widget State Event

toger5:
[rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/widget-state-event/proposals/4411-widget-state-event.md)

### DIFF
--- a/proposals/4411-widget-state-event.md
+++ b/proposals/4411-widget-state-event.md
@@ -29,15 +29,26 @@ The scope is to define a widget for a single room only. We want to define a simp
   "type": "m.widget",
   "state_key": "some-uuid",
   "content": {
+    // copied widget data
     "name": "some-widget-name", // required
     "url": "https://custom.widget.app/widget", // required
     "avatar_url": "mxc://anAvatar", // optional
+    "version": 00.0.0,
+    // optional
+    "matrix_app_registry": "matrix-registry://feed"
   }
 }
 ```
+
+
+
+
  - `name`: The widget name SHOULD be treated similar to a room name. It should be meaningful to all room members. It is expected to be the same string on all clients independent of localization.
  - `url`: The actual widget url where the widget is loaded from. This also takes the role as the widget type parameter. Most per room and per user fields are available through the widget api [MSC4412](https://github.com/matrix-org/matrix-spec-proposals/pull/4412) so the urls should be very short.
  - `avatar_url` The icon used to render the widget in the client list.
+ - `version` the currently in used version
+ - `matrix_app_registry` is a endpoint in which all versions of an app are listed. Its primary purpose is to verify the correctness of the other fields. (url, version and avatar_url)
+   Matrix apps can also be added without a registry endpoint. But then it is not possible find out if two app room installations are the same or not.
 
 Url template parameters are excluded. Additional per user and per room data are exposed via widget api.
 This has an advantage that it also solves the reactive data, example: theme, language. If the theme is passed over widget api
--- a/proposals/4411-widget-state-event.md
+++ b/proposals/4411-widget-state-event.md
@@ -1,4 +1,4 @@
-# Widget State Event Proposal
+# MSC4411: Widget State Event Proposal
 This MSC splits [MSC1236](https://github.com/matrix-org/matrix-spec-proposals/issues/3803) into two.
 - One responsible for the widget state event.
 - And one for the postmessage api.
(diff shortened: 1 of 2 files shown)


proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+
+
+
+ - `name`: The widget name SHOULD be treated similar to a room name. It should be meaningful to all room members. It is expected to be the same string on all clients independent of localization.
+ - `url`: The actual widget url where the widget is loaded from. This also takes the role as the widget type parameter. Most per room and per user fields are available through the widget api [MSC4412](https://github.com/matrix-org/matrix-spec-proposals/pull/4412) so the urls should be very short.
+ - `avatar_url` The icon used to render the widget in the client list.
+ - `version` the currently in used version
+ - `matrix_app_registry` is a endpoint in which all versions of an app are listed. Its primary purpose is to verify the correctness of the other fields. (url, version and avatar_url)
```
Half-Shot, line 50: I would say this probably isn't "core" to a working widget. It sounds plausibly like something you might want as a verification extension, but in practice we have survived many long years without it, so it's probably not core.

proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+}
+```
+
+
+
+
+ - `name`: The widget name SHOULD be treated similar to a room name. It should be meaningful to all room members. It is expected to be the same string on all clients independent of localization.
+ - `url`: The actual widget url where the widget is loaded from. This also takes the role as the widget type parameter. Most per room and per user fields are available through the widget api [MSC4412](https://github.com/matrix-org/matrix-spec-proposals/pull/4412) so the urls should be very short.
+ - `avatar_url` The icon used to render the widget in the client list.
+ - `version` the currently in used version
```
Half-Shot, line 49: Version of...what?

proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+    "matrix_app_registry": "matrix-registry://feed"
+  }
+}
+```
+
+
+
+
+ - `name`: The widget name SHOULD be treated similar to a room name. It should be meaningful to all room members. It is expected to be the same string on all clients independent of localization.
+ - `url`: The actual widget url where the widget is loaded from. This also takes the role as the widget type parameter. Most per room and per user fields are available through the widget api [MSC4412](https://github.com/matrix-org/matrix-spec-proposals/pull/4412) so the urls should be very short.
```
Half-Shot, line 47: This could do with specifying what URLs are valid (e.g. do we want localhost?)? 

>Most per room and per user fields are available through the widget api [MSC4412](https://github.com/matrix-org/matrix-spec-proposals/pull/4412) so the urls should be very short.

I don't follow, why does this mean the URL needs to be short? What is short for a URL?

proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+ - `avatar_url` The icon used to render the widget in the client list.
+ - `version` the currently in used version
+ - `matrix_app_registry` is a endpoint in which all versions of an app are listed. Its primary purpose is to verify the correctness of the other fields. (url, version and avatar_url)
+   Matrix apps can also be added without a registry endpoint. But then it is not possible find out if two app room installations are the same or not.
+
+Url template parameters are excluded. Additional per user and per room data are exposed via widget api.
```
Half-Shot, line 53: I actually think we should specify at least a few of the really common ones like RoomID, it doesn't add a lot of technical burden onto the spec.

proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+ - `data`: This is used for the url template concept (which we will not continue to propose in this PR) All this can be achieved by just
+    using a modified url. e.g. https://widget.repositry.org/notepad?padName=Example
+    Additional data can be passed over the widget api. (TODO: Backwards comapt comment)
+ - `roomId` Widgets are only for the room they are added to. This data is implicitly known by the client.
+ - `type` Is only used for allowing clients to categorize widgets. A dedicated category concept is desired here. Not part of the core msc.
+
+## Closes
+https://github.com/matrix-org/matrix-spec-proposals/pull/2764
+https://github.com/matrix-org/matrix-doc/pull/2774
+https://github.com/matrix-org/matrix-spec-proposals/issues/3803
```
Half-Shot, line 88: That might be a bit ambitious. It certainly starts to close it out, but we're not really defining the API here.

Half-Shot (commented):
I'd really love for this to progress and feels like such an obviously easy thing to write up, we should just write it up.

proposals/4411-widget-state-event.md
```diff
@@ -0,0 +1,88 @@
...
+  "url": "https://widget.repositry.org/notepad?padName=$padName&userName=$matrix_user_id" // Keep
+}
+```
+ - `creatorUserId`: Can be lied about it on each update. The truth can only get aquired via back pagination. It is redundent.
+ - `data`: This is used for the url template concept (which we will not continue to propose in this PR) All this can be achieved by just
+    using a modified url. e.g. https://widget.repositry.org/notepad?padName=Example
+    Additional data can be passed over the widget api. (TODO: Backwards comapt comment)
+ - `roomId` Widgets are only for the room they are added to. This data is implicitly known by the client.
+ - `type` Is only used for allowing clients to categorize widgets. A dedicated category concept is desired here. Not part of the core msc.
+
```
Half-Shot, line 84: We might as well point to `im.vector.modular.widgets` as the unstable version of this since it has such widespread usage. It has more data fields for sure, but I think we just need to soundly explain why the removed fields are not relevant.

proposals/4411-widget-state-event.md
yncyrydybyl: What is the roomID needed for? 
The Host is calling the client in the context of a room. A way to pass initial parameters to a client can be part of another MSC.

PR #4412, opened 2026-02-02
MSC4412: Widget Base postMessage API

dbkr:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/dbkr/widget-base-postmessage-api/proposals/4412-widget-base-postmessage-api.md)

Disclosure: I'm an Element employee but this MSC is done with my Matrix Spec Core team hat on, helping @toger5 with his hackathon idea to finally get widgets specced.


proposals/4412-widget-base-postmessage-api.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Widget

(I believe these are already satisfied, but links to such examples would be good. ie: `matrix-widget-api` for both)

toger5: There are also little changes (`api` is dropped, `capabilites` gets dropped and the `request_capabilities` action is used.) which need to be added to make sure we have msc conform impls.

PR #4413, opened 2026-02-02
MSC4413: remove `private` join_rule

ara4n:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/msc4413/proposals/4413-remove-private-join-rule.md)

### DIFF
--- a/proposals/4413-remove-private-join-rule.md
+++ b/proposals/4413-remove-private-join-rule.md
@@ -30,10 +30,10 @@ This is unhelpful, because:
 
 ## Proposal
 
-We remove `private` from the spec entirely as of the next room version.
+We remove `private` join_rules from the spec entirely.
 
 This means that it will not be explicitly "reserved" per se, but given join_rules doesn't allow extensible
-identifiers, all strings are reserved there anyway.
+identifiers as values, all strings are effectively reserved there anyway.
 
 This then means that libraries like Ruma can then remove `private` from their join_rules enum and avoid future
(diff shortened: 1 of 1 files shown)


proposals/4413-remove-private-join-rule.md
```diff
@@ -0,0 +1,66 @@
...
+   bugs when configuring join_rules - see https://github.com/ruma/ruma/issues/2354, and both Element X iOS
+   and Element X Android incorrectly setting `private` join_rules at the time of writing.
+ * In practice, we're not aware of anyone ever deliberating using the `private` join_rule, either clientside
+   or serverside.
+ * Rooms already have too many ways in which they can be 'public' or 'private' around various different axes;
+   this eliminates one of the possible flavours of confusion.
+
+## Proposal
+
+We remove `private` from the spec entirely as of the next room version.
```
richvdh: I'd strongly prefer to divorce this from room versions.

The only place room versions reference join rules is in the auth rules, and we're not changing them. I don't think we need to complicate the spec with a difference between room v12 and room v13 that is actually a no-op.

I see this change as very much something that can just change in the C-S API.

proposals/4413-remove-private-join-rule.md
turt2live, line 1: Implementation requirements:

* [ ] A server removes `private` special casing and existing rooms with `join_rule: private` don't break/explode on that server.

PR #4414, opened 2026-02-03
MSC4414: Design decision - Errors

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/arch-errors/proposals/4414-design-decision-errors.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published with my Director of Standards Development role to facilitate the SCT and spec process more broadly.


proposals/4414-design-decision-errors.md
```diff
@@ -0,0 +1,113 @@
...
+understanding can and does differ from readability. "Invalid signature" is better understood by a
+server (or possibly client) developer, but is meaningless to an end user. "Failed to join room due to
+a server error" would be a better representation of the problem for an end user.
+
+Overall, this proposal intends to solve [matrix-spec#377](https://github.com/matrix-org/matrix-spec/issues/377)
+by clarifying whether clients should rely on their own generation for errors or the server's generation
+of errors. If (continuing to) rely on their own code, servers will need to supply more detail to clients
+so they can do so. If (returning to) reliance on the server's own message, servers may have to support
+translations. Servers may also be interested in being able to offer not just plaintext error messages
+to create a richer experience in the client.
```
Johennes, line 77: A client generally is in control of all aspects of its user interface, especially the copy used. This includes languages but also terminology or use of emoji. It seems unlikely to me that a server would be able to blanketly provide copy for errors that would be deemed appropriate by all possible clients. Therefore, I'd be pro keeping error visualization on the client.

Clients seeing a suboptimal error message for a limited amount of time while an update is pending doesn't seem overly critical to me. Without having any hard evidence at hand, at least it _feels_ to me as if clients normally update faster than servers.

turt2live: I also don't have update data, but generally there's a lot of immediate upgrades followed by a long tail for both. Some servers would skew the data, like matrix.org upgrading at (almost) every Synapse release. Mobile apps tend to take longer to roll out upgrades because their respective app stores somewhat arbitrarily decide when an update is available. Web/desktop clients typically upgrade right away (within ~2 days).

Servers which also offer clients directly tend to upgrade at the same time (if at all), though this isn't a guarantee.

hughns: > A client generally is in control of all aspects of its user interface, especially the copy used. This includes languages but also terminology or use of emoji. It seems unlikely to me that a server would be able to blanketly provide copy for errors that would be deemed appropriate by all possible clients. Therefore, I'd be pro keeping error visualization on the client.

A significant exception to this is with the OAuth 2.0 API where a WebView (or whatever platform equivalent) is used to show the server provided user interface for authentication, device management and account management.

> I also don't have update data, but generally there's a lot of immediate upgrades followed by a long tail for both. Some servers would skew the data, like matrix.org upgrading at (almost) every Synapse release. Mobile apps tend to take longer to roll out upgrades because their respective app stores somewhat arbitrarily decide when an update is available. Web/desktop clients typically upgrade right away (within ~2 days).

Perhaps part of the discussion here should be about whether a server operator *can* do this (whilst maintaining an open client policy) for some errors if it chooses.

turt2live: I would expect that authentication flows are out of scope because they're not "inside" the scope of the spec anymore.

proposals/4414-design-decision-errors.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+## Proposal
```
clokep, line 128: Is there pretty much 3 options:

1. Leave stuff alone.
2. Add more specific (or secondary or something) error codes.
3. Provide translated server error via the server.

I suspect just list/table of pros/cons would go a huge way to figuring out what's appropriate?

I also suspect looking at other protocols is reasonable here -- I don't think I've seen option 3 implemented in an open protocol/API.

turt2live: Those options are pretty much it, with the understanding that option 1 isn't sustainable anymore. I'd like us (the SCT) to decide on a direction to take, fueled by what product teams want out of the error system. 

I'm hesitant to add a pros/cons list at this stage while folks are still forming opinions on what pros/cons actually exist, but possibly in a later stage of the proposal it could be added.

clokep: I pretty much think 2 is the likely way forward, but that's mostly a gut feeling without data.

Half-Shot: As far as I'm aware Element's client policy is that out of caution we only accept machine error codes and translate them in the client. The english text is useful for developers, but users will only ever see what's been translated in app.

With that in mind, I'd vote 2 to have more nuanced error codes than our current set.

ASR-ASU: It is indeed useful to maintain a clear separation between technical data and its transcription for users. Furthermore, it may be necessary for certain errors to be more specific in order to provide the information required to develop more contextualized error messages, so 2.

proposals/4414-design-decision-errors.md
turt2live: This MSC has adopted a direction of client-rendered errors. I'm resolving this thread for now, pending discussion on the newly available diff.

proposals/4414-design-decision-errors.md
turt2live: This MSC has adopted a direction of client-rendered errors. I'm resolving this thread for now, pending discussion on the newly available diff.

PR #4284, opened 2025-04-07
[WIP] MSC4284: Placeholder

turt2live:
TODO: Rendered link

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published as a Trust & Safety team member allocated in full to the Foundation.

PR #4416, opened 2026-02-06
MSC4416: Optionally requiring policy server signatures in a room

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/policy-servers-room-version/proposals/4416-policy-servers-room-version.md)

----

Depends on https://github.com/matrix-org/matrix-spec-proposals/pull/4284

----

### DIFF
--- a/proposals/4416-policy-servers-room-version.md
+++ b/proposals/4416-policy-servers-room-version.md
@@ -1,4 +1,4 @@
-# MSC4415: Optionally requiring policy server signatures in a room
+# MSC4416: Optionally requiring policy server signatures in a room
 
 [MSC4284 (proposed-FCP)](https://github.com/matrix-org/matrix-spec-proposals/pull/4284) introduces a
 concept of Policy Servers to Matrix. These servers produce recommendations on events attempting to be
@@ -153,7 +153,7 @@ This proposal's functionality will exist in an unstable room version until anoth
 to a stable room version. This is normal process for the specification. See [MSC4304 (merged)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4304-room-version-12.md)
 for an example of this process.
 
-Implementations should use `org.matrix.msc4415.v1` as an unstable room version using v12 as a base
+Implementations should use `org.matrix.msc4416.v1` as an unstable room version using v12 as a base
 until this MSC can be adopted into a stable room version.
 
 


proposals/4416-policy-servers-room-version.md
turt2live, line 1: Implementation requirements:

* [ ] Unstable room version
* [ ] Complement tests

proposals/4416-policy-servers-room-version.md
```diff
@@ -0,0 +1,165 @@
...
+does *not* override the auth rules which follow the new rule - the new rule falls through to the
+remaining rules.
+
+### Changes: Auth events selection
+
+Reference: [Current auth events selection algorithm](https://spec.matrix.org/v1.17/server-server-api/#auth-events-selection)
+
+To support the auth rule changes above, the auth events selection algorithm appends the following:
+
+* Depending on the room version, the `m.room.policy` state event, if any.
```
kegsay, line 99: If MSC4242 lands before this, then this isn't necessary.

turt2live: does MSC4242 remove auth events selection?

proposals/4416-policy-servers-room-version.md
```diff
@@ -0,0 +1,165 @@
...
+
+MSC4284 says "If the [policy server's] signature is plainly missing, the homeserver SHOULD call `/sign`
+on the policy server and use that result to determine whether to pass the event through unimpeded or
+soft fail it.". This recommendation is changed to be a SHOULD NOT in this proposal's new room version.
+Removing the recommendation to acquire a signature when processing a received event which is missing
+it encourages homeservers to get that signature themselves or face rejection under the new auth rules
+above.
+
+
+## Potential issues
```
Gnuxie, line 120: @velikopter [asked](https://matrix.to/#/!nD4Jy1hp0We0VmIM9ubjqWLBX_uV8YlTBBPa3a_v2uk/%24Eo9YIpHqSKO0OT9NGNffsF_qzaUSD2DwUJor25YUX14?via=matrix.org&via=element.io&via=starstruck.systems) whether policy servers could be used to forbid leaving a room. 

Could the auth rules be modified to allow users to leave without requiring a signature from the policy server?

turt2live: Writing a comment to acknowledge the concern, though I'm not committing to a particular solution yet. I do agree it's important to maintain the ability for people to leave rooms without policy server approval. There are some spam vectors in the area (in combination with other (not yet written?) proposals) we'd like to prevent too though.

Early thoughts on this are we might not be able to add it to the auth rules, but we might be able to make it a spec compliance requirement. For example, "policy servers MUST NOT inhibit typical membership transitions". 

There's more to think about on this - that's happening in my head, so won't really appear as github comments for now (sorry).

Gnuxie: Are you saying that there is a security embargo over the details of an attack that could exploit an auth rule that allows leave membership? Or have the details just not been written down?

turt2live: There's no security embargo I'm aware of, but adding such an auth rule creates different problems for rooms.

Gnuxie: Can these problems be written down and discussed then?

turt2live: I will when I'm done thinking about them, per my first reply.

Edit: part of the reason why I'm not writing them down now is I want to get specific feedback from individuals before general feedback.

PR #4417, opened 2026-02-12
MSC4417: URL Previews via Appservices

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/as-url-previews/proposals/4417-appservice-url-previews.md)

Conflict of Interest declaration: I am employed by Element. This MSC was written as part of my work on improving integrations in Element with other apps.

### DIFF
--- a/proposals/4417-appservice-url-previews.md
+++ b/proposals/4417-appservice-url-previews.md
@@ -81,14 +81,14 @@ as no backend service other than the URL host itself will see the request, howev
 
 ### [MSC4095 Bundled URL previews](https://github.com/matrix-org/matrix-spec-proposals/pull/4095)
 
-MSC4095 suggests that receving users can instead read the previews from the Matrix event contents rather
+MSC4095 suggests that receiving users can instead read the previews from the Matrix event contents rather
 than generating their own preview, which reduces the amount of metadata that leaks. It does however not
 solve the problem of fetching data behind an authentication gate, or clients which obey CORS, or sites
 which do not provide standard OpenGraph data.
 
 This MSC is entirely compatible with that concept, as only the sender would need to request the information
 from the application service which could then be forwarded onto other users. However, users would need to
-keep in mind that it would be possible to foward sensitive data too and would need to be given the choice
+keep in mind that it would be possible to forward sensitive data too and would need to be given the choice
 about what they would like to have forwarded.
 
 


proposals/4417-appservice-url-previews.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Appservice

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,110 @@
...
+URL lookup, and return some presentation data to show a hint as to what the URL contains. This is
+performed server-side as not all clients are able to perform their own lookup (e.g. Web clients are
+usually prevented by [CORS](https://developer.mozilla.org/en-US/docs/Glossary/CORS)).
+
+In most cases, servers can only do unauthenticated requests which means that URL previews become mostly
+useless in environments where a lot of resources are blocked behind things such as SSO. For many reasons,
+it's neither plausible to augment the client to do these lookups nor modify the homeserver to support
+specific services.
+
+This proposal suggests using the a new Appservice API to offload URL lookups to services that already
```
clokep:
```suggestion
This proposal suggests using a new Appservice API to offload URL lookups to services that already
```

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,110 @@
...
+    ...
+  aliases:
+    ...
+  preview_urls:
+    - regex: "https:\/\/github.com.*"
+```
+
+When a homeserver receives a request to preview a URL via `GET /_matrix/client/v1/media/preview_url`, it MUST
+first check all appservices to see if any contain a matching entry in `namespaces.preview_urls`. For each matching
+appservice, the homeserver should make a request to `/_matrix/app/preview_url`.
```
clokep: Often a URL preview includes an image, how does that enabled in this situation?

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,110 @@
...
+The appservice should handle the request appropriately by sending a HTTP 200 response with
+the required data. If the AS response with another status code, the homeserver should try the next
+matching application service. If no service can handle the request, then the homeserver MAY perform
+a lookup itself.
+
+The homeserver MAY choose to respect any caching headers returned by the application service, but it MUST
+also ensure caches are keyed by the user who requested it. It's also acceptable for the homeserver to avoid
+caching the request in favour of the application service handling the caching.
+
+## Potential issues
```
clokep, line 60: Is there any potential leak of internal resources if someone posts something in a public room? I guess no since URL previews are displayed only for the requesting user and you call out the appservice ones shouldn't be cached across users. I think maybe that's it then.

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,110 @@
...
+first check all appservices to see if any contain a matching entry in `namespaces.preview_urls`. For each matching
+appservice, the homeserver should make a request to `/_matrix/app/preview_url`.
+
+The appservice should handle the request appropriately by sending a HTTP 200 response with
+the required data. If the AS response with another status code, the homeserver should try the next
+matching application service. If no service can handle the request, then the homeserver MAY perform
+a lookup itself.
+
+The homeserver MAY choose to respect any caching headers returned by the application service, but it MUST
+also ensure caches are keyed by the user who requested it. It's also acceptable for the homeserver to avoid
```
clokep, line 51: Do we want a way for the appservice to tell the homeserver if it is per user or not or just always assume that? Always is probably safer.

proposals/4417-appservice-url-previews.md
Half-Shot: https://github.com/element-hq/synapse/pull/19458 exists now 
https://github.com/matrix-org/matrix-hookshot/pull/1217 exists nower
https://github.com/element-hq/matrix-bot-sdk/pull/85 exists even more nower

proposals/4417-appservice-url-previews.md
Half-Shot: In those cases we'd expect the AS to process this into a MXC (probably asynchronously via the /create endpoint) and then parse that back up, which is what the homeserver does anyway. I can spell that out, but I think it's in the implementation text of the CS API anyway.

proposals/4417-appservice-url-previews.md
Half-Shot: I feel like safe defaults here are to just always assume sensitive, perhaps we could add a flag to tell the HS it's safe to cache but I feel like the gains ain't massive.

proposals/4417-appservice-url-previews.md
Half-Shot: Yep, we wouldn't want to change the semantics of how url previews work fundamentally. It's obviously the case that if someone ignores the caching advice and creates a global cache then there is a leak, but the answer is to be careful :)

proposals/4417-appservice-url-previews.md
clokep: I'm not sure that's really clear from here that it's expected for the AS to take that over though.

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,111 @@
...
+it's neither plausible to augment the client to do these lookups nor modify the homeserver to support
+specific services.
+
+This proposal suggests using a new Appservice API to offload URL lookups to services that already
+have authenticated access on behalf of a user, are homeserver agnostic by design, and can even generate
+previews without making any HTTP hits for better privacy (e.g. by using local caches).
+
+## Proposal
+
+A new endpoint is introduced on the Application Service API `GET /_matrix/app/preview_url`, which takes:
```
turt2live: This probably needs an endpoint version

proposals/4417-appservice-url-previews.md
Half-Shot: Ack, let me get my clarity quill.

proposals/4417-appservice-url-previews.md
Half-Shot: Yup. Also to be safe I think this really needed to be `/_matrix/app/v1/>media</preview_url`.

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,117 @@
...
+
+  - This is not possible for pure web clients, which is still a large number of Matrix users.
+  - Clients may not want to store the authentication details of a wide array of services.
+  - For sites that do not follow the opengraph standard, it may be hard for clients to support. Appservices can be specific and workaround this.
+
+### [MSC4095 Bundled URL previews](https://github.com/matrix-org/matrix-spec-proposals/pull/4095)
+
+Another proposal exists which suggests that Matrix events instead bundle the URL preview data, which again is useful
+from privacy perspective because it reduces the leakage of event information, but would not be suitable for authenticated
+services as it would also expose this information to everyone in the room.
```
pixlwave: If these 2 MSCs aren't compatible then I would strongly favour MSC4095 over this one:
- It provides a privacy preserving way to use URL previews in encrypted rooms which Matrix is sorely missing right now.
- It provides stable URL previews in chat history even after a page has been changed/removed.
- It improves privacy for single-user homeservers (maybe this is arguable, but if the homeserver isn't proxying previews for multiple user's, then that server's IP is fundamentally an identifier for its user).

Half-Shot: Right, I need to clarify this bit. The MSCs aren't inherently incompatible.

This MSC is focused on how we generate previews for resources which neither the homeservers can access (lack of access / non-standard / non-opengraph source url) or the client (since the client shouldn't be holding your GitHub / GitLab / JIRA token, nor can we assume it can even make web requests to these resources).

MSC4095 is focused on how the sender can send along the URL preview to the receiver so that we reduce the amount of data leakage. So in effect, you could very easily use MSC4417 alongside MSC4095. The only gotcha is I believe clients need to offer the choice as to whether to include the data on send, as you could accidentally send the wrong thing. (Myself and @anoadragon453 were chatting about how Signal actually does this).

Moreover, I think these MSCs could land in either order without significantly making things worse. If MSC4095 lands first then we can happily bundling the derived data from this MSC in that.

If this lands first, clients should continue to give users control over how much data they leak. For enterprise or trusted use cases, users are less bothered about leaking URLs within their closed ecosystem and would like richer previews over preserving privacy against their own infra.

Half-Shot: (I've updated the text to make that clear) https://github.com/matrix-org/matrix-spec-proposals/pull/4417/commits/c5ee95f67bd4fc944c383f5affbbf8c30a2b4ec5

proposals/4417-appservice-url-previews.md
Half-Shot: https://github.com/matrix-org/matrix-spec-proposals/pull/4417/commits/11adb3c376e82a45ab4649f08bf421f623a6bc6c

proposals/4417-appservice-url-previews.md
Half-Shot: I think we can close this.

proposals/4417-appservice-url-previews.md
```diff
@@ -0,0 +1,124 @@
...
+also ensure caches are keyed by the user who requested it. It's also acceptable for the homeserver to avoid
+caching the request in favour of the application service handling the caching.
+
+### Media
+
+[URL Previews require](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediapreview_url) that
+the `og:image` specified in OpenGraph data be in MXC format. The appservice MUST do this upload step and respond
+to the homeserver with a MXC URI.
+
+## Potential issues
```
clokep, line 60: A downside of this is also that it requires admins to setup a URL provider as an appservice, instead of something more self-serve / similar to Slack plugins that allow specific preview providers.

That's also true today, but maybe worth considering.

Half-Shot: Yup, it's certainly true. There has been the idea floated about that we just do these previews on the client in the future, but that limits you to native clients and it doesn't work if you have specific rules in place about who can access what.

I think the expectation would be that you'd either have an appservice hanging around (a lot of folks run a hookshot on the side these days :innocent: ) or your provider would allow you to do trivial steps to install one (ESS provides most of this today).

I think if we were a closed protocol and there was a whole app store infrastructure in front of you we wouldn't even bother with this design, but thems the limitations.

PR #4418, opened 2026-02-14
MSC4418: Make `destination` a required server authentication field

1vel:
[Rendered](https://github.com/velikopter/matrix-spec-proposals/blob/require-destination/proposals/4418-require-destination.md)

Implementations requiring value:
 - [Continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/src/commit/8d66500c991202f464785d34f3cfa31a1c831997/src/api/router/auth.rs#L327-L329)
 - [Tuwunel](https://github.com/matrix-construct/tuwunel/blob/6c91aa1ddcfc2c25202dc9a1700ee7d11305fd6a/src/api/router/auth/server.rs#L99-L101)

Implementations permitting optional value:
 - [Conduit](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/ruma_wrapper/axum.rs#L198)
 - [gomatrixserverlib](https://github.com/matrix-org/gomatrixserverlib/blob/20c9de33969e2df883eaa69c283d6a38efce9284/fclient/request.go#L309)
 - [Synapse](https://github.com/element-hq/synapse/blob/7e4588ac4f2d18bab150a2c1a123ecb22e535534/synapse/federation/transport/server/_base.py#L111-L118)
 - [Vona](https://foundry.fsky.io/vel/matrix-vona/src/commit/b413e4b05ba3260c0fc5426d3cac8a4cab1f2f5c/vona/__main__.py#L94)

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4418#issuecomment-4114144588)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4418#issuecomment-4367206614)

### DIFF
--- a/proposals/4418-require-destination.md
+++ b/proposals/4418-require-destination.md
@@ -1,0 +1,31 @@
+# MSC4418: Make `destination` a required server authentication field
+
+Currently the server to server authentication scheme allows `destination` to be optional for
+backwards-compatibility, however most homeserver implementations today no longer permit this.
+It additionally makes vHosting more difficult to implement and utilize.
+
+## Proposal
+
+Require `destination` for authenticated federation requests.
+
+## Potential issues
+
+Older homeservers may be unable to federate with servers that implement this MSC,
+however all major implementations have caught up by now, and any older Synapse
+versions that don't send the field are likely riddled with security holes.
+
+## Alternatives
+
+None?
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+None.
+
+## Dependencies
+
+None.
--- a/proposals/4418-require-destination.md
+++ b/proposals/4418-require-destination.md
@@ -6,7 +6,8 @@ It additionally makes vHosting more difficult to implement and utilize.
 
 ## Proposal
 
-Require `destination` for authenticated federation requests.
+Require `destination` for
+[authenticated federation requests](https://spec.matrix.org/v1.17/server-server-api/#request-authentication).
 
 ## Potential issues
 
--- a/proposals/4418-require-destination.md
+++ b/proposals/4418-require-destination.md
@@ -11,8 +11,9 @@ Require `destination` for
 
 ## Potential issues
 
-Older homeservers may be unable to federate with servers that implement this MSC,
-however all major implementations have caught up by now:
+Older homeservers which do not send the field may be unable to federate
+with servers that implement this MSC, however all major implementations
+have caught up by now:
 
 - Continuwuity:
   [read](https://forgejo.ellis.link/continuwuation/continuwuity/src/commit/8d66500c991202f464785d34f3cfa31a1c831997/src/api/router/auth.rs#L327-L329),
--- a/proposals/4418-require-destination.md
+++ b/proposals/4418-require-destination.md
@@ -25,7 +25,7 @@ however all major implementations have caught up by now:
   [write](https://github.com/matrix-construct/tuwunel/blob/da3e7539f4753e8c681c8f0ade447dfab201f408/src/service/federation/execute.rs#L280)
 - Conduit:
   [read](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/ruma_wrapper/axum.rs#L198),
-  [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L125)
+  [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L282)
 - gomatrixserverlib (Dendrite):
   [read](https://github.com/matrix-org/gomatrixserverlib/blob/6c4c6f7d0d301cef23e84d237b27d9e9ff4562d7/fclient/request.go#L216),
   [write](https://github.com/matrix-org/gomatrixserverlib/blob/6c4c6f7d0d301cef23e84d237b27d9e9ff4562d7/fclient/request.go#L158)
--- a/proposals/4418-require-destination.md
+++ b/proposals/4418-require-destination.md
@@ -26,9 +26,6 @@ however all major implementations have caught up by now:
 - Conduit:
   [read](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/ruma_wrapper/axum.rs#L198),
   [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L125)
-- Conduit:
-  [read](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/ruma_wrapper/axum.rs#L198),
-  [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L125)
 - gomatrixserverlib (Dendrite):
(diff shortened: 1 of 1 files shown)


proposals/4418-require-destination.md
```diff
@@ -0,0 +1,31 @@
+# MSC4418: Make `destination` a required server authentication field
+
+Currently the server to server authentication scheme allows `destination` to be optional for
+backwards-compatibility, however most homeserver implementations today no longer permit this.
+It additionally makes vHosting more difficult to implement and utilize.
+
+## Proposal
+
+Require `destination` for authenticated federation requests.
```
turt2live: Link to spec please.

proposals/4418-require-destination.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Potential issues
+
+Older homeservers may be unable to federate with servers that implement this MSC,
+however all major implementations have caught up by now, and any older Synapse
+versions that don't send the field are likely riddled with security holes.
```
turt2live: Successful MSCs here cite their sources in the proposal itself. Listing the implementations from the matrix.org website and permalinking to where they or their libraries read *and* write `destination` is helpful.

turt2live (changes requested):
This proposal seems extremely close to proposed-FCP - thanks for writing it! Once the implementation evidence is provided, I think we can push it forward in the process.

Please re-raise the MSC in the [SCT Office](https://matrix.to/#/#sct-office:matrix.org) once the implementation evidence is incorporated in the MSC, as we might miss mentions on github.

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

proposals/4418-require-destination.md
```diff
@@ -0,0 +1,47 @@
...
+  [write](https://forgejo.ellis.link/continuwuation/continuwuity/src/commit/754959e80d4865cfcfd9c0de10ab9391b11bac39/src/service/federation/execute.rs#L251)
+- Synapse:
+  [read](https://github.com/element-hq/synapse/blob/7e4588ac4f2d18bab150a2c1a123ecb22e535534/synapse/federation/transport/server/_base.py#L111-L118),
+  [write](https://github.com/element-hq/synapse/blob/df24e0f30244b1c423f4130d64c6008be341d0b7/synapse/http/matrixfederationclient.py#L956)
+- Tuwunel:
+  [read](https://github.com/matrix-construct/tuwunel/blob/da3e7539f4753e8c681c8f0ade447dfab201f408/src/api/router/auth/server.rs#L99-L101),
+  [write](https://github.com/matrix-construct/tuwunel/blob/da3e7539f4753e8c681c8f0ade447dfab201f408/src/service/federation/execute.rs#L280)
+- Conduit:
+  [read](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/ruma_wrapper/axum.rs#L198),
+  [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L125)
```
tulir: The line here seems wrong

```suggestion
  [write](https://gitlab.com/famedly/conduit/-/blob/8def22bfb8b9b23bbd47e17772f2bd80500eacf6/src/api/server_server.rs#L282)
```

tulir:
The implementations linked in the proposal doc seem to be accurate

@mscbot fcp merge

proposals/4418-require-destination.md
```diff
@@ -0,0 +1,47 @@
...
+It additionally makes vHosting more difficult to implement and utilize.
+
+## Proposal
+
+Require `destination` for
+[authenticated federation requests](https://spec.matrix.org/v1.17/server-server-api/#request-authentication).
+
+## Potential issues
+
+Older homeservers may be unable to federate with servers that implement this MSC,
```
turt2live: The MSC should spell out these effects in more detail. Is it an issue for servers which expect the value but don't receive it, servers which provide it unexpectedly, or both? Can they recover from the situation? If so, how?

turt2live: @mscbot concern Effects of requiring the flag are not detailed enough for review.

turt2live (changes requested):
I suspect this is probably fine to land, but it needs more detail added to the MSC before I'm comfortable giving it a review.

proposals/4418-require-destination.md
1vel: It would be an issue for servers that expect the value but don't receive it. [The spec](https://spec.matrix.org/v1.18/server-server-api/#request-authentication) already says:

> Unknown parameters are ignored.

so "servers which provide it unexpectedly" being an error case should be impossible.

turt2live: I'm not overly in favour of this MSC given the breaking changes, but we don't need a concern to track that.

@mscbot resolve Effects of requiring the flag are not detailed enough for review.

proposals/4418-require-destination.md
```diff
@@ -0,0 +1,48 @@
+# MSC4418: Make `destination` a required server authentication field
```
dbkr, line 1: Sorry, I don't understand exactly what the proposal is here. The spec already requires that servers must always send it, so it's already required? Are you suggesting that servers must reject requests without this header lest they be non-spec compliant? If so, I don't see how this makes vHosting more difficult (and the data in your PR description suggests the majority of homeservers allow requests without a destination which is not what your MSC says).

erikjohnston: Currently, the spec says:

> For compatibility with older servers, **recipients should accept requests without this parameter**, but MUST always send it. 

I actually thought that it was a stronger suggestion than SHOULD there. However, I think that it's still helpful to further relax that and allowing dropping compatibility with older servers.

The reason doing so is useful is that it makes it much easier for implementations to justify not supporting remote servers that don't include the `destination` field. At the moment the spec heavily suggests that servers should continue supporting such servers.

dbkr: I'd argue SHOULD means they're already allowed to drop compat, but I'm happy to change it to "MAY accept requests without" or indeed just drop the backwards compat sentence entirely (ie. not sending destination would be out-of-spec so HSes might choose to accept/reject/turn your homeserver into a goldfish/whatever) which is not massively different from now. I do think this MSC needs to be be very explicit in what change it's actually proposing though, as the language in this section is very specific and this proposal could be interpreted in all sorts of different ways.

clokep: My understanding was that it dropped any requirement for backwards compatibility. But maybe I read too much into it.

uhoreg: I agree that this MSC needs to be more explicit about what it's trying to do.  It looks like possible readings  are to:

- say that recipients MAY reject requests without the `destination` parameter (this is largely the same as what we currently have)
- say that recipients SHOULD reject requests without the `destination` parameter (or MAY accept requests without `destination`)
- say that recipients MUST reject requests without the `destination` parameter
- drop any mention of backwards compatibility, and let the recipient do whatever the feel like.

@velikopter can you clarify what you're intending?

uhoreg:
@mscbot concern unclear what exactly is being proposed

turt2live:
@mscbot fcp cancel

proposals/4418-require-destination.md
erikjohnston: Given it says "Require `destination`" and the context a) says most most homeservers will reject requests that are missing a `destination`, and b) it's required by all sending servers already, I think its pretty clear. Agree we could make it more explicit, but let's suggest the change, such as:

> Require `destination` for [authenticated federation requests](https://spec.matrix.org/v1.17/server-server-api/#request-authentication). Homeservers MUST include the `destination` field when sending requests, and MUST reject inbound requests that do not include the `destination` field.

PR #4427, opened 2026-02-25
MSC4427: Custom banners for user profiles

cyrneko:
[Rendered](https://github.com/cyrneko/matrix-spec-proposals/blob/main/proposals/4427-user-banners.md)

Signed-off-by: Alexia <cyrus42@disroot.org>

### DIFF
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -19,10 +19,10 @@ We add a new field to represent banners:
 ```json
 {
   "avatar_url": "mxc://matrix.org/SDGdghriugerRg",
-  "banner_url": "mxc://matrix.org/example123",
   "displayname": "Alice Margatroid",
   "m.example_field": "custom_value",
-  "m.tz": "Europe/London"
+  "m.tz": "Europe/London",
+  "m.banner_url": "mxc://matrix.org/example123"
 }
 ```
 
@@ -41,14 +41,9 @@ The same security precautions apply here as they do for avatars. Uploaded
 banners should always be some kind of image data (i.e image/* mime-type)
 and never other arbitrary files.
 
-## Alternatives
-
-This field could also be implemented as `m.banner_url`, as per MSC4133.
-
 ## Unstable prefix
 
-This MSC is a non-breaking change, and as such `banner_url` can be used as-is
-even before this MSC is merged.
+`eu.cyrneko.msc4427.banner_url`
 
 ## Dependencies
 
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -43,7 +43,7 @@ and never other arbitrary files.
 
 ## Unstable prefix
 
-`eu.cyrneko.msc4427.banner_url`
+`chat.commet.profile_banner`
 
 ## Dependencies
 
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -1,4 +1,4 @@
-# MSC0000: Custom banners for user profiles
+# MSC4427: Custom banners for user profiles
 
 On many platforms, users have additional customization options for 
 their profile, including but not limited to banners.
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -28,6 +28,9 @@ We add a new field to represent banners:
 
 Clients can then use this field to load user banners.
 
+It is recommended that clients stick to a horizontal aspect ratio, 
+like 3:1 or 2.8:1.
+
 ## Potential issues
 
 Some clients, like [extera](https://matrix.org/ecosystem/clients/extera-next/)
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -22,13 +22,13 @@ We add a new field to represent banners:
   "displayname": "Alice Margatroid",
   "m.example_field": "custom_value",
   "m.tz": "Europe/London",
-  "m.banner_uri": "mxc://matrix.org/example123"
+  "m.banner_url": "mxc://matrix.org/example123"
 }
 ```
 
 Clients can then use this field to load user banners.
 
-The `m.banner_uri` field MUST be an MXC URI. External URIs/URLs, i.e 
+The `m.banner_url` field MUST be an MXC URI. External URIs/URLs, i.e 
 to image hosting services, should not be used.
 
 It is recommended that clients stick to a horizontal aspect ratio, 
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -22,7 +22,7 @@ We add a new field to represent banners:
   "displayname": "Alice Margatroid",
   "m.example_field": "custom_value",
   "m.tz": "Europe/London",
-  "m.banner_url": "mxc://matrix.org/example123"
+  "m.banner_uri": "mxc://matrix.org/example123"
 }
 ```
 
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -28,6 +28,9 @@ We add a new field to represent banners:
 
 Clients can then use this field to load user banners.
 
+The `m.banner_uri` field MUST be an MXC URI. External URIs/URLs, i.e 
+to image hosting services, should not be used.
+
 It is recommended that clients stick to a horizontal aspect ratio, 
 like 3:1 or 2.8:1.
 
--- a/proposals/4427-user-banners.md
+++ b/proposals/4427-user-banners.md
@@ -28,8 +28,7 @@ We add a new field to represent banners:
 
 Clients can then use this field to load user banners.
 
-The `m.banner_url` field MUST be an MXC URI. External URIs/URLs, i.e 
-to image hosting services, should not be used.
+The `m.banner_url` field MUST be an MXC URI.
 
 It is recommended that clients stick to a horizontal aspect ratio, 
 like 3:1 or 2.8:1.
--- a/proposals/0000-user-banners.md
+++ b/proposals/0000-user-banners.md
@@ -1,0 +1,55 @@
+# MSC0000: Custom banners for user profiles
+
+On many platforms, users have additional customization options for 
+their profile, including but not limited to banners.
+
+Banners are displayed above, and often slightly behind a users' avatar,
+most commonly featuring landscape content.
+
+This MSC aims to standardize these banners across the matrix ecosystem.
+
+## Proposal
+
+Banners are something that is already widely represented on instant
+messaging platforms, and even in some UI designs in the Matrix
+ecosystem. Yet, there isn't a standardized field (outside of 
+non-specced implementations in some clients) to read banners from.
+
+We add a new field to represent banners:
+```json
+{
+  "avatar_url": "mxc://matrix.org/SDGdghriugerRg",
+  "banner_url": "mxc://matrix.org/example123",
+  "displayname": "Alice Margatroid",
+  "m.example_field": "custom_value",
+  "m.tz": "Europe/London"
+}
+```
+
+Clients can then use this field to load user banners.
+
+## Potential issues
+
+Some clients, like [extera](https://matrix.org/ecosystem/clients/extera-next/)
+already had user banners in the past. This field may not be read by
+those clients and users which wish to keep their banners once this is
+implemented will have to reupload their banner.
+
+## Security considerations
+
+The same security precautions apply here as they do for avatars. Uploaded 
+banners should always be some kind of image data (i.e image/* mime-type)
+and never other arbitrary files.
+
+## Alternatives
+
+This field could also be implemented as `m.banner_url`, as per MSC4133.
+
+## Unstable prefix
+
+This MSC is a non-breaking change, and as such `banner_url` can be used as-is
+even before this MSC is merged.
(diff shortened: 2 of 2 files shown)


proposals/4427-user-banners.md
turt2live, line 1: Implementation requirements:

* [ ] Client (sending)
* [ ] Client (rendering)

proposals/4427-user-banners.md
```diff
@@ -0,0 +1,55 @@
...
+and never other arbitrary files.
+
+## Alternatives
+
+This field could also be implemented as `m.banner_url`, as per MSC4133.
+
+## Unstable prefix
+
+This MSC is a non-breaking change, and as such `banner_url` can be used as-is
+even before this MSC is merged.
```
tulir: This is not how unstable prefixes work. Any new properties must be prefixed as explained in https://spec.matrix.org/proposals/#unstable-endpoints-features-and-vendor-prefixes

proposals/4427-user-banners.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Security considerations
+
+The same security precautions apply here as they do for avatars. Uploaded 
+banners should always be some kind of image data (i.e image/* mime-type)
+and never other arbitrary files.
+
+## Alternatives
+
+This field could also be implemented as `m.banner_url`, as per MSC4133.
```
tulir: I don't see any reason for it not to be `m.banner_url` when MSC4133 defined that as the convention and this MSC directly builds on MSC4133

cyrneko: fair enough! I've not been around Matrix for a long time so I just skimmed over whatever I might've missed in the relevant sections; I'll push something to fix this (and the other comment) in a moment

cyrneko: done :)

proposals/4427-user-banners.md
cyrneko: fixed this after a push; Sorry for the confusion

proposals/4427-user-banners.md
cyrneko: Question, for an implementation to be valid...can it be one that is compatible but uses a different field?

I've just found out that [Commet](https://commet.chat) apparently implements this, with the same semantics, however uses `chat.commet.profile_banner` instead of `eu.cyrneko.msc4427.banner` / `m.banner`

I assume not, but I thought I'd ask just in case

cyrneko: I've pushed a commit that makes the unstable prefix `chat.commet.profile_banner`

This means there is now two known implementations for this feature:
- https://github.com/commetchat/commet
- https://github.com/7w1/sable

respective commits:
- Commet: https://github.com/commetchat/commet/commit/95b69ff5d65eaf9fc87d9054547397697f5a7d61
- Sable: https://github.com/7w1/sable/commit/e4f508709149c9765155ee72a755eb0182671603

Both of these can set a banner as well as display it on profiles.

Half-Shot: > Question, for an implementation to be valid...can it be one that is compatible but uses a different field?

Sure, as long as the spec your describing in this MSC matches an implementation it's valid. The reason for this check is to ensure that the proposal *works* and the field names do not need to align for that.

And there is precedence for using a existing field name as a unstable key 👍

proposals/4427-user-banners.md
```diff
@@ -0,0 +1,50 @@
...
+  "displayname": "Alice Margatroid",
+  "m.example_field": "custom_value",
+  "m.tz": "Europe/London",
+  "m.banner_url": "mxc://matrix.org/example123"
+}
+```
+
+Clients can then use this field to load user banners.
+
+## Potential issues
```
Johennes, line 36: Would we benefit from a recommendation for the banner dimensions or is this something you'd expect to work itself out organically across clients and users?

cyrneko: I think this will probably end up being an implementation detail, though I feel guidelines around *aspect ratio* would make sense probably? Something along the lines of perhaps `1:√2`[^1] or another aspect ratio

[^1]: okay, I admit, that's me being a nerd for DIN standard paper sizes. There isn't really necessarily a reason I suggest this ratio in particular

Johennes: Maybe worth comparing with other platforms. From a quick check:

* [Mastodon](https://docs.joinmastodon.org/user/profile/#header) uses 1500x500 pixels (3:1)
* [Discord](https://support.discord.com/hc/en-us/articles/4403147417623-Custom-Profiles#h_01JPQNH7CRX2642S6Q3ES5QZK1) uses 680x240 pixels (2.8:1)
* [LinkedIn](https://www.linkedin.com/help/linkedin/answer/a563309/image-specifications-for-your-linkedin-pages-and-career-pages) uses 4200x700 pixels (6:1)

Half-Shot: I can see the point on both to be honest. Mandating an aspect ratio is almost certainly going to be impossible, as clients require their own agency about how to display the banner.

That said, perhaps some wording that banners should be landscale an approximately 3:1 (clients MAY use techniques such as letterboxing or scaling if a banner doesn't fit)

cyrneko: that sounds good! I'll add that to the doc 
###### if I don't forget at least...

cyrneko: I did, indeed, forget!

I'll set a reminder with a link to this...

1vel:
(psst, you made the rendered link point to the first revision only)

proposals/4427-user-banners.md
```diff
@@ -0,0 +1,53 @@
...
+non-specced implementations in some clients) to read banners from.
+
+We add a new field to represent banners:
+```json
+{
+  "avatar_url": "mxc://matrix.org/SDGdghriugerRg",
+  "displayname": "Alice Margatroid",
+  "m.example_field": "custom_value",
+  "m.tz": "Europe/London",
+  "m.banner_url": "mxc://matrix.org/example123"
```
1vel, line 25: Perhaps it should be `m.banner_uri` instead, given this is a new field and it doesn't accept URLs (seemingly). Maybe it should also be noted that this must be an MXC URI.

cyrneko:
> (psst, you made the rendered link point to the first revision only)

oops, totally missed that, but it's fixed now, ty!

proposals/4427-user-banners.md
cyrneko: that checks out, actually; it wasn't intended that non-mxc-URIs are to be used here and I just wrote `url` from muscle-memory, I'll fix that in a little bit

q-TheDog: "url" is currently the standard in other places we have mxc's, e.g. avatar_url

cyrneko: hm, that is true...but perhaps this should be changed for new fields...? either way I intend MXC URIs so I'll probably clarify that and push. (finally coming back to this)

catfromplan9:
@cyrneko consider reverting url -> uri, url matches similar syntax elsewhere, namely avatar_url on rooms and memberships. I think that even if uri is pedantically correct it would be better to adhere to pre-established stylistic choices that exist in the spec

proposals/4427-user-banners.md
catfromplan9: avatar_url does not support arbitrary URLs either, it mandates MXC URIs, yet it is labeled avatar_url

cyrneko:
@catfromplan9 done.

proposals/4427-user-banners.md
1vel: I believe `url` in places that only accept MXC URIs is due to the fields previously accepting URLs, however I haven't verified this. Correction appreciated.

catfromplan9: @velikopter no thats not the case

1vel: The sauce gets saucier. The spec doesn't mandate `avatar_url` being an MXC URI in all places.

For example:

1. 
   https://spec.matrix.org/v1.19/client-server-api/#mroommember
2. 
   https://spec.matrix.org/v1.19/client-server-api/#get_matrixclientv3roomsroomidjoined_members

catfromplan9: > The sauce gets saucier. The spec doesn't mandate `avatar_url` being an MXC URI in all places.
> 
> For example:
> 
>     1. 
>        https://spec.matrix.org/v1.19/client-server-api/#mroommember
> 
>     2. 
>        https://spec.matrix.org/v1.19/client-server-api/#get_matrixclientv3roomsroomidjoined_members

This is a documentation issue, for privacy reasons non-MXC URLs must **NOT** be fetched by client. You can refer to matrix client implementations for that

cyrneko: I'll leave it as URL and keep the warning for implementers to only load MXC:// URIs in

cyrneko:
smh the spellcheck still thinks `commet` is wrong

catfromplan9:
Is this also supported per-room (as in, on join events)?

cyrneko:
> Is this also supported per-room (as in, on join events)?

Not through join events, only *maybe* through things such as [MSC4144](https://github.com/matrix-org/matrix-spec-proposals/pull/4144), but that is out-of-scope for this MSC.

catfromplan9:
> > Is this also supported per-room (as in, on join events)?
> 
> Not through join events, only _maybe_ through things such as [MSC4144](https://github.com/matrix-org/matrix-spec-proposals/pull/4144), but that is out-of-scope for this MSC.

Why not `m.room.member` state events? Its very straightforward to support

catfromplan9:
I implemented the `m.room.member` state event support for banners in my client, would like if it were in this MSC

catfromplan9:
Should also be kept in mind that profile lookup over federation may be disabled. If this is the case, then the only way a banner could be visible is to support it on `m.room.member` state events. Since the membership state event is per-room, this requires this MSC to also support propagating a global banner change server-side into all rooms the user is a member of when it is changed, as is already done for display name and avatar changes.

cyrneko:
hm, I'm not sure if this MSC should be the one to introduce that, I did initially just want to standardize existing behavior based on extended profile data, `m.room.member` state events are a different mechanism altogether with different considerations[^1], and I don't know many clients that implement this behavior in `m.room.member` events as opposed to using extended profile data. I think Extera might still be doing that, but they don't follow this MSC to begin with.

I am also not aware of the specifics of that mechanism so I'd have to read the spec for that part first, but that's the lesser of my worries

[^1]: My primary concern is that they're *another* reason state-events are sent to potentially hundreds or thousands of rooms, in addition to name and avatar. I was never really a huge fan of that as it means more waiting time for changes to propagate, as opposed to the basically-instant nature of extended profile data and PMPs, which also discourages changing this data often (which is something I know some like to do).

catfromplan9:
> hm, I'm not sure if this MSC should be the one to introduce that, I did initially just want to standardize existing behavior based on extended profile data, `m.room.member` state events are a different mechanism altogether with different considerations[1](#user-content-fn-1-63d25e82ccf041064ede23fd5d833ebe), and I don't know many clients that implement this behavior in `m.room.member` events as opposed to using extended profile data. I think Extera might still be doing that, but they don't follow this MSC to begin with.
> 
> I am also not aware of the specifics of that mechanism so I'd have to read the spec for that part first, but that's the lesser of my worries
> ## Footnotes
> 
>     1. My primary concern is that they're _another_ reason state-events are sent to potentially hundreds or thousands of rooms, in addition to name and avatar. I was never really a huge fan of that as it means more waiting time for changes to propagate, as opposed to the basically-instant nature of extended profile data and PMPs, which also discourages changing this data often (which is something I know some like to do). [↩](#user-content-fnref-1-63d25e82ccf041064ede23fd5d833ebe)

Well what I mean is, if profile lookups are disabled on the homeserver then you rely on member state events. You could leave propagation of the state event up to the client (meaning no server side change is necessary) so all you'd have to do is add a section about them being defined first by per-room m.room.member state events

This lets you have per-room banners supported, and if you want then you don't need to propagate it into all the rooms when changed on your client (if you assume your userbase is on servers with profile lookup enabled)

proposals/4427-user-banners.md
q-TheDog: Client ([Haven Desktop](https://github.com/Haven-Organization/haven-desktop)): https://github.com/Haven-Organization/haven-desktop/commit/d6269d48ab118bb89efe16aea2bfa3b88c353737

proposals/4427-user-banners.md
```diff
@@ -0,0 +1,56 @@
...
+}
+```
+
+Clients can then use this field to load user banners.
+
+The `m.banner_url` field MUST be an MXC URI. External URIs/URLs, i.e 
+to image hosting services, should not be used.
```
tulir: The second sentence is redundant and is more likely to cause confusion than help (if your image hosting service provides mxc uris, using that is perfectly fine)

```suggestion
The `m.banner_url` field MUST be an MXC URI.
```

proposals/4427-user-banners.md
nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/15b47291c204c1ac21012429aa207cc0ebd60336

PR #3391, opened 2021-09-09
MSC3391: API to delete account data

ShadowJonathan:
[Rendered](https://github.com/ShadowJonathan/matrix-doc/blob/account-data-delete/proposals/3391-account-data-delete.md)

Implementations;
- Synapse
  - PR: https://github.com/matrix-org/synapse/pull/14714 (design: https://github.com/matrix-org/synapse/issues/14244) + https://github.com/matrix-org/synapse/pull/15562 (add an unstable features flag)
- Ruma
  - #TBD
- JS SDK
  - https://github.com/matrix-org/matrix-js-sdk/commit/b2a10e6
- Element iOS
  - https://github.com/vector-im/element-ios/pull/7164

Fixes https://github.com/matrix-org/matrix-spec/issues/190

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`

FCP tickyboxes for **closure:** https://github.com/matrix-org/matrix-spec-proposals/pull/3391#issuecomment-2637454152

PR #4429, opened 2026-03-05
MSC4429: Profile Updates for Legacy Sync

anoadragon453:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/anoa/legacy_sync_profile_updates/proposals/4429-legacy-sync-profile-updates.md)

Conflict of Interest declaration: I am employed by Element. This MSC was written as part of my work on the Element Backend Team to land a "user status" feature (based on top of custom profiles, see MSC4426) for a customer.

Implementations:
 * Element Web (merged)
 * Synapse (`include_profile_updates_in_sync` option)


proposals/4429-legacy-sync-profile-updates.md
turt2live, line 1: Implementation requirements:

* [ ] Client (using)
* [ ] Server (sending)

scutuatua-crypto (approved):

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,233 @@
...
+`since` query parameter in the next request. This allows the homeserver to track
+which changes the client has yet to see. This does not mean homeservers now need
+to track historical values of profiles - users will typically only ever care
+about the latest profile information for another user.
+
+Homeservers may instead assign every profile update an incrementing identifier,
+and compare that against an identifier stored within the provided `since` token
+to know whether a client has yet to receive a given profile update.
+
+### Clients
```
Half-Shot: How do clients know when to expire the stored cache of profile updates? ~~If the user deletes a profile key, do we get that down sync?~~ (Covered by `null` above) If another user leaves the scope of the authenticated user, can the homeserver notify that we are no longer interested in the keys for that user?

anoadragon453: Good point. If a user leaves all rooms that you share with them, then it would make sense for homeservers to inform clients of that.

And as you suggested in DM, it would be nice to be able to send:

```
{
  "users": {
    "@leftdude:example.org": {
      "profile_updates": null
    }
  }
}
```

down `/sync` in order to indicate that all profile fields should be forgotten by the client.

proposals/4429-legacy-sync-profile-updates.md
Half-Shot: - Server https://github.com/element-hq/synapse/pull/19556
- Client https://github.com/matrix-org/matrix-js-sdk/pull/5246

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,277 @@
...
+
+Clients are recommended to enable lazy-loading in their `/sync` requests to
+limit the profile information to only those users that are initially relevant
+(i.e. senders of events that are initially returned to the client).
+
+Client MAY clear their local cache of another user's profile upon a
+`m.room.member` event of type `leave` appearing down `/sync`. The client can
+calculate which users it no longer shares a room with and remove the local cache
+for those user(s) profiles.
```
anoadragon453: @Half-Shot (and other client developers) I'm curious for your thoughts on how sensible this is.

proposals/4429-legacy-sync-profile-updates.md
anoadragon453: Done in https://github.com/matrix-org/matrix-spec-proposals/pull/4429/commits/5ab7aede963db279370e80673f80d2bf58934c00

proposals/4429-legacy-sync-profile-updates.md
anoadragon453: This was [discussed in a private matrix room](https://matrix.to/#/!ydsiRWyToDQuTImJKD:element.io/$GIWvb6EXmqScLvMXZqQDg4-mFmxvU48CkLO98CkYKgY?via=element.io&via=matrix.org&via=half-shot.uk). The conclusion was that it was not viable due to the fact that clients would not have a full view of who's in what room due to lazy-loading.

Thus, in order for lazy-loading clients to discover whether they still shared a room with a user, they would have to query the room membership of *every room*. Which makes little sense.

Instead, clients can rely on the homeserver [sending down a `profile_updates` field of `null`](https://github.com/matrix-org/matrix-spec-proposals/pull/4429/changes/5ab7aede963db279370e80673f80d2bf58934c00) to inform them that they no longer share a room with a user.

anoadragon453: Removed the section that this thread was attached to in c8d2277c.

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,289 @@
...
+updates to only those from users you share a DM with, or a `users` field to
+filter updates by specific user IDs.
+
+### Scope of profile updates
+
+Homeservers SHOULD include updates for all users that a user shares a room with
+- however it is left up to implementations if they wish to restrict that set
+further by default. Homeservers SHOULD NOT send profile updates for users who do
+not share at least one room (lest you receive updates for the entire
+federation).
```
jaywink, line 152:
```suggestion
Homeservers SHOULD include updates for all users that a user shares a room with,
including the user themselves who made the profile update, to ensure their
potential other clients get the same update. However it is left up to implementations 
if they wish to restrict that set further by default. Homeservers SHOULD NOT send
profile updates for users who do not share at least one room (lest you receive updates
for the entire federation).
```

Maybe a clarification that the user making the update should also receive updates, otherwise their other clients will be confused?

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+The field name `users` fits in alongside `rooms`. Using a nested
+`profile_updates` dictionary allows extending the dictionary beneath each user
+ID in the future if desired.
+
+Each entry in `profile_updates` represents a profile field and any changes to
+its value. If its value is `null`, the field is treated as having been removed
+from the user's profile (rather than literally being set to `null`).
```
anoadragon453, line 73: @zecakeh [pointed out on the equivalent Sliding Sync MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/4262/changes#r3466830101) that:

> From [the definition of `PUT /profile/{userId}/{keyName}`](https://spec.matrix.org/v1.18/client-server-api/#put_matrixclientv3profileuseridkeyname):
> 
> > Servers MAY reject `null` values. Servers that accept `null` values SHOULD store them rather than treating `null` as a deletion request. Clients that want to delete a field, including its key and value, SHOULD use the `DELETE` endpoint instead.

So actually `null` *is* a valid value, and we can't use it as a signal to clients that a field has been deleted, unfortunately.

Instead, we'll probably want a `removed_profile_fields` array or similar:

```jsonc
{
  "users": {
    "@user:example.org": {
      "profile_updates": {
          "m.status": {
	        "text": "Swimming in the Great Lakes!",
	        "emoji": "🏊️"
	      },
      },
      "removed_profile_fields": ["m.other_field"]
    }
  }
}
```

proposals/4429-legacy-sync-profile-updates.md
anoadragon453: Good point - I've done this in f4a9545f.

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,292 @@
...
+
+A new top-level key is added to the response body of [`GET
+/_matrix/client/v3/sync`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv3sync)
+named `users`. It features the following implied schema:
+
+```jsonc
+{
+  "users": {
+    "@user:example.org": {
+      "profile_updates": {
```
reivilibre, line 50: It would be very beneficial if the server could flag the difference between a profile **delta** and a **complete** profile to replace whatever the client has in cache.

Without that distinction, the server is limited in how much clean-up it can do; it has to (at the very least) remember the whole set of 'removed profile fields' for all of time.

Why? Because we never know what state the client is in. If we want to clean up our deltas server-side, we will have to, at some point, tell clients 'you are too old to get just the delta, here's the whole profile instead'.
Since we can only currently speak in deltas, then sending a 'whole profile' involves tracking all field names that we *might* have sent to a client in the past.


Maybe the keys we reserve are something like:

- `profile`: (fields)
- `profile_reset`: `true` (marks this as a 'whole profile', not a delta. Should this be automatically implied for an initial sync?)
- `profile_remove`: list of removed field names

You could, I suppose, make `profile_remove: true` the 'whole profile' marker. I don't know if that kind of double-duty is a good or bad thing. `profile_reset` and `profile_remove` do in fact clash so this 'naturally' resolves that clash by making it impossible

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,292 @@
...
+    }
+  },
+  // Other `/sync` fields...
+}
+```
+
+The field name `users` fits in alongside `rooms`. Using a nested
+`profile_updates` dictionary allows extending the dictionary beneath each user
+ID in the future if desired.
+
```
reivilibre, line 70: It is what it is, but `profile_updates` feels like a chunky key when it could have been called `profile`. It will be repeated a lot of times in the response body potentially.

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,292 @@
...
+These injections of *current* user profiles allow a client to always display
+the profile information for a user, even if it hasn't changed recently.
+
+Profile *updates*/diffs should *always* be sent in an incremental sync response.
+They're naturally infrequent enough that there's little need to filter them with
+the currently defined use cases.
+
+### Clients
+
+Clients are recommended to enable lazy-loading in their `/sync` requests to
```
reivilibre, line 234: Should `include_redundant_members` semantics apply also to profiles?

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,292 @@
...
+ID in the future if desired.
+
+Each entry in `profile_updates` represents a profile field and any changes to
+its value. If its value is `null`, the field is treated as having been removed
+from the user's profile (rather than literally being set to `null`).
+
+The `profile_updates` field MAY optionally be set to `null` itself. A homeserver
+MAY do this in order to inform a client that it should stop tracking the profile
+of a given user; such as when a user no longer shares any room with another
+user. For example:
```
reivilibre, line 78: I am partly wondering if this is actually necessary; would clients know to delete their profile cache because they already saw the user leave on their own?

Half-Shot: Clients will not always know if they no longer share a room with a user, for instance if the room was lazy-loaded.

proposals/4429-legacy-sync-profile-updates.md
```diff
@@ -0,0 +1,292 @@
...
+
+Clients are recommended to enable lazy-loading in their `/sync` requests to
+limit the profile information to only those users that are initially relevant
+(i.e. senders of events that are initially returned to the client).
+
+## Potential issues
+
+None identified.
+
+## Alternatives
```
benkuly, line 242: More general: Why not put them into the member events like already done with displayname and avatar? Now we have two places to gather profile information.

proposals/4429-legacy-sync-profile-updates.md
clokep: This sounds like it is re-treading conversations in https://github.com/matrix-org/matrix-spec-proposals/pull/3391

dbkr: > This sounds like it is re-treading conversations in [#3391](https://github.com/matrix-org/matrix-spec-proposals/pull/3391)

I'm not sure, in this case there's already an established mechanism to delete a profile field (DELETE), that was about whether to add one.

What about either defining `undefined` as a deletion or else just changing it in this MSC so null is not an acceptable value? Having to add a separate field to track removed fields seems just unnecessary.

proposals/4429-legacy-sync-profile-updates.md
dbkr: Could you clarify what you mean here? Profile fields are sent when the profile updates, not with every event like member events?

proposals/4429-legacy-sync-profile-updates.md
dbkr: Profile fields are global and not based on per-room member events so I don't believe tacking them onto existing member events would make sense.

benkuly: But who decides what is global? Shouldn't we consider to remove these information from the member event to have one single source of truth?

reivilibre: Putting them in per-room member events means we have the same ambiguity problems about what is intended to be global and what is intended to be per-room (e.g. per-room nicknames), so you don't know what you can and can't put in the user directory (for example) or use as a response to `/profile`.
The current profiles system (which is global) makes a lot of sense for this reason.

At the same time, removing the displayname and avatar from member events means that if the user's homeserver goes down, a new user joining the room won't know their name (or avatar). So there is some value in having these baked into the room state, even though the state bloat is undesirable.
(Either way, they can't be removed from room state yet for backwards compatibility reasons)

proposals/4429-legacy-sync-profile-updates.md
reivilibre: As far as I understand it, `include_redundant_members` means that when using lazy-loading sync, you get sent those user's membership events every time they appear in the timeline, not just the first time they appear in the timeline.
This is useful for clients that don't maintain a persistent cache of users, so they can always display the relevant information in-context without having to make CS API requests first.

The question is whether this exact same system should apply also to the profiles sent down when lazy-loading. You could imagine the same argument applying: clients might want to show the 'away' state next to the user's name and may not maintain a cache. The client might have been 're-launched' since witnessing the first timeline event from a given user, which is how they would have received their profile fields.

PR #4431, opened 2026-03-09
MSC4431: Personalised room name overrides

Johennes:
[Rendered](https://github.com/gematik/matrix-spec-proposals/blob/johannes/personalised-room-name-overrides/proposals/4431-personal-room-name-overrides.md)

### DIFF
--- a/proposals/4431-personal-room-name-overrides.md
+++ b/proposals/4431-personal-room-name-overrides.md
@@ -1,0 +1,79 @@
+# MSC4431: Personalised room name overrides
+
+Matrix includes [recommendations] for how to compute a room's display name in clients. The logic
+starts with the [`m.room.name`] state event, falls back to the [`m.room.canonical_alias`] state
+event and finally defaults to concatenating member display names. None of these branches allows one
+party in the room to *set* a room name that only they can see. This would, however, be desirable in
+certain cases:
+
+- You may want to file a DM under something else than the other user's display name because their
+  display name is ambiguous (you may know more than one person called "Charlie") or otherwise
+  confusing (you don't interact with Charlie often enough to remember "SpursLover86").
+- The logic of concatenating member names doesn't work well in DMs where one party is joined with
+  multiple accounts (e.g. a work and a personal account or a regular and a backup account).
+- You want to use a personalised naming scheme such as "Lastname, Firstname" for DMs or "External:
+  ..." for chats outside your organisation.
+
+The present proposal addresses these use cases by introducing a way to create personalised overrides
+for room names via account data.
+
+## Proposal
+
+A new room account data type `m.room.name.override` is introduced. The schema is identical to the
+schema of `content` in [`m.room.name`].
+
+    GET /_matrix/client/v3/user/@john:example.com/account_data/m.room.name.override
+    {
+      "name": "Charlie Meyers (the Hotspurs guy)"
+    }
+
+A new first step is inserted at the very beginning of the [room name computation
+RECOMMENDATIONs][recommendations]: If `m.room.name.override` exists in the room's account data and
+is not the empty object (`{}`), use the name given by that item.
+
+Clients MAY additionally display the room name implied by [`m.room.name`] and
+[`m.room.canonical_alias`] as an indicator for how the room may be displayed for other users.
+
+Since account data cannot be deleted, clients MAY set the value to the empty object (`{}`) to remove
+the override. Contrarily, to force an empty room name, clients MAY use a value of `""` for `name` in
+the override.
+
+Finally, clients MAY encrypt `m.room.name.override` using the [Secrets module].
+
+## Potential issues
+
+Confusion may arise when people refer to the same room under different names due to their personal
+overrides. Clients can display both the personal and the official name to mitigate this problem.
+
+Users may miss changes of a room's purpose due to their personal name override. Again, clients can
+display both names to help alleviate this issue.
+
+## Alternatives
+
+This proposal largely follows [MSC3015] which appears to have been abandoned.
+
+An actual alternative would be to store personal overrides in `m.room.name` state events by using
+the MXID as `state_key`. This doesn't provide confidentiality for overrides and may pollute the room
+state, slowing down state resolution, though. Additionally users may not have sufficient power level
+to send state events in all rooms.
+
+## Security considerations
+
+Personalised room name overrides may leak metadata to the home server. Clients MAY use the [Secrets
+module] to prevent this.
+
+## Unstable prefix
+
+| Stable identifier      | Purpose           | Unstable identifier                     |
+|------------------------|-------------------|-----------------------------------------|
+| `m.room.name.override` | Account data type | `de.gematik.msc4431.room.name.override` |
+
+## Dependencies
+
+None.
+
+  [recommendations]: https://spec.matrix.org/v1.17/client-server-api/#calculating-the-display-name-for-a-room
+  [`m.room.name`]: https://spec.matrix.org/v1.17/client-server-api/#mroomname
+  [`m.room.canonical_alias`]: https://spec.matrix.org/v1.17/client-server-api/#mroomcanonical_alias
+  [Secrets module]: https://spec.matrix.org/v1.17/client-server-api/#secrets
+  [MSC3015]: https://github.com/matrix-org/matrix-spec-proposals/pull/3015
--- a/proposals/4431-personal-room-name-overrides.md
+++ b/proposals/4431-personal-room-name-overrides.md
@@ -57,7 +57,7 @@ This proposal largely follows [MSC3015] which appears to have been abandoned. On
 the usage of `private` rather than `override` in the account data type. The fact that
 `m.room.name.private` in account data overrides `m.room.name` in the room state seems sufficiently
 clear even without an explicit `override` suffix. This is because the room state is shared by all
-room members whereas room account data is storage specifially meant for a single user. In contrast,
+room members whereas room account data is storage specifically meant for a single user. In contrast,
 the semantics of an `m.room.name` account data type would not be clear if another kind of override
 is introduced in an `m.room.name.whatever` account data type in future.
 
--- a/proposals/4431-personal-room-name-overrides.md
+++ b/proposals/4431-personal-room-name-overrides.md
@@ -38,8 +38,6 @@ Since account data cannot be deleted, clients MAY set the value to the empty obj
 the override. Contrarily, to force an empty room name, clients MAY use a value of `""` for `name` in
 the override.
 
-Finally, clients MAY encrypt `m.room.name.override` using the [Secrets module].
-
 ## Potential issues
 
 Confusion may arise when people refer to the same room under different names due to their personal
@@ -59,8 +57,9 @@ to send state events in all rooms.
 
 ## Security considerations
 
-Personalised room name overrides may leak metadata to the home server. Clients MAY use the [Secrets
-module] to prevent this.
+Just like the `m.room.name` state event, personalised room name overrides may leak metadata to the
+home server. A future MSC may introduce a practical encryption mechanism for account data events,
+for instance, by building upon the [Secrets module] to prevent this.
 
 ## Unstable prefix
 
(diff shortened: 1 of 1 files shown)


proposals/4431-personal-room-name-overrides.md
```diff
@@ -0,0 +1,79 @@
...
+is not the empty object (`{}`), use the name given by that item.
+
+Clients MAY additionally display the room name implied by [`m.room.name`] and
+[`m.room.canonical_alias`] as an indicator for how the room may be displayed for other users.
+
+Since account data cannot be deleted, clients MAY set the value to the empty object (`{}`) to remove
+the override. Contrarily, to force an empty room name, clients MAY use a value of `""` for `name` in
+the override.
+
+Finally, clients MAY encrypt `m.room.name.override` using the [Secrets module].
```
benkuly: A bit more context is missing. Currently, the recovery key as default key is used to encrypt and decrypt secrets. But the recovery key is only saved in memory for a very short time and forgotten after used. I think encrypting should be out of scope of this MSC.

Johennes: Hm, good point.

Well encryption is out of scope one way or another because the "MAY" applies even if it's not spelled out. The only reason I mentioned it is to safe guard against potential metadata leak criticism.

MurzNN:
I am the author of the [MSC3015: Room state personal overrides](https://github.com/matrix-org/matrix-spec-proposals/pull/3015) - it looks abandoned because I just described the pain points and the idea, and found no sponsors or other involved people to move on with this idea.

That's great that we have progress in this topic in a new MSC4431! But it looks like it covers only the room name, but MSC3015 also covers the room avatar and topic. So, maybe include in this MSC overriding the avatar too, at least?  

Let's look at the classic "Contact book" feature in smartphones: we can set not only the custom name, but also attach a personal photo, and other notes to the contact.  So, I'd prefer to see my Matrix contacts with a real photo of the person (if I have it) instead of the avatar that they can change at any time to a new ugly image unrelated to the contact, and I will forget who it is.  The same is with group chats, that like to change avatars pretty often.

What do you think about adding overriding avatar to this MSC?

Johennes:
Apologies, I wasn't aware that you're still around. The (unaddressed) comments on https://github.com/matrix-org/matrix-spec-proposals/pull/3015 led me to believe that it was abandoned. I didn't mean to duplicate your proposal needlessly.

For our use case at gematik, we really just need the name at the moment (same in https://github.com/matrix-org/matrix-spec-proposals/pull/4432). Being able to override avatars sounds like a helpful feature. I'm slightly wary of broadening the scope of the proposal though because that'll possibly complicate review / acceptance. Maybe avatars could be covered in your proposal instead?

MurzNN:
> Apologies, I wasn't aware that you're still around. The (unaddressed) comments on #3015 led me to believe that it was abandoned. I didn't mean to duplicate your proposal needlessly.

Yes, it's kind of true, in recent years I had to switch from Matrix-based development and services to another project not related to Matrix at all, so now I became just a regular annoying user of the Matrix clients ;) 

So, actually, I have no time to move the #3015 forward with detailed discussion and implementation, but I'd be happy to transfer it to anyone else to finalize and implement, for example, to you, @Johennes :wink: 

So, if those two MSCs are pretty similar, maybe just merge them into one?

proposals/4431-personal-room-name-overrides.md
turt2live, line 1: Implementation requirements:

* [ ] Client (setting)
* [ ] Client (showing)
* [ ] Server (showing)

proposals/4431-personal-room-name-overrides.md
```diff
@@ -0,0 +1,79 @@
...
+display both names to help alleviate this issue.
+
+## Alternatives
+
+This proposal largely follows [MSC3015] which appears to have been abandoned.
+
+An actual alternative would be to store personal overrides in `m.room.name` state events by using
+the MXID as `state_key`. This doesn't provide confidentiality for overrides and may pollute the room
+state, slowing down state resolution, though. Additionally users may not have sufficient power level
+to send state events in all rooms.
```
tulir, line 67: The account data event type could just be `m.room.name`, the override part is implied by it being in room account data

proposals/4431-personal-room-name-overrides.md
uhoreg: Yeah, I don't think that the Secrets module can be used for this yet.  IMHO, it would be best to replace this sentence with something saying that encryption may be defined in a separate MSC (and remove the sentence in the Security Considerations section about using using the Secrets Module).

And then at some point, someone can write an MSC that describes how to to the encryption, possibly as a generalized account-data encryption mechanism.

proposals/4431-personal-room-name-overrides.md
Johennes: There was a notion on https://github.com/matrix-org/matrix-spec-proposals/pull/3015#discussion_r664290591 that a prefix / suffix would help differentiate user-managed from server-managed data. I don't feel strongly about this personally though.

proposals/4431-personal-room-name-overrides.md
Johennes: Makes sense. I have descoped encryption to a future proposal with 99fabbfdb4e485cd526e523f4260afd6f73d7a42. Thanks for the input both.

proposals/4431-personal-room-name-overrides.md
Johennes: There is also an open question on https://github.com/matrix-org/matrix-spec-proposals/pull/4432#discussion_r2925881711 whether to convert this into a server-managed account data event. Maybe a suffix expressing the managed/unmanaged nature could be more worthwhile if that happens?

Johennes: Actually, I guess capabilities would be better to express whether the data is managed or unmanaged. That would also allow admins to configure whether something is managed or not based on their needs.

If both the local name and the server-wide name from https://github.com/matrix-org/matrix-spec-proposals/pull/4432 enter account data, we could use a `.local` and `.server` suffix on the account data type to differentiate them though.

Johennes: I have changed the suffix to `.private` in 6a9d179e73f2f77afa74179d509517af23ed776d to enable future extensibility through different types of overrides such as the ones from https://github.com/matrix-org/matrix-spec-proposals/pull/4432.

scutuatua-crypto (approved):

elbenditu:
Hello I have been _tracking_ this issue, tired of pretty much every AI hallucinating that this feature is already implemented when it is clearly not. Is there any chance that this will get through? Many thanks!

SpiritCroc:
Implemented client setting + showing in SchildiChat Revenge https://github.com/SchildiChat/schildi-revenge/commit/2dde26ce6245d2a0506699a8c6fa5839ebfcc16a

catfromplan9:
does this support avatar overrides? would really appreciate if it did. if not, is it too late for that to be added in?

catfromplan9:
Seems not. Bummer. An idea that I had is that if we wanted to, we could move `name` to `displayname` and add an `avatar_url` parameter, allowing for an override of either component.

SpiritCroc:
I'm implementing this feature in Beeper Android right now, as we're planning to adopt personal room overrides into our clients as well.

We'd also love to see some avatar coverage (I suppose `de.gematik.msc4431.room.avatar.private`) in this proposal. Since avatars are closely related to room names, and straightforward enough to handle in the same way as the room name overrides on the data layer, I think it would make sense to handle in the same proposal. I wouldn't include the room topic, which was suggested in MSC3015, as I think that may be more open to debate if it's a good idea being able to overwrite conceptually.

I wonder if it would make sense to also explicitly state that [MSC4483](https://github.com/matrix-org/matrix-spec-proposals/pull/4483) (or any other future MSC with the same goal) should be compatible with this approach, to future-proof this proposal for account data encryption. Since MSC4483 requires to only encrypt fields that are explicitly allowed for encryption, and there's no technical requirement to not have personal overrides encrypted, these two MSCs feel like a natural fit. (Of course neither MSC should depend on the other, as each also works on their own.)

So, in case this proposal will allow avatars too, it would make sense to also allow encrypted attachments as avatars (otherwise encrypted account data alone will not help much), via some content like:

```json
"file": {
  "hashes": {
    "sha256": "..."
  },
  "iv": "...",
  "key": {
    "alg": "A256CTR",
    "ext": true,
    "k": "...",
    "key_ops": [
      "encrypt",
      "decrypt"
    ],
    "kty": "oct"
  },
  "url": "mxc://...",
  "v": "v2"
}
```

Where plaintext avatar overrides would look as expected from normal avatars, like

```json
{
  "url": "mxc://..."
}
```

catfromplan9:
Looking for a similar proposal that allows personalized user profile overrides (customizable display name and avatar for others), does such a proposal exist? If not, may you create one @Johennes? If I am to implement this in my client, I would like a per-user profile overrides proposal to be made and @SpiritCroc's avatar coverage being scoped into this proposal before I do so.

KitsuneRal:
I'd recommend this MSC go with the scope as it is, making sure the mechanism is extensible to other room attributes; and put avatar/topic overrides in another MSC that would build on both this one and MSC3015.
(Please use in-text comments if you want SCT not to miss them while reviewing.)

proposals/4431-personal-room-name-overrides.md
Johennes: Client (setting & showing): https://github.com/SchildiChat/schildi-revenge/commit/2dde26ce6245d2a0506699a8c6fa5839ebfcc16a (thanks @SpiritCroc)

I don't think this proposal contains any server implementation part. So this should be ready to go ahead but given the ongoing Matrix 2.0 endeavor, I don't think it's sensible to raise it with the SCT until after the Matrix conference.

catfromplan9:
> I'd recommend this MSC go with the scope as it is, making sure the mechanism is extensible to other room attributes; and put avatar/topic overrides in another MSC that would build on both this one and [MSC3015](https://github.com/matrix-org/matrix-doc/pull/3015). (Please use in-text comments if you want SCT not to miss them while reviewing.)

Why do we need @Johennes proposal if MSC3015 can do the same thing but is more powerful?

proposals/4431-personal-room-name-overrides.md
Johennes, line 1: @catfromplan9 [said](https://github.com/matrix-org/matrix-spec-proposals/pull/4431#issuecomment-5058704256):

> Why do we need @Johennes proposal if MSC3015 can do the same thing but is more powerful?

Johennes: As mentioned in https://github.com/matrix-org/matrix-spec-proposals/pull/4431#issuecomment-4045610256, I opened this PR because MSC3015 appeared abandoned to me.

catfromplan9: > As mentioned in [#4431 (comment)](https://github.com/matrix-org/matrix-spec-proposals/pull/4431#issuecomment-4045610256), I opened this PR because [MSC3015](https://github.com/matrix-org/matrix-doc/pull/3015) appeared abandoned to me.

I see, why not continue with MSC3015's proposal then? It is more powerful than your proposal

catfromplan9: Since this MSC only allows for display name override and 3015 seems to be a more versatile proposal, I think it would be counterproductive to go forward with this proposal. If customizable room display names are proposed, then I believe avatars should be customizable too, and with arbitrary state event overrides you get that plus the ability to override more things (room banners, for example)

Johennes: Well this proposal here was sponsored by the company I was working for at the time. It's part of their fork of the Matrix spec and has implementations. So I think there's some momentum here. Ultimately a smaller scope also increases the chances of it being accepted because there's just less things to discuss. This doesn't preclude another small proposal adding personalised room avatar overrides later either. In any case, I don't get to decide which proposal the SCT moves forward with though.

catfromplan9: > Well this proposal here was sponsored by the company I was working for at the time. It's part of their fork of the Matrix spec and has implementations. So I think there's some momentum here. Ultimately a smaller scope also increases the chances of it being accepted because there's just less things to discuss. This doesn't preclude another small proposal adding personalised room avatar overrides later either. In any case, I don't get to decide which proposal the SCT moves forward with though.

Ah alright. Well that's a bummer. I think a good way to handle the scope would be to leave support up to clients. For example, clients may choose to only support room `m.room.name` overrides, or `m.room.avatar` overrides too. Others may support overriding everything. Anyway, I think for my client I will go ahead with MSC3015 but limited to avatar, display name, and banner

PR #4435, opened 2026-03-17
MSC4435: Room slowmode

1vel:
[Rendered](https://github.com/velikopter/matrix-spec-proposals/blob/slowmode/proposals/4435-room-slowmode.md)

### DIFF
--- a/proposals/4435-room-slowmode.md
+++ b/proposals/4435-room-slowmode.md
@@ -38,7 +38,7 @@ disabling the message input bar, or disabling certain buttons.
 If this state PDU is not present in the room, slowmode rate limits
 do not apply.
 
-Servers MUST reject PDUs sent over C2S with `M_LIMIT_EXCEEDED` if a user
+Servers MUST reject PDUs sent over client APIs with `M_LIMIT_EXCEEDED` if a user
 attempts to send a PDU of a given type before `rate_limit` milliseconds
 have elapsed since their last PDU of that type, and the user's power level
 is not greater than or equal to the `exempt_power_level` key (if present)
@@ -58,6 +58,7 @@ Some servers in a room may not enforce these limits. This was deemed
 acceptable as several other things in Matrix are also like this,
 for example
 [history visibility restrictions](https://spec.matrix.org/v1.17/client-server-api/#mroomhistory_visibility).
+A policy server could also be used to enforce slowmode rate limits.
 
 ## Alternatives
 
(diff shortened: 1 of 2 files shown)


proposals/4435-room-slowmode.md
turt2live, line 1: Implementation requirements:

* [ ] Client (setting)
* [ ] Client (rendering settings)
* [ ] Server (enforcing)

proposals/4435-room-slowmode.md
```diff
@@ -0,0 +1,78 @@
+# MSC4435: Room slowmode
+
+Some room administrators may want to limit the rate of
+messages in a room.
+
+## Proposal
+
+### `m.room.slowmode`
```
Johennes, line 8: Probably a matter of taste but maybe this could be `m.room.rate_limits` instead? "slowmode" feels slightly odd to me because in general we're trying to make Matrix faster rather than slower. 😅

clokep: Yeah this is definitely describing rate limits. I had the same thought.

proposals/4435-room-slowmode.md
```diff
@@ -0,0 +1,78 @@
...
+
+Some servers in a room may not enforce these limits. This was deemed
+acceptable as several other things in Matrix are also like this,
+for example
+[history visibility restrictions](https://spec.matrix.org/v1.17/client-server-api/#mroomhistory_visibility).
+
+## Alternatives
+
+None.
+
```
Gnuxie, line 66: Policy servers should be able to provide completely arbitrary and granular rate limits. Particularly in ways that are difficult to express in the power levels event/users format. Which also cannot consider event mixins from extensible events.

1vel: A policy server to strictly enforce the slowmode configuration in the room in the federation aspect may be a nice idea, but a policy server on its own would not be great UX, as it would just show to the user as being flagged as spam.

Gnuxie: Yeah and this UX problem is relevant to policy servers in all use cases, and needs additional work to allow arbitrary feedback to the user, such as https://github.com/matrix-org/matrix-spec-proposals/pull/4387.

Gnuxie: And MSC4435 shares the same concern, as it would bring dedicated error messaging in addition to the rate limit event and semantics. Instead of just dedicated error messaging for policy servers that implement rate limiting, or more to the point just general error messaging for policy servers.

1vel: Actually, it looks like the stable policy server sign API supports returning actual errors.. my ideas are being replaced by policy servers :(

Now the only real need for this would be people that can't host their own PS (and don't want to use someone elses)

scutuatua-crypto (approved):

PR #4436, opened 2026-03-17
MSC4436: Make server ACLs case insensitive

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/acl-case-insensitive/proposals/4436-case-insensitive-acls.md)

Disclosure: I am Director of Standards Development at The Matrix.org Foundation C.I.C., Matrix Spec Core Team (SCT) member, employed by Element, and operate the t2bot.io service. This proposal is written and published with my Director of Standards Development role to facilitate the SCT and spec process more broadly.

----

SCT Stuff:

No MSC checklist

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4436#issuecomment-4076624383)

### DIFF
--- a/proposals/4436-case-insensitive-acls.md
+++ b/proposals/4436-case-insensitive-acls.md
@@ -2,7 +2,7 @@
 
 [MSC1383](https://github.com/matrix-org/matrix-spec-proposals/issues/1383) originally introduced
 [server ACLs](https://spec.matrix.org/v1.17/server-server-api/#server-access-control-lists-acls). At
-the time, the MSC failed to clarify whether ACLs are case sensitive or insenstive. The
+the time, the MSC failed to clarify whether ACLs are case sensitive or insensitive. The
 [Synapse implementation](https://github.com/matrix-org/synapse/commit/3cf3e08a97f4617763ce10da4f127c0e21d7ff1d#diff-996dafa2cabdf043e51a444e5d3c5ba8389d48e6cdab0d469677974bd470d58fR726)
 used `re.IGNORECASE` to match entries, indicating preference that the ACLs should be case insensitive.
 Implementations based on Ruma will also [check case insensitively](https://github.com/ruma/ruma/blob/c242467a8ac13e691af10d69f76ca8c293e165db/crates/ruma-events/src/room/server_acl.rs#L64-L65).
--- a/proposals/4436-case-insensitive-acls.md
+++ b/proposals/4436-case-insensitive-acls.md
@@ -1,0 +1,61 @@
+# MSC4436: Make server ACLs case insensitive
+
+[MSC1383](https://github.com/matrix-org/matrix-spec-proposals/issues/1383) originally introduced
+[server ACLs](https://spec.matrix.org/v1.17/server-server-api/#server-access-control-lists-acls). At
+the time, the MSC failed to clarify whether ACLs are case sensitive or insenstive. The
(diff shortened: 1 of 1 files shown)


proposals/4436-case-insensitive-acls.md
turt2live, line 1: Implementation requirements:

* Already satisfied, per intro.

turt2live:
Inserting to relatively low priority in the stack:

@mscbot fcp merge

proposals/4436-case-insensitive-acls.md
```diff
@@ -0,0 +1,61 @@
...
+> issues with enforcing this. A similar process was used to enforce ACLs upon EDUs in
+> [MSC4163](https://github.com/matrix-org/matrix-spec-proposals/pull/4163).
+
+
+## Proposal
+
+Server ACLs become case insensitive matches. `EXAMPLE.org` and `example.ORG` are the same.
+
+
+## Potential issues
```
tulir, line 25: I wonder if [m.policy.rule.server](https://spec.matrix.org/v1.17/client-server-api/#mpolicyruleserver) needs touching 🤔

turt2live: If it does, that'd definitely be a different MSC. For now, considering the sole implementation of that event is moderation bots which proxy the event to ACLs, it's probably fine. There's no requirement that they get translated to an ACL though, which is why an MSC might be needed to expand the use cases of the event.

richvdh (approved):
anything else would be insane

proposals/4436-case-insensitive-acls.md
```diff
@@ -0,0 +1,61 @@
...
+
+* [Server names](https://spec.matrix.org/v1.17/appendices/#server-name) are case sensitive in Matrix.
```
Johennes, line 27: Is it not a contradiction that server names respect casing but ACLs which are evaluated against server names do not? 🤔

turt2live: that's why it's a potential issue :) 

I feel it's acceptable for tools built on top of the core protocol to overload behaviour like this. ACLs are one of those tools.

Johennes: Ok, fair enough as long as it's explicit, I suppose. 👍

scutuatua-crypto (approved):

velikopter (approved):

proposals/4436-case-insensitive-acls.md
TheArcaneBrony: https://cgit.rory.gay/matrix/tools/MatrixAntiDmSpam.git/tree/MatrixAntiDmSpam.Core/InviteManager.cs#n58

Example implementation consuming server policies directly, though I see no reason not to make entity matching case insensitive for `m.policy.rule.server` entities, at the SDK layer, am I missing something here?

turt2live: The entire implementation for policy events is left as an exercise for the developer. If they want to treat it case sensitively, insensitively, always upper case, or whatever, they can. 

A future MSC may (or may not) clarify that further.

turt2live:
Spec PR: https://github.com/matrix-org/matrix-spec/pull/2334

turt2live:
Looks like the bot got stuck finishing FCP here. It should have finished FCP yesterday because there are no outstanding concerns. 

I'm manually pushing it forward.

turt2live:
Merged 🎉

PR #4439, opened 2026-03-26
MSC4439: Encryption key URIs in `/.well-known/matrix/support`

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/msc4439/proposals/4439-support-contact-encryption.md)

Implementations:
* Server (publishing):
  * MDAD: https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/5090
  * [Continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1609)
  * [Tuwunel](https://github.com/matrix-construct/tuwunel/commit/eaf1b93201f3f97513d16815f54043d93e021e38)
  * Example servers using this MSC:
    * `openpgp4fpr:` URI: https://electrologic.org/.well-known/matrix/support
    * `https` URI: https://zirco.dev/.well-known/matrix/support
* Client (using):
  * Ruma: https://github.com/ruma/ruma/pull/2446
  * `go-msc1929`: https://github.com/etkecc/go-msc1929/pull/1
  * [Connectivity Tester](https://github.com/MTRNord/matrix-connection-tester-ui/releases/tag/v0.5.1)
  * Any other tool which displays `.well-known/matrix/support` fields raw

Signed-Off-By: Logan Devine <logan@zirco.dev>

### DIFF
--- a/proposals/4439-support-contact-encryption.md
+++ b/proposals/4439-support-contact-encryption.md
@@ -48,7 +48,11 @@ or mandate out-of-band channels, which this MSC accommodates.
 
 Sensitive information may simply be encrypted to the key already listed in a site's security.txt file, however, at
 larger organizations this does not provide the same per-contact granularity (and may not even reference someone
-responsible for hosting the Matrix homeserver at all).
+responsible for hosting the Matrix homeserver at all). A homeserver is unlikely to be the only service running on a
+domain. Especially for larger organizations, where the single key that security.txt provides may not actually represent
+anyone responsible for the Matrix homeserver itself. This is a tool to aid in discovery alongside WKD or key servers.
+
+Additionally, this helps obtain per-contact granularity in contrast to just one "this is our security team's key."
 
 ## Unstable prefix
 
--- a/proposals/4439-support-contact-encryption.md
+++ b/proposals/4439-support-contact-encryption.md
@@ -46,6 +46,10 @@ None identified.
 Sensitive communications may instead be conducted over Matrix, where E2EE is native. However, some researchers prefer
 or mandate out-of-band channels, which this MSC accommodates.
 
+Sensitive information may simply be encrypted to the key already listed in a site's security.txt file, however, at
+larger organizations this does not provide the same per-contact granularity (and may not even reference someone
+responsible for hosting the Matrix homeserver at all).
+
 ## Unstable prefix
 
 While this proposal is unstable, `pgp_key` should be referred to as `dev.zirco.msc4439.pgp_key`.
--- a/proposals/4439-support-contact-encryption.md
+++ b/proposals/4439-support-contact-encryption.md
@@ -12,9 +12,12 @@ other insecure channels.
 ## Proposal
 
 A new optional property `pgp_key` (unstable prefix: `dev.zirco.msc4439.pgp_key`) is added to the [`Contact`]
-response from [`/.well-known/matrix/support`]. This field indicates a PGP key that should be used for encrypted communication to that particular contact.
+response from [`/.well-known/matrix/support`]. This field indicates a PGP key that should be used for encrypted
+communication to that particular contact.
 
-The value of this field MUST be a URI pointing to a location where the key can be retrieved. Raw key material MUST NOT appear as the value of this field. As with [RFC9116], it is always the responsibility of the sender to ensure they trust the key provided.
+The value of this field MUST be a URI pointing to a location where the key can be retrieved. Raw key material MUST
+NOT appear as the value of this field. As with [RFC9116], it is always the responsibility of the sender to ensure they
+trust the key provided.
 
 Example of an OpenPGP key available from a web URI:
 
@@ -51,5 +54,4 @@ While this proposal is unstable, `pgp_key` should be referred to as `dev.zirco.m
   [`email_address`]: https://spec.matrix.org/unstable/client-server-api/#getwell-knownmatrixsupport_response-200_contact
   [`Contact`]: https://spec.matrix.org/unstable/client-server-api/#getwell-knownmatrixsupport_response-200_contact
   [RFC9116]: https://www.rfc-editor.org/info/rfc9116
-  [RFC7230]: https://www.rfc-editor.org/info/rfc7230
   [RFC7929]: https://www.rfc-editor.org/info/rfc7929
--- a/proposals/4439-support-contact-encryption.md
+++ b/proposals/4439-support-contact-encryption.md
@@ -6,16 +6,16 @@ as `security.txt`\) serves a similar purpose and defines an `Encryption` field (
 advertise a key URI for encrypted communication with security researchers.
 
 This proposal adds a similar `pgp_key` field to the [`Contact`] entry on [`/.well-known/matrix/support`], enabling
-homeserver operators to indicate a key that senders should use when communicating sensitive information over email or
+homeserver operators to indicate a key that senders may use when communicating sensitive information over email or
 other insecure channels.
 
 ## Proposal
 
 A new optional property `pgp_key` (unstable prefix: `dev.zirco.msc4439.pgp_key`) is added to the [`Contact`]
-response from [`/.well-known/matrix/support`]. This field indicates a PGP key that should be used for encrypted
-communication to that particular contact.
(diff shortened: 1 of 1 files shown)


proposals/4439-support-contact-encryption.md
turt2live, line 1: Implementation requirements:

* [ ] Server (publishing)
* [ ] Client (using)

proposals/4439-support-contact-encryption.md
```diff
@@ -0,0 +1,55 @@
...
+The value of this field MUST be a URI pointing to a location where the key can be retrieved. Raw key material MUST NOT appear as the value of this field. As with [RFC9116], it is always the responsibility of the sender to ensure they trust the key provided.
+
+Example of an OpenPGP key available from a web URI:
+
+```
+{
+    "contacts": [
+        {
+            "email_address": "logan@zirco.dev",
+            "encryption": "https://zirco.dev/pgp/logn.pub",
```
1vel: I feel `encryption` is a bit of a vague key name.

thetayloredman: That is fair enough. I originally considered `pgp_key` but thought to follow the precedent from security.txt. I am open to any suggestions for clearer names :)

1vel: `pgp_key` sounds fine to me

1vel:
(please avoid force-pushing after review is given, and please don't change the unstable prefix; it doesn't need to match the stable one)

proposals/4439-support-contact-encryption.md
CrazyNicc: another Server publishing the Changes: https://hnvn.de/.well-known/matrix/support

proposals/4439-support-contact-encryption.md
```diff
@@ -0,0 +1,57 @@
...
+as `security.txt`\) serves a similar purpose and defines an `Encryption` field (&sect;2.5.4) allowing operators to
+advertise a key URI for encrypted communication with security researchers.
+
+This proposal adds a similar `pgp_key` field to the [`Contact`] entry on [`/.well-known/matrix/support`], enabling
+homeserver operators to indicate a key that senders should use when communicating sensitive information over email or
+other insecure channels.
+
+## Proposal
+
+A new optional property `pgp_key` (unstable prefix: `dev.zirco.msc4439.pgp_key`) is added to the [`Contact`]
```
Johennes, line 14: I assume the field should only be set when `email_address` is set?

thetayloredman: Realistically, probably. Is there a reason to specify that though?

Hypothetically someone could use PGP-over Matrix and have no issue with it.

Johennes: Hm, but PGP-over Matrix is not part of the spec itself, so I guess that would be non-compliant behaviour to begin with? Anyway, I just feel that if there's a bunch of optional properties, it makes sense to document which ones are meant to be used together. Not sure if it needs to be normative.

thetayloredman: Yeah it seems pretty useless besides to maybe get initial trust in the user. There really is probably no scenario where they wouldn't come together. So I guess that's a SHOULD [RFC2119].

thetayloredman: eaf62815

proposals/4439-support-contact-encryption.md
```diff
@@ -0,0 +1,57 @@
...
+
+Other URI schemes other than `https` may also be used, common examples include, but are not limited to:
+- `openpgp4fpr:67FAAA655DBD691E7957E0951594E544D8F8F21E` (key fingerprint)
+- `dns:HASH._openpgpkey.zirco.dev?type=OPENPGPKEY` (`OPENPGPKEY` DNS record) ([RFC7929])
+
+## Potential issues
+
+None identified.
+
+## Alternatives
```
clokep, line 45: Why redefine what RFC9116 already defines? Would it make more sense to just use `security.txt`?

thetayloredman: A homeserver is unlikely to be the only service running on a domain. Especially for larger organizations, where the single key that security.txt provides may not actually represent anyone responsible for the Matrix homeserver itself. I would say that same logic applies to the very existence of the security role in the file at all in contrast to just using security.txt overall.

Additionally, this gets you per-contact granularity in contrast to just one "this is our security team's key."

clokep: I would include all that in the MSC.

turt2live (commented):
This has been queued for SCT review, but I think the Security team's opinion would be more helpful at this stage. Specifically, I'm interested to see if they'd use this and whether it's generic enough for their purposes. The remaining feature design components of whether to use a redirect, URL, base64, etc can likely be worked out by the SCT in a later stage, though the Security team might also have ideas to improve trust and security in the key availability too.

Other SCT members might have different opinions.

proposals/4439-support-contact-encryption.md
x86pup: Server publishing: https://woof.gay/.well-known/matrix/support

thetayloredman: no need to list your server here, I have an automated scanner

MTRNord: The Connectivity Tester now has support to display this https://github.com/MTRNord/matrix-connection-tester-ui/releases/tag/v0.5.1

proposals/4439-support-contact-encryption.md
```diff
@@ -0,0 +1,66 @@
...
+## Proposal
+
+A new optional property `pgp_key` (unstable prefix: `dev.zirco.msc4439.pgp_key`) is added to the [`Contact`]
+response from [`/.well-known/matrix/support`]. This field indicates a PGP key that may be used for encrypted
+communication to that particular contact. If the field is used, the `email_address` field SHOULD also be present.
+
+The value of this field MUST be a URI pointing to a location where the key may be retrieved. Raw key material MUST
+NOT appear as the value of this field. If a key fingerprint is to be used as this field, it MUST be prefixed with
+the `openpgp4fpr:` URI scheme. As with [RFC9116], it is always the responsibility of the sender to ensure they
+trust the key provided.
```
gewitternacht: As this explicitly references RFC9116, and reads very similar to its [section 2.5.4](https://www.rfc-editor.org/info/rfc9116/#section-2.5.4): Is there a specific reason the following bit from that section was omitted  in this MSC?
> If this field indicates a web URI, then it MUST begin with "https://" (...).

If there isn't, I think it wouldn't hurt to add it here too.

Or, alternatively, if the ways to specify a PGP key are supposed to be the exact same as in RFC9116, we could just omit this paragraph entirely and reference RFC9116 for details on the format of this field.

thetayloredman: That could definitely be added.

I intentionally avoided defining explicitly permitted types of URIs to allow for future expansion, only providing the *examples* of `https`, `openpgp4fpr`, and `dns`. It seems a couple implementations have interpreted this as "MUST be one of the three" and not "here are 3 examples." it might be good to clarify this and that seems like a potential route in that direction.

proposals/4439-support-contact-encryption.md
turt2live: Noting that the Security team ran into this exact alternative. Homeservers using a security disclosure program would use the widely adopted `security.txt` while others might use this MSC instead.

The Security team sees value in this MSC and suggests building something into the spec.

proposals/4439-support-contact-encryption.md
```diff
@@ -0,0 +1,67 @@
...
+        }
+    ]
+}
+```
+
+Other URI schemes other than `https` may also be used, common examples include, but are not limited to:
+- `openpgp4fpr:67FAAA655DBD691E7957E0951594E544D8F8F21E` (key fingerprint)
+- `dns:HASH._openpgpkey.zirco.dev?type=OPENPGPKEY` (`OPENPGPKEY` DNS record) ([RFC 7929])
+
+## Potential issues
```
thetayloredman, line 41: This proposal lacks a Security Considerations section, i.e. that of [RFC 9116](https://datatracker.ietf.org/doc/html/rfc9116#name-security-considerations)

PR #4440, opened 2026-03-27
MSC4440: Profile Biography via Global Profiles

RoootTheFox:
[Rendered](https://github.com/RoootTheFox/matrix-spec-proposals/blob/main/proposals/4440-profile-biography.md)

Implementations:
- Client (modifying bio):
  - Sable: https://github.com/SableClient/Sable/pull/559
  - gomuks: https://github.com/gomuks/gomuks/commit/753ddac15d7c17e94be9b0208c9cf200c3d96992 (+ follow-up commits)
  - Commet: https://github.com/commetchat/commet/pull/893
- Client (rendering bio):
  - Sable: https://github.com/SableClient/Sable/pull/559
  - gomuks: https://github.com/gomuks/gomuks/commit/753ddac15d7c17e94be9b0208c9cf200c3d96992 (+ follow-up commits)
  - Commet: https://github.com/commetchat/commet/pull/893

### DIFF
--- a/proposals/4440-profile-biography.md
+++ b/proposals/4440-profile-biography.md
@@ -1,0 +1,62 @@
+# MSC4440: Profile Biography via Global Profiles
+
+A lot of users, especially those switching to Matrix from other platform (e.g. Discord) are finding Matrix
+lacking for not providing a way to write information about themselves, e.g. their interests, and having
+others be able to view that. Several clients already have their own, client-specific implementations of a
+biography field via extended global profiles, creating a split in the ecosystem.
+To bring back standardization, this proposal defines a standardized biography field on top of [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+
+## Proposal
+
+Profiles may have an optional `m.biography` field as object. These fields can be fetched through the [profile API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles).
+Clients should display the biography of a user in their respective user popup/profile view UI.
+
+### Examples
+
+Plaintext only:
+```json
+{
+  "m.biography": {
+    "body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats"
+  }
+}
+```
+
+with HTML formatting:
+```json
+{
+  "m.biography": {
+    "body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats",
+    "format": "org.matrix.custom.html",
+    "formatted_body": "hello world!<br/><br/>interests:<br/><ul data-md=\"-\"><li><p>programming</p></li><li><p>matrix</p></li><li><p>sleeping</p></li><li><p>petting cats</p></li></ul><br/><img data-mx-emoticon src=\"mxc://fomx.gay/ICgRWFY6HWvMVVrHRqr7MYMLiTCgWxpl\" alt=\"bwaa\" title=\"bwaa\" height=\"32\" />"
+  }
+}
+```
+
+The `body` field should hold a plaintext representation of the users' biography, similar to room messages.
+Clients can optionally set the `formatted_body` field to an HTML-formatted representation of the body in the
+same format as [formatted messages](https://spec.matrix.org/v1.17/client-server-api/#mroommessage-msgtypes),
+in which case the format field has to be set to `org.matrix.custom.html`.
+When rendering a user's biography, the `formatted_body` should be preferred if set, but clients are free to
+render the plaintext representation in `body` instead.
+
+## Potential issues
+
+Unsure?
+
+## Alternatives
+
+None, really. Maybe linking a pastebin URL in your username?
+
+## Security considerations
+
+Malicious actors could set an unreasonably long bio, potentially lagging or even crashing clients, if length
+stays unlimited.
+
+## Unstable prefix
+
+Clients implementing this MSC early may use the profile key `gay.fomx.biography`.
+
+## Dependencies
+
+This MSC builds on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
--- a/proposals/4440-profile-biography.md
+++ b/proposals/4440-profile-biography.md
@@ -17,7 +17,9 @@ Plaintext only:
 ```json
 {
   "m.biography": {
-    "body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats"
+    "m.text": [
+      { "body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats" }
+    ]
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


proposals/4440-profile-biography.md
tulir, line 1: Implementation requirements:

* [x] Client (modifying bio)
* [x] Client (rendering bio)

Since there might be moderation concerns (the reason why MSC4133 didn't define freeform input fields):

* [ ] Review from trust & safety

proposals/4440-profile-biography.md
```diff
@@ -0,0 +1,62 @@
...
+Malicious actors could set an unreasonably long bio, potentially lagging or even crashing clients, if length
+stays unlimited.
+
+## Unstable prefix
+
+Clients implementing this MSC early may use the profile key `gay.fomx.biography`.
+
+## Dependencies
+
+This MSC builds on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
```
tulir: You don't need to list dependencies that are in the spec. If you do, it should say that the dependency is already in the spec

proposals/4440-profile-biography.md
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+with HTML formatting:
+```json
+{
+  "m.biography": {
+    "body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats",
+    "format": "org.matrix.custom.html",
+    "formatted_body": "hello world!<br/><br/>interests:<br/><ul data-md=\"-\"><li><p>programming</p></li><li><p>matrix</p></li><li><p>sleeping</p></li><li><p>petting cats</p></li></ul><br/><img data-mx-emoticon src=\"mxc://fomx.gay/ICgRWFY6HWvMVVrHRqr7MYMLiTCgWxpl\" alt=\"bwaa\" title=\"bwaa\" height=\"32\" />"
```
tulir: Copying the old message format is probably not a great idea since all other new things (like html-formatted room topics) use extensible events

proposals/4440-profile-biography.md
```diff
@@ -0,0 +1,62 @@
...
+When rendering a user's biography, the `formatted_body` should be preferred if set, but clients are free to
+render the plaintext representation in `body` instead.
+
+## Potential issues
+
+Unsure?
+
+## Alternatives
+
+None, really. Maybe linking a pastebin URL in your username?
```
nushea: Not a great option but the m.presence status_msg field which is user modifiable is an alternative if not blocked by the homeserver (which only allows regular text, no formatting or anything)

FSG-Cat: I mean alternatives exist considering proposals like https://github.com/matrix-org/matrix-spec-proposals/pull/4201 exists and its all about an alternative way to get this done.

So clearly i would say that this should be updated to address the list of alternatives and why this proposal should be considered instead of other approaches.

I will say that 4201 needs work to be considered (As in its current state as a MSC is not good enough. It was a rushjob of a MSC tbh and it needs work if it ever wants to be viable for FCP assuming implementations existed.)

proposals/4440-profile-biography.md
RoootTheFox: so, essentially changing the format to this?
```json
{
  "m.biography": {
    "m.text": [
      {"body": "hello world!<br/><br/>interests:<br/><ul data-md=\"-\"><li><p>programming</p></li><li><p>matrix</p></li><li><p>sleeping</p></li><li><p>petting cats</p></li></ul><br/><img data-mx-emoticon src=\"mxc://fomx.gay/ICgRWFY6HWvMVVrHRqr7MYMLiTCgWxpl\" alt=\"bwaa\" title=\"bwaa\" height=\"32\" />", "mimetype": "text/html"},
      {"body": "hello world!\n\ninterests:\n-  programming\n-  matrix\n-  sleeping\n-  petting cats"} 
    ]
  }
}
```

proposals/4440-profile-biography.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Security considerations
```
FSG-Cat, line 54: I would say that the sizable Trust and Saftey considerations that need to be made should belong here or in a dedicated section.

Because malicious actors setting large info is the least of the concerns.

I would be concerned with all the various types of harm that will be put into these bios because you need only look at Discord, Steam and various videogames to know people will abuse this.

A system like this needs serious design considerations for how to allow moderators to handle it. This is why the MSC that added the Extended global profiles in the first place had to stay away from this as its survival as a proposal was essentially made possible by postponing this question to this MSC.

proposals/4440-profile-biography.md
RoootTheFox: i feel like #4201 is more of an alternative to #4133 instead of this proposal specifically; i do get your point though.
when #4201 is in-spec, this proposal could be reimplemented on top of that, though i feel like it's much better to base this proposal on top of existing infrastructure which is in-spec (in this case, #4133, which other MSCs such as #4247 also base themselves on)

proposals/4440-profile-biography.md
snoweuph: > I would be concerned with all the various types of harm that will be put into these bios because you need only look at Discord, Steam and various videogames to know people will abuse this.

Isn't that a general moderation concern, you can have about most things in matrix?

In the sense of moderation, this doesn't really differ from profile pictures, which can hold bad content like CSAM.

As far as I understand matrix, the moderation, is always up to instance owners and room administrators, otherwise you would need to centralize.

In that sense, the instance trust system also plays into this general idea of moderation and keeping safe spaces. For example, a homeserver could block all profile pictures and descriptions from untrusted instances.

FSG-Cat: Room administrators have absolutely no possible say under this proposal is an example of a concern because it builds on a proposal that is global profile only scoped not per room configurable.

snoweuph: Room Administrators, can always ban/kick people, that have things that aren't wanted in the space, and can report to the homeserver administrators. So no, they don't have a say in this proposal directly, but the general moderation flow still applies, that's what I wanted to point at.

It's basicly the same on Discord if you think about it, where Guild Admins can ban people that have stuff they don't want on their Guild, and report illegal stuff to discord.

proposals/4440-profile-biography.md
RoootTheFox: fwiw, there are two implementations now (one merged) ^^

proposals/4440-profile-biography.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Alternatives
+
+Maybe linking a pastebin URL in your username?
+Potentially setting an m.presence status_msg could work too, though that is much more limited.
+
+## Security considerations
+
+- Malicious actors could set an unreasonably long bio, potentially lagging or even crashing clients, if length
+stays unlimited.
```
tulir, line 57: This is probably not a real problem, profiles have the same 64 KiB limit as messages and clients already render html messages

(of course this relies on servers implementing the profile size limit correctly both locally and over federation)

proposals/4440-profile-biography.md
tulir: https://github.com/gomuks/gomuks/commit/9fd0c63243d48ee3509c74147747d2ca88d3654e handles the T&S side: when viewing the profile of a banned user, all fields with arbitrary user input are blurred

RoootTheFox: > [gomuks/gomuks@9fd0c63](https://github.com/gomuks/gomuks/commit/9fd0c63243d48ee3509c74147747d2ca88d3654e) handles the T&S side: when viewing the profile of a banned user, all fields with arbitrary user input are blurred

perhaps behavior like this could become part of this MSC (or even the spec in general)?
i feel like this is a pretty reasonable and good default to have, not just for biographies but also names and avatars.

that'd pretty much solve the T&S issues of not just this MSC but profiles in general

tulir: Yeah the security consideration should probably be reworded to say clients SHOULD hide user-definable global profile fields when viewing the profile of a banned user

PR #4441, opened 2026-03-30
MSC4441: User Profile Annotations ("Profile Notes")

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/profile-annotations/proposals/4441-user-profile-annotations.md)

Implementations:
* [Rust SDK](https://github.com/matrix-org/matrix-rust-sdk/commit/d617a9d53e48c259f83873a87af7379ff22fbfa4) (outdated - implements up to 8b0dec4)
* gomuks (no E2EE): https://github.com/gomuks/gomuks/pull/744

Dependencies:
* https://github.com/matrix-org/matrix-spec-proposals/pull/4483

### DIFF
--- a/proposals/4441-user-profile-annotations.md
+++ b/proposals/4441-user-profile-annotations.md
@@ -127,10 +127,11 @@ etc.
 
 Before this MSC is accepted, implementations should use the unstable `account_data` events:
 
-| Stable identifier       | Purpose                                  | Unstable identifier                     |
-| ----------------------- | ---------------------------------------- | --------------------------------------- |
-| `m.profile_annotations` | Storage of User Profile Annotations      | `dev.zirco.msc4441.profile_annotations` |
-| `m.account_data.key`    | ADK secret storage for encrypted content | `dev.zirco.msc4441.account_data.key`    |
+| Stable identifier        | Purpose                                  | Unstable identifier                      |
+| ------------------------ | ---------------------------------------- | ---------------------------------------- |
+| `m.profile_annotations`  | Storage of User Profile Annotations      | `dev.zirco.msc4441.profile_annotations`  |
+| `m.profile_annotations/` | Prefix of HKDF `info`                    | `dev.zirco.msc4441.profile_annotations/` |
+| `m.account_data.key`     | ADK secret storage for encrypted content | `dev.zirco.msc4441.account_data.key`     |
 
 ## Dependencies
 
--- a/proposals/4441-user-profile-annotations.md
+++ b/proposals/4441-user-profile-annotations.md
@@ -1,7 +1,10 @@
 # MSC4441: Encrypted User Profile Annotations via Account Data
 
 Many platforms, such as Discord, provide a capability for a user to leave personal "notes" on a user's profile. Matrix,
-however, has no similar functionality that can be shared between clients. This proposal builds User Profile Annotations,
+however, has no similar functionality that can be shared between different clients. This proposal builds User Profile
+Annotations, a method for storing encrypted, private metadata on user and room profiles.
+
+This proposal builds User Profile Annotations,
 a method of storing personal context on other users' profiles. This is made possible using a new `m.profile_annotations`
 Account Data event, allowing the storage of free text (and other annotations) for a user's reference.
 
@@ -11,11 +14,18 @@ annotations is also encrypted using an account secret scoped specifically for ac
 
 ## Proposal
 
+This proposal introduces a new `m.profile_annotations` Account Data event, allowing the storage of free text (and other
+future annotations) on users. The framework introduced within this MSC builds on [Extensible Events][MSC1767] and
+provides room for future extension into other forms of user context, such as contact labels, custom nickname overrides,
+or annotations on rooms.
+
 ### Plaintext `m.profile_annotations`
 
-A new event, `m.profile_annotations`, is stored in [`account_data`][account_data]. The `content` of this event is a JSON
-mapping between MXIDs (user or room IDs) and annotations stored on that user/room. Within a user, the `m.text`
+A new event, `m.profile_annotations`, is stored in [`account_data`][account_data]. The `content` of this event MUST be a
+mapping between valid user MXIDs or room IDs and annotations stored on that user/room. Within a user, the `m.text`
 property represents a text annotation ("note") on that user. It follows the text format as defined in [MSC1767][MSC1767].
+
+Unknown keys within user annotations MUST be ignored.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


proposals/4441-user-profile-annotations.md
tulir, line 1: Implementation requirements:

* [ ] Client (creating annotations)
* [ ] Client (rendering annotations)

turt2live:
The author would like review of this MSC prior to implementation, especially around crypto. See https://matrix.to/#/!0KNSXYXB_2xtEUkQ9MGBRy5oNIOfAKoq2uIqPZCJbI8/$Zss0QZ3VKtAqDlOCj4oBD2U7EO3UWVSgQ5xX3DuXe0Y?via=element.io&via=matrix.org&via=codestorm.net

proposals/4441-user-profile-annotations.md
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+}
+```
+
+### Encrypted `m.profile_annotations`
+
+Using the ADK bytes introduced above, contents of `m.profile_annotations` can be encrypted using
+[`m.secret_storage.v1.aes-hmac-sha2`][SS_crypto]. The info property of the HKDF is the value `m.profile_annotations`,
+an ASCII `/`, concatenated with the MXID of the user it represents (e.g. `m.profile_annotations/@logn:zirco.dev`).
```
thetayloredman: Should the unstable identifier be used here as well?

thetayloredman:
This may also benefit from the attempts at solving RMW races in https://github.com/matrix-org/matrix-spec-proposals/pull/4438. The addition of an ADK for encrypting account data could also potentially benefit that MSC and could potentially be factored out into a separate proposal?

proposals/4441-user-profile-annotations.md
```diff
@@ -0,0 +1,142 @@
...
+Account data is last-write-wins. The usage of one key representing all users may lead to a data race in rare scenarios
+(although this tradeoff was accepted to keep the behavior similar to existing state like `m.direct`).
+
+## Alternatives
+
+#### Store annotations as a new API endpoint
+Rather than using `account_data`, a new dedicated endpoint for annotations would be created. Rejected because it
+requires substantial serverside changes, this MSC aims to be entirely clintside.
+
+#### Encrypt the entire event as one blob
```
Johennes: If the contents of each item are encrypted separately, how does storing all items in a single account data entry compare to using one account data entry per item (with a common prefix)?

proposals/4441-user-profile-annotations.md
```diff
@@ -0,0 +1,142 @@
...
+
+This proposal introduces a new `m.profile_annotations` Account Data event, allowing the storage of free text (and other
+future annotations) on users. The framework introduced within this MSC builds on [Extensible Events][MSC1767] and
+provides room for future extension into other forms of user context, such as contact labels, custom nickname overrides,
+or annotations on rooms.
+
+### Plaintext `m.profile_annotations`
+
+A new event, `m.profile_annotations`, is stored in [`account_data`][account_data]. The `content` of this event MUST be a
+mapping between valid user MXIDs or room IDs and annotations stored on that user/room. Within a user, the `m.text`
```
Johennes: Could it be easier to just use room account data if the subject is a room?

I also find it slightly unintuitive that `m.profile_annotations` contains rooms because "room profile" is not a common term in the ecosystem to my knowledge.

proposals/4441-user-profile-annotations.md
thetayloredman: That could definitely be considered. It would reduce the risk of a data race, however my concerns would be:
- There is existing precedent for keeping it together, e.g. in `m.direct`.
- How well do homeservers handle lots and lots of account_data events?

proposals/4441-user-profile-annotations.md
thetayloredman: I do not build much on the idea of similar annotations on rooms in this MSC, so it could be eliminated entirely as out of scope for this first round. Probably not a bad idea to keep it dedicated to users especially when it implies "profiles" and the use case for rooms would likely be drastically different.

thetayloredman: 3c05a84e

proposals/4441-user-profile-annotations.md
```diff
@@ -0,0 +1,143 @@
+# MSC4441: Encrypted User Profile Annotations via Account Data
+
+Many platforms, such as Discord, provide a capability for a user to leave personal "notes" on a user's profile. Matrix,
+however, has no similar functionality that can be shared between different clients. This proposal builds User Profile
+Annotations, a method for storing encrypted, private metadata on user profiles.
```
turt2live, line 5: the introduction should explain why encryption is required for this feature

proposals/4441-user-profile-annotations.md
```diff
@@ -0,0 +1,143 @@
...
+
+This proposal builds User Profile Annotations,
+a method of storing personal context on other users' profiles. This is made possible using a new `m.profile_annotations`
+Account Data event, allowing the storage of free text (and other annotations) for a user's reference.
+
+The framework introduced within this MSC builds upon [Extensible Events][MSC1767] and provides room for future extension
+into other forms of user context, such as contact labels or custom nickname overrides. The content stored within these
+annotations is also encrypted using an account secret scoped specifically for account data.
```
turt2live: We've reclassified account data to *not* be an event in Matrix. Extensible events are also meant to improve rendering ability, not to provide a fallback structure - using them for profiles feels wrong, especially if those profiles are in account data rather than room events.

turt2live (commented):
Some early review as an individual on the SCT - other SCT members might have other opinions. Sorry it's taken so long to get feedback.

Overall: it's not clear to me why encryption is needed or why extensible events is used. The title and actual proposal appear to be different as well: this proposal adds a local notes field which isn't mentioned in the title at all.

turt2live:
@thetayloredman note that for this MSC to be eligible to enter FCP, it will need [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md) as well.

proposals/4441-user-profile-annotations.md
thetayloredman: 9f4151c7 makes this no longer relevant - all annotations are now encrypted as one blob.

proposals/4441-user-profile-annotations.md
thetayloredman: 9f4151c7 rephrased this introduction to clarify that encryption benefits users who may place sensitive information in these profile annotations, alongside rephrasing the notes field.

proposals/4441-user-profile-annotations.md
thetayloredman: 9f4151c7 makes it clearer that the annotation data is not an event, it simply uses the same fallback structure as extev, in a similar light to [MSC4440](https://github.com/RoootTheFox/matrix-spec-proposals/blob/main/proposals/4440-profile-biography.md#examples).

proposals/4441-user-profile-annotations.md
nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/fdd0308627f0946541326d57aefff0bd91499e7b

PR #4445, opened 2026-04-01
MSC4445: Clarify `/sync` timeline order

### LINKED ISSUE #1917: Clarify order of initial vs incremental `/sync`
**Link to problem area**:

Spec: https://spec.matrix.org/v1.10/client-server-api/#syncing

> Events are ordered in this API according to the arrival time of the event on the homeserver. This can conflict with other APIs which order events based on their partial ordering in the event graph. This can result in duplicate events being received (once per distinct API called). Clients SHOULD de-duplicate events based on the event ID when this happens.

**Issue**

*Spawning from a [discussion](https://github.com/element-hq/synapse/pull/17510#discussion_r1699105569) with @erikjohnston,*

The spec clearly states that `/sync` should return events according to the "arrival time of the event on the homeserver" (`stream_ordering`). This is not how [Synapse has been behaving](https://github.com/element-hq/synapse/blob/e3db7b2d81cabc7e4335afc051e28678e3a9dd02/synapse/handlers/sync.py#L882-L897) since the [beginning](https://github.com/element-hq/synapse/commit/a56008842b43089433768f569f35b2d14523ac39). https://github.com/element-hq/synapse/commit/e2accd7f1d21e34181dd4543eca30ad1ea971b4c even claims the following in the commit message:

> The sync API often returns events in a topological rather than stream
ordering, e.g. when the user joined the room or on initial sync. When
this happens we can reuse existing pagination storage functions.

Normally, this would just be a spec-compliance problem in Synapse but I think this may be the right way to think about it.

### Proposal

For initial `/sync`, we want to view a historical section of the timeline; to fetch events by `topological_ordering` (partial ordering in the event graph) (best representation of the room DAG as others were seeing it at the time). This also aligns with the order that `/messages` returns events in. We should also extend this to any time we're initially returning a room (therefore historical events) to the user (like a newly joined room). This behavior basically results in the same outcome as if no history was sent down `/sync` and `/messages` was used instead.

For incremental `/sync`, we want to get all updates for rooms since the last `/sync` (regardless if those updates arrived late or happened a while ago in the past); to fetch events by `stream_ordering` (in the order they were received by the server).

### Related issues

 - https://github.com/matrix-org/matrix-spec/issues/1334
 - https://github.com/matrix-org/matrix-spec/issues/125
 - https://github.com/matrix-org/matrix-spec/issues/852
 - https://github.com/matrix-org/matrix-spec-proposals/pull/4033
     - https://github.com/matrix-org/matrix-spec/issues/1167

MadLittleMods:
MSC4445: Clarify `/sync` timeline order

Fix https://github.com/matrix-org/matrix-spec/issues/1917

---

[Rendered](https://github.com/MadLittleMods/fork-matrix-spec-proposals/blob/madlittlemods/sync-timeline-order/proposals/4445-sync-timeline-order.md)

Server implementations:

 - [Synapse](https://github.com/element-hq/synapse) since inception (2015)
 - Unstable flag in Synapse: https://github.com/element-hq/synapse/pull/19642

### DIFF
--- a/proposals/4445-sync-timeline-order.md
+++ b/proposals/4445-sync-timeline-order.md
@@ -1,6 +1,6 @@
 # MSC4445: Clarify `/sync` timeline order
 
-Currently, the spec describes order of events in the `timeline` of `/sync` in this manner:
+Currently, the spec describes the order of events in the `timeline` of `/sync` in this manner:
 
 > Events are ordered in this API according to the arrival time of the event on the
 > homeserver. This can conflict with other APIs which order events based on their
--- a/proposals/4445-sync-timeline-order.md
+++ b/proposals/4445-sync-timeline-order.md
@@ -23,7 +23,7 @@ claims the following:
 ordering, e.g. when the user joined the room or on initial sync. When
 this happens we can reuse existing pagination storage functions.
 >
-> *-- [Synapse `e2accd7` commit message*](https://github.com/element-hq/synapse/commit/e2accd7f1d21e34181dd4543eca30ad1ea971b4c)
+> *-- [Synapse `e2accd7` commit message](https://github.com/element-hq/synapse/commit/e2accd7f1d21e34181dd4543eca30ad1ea971b4c)*
 
 Normally, this would just be a spec-compliance problem in Synapse but I think this may
 be the right way to think about it.
--- a/proposals/4445-sync-timeline-order.md
+++ b/proposals/4445-sync-timeline-order.md
@@ -1,4 +1,4 @@
-# MSC0000: Clarify `/sync` timeline order
+# MSC4445: Clarify `/sync` timeline order
 
 Currently, the spec describes order of events in the `timeline` of `/sync` in this manner:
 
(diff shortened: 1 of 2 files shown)


proposals/4445-sync-timeline-order.md
turt2live, line 1: Implementation requirements:

* [x] Evidence of long-standing non-compliance
* [ ] Client (unstable flag: topological ordering)
* [ ] Client (unstable flag: stream ordering)
* [ ] Backwards compatibility research ("will clients break with this change?" and "what do non-Synapse servers do?")

proposals/4445-sync-timeline-order.md
```diff
@@ -0,0 +1,136 @@
...
+
+Clients can rely on the presence of the
+`org.matrix.msc4445.initial_sync_timeline_stream_ordering` and
+`org.matrix.msc4445.initial_sync_timeline_topological_ordering` flags in
+`unstable_features` as an explicit indicator to tell how a homeserver is behaving. If
+these flags are absent, the client has to assume the server is behaving as currently
+specc'ed but has no guarantee of this.
+
+Synapse already behaves in the way described in this proposal (since inception in 2015)
+and is unlikely to change it's behavior until a conclusion is made with this MSC.
```
andybalaam:
```suggestion
and is unlikely to change its behavior until a conclusion is made with this MSC.
```

PR #4448, opened 2026-04-09
Add MSC4448: Preview URL Site Logos

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/msc4448-preview-url-site-logos/proposals/4448-preview-url-site-logo.md)


proposals/4448-preview-url-site-logo.md
turt2live, line 1: Implementation requirements:

* [ ] Client (rendering)
* [ ] Server (serving)

proposals/4448-preview-url-site-logo.md
```diff
@@ -0,0 +1,98 @@
...
+Therefore, this proposal suggests an extension to the existing API to provide a site logo in addition to
+other content.
+
+
+## Proposal
+
+Additional fields MAY be provided in the response to `GET /_matrix/client/v1/media/preview_url`:
+
+ - `matrix:site_logo` which is a `mxc://` URI to the site's logo. Omitted if there is no logo.
+ - `matrix:site_logo:size` which is the byte-size of the image. Omitted if there is no logo.
```
turt2live, line 27: In general we should be keeping our opengraph extensions to a minimum. If there's already a field we can reuse, servers can use that. The MSC would simply become adding an awareness mechanism to the spec so other implementations know about it.

Half-Shot: Yeah, looked far and wide and there is nothing for it. Sites like BlueSky use `og:logo` but it's non-standard. https://ogp.me/ is rather underwhelming, and I'm not sure we can expect any movement for the foreseeable.

Given I think this is a valuable and reasonably trivial extension, I think it's excusable for the moment. If there was a change to the OpenGraph spec, I would most certainly open a subsequent MSC to deprecate and move to that.

proposals/4448-preview-url-site-logo.md
```diff
@@ -0,0 +1,98 @@
...
+This introduces a custom field into a spec largely defined outside of Matrix, and could perhaps be floated
+as an extension to the standard itself. For the moment though, it feels fine to experiment at the Matrix layer
+to prove the concept.
+
+### Local URL Previews
+
+This proposal is centred around extending the server-side API to render previews, rather than working to
+make client-side previews easier. There are arguments that client-side previews are the better bet for the
+future of Matrix, however this proposal at least standardises the response body for server-side responses
+which could be used by client-side libraries, so there is still some benefits which aid both approaches.
```
tulir, line 59: I recently updated MSC4095 to specify that anything added to the `/preview_url` response can also be used in bundled URL previews

Half-Shot: Reminder to myself to reference this MSC :)

PR #4450, opened 2026-04-14
MSC4450: Identity Provider selection for User-Interactive Authentication with Legacy Single Sign-On

reivilibre:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rei/uiaa_sso_web_fallback_idp_id/proposals/4450-identity-provider-selection-uia-legacy-sso.md)

### DIFF
--- a/proposals/4450-identity-provider-selection-uia-legacy-sso.md
+++ b/proposals/4450-identity-provider-selection-uia-legacy-sso.md
@@ -1,0 +1,70 @@
+# MSC4450: Identity Provider selection for User-Interactive Authentication with Legacy Single Sign-On
+
+When a client presents a user with a User-Interactive Authentication flow using the Single Sign-On
+authentication type (`m.login.sso`), the client must direct the user to complete the flow in the
+browser, using the fallback URL at `/_matrix/client/v3/auth/m.login.sso/fallback/web?session=<session ID>`.
+
+(Note that despite being called 'fallback', there is no other method for completing `m.login.sso`.)
+
+It is possible for a homeserver to have multiple Single Sign-On providers (also known as 'Identity Providers').
+
+There is currently no mechanism for the client to indicate a preference/selection of identity provider
+to be used for the User-Interactive Authentication flow.
+
+This proposal adds such a mechanism.
+
+
+## Proposal
+
+An `idp_id` query parameter is introduced to the `m.login.sso` fallback endpoint:
+
+```
+/_matrix/client/v3/auth/m.login.sso/fallback/web?session=<session ID>&idp_id=<IDP ID>
+```
+
+The IDP IDs (Identity Provider identifiers) are the same as used in the [`GET /_matrix/client/v3/login/sso/redirect/{idpId}`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3loginssoredirectidpid) endpoint.
+
+If the Identity Provider identifier is not set, the homeserver MUST act the same as before this MSC
+was introduced, for example automatically selecting an Identity Provider or offering a web user interface
+for the user to do so manually.
+
+If the Identity Provider identifier refers to an Identity Provider that is non-existent,
+not associated to the user or otherwise invalid,
+the homeserver MAY either treat the parameter as not having been set, or display an error to the user,
+depending on which is more appropriate for the homeserver's circumstances.
+
+
+## Potential issues
+
+- Legacy authentication is already somewhat in the process of being replaced by OAuth 2.0 authentication,
+  so perhaps it's not worth proceeding with changes to legacy authentication.
+
+
+## Alternatives
+
+- We could instead make it possible for the homeserver to offer multiple `m.login.sso` authentication options
+  and have a generic authentication 'subtype' system for selecting one of multiple independent `m.login.sso`
+  authentication options.
+  
+  This 'subtype' approach could generalise beyond `m.login.sso`.
+  
+  The fallback for authentication subtypes could then be exposed at e.g. `/_matrix/client/v4/auth/<auth type>/<auth subtype>/fallback/web?session=<session ID>`
+
+
+## Security considerations
+
+None anticipated.
+
+
+## Unstable prefix
+
+Whilst this MSC is unstable, the unstable-prefixed query parameter `io.element.idp_id` should be used in place of `idp_id`.
+
+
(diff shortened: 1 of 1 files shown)


proposals/4450-identity-provider-selection-uia-legacy-sso.md
```diff
@@ -0,0 +1,70 @@
+# MSC4450: Identity Provider selection for User-Interactive Authentication with Legacy Single Sign-On
+
+When a client presents a user with a User-Interactive Authentication flow using the Single Sign-On
+authentication type (`m.login.sso`), the client must direct the user to complete the flow in the
+browser, using the fallback URL at `/_matrix/client/v3/auth/m.login.sso/fallback/web?session=<session ID>`.
+
+(Note that despite being called 'fallback', there is no other method for completing `m.login.sso`.)
```
MadLittleMods, line 7: As-is, this statement is confusing as [`m.login.sso` also applies to login](https://spec.matrix.org/v1.18/client-server-api/#sso-client-loginauthentication).

More clear:

```suggestion
(Note that despite being called 'fallback', there is no other method for completing User-Interactive Authentication for the `m.login.sso` type.)
```


For my own reference, "User-Interactive Authentication" is separate from the login flow. User-Interactive Authentication is used to authorize a specific action (like when changing your password, deleting devices, etc) when you're already logged in.

MadLittleMods: Feels like this should instead be re-framed to explain that it's a fallback for when a client doesn't know how to handle a specific login type.

proposals/4450-identity-provider-selection-uia-legacy-sso.md
```diff
@@ -0,0 +1,70 @@
...
+
+The IDP IDs (Identity Provider identifiers) are the same as used in the [`GET /_matrix/client/v3/login/sso/redirect/{idpId}`](https://spec.matrix.org/v1.15/client-server-api/#get_matrixclientv3loginssoredirectidpid) endpoint.
+
+If the Identity Provider identifier is not set, the homeserver MUST act the same as before this MSC
+was introduced, for example automatically selecting an Identity Provider or offering a web user interface
+for the user to do so manually.
+
+If the Identity Provider identifier refers to an Identity Provider that is non-existent,
+not associated to the user or otherwise invalid,
+the homeserver MAY either treat the parameter as not having been set, or display an error to the user,
```
MadLittleMods, line 33: > the homeserver MAY either treat the parameter as not having been set

Feels like the homeserver should always error if it supports this parameter.

For homeservers that don't support this parameter, chances are it will be ignored though but that should just be noted as a client expectation (as it is already).

proposals/4450-identity-provider-selection-uia-legacy-sso.md
```diff
@@ -0,0 +1,70 @@
...
+depending on which is more appropriate for the homeserver's circumstances.
+
+
+## Potential issues
+
+- Legacy authentication is already somewhat in the process of being replaced by OAuth 2.0 authentication,
+  so perhaps it's not worth proceeding with changes to legacy authentication.
+
+
+## Alternatives
```
MadLittleMods, line 43: Given the homeserver is the one returning the `session` ID to be used in the User-Interactive Authentication flow, doesn't the homeserver have some idea about which SSO provider was used to login the user in the first place. And the homeserver could be using the same SSO provider for the fallback endpoint.

For my reference on how "User-Interactive Authentication" works:

 1. Client tries to change password: `POST /_matrix/client/v3/account/password` with `{ "new_password": "newpass" }` -> `401` `{ "flows": [{ "stages": ["m.login.sso"] }], "params":  { ... }, "session": "xyz123" }`
 1. Client directs the user to `/_matrix/client/v3/auth/m.login.sso/fallback/web?session=xyz123`
 1. User authorizes and the final HTML page has a few triggers to notify the opener about the flow being complete. For client apps that can't hook the `postMessage` notification, they usually just ask the user to press next when it's complete.
 1. The client tries the same `POST /_matrix/client/v3/account/password` request but with `{ "new_password": "newpass", "auth": { "session": "xyz123" } }` and the homeserver validates that the user ID of the requester [matches](https://github.com/element-hq/synapse/blob/943da0ace86cc1d608696ce23960269f75d87f5f/synapse/handlers/auth.py#L390-L392) the one in the session.

proposals/4450-identity-provider-selection-uia-legacy-sso.md
```diff
@@ -0,0 +1,70 @@
...
+
+There is currently no mechanism for the client to indicate a preference/selection of identity provider
+to be used for the User-Interactive Authentication flow.
+
+This proposal adds such a mechanism.
+
+
+## Proposal
+
+An `idp_id` query parameter is introduced to the `m.login.sso` fallback endpoint:
```
MadLittleMods, line 19: How do we expect the client to know which `idp_id` to use from the list of [`identity_providers`](https://spec.matrix.org/v1.18/client-server-api/#definition-mloginsso-flow-schema)?

Especially in the face of:

> Presenting a web UI for manual selection is not useful in this customer's situation as it would only confuse the user.
>
> *-- https://github.com/element-hq/synapse/issues/19691*

proposals/4450-identity-provider-selection-uia-legacy-sso.md
MadLittleMods: To answer my own question, Synapse already does this but users can be associated with multiple identity providers and we just [arbitrarily select one](https://github.com/element-hq/synapse/blob/0be7fe926d54c9b4de87fe83d8b75a82482b10d3/synapse/handlers/auth.py#L1732-L1750). People want to be able to use whichever identity provider to authorize the action.

proposals/4450-identity-provider-selection-uia-legacy-sso.md
MadLittleMods: Reading https://github.com/element-hq/synapse/issues/19691 a bit closer, in this case, it sounds like each client has a dedicated identity provider which is probably hard-coded.

(I'm unable to resolve this discussion)

proposals/4450-identity-provider-selection-uia-legacy-sso.md
MadLittleMods: hmm, thinking about this more, this could also be solved by Synapse picking a more appropriate identity provider (the one you logged in with that particular the client) instead of picking an arbitrary identity provider used on your account.

PR #4451, opened 2026-04-16
MSC4451: Deprecate notifications endpoint

JadedBlueEyes:
[Rendered](https://github.com/JadedBlueEyes/matrix-spec-proposals/blob/jade/reprecate-notifications-endpoint/proposals/4451-notification-endpoint.md)

For an implementation of notification history without this endpoint, see [gomuks](https://github.com/gomuks/gomuks/tree/main)
For an example of a client disabling functionality based on this endpoint, see [Element Web](https://github.com/element-hq/element-web/commit/a5e186b9786c387c999ecc70251ba01059add194)

### DIFF
--- a/proposals/4451-notification-endpoint.md
+++ b/proposals/4451-notification-endpoint.md
@@ -1,0 +1,54 @@
+# MSC4451: Deprecate Notifications Endpoint
+
+The `/_matrix/client/v3/notifications` endpoint was added more than a decade ago. In the meantime,
+end-to-end encryption was added, and the server can no longer accurately evaluate notifications for
+all events.
+
+Because the server cannot see the event type or event contents of E2EE messages, it cannot reliably
+decide whether a given message should trigger a notification. It cannot accurately evaluate push rules
+based on event types being visible, or based on mention keywords within the message contents.
+As a result, the server's view of notifications is inherently incomplete. It must either miss all
+encrypted events, or incorrectly assume that all encrypted events cause notifications.
+
+This means that this endpoint necessarily returns incomplete and incorrect results to the client,
+encouraging an incomplete and buggy implementation on the client side.
+
+The endpoint is also relatively unpopular, with few clients using it and few servers implementing it.
+
+[MSC4076](https://github.com/matrix-org/matrix-spec-proposals/pull/4076) (Let E2EE clients calculate app badge counts themselves)
+also exists for much the same reason.
+
+## Proposal
+
+Deprecate the `GET /_matrix/client/v3/notifications` endpoint.
+
+In a future version of the Matrix specification, the endpoint will be removed entirely. For now, it
+will be marked as deprecated in the specification. Clients are advised to calculate notifications
+locally using data stored from sync, locally executing push rules against decrypted event contents
+to accurately determine if an event should notify.
+
+## Potential Issues
+
+The few clients using this endpoint will have to start moving towards alternative, more correct
+implementations (i.e. using locally stored data and evaluating push rules client-side). If servers
+chose to remove the endpoint before it is officially removed from the specification, it could result
+in a reduction in functionality for those legacy clients.
+
+## Alternatives
+
+Adding a warning to the specification, and clarifying the limited behavior of the endpoint.
+Perhaps clients could use the endpoint to optimise a more complete implementation - however,
+it's not possible to skip "correctly" indexing a room because encrypted events can practically
+be sent in rooms without the `m.encryption` state event.
+
+## Security considerations
+
+None.
+
+## Unstable prefix
+
+Unneeded, as this is a deprecation of an existing endpoint.
+
+## Dependencies
+
+None.


proposals/4451-notification-endpoint.md
turt2live, line 1: Implementation requirements:

* [ ] Evidence that the endpoint is not fit for purpose, or has a replacement. For example: clients *not* using it, clients having bug reports about it, or an MSC which handles the desired functionality better.

proposals/4451-notification-endpoint.md
```diff
@@ -0,0 +1,54 @@
...
+all events.
+
+Because the server cannot see the event type or event contents of E2EE messages, it cannot reliably
+decide whether a given message should trigger a notification. It cannot accurately evaluate push rules
+based on event types being visible, or based on mention keywords within the message contents.
+As a result, the server's view of notifications is inherently incomplete. It must either miss all
+encrypted events, or incorrectly assume that all encrypted events cause notifications.
+
+This means that this endpoint necessarily returns incomplete and incorrect results to the client,
+encouraging an incomplete and buggy implementation on the client side.
```
1vel, line 14: Not everyone uses E2EE, and the large majority of public rooms aren't encrypted either.

This also encourages less efficient implementations, as, if the client doesn't have all of the messages in every room the user is in (which is a quite common case I'd imagine), it would then need to paginate all of the history of every room the user is in, which doesn't scale very well. (perhaps a client could limit the pagination it does, but that also results in being potentially incomplete)

Having the server process this scales way better as the server has the events after the user joined (which are likely to be the most important to the user) already.

I think a more rational approach would be to keep the endpoint and for clients to inform the user that encrypted messages won't show.

JadedBlueEyes: A few points:

Having a partially working feature is poor UX and creates a split world of functionality - directly contrary to end to end encryption. For many people a significant portion or all of the rooms they are in are encrypted (I'm one of them). Adding some text doesn't fix this.

As far as I'm aware for shipping functionality, only Cinny (and forks) uses this endpoint anyway. Only Cinny and Gomuks implement notification history.

I expect that for implementing notification history by itself, indexing/pagination will not be needed. Gomuks doesn't do this, for example, as it assumes all notifications a user cares about on a device are after they logged in. Because client should be woken by push, they can add notifications as they receive them. If clients aren't woken, they can't reliably receive notifications anyway.

As clients implementing search will need to index events anyway, this can feed into the same system as push history to get a historic view of events. However, this is an optional enhancement.

I think I proposed a combined approach in the "alternatives" section (maybe I deleted it?), but I don't think it justifies the additional complexity.

1vel: For many people a significant portion or all of the rooms they are in _aren't_ encrypted. Sure, adding some text won't fix the core issue, but it sets user expectations correctly.

Many would also expect being able to see what could have notified them on a fresh login.

Since Cinny is also a quite widely used client, it could break users expectations if the button either doesn't work anymore, or disappears entirely.

LaPingvino: @velikopter in my Cinny-fork, I sidestep the endpoint on Continuwuity with basically the same functionality and probably better: https://codeberg.org/lapingvino/cinny/src/commit/15e471bb9a8f6d9fdd92083167e244ea2423da9d/src/app/pages/client/inbox/Inbox.tsx#L21-L78 and https://codeberg.org/lapingvino/cinny/src/commit/15e471bb9a8f6d9fdd92083167e244ea2423da9d/src/app/pages/client/inbox/Unread.tsx#L46-L65 . There is absolutely no need to have the endpoint to provide the functionality.

proposals/4451-notification-endpoint.md
```diff
@@ -0,0 +1,54 @@
+# MSC4451: Deprecate Notifications Endpoint
+
+The `/_matrix/client/v3/notifications` endpoint was added more than a decade ago. In the meantime,
```
jevolk, line 3: Note E2EE dates back to at least 2016 if not earlier; the same decade ago.

proposals/4451-notification-endpoint.md
```diff
@@ -0,0 +1,54 @@
...
+Because the server cannot see the event type or event contents of E2EE messages, it cannot reliably
+decide whether a given message should trigger a notification. It cannot accurately evaluate push rules
+based on event types being visible, or based on mention keywords within the message contents.
+As a result, the server's view of notifications is inherently incomplete. It must either miss all
+encrypted events, or incorrectly assume that all encrypted events cause notifications.
+
+This means that this endpoint necessarily returns incomplete and incorrect results to the client,
+encouraging an incomplete and buggy implementation on the client side.
+
+The endpoint is also relatively unpopular, with few clients using it and few servers implementing it.
```
jevolk, line 16: The endpoint is prominently implemented by the flagship client as the bell icon.

Image

It was requested and had attraction for implementation in Tuwunel: https://github.com/matrix-construct/tuwunel/issues/201

Conduwuit-based servers are free to use this implementation: https://github.com/matrix-construct/tuwunel/commit/48923b36576e194b37c8e7fb4c7721241f485284 rather than opting to remove it from the spec.

JadedBlueEyes: The bell has been disabled by default since [Aug 31, 2023](https://github.com/element-hq/element-web/commit/a5e186b9786c387c999ecc70251ba01059add194)

proposals/4451-notification-endpoint.md
JadedBlueEyes: I believe the point that E2EE was added after this endpoint is clear and correct?

proposals/4451-notification-endpoint.md
timedoutuk: > Not everyone uses E2EE, and the large majority of public rooms aren't encrypted either.

I disagree. Most notifications end up coming from direct messages (one-to-one rooms), which are typically end to end encrypted. Every other notification is typically a reply ping or direct mention, which usually only happens after joining a room, meaning a client can easily continue to incrementally sync and process notifications from new incoming events locally.

> This also encourages less efficient implementations, as, if the client doesn't have all of the messages in every room the user is in (which is a quite common case I'd imagine), it would then need to paginate all of the history of every room the user is in, which doesn't scale very well. (perhaps a client could limit the pagination it does, but that also results in being potentially incomplete)

This is not the case - clients that run in secure contexts already download history in the background, and index incoming events as they happen, in order to provide full-text search for encrypted events (for example, Element Desktop), so those indexes could be re-used for push rule evaluation too. And, considering Element X is also (I believe) going in the same direction, there's precedent that the ecosystem is moving in this direction anyway.

> Having the server process this scales way better as the server has the events after the user joined (which are likely to be the most important to the user) already.

This isn't true, at least not in a way I can see. The server either has to store every notification a user has ever received, or re-evaluate the notifications for each call. The former is expensive on disk, the latter is expensive on compute. Pick your poison. That won't scale any better than client-side evaluation will.

proposals/4451-notification-endpoint.md
timedoutuk: Clients performing client-side notification aggregation (without the endpoint): 
- [Sable (Web)](https://github.com/SableClient/Sable/pull/1372)
- [Gomuks](https://github.com/gomuks/gomuks/commit/61cafc00afa4acc37062a1bd98eef29c9c148a3b)

Clients depending on server-side notification aggregation `/_matrix/client/v3/notifications`:
- Element Web (**labs opt-in only since 2023**)
- NeoChat
- Ement.el (based on source code find & search)
- Cinny

Clients that just don't bother with a notifications list at all (or at least I couldn't find one on my testing account): 
- Chatty
- FluffyChat
- Quaternion
- Watch the Matrix (based on source code find & search)
- Zooid (based on source code find & search)
- Tammy
- Nheko
- chatrix
- Singularity (based on source code find & search)
- Thunderbird
- Commet
- Element X
- kazv (based on source code find & search)
- Extera Next
- SchildiChat Revenge
- Lightning (based on source code find & search)
- SchildiChat Next
- iamb
- Relay (based on source code find & search)
- Mages
- Fractal
- matrix-commander-rs
- Mactrix (based on source code find & search)
- Neoment (based on source code find & search)
- matrix-commander
- Tacit (based on source code find & search)
- Polycule
- WizardChat
- WanChat (assumed)
- Tesseract

Finding clients with bug reports for the notifications endpoint is hard as very few clients even implement anything that uses it anyway. Likewise, while there are a couple that exist, finding MSCs that try to improve on the current endpoint are sparse. [Even Synapse doesn't fully support server-side notifications properly, delegating them to client handling for this reason.](https://github.com/element-hq/synapse/blob/c0357de4eda2f32919d89b4cffa306d68562c022/synapse/handlers/sliding_sync/__init__.py#L1519-L1521)

I'm confident that the clients that DO depend on the endpoint could easily be adapted to perform local evaluation. Sable did this, which proves that it's also possible to do in a browser environment.

proposals/4451-notification-endpoint.md
gamesguru: I think I'm starting to agree with the Continuwuity devs here.

The sever has to store a number, say up to 10,000, of notifications on disk and serve them up arbitrarily depending on client load. (It can also keep min heaps in memory, but in general to guarantee 10,000 notifications per user, this must be done to arbitrary depth or per-user, which are both RAM-demanding).

This is theoretically done based on information the client is already sent thru sync requests (although gaps and edge cases here can occur and arguably pollute the user experience).

E2EE is not a big deal for me, since only mentions or tags of your username/handle will fail server side. Everything else should work. Push rules for encrypted rooms in general, if set to permissively notify on every message (semi-reasonable given encrypted rooms tend to be smaller) will still work and alert will full message contents on devices. Client devices could in theory evaluate mentions/matching substrings via a hybrid approach.

I don't know if deleting the endpoint today is the answer, but we should weigh the pros and cons and the relative difficulty of both approaches. If it greatly simplifies the server-side ecosystem while creating very easily solved client inconveniences, it might make sense to move in that direction. But, conversely, if the `/sync` endpoint is ultimately deemed too unreliable or finicky to allow clients to do these evals and provide a consistent user experience, it may make sense keeping the logic on the server side.

proposals/4451-notification-endpoint.md
kfiven: > The endpoint is also relatively unpopular

The proposal is backed saying it's unpopular but the implementation in flagship client (even behind labs flag) says otherwise.

proposals/4451-notification-endpoint.md
kfiven: E2EE is optional, so the point that we should remove the endpoint because it does not work with that is not very convincing. If the endpoint has issues with E2EE, on-client processing can be proposed for E2EE messages and the endpoint can be kept for non e2ee messages (which are majority of messages from my finding in client where this endpoint is being used).

If Continuwuity devs are having issues in implementing the endpoint, then maybe they should look for better solution for implementation rather than proposing to remove the endpoint.

proposals/4451-notification-endpoint.md
kfiven: Gomuks is running backend server so it's effectively equals to server providing the functionality.

JadedBlueEyes: Gomuks doesn't need to run as a daemon, see gomuks desktop and nexus client. I also don't think this is the correct thread to debate intracacies - if you have a specific concern, please start a new thread.

proposals/4451-notification-endpoint.md
timedoutuk: There are two, maybe three clients that implement it, of which the "flagship" client has deemed it *too buggy* to enable by default.

proposals/4451-notification-endpoint.md
timedoutuk: The better solution is client-side aggregation.

proposals/4451-notification-endpoint.md
timedoutuk: Just updated my original comment after investigating every single listed client at https://matrix.org/ecosystem/clients/. It is painfully apparent that the notifications endpoint is practically unused, and it has known flaws that are easily worked around with local evaluation.

proposals/4451-notification-endpoint.md
```diff
@@ -0,0 +1,54 @@
+# MSC4451: Deprecate Notifications Endpoint
```
wigmarbled, line 1: I tried the Sable web implementation, but I don’t receive any notifications unless I keep my session logged in. This essentially means I have to remain logged in to receive notifications, which is a major security concern, especially since I access my account on public devices.

proposals/4451-notification-endpoint.md
wigmarbled: Server implementation that supports/not support this endpoint (Filtered by maturity stable from https://matrix.org/ecosystem/servers/):
Supports:
* Synapse 
* Tuwunel 
* Synapse Pro

Doesn't support:
* Continuwuity

PR #4452, opened 2026-04-21
MSC4452: Preview URL capabilities API

Half-Shot:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/msc4452-url-preview-cap/proposals/4452-preview-url-capability.md)

Signed-off-by: Will Hunt <signoff@half-shot.uk>

----

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4452#issuecomment-4327157089)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4452#issuecomment-4327158285)


proposals/4452-preview-url-capability.md
turt2live, line 1: Implementation requirements:

* [x] Server (capability)
* [x] Client (respecting capability)
* [x] Client (unaware of capability)

Half-Shot: - Capability https://github.com/element-hq/synapse/pull/19715
 - Element Web (aware) -> https://github.com/element-hq/element-web/pull/33279
 - Element Web (unaware) -> 
   - When the capability is OFF, an unaware version of Element Web continues to work. It doesn't show the user any errors, and the setting can be toggled in settings.
   - When the capability is ON, the /preview_url experience is identical to the spec. The extra capability didn't cause any issues.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,50 @@
+# MSC4452: Preview URL capabilities API
+
+It's possible for server-side implementations to want to disallow users the [GET /_matrix/client/v1/media/preview_url](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediapreview_url) endpoint. This is often done to either reduce excess
```
clokep: This is pretty much making a synapse config flag spec compliant, right?

Half-Shot: Yep! Basically A) Telling the clients if they can perform an action and B) using a proper error.

clokep:
MSCs proposed for Final Comment Period (FCP) should meet the requirements outlined in the [checklist](https://github.com/matrix-org/matrix-spec-proposals/blob/main/MSC_CHECKLIST.md?rgh-link-date=2026-04-27T04%3A22%3A51Z) prior to being accepted into the spec. This checklist is a bit long, but aims to reduce the number of follow-on MSCs after a feature lands.

Spec Core Team (SCT) members, please ensure that all of the following checks
pass before accepting a given Matrix Spec Change (MSC).

MSC authors, feel free to ask in a thread on your PR or in the
[#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org) room for
clarification of any of these points.

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Unstable prefix
  - [x] Dependencies

clokep:
@mscbot fcp merge

turt2live:
@mscbot concern Lacking implementation - see implementation requirements

clokep:
I checked the implementations last night and they seemed good. I've checked the related boxes. The checklist is now complete.

@mscbot resolve Lacking implementation - see implementation requirements

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,50 @@
...
+    "m.preview_url": {
+      "enabled": true
+    }
+  }
+}
+```
+
+If `enabled` is false, any attempt to use `GET /_matrix/client/v1/media/preview_url` MUST be rejected with
+a `403` `M_FORBIDDEN` error.
+
```
uhoreg, line 34: Should probably say something about how clients should treat a missing value (I'd assume either treat it as `"enable": true`, or else treat it as unknown).

Half-Shot: Yep, Element treats it as effectively true (which is safe, as that's what we've been doing for forever anyway..)

Half-Shot: [5b338f6](https://github.com/matrix-org/matrix-spec-proposals/pull/4452/commits/5b338f626c1b5e6189109a21c032c43aab9b98b2)

proposals/4452-preview-url-capability.md
Half-Shot: Both the client and server changes have now merged!

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,54 @@
+# MSC4452: Preview URL capabilities API
+
+It's possible for server-side implementations to want to disallow users the [GET /_matrix/client/v1/media/preview_url](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv1mediapreview_url) endpoint. This is often done to either reduce excess
```
turt2live: `latest` can and does change in ways which make history harder - we prefer static links where possible:

```suggestion
It's possible for server-side implementations to want to disallow users the [GET /_matrix/client/v1/media/preview_url](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediapreview_url) endpoint. This is often done to either reduce excess
```

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,54 @@
...
+load on servers, for security reasons to avoid metadata leakage, or safety to prevent a
+homeserver from querying potentially untrusted services.
+
+This proposal suggests informing clients of this configuration through the standard
+capabilities API.
+
+## Proposal
+
+The [GET /_matrix/client/v3/capabilities](https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3capabilities)
```
turt2live:
```suggestion
The [GET /_matrix/client/v3/capabilities](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3capabilities)
```

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,54 @@
...
+Servers MAY elect to allow some users access to this endpoint, so the response may be different
+depending on the authenticated user (e.g. disabled for guest users).
+
+## Potential issues
+
+None.
+
+## Alternatives
+
+None.
```
turt2live: This should describe why the client always being met with an error is worse/bad than having a capability.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,54 @@
...
+a `403` `M_FORBIDDEN` error.
+
+If `enabled` is `true` or not defined, then the client can make queries as normal. This means that the existing
+behaviour is preserved so that servers who do not support this capability will not prevent clients from
+requesting previews.
+
+Servers MAY elect to allow some users access to this endpoint, so the response may be different
+depending on the authenticated user (e.g. disabled for guest users).
+
+## Potential issues
```
turt2live, line 42: This isn't necessarily something for this proposal to acknowledge or worry about, but as we continue to add more capabilities we risk the spec being *too* elective, which adds complexity to primarily clients. Currently, the spec is intended to be monolithic so clients don't need to version-check everything they want to use (or in some cases, can't operate without). 

This proposal introduces a capability for something which is already reasonably optional: a server can be easily configured to deny URL previews entirely, always producing an error. We can save an HTTP hit by telling the client that it won't work.

We *shouldn't* rely on capabilities too much, though.

turt2live (approved):
generally this seems fine - see threads for detailed comments.

I would consider the Alternatives section being populated as a concern which blocks FCP from finishing, but not from starting. The alternative is unlikely to sway opinions, but should be documented.

proposals/4452-preview-url-capability.md
Half-Shot: Done in 6a92b424

proposals/4452-preview-url-capability.md
Half-Shot: Have added some words. I think in this case the capability is well justified as a way for the client to reduce metadata leakage and reduce homeserver traffic, and using existing capability primitives means clients shouldn't have to build in a ton of new behaviors (as demonstrated by the client impl).

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
...
+{
+  "capabilities": {
+    "m.preview_url": {
+      "enabled": true
+    }
+  }
+}
+```
+
+If `enabled` is `false`, any attempt to use `GET /_matrix/client/v1/media/preview_url` MUST be rejected with
```
dbkr: Why a 'MUST' here? If a server wants to claim url previews are not allowed but actually allow them, who are we to judge?

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
...
+behaviour is preserved so that servers who do not support this capability will not prevent clients from
+requesting previews.
+
+Servers MAY elect to allow some users access to this endpoint, so the response may be different
+depending on the authenticated user (e.g. disabled for guest users).
+
+## Potential issues
+
+This increases complexity for clients who offer URL previews as a feature. They will need to check
+the capabilities offered by the homeserver before they can start making requests. This has been mitigated
```
dbkr: Not really, surely they are still free to give it a go? This is more for information purposes.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
...
+by reusing the capabilities endpoint. For the reasons outlined in Alternatives, it's an acceptable
+tradeoff to include a capability for this feature.
+
+## Alternatives
+
+In practice most implementations of URL previews either allow the server admin to disable/enable it, or  do
```
dbkr:
```suggestion
In practice most implementations of URL previews either allow the server admin to disable/enable it, or do
```

proposals/4452-preview-url-capability.md
Half-Shot: Depends. If I were developing a client and wanted to reduce metadata leakage, *i'd* check before making requests. You could argue this isn't absolutely necessary, but I think it'd be *weird* to make requests before checking caps given it's cheap.

proposals/4452-preview-url-capability.md
Half-Shot: Sure they could, but it would feel non-compliant to me and the spec is meant to set the expectations for clients. In practice, this is all unenforceable words and implementations can ignore them as they choose, but we should strive to write the expected use case in here?

I think unless there is a justified use case for "If a server wants to claim url previews are not allowed but actually allow them" then this should remain a MUST, imo.

proposals/4452-preview-url-capability.md
Susurrus: Should there be a SHOULD here for a best practice?

proposals/4452-preview-url-capability.md
richvdh: I'm with Dave here. This feels like unnecessarily binding the server implementation. How about:

> If `enabled` is `false`, clients SHOULD NOT use `GET /_matrix/client/v1/media/preview_url`, to avoid unnecessary traffic and information leakage.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
+# MSC4452: Preview URL capabilities API
+
+It's possible for server-side implementations to want to disallow users the [GET /_matrix/client/v1/media/preview_url](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1mediapreview_url) endpoint. This is often done to either reduce excess
+load on servers, for security reasons to avoid metadata leakage, or safety to prevent a
+homeserver from querying potentially untrusted services.
+
+This proposal suggests informing clients of this configuration through the standard
+capabilities API.
```
richvdh: This introduction doesn't really explain why a capability is a useful thing. You can kindof infer it by reading the "Alternatives", but it should be clarified here, imho.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
...
+
+In practice most implementations of URL previews either allow the server admin to disable/enable it, or do
+not support it. With that in mind, an alternative could be to just standardise the error code without the capability.
+However, this is a problem for three reasons.
+
+1. Clients will continue to make requests to the homeserver even if the feature is disabled, increasing the amount of traffic
+   the server needs to handle.
+2. Clients will leak URLs to the homeserver (which may be within encrypted messages). While this is an acceptable trade-off
+   if the feature works, it's pointless if the requests are being failed.
+3. Clients often allow users to choose whether they want the feature on. It's useful to show the option as disabled
```
richvdh, line 56: > it's pointless if the requests are being failed.

meh, it's pointless, but if you care about the security concern, you probably shouldn't turn on the feature in your client anyway.

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,68 @@
...
+   if the homeserver doesn't support it.
```
richvdh, line 57: I am not entirely convinced that the benefit is worth the added complexity. I'm prepared to be disagreed with here though, but I'm wondering if anyone actually really cares about these things.

Do we have people that deliberately turn off preview support on their HS and are concerned about the bandwidth? Do we have client product people who want to grey out the "enable preview" option if it's unsupported on the server?

richvdh (commented):
Not sure I buy it, but I'm happy to be convinced.

proposals/4452-preview-url-capability.md
timedoutuk: Not sure about the bandwidth perspective specifically, but Continuwuity ships with URL previews off by default, and we've had many users ask about the security and privacy ramifications of enabling the feature, whom then often continue leaving it disabled. Concerns have been raised about the bandwidth usage before, particularly for more open/public servers, but that usually seems to be a secondary concern to privacy+security.

proposals/4452-preview-url-capability.md
richvdh: Element Web disables URL previews in encrypted rooms by default. I haven't tested EX, but I'd expect it to go the same way. So you definitely have to *opt in* to send your encrypted URLs to the server.

I'm really having a tough time justifying that we need to add a load of complexity to the protocol in order to mitigate this.

proposals/4452-preview-url-capability.md
richvdh: [Synapse also ships with URL previews disabled, fwiw. When I say "deliberately turn off", I mean "make the conscious decision not to enable it, as opposed to accepting the defaults and not thinking about it"]

> Concerns have been raised about the bandwidth usage before, particularly for more open/public servers

IME, people often get unnecessarily concerned about bandwidth usage because something "feels" inefficient, but have no data to back this up. It's a form of premature optimisation. Is the bandwidth usage of (failing) URL previews *actually* significant compared to other HS traffic? Is it *actually* causing performance or cost problems? I'd be a lot more persuaded if someone measured some bandwidth figures on a HS or client.

> that usually seems to be a secondary concern to privacy+security.

Let's discuss the privacy concern, or lack thereof, in the [other thread](https://github.com/matrix-org/matrix-spec-proposals/pull/4452#discussion_r3343401813)

richvdh:
For clarity: I'm not going to block this landing if 8 other SCT members agree it is needed (hence not raising a formal concern), but as of right now, I'm unconvinced and am not ticking it.

proposals/4452-preview-url-capability.md
Half-Shot: Addressing the original post, the "added complexity" is a boolean field in an existing endpoint that most clients are probably already calling (we discussed this [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4452#discussion_r3269002546)) and the server implementation for Synapse at least was pretty much reflecting a configuration option in the endpoint. I think we should be clear here the added complexity is not a huge burden.

On the other point:

>Do we have client product people who want to grey out the "enable preview" option if it's unsupported on the server?

This is the main driver for the feature. We want to reflect the state of the feature in the clientside settings menus correctly because users have tried to enable the setting, found it didn't work, and cannot understand why. IMO this feels pretty much the same use case as `m.profile_fields` (and previously `m.set_avatar_url`) in that we are providing a capability to largely improve the client UX.

The other benefits listed are largely secondary, but no less worth mentioning.

proposals/4452-preview-url-capability.md
Half-Shot: There are cases where clients would like to have features on by default, and Element Web chooses to do this for unencrypted URLs. This is by no means the primary benefit of the feature, but it's not an entirely pointless consideration.

I also really want to understand where you think this proposal adds a "load of complexity". I am not seeing that at all.

proposals/4452-preview-url-capability.md
richvdh: > Addressing the original post, the "added complexity" is a boolean field in an existing endpoint that most clients are probably already calling (we discussed this [here](https://github.com/matrix-org/matrix-spec-proposals/pull/4452#discussion_r3269002546)) and the server implementation for Synapse at least was pretty much reflecting a configuration option in the endpoint. I think we should be clear here the added complexity is not a huge burden.

Sadly your link doesn't work, but you're right that it's not a significant amount of complexity and not a huge burden. I was just a bit mystified about what problem we were actually trying to solve here, as most of the purported benefits seem to be negligible. However ...

> > Do we have client product people who want to grey out the "enable preview" option if it's unsupported on the server?
> 
> This is the main driver for the feature. We want to reflect the state of the feature in the clientside settings menus correctly because users have tried to enable the setting, found it didn't work, and cannot understand why.

Right, fair enough. This does seem like a tangible benefit. Could you make it a bit more pronounced in the MSC? Putting it at number 3 in the list of problems with the "Alternatives" made it feel like an afterthought rather than the whole driver of the feature.

proposals/4452-preview-url-capability.md
Susurrus: Note that capabilities already need to be checked for several things (editing use avatar, etc.). I was just working on that for Fractal, and we cache the capabilities for 24h I think? So in practice, this checking is almost free.

proposals/4452-preview-url-capability.md
Half-Shot: Fair feedback, I suppose I rather jumbled a bunch of benefits rather than leading with the main one. Will do so :+1:

Half-Shot: I hope that f1adda9ae16f42bee1ceb9c763abb6e80ffc44d2 gives a stronger rationale

proposals/4452-preview-url-capability.md
Half-Shot: Hopefully done so in f1adda9ae16f42bee1ceb9c763abb6e80ffc44d2

proposals/4452-preview-url-capability.md
Half-Shot: I'm won over, updated in c1f99a7a

proposals/4452-preview-url-capability.md
Half-Shot: I'm not sure if there is something actionable here?

Susurrus: > They will need to check
the capabilities offered by the homeserver before they can start making requests. 

Would making this a formal recommendation using SHOULD be appropriate here? I think that's the key point here.

Half-Shot: Ah, instead of the "need"? I can change the language a bit here but I don't think this needs a MUST/SHOULD.

It's ultimately up to the client how they choose to ingest the capability and it's not for the spec to mandate that imo

proposals/4452-preview-url-capability.md
```diff
@@ -22,8 +30,8 @@ will include a new capability key, `m.preview_url` containing a single `enabled`
 }
 ```
 
-If `enabled` is `false`, any attempt to use `GET /_matrix/client/v1/media/preview_url` MUST be rejected with
-a `403` `M_FORBIDDEN` error.
+If `enabled` is `false`,  `GET /_matrix/client/v1/media/preview_url` SHOULD reject with
+a `403` `M_FORBIDDEN` error to avoid unnecessary traffic and information leakage.
```
richvdh: How does a 403 avoid unnecessary traffic?

proposals/4452-preview-url-capability.md
richvdh: > Not really, surely they are still free to give it a go? This is more for information purposes.

+1.

We can surely remove this whole paragraph from potential issues. Clients don't *have* to check the caps. They don't even SHOULD check the caps. They can if they find it useful, to show in the settings, or to reduce traffic, but that's up to them.

I don't think there is any increase in complexity for clients over what's there today.

(If anything, there is an increase in complexity for servers, which now must/should support this cap. Probably sufficiently marginal not to be worth mentioning.

proposals/4452-preview-url-capability.md
Half-Shot: :facepalm: . It doesn't really does it. I'll just remove that bit.

proposals/4452-preview-url-capability.md
```diff
@@ -1,15 +1,22 @@
...
+for not supplying a preview they have asked for. Hence, having proper access to the
+state is useful.
 
 This proposal suggests informing clients of this configuration through the standard
 capabilities API.
 
 ## Proposal
 
-The [GET /_matrix/client/v3/capabilities](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3capabilities)
+The [GET /\_matrix/client/v3/capabilities](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3capabilities)
```
richvdh:
```suggestion
The [GET /\_matrix/client/v3/capabilities](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3capabilities) endpoint
```

proposals/4452-preview-url-capability.md
richvdh: > Element Web chooses to do this for unencrypted URLs.

Unencrypted URLs are visible to the server admin anyway, though. I suppose there's a difference between a URL appearing unencrypted in the database and said URL appearing in the http access logs, but if that's the difference you're basing your argument on, it would be better to make it explicit.

proposals/4452-preview-url-capability.md
richvdh: That change certainly helps clarify, but I still think this list of problems with the alternative is very weak and fails to highlight the main problem.

You've done enough to convince me that this MSC is worthwhile, but MSCs have an important job as historical documents that we can look back at in future, setting out our reasoning for a change. Could you have another go at this list, making sure you highlight the main driver, and maybe mentioning that the other points are just bonuses?

proposals/4452-preview-url-capability.md
```diff
@@ -0,0 +1,77 @@
...
+  "capabilities": {
+    "m.preview_url": {
+      "enabled": true
+    }
+  }
+}
+```
+
+If `enabled` is `false`, `GET /_matrix/client/v1/media/preview_url` SHOULD reject with
+a `403` `M_FORBIDDEN` error.
```
anoadragon453, line 33: Note that the TI-Messenger specification [currently mandates](https://gemspec.gematik.de/docs/gemSpec/gemSpec_TI-M_Basis/gemSpec_TI-M_Basis_V1.2.0/#A_26344) that a 404 be returned from this endpoint.

This is more something for the TI-M spec authors to consider than yourself, but just dropping it here as I'm lacking a better spot.

Half-Shot: I suppose you could argue it either way. I felt using a status code for *explicitly* "this is the correct endpoint but you are not allowed" was more appropriate than a not found.

I don't know if TI-M would resolve to adopt this spec if it gets merged?

proposals/4452-preview-url-capability.md
dbkr: Yep, just to note that rich has captured my drift here. I'll put a concern on this just to note where I am with this proposal.

dbkr:
@mscbot concern complexity increase for clients

proposals/4452-preview-url-capability.md
Half-Shot: Okay, I went over this in my head and came up short of anything useful to put in Potential Issues. I don't think we need to refer to "previously though to be potential issues" given the rich comment history here, so leaving this as is.

@dbkr is this what you had in mind?

dbkr: works for me!

dbkr:
@mscbot resolve complexity increase for clients

PR #4454, opened 2026-04-22
MSC4454: Deprecating Spoiler Fallback In Media Repository

dozro:
[Rendered](https://github.com/dozro/matrix-spec-proposals/blob/deprecate-spoiler-plaintext-fallback/proposals/4454-deprecating-spoiler-fallback-in-media-repository.md)

**Implementations**:

- not relevant, doesn't have implementation requirements (see [comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4454#discussion_r3132022760))

Signed-off-by: Rye <git@itsrye.dev>


proposals/4454-deprecating-spoiler-fallback-in-media-repository.md
```diff
@@ -0,0 +1,142 @@
...
+
+This MSC proposes removing specification language describing unused client behaviour for spoiler
+fallbacks, which is ambiguous and could be interpreted as encouraging plaintext storage of
+spoilered content in End-to-end encryption contexts.
+
+This proposal prioritizes preventing unintended disclosure of spoilered content over preserving plaintext fidelity.
+
+## Proposal
+
+The spec now reads that a client should provide a plaintext (`body`) of
```
clokep: Please link to the spec!

proposals/4454-deprecating-spoiler-fallback-in-media-repository.md
tulir, line 1: Implementation requirements: None, the latest version of the proposal only removes parts of the spec

proposals/4454-deprecating-spoiler-fallback-in-media-repository.md
```diff
@@ -0,0 +1,142 @@
...
+  "body": "Alice [Spoiler] in the movie.",
+  "formatted_body": "Alice <span data-mx-spoiler>lived happily ever after</span> in the movie."
+}
+```
+
+Clients generating messages with spoilers:
+
+- MUST NOT include spoilered content in the `body` field
+- MUST replace spoilered sections with a neutral placeholder (e.g. `[Spoiler]`)
+- MUST encode the full spoilered content only in `formatted_body`
```
tulir: The spec doesn't currently require anything like this, all the fallback stuff is `SHOULD`. If it's intended, might be worth calling out that change more, especially since that's the only thing that requires an implementation here

dozro: So it should rather be a `SHOULD` instead of a `MUST`?

(changed it to `SHOULD`)

proposals/4454-deprecating-spoiler-fallback-in-media-repository.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Plaintext-only clients will not receive spoilered content
+
+Because notifications and message previews are often derived from the `body` field, they will contain less information for spoilered
+messages under this proposal.
+
+For example, a notification may display `Alice [Spoiler] in the movie.` rather than including the spoilered text itself.
+
+While this reduces the usefulness of previews, it avoids leaking spoilered content in contexts where it may be displayed without
+explicit user interaction (such as push notifications or lockscreen previews).
```
JadedBlueEyes: Clients like gomuks already handle this by replacing the whole message in plaintext contexts, for example: 

Image

In my opinon this particular trafeoff isn't worth the potential loss of information.

dozro: didn't know that. thanks for the info.

will adjust the MSC as soon as i come to it :3

dozro: probably to something like `[Spoiler](text goes here)` so that the plain text is still usable while still containing information about it being a spoiler.

dozro: updated the MSC accordingly in 7e0c27664f5d56ee2e32f97da8e758ac5c2fb217

proposals/4454-deprecating-spoiler-fallback-in-media-repository.md
```diff
@@ -0,0 +1,101 @@
...
+
+This behavior is not implemented by known clients and represents unnecessary
+specification complexity. It is not appropriate to keep this text in the spec,
+because a client implementing it could leak message text when sending spoilered
+content in an end-to-end encrypted (E2EE) room. The current spec does not
+specify how spoilers should be handled in E2EE rooms and could be read as
+suggesting uploading spoilered text unencrypted to the media repository.
+
+This proposal removes those recommendations on how clients should set the plain
+text `body` for spoilered messages.
```
tulir: The proposal should explicitly say what the new behavior is (it's now an implementation detail). Something along the lines of

> This proposal removes the recommended fallback for spoilers in `body`. The plaintext representation of spoilers is left as an implementation detail.

Maybe also mention something about how clients shouldn't use `body` directly

> Spoiler-aware clients SHOULD NOT display `body` anywhere when there are spoilers in `formatted_body`. For contexts like room list previews and notifications where a plain text representation is needed, clients SHOULD either generate plain text from the HTML locally, or simply display a placeholder message when any spoilers are detected.

JadedBlueEyes (approved):

PR #4455, opened 2026-04-25
MSC4455: Catch-all property for spaces

SpiritCroc:
[Rendered](https://github.com/SpiritCroc/matrix-spec-proposals/blob/space_catchall/proposals/4455-space-catch-all.md)

### DIFF
--- a/proposals/4455-space-catch-all.md
+++ b/proposals/4455-space-catch-all.md
@@ -1,4 +1,4 @@
-# MSC0000: Space catch-all
+# MSC4455: Space catch-all
 
 This MSC introduces a new state event for `m.space` rooms that allows spaces to instruct clients to
 include "space orphans" (rooms not part of any space that the user knows of).
--- a/proposals/4455-space-catch-all.md
+++ b/proposals/4455-space-catch-all.md
@@ -10,8 +10,8 @@ of spaces that you use for day-to-day communication, ignoring a majority of othe
 joined. However, sometimes you may get invited to a new chat and not immediately remember
 to put it into your day-to-day spaces. A catch-all space property would allow you to upgrade
 your space to include any room that you haven't categorized into an appropriate space yet, and thus
-ensure you don't miss out. In this scenario, you would then remove rooms from the catch-all
-space by ensuring they are assigned to another space, e.g. a "verbose" space.
+ensure you don't miss out. In this scenario, in order to remove rooms from a catch-all space,
+users could then assign the respective rooms to another space, e.g. a "verbose" space.
 
 Accordingly, space rooms can include a `m.space.catch_all` state event to instruct clients to perform
 such catch-all functionality. For the time being, server implementations are not required to apply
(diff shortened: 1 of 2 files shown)


SpiritCroc:
Initial client implementation in SchildiChat Next & Revenge for filtering:
- ruma: https://github.com/SchildiChat/ruma/commit/2b36edf05634795c1771c97a4097f262392c5e4f
- matrix-rust-sdk: https://github.com/SchildiChat/matrix-rust-sdk/commit/5bc2f645797a2fd4cd6429b98227e07921f6b816
- SchildiChat Revenge: https://github.com/SchildiChat/schildi-revenge/commit/cce2a015f3f2f95fc73c450b6d09bdfd61172aad
- SchildiChat Next: https://github.com/SchildiChat/schildichat-android-next/commit/8264c71192c1f6a76185f6a36b8f4e986df592ac

UI Setting:
- SchildiChat Revenge: https://github.com/SchildiChat/schildi-revenge/commit/e74c88b9c8fffba1b80a742b408e5e01936f3f72

proposals/4455-space-catch-all.md
turt2live, line 1: Implementation requirements:

* [ ] Client
* [ ] Product review

proposals/4455-space-catch-all.md
```diff
@@ -0,0 +1,69 @@
...
+
+This MSC introduces a new state event for `m.space` rooms that allows spaces to instruct clients to
+include "space orphans" (rooms not part of any space that the user knows of).
+
+## Proposal
+
+Personal spaces are a great way to organize your own inbox. Accordingly, you may want to have a set
+of spaces that you use for day-to-day communication, ignoring a majority of other chats you may have
+joined. However, sometimes you may get invited to a new chat and not immediately remember
+to put it into your day-to-day spaces. A catch-all space property would allow you to upgrade
```
ashkitten, line 11: Since this is meant for personal use, I think it would make more sense to use account data than a state event that is synced to all space participants and requires permission to send.

SpiritCroc: Fair suggestion, I guess I can put that into the alternatives section.

Why I think state events are more suitable here:
Conceptually I see it as a property of a space, not for your personal view on a space. I don't think it makes sense to allow users to enable this for public spaces where they don't have state event permission (you can still put the public space into your private spaces for organization anyway to achieve the same, provided a client that recurses spaces when filtering the room list). Furthermore just because the feature is meant for personal spaces, doesn't mean you may not want to share it across user accounts - imagine a multi-account Matrix client that merges all your accounts' space lists into one (like [SchildiChat Revenge](https://github.com/schildiChat/schildi-revenge)), then if you put some of your accounts into the same personal space, you probably want it to behave the same, without having to configure it for each account separately.

ashkitten: What if a trollish space admin sends this state event in a public space?

SpiritCroc: Clients that are concerned with that may chose to ignore catch-all properties set on public spaces.

PR #4456, opened 2026-04-28
MSC4456: Harms taxonomy

turt2live:
> [!WARNING]
> **Content Warning**: This proposal discuses and identifies harmful content, but does not attempt
> to describe the harm posed in detail. This includes identifiers for child safety, sexual abuse,
> self-harm, and other types of harm a user may encounter on the open internet.

----

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/harms-appendix/proposals/4456-safety-harms-appendix.md)

*This proposal was split out from [MSC4387](https://github.com/matrix-org/matrix-spec-proposals/pull/4387).*

### DIFF
--- a/proposals/4456-safety-harms-appendix.md
+++ b/proposals/4456-safety-harms-appendix.md
@@ -1,0 +1,145 @@
+# MSC4456: Harms taxonomy
+
+> [!WARNING]
+> **Content Warning**: This proposal discuses and identifies harmful content, but does not attempt
+> to describe the harm posed in detail. This includes identifiers for child safety, sexual abuse,
+> self-harm, and other types of harm a user may encounter on the open internet.
+
+*This MSC is part of "Reporting v2" - a project led by the Foundation’s T&S team to improve communication
+and effectiveness of reports on Matrix.*
+
+When a user reports something, they ideally are able to express their opinion for what kind of harm
+they believe is caused by that something. Safety teams use this information to guide their initial
+investigation, and can reclassify a report as needed to better match the harm actually caused. This
+is in contrast to an (often blank) free-form report reason where the safety team needs to guess at
+the harm caused.
+
+Identified harms are also helpful when events or searches are rejected due to backend safety tooling,
+like in [MSC4387](https://github.com/matrix-org/matrix-spec-proposals/pull/4387).
+
+This proposal introduces a new [appendix](https://spec.matrix.org/v1.18/appendices/) to list out
+harms common to safety legislation across the world. Other proposals are expected to actually use
+this taxonomy - this proposal simply introduces them as a dependency for other MSCs.
+
+
+## Proposal
+
+The following standardized harm identifiers are expected to be used by Matrix safety tooling and features.
+For example, when reporting something, the user can express their opinion of what kind of harm is caused
+by the thing they're reporting. Similarly, a (policy) server might reject an event due to a common
+harm.
+
+The identifiers chosen represent the similarities between various safety legislations (UK, Australia,
+Canada, EU, etc) and what other services offer to their users in reporting flows. The identifiers use
+the [Common Namespaced Identifier Grammar](https://spec.matrix.org/v1.18/appendices/#common-namespaced-identifier-grammar),
+and therefore allow custom harms to be represented. When a proposal uses harm identifiers, it SHOULD
+ensure that custom identifiers are accompanied by a specified identifier for added clarity/compatibility.
+
+The harms, their categories, and suggested names are:
+
+**Spam**
+
+* `m.spam` - General/Other
+* `m.spam.fraud` - Fraud/Phishing
+* `m.spam.impersonation` - Impersonation
+* `m.spam.election_interference` - Election Interference
+* `m.spam.flooding` - Flooding
+
+**Adult Content & Safety**
+
+* `m.adult` - General/Other
+* `m.adult.sexual_abuse` - Sexual Abuse
+* `m.adult.ncii` - Non-Consensual Intimate Imagery
+  * Recognizing that these links deal with sexual abuse topics, more information about NCII can be
+    found at [StopNCII](https://stopncii.org/), [INHOPE](https://inhope.org/EN/articles/what-is-ncii),
+    and [Meta's Safety Center](https://www.meta.com/safety/topics/bullying-harassment/ncii/).
+* `m.adult.deepfake` - Deepfake
+* `m.adult.animal_sexual_abuse` - Animal Sexual Abuse
+* `m.adult.sexual_violence` - Sexual Violence
+
+**Harassment**
+
+* `m.harassment` - General/Other
+* `m.harassment.trolling` - Trolling
+* `m.harassment.targeted` - Targeted
+* `m.harassment.hate` - Hate
+* `m.harassment.doxxing` - Doxxing/Personal Information
+
+**Violence**
+
+* `m.violence` - General/Other
+* `m.violence.animal_welfare` - Animal Welfare
+* `m.violence.threats` - Threatening/Threats
+* `m.violence.graphic` - Graphic/Gore
+* `m.violence.glorification` - Glorification/Promotion
+* `m.violence.extremist` - Extremism
+* `m.violence.human_trafficking` - Human Trafficking
+* `m.violence.domestic` - Domestic/Intimate Partner
+
+**Child Safety**
+
+* `m.child_safety` - General/Other
+* `m.child_safety.csam` - Child Sexual Abuse Material (CSAM)
+* `m.child_safety.grooming` - Grooming
+* `m.child_safety.privacy_violation` - Privacy
+* `m.child_safety.harassment` - Harassment
+
+**Dangers**
+
+* `m.danger` - General/Other
+* `m.danger.self_harm` - Self Harm
+* `m.danger.eating_disorder` - Eating Disorder
+* `m.danger.challenges` - Challenges, including Social Media Challenges
+* `m.danger.substance_abuse` - Substance Abuse
+
+**Terms of Service**
+
+* `m.tos` - General/Other
+* `m.tos.hacking` - Hacking/Computer Misuse
+* `m.tos.prohibited` - Prohibited Items (Drugs, Weapons, etc)
+* `m.tos.ban_evasion` - Ban Evasion
+
+
+## Implementation considerations
+
+* A reporting dialog using these harms might have a two-tier dropdown: one for the category (spam,
+  harassment, etc) and another for the specific harm caused (defaulting to "General/Other").
+
+* Clients should generally aim to keep definitions/titles of the harms brief to be as broadly applicable
+  as possible.
+
+* Future MSCs are encouraged to explore using the harms list *safely*. [MSC4204](https://github.com/matrix-org/matrix-spec-proposals/pull/4204)
+  and [MSC4205](https://github.com/matrix-org/matrix-spec-proposals/pull/4205) both discuss challenges
+  related to classifying content against users.
+
+
+## Potential issues
+
+* The specified harms list might not be extensive enough to encompass all possible harms in the online
+  world. It's expected that custom identifiers for commonly used harms will become MSCs to expand the
+  list as needed. Other MSCs are expected to describe ways of expressing custom namespaced harms that
+  are usable in different environments.
+
+
+## Alternatives
+
+No significant alternatives. Having a standard taxonomy of harms is important for safety teams to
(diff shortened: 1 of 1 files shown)


proposals/4456-safety-harms-appendix.md
turt2live, line 1: Implementation requirements:

* [ ] 2+ MSCs in different areas using this. For example, search redirection and reporting.

proposals/4456-safety-harms-appendix.md
```diff
@@ -0,0 +1,145 @@
...
+by the thing they're reporting. Similarly, a (policy) server might reject an event due to a common
+harm.
+
+The identifiers chosen represent the similarities between various safety legislations (UK, Australia,
+Canada, EU, etc) and what other services offer to their users in reporting flows. The identifiers use
+the [Common Namespaced Identifier Grammar](https://spec.matrix.org/v1.18/appendices/#common-namespaced-identifier-grammar),
+and therefore allow custom harms to be represented. When a proposal uses harm identifiers, it SHOULD
+ensure that custom identifiers are accompanied by a specified identifier for added clarity/compatibility.
+
+The harms, their categories, and suggested names are:
```
thetayloredman, line 58: Is there anticipated to be a generic `m.other` harm that can be used when no other explicit category from this list applies? Free-form text can still be very useful, especially for reporting specific harms.

proposals/4456-safety-harms-appendix.md
```diff
@@ -0,0 +1,145 @@
...
+
+**Spam**
+
+* `m.spam` - General/Other
+* `m.spam.fraud` - Fraud/Phishing
+* `m.spam.impersonation` - Impersonation
+* `m.spam.election_interference` - Election Interference
+* `m.spam.flooding` - Flooding
```
thetayloredman, line 66: It may be worthwhile to have a separate `m.misinformation` category, especially to include other forms of synthetic media/deepfakes besides `m.adult.deepfake` (not all deepfakes are inherently sexual, so there could be e.g. `m.misinformation.deepfake`). "Election interference" feels like it wouldn't always be a subcategory of spam.

Having them be independent also means you can classify `m.misinformation.fraud` *alongside* `m.spam` where they come in a list.

Additionally, if the UX was designed to align with these categories, it wouldn't make sense for a user to view these under "spam," IMHO.

proposals/4456-safety-harms-appendix.md
turt2live: Oops, yes, we do want `m.other` for "idk, it's just bad".

proposals/4456-safety-harms-appendix.md
turt2live: Technically met by the following two MSCs, but I'd like to see each further along in the spec process before considering them true implementations of this MSC:

* https://github.com/matrix-org/matrix-spec-proposals/pull/4387
* https://github.com/matrix-org/matrix-spec-proposals/pull/4457

proposals/4456-safety-harms-appendix.md
```diff
@@ -0,0 +1,145 @@
+# MSC4456: Harms taxonomy
```
ara4n, line 1: I am very dubious that we should be baking a taxonomy like this into the Matrix spec, because:

 1. The spec is already huge
 2. Harms can be very subjective and will encourage bikeshedding or bloat. e.g. where is [Lese-majesty](https://en.wikipedia.org/wiki/L%C3%A8se-majest%C3%A9#Current_laws) on the list?
 3. In practice we'll always need an 'other' fallback with a natural language explanation anyway - why not use natural language all along?
 4. Why do we care about semantic codes here at all?
 5. I suspect that we're going to see more and more LLM-based moderation functionality in future, which will be quite happy to process natural language reasons rather than trying to create a set of reason enumerations

At the least, i'd expect the reasons to sit in an external registry somewhere to avoid bloating the spec.

thetayloredman: To be fair, many online reporting platforms have a flow for specifically selecting harms information which I believe to be the point of the MSC:
image

It feels reasonable to want to put a list of some common harms in to make the interface better and potentially aid in tooling *without* the intrinsic requirement for LLMs :D

I do agree that it might be better outside of the spec, but the question for me is *where* does this get defined, because it feels very necessary.

jimmackenzie: > 1. The spec is already huge

Building a better way to communicate necessarily involves a bunch of detail! 
> 2. Harms can be very subjective and will encourage bikeshedding or bloat.

That is a risk. Defining a baseline taxonomy rather than a comprehensive one may help to mitigate that risk.
> 3. In practice we'll always need an 'other' fallback with a natural language explanation anyway - why not use natural language all along?
> 4. Why do we care about semantic codes here at all?

With semantic codes, clients can more easily build better reporting flows, offering tailored advice based on the type of harm the user has experienced (e.g. direction to helplines, law enforcement, how to keep themselves safe). Servers and communities can use the codes to communicate why enforcement action was taken against a user or piece of content (a requirement in many safety laws). Safety teams can use user-provided codes to triage reports more effectively, both with human teams, and by routing to the most cost/time-effective automated flows. When all servers in a Matrix federation share a common taxonomy of harms, it simplifies sharing details of those harms over federation.
> 5. I suspect that we're going to see more and more LLM-based moderation functionality in future, which will be quite happy to process natural language reasons rather than trying to create a set of reason enumerations

Using semantic reasons enables the use of more cost-effective & faster single purpose models rather than slower, more expensive general models, and enables routing to appropriate humans for review.
> At the least, i'd expect the reasons to sit in an external registry somewhere to avoid bloating the spec.

An alternative: reference an external standard, as we do with RFCs elsewhere in the spec. Unfortunately, there doesn't appear to be a suitable standard that provides this taxonomy at present, but this could be something to explore and then replace this proposal down the line. The DTSP framework (ISO/IEC 25389) is an example of nascent work in safety standardisation, that doesn't do what we need here. The spec could also offer appendices for this type of content, if there are concerns? I think the spec should contain appropriate guidance to building safe servers and clients, so I'd be comfortable with us including it in the body of the spec.

proposals/4456-safety-harms-appendix.md
turt2live: Looking at Bluesky (where the list is inspired in part from), they label "spam" as "Misleading - spam or other inauthentic behaviour or deception". We might want to adopt similar labeling for the "Spam" category we have here.

They also consider deepfakes to be primarily adult content. I suspect that if a user was reporting a deepfake that wasn't easily classified as adult content then they'd use "impersonation" or "other misleading content" (to use the Bluesky terms).

proposals/4456-safety-harms-appendix.md
turt2live: 50d17816c020232829654306e326ab1bfc375cd5

proposals/4456-safety-harms-appendix.md
turt2live: The closest I can find for an existing reference are AT Proto's [`com.atproto.moderation.defs`](https://atproto.blue/en/latest/atproto/atproto_client.models.com.atproto.moderation.defs.html) and [`tools.ozone.report.defs`](https://atproto.blue/en/latest/atproto/atproto_client.models.tools.ozone.report.defs.html) models. Obviously, these definitions are highly targeted at AT Proto's use cases, but the parallels in this MSC should be fairly evident as well :) 

We may benefit from just copying AT Proto's definitions directly, or working with them to create an external standard that works for both of us. This MSC currently suggests we do something similar to what AT Proto did: create an appendix/definition that exists within their world and refer to it as needed.

The next closest I can find for an existing reference is the European Commission's Transparency Database API which describes content in two ways: a [Category](https://transparency.dsa.ec.europa.eu/page/api-documentation#category-category) and a [Category Specification](https://transparency.dsa.ec.europa.eu/page/api-documentation#category-specification-category-specification). By nature of it being backed by the Digital Services Act (DSA), it's *highly* focused on that particular regulatory environment - it does not easily apply to other environments such as the UK, US, Australia, or Canada (despite these places copying most of each other's work in law creation). 

Related work is from the World Economic Forum (WEF) which attempts to describe harms in ways that users can understand, but is hardly a "harm identifier" list. Their report can be found [here](https://www.weforum.org/publications/toolkit-for-digital-safety-design-interventions-and-innovations-typology-of-online-harms/).

The Trust & Safety Professional Association (TSPA) [attempts to list the types of abuse](https://www.tspa.org/curriculum/ts-fundamentals/policy/abuse-types/), but also doesn't define machine-friendly identifiers for those abuse types. It may be possible to ask them to create a machine-friendly taxonomy for their list, though I expect it'll be too broad for our purposes in Matrix.

[IFTAS](https://about.iftas.org/) is primarily used by ActivityPub, but is not a standards organization. They do however provide definitions for 3 types of harmful content (and how to deal with it): by [actor](https://about.iftas.org/wiki_cats/actors/), [behaviour](https://about.iftas.org/wiki_cats/behaviours/), or [content](https://about.iftas.org/wiki_cats/content/). Like TSPA, we might be able to ask them to consider a machine-friendly specification for these types of harms. Being associated with ActivityPub might make them more applicable to Matrix too.

If we *really* don't want to host the list as Matrix, we can probably look to the [W3C Data Privacy Vocabularies and Controls Community Group (DPVCG)](https://www.w3.org/groups/cg/dpvcg/) to establish a set of identifiers. The DPVCG might not take on the work because not all harms are privacy related. 

[OASIS](https://www.oasis-open.org/) might be able to help create a standard external to Matrix as well, though their primary output locations are the [ISO](https://www.iso.org/home.html) and [IEC](https://iec.ch/homepage). It may be faster/easier/different to go through a local national body instead, like the [Standards Council of Canada (SCC)](https://scc-ccn.ca/). The [DTSP Safe Framework Specification](https://dtspartnership.org/the-safe-framework-specification/) is hosted as ISO/IEC 25389 (as Jim mentions), so it's plausible that we could get a similar harms taxonomy specification there too.

[DTSP](https://dtspartnership.org/) might also be able to help create a standard to reference.

turt2live: I've updated this MSC in https://github.com/matrix-org/matrix-spec-proposals/pull/4456/commits/acddc1610bea120a34ccc006e17b3633e96451eb to clarify that the intended destination for this MSC is not the "core" spec but rather something vaguely adjacent. This is done with an async dependency on [MSC4518: Registries](https://github.com/matrix-org/matrix-spec-proposals/pull/4518). While MSC4518 progresses through its rounds of review, this proposal can be landed in the appendices.

This should hopefully make it easier to lift & shift the identifiers out of the Matrix spec entirely if/when needed. For now though, I'm strongly in favour of placing the list external to the Client-Server API so we can start using them.

proposals/4456-safety-harms-appendix.md
turt2live: Timeline update: Search Redirection is still progressing at the architecture layer, but the design to have an error (with harm) be returned seems consistent enough. Reporting v2's generic reporting API needs a bit of battle testing in production before I'm comfortable declaring it "implemented".

proposals/4456-safety-harms-appendix.md
erikjohnston: This is tangentially being discussed in a registries MSC thread: https://github.com/matrix-org/matrix-spec-proposals/pull/4518#discussion_r3664918019

I have a strong feeling that if something is required in a spec version then it lives within the spec. This MSC currently requires that servers support certain harm identifiers in certain spec versions, which I think make it unsuitable for a registry as things stand.

(I don't think I've fully understood the wider context of this proposal, so take the following with a grain of salt)

However, I am a bit dubious about how the versioning is happening here and whether it is actually required in practice. Specifically, if these are being sent to clients then what spec version a server supports isn't the issue: it's what version the client supports (so it can render it sensibly). If these errors are actually coming from a policy server then the server doesn't even need to care about the error code and can happily forward one it doesn't understand. 

This is really reminding me of normal error codes, where you really want a tree shape such that it is easier to add new error codes and clients will gracefully fallback to the most specific error code they support. With a scheme somewhat like that (I'm sure it's not an exact fit), then these wouldn't even need to be versioned alongside the main spec.

turt2live: (these identifiers are both emitted by servers (via policy servers, in some cases) and submitted to servers via reporting endpoints, fwiw).

turt2live: Per changes in MSC4518, the registry in this proposal will need updating to not list required functionality inside the registry itself.

PR #4457, opened 2026-04-28
MSC4457: Generic reporting API

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/reporting-v2/generic-report-api/proposals/4457-generic-reporting-api.md)

### DIFF
--- a/proposals/4457-generic-reporting-api.md
+++ b/proposals/4457-generic-reporting-api.md
@@ -34,8 +34,8 @@ Authorization: <normal Client-Server API authentication>
 Content-Type: application/json
 
 {
-	"type": "complaint", // future scope: "appeal" and possibly other types
-	// ... other fields per `type`
+  "type": "complaint", // future scope: "appeal" and possibly other types
+  // ... other fields per `type`
 }
 ```
 
@@ -49,38 +49,38 @@ the top level:
 
 ```jsonc
 {
-	// The *primary* identifier (user ID, etc) the complaint is regarding. Currently can be one of the following:
-	// * A user ID
-	// * An event ID
-	// * A room ID
-	// * A room alias (noting that room ID reports are more reliable because aliases can drift between rooms)
-	// * A server name (prefixed with "server:" to distinguish it from a namespaced ID below)
-	// * An MXC URI (media URI)
-	// * The string "m.system" to denote a complaint regarding the reporting system itself
-	// * A common namespaced identifier (https://spec.matrix.org/v1.18/appendices/#common-namespaced-identifier-grammar)
-	//
-	// The above identifiers are structured so the server can identify each one individually. For example, the server
-	// knows it's dealing with a user report if `regarding` starts with `@`.
-	//
-	// REQUIRED.
-	"regarding": "<identifier>",
+  // The *primary* identifier (user ID, etc) the complaint is regarding. Currently can be one of the following:
+  // * A user ID
+  // * An event ID
+  // * A room ID
+  // * A room alias (noting that room ID reports are more reliable because aliases can drift between rooms)
+  // * A server name (prefixed with "server:" to distinguish it from a namespaced ID below)
+  // * An MXC URI (media URI)
+  // * The string "m.system" to denote a complaint regarding the reporting system itself
+  // * A common namespaced identifier (https://spec.matrix.org/v1.18/appendices/#common-namespaced-identifier-grammar)
+  //
+  // The above identifiers are structured so the server can identify each one individually. For example, the server
+  // knows it's dealing with a user report if `regarding` starts with `@`.
+  //
+  // REQUIRED.
+  "regarding": "<identifier>",
 
-	// The type of harm being reported in this complaint. Currently, the available harms are defined
-	// by MSC4456: https://github.com/matrix-org/matrix-spec-proposals/pull/4456
-	// A future MSC is expected to advertise which custom harms (if any) the server supports.
+  // The type of harm being reported in this complaint. Currently, the available harms are defined
+  // by MSC4456: https://github.com/matrix-org/matrix-spec-proposals/pull/4456
+  // A future MSC is expected to advertise which custom harms (if any) the server supports.
   //
   // Note: This is the reporter's opinion and is not necessarily fact.
-	//
-	// REQUIRED.
-	"harm": "<harm identifier>",
+  //
+  // REQUIRED.
+  "harm": "<harm identifier>",
 
-	// The text the user supplied to support this complaint. The input field presented to the user SHOULD ask them
-	// to *briefly* describe the content or harm caused.
+  // The text the user supplied to support this complaint. The input field presented to the user SHOULD ask them
+  // to *briefly* describe the content or harm caused.
   //
   // Cannot exceed 1024 bytes (before trimming whitespace).
-	//
-	// REQUIRED (cannot be an empty string, after trimming whitespace).
-	"description": "This user is spamming",
+  //
+  // REQUIRED (cannot be an empty string, after trimming whitespace).
+  "description": "This user is spamming",
 }
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/4457-generic-reporting-api.md
turt2live, line 1: Implementation requirements:

* [ ] Client (using new API)
* [ ] Server (offering new API)

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+  caller from filing reports or the server has determined that the caller does not have visibility
+  on the reported object/entity (ie: can’t see the event they’re complaining about, or can’t appeal
+  someone else’s ban).
+
+
+*Note:* Servers MAY use `404 M_NOT_FOUND` and `403 M_FORBIDDEN`. The remaining error codes MUST only
+be returned in applicable cases. This optionality is to allow servers to customize their operations
+to their individual regulatory requirements and needs. For example, a server might choose to validate
+that a user can see an event, but also choose *not* to return an error if they can’t (instead, it’d
+get flagged internally on the report).
```
turt2live: "MUST only" is trying to carry a requirement for a future compliance suite: a server which always returns a 404 for local users, even if they exist, is non-compliant.

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+
+The new report endpoint is available to [guests](https://spec.matrix.org/v1.18/client-server-api/#guest-access).
+
+How the server processes the report remains an implementation detail. A future MSC will clarify where
+exactly a report should be routed to. Implementations SHOULD NOT assume that reports will only go to
+a single place in the future. This is important for implementations which provide the endpoint as
+separate software from the bulk of the homeserver, as otherwise that software might assume that it’s
+only going to have to populate a single destination queue.
```
turt2live, line 146: "Implementations SHOULD NOT assume..." might make more sense as future MSCs are written. I'm mostly trying to say "don't treat this endpoint as **only** reporting to server admins - it'll have routing information later, and your software will need to be capable of sending communications as needed"

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+    can optionally send it to the mods too").
+  * Possibly based on choice too ("do you want to send this to `remote.example.org` too? Should we
+    tell them you send the report?").
+* Adding attachments (screenshots, events, additional info) and amendments.
+* Closing/cancelling reports (and generally the open/closed status of a report).
+* Communication frameworks, like who to send updates to. Possibly including a "reply to" address to
+  direct safety teams at a more responsive user ID, for example.
+* Appeals with an actions database/ID. For example: "appeal {$this} ban".
+* Possibly adding evidence to someone else's report, similar to a character witness statement.
+  * This may be better handled by a new `information` report type.
```
turt2live, line 164: this may further be best to just leave to the backend. Individual users can submit distinct complaints that are deduplicated/aggregated on the backend.

turt2live (commented):
Bringing over comments from the private draft doc (all authored by me).

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+  // REQUIRED.
+  "harm": "<harm identifier>",
+
+  // The text the user supplied to support this complaint. The input field presented to the user SHOULD ask them
+  // to *briefly* describe the content or harm caused.
+  //
+  // Cannot exceed 1024 bytes (before trimming whitespace).
+  //
+  // REQUIRED (cannot be an empty string, after trimming whitespace).
+  "description": "This user is spamming",
```
JadedBlueEyes, line 87: For some harms it doesn't make sense to require a description - the subject is enough to make a decision. Requiring it is likely to cause some portion of reports to have content like "."

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+*Author's note*: The choice to use `regarding` is deliberate for a bit of natural language: a user is
+**reporting** a **complaint** **regarding** something/someone which caused **harm**. `description`
```
JadedBlueEyes: I think this is slightly clunky, although not too important.

- Reporting an Appeal (or other type of object) doesn't make as much sense
- The user is *making* a complaint with a *subject* and a *description*, and the moderator looks at the *subject* of the complaint, not the *regarding* of the complaint.

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+
+Implementation and unstable usage of this MSC is expected to populate this section. Currently, expected
+risks include:
+
+* Not having a generic enough API to support appeals.
+* Clients implementing something which makes it harder to add routing later.
+* Returning a report ID only to discard it is a bit strange.
+* Internal handling of reports may be difficult.
+* Limiting to authenticated users (including guests) may prove to be an issue.
+
```
JadedBlueEyes, line 182: - A user filling out a report only to hit a forbidden or rate limited error code, wasting the work / losing the data

JadedBlueEyes (commented):
Just a quick skim

proposals/4457-generic-reporting-api.md
CrazyNicc: Having a content of "." in some proposals where it isn't needed is imo better than people not providing a description where it would be necessary.

turt2live: We're currently thinking that `description` is probably going to become required for some types of harm and optional for others. Real world data shows most reports are filed with unhelpful reasons - asking the user to pick the type of harm is infinitely more helpful than the reasons currently filed. 

Some examples of reasons we've seen (based on clicking random reports in our system):

* "Reporting user `@someone:example.org`" - No indication for *why* they're being reported. The event referenced is their membership event in a room.
* "spam" - The event sometimes is obviously spam, but other times it's things like "hello" or a github URL.
* "embargo" - Presumably this is either a translation issue or the word means something to someone else.
* "SPAM" - A few users have pressed extra keys to make it all caps instead.

Of the ~50 reports I opened, 3 had actionable reasons beyond "spam" and 1 used a slur to describe the user they were reporting. The remainder had one of the above 4 reasons.

proposals/4457-generic-reporting-api.md
turt2live: hmm, yea, the example breaks down quite a bit with appeals. The important distinction I'm trying to make is it's not a *reason* for the concern, it's a *description* of the concern. It's subtle, but "reason" implies that the report is being made in error up front while "description" asks for more information.

proposals/4457-generic-reporting-api.md
turt2live: We'll need to prove this during implementation, but we're expecting that only users which intentionally work around a compliant client's reporting flow will encounter these errors. For example, hammering the report endpoint with curl or force-opening a dialog on an event where the button to report isn't shown. 

In other words: users who see forbidden or rate limit errors should already be expecting that outcome.

proposals/4457-generic-reporting-api.md
```diff
@@ -0,0 +1,219 @@
...
+* `429 M_LIMIT_EXCEEDED` - Rate limited (“too many reports too quickly”).
+* `400 M_BAD_JSON` - A `REQUIRED` field is missing, especially for the `type`, or a field is too long/short.
+* `400 M_INVALID_PARAM` - The `type` is unknown to the server. (Note: servers MUST support at least
+  `complaint` and more in future MSCs).
+* `404 M_NOT_FOUND` - The server cannot process the report because the reported identifier does not
+  exist (or is unknown). Servers MUST NOT return this error for `m.system` complaints.
+* `403 M_FORBIDDEN` - The caller cannot file this report. For example, the server has banned the
+  caller from filing reports or the server has determined that the caller does not have visibility
+  on the reported object/entity (ie: can’t see the event they’re complaining about, or can’t appeal
+  someone else’s ban).
```
turt2live, line 131: This should clarify that self-reports are valid, even outside of appeals. We don't currently see such reports in the wild because clients disallow it, but self-reporting is a way for some people to reach out to get help.

During implementation I think we should experiment with providing self-report as an option and measure how effective it is.

PR #4458, opened 2026-04-28
MSC4458: Handling incoming JSON in the server-server API

richvdh:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/rav/proposal/json_handling_clarifications/proposals/4458-handling-received-json.md)

Implementation: Synapse already does all this.

---

SCT stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4458#issuecomment-5040151220)
[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4458#issuecomment-5040152537)

### DIFF
--- a/proposals/4458-handling-received-json.md
+++ b/proposals/4458-handling-received-json.md
@@ -71,6 +71,34 @@ clarifications.
    Implementations are free to implement this requirement by dropping one or
    other duplicate, or by rejecting the JSON as a whole.
 
+   > ### Justification
+   >
+   > The previous paragraph merits some explanation as to its safety.
+   >
+   > Duplicate keys can only arise when the sending server is buggy or
+   > malicious. There are two possibilities:
+   >
+   >  * Having dropped one or both of the duplicates, we find the signature
+   >    matches. We now have a regular event or request containing valid JSON; in
+   >    the case of an event, the deduplicated version becomes our
+   >    authoritative reference.
+   >
+   >  * Having dropped one or both of the duplicates, the signature does not
+   >    match. We reject the request or event. In the case of an event which
+   >    other servers accept via the previous possibility, we may end up later
+   >    requesting a copy of that event from an honest server — in which case,
+   >    we will receive a deduplicated copy that everyone in the room can agree
+   >    on.
+   >
+   > In either case, everyone in the room except the originating server ends up
+   > coming to the same eventual conclusion about the room DAG. The originating
+   > server may have a different idea of the room DAG, but that is the expected
+   > fate of a buggy server implementation.
+   >
+   > Note also that neither possibility allows the sending server to do
+   > anything that they wouldn't already be able to do by sending well-formed
+   > JSON without duplicates.
+
    Assuming that the internal data structures mentioned above only allow one
    value per key, no action will be needed by homeservers to comply with this
    requirement, as duplicate keys will be dropped automatically.
--- a/proposals/4458-handling-received-json.md
+++ b/proposals/4458-handling-received-json.md
@@ -401,3 +401,14 @@ N/A
 ## Dependencies
 
 None
+
+## Appendix: test vectors for Canonical JSON
+
+The existing spec contains a number of
+[examples](https://spec.matrix.org/v1.19/appendices/#examples) of the Canonical
+JSON representation of a number of JSON objects. This MSC is accompanied by an
+extended set of test vectors which come as corrolaries of the existing rules,
+or the clarifications set out above.
+
+These test cases are at
+https://github.com/matrix-org/matrix-spec-proposals/tree/rav/proposal/json_handling_clarifications/proposals/data/4458.
--- a/proposals/4458-handling-received-json.md
+++ b/proposals/4458-handling-received-json.md
@@ -307,7 +307,7 @@ Such an change remains under consideration, but is left for a future MSC.
    duplicate. A split-brain *can* occur between A and the rest of the room, but
    that is acceptable because A is buggy/malicious. (There are plenty of ways
    for a server to split-brain itself from the rest of the room via
-   non-conformant behaviour: poor JSON implentation is not special in this
+   non-conformant behaviour: poor JSON implementation is not special in this
    regard).
 
 ## Unstable prefix
--- a/proposals/4458-handling-received-json.md
+++ b/proposals/4458-handling-received-json.md
@@ -1,0 +1,293 @@
+# MSC4458: Handling incoming JSON in the server-server API
+
+The structure of Matrix rooms, and the security of the server-server API,
+relies on the calculation and checking of signatures and hashes of data
+structures. Examples include:
+
+ * Server-server API requests are
+   [authenticated](https://spec.matrix.org/v1.17/server-server-api/#request-authentication)
+   by a signature on a data structure containing the details of the request.
+
+ * Room events must be
+   [signed](https://spec.matrix.org/v1.17/server-server-api/#signing-events) by
+   the homeserver that created them (and, in the case of invite events, by the
+   target of the invite), to guard against impersonation.
+
+ * [Event IDs](https://spec.matrix.org/v1.17/rooms/v12/#event-ids) are hashes of
+   events, to guard against equivocation (i.e. an attacker sending different
+   event contents to different recipients, claiming that they fit in the same
+   place in room history).
+
+ * Room events also include a [content
+   hash](https://spec.matrix.org/v1.17/server-server-api/#calculating-the-content-hash-for-an-event),
+   to guard against tampering of the `contents` (which are not directly
+   protected by the signature or Event ID).
+
+All of these processes rely on encoding the data structure in question as
+[Canonical JSON](https://spec.matrix.org/v1.17/appendices/#canonical-json), and
+feeding the bytes thus obtained into the relevant hashing or signing algorithm.
+
+However, the specification today is not sufficiently clear on how exactly
+Canonical JSON should be used. This MSC seeks to provide the necessary
+clarifications.
+
+## Proposal
+
+1. When receiving JSON over the server-server API (either in a request or
+   response body), server implementations MUST first deserialize the incoming
+   JSON into internal data structures. These data structures must be able to
+   fully represent the data types permissible in Canonical JSON:
+     * strings
+     * numbers (or at least, integers in the range `[-(2**53)+1, (2**53)-1]`)
+     * objects
+     * arrays
+     * `true`, `false`, `null`.
+
+   The original JSON MUST then be discarded, and all future operations must be
+   based on the deserialized structures. This includes hashing and signature
+   checking, which must be done be encoding the deserialized structures as
+   Canonical JSON.
+
+   In particular: it is **not** sufficient to construct the Canonical JSON by
+   modifying the unparsed JSON data, since this brings the risk of inconsistent
+   parsing when the JSON is later parsed. Further discussion of the potential
+   problems can be found in the [Alternatives](#altenatives) below.
+
+2. Implementations MUST guard against duplicate keys in the incoming JSON, and
+   ensure that duplicates are dropped before encoding to Canonical JSON. (In
+   JSON, a "duplicate key" refers to a JSON object which contains two or more
+   entries using the same key; for example: `{"a": 1, "a": 2}`.)
+
+   Implementations are free to implement this requirement by dropping one or
+   other duplicate, or by rejecting the JSON as a whole.
+
+   Assuming that the internal data structures mentioned above only allow one
+   value per key, no action will be needed by homeservers to comply with this
+   requirement, as duplicate keys will be dropped automatically.
+
+   To be explicit: Canonical JSON encodings MUST NOT contain duplicate keys.
+
+3. A clarification to the handling of the [UTF-16
+   surrogates](https://www.unicode.org/faq/utf_bom#utf16-2) `U+D800`-`U+DFFF`.
+   It is possible that implementations will encounter these codepoints in
+   incoming JSON (possibly, though not necessarily, represented with a `\uXXXX`
+   escape). For example, the JSON `{"emoji": "\ud83d\ude00"}` contains one possible
+   representation of the smiley face, 😀.
+
+   There is no way to legally express the UTF-16 surrogates in Canonical JSON,
+   since Canonical JSON forbids the use of `\uXXXX` escapes other than for
+   some of the C0 control characters, and Canonical JSON is defined to use
+   the UTF-8 encoding (which
+   [forbids](https://www.unicode.org/faq/utf_bom#utf8-4) encoding of the
+   surrogates).
+
+   To be explicit, then: Canonical JSON MUST NOT encode the UTF-16
+   surrogates.
+
+   Valid surrogate pairs SHOULD be decoded to their Unicode code point on
+   parsing, and then encoded as a 4-byte UTF-8 sequence when encoding as
+   Canonical JSON. JSON containing unpaired surrogates MUST be rejected
+   (either on parsing, or when attempting to encode as Canonical JSON).
+
+## Potential issues
+
+TODO
+
+## Alternatives
+
+A number of alternative approaches have been proposed, and are discussed below.
+
+### Require servers to transmit Canonical JSON on the wire
+
+One approach would be to mandate that federation request bodies be serialized
+as Canonical JSON. Doing so could bring performance benefits: the signature on
+an incoming federation request could be checked much more efficiently than
+today.
+
+The benefit is less clear for *response* bodies, since they are not signed in
+their own right, though some responses, such as
+[`/backfill`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1backfillroomid),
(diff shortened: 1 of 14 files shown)


proposals/4458-handling-received-json.md
turt2live, line 1: Implementation requirements:

* [ ] Server (sending acceptable JSON)
* [ ] Server (receiving JSON and parsing per requirements)

Specifically for duplicate keys (to prove there's compatibility with both approaches):
* [ ] A server which *drops* one or the other duplicate fields.
* [ ] A server which *discards* the entire JSON object.

richvdh: > Implementation requirements:
> 
>     * [ ]  Server (sending acceptable JSON)

The MSC doesn't say much about *sending* JSON, so I don't think there's anything to do here.

>     * [ ]  Server (receiving JSON and parsing per requirements)

Synapse follows the requirements in the MSC. For federation requests, parsing the incoming request happens [here](https://github.com/element-hq/synapse/blob/76b4fdceed0739d83ac79588416dc88f25d8d14e/synapse/federation/transport/server/_base.py#L308): if you chase the logic deep enough you'll see it gets parsed into `dict`s, `list`s, etc.

> Specifically for duplicate keys (to prove there's compatibility with both approaches):
> 
>     * [ ]  A server which _drops_ one or the other duplicate fields.

Again, Synapse does this, by virtue of parsing JSON objects into dicts.

>     * [ ]  A server which _discards_ the entire JSON object.

I'm not sure we have such an impl, though I'm unconvinced its necessary: duplicate fields can only possibly be malicious, and discarding such requests is surely sensible.

turt2live: The discard implementation I feel is required because discarding an event has consequences the server needs to be prepared for (namely, that it discarded a valid event that may be important to its understanding of the room going forward). This may also require protocol changes if the eventual consistency mechanism doesn't work properly to remedy those consequences.

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+
+ * The sort order for object keys is subtly different: JCS encodes the keys
+   using UTF-16 and sorts the resulting code units, whilst Canonical JSON sorts
+   the UTF-32 code points.
+
+   In other words, JCS sorts the smiley face 😀 (U+1F600) before the Hebrew
+   letter דּ (U+FB33), whilst Canonical JSON uses the opposite sorting.
+
+It's not obvious that the benefits of switching to an alternative Canonical
+JSON format are worthwhile at this time.
```
clokep, line 325: I'm assuming JCS also only solves the canonicalization portion and not the other aspects discussed here, e.g. handling of hashes/signatures. So the contents of this MSC would essentially still apply, just with slightly different canonicalization rules?

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+
+All of these processes rely on encoding the data structure in question as
+[Canonical JSON](https://spec.matrix.org/v1.17/appendices/#canonical-json), and
+feeding the bytes thus obtained into the relevant hashing or signing algorithm.
+
+However, the specification today is not sufficiently clear on how exactly
+Canonical JSON should be used. This MSC seeks to provide the necessary
+clarifications.
+
+## Proposal
```
clokep, line 34: Maybe not necessary for this proposal, but it would be very nice to expand on https://spec.matrix.org/v1.18/appendices/#examples with examples of bytes -> canonicalized version instead of bus object -> bytes. Especially if we could include malicious example.

proposals/4458-handling-received-json.md
richvdh: Yup, as we say further up: "This is a somewhat orthogonal question". It doesn't actually solve anything in this MSC, it's just a topic that gets bandied around in the same conversation. Thoughts on clarifying this (or just removing the section altogether) welcome.

clokep: Not sure if there's any need to make it more obvious, was mostly clarifying what orthogonal meant. I'll think about it.

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+   some of the C0 control characters, and Canonical JSON is defined to use
+   the UTF-8 encoding (which
+   [forbids](https://www.unicode.org/faq/utf_bom#utf8-4) encoding of the
+   surrogates).
+
+   To be explicit, then: Canonical JSON MUST NOT encode the UTF-16
+   surrogates.
+
+   When parsing incoming JSON, valid surrogate pairs SHOULD be decoded to their
+   Unicode codedpoint. Such codepoints MUST then be encoded as a 4-byte UTF-8
```
ara4n:
```suggestion
   Unicode codepoint. Such codepoints MUST then be encoded as a 4-byte UTF-8
```

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Use an alternative Canonical JSON format
+
+Rather than using a Canonical JSON format specific to Matrix, could we use a
+standardised format such as "JSON Canonicalization Scheme" (JCS, specified by
+[RFC8785](https://www.rfc-editor.org/rfc/rfc8785.html))?
+
+This is a somewhat orthogonal question, but worth discussing while we're in the
+area. Adopting a standardized canonical JSON format might mean that we could
+benefit from existing implementations.
```
clokep:
```suggestion
This is a somewhat orthogonal question as it does not help solve the issues discussed above with verifying signatures and hashes, but it is worth discussing while we're in the
area. Adopting a standardized canonical JSON format might mean that we could
benefit from existing implementations.
```

Maybe this would help?

proposals/4458-handling-received-json.md
clokep: https://github.com/matrix-org/matrix-spec-proposals/pull/4458#discussion_r3164048512 might resolve this?

I'm also happy with no changes though.

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+   parsing when the JSON is later parsed. Further discussion of the potential
+   problems can be found in the [Alternatives](#allow-servers-to-canonicalise-incoming-json-separately-to-parsing) below.
+
+2. Implementations MUST guard against duplicate keys in the incoming JSON, and
+   ensure that duplicates are dropped before encoding to Canonical JSON. (In
+   JSON, a "duplicate key" refers to a JSON object which contains two or more
+   entries using the same key; for example: `{"a": 1, "a": 2}`.)
+
+   Implementations are free to implement this requirement by dropping one or
+   other duplicate, or by rejecting the JSON as a whole.
```
ara4n, line 72: This jars a lot when reading. I assume the reason you can solve the duplication any way you like is because that if you choose a way that doesn't match the sender's intent, the hash won't match and you'll reject the event. Meanwhile, the sender got it wrong if they sent you dups anyway. However, rather than the reader having to figure this out themselves at this point, i'd spell it out as an explanation.

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,319 @@
...
+The benefit is less clear for *response* bodies, since they are not signed in
+their own right, though some responses, such as
+[`/backfill`](https://spec.matrix.org/v1.17/server-server-api/#get_matrixfederationv1backfillroomid),
+contain event bodies, which are signed, so requiring them to be transmitted as
+Canonical JSON could be beneficial.
+
+In any case, the main difficulty is that such a rule has no benefit unless it
+can be enforced. Receiving servers would have to verify that incoming JSON is
+actually compliant with Canonical JSON, to prevent attacks from malicious
+servers. This is not a trivial operation.
```
ara4n, line 185: We should perhaps also point out that existing JSON parsers tend not to provide this verification mode - and even if they did, they tend to be wildly inconsistent (and link the various sad examples of inconsistent JSON parsers)

ara4n: oh, the next para had this. doh.

proposals/4458-handling-received-json.md
richvdh: Have accepted your proposal.

proposals/4458-handling-received-json.md
richvdh: Have added some words.

erikjohnston (approved):

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,403 @@
...
+  Further, manipulating the raw JSON in this way is very unnatural in some
+  languages: it is much easier to parse the incoming JSON, manipulate the data,
+  and re-serialize.
+
+### Allow servers to canonicalise incoming JSON separately to parsing
+
+Some implementations (for example, Dendrite) operate on incoming JSON at the
+byte level to transform it to Canonical JSON, before checking
+signatures/hashes. They then retain the *original* JSON, deserializing it where
+necessary for later operation. Should we allow such an approach?
```
tulir, line 250: I think retaining the original JSON is likely never safe, but retaining the transformed canonical version probably is (assuming the transformer actually works). Go's jsonv2 supports reformatting json in-place <https://pkg.go.dev/encoding/json/jsontext#Encoder.WriteValue> and mautrix-go now has a canonical json package built on top of that (with the non-RFC8785 utf-8 object reordering and float rejection as a step after the standard reformat): <https://github.com/mautrix/go/blob/main/crypto/canonicaljson/jsonv2.go>. There are a bunch of tests which pass both with a roundtrip and just in-place reformatting.

proposals/4458-handling-received-json.md
richvdh: > The discard implementation I feel is required because discarding an event has consequences the server needs to be prepared for 

In theory, we could set up a test environment to simulate this. Eve generates a state event with duplicate fields; Alice rejects the event; Bob discards one of the duplicates and accepts it. The question then is: if Bob generates a new event which relies on having accepted Eve's event, can Alice continue to operate in the room?  But we know the answer to that, because it's identical to Alice not having seen the event in the first place: she can just request the event from Bob.

So, I'm not sure what such a test would achieve. Meanwhile it would take a bunch of time and effort and hinder us just merging this clarification to the spec.

proposals/4458-handling-received-json.md
richvdh: > I think retaining the original JSON is likely never safe, but retaining the transformed canonical version probably is (assuming the transformer actually works). 

Yeah. I think that's basically the next section ("canonicalise at the byte level, then parse the JSON")?

> Go's jsonv2 supports reformatting json in-place [pkg.go.dev/encoding/json/jsontext#Encoder.WriteValue](https://pkg.go.dev/encoding/json/jsontext#Encoder.WriteValue) and mautrix-go now has a canonical json package built on top of that (with the non-RFC8785 utf-8 object reordering and float rejection as a step after the standard reformat): [mautrix/go@`main`/crypto/canonicaljson/jsonv2.go](https://github.com/mautrix/go/blob/main/crypto/canonicaljson/jsonv2.go?rgh-link-date=2026-05-09T13%3A09%3A07.000Z).

I'm slightly confused what `Encoder.WriteValue` does, and the mautrix-go code you link to seems to use [`Value.Format`](https://pkg.go.dev/encoding/json/jsontext#Value.Format) instead?

tulir: > Yeah. I think that's basically the next section ("canonicalise at the byte level, then parse the JSON")?

Oh yeah that does cover it

> I'm slightly confused what Encoder.WriteValue does, and the mautrix-go code you link to seems to use [Value.Format](https://pkg.go.dev/encoding/json/jsontext#Value.Format) instead?

`Value.Format` calls `WriteValue` internally https://github.com/golang/go/blob/go1.26.3/src/encoding/json/jsontext/value.go#L113-L123. `WriteValue` basically walks the input like any parser does, but instead of parsing into an in-memory type, it emits equivalent JSON with whatever formatting rules it was given.

jsonv2 rejects duplicate keys and invalid utf-8 rather than silently allowing them, so those parts shouldn't cause problems.

proposals/4458-handling-received-json.md
richvdh: Hrm, what are you thinking of here? Maybe a table like this:

| *Input JSON bytes* | *Input JSON string* | *Output Canonicaljson string* | *Output Canonicaljson bytes* |
| --- | --- | --- | --- | 
| `7b7d` | `{}` | `{}` | `7b7d` |
| `7b20226f6e65223a20312c202274776f223a202254776f22207d` | `{ "one": 1, "two": "Two" }` | `{"one":1,"two":"Two"}` | `7b226f6e65223a312c2274776f223a2254776f227d` |
| `7b613a20312c20613a20327d` | `{a: 1, a: 2}` | `{a:1}` | `7b613a317d` |
| | | or `{a:2}` | `7b613a327d` |
| | | or reject altogether | |

Other useful examples might be:

 * `31303030...30652d3130` (i.e. `1000...0e-10`), where `...` represents some large number of zeros, can legitimately be rejected, or canonicalise to a slightly smaller huge int, depending in the size of your floats
 * `2261205c7564383364207a22` (i.e. `"a \ud83d z"`) should be rejected, because it contains an unpaired UTF-16 surrogate
 * `2261205c75643833645c7564653030206222` (i.e. `"a \ud83d\ude00 b"`) canonicalises to `"a \xf0\x9f\x98\x80 b"` (because the UTF-16 pair must be encoded as a UTF-8 sequence).
 * an example of objects with keys containing characters outside the BMP, which is important for distinction from JCS.

proposals/4458-handling-received-json.md
richvdh: Ok. Do you think we need changes to the text of the MSC here?

tulir: Probably not

proposals/4458-handling-received-json.md
turt2live: Revisting this some time later: Synapse does indeed appear to be a suitable implementation.

proposals/4458-handling-received-json.md
richvdh: I got a bit nerdswiped, and added a set of test vectors at https://github.com/matrix-org/matrix-spec-proposals/tree/rav/proposal/json_handling_clarifications/proposals/data/4458.

richvdh:
MSCs proposed for Final Comment Period (FCP) should meet the requirements outlined in the [checklist](https://github.com/matrix-org/matrix-spec-proposals/blob/main/MSC_CHECKLIST.md?rgh-link-date=2026-06-30T14%3A25%3A30Z) prior to being accepted into the spec. This checklist is a bit long, but aims to reduce the number of follow-on MSCs after a feature lands.

- [n/a] For each endpoint that is introduced or modified:
  - [n/a] Have authentication requirements been specified?
  - [n/a] Have rate-limiting requirements been specified?
  - [n/a] Have guest access requirements been specified?
  - [n/a] Are error responses specified?
    - [n/a] Does each error case have a specified `errcode` (e.g. `M_FORBIDDEN`) and HTTP status code?
      - [n/a] If a new `errcode` is introduced, is it clear that it is new?
  - [n/a] Are the [endpoint conventions](https://spec.matrix.org/latest/appendices/#conventions-for-matrix-apis) honoured?
    - [n/a] Do HTTP endpoints `use_underscores_like_this`?
    - [n/a] Will the endpoint return unbounded data? If so, has pagination been considered?
    - [n/a] If the endpoint utilises pagination, is it consistent with [the appendices](https://spec.matrix.org/latest/appendices/#pagination)?
  - [n/a] Is the reason for a new room version clearly stated? For example, modifying the set of redacted fields changes how event IDs are calculated, thus requiring a new room version.
  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Unstable prefix
  - [x] Dependencies
  - [n/a] Unstable prefixes [consider](https://github.com/matrix-org/matrix-spec-proposals/blob/main/README.md#unstable-prefixes) the awkward accepted-but-not-merged state
  - [n/a] Chosen unstable prefixes do not pollute any global namespace (use “org.matrix.mscXXXX”, not “org.matrix”).

richvdh:
@mscbot fcp merge

proposals/4458-handling-received-json.md
```diff
@@ -0,0 +1,414 @@
...
+the uncanonicalised data on the wire rather than a canonicalised representation
+of it. Both approaches are discussed in more detail under
+[Alternatives](#alternatives).
+
+However, requiring servers to transmit Canonical JSON is currently believed to
+be technically infeasible, as discussed below.
+
+Having implementations sign and hash the data on the wire would mean that
+servers could at least get as far as agreeing on whether an event was correctly
+signed, even if their JSON parsers then treat the event differently. However,
```
erikjohnston: I'm not overly fond of the framing that we're signing the *event* in this case, even if it is immediately clarified in the next sentence.


```suggestion
Having implementations sign and hash the data on the wire would mean that
servers could at least get as far as agreeing on whether the event bytes were correctly
signed, even if their JSON parsers then treat the parsed event differently. However,
```

Maybe?

PR #4460, opened 2026-05-01
MSC4460: Extensible Events - Alternative unstable support

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/extev/unblock/proposals/4460-extensible-events-unstable.md)

### DIFF
--- a/proposals/4460-extensible-events-unstable.md
+++ b/proposals/4460-extensible-events-unstable.md
@@ -1,0 +1,172 @@
+# MSC4460: Extensible Events - Alternative unstable support
+
+[MSC1767 (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+outlines a system for rendering unknown event types in clients called "Extensible Events".
+
+The original proposal gates using extensible events (meaning "content blocks") behind a room version
+to ensure that clients are not exposed to events they might not be able to render. A previous draft
+of the proposal tried to allow `m.room.message` and similar event types to use content blocks, though
+that was later removed to discourage long-term use of `m.room.message`.
+
+It's also been over 3 years since the extensible events design was accepted by the Spec Core Team (SCT).
+Back in 2023, it was intended to keep momentum moving and land the remainder of the core content block
+types. Changes to project priority ended up causing the extensible events MSCs to fall below an activity
+threshold, but the intent to land the overall system remains.
+
+Over the last 3 years there's been interest in using extensible events for more functionality, and the
+SCT is interested in actually merging proposals like [MSC3381: Polls (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3381-polls.md)
+which use extensible events. [MSC3765: Rich topics (merged)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3765-rich-room-topics.md)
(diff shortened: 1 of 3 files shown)


proposals/4460-extensible-events-unstable.md
```diff
@@ -0,0 +1,172 @@
...
+
+   [^2]: Location messages have a few different options, so are limited to only `m.text` content blocks
+   for now. This may change in future when one of the location sharing proposals is accepted.
+
+   **Note**: Many of the listed content blocks are *unstable*, therefore requiring prefixed content
+   block types.
+
+   All content blocks MUST additionally match their non-content block form. Events MUST have all
+   allowed and required content blocks if they use any content blocks. For example, an `m.image`
+   message type using the `m.text` content block *must* also have `m.file` and `m.image_details`.
```
HarHarLinks, line 81: > all allowed and required content blocks

maybe i'm misunderstanding, but wouldn't that mean that a `msgtype` of `m.image` would also require an `m.thumbnail`? i don't think that's the idea though; it seems more sensible to require events to have all required content blocks plus any optional ones that have equivalents of the old format present. e.g. if there is a `thumbnail_url` and `thumbnail_info`, then there MUST also be a `m.thumbnail` content block with matching values.

proposals/4460-extensible-events-unstable.md
```diff
@@ -0,0 +1,172 @@
+# MSC4460: Extensible Events - Alternative unstable support
+
+[MSC1767 (accepted)](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md)
+outlines a system for rendering unknown event types in clients called "Extensible Events".
+
+The original proposal gates using extensible events (meaning "content blocks") behind a room version
+to ensure that clients are not exposed to events they might not be able to render. A previous draft
+of the proposal tried to allow `m.room.message` and similar event types to use content blocks, though
+that was later removed to discourage long-term use of `m.room.message`.
```
HarHarLinks, line 9: so just to be clear, the idea is to retract on this issue and accept the risk?

proposals/4460-extensible-events-unstable.md
```diff
@@ -0,0 +1,172 @@
...
+  content blocks as though they are `m.room.message` events with `msgtype: m.image`.
+* `m.poll.start`, `m.poll.response`, and `m.poll.end` from MSC3381 permit content blocks as proposed.
+* [`m.room.topic`](https://spec.matrix.org/v1.18/client-server-api/#mroomtopic) permits the `m.topic`
+  content block (with `m.text` sub-block). This is already the case in the spec, but is listed here
+  for clarity.
+
+For all other event types, only unstable content block types are permitted. A room version is still
+required to use stable content block types, as per MSC1767. Clients SHOULD NOT render `m.text`,
+MSC3551 (files), MSC3552 (images), MSC3553 (videos), MSC3955 (notices), and MSC3927 (audio) content
+blocks on non-`m.room.message` event types outside of an extensible events-supporting room version.
```
HarHarLinks, line 110: that is not what I can read from https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1767-extensible-events.md#unstable-prefix.

I understand MSC1767 as "you can use unstable ID types in any room, or stable ID types in unstable rooms". Was this modified by another MSC? I also don't really see why this is necessary.

This also seems to decrease the value of this MSC for "extensible events forward compatibility"?

PR #4461, opened 2026-05-05
MSC4461: Storing per-message profiles for users

tulir:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/tulir/per-message-profile-storage/proposals/4461-per-message-profile-storage.md)

### DIFF
--- a/proposals/4461-per-message-profile-storage.md
+++ b/proposals/4461-per-message-profile-storage.md
@@ -1,0 +1,84 @@
+# MSC4461: Storing per-message profiles for users
+[MSC4144] introduces per-message profiles, which allow users and bots to
+override their profile info for a single message. The original use case was
+primarily bots, but the feature is useful for humans too. However, humans don't
+like re-entering their profile info every time they send a message. Therefore,
+a method for storing reusable profiles is needed.
+
+[MSC4144]: https://github.com/matrix-org/matrix-spec-proposals/pull/4144
+
+## Proposal
+A new `m.per_message_profiles` account data event is introduced. The event
+content is a map from a user-defined profile "shortcode" to the per-message
+profile data that gets sent in events. All fields inside the profile object
+are defined by [MSC4144].
+
+The shortcode is an arbitrary user-defined string, which can be used as a key
+when selecting a profile to use. Arbitrary unicode is allowed, but spaces
+SHOULD NOT be used, as clients that use a command-based UI might split
+parameters on spaces.
+
+```json
+{
+  "type": "m.per_message_profiles",
+  "content": {
+    "meow": {
+      "id": "cat",
+      "displayname": "Cat 🐈️"
+    },
+    "mrrp": {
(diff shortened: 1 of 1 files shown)


proposals/4461-per-message-profile-storage.md
tulir, line 1: Implementation requirements:

* [ ] Client that can create reusable profiles
* [x] Client that can use reusable profiles when sending messages

proposals/4461-per-message-profile-storage.md
```diff
@@ -0,0 +1,84 @@
...
+Users with lots of profiles can end up with a large account data event, but
+it's unlikely to be larger than existing big account data like push rules or
+`m.direct`.
+
+Since it's a single account data event, editing can hit race conditions. It's
+likely not a problem, as it's only edited directly by humans (as opposed to
+something like `m.direct`, which clients may update in the background).
+
+## Alternatives
+There are various other places where profiles could be stored, like
```
onestacked: I think in room profiles (probably as state events) might be a better idea, as these could then also be shown in the user list. That would make MSC4144/non-appservice bridges nicer to use.

tulir: That wouldn't help normal users at all since users can't send state events in most rooms. It'd also mean profiles have to be added to every room instead of just once in global account data. Definitely not an option for this MSC, but I guess I should go into more detail in the alternatives section

proposals/4461-per-message-profile-storage.md
tulir: https://github.com/gomuks/gomuks/commit/2e5489a5d7bbed8b026bb2eea82135943fa4f3d5 (+ https://github.com/gomuks/gomuks/commit/4daa127ac894d3301133973160ba2d47fc564466 and https://github.com/gomuks/gomuks/commit/fbe6fdbaabeaae46fb4c03e0d18260941f235d49) is an implementation of sending profiles, no UI for creating yet

https://github.com/gomuks/gomuks/commit/951bac5f2fe680d9cbcbeb98cd468ecd06ae16a1 updates the implementation to the v2 format added in 61aec706a25c2157898c0bbef75b7d32cde2a41c

https://github.com/gomuks/gomuks/commit/c416f43181fce71b9ffbb1d1c535446aa0e31715 updates it to v3

proposals/4461-per-message-profile-storage.md
j0lol: Room profiles could be a good idea for privacy reasons. A client could in theory have 'private' PMPs and then 'share' them with the room which would add it to the room profile. (Maybe this would be good for extended profiles, or something?) See also: https://pluralkit.me/guide/#privacy

tulir: All profiles in this MSC are private until you use them in a room. Publishing profiles outside of sending messages should be a separate MSC

PR #4462, opened 2026-05-06
MSC4462: Links in Profile

dozro:
[Rendered](https://github.com/dozro/matrix-spec-proposals/blob/proposal/links-in-bio/proposals/4462-links-in-profile.md)

Signed-off-by: Rye <git@itsrye.dev>


proposals/4462-links-in-profile.md
tulir, line 1: Implementation requirements:

* [ ] Client that lets the user set links
* [ ] Client that displays links in profile

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
+# MSC4462: Links in Profile
+
+*This Proposal is related to future [MSCxxxx: Verifiable Links in Profile](https://github.com/matrix-org/matrix-spec-proposals/pull/xxxx) but not dependent on it*
```
thetayloredman: I would advise not linking to a proposal that doesn't exist until you've published the draft.

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
...
+        "description": "alt-account",
+        "link": "matrix:u/alice:example.org"
+     }
+   ]
+}
+```
+
+Content of a object in `m.links`:
+
+- `description`: human readable description of what a client should show next to the link as label (free-form)
```
thetayloredman, line 40: This does not allow clients or servers to easily identify logos (and in the future, verification methods) for links shared on user profiles. It will likely become necessary to explicitly numerate platforms as a list of common namespaced identifiers, or simply an array of URIs (e.g. instead of `m.website https://zirco.dev` `m.matrix_user @logn:zirco.dev` `m.domain zirco.dev` the URI could just be `matrix:u/@logn:zirco.dev` `https://zirco.dev` `dns:zirco.dev`). This removes the need for an explicit enumeration in the spec.

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
...
+
+Profiles MAY have a `m.links` field as an array of objects. These fields can be fetched through the
+[profile API endpoints](https://spec.matrix.org/unstable/client-server-api/#profiles).
+
+```json
+{
+   "m.links": [
+     {
+        "description": "homepage",
+        "link": "https://example.org"
```
thetayloredman: I feel like `uri` is a better name for this key, considering a `matrix:` ID is not a "link" per se. I honestly feel like overall calling these "connections" and not links is better.

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
...
+- `description`: human readable description of what a client should show next to the link as label (free-form)
+- `link`: the link to show and render as link (it SHOULD be a valid http/https/matrix url)
+
+Clients can then display the links of a user in their respective user popup/profile view UI.
+
+Clients SHOULD limit the length of `description` to 200 characters and the amount of links to at most 20.
+If a profile does not conform to the length/amount of links limit a receiving client MAY truncate or ignore
+entries exceeding these limits.
+
+If a `link` is some scheme other than `http`, `https`, or `matrix` a receiving client SHOULD consider these
```
thetayloredman: [See above](https://github.com/matrix-org/matrix-spec-proposals/pull/4462/changes#r3197130981) it may be beneficial to allow other URI types, like `dns:` or [`openpgp4fpr:`](https://github.com/matrix-org/matrix-spec-proposals/pull/4439/changes#diff-855306dbcbefbf1e81a594ad92cbcc39b45cc719322526e17081470ed1e48c12R37)

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
...
+This proposal doesn't introduce a way to verify link ownership (i.e. is that link really belonging
+to @alice:example.org).
+
+A followup MSC ([MSCxxxx: Verifiable Links in Profile](https://github.com/matrix-org/matrix-spec-proposals/pull/xxxx)) will tackle that.
+
+### Trust & Safety
+
+Links may lead to sites deemed inappropriate for a given context (e.g. NSFW sites, illegal content, etc...).
+This could be solved by introducing room-local extensible profiles (not part of this proposal), so that
+room moderators can do meaningful moderation of links in profiles.
```
thetayloredman, line 67: This is a sizeable concern with any free-text input in user profiles. As with [this review on MSC4440](https://github.com/matrix-org/matrix-spec-proposals/pull/4440#discussion_r3000811772) and with more detail [this review on MSC4426](https://github.com/matrix-org/matrix-spec-proposals/pull/4426#discussion_r2855381196). This is a major decision to make regarding the ability room moderators do or don't have to manage profiles.

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,95 @@
...
+### Listing them in Bio
+
+If MSC4440 gets accepted an alternative to this proposal would be to just put the links in the bio.
+
+Most important difference being that this proposal introduces a structured way of saving it in the
+profile, and so making it better readable for machines.
+Clients can, when they recognize a link, for example `https://bsky.app/profile/alice.example.org`
+display a appropriate logo next to the link.
+Note that this proposal specifically doesn't include a `platform` or `logo` field, as that could
+be abused way too easily.
```
thetayloredman, line 80: What would the abuse vector of a `platform` field be?

dozro: saying it is a link to a mastodon page while it leading somewhere completely different

proposals/4462-links-in-profile.md
dozro: i thought about adding a `verification_method` field in the follow-up msc to specify if it’s a dns, http, matrix verification

proposals/4462-links-in-profile.md
dozro: yeah “connections” does have a nice ring to it, you’re right, it sounds a bit vague tho 🤔 

but wouldn’t naming it `uri` imply also accepting `mailto` uris and so on?
[RFC 3986 - Example URIs](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2)

thetayloredman: @dozro Yes, and that also sounds like a good use for this!

dozro: > @dozro Yes, and that also sounds like a good use for this!

so it *should* allow email addresses and the like?

thetayloredman: > so it *should* allow email addresses and the like?

Email addresses especially sound useful in corporate environments. It would be far harder to verify independently but I feel like more is better here. Hence why maybe even just using valid URIs would be cool, although that brings its own issues (file: for example)

dozro: > Email addresses especially sound useful in corporate environments. It would be far harder to verify independently but I feel like more is better here. Hence why maybe even just using valid URIs would be cool, although that brings its own issues (file: for example)

would still prefer to have the set of allowed schemes to be limited. but yeah allowing email addresses does sound like a good idea :)

proposals/4462-links-in-profile.md
dozro: agreed, but i sadly don't think this MSC could do anything to satisfy these concerns

thetayloredman: I do believe a wider solution to this problem (as it impacts at least 3 open MSCs) is out of scope, but it's definitely a dependency

proposals/4462-links-in-profile.md
cyrneko: Email sounds reasonable, could be a little one-click "Make my email public" toggle in situations where you already have an Email associated with your account.

proposals/4462-links-in-profile.md
cyrneko: Yea the concern is that someone crafts a link that looks like it would lead to a trusted site that the user knows, when in reality it leads to a phishing scam or something.

1vel: If the full URI is shown this wouldn't really be an issue.

dozro: Would `platform` just be something like `m.platform.mastodon`, `m.platform.twitter`, etc? 

That could replace the `description` text on clients that are worried about displaying free form text in a profile 🤔

proposals/4462-links-in-profile.md
Johennes: I think at least to some extent you could identify services or types of services based on the URL?

proposals/4462-links-in-profile.md
```diff
@@ -0,0 +1,105 @@
...
+- `uri`: the link to show and render as link (it SHOULD be a valid http/https/matrix/mailto uri)
+
+Clients can then display the links of a user in their respective user popup/profile view UI.
+
+Clients SHOULD limit the length of `description` to 200 characters and the amount of links to at most 20.
+If a profile does not conform to the length/amount of links limit a receiving client MAY truncate or ignore
+entries exceeding these limits.
+
+If a `uri` is some scheme other than `http`, `https`,`mailto`, or `matrix` a receiving client SHOULD consider these
+links invalid and either hide them or mark them as being invalid.
```
Johennes, line 50: Would hiding make it difficult to add further schemes in future? They would never be shown on existing clients this way.

dozro: my motivation was mainly to prevent potentially dangerous schemes like `file`, or `data` by encouraging (`SHOULD`) clients to ignore schemes other than the allowed ones (http, https, matrix, mailto).

as discussed here: https://github.com/matrix-org/matrix-spec-proposals/pull/4462#discussion_r3197463568

or would a denylist (aka saying `clients SHOULD NOT display potentially abusable uri schemes`) make more sense?

PR #4463, opened 2026-05-08
MSC4463: Backfilling Pinned Events

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/pin-backfill/proposals/4463-backfilling-pinned-events.md)

Superseded by https://github.com/matrix-org/matrix-spec-proposals/pull/4465.

### DIFF
--- a/proposals/4463-backfilling-pinned-events.md
+++ b/proposals/4463-backfilling-pinned-events.md
@@ -1,0 +1,70 @@
+# MSC4463: Backfilling Pinned Events
+
+When a homeserver joins a room it lacks prior history for, any pinned events in its
+history are often invisible to clients, as server implementations typically respond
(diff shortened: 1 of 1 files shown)


JadedBlueEyes:
Some implementations, e.g. Continuwuity, already fetch missing events on demand, and so don't exhibit this flaw (which also technically affects other event references, like replies).

thetayloredman:
> Some implementations, e.g. Continuwuity, already fetch missing events on demand, and so don't exhibit this flaw (which also technically affects other event references, like replies). 

@JadedBlueEyes True. It might just flat out be better to say that servers SHOULD fetch events when the CS endpoint is hit. Would you say there are any substantial tradeoffs in that?

thetayloredman:
Superseded by https://github.com/matrix-org/matrix-spec-proposals/pull/4465.

PR #4464, opened 2026-05-08
MSC4464: verifiable links in profile

dozro:
[Rendered](https://github.com/dozro/matrix-spec-proposals/blob/proposal/verifiable-links-in-profile/proposals/4464-verifiable-links-in-profile.md)

Follow-up for [MSC4462](https://github.com/matrix-org/matrix-spec-proposals/pull/4462)

### DIFF
--- a/proposals/4464-verifiable-links-in-profile.md
+++ b/proposals/4464-verifiable-links-in-profile.md
@@ -142,7 +142,7 @@ the following:
         "result": "success",
         "method": "m.method.dns" | "m.method.relation" | "m.method.matrix",
         // expiry should be defined as the unix epoch after which to try again
-        "expires": "1778107123",
+        "expires": 1778107123,
         "scope": "m.scope.site" | "m.scope.domain" | "m.scope.account"
     },
     // optional additional url preview stuff
--- a/proposals/4464-verifiable-links-in-profile.md
+++ b/proposals/4464-verifiable-links-in-profile.md
@@ -1,4 +1,4 @@
-# MSC0000: Verifiable Links in Profile
+# MSC4464: Verifiable Links in Profile
 
 A lot of other chat platforms (like Discord) and social media platforms (like
 Mastodon, Sharkey, ...) allow you to put a list of links in your profile which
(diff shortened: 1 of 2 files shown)


proposals/4464-verifiable-links-in-profile.md
turt2live, line 1: Implementation requirements:

* [ ] Server
* [ ] Client

DidiDidi129:
If it’s a work in progress should u mark as a draft for now?
Edit: Nvm I think it was a draft earlier but not anymore

DidiDidi129:
I’m a little bit concerned that this would be used by bad actors for impersonation attacks

Eg they could spin up a homeserver without the requirements you outlined and make their profile link to any account they want.

dozro:
> I’m a little bit concerned that this would be used by bad actors for impersonation attacks
> 
> Eg they could spin up a homeserver without the requirements you outlined and make their profile link to any account they want.

@DidiDidi129  Since the verification is handled via the **Client-to-Server (C2S) API** rather than a Server-to-Server (S2S) protocol, the validation logic resides within the **user's own homeserver**.

Consequently, the security of this implementation relies on the integrity of the specific homeserver the user is registered with. While the trustworthiness of public homeservers is a valid point of discussion, the specific impersonation vector you mentioned, where a malicious actor spins up a rogue homeserver to affect others, is not feasible here, as their server would have **no authority** over the victim's verification process.

Note: Please use threads (reviews) for further technical discussions to keep the PR history organized.

DidiDidi129:
> > I’m a little bit concerned that this would be used by bad actors for impersonation attacks
> 
> > 
> 
> > Eg they could spin up a homeserver without the requirements you outlined and make their profile link to any account they want.
> 
> 
> 
> @DidiDidi129  Since the verification is handled via the **Client-to-Server (C2S) API** rather than a Server-to-Server (S2S) protocol, the validation logic resides within the **user's own homeserver**.
> 
> 
> 
> Consequently, the security of this implementation relies on the integrity of the specific homeserver the user is registered with. While the trustworthiness of public homeservers is a valid point of discussion, the specific impersonation vector you mentioned, where a malicious actor spins up a rogue homeserver to affect others, is not feasible here, as their server would have **no authority** over the victim's verification process.
> 
> 
> 
> Note: Please use threads (reviews) for further technical discussions to keep the PR history organized.

Ahh yes that makes sense. Sorry for my comprehension issues lmao

PR #4465, opened 2026-05-08
MSC4465: On-Demand Fetch for Missing Events

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/2bab4ed686bae6eda88d801849a56aa8fcc6b26a/proposals/4465-on-demand-fetch-for-missing-events.md)

Implementations:
- [Continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/976)

Supersedes [MSC4463](https://github.com/matrix-org/matrix-spec-proposals/pull/4463)

### DIFF
--- a/proposals/4465-on-demand-fetch-for-missing-events.md
+++ b/proposals/4465-on-demand-fetch-for-missing-events.md
@@ -1,0 +1,52 @@
+# MSC4465: On-Demand Fetch for Missing Events
+
+When a homeserver joins a room it lacks prior history for, it is often missing historical
+event PDUs that may be relevant or visible to users (e.g. via [`m.room.pinned_events`][pins],
+replies, etc.). Many implementations respond to [requests for unknown events][CS_get_evt]
+with `M_NOT_FOUND` is it does not exist locally, which this MSC addresses.
+
+When conversation within the room refers to prior (pinned) events that a member does
+not have access to, this increases friction and eliminates one of the primary use
+cases of referencing events: to share important historical information with new room
+participants.
+
+This proposal supersedes [MSC4463][MSC4463] by generalizing it to all forms of references,
+including message replies.
+
+## Proposal
+
+If a server receives a request to [`/_matrix/client/v3/rooms/{roomId}/event/{eventId}`][CS_get_evt]
+for an `eventId` it is not aware of, it SHOULD attempt to fetch the event from other
+servers in the room and return it to the client.
+
+Homeservers MAY use backfill or the [`/_matrix/federation/v1/event/{eventId}`][SS_get_evt] for this
+purpose. Homeservers use a best-effort mechanism to select servers believed to have the event (such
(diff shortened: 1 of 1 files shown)


turt2live (commented):
Implementation requirements:

* [ ] Server

proposals/4465-on-demand-fetch-for-missing-events.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4465-on-demand-fetch-for-missing-events.md
```diff
@@ -0,0 +1,52 @@
...
+## Proposal
+
+If a server receives a request to [`/_matrix/client/v3/rooms/{roomId}/event/{eventId}`][CS_get_evt]
+for an `eventId` it is not aware of, it SHOULD attempt to fetch the event from other
+servers in the room and return it to the client.
+
+Homeservers MAY use backfill or the [`/_matrix/federation/v1/event/{eventId}`][SS_get_evt] for this
+purpose. Homeservers use a best-effort mechanism to select servers believed to have the event (such
+as the alias server or room moderators' servers), but this is left as an implementation decision
+as with all other forms of backfill.
```
turt2live, line 25: This feels like something that can/should already be done without an MSC? The proposal should clarify why an MSC is required for this change.

turt2live (commented):
Early review from approximately 30 seconds of reading

proposals/4465-on-demand-fetch-for-missing-events.md
thetayloredman: @turt2live I was unsure if a MSC was necessary here, but I felt like it was more than just a clarification because it's a SHOULD, not a MAY (and the MSC process would help bring more attention to it). That is up to you/the SCT if you believe it could be PR'd directly into the spec.

turt2live: Can we get more of the rationale added to the MSC to help make that determination? It's not clear to me why this MSC is even required or why a similar spec PR would be required.

thetayloredman: I'm not quite sure how it could be explained any better than the pins example: I join a room, someone tells me "hey, look at the pinned message for the information you just asked for!" My client tries to fetch the event it doesn't have and then receives M_NOT_FOUND (Synapse) and now I'm out of luck on seeing what that information was without trying to paginate back and get backfill for it all. This change strongly encourages servers to do what Continuwuity does, which is attempt to get the event instead of just erroring.

turt2live: right, but why does that need to be spec behaviour? Is this "just" a Synapse bug?

thetayloredman:
This is clearly not something that requires an MSC, I am abandoning this proposal

PR #4466, opened 2026-05-09
MSC4466: Altering profile change propagation

gingershaped:
[Rendered](https://github.com/gingershaped/matrix-spec-proposals/blob/ginger/msc/profile-change-propagation/proposals/4466-profile-change-propagation.md)

Signed-off-by: Ginger <ginger@gingershaped.computer>

---

Implementations:
- Server: [Continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1763)
- Client: [Sable](https://github.com/SableClient/Sable/pull/1117)

### DIFF
--- a/proposals/4466-profile-change-propagation.md
+++ b/proposals/4466-profile-change-propagation.md
@@ -54,7 +54,6 @@ When sending a new `m.room.member` event with the intent of updating `displaynam
 `avatar_url`, the server:
 - MUST set `displayname` or `avatar_url`, whichever is applicable, to the newly provided
 value. 
-- MUST copy `third_party_invite` from the previous membership event, if it is present.
 - MUST NOT copy
 `join_authorised_via_users_server` from the previous membership event. The presence of this field requires the event
 to be signed by the authorizing server as well, which is of course not possible without that server's participation.
--- a/proposals/4466-profile-change-propagation.md
+++ b/proposals/4466-profile-change-propagation.md
@@ -40,13 +40,10 @@ keeping with current behavior.
 
 Example for the behavior of `unchanged`:
 
-| `displayname` in global account data | `displayname` in room membership | `displayname` in PUT request | New
-`displayname` in room membership | | ------------------------------------ | -------------------------------- |
----------------------------- | ------------------------------------ | | Alice Margatroid                     |
-Alice "Nickname" Margatroid      | Alice                        | Unchanged from Alice "Nickname" Margatroid,
-because `displayname` in room membership is different from displayname in global account data | Alice Margatroid |
-Alice Margatroid | Alice | Changed to Alice, because `displayname` in room membership is identical to `displayname`
-in global account data
+| `displayname` in global account data | `displayname` in room membership | `displayname` in PUT request | New `displayname` in room membership |
+| ------------------------------------ | -------------------------------- | ---------------------------- | ------------------------------------ |
+| Alice Margatroid                     | Alice "Nickname" Margatroid      | Alice                        | Unchanged from Alice "Nickname" Margatroid, because `displayname` in room membership is different from displayname in global account data |
+| Alice Margatroid | Alice Margatroid | Alice | Changed to Alice, because `displayname` in room membership is identical to `displayname` in global account data |
 
 
 ### Key copying rules
--- a/proposals/4466-profile-change-propagation.md
+++ b/proposals/4466-profile-change-propagation.md
@@ -128,7 +128,7 @@ Servers which implement MSC4437 as well as this proposal should strongly conside
 extensions to MSC4437. While this proposal is unstable, servers which support the extensions should use
 `computer.gingershaped.msc4466.propagate_to` instead of `m.propagate_to` as a query parameter for the `PUT
 /_matrix/client/v3/profile/{userId}` endpoint and `PUT /_matrix/client/unstable/com.beeper.msc4437/profile/{userId}`
-unstable endpoint. Servers may use the `computer.gingershaped.msc4466.msc4437` unstable feature flag to indicate
+unstable endpoint. Servers may use the `computer.gingershaped.msc4466.msc4437_extensions` unstable feature flag to indicate
 that they support the extensions.
 
 ## Dependencies
--- a/proposals/4466-profile-change-propagation.md
+++ b/proposals/4466-profile-change-propagation.md
@@ -126,7 +126,7 @@ this proposal.
 
 Servers which implement MSC4437 as well as this proposal should strongly consider supporting this proposal's
 extensions to MSC4437. While this proposal is unstable, servers which support the extensions should use
-`computer.gingershaped.msc4466.propagate_to` instead of `m.propagate_to` as a query parameter for the `PUT
+`computer.gingershaped.msc4466.propagate_to` instead of `propagate_to` as a query parameter for the `PUT
 /_matrix/client/v3/profile/{userId}` endpoint and `PUT /_matrix/client/unstable/com.beeper.msc4437/profile/{userId}`
 unstable endpoint. Servers may use the `computer.gingershaped.msc4466.msc4437_extensions` unstable feature flag to indicate
 that they support the extensions.
--- a/proposals/4466-profile-change-propagation.md
+++ b/proposals/4466-profile-change-propagation.md
@@ -21,14 +21,16 @@ behavior when sending an updated `m.room.member` event:
 
 - `all`: The server MUST send a `m.room.member` event for `{userId}` to every room that `{userId}` has a `join`
 membership in, with `{keyName}` set to the newly provided value, or removed entirely, in accordance with the
-[key copying rules](#Key-copying-rules).  - `unchanged`: For each room that `{userId}` has a `join` membership
+[key copying rules](#key-copying-rules).
+- `unchanged`: For each room that `{userId}` has a `join` membership
 in, the server MUST check the value of `{keyName}` in the `m.room.member` event's content. If `PUT` was used and
 the content's current value of `{keyName}` is exactly equivalent to the value of `{keyName}` in the user's global
 profile data, or `DELETE` was used and the user's global profile data is already missing an entry for `{keyName}`,
 the server MUST send a `m.room.member` event to that room with `{keyName}` changed to the newly provided value or
-removed entirely, in accordance with the [key copying rules](#Key-copying-rules). For all other rooms, the server
+removed entirely, in accordance with the [key copying rules](#key-copying-rules). For all other rooms, the server
 MUST NOT send a `m.room.member` event to that room. This allows clients to easily offer a way to change a user's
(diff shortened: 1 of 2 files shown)


proposals/4466-profile-change-propagation.md
tulir, line 1: Implementation requirements:

* [ ] Server that implements `propagate_to`
* [ ] Client that uses `propagate_to=unchanged`
* [ ] Client that uses `propagate_to=none`

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,139 @@
...
+| Alice Margatroid | Alice Margatroid | Alice | Changed to Alice, because `displayname` in room membership is identical to `displayname` in global account data |
+
+
+### Key copying rules
+
+When sending a new `m.room.member` event with the intent of updating `displayname` or
+`avatar_url`, the server:
+- MUST set `displayname` or `avatar_url`, whichever is applicable, to the newly provided
+value. 
+- MUST copy `third_party_invite` from the previous membership event, if it is present.
```
zecakeh: Why would we encounter this field? According to the spec, this field is only present with an `invite` membership, and profile changes propagation only occurs with `join` memberships.

gingershaped: Removed. That was a mistake on my part while reading the list of fields in the spec.

proposals/4466-profile-change-propagation.md
gingershaped: Implemented in Continuwuity: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1763

gingershaped: Implemented in Sable: https://github.com/SableClient/Sable/pull/1117

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+a method to give a reason for a profile change, and describing such a method is out of the scope of this MSC.
+- MUST copy ALL OTHER fields from the previous membership event's content unchanged, including whichever of
+`displayname` or `avatar_url` was not changed.
+
+Future MSCs which add new fields to the content of `m.room.member` events may specify additional rules.
+
+### Extensions to MSC4437
+
+For servers which implement
+[MSC4437](https://github.com/beeper/matrix-spec-proposals/blob/replace-entire-profile/proposals/4437-replace-entire-profile.md),
```
turt2live, line 70:
```suggestion
[MSC4437](https://github.com/matrix-org/matrix-spec-proposals/pull/4437),
```

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+If the `propagate_to` query parameter is not provided, the server MUST behave as if it were set to `all`, in
+keeping with current behavior.
+
+Example for the behavior of `unchanged`:
+
+| `displayname` in global account data | `displayname` in room membership | `displayname` in PUT request | New `displayname` in room membership |
+| ------------------------------------ | -------------------------------- | ---------------------------- | ------------------------------------ |
+| Alice Margatroid                     | Alice "Nickname" Margatroid      | Alice                        | Unchanged from Alice "Nickname" Margatroid, because `displayname` in room membership is different from displayname in global account data |
+| Alice Margatroid | Alice Margatroid | Alice | Changed to Alice, because `displayname` in room membership is identical to `displayname` in global account data |
```
turt2live, line 48: This table should have columns aligned in the diff for ease of review.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+The server MUST additionally update the user's global profile data with the new value of
+`{keyName}`, regardless of the value of `propagate_to`. This is the data that is returned by [`GET /_matrix/client/v3/profile/{userId}`](https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv3profileuserid)
+and related endpoints.
+
+If the `propagate_to` query parameter is not provided, the server MUST behave as if it were set to `all`, in
+keeping with current behavior.
+
+Example for the behavior of `unchanged`:
+
+| `displayname` in global account data | `displayname` in room membership | `displayname` in PUT request | New `displayname` in room membership |
```
turt2live, line 45: Profiles aren't stored in account data at the spec level - they are their own concept.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+`join_authorised_via_users_server` from the previous membership event. The presence of this field requires the event
+to be signed by the authorizing server as well, which is of course not possible without that server's participation.
+- MUST NOT copy `reason` from the previous membership event. The profile change API does not currently specify
+a method to give a reason for a profile change, and describing such a method is out of the scope of this MSC.
+- MUST copy ALL OTHER fields from the previous membership event's content unchanged, including whichever of
+`displayname` or `avatar_url` was not changed.
+
+Future MSCs which add new fields to the content of `m.room.member` events may specify additional rules.
+
+### Extensions to MSC4437
```
turt2live, line 67: This section effectively creates the dependency. It would be best to just fold the section into the rest of the proposal, making it shorter in the process.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+a request to this endpoint.
+
+If the `propagate_to` query parameter is not provided, the server MUST behave as if it were set to `all`, in
+keeping with the current behavior of MSC4437.
+
+## Potential issues
+
+If `propagate_to` is set to `unchanged`, the server will need to access state information for
+every room which the user is currently joined to. Since the server should already have this information in its
+database, the effort required to perform this task is deemed to be acceptable.
```
turt2live, line 94: This is still a performance issue. The work required to look up state events is non-trivial in a few cases, and preparing to send events is not all that challenging for a server (or shouldn't be - it's one of the server's core functions). 

The proposal certainly has benefit to avoid excess events from being sent (custom -> global -> reset to custom profile), but it can't as easily write off performance impact when it's a core driver in its own introduction.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+No method is provided by this MSC for clients to specify a list of rooms to send a `m.room.member` event in. A client
+may implement this easily by updating the user's global profile without sending any events (setting `propagate_to`
+to `none`), and then sending `m.room.member` events itself in whatever rooms it chooses.
+
+This MSC does not specify a solution to the problem of the server being flooded with thousands of `state_ids` and
+`get_missing_events` requests when updating a user's membership in a large number of rooms at once. The server
+MAY choose to send the new membership events at a slow pace to manage the load.
```
turt2live, line 102: The introduction cites this very problem as a motivation for the proposal, so it would need to be addressed as-written. If the proposal doesn't intend to fix this issue, it can't use that issue as a reason to merge the proposal.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Alternatives
+
+[MSC4069](https://github.com/matrix-org/matrix-spec-proposals/pull/4069) is a similar proposal which
+only implements a Boolean flag to suppress propagation (equivalent to setting this proposal's `propagate_to` to
+`none`), and tasks clients with deciding how to update a user's profile if they wish to do anything more complex. The
+use case of "update my profile but not in rooms where I changed it" seems common enough to the author of this
+proposal that having a server-side implementation of it, which will likely be more efficient since the server
+already has access to relevant state, would be useful. An implementation of MSC4069 could likely be converted to
+an implementation of this proposal with little effort.
```
turt2live, line 112: For future proposals: it's generally best to adapt existing proposals through comments on the diff than it is to open a whole new proposal. For this particular case it's fine and doesn't cause any process issues, but other proposals can quickly create a headache for reviewers as they need to compare other options.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Security considerations
+
+None.
```
turt2live, line 116: This section requires more detail. *Every* MSC has security considerations, or at least privacy considerations. If the proposal truly doesn't have those considerations, that needs to be explained in detail.

proposals/4466-profile-change-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Unstable prefix
+
+While this proposal is unstable, implementations should use
+`computer.gingershaped.msc4466.propagate_to` instead of `propagate_to` as a query parameter for the `PUT
+/_matrix/client/v3/profile/{userId}/{keyName}` endpoint and `DELETE /_matrix/client/v3/profile/{userId}/{keyName}`
+endpoint. Servers may use the `computer.gingershaped.msc4466` unstable feature flag to advertise support for
+this proposal.
```
turt2live, line 124:
```suggestion
endpoint. Servers may use the `computer.gingershaped.msc4466` unstable feature flag on `/versions` to advertise support for
this proposal.
```

turt2live (changes requested):
Thanks for writing this up! This review is a first pass editorial review ahead of the proposal entering the (somewhat lengthy) proposed-for-proposed-FCP queue. More comments are possible (sorry) as the SCT completes its rounds of review.

proposals/4466-profile-change-propagation.md
timedoutuk: I'm not sure why this is even listed as a problem at all. Servers already need to access said state for every room during a global profile update, this is no different for "unchanged".
Furthermore, "looking up state events may be non-trivial" and "preparing to send events is not challenging" are contradictory to each other - the state has to be looked up to *send an event* (populate auth events, run auth checks). However, by choosing not to send an event, the server doesn't have to modify the room, which *is* an expensive operation in comparison to... not doing that.

gingershaped: I've removed this bullet point from this section entirely (https://github.com/matrix-org/matrix-spec-proposals/pull/4466/commits/dfafc49a872307102a23d3378f2859572cf7e868). As Nex mentions, this MSC can only reduce the amount of state lookups that the server would have to do by virtue of letting the server send fewer state events during a profile update.

proposals/4466-profile-change-propagation.md
gingershaped: I've elaborated on this point. (https://github.com/matrix-org/matrix-spec-proposals/pull/4466/commits/dfafc49a872307102a23d3378f2859572cf7e868)

proposals/4466-profile-change-propagation.md
gingershaped: I've updated the introduction to better justify the benefits of the proposal. (https://github.com/matrix-org/matrix-spec-proposals/pull/4466/commits/dfafc49a872307102a23d3378f2859572cf7e868)

PR #4470, opened 2026-05-13
MSC4470: Routing reports to non-local destinations

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/reporting-v2/x-routing/proposals/4470-report-routing.md)

*Note*: This is a draft while it undergoes very early review. The idea needs validation prior to any implementation begins.

### DIFF
--- a/proposals/4470-report-routing.md
+++ b/proposals/4470-report-routing.md
@@ -1,0 +1,127 @@
+# MSC4470: Routing reports to non-local destinations
+
+*This MSC is part of “Reporting v2” - a project led by the Foundation’s T&S team to improve communication
+and effectiveness of reports on Matrix.*
+
+**Note**: This proposal is not intended to be a complete idea. It should be heavily reviewed prior
+to implementation. It may be closed or replaced by a different approach as needed.
+
+[MSC4457](https://github.com/matrix-org/matrix-spec-proposals/pull/4457) introduces a generic reporting
+API which can be used to report a wide variety of content and material. The API lacks an ability to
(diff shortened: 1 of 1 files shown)


proposals/4470-report-routing.md
turt2live, line 1: Implementation requirements:

* Withheld pending review.

proposals/4470-report-routing.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+## Security considerations
+
+* This can allow a reporter to spam-forward reports at unrelated destinations. Local servers are
+  encouraged to make use of the `429` error code to limit a sender's ability to spam-forward reports
+  at multiple destinations, especially when requests to forward reports to more than 4 destinations
+  are received. Additionally, clients SHOULD NOT provide UI/UX which allows a user to enter their own
+  destinations. Users SHOULD be given a (short) list of checkboxes to choose from, such as "Report to
+  this user's homeserver admins" and "Report to this room's moderators".
```
turt2live, line 104: It feels important to call out that this bit of the proposal causes the client to be the decision maker for where the user is able to forward their report. The proposal is capable of permitting the user to be the sole decision maker by nature of the API shape (and clients providing text boxes instead of checkboxes), but that doesn't feel entirely desirable.

This structure does exclude the user's local server from making the initial routing decisions clear to the user, though. A server might *always* send reports to a nearby community, for example, or might want to suggest that a user routes a report to a specific place by having a default-checked box in the client.

PR #4140, opened 2024-05-07
MSC4140: Cancellable delayed events

toger5:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4140-delayed-events-futures.md)

This could also supersede [MSC2228](https://github.com/matrix-org/matrix-spec-proposals/pull/2228) (by making it possible to send a redaction with the `/send` endpoint. This is the case as mentioned [here](https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$r_PveZzw9TqvN3qKjCmxnIavR2CCKRWvTPJoLW6BA8k?via=t2l.io&via=matrix.org&via=element.io))

**Implementations**

- [x] Synapse
  - https://github.com/element-hq/synapse/pull/17326
  - https://github.com/element-hq/synapse/pull/19038
  - https://github.com/element-hq/synapse/pull/19354
   - https://github.com/element-hq/synapse/pull/19479
   - https://github.com/element-hq/synapse/pull/19539
   - https://github.com/element-hq/synapse/pull/19794
   - https://github.com/element-hq/synapse/pull/19926
   - https://github.com/element-hq/synapse/pull/20156
- [x] JS-SDK
  - https://github.com/matrix-org/matrix-js-sdk/pull/4294 
  - https://github.com/matrix-org/matrix-js-sdk/pull/5066
  - https://github.com/matrix-org/matrix-js-sdk/pull/5133
  - https://github.com/matrix-org/matrix-js-sdk/pull/5340
  - https://github.com/matrix-org/matrix-js-sdk/pull/5504
	
**Known differences between current implementations and the proposal**

- The implementations use the existing endpoints `/send` and `/state` with a new query parameter `delay` for scheduling delayed events. The proposal has since switched to a dedicated endpoint `PUT /_matrix/client/v3/rooms/{roomId}/delayed_event/{eventType}/{txnId}` where the delay is included in the request body.
  - The work to update the implementations was started in https://github.com/element-hq/synapse/pull/19354 and https://github.com/matrix-org/matrix-js-sdk/pull/5133. To finish it, another `/versions` key would be required to manage the migration from the previous unstable implementation.
- The implementations do not yet support finalised delayed events and the HTTP 409 response to conflicting management requests on an already-sent/cancelled delayed event.  The work to support this was started in https://github.com/element-hq/synapse/pull/19038
[...]

PR #4477, opened 2026-05-20
*MSC4140: Convert cancelling delayed state events alternative into briefer potential issue

Johennes:
This addresses the concern discussed in https://github.com/matrix-org/matrix-spec-proposals/pull/4140/changes#r3259703571.

### DIFF
--- a/proposals/4140-delayed-events-futures.md
+++ b/proposals/4140-delayed-events-futures.md
@@ -601,9 +601,9 @@ it would make it transparent to clients, when an event was scheduled and when it
 
 ### Conflicting delayed state events
 
-A delayed state event can overwrite other state events that were sent inbetween the delayed event
-being scheduled and it being sent. Whether or not this is problematic, strongly depends on the
-use case though. When the overwrite is undesired, a possible remedy could be to cancel the scheduled
+A delayed state event can overwrite other state events that were sent in between the delayed event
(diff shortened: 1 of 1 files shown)


proposals/4140-delayed-events-futures.md
```diff
@@ -599,6 +599,17 @@ Both are reasonable data points that clients might want to use.
 This would solve issues related to delayed events since
 it would make it transparent to clients, when an event was scheduled and when it was distributed over federation.
 
+### Conflicting delayed state events
+
+A delayed state event can overwrite other state events that were sent inbetween the delayed event
+being scheduled and it being sent. Whether or not this is problematic, strongly depends on the
+use case though. When the overwrite is undesired, a possible remedy could be to cancel the scheduled
```
AndrewFerr: Nitpicks
```suggestion
A delayed state event can overwrite other state events that were sent in between the delayed event
being scheduled and it being sent. Whether or not this is problematic strongly depends on the
use case, though. When the overwrite is undesired, a possible remedy could be to cancel the scheduled
```

AndrewFerr (approved):
Thanks for dropping this in a clean way.

proposals/4140-delayed-events-futures.md
Johennes: Ah, thanks. 👍

PR #4478, opened 2026-05-20
*MSC4140: Blend use case specific considerations into the introduction directly

Johennes:
This is to address https://github.com/matrix-org/matrix-spec-proposals/pull/4140/changes#r3269609956 and https://github.com/matrix-org/matrix-spec-proposals/pull/4140/changes#r3259100997.


AndrewFerr (commented):
I like this, but since it impacts the MSC's relevance to VoIP, I'll defer approval to someone from the VoIP team (pinging @toger5 who is already active on the MSC)

Johennes:
@toger5 and I synced out of band and he has no concerns with this landing.

PR #2723, opened 2020-08-12
MSC2723: Forwarded message metadata

hummlbach:
Signed-off-by: Johannes Renkl <hummlbach@posteo.de>

A proposal to properly support marking forwards as what they are.

Thx to @Sorunome and @krille for getting things straight.

PR #2437, opened 2020-02-18
MSC2437: Store tagged events in Room Account Data

giomfo:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/giomfo/tagged_events/proposals/2437-tagged-events-account-data.md)

PR #1767, opened 2019-01-01
MSC1767: Extensible event types & fallback in Matrix (v2)

ara4n:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/matthew/msc1767/proposals/1767-extensible-events.md).

Obsoletes matrix-org/matrix-spec-proposals#1225

Shepherd: @turt2live 

----

Implementations:
* SDK: https://github.com/matrix-org/matrix-events-sdk
  * Current version of this MSC in the SDK: https://github.com/matrix-org/matrix-events-sdk/pull/22
  * Usage in Element Web (current version of this MSC): [React](https://github.com/matrix-org/matrix-react-sdk/pull/9915), [JS](https://github.com/matrix-org/matrix-js-sdk/pull/3067)
  * Usage in Element Web (v1 of this MSC): [React](https://github.com/matrix-org/matrix-react-sdk/pull/7462), [JS](https://github.com/matrix-org/matrix-js-sdk/pull/2091)
  
Proven by other MSC implementations as well (may be an older version of this MSC, but still compatible on a technical level):
* Polls: https://github.com/matrix-org/matrix-doc/pull/3381
* Voice Messages: https://github.com/matrix-org/matrix-doc/pull/3245
* Static Location Sharing: https://github.com/matrix-org/matrix-doc/pull/3488

Early conceptual work for extensible events usage:
* matrix-art: https://github.com/MTRNord/matrix-art/blob/154c3af36b7db5502fec26ad380079c712de0756/helpers/event_types.ts

Articles:
* https://www.artificialworlds.net/blog/2022/03/08/comparison-of-matrix-events-before-and-after-extensible-events/

----

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/1767#issuecomment-1382598157)

PR #4482, opened 2026-05-23
MSC4482: Ability to bookmark messages

Gernomaly:
[rendered](https://github.com/Gernomaly/matrix-spec-proposals/blob/fcbfd5e2de81fa209057039493dff5c2b9a3a5df/proposals/4482-ability-to-bookmark-messages.md)

### DIFF
--- a/proposals/4482-ability-to-bookmark-messages.md
+++ b/proposals/4482-ability-to-bookmark-messages.md
@@ -22,7 +22,7 @@ This kind of room SHOULD only hold some `m.bookmark` events defined below.
 
 This MSC introduces a new kind of event which is a timeline event that uses extensible events blocks.
 It introduces a new `m.pointer` block that references a particular event sitting in a given room.
-The pointer is provided as is: it may not be "valid" (i.e. the user does not have access / permission to view the referenced event).
+The pointer is provided as is: it may not be "valid" (i.e. the user does not have access / permission to view the referenced event). However, in case the bookmarked / referenced event has a `m.replace` relation, the sending client should use the original event_id of the relation chain. This assumes that a `m.bookmark` event references an event and all its "versions", not a particular version of this event. 
 
 ```json5
 {
@@ -41,8 +41,7 @@ The pointer is provided as is: it may not be "valid" (i.e. the user does not hav
       "room_id": "$ROOM_ID",
       "event_id": "$EVENT_ID",
       "via": ["$SERVER_NAME"]
-    
-  } 
+    }
 }
 ```
 
@@ -50,16 +49,32 @@ The fallback text block of this event should contain at least the [Matrix URI](h
 
 ## Client implementation recommendations 
 
-## General
+### General
 
-If the client support the `m.bookmarks` room type, it should'nt be displayed as a normal room in the regular room list. Instead the client could display bookmarks in dedicated sections of the client. 
-The client should ensure that only one `m.bookmarks` room is existing at once.
-If there are multiple `m.bookmarks` rooms, we leave this as an implementation detail.
+If the client support the `m.bookmarks` room type, it shouldn't be displayed as a normal room in the regular room list. Instead the client could display bookmarks in dedicated sections of the client. 
+The client should ensure that only one `m.bookmarks` room is existing at once. 
+To enforce this, the client MUST use the `m.bookmarks_room` global account data event that contains a single `active_bookmarks_room_id` parameter with the Room ID of the bookmarks room. 
 
-## Room creation
+### Room creation
 
-The room should be created upon first bookmark creation.
-The client should check if a `m.bookmarks` room already exists before creating a new one.
+The room should be created upon first bookmark creation, and only if the global account data `m.bookmarks_room` event isn't set. When creating the room, the client must set the
+`m.bookmarks_room` account data event with the newly defined room ID.
+
+### Unbookmarking a message
+
+A message that is not bookmarked is a message that doesn't have a `m.bookmark` event referencing it in the `m.bookmarks` room. Thus, redacting a `m.bookmark` event is equivalent to not having its referenced event bookmarked.
+
+### Bookmark unicity
+
+The client should ensure that only one bookmark at the time exists for a given event and its versions. This assures that each message has a clean binary state: either bookmarked or not. This may be achieved by redacting the bookmark event. A message may be bookmarked again after being unbookmarked, by creating a new `m.bookmark` event referencing it.
+
+### Original message redaction / edition
+
+A bookmarked message may be redacted or edited by its sender. It is up to the client to respect the redaction of the referenced message, or resolve its latest version.
+
+### Bookmark event editions
+
+As all timeline events, the `m.bookmark` event may be replaced / edited through a `m.replace` relation. The client must use the latest version of the bookmark event, notably the referenced event of the `m.pointer` block.
 
 ## Potential issues
 
--- a/proposals/4482-ability-to-bookmark-messages.md
+++ b/proposals/4482-ability-to-bookmark-messages.md
@@ -16,7 +16,7 @@ This proposal leverages extensible timeline events and the fact that unknown roo
 ### New room type
 
 This MSC introduces a new room type called `m.bookmarks`.
-This kind of room SHOULD hold some `m.bookmark` events defined below. 
+This kind of room SHOULD only hold some `m.bookmark` events defined below. 
 
 ### Event structure
 
@@ -33,7 +33,8 @@ The pointer is provided as is: it may not be "valid" (i.e. the user does not hav
         "body": "You bookmarked matrix:r/matrix-dev:matrix.org in Matrix Dev"
       },
       {
-        { "body": "You bookmarked <a href='matrix:r/matrix-dev:matrix.org'>matrix:r/matrix-dev:matrix.org</a> in Matrix Dev", "mimetype": "text/html" },
+        "body": "You bookmarked <a href='matrix:r/matrix-dev:matrix.org'>matrix:r/matrix-dev:matrix.org</a> in Matrix Dev",
+        "mimetype": "text/html"
       }
     ],
     "m.pointer": {
--- a/proposals/4482-ability-to-bookmark-messages.md
+++ b/proposals/4482-ability-to-bookmark-messages.md
@@ -10,7 +10,19 @@ This feature can be extended for example to allow clients to add reminders after
 
 ## Proposal
 
+This proposal requires 2 additions to the spec : one new room type and one new event that would be sent in this new kind of room.
+This proposal leverages extensible timeline events and the fact that unknown room types are treated as regular rooms to provide a nice retrocompatibility experience with clients that do not implements this feature. 
+
+### New room type
+
+This MSC introduces a new room type called `m.bookmarks`.
+This kind of room SHOULD only hold some `m.bookmark` events defined below. 
+
 ### Event structure
+
+This MSC introduces a new kind of event which is a timeline event that uses extensible events blocks.
+It introduces a new `m.pointer` block that references a particular event sitting in a given room.
+The pointer is provided as is: it may not be "valid" (i.e. the user does not have access / permission to view the referenced event).
 
 ```json5
 {
@@ -29,9 +41,6 @@ This feature can be extended for example to allow clients to add reminders after
 }
 ```
 
-
-### New room type : `m.bookmark`
-
 ## Client implementation recommendations 
 
 ## General
@@ -45,13 +54,10 @@ If there are multiple `m.bookmark` rooms, we leave this as an implementation det
 The room should be created on first bookmark creation.
 The client should check if a `m.bookmark` room already exists before creating a new one.
 
-
-
 ## Potential issues
 
 - Room upgrades: could lead to bookmarks going missing if not handled properly. 
 - Multiple people: having multiple people in a room will lead to unintended consequences. 
-
 
 ## Alternatives
 
--- a/proposals/4482-ability-to-bookmark-messages.md
+++ b/proposals/4482-ability-to-bookmark-messages.md
@@ -30,7 +30,7 @@ The pointer is provided as is: it may not be "valid" (i.e. the user does not hav
   "content": {
     "m.text": [
       {
-        "body": "You bookmarked an image from "Alice" in Matrix Dev"
+        "body": "You bookmarked an image from Alice in Matrix Dev"
       },
       {
         "body": "You bookmarked an <a href='matrix:roomid/$ROOM_ID/e/$EVENT_ID?via=$SERVER_NAME'>image</a> from <a href='matrix:u/alice:example.org'>Alice</a> in Matrix Dev",
--- a/proposals/4482-ability-to-bookmark-messages.md
+++ b/proposals/4482-ability-to-bookmark-messages.md
@@ -30,17 +30,17 @@ The pointer is provided as is: it may not be "valid" (i.e. the user does not hav
   "content": {
     "m.text": [
       {
-        "body": "You bookmarked matrix:r/matrix-dev:matrix.org in Matrix Dev"
+        "body": "You bookmarked an image from "Alice" in Matrix Dev"
       },
       {
-        "body": "You bookmarked <a href='matrix:r/matrix-dev:matrix.org'>matrix:r/matrix-dev:matrix.org</a> in Matrix Dev",
+        "body": "You bookmarked an <a href='matrix:roomid/$ROOM_ID/e/$EVENT_ID?via=$SERVER_NAME'>image</a> from <a href='matrix:u/alice:example.org'>Alice</a> in Matrix Dev",
         "mimetype": "text/html"
       }
     ],
     "m.pointer": {
-      "room_id": "<room_id>",
-      "event_id": "<event_id>",
-      "via": ["<server_name>"]
+      "room_id": "$ROOM_ID",
+      "event_id": "$EVENT_ID",
+      "via": ["$SERVER_NAME"]
     
   } 
 }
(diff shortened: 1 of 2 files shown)


Gernomaly:
Signed-off-by: xeni <xeni@koeln.ccc.de>

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+
+## Potential issues
+
+- Room upgrades: could lead to bookmarks going missing if not handled properly. 
+- Multiple people: having multiple people in a room will lead to unintended consequences. 
+
+
+## Alternatives
```
turt2live, line 85: This MSC should compare itself to https://github.com/matrix-org/matrix-spec-proposals/pull/4438

turt2live (commented):
This MSC appears to be a work in progress, so I haven't done the normal "implementation requirements" comment yet. When ready, feel free to request such a comment in the [SCT Office](https://matrix.to/#/#sct-office:matrix.org).

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,84 @@
...
+    "m.pointer": {
+      "roomId": "<roomId>",
+      "eventId": "<eventId>"
+    }
+  } 
+}
+```
+
+
+### New room type : `m.bookmark`
```
IT-ess: Obviously we may need to complete this part. 
Also, I'm not sure about the `m.bookmark` naming anymore, since this should be the sole place where we store all of our bookmarks. What about `m.bookmarks` (plural) ?

IT-ess:
I completed the proposal part. Tell me if I'm too verbose @Gernomaly
We probably need to elaborate the last parts too.

Gernomaly:
Love the additions! I am going to elaborate on the alternatives parts later today.  Including the comparison that travis suggested 👍

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: has been added, tho there can probably be added a lot more.

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: I am a fan of using m.bookmarks!

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Proposal
+
+This proposal requires 2 additions to the spec : one new room type and one new event that would be sent in this new kind of room.
+This proposal leverages extensible timeline events and the fact that unknown room types are treated as regular rooms to provide a nice retrocompatibility experience with clients that do not implements this feature. 
+
+### New room type
+
+This MSC introduces a new room type called `m.bookmarks`.
+This kind of room SHOULD only hold some `m.bookmark` events defined below.
```
IT-ess, line 19: @Gernomaly I'm not sure anymore about this one.
As said in private, I want to adapt my Refs client so it supports this MSC. However, when I need to create a personal bookmark _ex-nihilo_ (not from an existing Matrix event), I need to send my custom Ref event somewhere, and then reference it with our new `m.bookmark` "pointer event".
Currently I'm using my own custom room type to store my custom events, so using just this `m.bookmarks` instead would be nice.
So my proposition is that instead of `SHOULD only` we put `MAY` ? Does that sound good ?

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+    ],
+    "m.pointer": {
+      "roomId": "<roomId>",
+      "eventId": "<eventId>"
+    }
+  } 
+}
+```
+
+The fallback text block of this event should contain at least the [Matrix URI](https://spec.matrix.org/v1.18/appendices/#matrix-uri-scheme) pointing to the bookmarked event. The rest of the body of this text block is left as an implementation detail of each client.
```
IT-ess, line 48: Now that I'm working on an implementation, I'm figuring out that detecting `matrix:` URIs in a plain text body is not that easy. 
I think additionally to the plain text part of this `m.text` block fallback, the bookmark sender should include a `org.matrix.custom.html` (rich text) that already wraps the Matrix URI so it is easy to detect. 
The rest may be left as implementation details. Thoughts @Gernomaly ?

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: i think should is fine, its not a must and may is too soft imo

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: ngl i am not knowledgeable enough on the implementation side of things to be able to give a useful opinion on this. Maybe some of the others have some thoughts. Ill send it into the matrix room

evlli: fallback representations don't need to be parsed because they're fallback for clients who don't support the event itself, no?

having those clients provide a clickable link for URI's is admittedly a hard problem.

`m.text` already has everything we need to support html representations though.
as defined in the extensible events MSC https://github.com/matrix-org/matrix-spec-proposals/pull/1767/changes#diff-242c41793e638a8acaf2b7ea5870ca0249de76f00b36e8fb2c2e5ed4e106872aR69-R129


``` json
"m.text": [
  { "body": "<a href='matrix://drachen-einfach.de'>matrix://drachen-einfach.de</a>", "mimetype": "text/html" },
  { "body": "matrix://drachen-einfach.de" }
]
```

we should probably explicitly reference this here, as people might not be familiar with extensible events

Gernomaly: sounds good to me @evlli

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+  "type": "m.bookmark",
+  "content": {
+    "m.text": [
+      {
+        "body": "You bookmarked <matrix_uri> in <room_name>" // Example fallback text
+      }
+    ],
+    "m.pointer": {
+      "roomId": "<roomId>",
+      "eventId": "<eventId>"
```
Johennes: I couldn't find it spelled out in the spec but I think most / all endpoints in Matrix use snake case for JSON properties (e.g. `m.relates_to` rather than `m.relatesTo`).

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+{
+  "type": "m.bookmark",
+  "content": {
+    "m.text": [
+      {
+        "body": "You bookmarked <matrix_uri> in <room_name>" // Example fallback text
+      }
+    ],
+    "m.pointer": {
+      "roomId": "<roomId>",
```
Johennes: Since room IDs alone are not routable in Matrix, would you also need a `via` property possibly? I assume you'd normally already be in the room so maybe not? 🤔

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+The fallback text block of this event should contain at least the [Matrix URI](https://spec.matrix.org/v1.18/appendices/#matrix-uri-scheme) pointing to the bookmarked event. The rest of the body of this text block is left as an implementation detail of each client. 
+
+## Client implementation recommendations 
+
+## General
+
+If the client support the `m.bookmarks` room type, it should'nt be displayed as a normal room in the regular room list. Instead the client could display bookmarks in dedicated sections of the client. 
+The client should ensure that only one `m.bookmarks` room is existing at once.
```
Johennes: Could an account data event help to let clients store the room ID of _the one_ bookmarks room?

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,123 @@
...
+If there are multiple `m.bookmarks` rooms, we leave this as an implementation detail.
+
+## Room creation
+
+The room should be created upon first bookmark creation.
```
Johennes: This feels like it could be an implementation detail. Would there be any issues with letting clients create the room prematurely?

proposals/4482-ability-to-bookmark-messages.md
IT-ess: Yeah that's what I meant, that the fallback `m.text` representation should be "complete", not only the plain text representation but also the html one. 
Your proposition looks good to me too !

proposals/4482-ability-to-bookmark-messages.md
IT-ess: Our approach of pointers is kinda simple, we are ok if some pointers are just "dead". It should be up to the client to eventually clean up dead pointers (bookmarks pointing to rooms you're no longer in for instance). 

Also, I guess if you don't clean a bookmark to a public room you're no longer part of, having the `via` parameter could help you joining this room back, so I have mixed opinions on this.

proposals/4482-ability-to-bookmark-messages.md
IT-ess: We imagined that but it would probably require extra work if we need to upgrade the `m.bookmarks` room at some point.

proposals/4482-ability-to-bookmark-messages.md
IT-ess: When we discussed it IRL I remember two arguments : 
- waste of ressources
- zero click creation of a room that contains nothing yet, so it could be weird for a user that also uses a client that does not support `m.bookmarks` rooms, and thus display the fallback room containing nothing. 

We're aware that the first bookmark creation should be a little longer, but that is probably up to the client to make it smooth.

proposals/4482-ability-to-bookmark-messages.md
IT-ess: Is `SHOULD` without `only` ok for you ?

Gernomaly: id just stay with should only tbh, its not like the should only prevents you of not adding anything else imo, and from my point of view your events are a kind of bookmark as well :P . To me it its important to at least for the moment keep the scope together a bit and adapt it if needed in the future if the protocol develops into that direction (referencing msc guide)

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: thats true, we should do that!

proposals/4482-ability-to-bookmark-messages.md
IT-ess: Ok I will put back the `only` then. You're right we should keep it scoped to the generic usecase for now

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: @Johennes is that a sufficient answer to that question or do  you still think we should change it?

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,128 @@
...
+        "body": "You bookmarked an image from Alice in Matrix Dev"
+      },
+      {
+        "body": "You bookmarked an <a href='matrix:roomid/$ROOM_ID/e/$EVENT_ID?via=$SERVER_NAME'>image</a> from <a href='matrix:u/alice:example.org'>Alice</a> in Matrix Dev",
+        "mimetype": "text/html"
+      }
+    ],
+    "m.pointer": {
+      "room_id": "$ROOM_ID",
+      "event_id": "$EVENT_ID",
```
IT-ess, line 42: I think we should add the constraint to the value of this event_id, that it should always be a "root" event_id in case the original event has been (or will be) replaced by another one (`m.replace` relation).
It is then the client's role to resolve or update the content of the bookmark taking into account the eventual edits.

proposals/4482-ability-to-bookmark-messages.md
evlli: > Since room IDs alone are not routable in Matrix, would you also need a `via` property possibly? I assume you'd normally already be in the room so maybe not? 🤔

you should be in the room, or at the very least join via the server of the user who created the bookmark event. if their homeserver isn't participating in that room anymore, any via's you'd have set would also likely be useless.

if you really want to, we could use the most popular server by users that's in that room, hence having a decent chance of still being part of that room?

I don't particularly care either way, for my use case it's not necessary but I'm open to adding it.

proposals/4482-ability-to-bookmark-messages.md
evlli: I agree, feel free to add this

Gernomaly: yeah

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,128 @@
...
+      "via": ["$SERVER_NAME"]
+    
+  } 
+}
+```
+
```
IT-ess, line 47: Suggestion: Add the fact that each message of a room should be bookmarked at most once (maximum one `m.bookmark` event pointing to an event). This ensures that we have a binary state for each message: either bookmarked or not.

proposals/4482-ability-to-bookmark-messages.md
```diff
@@ -0,0 +1,128 @@
...
+        "body": "You bookmarked an <a href='matrix:roomid/$ROOM_ID/e/$EVENT_ID?via=$SERVER_NAME'>image</a> from <a href='matrix:u/alice:example.org'>Alice</a> in Matrix Dev",
+        "mimetype": "text/html"
+      }
+    ],
+    "m.pointer": {
+      "room_id": "$ROOM_ID",
+      "event_id": "$EVENT_ID",
+      "via": ["$SERVER_NAME"]
+    
+  }
```
IT-ess: Suggestion: add a `m.relates_to` block on this event so we can handle editing the event with `m.replace` relations. This isn't very useful as is because editing the pointer or the fallback text isn't very useful, but since it is an extensible event, one may want to add extra blocks to this bookmark. We want to have a future proof event.

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: agreed tho i dont really know how we should implement this

proposals/4482-ability-to-bookmark-messages.md
Gernomaly: sounds reasonable to me!

proposals/4482-ability-to-bookmark-messages.md
IT-ess: I already [started](https://github.com/IT-ess/matrix-rust-sdk/tree/msc4482-bookmarks) implementing it on the rust-sdk if you're curious. That's why I'm coming with all these questions and precisions. 
Implementation is not that obvious tbh.

stefanceriu:
I'm sorry I didn't notice this sooner but you should have a look at [MSC2437: Store tagged events in Room Account Data](https://github.com/matrix-org/matrix-spec-proposals/pull/2437) too:
> The first use case would be to let him mark some events as favorites to keep a reference on important messages or attachments. Clients would then be able to handle a global bookmark of these favourite events, or display a list of them at the room level.

IT-ess:
> I'm sorry I didn't notice this sooner but you should have a look at [MSC2437: Store tagged events in Room Account Data](https://github.com/matrix-org/matrix-spec-proposals/pull/2437) too:
> 
> > The first use case would be to let him mark some events as favorites to keep a reference on important messages or attachments. Clients would then be able to handle a global bookmark of these favourite events, or display a list of them at the room level.

This is a different approach. We describe why we think it is a better way from line 108, but here are the main points: 
- Account data is fetched in one go, we don't have a way to paginate it. When we store this kind of data in account data, we push data to clients that don't even need it. 
- Our approach provides a fallback "free" implementation for clients that implements Extensible Events. I admit they are few right now, but it will improve over time.
- This implementation is way more extensible, thanks to extensible events again. One could imagine adding additional blocks to the bookmarks to implement folders for instance. 

Let us know if you think this different approach is valuable enough ! @stefanceriu

stefanceriu:
I mostly wanted to link the two together. We [prototyped message bookmarks](https://www.youtube.com/watch?v=D5zAgVYBuGk&t=998s) ourselves in one of our hackatons (based on [bnjbvr/bookmarks](https://github.com/element-hq/matrix-rust-sdk/tree/bnjbvr/bookmarks)) in which we  decided to go against #2437 and store everything in the account data, as it made building the UI we wanted easier to build.

Your solution on the other hand would take care of both paths but I do wonder if it is a specialisation of message forwarding i.e. something like #2723 in which `m.forwarded` is your `m.pointer`.

IT-ess:
I missed that (great) demo, thank you for the link. I was aware of this other MSC implementing things through account data, but I never saw it landing in Element X (I don't have access to the branch you mentioned), so my guess was that no effort was made in the rust sdk to implement this already. 
For the record, I've made an implementation of this MSC for the Rust SDK [here](https://github.com/IT-ess/matrix-rust-sdk/pull/1/changes), there is also a small video demo linked. I may open smaller PRs to the upstream, but I'm not sure if its required or necessary since there is already a concurrent implementation :shrug: 

About the `m.forwarded` block, they are indeed very similar except we don't require the sender and timestamp in `m.pointer`. We could eventually reuse the same block, but I'm afraid that in terms of naming `m.forwarded` feels pretty specific related to `m.pointer`.

Gernomaly:
> About the `m.forwarded` block, they are indeed very similar except we don't require the sender and timestamp in `m.pointer`. We could eventually reuse the same block, but I'm afraid that in terms of naming `m.forwarded` feels pretty specific related to `m.pointer`.

It might actually be an idea to think about as a protocol if we should resuse "m.pointer" for "m.forwarded" in the future

PR #4484, opened 2026-05-27
MSC4484: Server Administration OAuth Scope

gingershaped:
[Rendered](https://github.com/gingershaped/matrix-spec-proposals/blob/ginger/msc/admin-oauth-scope/proposals/4484-admin-oauth-scope.md)

Signed-off-by: Ginger <ginger@gingershaped.computer>

### DIFF
--- a/proposals/4484-admin-oauth-scope.md
+++ b/proposals/4484-admin-oauth-scope.md
@@ -17,7 +17,7 @@ ensuring the user can give informed consent to clients that wish to perform admi
 ## Proposal
 
 A new scope is added to the specification, which clients using OAuth MAY request:
-`urn:matrix:client:api:server_administration` (herein referred to as "the scope" for brevity). Clients using OAuth
+`urn:matrix:client:server_administration` (herein referred to as "the scope" for brevity). Clients using OAuth
 MUST request the scope to use any of the endpoints in the Server Administration module (all of which are currently
 under the `/_matrix/client/*/admin/` path).
 
--- a/proposals/4484-admin-oauth-scope.md
+++ b/proposals/4484-admin-oauth-scope.md
@@ -86,3 +86,7 @@ While this proposal is unstable, clients and servers should use the
 `urn:matrix:client:api:server_administration` scope. Servers may use the `org.continuwuity.msc4484.unstable`
 unstable feature flag to indicate that clients using OAuth must request the unstable
 `urn:matrix:client:cc.c10y.msc4484.server_administration` scope to use the Server Administration endpoints.
+
+## Dependencies
+
+None.
--- a/proposals/4484-admin-oauth-scope.md
+++ b/proposals/4484-admin-oauth-scope.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Server Administration OAuth Scope
+# MSC4484: Server Administration OAuth Scope
 
 Since Matrix 1.18, the [Server Administration](https://spec.matrix.org/v1.18/client-server-api/#server-administration)
 API module has provided homeserver-agnostic endpoints to manage suspending and locking user accounts. These endpoints
@@ -82,7 +82,7 @@ user's extant OAuth sessions that have the scope.
 ## Unstable prefix
 
 While this proposal is unstable, clients and servers should use the
-`urn:matrix:client:cc.c10y.mscXXXX.server_administration` scope instead of the
-`urn:matrix:client:api:server_administration` scope. Servers may use the `org.continuwuity.mscXXXX.unstable`
+`urn:matrix:client:cc.c10y.msc4484.server_administration` scope instead of the
+`urn:matrix:client:api:server_administration` scope. Servers may use the `org.continuwuity.msc4484.unstable`
 unstable feature flag to indicate that clients using OAuth must request the unstable
-`urn:matrix:client:cc.c10y.mscXXXX.server_administration` scope to use the Server Administration endpoints.
+`urn:matrix:client:cc.c10y.msc4484.server_administration` scope to use the Server Administration endpoints.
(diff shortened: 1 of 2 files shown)


proposals/4484-admin-oauth-scope.md
turt2live, line 1: Implementation requirements:

* [ ] Server (supporting)
* [ ] Client (using)

proposals/4484-admin-oauth-scope.md
```diff
@@ -0,0 +1,88 @@
...
+Clients using user-interactive authentication SHOULD be handled by the server as before. The server MAY restrict
+administration endpoints to only clients using OAuth, and return a `M_FORBIDDEN` error code to clients using
+user-interactive authentication.
+
+## Potential issues
+
+This MSC is explicitly not backwards-compatible. Existing clients using OAuth will no longer be able to use the
+admin API endpoints unless and until they begin requesting the scope. To the best of the author's knowledge, there
+are no clients in widespread use that support OAuth _and_ the Server Administration endpoints, and therefore the
+impact of this change is currently minimal.
```
tulir, line 63: Doesn't Synapse/MAS already require a special scope for all admin endpoints?

gingershaped: It does, but I don't know if that scope is only for the Synapse-specific admin endpoints or if it also includes the specced admin endpoints.

proposals/4484-admin-oauth-scope.md
```diff
@@ -0,0 +1,88 @@
...
+`urn:matrix:client:api:*` scope, which grants full access to the entire client-server API. Therefore, all OAuth
+2.0 clients will be granted access to the endpoints in the Server Administration module by default, potentially
+including more dangerous actions as this module is extended. This MSC improves these endpoints' security by
+introducing a new OAuth scope that is REQUIRED for OAuth 2.0 clients to use administration endpoints, thereby
+ensuring the user can give informed consent to clients that wish to perform administrative actions.
+
+## Proposal
+
+A new scope is added to the specification, which clients using OAuth MAY request:
+`urn:matrix:client:api:server_administration` (herein referred to as "the scope" for brevity). Clients using OAuth
```
tonkku107: `urn:matrix:client:api:*` implies it includes everything in that namespace, so if you were to introduce a new scope that is _not_ to be granted by that scope, it should be different at a higher level (like `urn:matrix:client:admin` or `urn:matrix:admin` for example).

Worth considering if it should also be defined with a wildcard in the end (e.g. `urn:matrix:admin:*`) for more fine-grained control in the future like is planned for client-server API

proposals/4484-admin-oauth-scope.md
tulir: I'd assume it includes the specced ones, but someone should check. If it does, then this potential issue doesn't really exist in the first place (and the MSC may even already be implemented with a different prefix?)

proposals/4484-admin-oauth-scope.md
gingershaped: > `urn:matrix:client:api:*` implies it includes everything in that namespace, so if you were to introduce a new scope that is _not_ to be granted by that scope, it should be different at a higher level (like `urn:matrix:client:admin` or `urn:matrix:admin` for example).

Good point. I've changed the scope to `urn:matrix:client:server_administration`.

> Worth considering if it should also be defined with a wildcard in the end (e.g. `urn:matrix:admin:*`) for more fine-grained control in the future like is planned for client-server API

It seems unlikely to me on the face of it that this will happen -- I would assume that clients generally either want full admin access or no admin access.

proposals/4484-admin-oauth-scope.md
gingershaped: It appears that it does include the specced ones ([logic checking if a user is an admin](https://github.com/element-hq/synapse/blob/f9f08db4c3dab5bec742ea2476ed815806c584c8/synapse/api/auth/mas.py#L274), [call to that logic in the whois endpoint](https://github.com/element-hq/synapse/blob/f9f08db4c3dab5bec742ea2476ed815806c584c8/synapse/rest/admin/users.py#L742)). However, the `urn:synapse:admin:*` scope [also requires the `urn:matrix:client:api:*` scope](https://element-hq.github.io/matrix-authentication-service/reference/scopes.html#urnsynapseadmin), so it doesn't behave the same as the scope described in the proposal.

proposals/4484-admin-oauth-scope.md
```diff
@@ -0,0 +1,92 @@
...
+defined by the server. If the authorizing user is not a server administrator, the server MUST respond to the
+authorization request or device authorization request (depending on the grant being used) with an `access_denied`
+error, as described in [RFC 6749 § 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1).
+
+The server SHOULD, when a client requests the scope, indicate clearly to the user the dangers of granting a client
+this scope in its grant confirmation webpage.
+
+Clients intended for general-purpose use, as opposed to specifically administrative tasks, SHOULD NOT request the
+scope during normal use. If the user wishes to perform an administrative action, clients SHOULD attempt to obtain
+a new access token with the scope.
```
Johennes, line 52: Not sure if worth mentioning but this might in future be aided by https://github.com/matrix-org/matrix-spec-proposals/pull/4363 or a similar approach.

proposals/4484-admin-oauth-scope.md
```diff
@@ -0,0 +1,92 @@
...
+`urn:matrix:client:api:*` scope, which grants full access to the entire client-server API. Therefore, all OAuth
+2.0 clients will be granted access to the endpoints in the Server Administration module by default, potentially
+including more dangerous actions as this module is extended. This MSC improves these endpoints' security by
+introducing a new OAuth scope that is REQUIRED for OAuth 2.0 clients to use administration endpoints, thereby
+ensuring the user can give informed consent to clients that wish to perform administrative actions.
+
+## Proposal
+
+A new scope is added to the specification, which clients using OAuth MAY request:
+`urn:matrix:client:server_administration` (herein referred to as "the scope" for brevity). Clients using OAuth
```
Johennes, line 20: I wonder if it should be `urn:matrix:client:server_administration:*` to allow for potential later subdivision? Or maybe just `urn:matrix:client:admin:*` to limit the length a bit.

PR #4402, opened 2026-01-15
MSC4402: Consistent redirects for .well-known-files

derhagen:
[Rendered](https://github.com/derhagen/matrix-spec-proposals/blob/main/proposals/4402-consistent-redirects-for-well-known-files.md)

This MSC fixes https://github.com/matrix-org/matrix-spec/issues/445

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4402#issuecomment-4114176145)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4402#issuecomment-4114176916)

PR #4485, opened 2026-06-02
MSC4485: Homeserver feature controls

JadedBlueEyes:
[Rendered](https://github.com/JadedBlueEyes/matrix-spec-proposals/blob/jade/featured/proposals/4485-featured-endpoint.md)

### DIFF
--- a/proposals/4485-featured-endpoint.md
+++ b/proposals/4485-featured-endpoint.md
@@ -1,70 +1,114 @@
 # MSC4485: Homeserver feature controls
 
-Currently 'featured homeserver lists' are gaining more attention as community members think about user onboarding. These would direct new users to homeservers curated to provide a good experience. However, many homeservers don't have the ability to accommodate any rate of growth, or a global audience. This proposal provides a way to give server admins more responsive control over their visibility on such lists.
+Currently 'featured homeserver lists' are gaining more attention as community
+members think about user onboarding. These would direct new users to homeservers
+curated to provide a good experience. However, many homeservers don't have
+the ability to accommodate any rate of growth, or a global audience. This
+proposal provides a way to give server admins more responsive control over their
+visibility on such lists.
 
 ## Proposal
 
-
-This proposal adds a new well-known endpoint, `.well-known/matrix/featured`. This endpoint responds with the following fields:
+This proposal adds a new well-known endpoint, `.well-known/matrix/featured`.
+This endpoint responds with the following fields:
 
 ```jsonc
 {
-    "display": 0.5, // A number from 0 to 1 defining the probability that this homeserver should be shown in a list, with 0 being never and 1 being always. 
+    "display": 0.5, // A number from 0 to 1 defining the probability that this homeserver should be shown in a list, with 0 being never and 1 being always.
     "languages": { // optional
-        "allow": ["en", "en-US", "fr-FR"], // Optional. Language codes that the server should be featured in. If present, should not be shown for other languages. 
+        "allow": ["en", "en-US", "fr-FR"], // Optional. Language codes that the server should be featured in. If present, should not be shown for other languages.
         "deny": ["es"], // Optional. Language codes that the server should *not* be featured in.
     },
     "geos": { // optional
-        "allow": ["GB"], // Optional. Country codes that the server should be featured in. If present, should not be shown for other countries. 
+        "allow": ["GB"], // Optional. Country codes that the server should be featured in. If present, should not be shown for other countries.
         "deny": ["CN"], // Optional. Country codes that the server should *not* be featured in.
-    }
+    },
 }
 ```
 
-Where language codes are specified, BCP 47 tags MUST be used. Where country codes are specified, ISO 3166-1 alpha-2 codes MUST be used.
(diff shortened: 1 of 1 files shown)


proposals/4485-featured-endpoint.md
```diff
@@ -0,0 +1,71 @@
+# MSC4485: Homeserver feature controls
+
+Currently 'featured homeserver lists' are gaining more attention as community members think about user onboarding. These would direct new users to homeservers curated to provide a good experience. However, many homeservers don't have the ability to accommodate any rate of growth, or a global audience. This proposal provides a way to give server admins more responsive control over their visibility on such lists.
```
turt2live: Please aim to wrap lines at around 100 characters. It doesn't need to be precise - just enough to ensure that if someone leaves a comment and the diff changes, the comment has a good chance of remaining visible in the github UI.

turt2live (commented):
Early editorial review as part of determining labels - the actual content hasn't been reviewed in reasonable detail.

proposals/4485-featured-endpoint.md
turt2live, line 1: Implementation requirements:

* [ ] Server (offering)
* [ ] Application (using) - ideally multiple

CrazyNicc:
Should 30x redirects be possible on this endpoint? If so, please clarify in the proposal

CrazyNicc (commented):
Should 30x redirects be possible on this endpoint?

proposals/4485-featured-endpoint.md
CrazyNicc, line 1: Should 30x redirects be possible on this endpoint? (3rd time commenting this because I wanted this to be a thread)

JadedBlueEyes: It should be handled in the same way as the other client-server well known endpoints. The spec doesn't specify, but the server-server one must follow redirects

JadedBlueEyes: Side note: looks like it's about to get spec'd https://github.com/matrix-org/matrix-spec-proposals/pull/4402

proposals/4485-featured-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+## Potential issues
+
+Because the algorithm includes randomness, reevaluating the algorithm on
+each page load may result in a poor user experience as servers disappear
+and reappear. Implementations should seed their random number generation
+appropriately to ensure the list remains reasonably stable.
+
+Methods for determining a user's geographic region and language often run into
+cases where they do not get an accurate or useful result. Because of this,
+clients may want to allow manual selection of these values.
```
Johennes, line 78: This seems to conflict with the following requirement?

>     - If the user's geolocation is present in the `deny` property, it MUST NOT display
>       the server.

proposals/4485-featured-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+      it SHOULD NOT display the server.
+- Otherwise, it SHOULD display the server.
+
+
+## Potential issues
+
+Because the algorithm includes randomness, reevaluating the algorithm on
+each page load may result in a poor user experience as servers disappear
+and reappear. Implementations should seed their random number generation
+appropriately to ensure the list remains reasonably stable.
```
Johennes, line 74: Does seeding appropriately mean using a constant seed?

proposals/4485-featured-endpoint.md
```diff
@@ -0,0 +1,115 @@
...
+would make the proposal less useful for servers that want to maintain a specific
+rate of growth.
+
+## Security considerations
+
+If a client fetches the rules by itself for evaluation, it leaks technical data
+like the IP address to (a potentially large number of) third party servers.
+For this reason, and to reduce load on servers, it is heavily recommended that
+implementations proxy these rules, potentially bundling them with the server
+list itself.
```
Johennes, line 106: Does this mean bundling the server list and well-known responses in the client? Or do you mean creating a separate service that hosts them for the client to load?

proposals/4485-featured-endpoint.md
JadedBlueEyes: Not particularly, that paragraph just hints that implementations may want to allow the user to declare what their geolocation is. Which is mildly more convenient than using a VPN or other location spoofing.

proposals/4485-featured-endpoint.md
JadedBlueEyes: It probably means persisting the seed in the session, or similar.

proposals/4485-featured-endpoint.md
JadedBlueEyes: A client might have a server list service that returns a response containing a list of vetted servers with each server's respective rules. A stand-alone implementation might build it into the webpage itself.

This explicitly does *not* mean bundling the responses at compile time (hence proxy), because otherwise servers would no longer be able to update rules on the fly, in response to load or other conditions.

proposals/4485-featured-endpoint.md
Johennes: Ah, got it. Thank you.

PR #4488, opened 2026-06-05
Fix `height` definition on MSC2545: Image Packs

turt2live:
Also remove stray whitespace my editor didn't like.

The [`img` height](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/height) is an *integer*, not a CSS-style string, so must be a value like `32`. This matches the examples throughout the proposal. 

Some implementations do send `height="32px"` as the MSC recommended, however this is incorrect behaviour and may lead to clients/browsers ignoring the `height` attribute entirely. Clients which are sending this invalid value SHOULD NOT continue doing so, and instead SHOULD use `"32"` instead. 

**Process note**: I've added a warning to https://github.com/matrix-org/matrix-spec-proposals/pull/2545 to highlight that this PR changes it a bit.

### DIFF
--- a/proposals/2545-emotes.md
+++ b/proposals/2545-emotes.md
@@ -136,10 +136,10 @@ An example of a pack object:
 An image object consists of the following keys:
 
 - `url`: **Required, String**. The MXC URL for this image.
-- `body`: **Optional, String**. A textual representation or associated 
+- `body`: **Optional, String**. A textual representation or associated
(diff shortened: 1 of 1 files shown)


anoadragon453 (approved):
Aha, I see. I was confusing it with CSS (which is happy to have a `px` suffix). Thanks!

PR #4468, opened 2026-05-12
MSC4468: Reporting to communities (via to-device)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/reporting-v2/x-report-to-community/proposals/4468-report-to-community.md)

*Note*: This is a draft while it undergoes very early review. The idea needs validation prior to any implementation begins.

Alternatives:
* https://github.com/matrix-org/matrix-spec-proposals/pull/2938
* https://github.com/matrix-org/matrix-spec-proposals/pull/4226
* Parts of https://github.com/matrix-org/matrix-spec-proposals/pull/3215

PR #4121, opened 2024-03-22
MSC4121: `m.role.moderator` `/.well-known/matrix/support` role.

FSG-Cat:
[Rendered](https://github.com/FSG-Cat/matrix-spec-proposals/blob/FSG-Cat-Moderation-Role-well-known-support-record/proposals/4121-m.role.moderator.md)

This MSC gives homeservers the ability to split administration and moderation in their support records.

Signed-off-by: Catalan Lover <catalanlover@protonmail.com>

PR #4489, opened 2026-06-05
MSC4489: Per-room contact state event

ll-SKY-ll:
Signed-off-by: Sky sky@codestorm.net

[rendered](https://github.com/ll-SKY-ll/matrix-spec-proposals/blob/a75746a30fb3912a5e46dbe2df3f120f6a3c571f/proposals/4489-per-room_contact_state_event.md)

### DIFF
--- a/proposals/4489-per-room_contact_state_event.md
+++ b/proposals/4489-per-room_contact_state_event.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Per-room contact state event
+# MSC4489: Per-room contact state event
 
 A room currently has no standard, machine-readable way to say *who to contact
 about it*. If you are confused about a room's rules, want to reach a moderator,
@@ -202,7 +202,7 @@ consumers to treat it as such.
 
 | Proposed (stable)            | Unstable                            |
 |------------------------------|-------------------------------------|
-| `m.room.contact` (event type) | `net.codestorm.mscXXXX.contact`    |
+| `m.room.contact` (event type) | `net.codestorm.msc4489.contact`    |
 
 The `role` values reuse the existing stable `m.role.*` identifiers from MSC1929
 and are unaffected by this proposal's unstable prefix.
--- a/proposals/4489-per-room_contact_state_event.md
+++ b/proposals/4489-per-room_contact_state_event.md
@@ -6,7 +6,7 @@ or have been banned by a policy list and want to appeal, there is nowhere to
 look: you are left guessing from the member list and power levels, neither of
 which was designed to answer the question "who do I talk to?". Matrix already
 solves the equivalent problem at the *homeserver* level with
-[`/.well-known/matrix/support`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixsupport)
+[`/.well-known/matrix/support`](https://spec.matrix.org/v1.18/client-server-api/#getwell-knownmatrixsupport)
 (originally [MSC1929](https://github.com/matrix-org/matrix-spec-proposals/pull/1929)),
 which lets a server advertise its administrative and security contacts. This
 proposal brings the same idea to rooms.
--- a/proposals/xxxx-per-room_contact_state_event
+++ b/proposals/xxxx-per-room_contact_state_event
@@ -1,0 +1,212 @@
+# MSCXXXX: Per-room contact state event
+
+A room currently has no standard, machine-readable way to say *who to contact
+about it*. If you are confused about a room's rules, want to reach a moderator,
+or have been banned by a policy list and want to appeal, there is nowhere to
+look: you are left guessing from the member list and power levels, neither of
+which was designed to answer the question "who do I talk to?". Matrix already
+solves the equivalent problem at the *homeserver* level with
+[`/.well-known/matrix/support`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixsupport)
+(originally [MSC1929](https://github.com/matrix-org/matrix-spec-proposals/pull/1929)),
+which lets a server advertise its administrative and security contacts. This
+proposal brings the same idea to rooms.
+
+This proposal introduces an `m.room.contact` state event that lets a room
+advertise its responsible contacts and an optional support page, reusing the
+data shape of the homeserver support document so that existing tooling can be
+adapted with minimal effort.
+
+## Proposal
+
+The set of people responsible for a room and/or a policy list is not reliably
+discoverable from the information available today (usually power levels):
+
+- Moderators frequently act through shared **moderation bot accounts** (e.g.
+  Draupnir, Meowlnir) rather than their personal accounts. The bot may hold a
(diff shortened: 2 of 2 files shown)


proposals/4489-per-room_contact_state_event.md
turt2live, line 1: Implementation requirements:

* [ ] Client (setting)
* [ ] Client (reading/using)

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
+# MSC4489: Per-room contact state event
+
+A room currently has no standard, machine-readable way to say *who to contact
+about it*. If you are confused about a room's rules, want to reach a moderator,
+or have been banned by a policy list and want to appeal, there is nowhere to
+look: you are left guessing from the member list and power levels, neither of
+which was designed to answer the question "who do I talk to?". Matrix already
+solves the equivalent problem at the *homeserver* level with
+[`/.well-known/matrix/support`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixsupport)
```
turt2live: (comment applies throughout)

Using `latest` can cause links to fail or become inaccurate as time goes on, especially after an MSC is accepted. Using a tagged version avoids these problems.

```suggestion
[`/.well-known/matrix/support`](https://spec.matrix.org/v1.18/client-server-api/#getwell-knownmatrixsupport)
```

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+  granularity improves automatically as that set grows, with no change to this
+  event.
+- **Discoverability before joining.** A user who cannot read room state (e.g.
+  has been banned, or has never been a member) cannot read this event through
+  normal means. Exposing it on a failed join or to non-members is deliberately
+  out of scope for this MSC and is left to a follow-up proposal, so that the
+  state event — which is independently useful to existing members and to bots
+  that are in the room — is not blocked on federation/client changes.
+
+## Alternatives
```
turt2live, line 159: Specifically for appeals, the Reporting v2 thought is currently https://github.com/matrix-org/matrix-spec-proposals/pull/4468 fwiw

turt2live (commented):
Early partial review of MSC

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+  etc.) programmatically resolve a contact without out-of-band coordination or
+  guessing from the room topic.
+
+### Event definition
+
+A new state event type `m.room.contact` is defined. It uses the empty string
+`""` as its state key; a room has at most one such event. It is ordinary room
+state and requires no new room version.
+
+Its `content` reuses the `contacts` and `support_page` fields, with the same
```
thetayloredman, line 58: As a room admin, I would be worried about these values becoming stale and needing to roll out new values to all of my rooms whenever these change. I wonder about the potential merits of a `"support_via": "matrix.org"` field that would "redirect" users to the `/.well-known/matrix/support` of a given server name?

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+contact fields, or new `m.role.*` values) apply here without requiring changes
+to this event.
+
+```json
+{
+  "contacts": [
+    {
+      "matrix_id": "@alice:example.org",
+      "email_address": "security@example.org",
+      "role": "m.role.security"
```
thetayloredman, line 71: What would a security contact on a room actually be responsible for?

proposals/4489-per-room_contact_state_event.md
ll-SKY-ll: i dont understand the issue. In that case you might wish to just populate the `support_page` and nothing else

proposals/4489-per-room_contact_state_event.md
ll-SKY-ll: The examples were made based on the currently available options in the `m.role.*` namespace, they don't necessarily all have a real world use case today, though that doesnt mean there wont be someone finding a use for it eventually. This proposal also hopes to inspire new proposals to expand the options available to the `m.role.*` namespace like [MSC4121](https://github.com/matrix-org/matrix-spec-proposals/pull/4121) already does.

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+
+- Clients render a "who runs this room / who do I contact for X" affordance,
+  independent of the member list and of the power levels event.
+- Bots consuming policy lists (Mjolnir/Draupnir/Meowlnir, ban-list aggregators,
+  etc.) programmatically resolve a contact without out-of-band coordination or
+  guessing from the room topic.
+
+### Event definition
+
+A new state event type `m.room.contact` is defined. It uses the empty string
```
Johennes, line 54: Low temperature: Since this is meant to be complementary to `.well-known/matrix/support`, maybe it could be `m.room.support` to keep the terminology on par?

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+      "email_address": "appeals@example.org",
+      "role": "m.role.admin"
+    }
+  ]
+}
+```
+
+The following constraints apply:
+
+- At least one of `contacts` or `support_page` SHOULD be present.
```
Johennes, line 98: This would allow sending the event with neither which feels slightly pointless. Maybe using the event itself is a MAY but if you use it, you MUST fill in either `contacts` or `support_page`. That would also match the rules for `/.well-known/matrix/support`.

thetayloredman: State events cannot be deleted, so if a room desires to no longer list contacts, an empty object is the best manner of doing so.

proposals/4489-per-room_contact_state_event.md
```diff
@@ -0,0 +1,212 @@
...
+- Within a contact, at least one of `matrix_id` or `email_address` MUST be
+  present; a contact with neither MUST be ignored by consumers.
+
+#### Roles
+
+A contact's `role` is a value from the same `m.role.*` set used by
+[`/.well-known/matrix/support`](https://spec.matrix.org/latest/client-server-api/#getwell-knownmatrixsupport);
+this proposal does not define its own roles. The set is open: consumers MUST
+gracefully handle role values they do not recognise (for example, by displaying
+the contact with no role label), and new `m.role.*` values defined by other
```
Johennes, line 108: Maybe this should be generalised to unknown roles that follow the common namespaced identifier syntax (as in `/.well-known/matrix/support`). MSC implementations and custom roles won't use the `m.` prefix.

proposals/4489-per-room_contact_state_event.md
Johennes: Ah, true. Maybe this could be clarified in the proposal.

PR #4495, opened 2026-06-25
MSC4495: Selective Presence

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/presence-v2/selective/proposals/4495-selective-presence.md)

Implementations:
- Ruma: https://github.com/ruma/ruma/pull/2546
- Client: none
- Server:
  - [Continuwuity](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/2034)

This proposal is written as part of a community-led effort called [Presence v2](https://ispresencefixedyet.com). Context and prior discussion can be found in [#presence-v2:zirco.dev](https://matrix.to/#/#presence-v2:zirco.dev) and in the [initial conversation that gave rise to this initiative](https://matrix.to/#/!nD4Jy1hp0We0VmIM9ubjqWLBX_uV8YlTBBPa3a_v2uk/%24pky9awpkMJUWeJPhTjcPaBZ7EPtAQ6rlKWUs9nCx66s?via=zirco.dev&via=matrix.org&via=element.io).

Author: @thetayloredman @lveneris

### DIFF
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Selective Presence
+# MSC4495: Selective Presence
 
 Presence is widely known to be one of the most expensive features in the Matrix specification. Even with continued
 efforts to improve server implementations[^1][^2], large public deployments still disable outbound federated
@@ -382,21 +382,21 @@ The following complementary proposals are planned:
 
 | Stable identifier         | Purpose                                                            | Unstable identifier                                          |
 | ------------------------- | ------------------------------------------------------------------ | ------------------------------------------------------------ |
-| `recipients`              | Recipient users to add and delete in federation presence EDU       | `org.continuwuity.presence_v2.mscXXXX.recipients`            |
-| `stream_id`               | Current recipient user set identifier federation presence EDU      | `org.continuwuity.presence_v2.mscXXXX.stream_id`             |
-| `prev_id`                 | Previous recipient user set identifier in federation presence EDU  | `org.continuwuity.presence_v2.mscXXXX.prev_id`               |
-| `m.presence.sharing`      | `account_data` entry for presence sharing config                   | `org.continuwuity.presence_v2.mscXXXX.presence.sharing`      |
-| `m.presence.prompted`     | `account_data` entry for persisting presence hint prompt responses | `org.continuwuity.presence_v2.mscXXXX.presence.prompted`     |
-| `m.room.presence_sharing` | Room state event for presence sharing hint                         | `org.continuwuity.presence_v2.mscXXXX.room.presence_sharing` |
+| `recipients`              | Recipient users to add and delete in federation presence EDU       | `org.continuwuity.presence_v2.msc4495.recipients`            |
+| `stream_id`               | Current recipient user set identifier federation presence EDU      | `org.continuwuity.presence_v2.msc4495.stream_id`             |
+| `prev_id`                 | Previous recipient user set identifier in federation presence EDU  | `org.continuwuity.presence_v2.msc4495.prev_id`               |
+| `m.presence.sharing`      | `account_data` entry for presence sharing config                   | `org.continuwuity.presence_v2.msc4495.presence.sharing`      |
+| `m.presence.prompted`     | `account_data` entry for persisting presence hint prompt responses | `org.continuwuity.presence_v2.msc4495.presence.prompted`     |
+| `m.room.presence_sharing` | Room state event for presence sharing hint                         | `org.continuwuity.presence_v2.msc4495.room.presence_sharing` |
 
-Servers may advertise support for Selective Presence by listing `org.continuwuity.presence_v2.mscXXXX` in the
+Servers may advertise support for Selective Presence by listing `org.continuwuity.presence_v2.msc4495` in the
 `unstable_features` section of the response to [`GET /_matrix/client/versions`][cs-versions].
 
 The stable endpoint `/_matrix/federation/v1/query/presence_recipients` will use the unstable identifier
-`/_matrix/federation/unstable/org.continuwuity.presence_v2.mscXXXX/query/presence_recipients`.
+`/_matrix/federation/unstable/org.continuwuity.presence_v2.msc4495/query/presence_recipients`.
 
 Once this proposal completes FCP, servers may advertise support for the stable identifiers by listing
-`org.continuwuity.presence_v2.mscXXXX.stable` in `unstable_features`; clients may use this while they
+`org.continuwuity.presence_v2.msc4495.stable` in `unstable_features`; clients may use this while they
 are waiting for the server to adopt a version of the spec that includes it.
 
 ## Dependencies
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -255,6 +255,12 @@ The 404 response uses the same format as in the existing query endpoints.
 }
 ```
 
+### Client-Server Capability
+
+Servers implementing Selective Presence MUST include the capability `m.selective_presence` in their response to the `GET
+/_matrix/client/v3/capabilities` endpoint. This allows clients to identify servers using Selective Presence so they MAY
+avoid showing inapplicable UI on servers without it, or inform users when their servers are upgraded to use it.
+
 ## Potential issues
 
 ### Presence Access
@@ -397,6 +403,7 @@ The following complementary proposals are planned:
 | `m.presence.sharing`      | `account_data` entry for presence sharing config                   | `org.continuwuity.presence_v2.msc4495.presence.sharing`      |
 | `m.presence.prompted`     | `account_data` entry for persisting presence hint prompt responses | `org.continuwuity.presence_v2.msc4495.presence.prompted`     |
 | `m.room.presence_sharing` | Room state event for presence sharing hint                         | `org.continuwuity.presence_v2.msc4495.room.presence_sharing` |
+| `m.selective_presence`    | Capability to announce Selective Presence behaviour to clients     | `org.continuwuity.presence_v2.msc4495.selective_presence`    |
 
 Servers may advertise support for Selective Presence by listing `org.continuwuity.presence_v2.msc4495` in the
 `unstable_features` section of the response to [`GET /_matrix/client/versions`][cs-versions].
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -79,6 +79,8 @@ present, the room is treated as though its value is `"forbid"` by default.
 Servers MUST NOT include rooms with a `presence_sharing` value of `"forbid"` when they calculate a user's recipient
 user set, even if the user `"allow"`s presence to be shared with the room.
 
+The `m.room.presence_sharing` hint SHOULD be [transferred upon room upgrade][room-upgrades].
+
 #### Presence Sharing Prompts
 
 A new [`account_data`] entry, `m.presence.prompted`, is introduced to improve the user experience for clients with
@@ -433,3 +435,4 @@ None.
 [cs-versions]: https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientversions
 [cs-createroom]: https://spec.matrix.org/v1.18/client-server-api/#post_matrixclientv3createroom
 [`m.ignored_user_list`]: https://spec.matrix.org/v1.18/client-server-api/#mignored_user_list
+[room-upgrades]: https://spec.matrix.org/v1.18/client-server-api/#server-behaviour-21
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -175,6 +175,11 @@ The **recipient user set** of a user's presence transition is determined as foll
    include users from this list that the sender shares at least one room with.
 5. Exclude any **users listed as `"deny"`** in the user's `m.presence.sharing` configuration.
 
+In addition to current presence update triggers, when a recipient user set changes to include new recipients, an
+update containing the user's presence state SHOULD be sent to those recipients' servers within a reasonable time
+frame. This is to prevent poor UX with not receiving presence until the next transition when you join a room, so
+this proposal recommends using 30 minutes.
+
 Determine the delta from the previous recipient user set. If there is no previous state, the entire set is an addition.
 If the recipient user set is left unchanged, emit a presence EDU using only the prior `stream_id` \- no `recipients`, no
 `prev_id`. Otherwise, generate a `stream_id` for the new state. Emit a presence EDU for each destination of the
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -42,8 +42,9 @@ This event contains four properties:
   2. `rooms` mapping [Room IDs][roomid-format] (default `{}`)
   3. `servers` mapping [Server Names][servername-format] (default `{}`)
 
-While homeservers SHOULD follow these defaults, operators and implementations MAY override them where appropriate.
-For example, an operator may default `share_locally` to `true` in an organisation environment.
+While homeservers SHOULD follow these defaults, operators and implementations MAY override them where appropriate;
+however, they MUST NOT overrule a user's wishes to deny presence sharing. For example, an operator may default
+`share_locally` to `true` in an organisation environment.
 
 No glob resolution is supported for any key in a presence sharing map.
 
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -42,9 +42,10 @@ This event contains four properties:
   2. `rooms` mapping [Room IDs][roomid-format] (default `{}`)
   3. `servers` mapping [Server Names][servername-format] (default `{}`)
 
-While homeservers SHOULD follow these defaults, operators and implementations MAY override them where appropriate;
-however, they MUST NOT overrule a user's wishes to deny presence sharing. For example, an operator may default
-`share_locally` to `true` in an organisation environment.
+If the homeserver requires different defaults, for example setting `share_locally` to true in an organisation
+environment, the server MAY create a pre-populated `m.presence.sharing` event for a user when the user is created
+or if the user exists already but does not have an `m.presence.sharing` event. Otherwise, a missing property MUST
+be assumed to use its above default.
 
 No glob resolution is supported for any key in a presence sharing map.
 
@@ -53,9 +54,6 @@ The string values of these presence sharing maps may be any of the following:
   The `"deny"` state MUST NOT be specified for room IDs under the `"rooms"` object.
 * `"allow"`: Permit sharing presence with the entity  
   The `"allow"` state MUST NOT be specified for entire homeservers under the `"servers"` object.
-
-Servers MUST treat users without `m.presence.sharing` in their `account_data` as though the default values were
-specified.
 
 Sample `m.presence.sharing` account data event:
 ```json
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -255,12 +255,6 @@ The 404 response uses the same format as in the existing query endpoints.
 }
 ```
 
-### Client-Server Capability
-
-Servers implementing Selective Presence MUST include the capability `m.selective_presence` in their response to the `GET
-/_matrix/client/v3/capabilities` endpoint. This allows clients to identify servers using Selective Presence so they MAY
-avoid showing inapplicable UI on servers without it, or inform users when their servers are upgraded to use it.
-
 ## Potential issues
 
 ### Presence Access
@@ -403,7 +397,6 @@ The following complementary proposals are planned:
 | `m.presence.sharing`      | `account_data` entry for presence sharing config                   | `org.continuwuity.presence_v2.msc4495.presence.sharing`      |
 | `m.presence.prompted`     | `account_data` entry for persisting presence hint prompt responses | `org.continuwuity.presence_v2.msc4495.presence.prompted`     |
 | `m.room.presence_sharing` | Room state event for presence sharing hint                         | `org.continuwuity.presence_v2.msc4495.room.presence_sharing` |
-| `m.selective_presence`    | Capability to announce Selective Presence behaviour to clients     | `org.continuwuity.presence_v2.msc4495.selective_presence`    |
 
 Servers may advertise support for Selective Presence by listing `org.continuwuity.presence_v2.msc4495` in the
 `unstable_features` section of the response to [`GET /_matrix/client/versions`][cs-versions].
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -34,7 +34,7 @@ communities, private rooms, and DMs, while reducing the load from federating pre
 
 A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
 
-This event contains four fields:
+This event contains four properties:
 * The boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
   (default `false`)
 * Three **presence sharing map** objects, mapping one of several entity types to presence sharing behaviour:
@@ -82,7 +82,7 @@ user set, even if the user `"allow"`s presence to be shared with the room.
 
 A new [`account_data`] entry, `m.presence.prompted`, is introduced to improve the user experience for clients with
 presence sharing prompts. To allow users to dismiss a prompt on one client and have this register across all of
-their clients, so that they do not have to answer the prompt several times, the following standard fields are
+their clients, so that they do not have to answer the prompt several times, the following standard properties are
 used:
 * `users`: An array of [User IDs][mxid-format] (default `[]`)
 * `rooms`: An array of [Room IDs][roomid-format] (default `[]`)
@@ -105,7 +105,8 @@ joined user or the room itself otherwise, in the following circumstances:
 * The relevant entity is not in the user's corresponding `m.presence.prompted` array or `m.presence.sharing` map
 
 Clients MAY wish to check room predecessor chains to avoid showing previously dismissed prompts in upgraded rooms.
-They MUST NOT prompt the user to share presence in a room with a `presence_sharing` value of `"forbid"`.
+They SHOULD carry settings over when starting and following a room upgrade. They MUST NOT prompt the user to share
+presence in a room with a `presence_sharing` value of `"forbid"`.
 
 #### Room Defaults
 
@@ -126,13 +127,14 @@ The federation [User Presence Update] type is modified to include:
 * A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
   be sequential or in any particular order, only unique.
 
+The `add` and `del` arrays of `recipients` represent an incremental update to a user's recipient user set, adding
+and removing users respectively. A server MUST NOT send updates lacking a populated `add` to a destination
+homeserver after their last user is updated out of the recipient set.
+
 `stream_id` and `prev_id` form a sequence representing the state of the user's recipient user set, similar to the
 behaviour of the [Device List Update] type. `recipients` may only be present if `prev_id` is present. If a user's
 `stream_id` is replaced but no changes are relevant to the receiving homeserver, the sending server uses empty `add`
 and `del` lists.
-
-A server MUST NOT send updates lacking a populated `add` to a destination homeserver after their last user is
-updated out of the recipient set.
 
 `m.presence` EDU (as received by `example.com`):
 ```json
@@ -173,10 +175,10 @@ The **recipient user set** of a user's presence transition is determined as foll
 5. Exclude any **users listed as `"deny"`** in the user's `m.presence.sharing` configuration.
 
 Determine the delta from the previous recipient user set. If there is no previous state, the entire set is an addition.
-If the recipient user set is left unchanged, emit a presence EDU using only the prior `stream_id` \- no `recipients`,
-no `prev_id`. Otherwise, generate a `stream_id` for the new state. For each server name represented in these updates,
-emit a presence EDU with `prev_id` set to the last `stream_id` sent to that destination if applicable, `stream_id` set
-according to the new state, and `recipients` populated with the updates.
+If the recipient user set is left unchanged, emit a presence EDU using only the prior `stream_id` \- no `recipients`, no
+`prev_id`. Otherwise, generate a `stream_id` for the new state. Emit a presence EDU for each destination of the
+recipients in these updates, with `prev_id` set to the last `stream_id` sent to the destination if applicable,
+`stream_id` set according to the new state, and `recipients` populated with the updates for the destination.
 
 #### Inbound Processing
 
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -121,18 +121,18 @@ They MUST NOT prompt the user to share presence in a room with a `presence_shari
 ### Modifications to the [`m.presence` EDU]
 
 The federation [User Presence Update] type is modified to include:
-* An optional map, `recipients`, containing two populated arrays of [User IDs][mxid-format] belonging to the
-  destination server, `add` and `del`.
+* An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
+  server, `add` and `del`.
 * A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
   be sequential or in any particular order, only unique.
 
 `stream_id` and `prev_id` form a sequence representing the state of the user's recipient user set, similar to the
 behaviour of the [Device List Update] type. `recipients` may only be present if `prev_id` is present. If a user's
-`stream_id` is incremented but no changes are relevant to the receiving homeserver, the sending server uses empty `add`
+`stream_id` is replaced but no changes are relevant to the receiving homeserver, the sending server uses empty `add`
 and `del` lists.
 
-A server MUST NOT send updates lacking `add` to a destination homeserver after their last user is updated out of the
-recipient set.
+A server MUST NOT send updates lacking a populated `add` to a destination homeserver after their last user is
+updated out of the recipient set.
 
 `m.presence` EDU (as received by `example.com`):
 ```json
@@ -191,17 +191,17 @@ change over both `/sync` and `/_matrix/client/v3/presence/{userId}/status`:
 * **`recipients` present, `prev_id` present:** Servers MUST apply the delta to the set corresponding to `prev_id`
   and only distribute the presence change to users in the resulting set, superseding `prev_id` in the process.
 
-For cases where `prev_id` is present but is not the previous `stream_id`, or `stream_id` is present on its own but
-unrecognised, servers SHOULD query the `GET /_matrix/federation/v1/query/presence_recipients` endpoint on the origin
-server.
+If `prev_id` is present but differs from the previous `stream_id`, or `stream_id` is present on its own but
+unrecognised, servers SHOULD query the `GET /_matrix/federation/v1/query/presence_recipients` endpoint on the
+origin server.
 
 ### Server-Server Endpoint `GET /_matrix/federation/v1/query/presence_recipients`
 
 Servers that do not recognise a `stream_id` or `prev_id` in an incoming `m.presence` EDU need a way to ask the
 origin server for the user's latest `stream_id` and a snapshot of its recipient user set. This proposal introduces
-a new federation endpoint to address this need. If a server is unable to determine the proper recipient user set
-(for example, if this endpoint returns an error), servers SHOULD treat it as the empty set, `[]` until it is able
-to fetch the full list.
+a new federation endpoint to address this need. If a server cannot determine the proper recipient user set \- for
+example, if this endpoint returns an error \- it SHOULD be assumed to be empty (`[]`) until the server can fetch
+the full set.
 
 <table>
   <tbody>
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -123,18 +123,18 @@ presence in a room with a `presence_sharing` value of `"forbid"`.
 
 The federation [User Presence Update] type is modified to include:
 * An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
-  server, `add` and `del`.
+  server, `add` and `delete`.
 * A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
   be sequential or in any particular order, only unique.
 
-The `add` and `del` arrays of `recipients` represent an incremental update to a user's recipient user set, adding
-and removing users respectively. A server MUST NOT send updates lacking a populated `add` to a destination
+The `add` and `delete` arrays of `recipients` represent an incremental update to a user's recipient user set,
+adding and removing users respectively. A server MUST NOT send updates lacking a populated `add` to a destination
 homeserver after their last user is updated out of the recipient set.
 
 `stream_id` and `prev_id` form a sequence representing the state of the user's recipient user set, similar to the
 behaviour of the [Device List Update] type. `recipients` may only be present if `prev_id` is present. If a user's
 `stream_id` is replaced but no changes are relevant to the receiving homeserver, the sending server uses empty `add`
-and `del` lists.
+and `delete` lists.
 
 `m.presence` EDU (as received by `example.com`):
 ```json
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -150,7 +150,7 @@ and `delete` lists.
                 "user_id": "@john:matrix.org",
                 "recipients": {
                     "add": ["@alice:example.com"],
-                    "del": ["@geoff:example.com"],
+                    "delete": ["@geoff:example.com"],
                 },
                 "stream_id": 1,
                 "prev_id": 0,
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -315,6 +315,10 @@ presence with an entire homeserver, users are almost certainly on the same local
 applies, or share a room with everyone in the organisation, which the users can set as `"allow"`. Excluding a room
 has no practical benefits as users can simply not include it in the configuration.
 
+Globs were prohibited in presence sharing maps to avoid introducing security issues for servers running glob parsing on
+arbitrary user-provided strings, and because users are not typically namespaced beyond servers in a way that would be
+suitable for glob selection.
+
 ### Entity Resolution Ordering
 
 The inclusion and exclusion lists could be applied in different orders; for example, if a user of a homeserver is set to
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -39,9 +39,9 @@ associated behaviours, unless they are explicitly declared to be OPTIONAL.
 A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
 
 This event contains four properties:
-* The boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
+* The OPTIONAL boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
   (default `false`)
-* Three **presence sharing map** objects, mapping one of several entity types to presence sharing behaviour:
+* Three OPTIONAL **presence sharing map** objects, mapping one of several entity types to presence sharing behaviour:
   1. `users` mapping [User IDs][mxid-format] (default `{}`)
   2. `rooms` mapping [Room IDs][roomid-format] (default `{}`)
   3. `servers` mapping [Server Names][servername-format] (default `{}`)
@@ -74,9 +74,9 @@ Sample `m.presence.sharing` account data event:
 
 ### Room Presence Sharing Hints
 
-A new room state event is introduced, `m.room.presence_sharing`. This state event contains one field,
+A new room state event is introduced, `m.room.presence_sharing`. This state event contains one OPTIONAL field,
 `presence_sharing`, which is a string with two possible values: `"suggest"` and `"forbid"`. If the event is not
-present, the room is treated as though its value is `"forbid"` by default.
+present, or the field is missing, the room is treated as though its value is `"forbid"` by default.
 
 Servers MUST NOT include rooms with a `presence_sharing` value of `"forbid"` when they calculate a user's recipient
 user set, even if the user `"allow"`s presence to be shared with the room.
@@ -85,12 +85,11 @@ The `m.room.presence_sharing` hint SHOULD be [transferred upon room upgrade][roo
 
 #### Presence Sharing Prompts
 
-A new [`account_data`] entry, `m.presence.prompted`, is introduced to improve the user experience for clients with
-presence sharing prompts. To allow users to dismiss a prompt on one client and have this register across all of
-their clients, so that they do not have to answer the prompt several times, the following standard properties are
-used:
-* `users`: An array of [User IDs][mxid-format] (default `[]`)
-* `rooms`: An array of [Room IDs][roomid-format] (default `[]`)
+A new OPTIONAL [`account_data`] entry, `m.presence.prompted`, is introduced to improve the user experience for clients
+with presence sharing prompts. To allow users to dismiss a prompt on one client and have this register across all of
+their clients, so that they do not have to answer the prompt several times, the following standard properties are used:
+* `users`: An OPTIONAL array of [User IDs][mxid-format] (default `[]`)
+* `rooms`: An OPTIONAL array of [Room IDs][roomid-format] (default `[]`)
 
 Sample `m.presence.prompted` account data event:
 ```json
--- a/proposals/4495-selective-presence.md
+++ b/proposals/4495-selective-presence.md
@@ -214,18 +214,9 @@ REQUIRED federation endpoint to address this need. If a server cannot determine 
 example, if this endpoint returns an error \- it SHOULD be assumed to be empty (`[]`) until the server can fetch the
 full set.
 
-<table>
-  <tbody>
-    <tr>
-      <th scope="row">Rate limited:</th>
-      <td>No</td>
-    </tr>
-    <tr>
-      <th scope="row">Requires authentication:</th>
-      <td>Yes</td>
-    </tr>
-  </tbody>
-</table>
+| Rate limited | Requires authentication |
+|--------------|-------------------------|
+| No           | Yes                     |
 
 #### Request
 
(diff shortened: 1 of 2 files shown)


proposals/4495-selective-presence.md
turt2live, line 1: Implementation requirements:

* [ ] Client (account data)
* [ ] Client (state events)
* [ ] Server (sending federation)
* [ ] Server (receiving federation)
* [ ] Server (backwards compatibility)
* [ ] Server (endpoint - using)
* [ ] Server (endpoint - serving)

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+2. users benefit from improved privacy by only sharing such sensitive information with those they want to have
+   access to it.
+
+## Prior Art
+
+Prior to [MSC1819], a mechanism known as "presence lists" allowed users to subscribe to others' presence. In
+theory, it also offered a mechanism for other users to request your presence. However, Synapse historically
+[accepted all presence requests automatically][#4417], ultimately resulting in the feature's removal prior to r0.
+
+[MSC4325: Presence privacy][MSC4325] attempts to solve this problem by reintroducing a concept similar to Selective
```
HarHarLinks, line 25: As author of the much rougher 4325, I endorse this MSC as it appears to follow a similar approach but has a very simple model worked out that is easy to understand as MSC reader. Particularly I think that dropping wildcards is (mostly) a good idea.

Regarding whether it solves the 2 problems of server performance and all it's features actually getting used (the presence list issue), we will need to watch how implementations fare in practice.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+    "users": {
+        "@alice:example.com": "allow",
+        "@mallory:example.com": "deny",
+    },
+    "rooms": { "!family-group-chat": "allow" },
+    "servers": { "matrix.org": "deny" }
+}
+```
+
+### Room Presence Sharing Hints
```
HarHarLinks, line 75: I was very wary about letting another party effectively control with whom I share my presence before I realized I need to opt into the rooms first to then follow their hints. I think this opens the door for simple UX that users might expect as an aspect of their choice, especially coming from other platforms, with a client-local setting to "allow all" rooms. In general my philosophy is that most users will likely want to set simple global toggles like "none", "all", "contacts", etc.

For this whole hints feature it will be important to see how well it gets adopted an actually used, based on wether clients can find reasonable UX for it.

Despite not supporting wildcards, we may end up wanting to allow some kind of "all" special value.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+This event contains four fields:
+* The boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
+  (default `false`)
+* Three **presence sharing map** objects, mapping one of several entity types to presence sharing behaviour:
+  1. `users` mapping [User IDs][mxid-format] (default `{}`)
+  2. `rooms` mapping [Room IDs][roomid-format] (default `{}`)
+  3. `servers` mapping [Server Names][servername-format] (default `{}`)
+
+While homeservers SHOULD follow these defaults, operators and implementations MAY override them where appropriate.
+For example, an operator may default `share_locally` to `true` in an organisation environment.
```
HarHarLinks: It would be reasonable for a server also to allow setting "share locally only", as well as disabling presence entirely as is possible today.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+would like to share their presence, or forbid sharing presence altogether. This keeps presence usable for small
+communities, private rooms, and DMs, while reducing the load from federating presence in large public rooms.
+
+## Proposal
+
+### User Presence Sharing Preferences
+
+A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
+
+This event contains four fields:
```
HarHarLinks: (iirc the spec likes to call these
```suggestion
This event contains four properties:
```
Just for consistency, no real impact of course)

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+| ---------------------- | ----------------------------------- |
+| `private_chat`         | `{ "presence_sharing": "suggest" }` |
+| `trusted_private_chat` | `{ "presence_sharing": "suggest" }` |
+| `public_chat`          | `{ "presence_sharing": "forbid" }`  |
+
+### Modifications to the [`m.presence` EDU]
+
+The federation [User Presence Update] type is modified to include:
+* An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
+  server, `add` and `del`.
```
HarHarLinks: I am missing a definition or reference what add and del are in this section. (There is only a mention of "addition" and "delta" in the algorithm subsection.)

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+}
+```
+
+#### Dispatch Algorithm
+
+The **recipient user set** of a user's presence transition is determined as follows:
+1. Include all members of rooms both **listed as `"allow"`** in the user's `m.presence.sharing` configuration and
+   with an `m.room.presence_sharing` state event with `presence_sharing` set to `"suggest"`. Servers MUST NOT
+   include rooms that do not satisfy this criteria.
+2. Exclude any user IDs with a server name corresponding to **any server listed as `"deny"`** in the user's
```
HarHarLinks, line 173: This means that individual "users allowed" override "servers denied". While I think it's reasonable that more specific settings should override less specific ones, it is unfortunate that this means a server could not trivially implement specific behavior that allows an operator to write to a server presence block list, e.g. of servers known to have presence disabled anyway. I wonder if this could be tweaked.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+   `m.presence.sharing` configuration.
+3. Include all local users if the user's `m.presence.sharing` configuration has `share_locally` set to `true`.
+   Servers SHOULD only include local users that the sender shares at least one room with.
+4. Include all **users listed as `"allow"`** in the user's `m.presence.sharing` configuration. Servers SHOULD only
+   include users from this list that the sender shares at least one room with.
+5. Exclude any **users listed as `"deny"`** in the user's `m.presence.sharing` configuration.
+
+Determine the delta from the previous recipient user set. If there is no previous state, the entire set is an addition.
+If the recipient user set is left unchanged, emit a presence EDU using only the prior `stream_id` \- no `recipients`,
+no `prev_id`. Otherwise, generate a `stream_id` for the new state. For each server name represented in these updates,
```
HarHarLinks:
```suggestion
no `prev_id`. Otherwise, generate a `stream_id` for the new state. For each server name represented among the recipients in these updates,
```

Alternatively maybe something like "group by server name of recipients, generate EDU for each"

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+emit a presence EDU with `prev_id` set to the last `stream_id` sent to that destination if applicable, `stream_id` set
```
HarHarLinks: I feel like sending EDUs without length cap is not reasonable.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+according to the new state, and `recipients` populated with the updates.
+
+#### Inbound Processing
+
+When a server receives a `m.presence` EDU over federation, servers follow these rules for distributing the presence
```
HarHarLinks:
```suggestion
When a server receives an `m.presence` EDU over federation, servers follow these rules for distributing the presence
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+the full set.
+
+<table>
+  <tbody>
+    <tr>
+      <th scope="row">Rate limited:</th>
+      <td>No</td>
+    </tr>
+    <tr>
+      <th scope="row">Requires authentication:</th>
```
HarHarLinks: Should servers only accept requests on this endpoint if they previously sent an EDU to the requesting server? I suppose there isn't really a security issue here.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+1. Include all members of rooms both **listed as `"allow"`** in the user's `m.presence.sharing` configuration and
+   with an `m.room.presence_sharing` state event with `presence_sharing` set to `"suggest"`. Servers MUST NOT
+   include rooms that do not satisfy this criteria.
+2. Exclude any user IDs with a server name corresponding to **any server listed as `"deny"`** in the user's
+   `m.presence.sharing` configuration.
+3. Include all local users if the user's `m.presence.sharing` configuration has `share_locally` set to `true`.
+   Servers SHOULD only include local users that the sender shares at least one room with.
+4. Include all **users listed as `"allow"`** in the user's `m.presence.sharing` configuration. Servers SHOULD only
+   include users from this list that the sender shares at least one room with.
+5. Exclude any **users listed as `"deny"`** in the user's `m.presence.sharing` configuration.
```
HarHarLinks, line 179: Would it not be reasonable to also exclude ignored users?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+achieve any privacy benefits. An auxiliary fetchable presence system is considered complementary follow-up work to
+allow users to make their presence public.
+
+### Receiver Resolution
+
+Instead of sending a finalised list of recipient users to remote servers, the user's preferences (allowed rooms and
+users, denied users) could be sent over federation. This proposal uses sender-side user resolution to avoid sending
+a "block list" to third parties, and to keep resolution consistent with the sending homeserver's state.
+
+### Entity Selection Affordances
```
HarHarLinks, line 310: Practically, the assumption is that it's not worth including whole servers because either they are small servers with few users you might as well add manually, or against the goal of the proposal to limit to whom my presence gets exposed. Idk this might be worth adding or not.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+recommendation purpose of the hints system.
+
+Allowing room administrators to suggest that their members share presence means users can still see presence for
+members of small rooms (assuming they opt in to this behaviour) without explicitly creating user lists. These are
+the rooms where presence matters most: small groups of friends, organisation environments, etc.
+
+Only allowing users to follow hints via `"allow"` means a `"forbid"` hint state explicitly forbids presence sharing
+for all users of a room. This allows room administrators to retroactively *disable* presence sharing for rooms as
+they grow; for example, if a previously private room grows to be public and contain many members, a room
+administrator may wish to disable presence sharing in the room.
```
HarHarLinks, line 343: Hm, why would an admin do that? I don't think room admins will want or be able to make judgements on the performance (?) of participating HS.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+{
+    "stream_id": 1,
+    "recipients": [
+        "@jesse:example.com",
+        "@jane:example.com"
+    ]
+}
+```
+
+## Potential issues
```
HarHarLinks, line 255: How will the account data list get cleaned up?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+but it is considered out of scope for this proposal.
+
+### Performance
+
+Homeserver operators and implementers may wish to limit the performance impacts of presence, particularly as the
+goal of Selective Presence is to make presence viable for popular use. Potential options for doing so may include:
+* Limiting the ability of users to share presence with rooms above a certain number of members
+* Limiting the number of entities (or total recipient users) users can have in their presence sharing maps
+
+### Presence EDU Batch Sizes
```
HarHarLinks, line 287: https://github.com/matrix-org/matrix-spec/issues/807

proposals/4495-selective-presence.md
lveneris: What do you mean by a length cap in this case? A maximum allowed number of entities, or a maximum size? The proposal originally borrowed 255 from 4325, and then calculated you could only fit 250 entities in a previous model of the EDU, and then we dropped it altogether noting that EDU size limits should be introduced to the spec before this can be sensibly decided on.

proposals/4495-selective-presence.md
lveneris: The spec uses both "an" and "a" in this case. I did check. Since the spec is already inconsistent, it made the most sense to me to use "a" as in "a Matrix presence EDU"

proposals/4495-selective-presence.md
lveneris: This is already technically the case, because if you've never sent an EDU to a server before that means you have no recipients on the server, so the endpoint should 404 anyway.

proposals/4495-selective-presence.md
lveneris: We opted against doing so with the rationale that ignoring users, by design, prevents you from seeing their events but not them from seeing yours. Ignored users are supposed to be one-way, so we're keeping with that theme.

proposals/4495-selective-presence.md
lveneris: We do not expect them to be automatically cleaned up, because it is difficult to design a sensible heuristic for doing so. Sometimes, you have to leave and rejoin rooms. It would be rather unfortunate if you had to reconfigure your presence because your client decided it was a good idea to remove the room from your list.

You could potentially design a system where they're removed if you're not in contact with them again within 24 hours, but this feels like more work than it's worth.

proposals/4495-selective-presence.md
lveneris: Performance, or perhaps privacy concerns, or to limit exposure temporarily if users are spamming inappropriate status messages, or if they just don't want presence to be shared in their room, etc.

proposals/4495-selective-presence.md
lveneris: I'll restructure this later.

proposals/4495-selective-presence.md
lveneris: We determined that the order should be users > servers > rooms. If you make servers override users in the deny case, it makes it impossible to provide exceptions for a server deny, which is unfortunate. Also, we determined that if you have individual users explicitly set to allow, expected behaviour would be that they get your presence. This is covered in "Entity Resolution Ordering" already.

proposals/4495-selective-presence.md
HarHarLinks: Yeah maybe it's fine given the comment below

/resolve

proposals/4495-selective-presence.md
TheArcaneBrony: I personally don't agree that dropping wildcards is a good idea, as it doesn't allow me to use presence the way I'd expect it to work (broadcast by default), or addressing a group of users by mxid format (eg. `{"allow_servers": ["rory.gay"], "deny_users": ["@ext.*:rory.gay"]}`)

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+## Potential issues
+
+### Presence Access
+
+Selective Presence *disables presence by default*. This may complicate future proposals that require publicly
+accessible presence information, and may change the way users interact with presence features. However, as
+previously mentioned, presence should not be public by default. A future proposal may address public, fetchable
+presence (see [Alternatives](#Alternatives)).
```
TheArcaneBrony, line 262: I feel like this may be in conflict with the expectations of users comming from other platforms, since the general behavior there is that presence is *public* with (partial) ability to opt out (eg. discord not allowing presence to be private beyond setting it to an "invisible" state, but allowing per-guild opt out of sharing *rich* presence info) - this is directly in conflict with the entire idea of using something like rich presence *itself* as an extension of your persona

proposals/4495-selective-presence.md
lveneris: There are two things to note here:

1. No matter how much you improve the performance of presence, broadcasting public presence creates unnecessary traffic for users that are not actively interested, and can lead to situations like the one presence is in now

2. Doing glob parsing on arbitrary user-provided strings is a security issue, and users are not typically grouped together by sections of their localparts except for in appservices, which is a narrow use case already but you can still send presence to them easily since they will probably share a room

This rationale is given in the proposal already.

TheArcaneBrony: for point 2, isn't this already the case for server ACLs for example anyways? Besides, could be done by the local server, I suppose?

proposals/4495-selective-presence.md
HarHarLinks: On the flip side, a client should probably carry settings over both when starting and following a room upgrade? Then this should be added.

proposals/4495-selective-presence.md
HarHarLinks: > I personally don't agree that dropping wildcards is a good idea, as it doesn't allow me to use presence the way I'd expect it to work (broadcast by default), or addressing a group of users by mxid format (eg. `{"allow_servers": ["rory.gay"], "deny_users": ["@ext.*:rory.gay"]}`)

This may be true, but in practice globbing by MXID (particularly allow) is a mostly unrealistic use case. Similarly I doubt that allowlisting a whole server that 1) isn't your own home 2) of nontrivial size is not an edge case. The real convincing point to me is though that this would be impossible to build usable UX for and is thus asking for trouble.

proposals/4495-selective-presence.md
HarHarLinks: If so, this should be reflected in UX somehow. I.e. users on such a server trying to allowlist others would be pretty confused (and wasting their time) if there was no UI feedback (e.g. that part of UI disabled).

proposals/4495-selective-presence.md
HarHarLinks: /resolve

proposals/4495-selective-presence.md
lveneris: I do not think it is fair to say that public presence is a general behaviour. There are several popular WhatsApp-style platforms where presence information is strictly limited to your contacts, or some other curated model. For platforms like Discord, a pubsub model is used internally, but with a looser mode of access control; the concept is similar to just adding rooms to the sharing map and removing them on request, although I would not advise this. We anticipate that a similar pubsub model will be used for the development of rich presence in the future.

proposals/4495-selective-presence.md
thetayloredman: bee426874d85da06689c5d292b305af9a3595eb7

proposals/4495-selective-presence.md
thetayloredman: bee426874d85da06689c5d292b305af9a3595eb7

proposals/4495-selective-presence.md
thetayloredman: Additionally, users can easily be taught. "You must share presence with who you want to see it" is not a particularly tough UI choice.

Half-Shot: > Additionally, users can easily be taught. "You must share presence with who you want to see it" is not a particularly tough UI choice.

I would caveat that with a gentle suggestion that we should prove this in implementation. It's not the first MSC where behaviors sounded easy on paper and were a nightmare to design for.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+
+This event contains four properties:
+* The boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
+  (default `false`)
+* Three **presence sharing map** objects, mapping one of several entity types to presence sharing behaviour:
+  1. `users` mapping [User IDs][mxid-format] (default `{}`)
+  2. `rooms` mapping [Room IDs][roomid-format] (default `{}`)
+  3. `servers` mapping [Server Names][servername-format] (default `{}`)
+
+While homeservers SHOULD follow these defaults, operators and implementations MAY override them where appropriate.
```
Half-Shot: We should be careful here that servers may be permitted to set defaults or add additional limitations to presence, but MUST not ignore the users request to deny presence further. Taking inspiration from https://github.com/matrix-org/matrix-spec-proposals/pull/4278, we allow a model where the server may default but does not overrule the users choice.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+For example, an operator may default `share_locally` to `true` in an organisation environment.
+
+No glob resolution is supported for any key in a presence sharing map.
```
Half-Shot, line 54: This sounds like a decision was made, can you expand on why?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+communities, private rooms, and DMs, while reducing the load from federating presence in large public rooms.
+
+## Proposal
+
+### User Presence Sharing Preferences
+
+A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
+
+This event contains four properties:
+* The boolean `share_locally` enables sharing presence with all users with mutual rooms on the local homeserver
```
Half-Shot: Can you expand on why we don't just have:

```json
    "servers": { "my-local-server": "allow|deny" }
```

?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+    "rooms": { "!family-group-chat": "allow" },
+    "servers": { "matrix.org": "deny" }
+}
+```
+
+### Room Presence Sharing Hints
+
+A new room state event is introduced, `m.room.presence_sharing`. This state event contains one field,
+`presence_sharing`, which is a string with two possible values: `"suggest"` and `"forbid"`. If the event is not
+present, the room is treated as though its value is `"forbid"` by default.
```
Half-Shot: This would presently break the current presence model? How would users/server admins migrate, would they need to send state events to all rooms (twice, because of the unstable prefix?)?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+
+### Modifications to the [`m.presence` EDU]
+
+The federation [User Presence Update] type is modified to include:
+* An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
+  server, `add` and `del`.
+* A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
+  be sequential or in any particular order, only unique.
+
+The `add` and `del` arrays of `recipients` represent an incremental update to a user's recipient user set, adding
```
Half-Shot: For the sake of a couple of bytes, could we call this `delete` (or remove, as you said a bit later). I feel passionately that shortening property names creates more confusion

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+    <tr>
+      <th scope="row">Rate limited:</th>
+      <td>No</td>
+    </tr>
+    <tr>
+      <th scope="row">Requires authentication:</th>
+      <td>Yes</td>
+    </tr>
+  </tbody>
+</table>
```
Half-Shot: meta note but this is hard to parse

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+There are multiple potential states for room hints, and just as many potential defaults. It was ultimately
+determined that this proposal should only specify the actual states of `"suggest"` or `"forbid"`, leaving the
+decision of when to ask users if they would like to enable presence up to the clients.
+
+### Room Account Data
+
+It was proposed that a user's `m.presence.sharing` `"rooms"` configuration could live in room account data, but
+this proposal opted against doing so to avoid fragmenting the configuration between different sources.
+
+### Fallback Behaviours
```
Half-Shot, line 356: Is this still part of the parent Alternatives section, if so this is a bit confusing

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,430 @@
...
+No glob resolution is supported for any key in a presence sharing map.
+
+The string values of these presence sharing maps may be any of the following:
+* `"deny"`: Explicitly forbid sending presence to the entity  
+  The `"deny"` state MUST NOT be specified for room IDs under the `"rooms"` object.
+* `"allow"`: Permit sharing presence with the entity  
+  The `"allow"` state MUST NOT be specified for entire homeservers under the `"servers"` object.
+
+Servers MUST treat users without `m.presence.sharing` in their `account_data` as though the default values were
+specified.
```
Half-Shot: I think it's going to be too radical a change to go from sending presence everywhere to not sending any without any user interaction. This will almost certainly break some workflows that are relied upon today.

Can we instead at least introduce a capability or some measure that allows servers to tell clients they are now using selective presence? I think more discussion about how folks migrate needs to be present in this proposal.

proposals/4495-selective-presence.md
thetayloredman: See https://github.com/matrix-org/matrix-spec-proposals/pull/4495#discussion_r3486370471. @lveneris maybe this should end up in the proposal text?

proposals/4495-selective-presence.md
thetayloredman: The section [Alternatives / Entity Selection Affordances](https://github.com/thetayloredman/matrix-spec-proposals/blob/presence-v2/selective/proposals/4495-selective-presence.md#entity-selection-affordances) describes why we chose to forbid `allow`ing entire servers in the `servers` map. We found there to be few real world use cases, and sending a full server `allow` list over federation (especially to large homeservers) would be difficult with the model we settled on.

Additionally, we anticipate few clients will actually utilize the `servers` map in their UX.

proposals/4495-selective-presence.md
thetayloredman: Yes, rooms will all default to having presence disabled, and their administrators will need to send the state event to change it to `suggest`. Clients could choose to e.g. display a prompt in 1:1 rooms to ask the user if they would like to both enable the hint and add the room to their sharing map, but this behavior is not specified.

I would be interested in investigating if it is possible to specify further suggestions for this sort of "auto-migration" behavior primarily for DMs.

proposals/4495-selective-presence.md
lveneris: Further to the points given by @thetayloredman, providing "allow" for one special case of this map would make things harder for implementations in strongly typed languages. Having the type of this map be the enum `"allow" | "deny"` when `"allow"` is only valid for your local server would be unfortunate.

proposals/4495-selective-presence.md
lveneris: I believe it was in the proposal text at one point. I will reintroduce it to the proposal shortly.

proposals/4495-selective-presence.md
lveneris: I would suggest that we implement this as a capability, since it would make no sense for clients to show the selective presence UI on servers without it regardless. I can see this being a problem for bots and appservices, which may want their users to know if they're online or not, but other than that it is hard to consider how people might migrate workflows without knowing what those workflows are. On the other hand, I am wary of falling into an XKCD 1172 situation.

proposals/4495-selective-presence.md
lveneris: What would the preferred formatting be? I formatted this as it is formatted in the specification itself, because I didn't see a style guide in the repo documents.

proposals/4495-selective-presence.md
lveneris: After some discussion with Ginger, I must ask, what would the purpose of introducing a capability be? It is my understanding that capabilities are for features that are implementation dependent, i.e. optional for a given spec version. Selective Presence is not intended to be an optional feature. It is a whole replacement of the existing presence broadcasting model. Clients already have a way to tell if a server is using Selective Presence in this regard, through `GET /_matrix/client/versions`.

proposals/4495-selective-presence.md
thetayloredman: Implementation in Continuwuity: https://forgejo.ellis.link/continuwuation/continuwuity/pulls/2034

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+or if the user exists already but does not have an `m.presence.sharing` event. Otherwise, a missing property MUST
+be assumed to use its above default.
+
+No glob resolution is supported for any key in a presence sharing map.
+
+The string values of these presence sharing maps may be any of the following:
+* `"deny"`: Explicitly forbid sending presence to the entity  
+  The `"deny"` state MUST NOT be specified for room IDs under the `"rooms"` object.
+* `"allow"`: Permit sharing presence with the entity  
+  The `"allow"` state MUST NOT be specified for entire homeservers under the `"servers"` object.
```
Bentheburrito: In the spirit of "make impossible states impossible to represent," why not substitute `rooms` for `allowed_rooms` as an array of room IDs?

Similarly, could do an array of `denied_servers` instead of `servers`

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+present, the room is treated as though its value is `"forbid"` by default.
+
+Servers MUST NOT include rooms with a `presence_sharing` value of `"forbid"` when they calculate a user's recipient
+user set, even if the user `"allow"`s presence to be shared with the room.
+
+The `m.room.presence_sharing` hint SHOULD be [transferred upon room upgrade][room-upgrades].
+
+#### Presence Sharing Prompts
+
+A new [`account_data`] entry, `m.presence.prompted`, is introduced to improve the user experience for clients with
```
Bentheburrito: Is `m.presence.prompted` necessary? Couldn't the `m.presence.sharing` event be queried for the target user/room to tell if they've been prompted before? e.g. if `m.presence.sharing` contains an allow/deny directive for a user, you can infer they've already been prompted.

I think since currently `rooms` only defines an `"allow"` value, that'd be a problem, so perhaps it could be `"allow" | "prompted"`?

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+| `trusted_private_chat` | `{ "presence_sharing": "suggest" }` |
+| `public_chat`          | `{ "presence_sharing": "forbid" }`  |
+
+### Modifications to the [`m.presence` EDU]
+
+The federation [User Presence Update] type is modified to include:
+* An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
+  server, `add` and `delete`.
+* A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
+  be sequential or in any particular order, only unique.
```
Bentheburrito: Just curious, if the only requirement is that these values are unique, why integers?

also, since `prev_id` is supposed to be the previous `stream_id`, I'd recommend `prev_stream_id` to make that explicit

I'd also like to call out that these IDs add a lot of additional book-keeping for homeservers to do. That's not necessarily bad on its own, though considering presence data is supposed to be ephemeral, I feel like it's not ideal. I am wondering, since there is a way for destination servers to query their recipient user set via `GET /_matrix/federation/v1/query/presence_recipients`, if using hashes of the recipient set would be better.

I haven't thought it through much, so please poke holes: I think if a scheme to produce a `sha256` of the recipient user set is defined, the sending server could include it in the `m.presence` EDU instead of `stream_id` and `prev_id`. If, after the destination server applies the incremental `recipients`, its hash does not match the one on the EDU, it will get the latest snapshot from the endpoint

thetayloredman: The EDU design (including the use of integers, alongside the naming of `stream_id` and `prev_id`) is heavily inspired by [device list update EDUs](https://spec.matrix.org/v1.19/server-server-api/#device-management), which form a DAG. We decided to limit this to a sequence (at most one `prev_id`), which made the approach much simpler.

<sub>It is worth noting that some implementations <a href="https://forgejo.ellis.link/continuwuation/continuwuity/src/commit/480e9758772e2e4f49718b4293e83b7e82fd4229/src/api/server/send.rs#L546">entirely ignore the device list update DAG</a>. This is also very possible, one can just fetch the recipient set on every inbound transaction.</sub>

When it comes to the tracking implementations need to do, for remote users it is as small as the most recent ID and a list of local recipients per remote user. [The logic on inbound transactions is then pretty simple](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/2034/files#diff-b18d44a2fa7605012739427bb0533318f4d86ebd).

Although the hash idea is interesting, I suspect that simply storing a `prev_id` is much cheaper than regularly recalculating hashes, especially if both cases require storing the recipient set.

proposals/4495-selective-presence.md
thetayloredman: & now, [MSC4514](https://github.com/matrix-org/matrix-spec-proposals/pull/4514)

proposals/4495-selective-presence.md
lveneris: This was explicitly considered in our drafting process, and decided against to keep the format consistent and to keep the system extensible in case other modes are desired in the future. Practically speaking, since unknown values are just ignored, this does not really matter.

proposals/4495-selective-presence.md
lveneris: It is also worth noting that anything you'd represent with text can be represented by an integer anyway. The only real difference is these have a bounded size, while a string would not, and I do not consider asking receiving servers to keep track of strings of potentially massive size acceptable in this case.

proposals/4495-selective-presence.md
lveneris: This was also explicitly considered in our drafting process, and we opted to take the approach laid out here to avoid polluting the sharing schema with states that are not a presence sharing state. Semantically, it does not make sense to include a "prompted" value there, and it would add to the work servers have to do to calculate a recipient set.

proposals/4495-selective-presence.md
Bentheburrito: That all makes sense, thanks for elaborating!

proposals/4495-selective-presence.md
Bentheburrito: I see, that makes sense. Maybe specifying the allowed values per mapping would make it read a little bit better, instead of saying each "may be any of the following[, unless it's `rooms` or `servers`]." e.g. something like

```suggestion
The string values of these presence sharing maps are defined as the following:
* `"users"`: `"allow" | "deny"`
* `"rooms"`: `"allow"`
* `"servers"`: `"deny"`

Where `"deny"` explicitly forbids sending presence to the entity, and `"allow"`
permits sharing presence with the entity. Additional values may be added in
future MSCs to support more use cases.
```

What do you think? Maybe the "MUST NOT"s could be kept, since they impact interpretation of the possible values as being strict. (Though if the goal is to be extensible maybe that's fine?)

lveneris: I do agree that this could be clearer. If you think it would achieve the same goal, I would prefer to flip them around, like so:
```md
The allowed string values for a presence sharing map are the following:
* `"deny"`: Explicitly forbid sending presence to the entity  
  Servers under the `"servers"` object MUST **only** use the `"deny"` state.
* `"allow"`: Permit sharing presence with the entity  
  Room IDs under the `"rooms"` object MUST **only** use the `"allow"` state.
```

Would this be acceptable to you? I think this is more explicit about the types.

Bentheburrito: That seems fine, though honestly the original wording is similar enough that it could go either way imo. To the point of the original comment, I was looking at this from the perspective of an implementer and wanting an explicit schema that limits room for error - I have noticed some Matrix structures/APIs are a bit awkward to implement, and could be simplified. (For example the `rooms` and `not_rooms` allow/block lists in `RoomEventFilter`s interact in weird ways).

I still like scoping the keys (`allowed_rooms`) to keep things specific to what is being introduced in this proposal, though I realize even if future proposals can extend things with new top-level keys, that then becomes annoying for implementers (have multiple places to check a room ID for)...all that to say, structuring things this way to keep it extensible makes sense to me. Thanks for the discussion :)

thetayloredman: [Here](https://github.com/ruma/ruma/pull/2546/changes#diff-1759953dd4700cf97b767a6ddf69421389a794bc353f940e47e3514cf94b5bd6R15-R52) is how it was done in Ruma, fwiw.

I'll leave it up to lveneris as to the final wording there.

proposals/4495-selective-presence.md
HarHarLinks: If a sufficiently large server were to survey the rooms it knows for the presence of this state event and value, that could be one factor to measuring success of this proposal?

proposals/4495-selective-presence.md
HarHarLinks: possibly https://github.github.com/gfm/#tables-extension-

but you might also do something simple that is primarily readable in the markdown source, like a bullet point list

proposals/4495-selective-presence.md
HarHarLinks: Hm, I think I get it, though I'm not sure if that behavior isn't a technical limitation rather than intended. I wouldn't be surprised if ignoring a user would cause message keys no longer being sent to them.

Anyway, I think my point is that profiles and presence aren't in the same area as messages: they don't even mean you're missing half the context of the convo in the room if you can't see whether someone is online. I think presence is "obviously" particularly relevant in terms of privacy. The spec could as an interim solution get written to contain a hint "if someone gets ignored, clients might also want to offer blocklisting them wrt presence". If you really disagree, I can accept this discussion getting postponed to #4283.

lveneris: I do believe the correct solution is for it to be determined by the access model of the system used at a broader level, ultimately. If the model of ignores is that you don't see their events but they see yours, which is currently the case, it is my view that we should follow that. In the same vein, if there is a block functionality that means neither party sees each other's events, if someone is blocked, I will expect them not to receive my presence. I hope this makes sense, although I can certainly see where you're coming from.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+[accepted all presence requests automatically][#4417], ultimately resulting in the feature's removal prior to r0.
+
+[MSC4325: Presence privacy][MSC4325] attempts to solve this problem by reintroducing a concept similar to Selective
+Presence. However, MSC4325 lacks some features integral to this proposal's goal of keeping UX similar to the
+current approach; room hints, for example, allow rooms to recommend that clients prompt users about whether they
+would like to share their presence, or forbid sharing presence altogether. This keeps presence usable for small
+communities, private rooms, and DMs, while reducing the load from federating presence in large public rooms.
+
+## Proposal
+
```
HarHarLinks, line 32:
```suggestion
## Proposal

The [presence module](https://spec.matrix.org/v1.19/client-server-api/#presence) of the Client-Server API and the [presence section](https://spec.matrix.org/v1.19/server-server-api/#presence) of the Server-Server API are adjusted as follows.

```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+### User Presence Sharing Preferences
+
+A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
```
HarHarLinks, line 39:
```suggestion
A new [`account_data`] entry, `m.presence.sharing`, is stored for modification by clients and use by homeservers.
Implementations of Matrix spec versions containing this feature MUST support this datum and associated behaviour.
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+}
+```
+
+### Room Presence Sharing Hints
+
+A new room state event is introduced, `m.room.presence_sharing`. This state event contains one field,
+`presence_sharing`, which is a string with two possible values: `"suggest"` and `"forbid"`. If the event is not
+present, the room is treated as though its value is `"forbid"` by default.
+
+Servers MUST NOT include rooms with a `presence_sharing` value of `"forbid"` when they calculate a user's recipient
```
HarHarLinks, line 81:
```suggestion
Implementations of Matrix spec versions containing this feature MUST support this datum and associated behaviour.
Servers MUST NOT include rooms with a `presence_sharing` value of `"forbid"` when they calculate a user's recipient
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+{
+    "users": [
+        "@joey:example.com",
+        "@chandler:example.com",
+    ],
+    "rooms": ["!apartment-group-chat"]
+}
+```
+
+Clients implementing presence sharing prompts SHOULD show a prompt resembling "Would you like to share your
```
HarHarLinks, line 105:
```suggestion
This is a purely client-side features and implementing it in clients is OPTIONAL.
Clients implementing presence sharing prompts SHOULD show a prompt resembling "Would you like to share your
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+
+Determine the delta from the previous recipient user set. If there is no previous state, the entire set is an addition.
+If the recipient user set is left unchanged, emit a presence EDU using only the prior `stream_id` \- no `recipients`, no
+`prev_id`. Otherwise, generate a `stream_id` for the new state. Emit a presence EDU for each destination of the
+recipients in these updates, with `prev_id` set to the last `stream_id` sent to the destination if applicable,
+`stream_id` set according to the new state, and `recipients` populated with the updates for the destination.
+
+#### Inbound Processing
+
+When a server receives a `m.presence` EDU over federation, servers follow these rules for distributing the presence
```
HarHarLinks:
```suggestion
When a server receives a `m.presence` EDU over federation, servers MUST follow these rules for distributing the presence
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+            }
+        ]
+    },
+    "edu_type": "m.presence"
+}
+```
+
+#### Dispatch Algorithm
+
+The **recipient user set** of a user's presence transition is determined as follows:
```
HarHarLinks:
```suggestion
The **recipient user set** of a user's presence transition MUST be determined as follows:
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+| room creation `preset` | `m.room.presence_sharing`           |
+| ---------------------- | ----------------------------------- |
+| `private_chat`         | `{ "presence_sharing": "suggest" }` |
+| `trusted_private_chat` | `{ "presence_sharing": "suggest" }` |
+| `public_chat`          | `{ "presence_sharing": "forbid" }`  |
+
+### Modifications to the [`m.presence` EDU]
+
+The federation [User Presence Update] type is modified to include:
+* An optional map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
```
HarHarLinks:
```suggestion
* An OPTIONAL map, `recipients`, containing two arrays of [User IDs][mxid-format] belonging to the destination
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+  server, `add` and `delete`.
+* A pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
```
HarHarLinks:
```suggestion
* A REQUIRED pair of integer identifiers, `stream_id` and `prev_id`, which are unique per `user_id`. These values do not need to
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+  be sequential or in any particular order, only unique.
+
+The `add` and `delete` arrays of `recipients` represent an incremental update to a user's recipient user set,
+adding and removing users respectively. A server MUST NOT send updates lacking a populated `add` to a destination
+homeserver after their last user is updated out of the recipient set.
+
+`stream_id` and `prev_id` form a sequence representing the state of the user's recipient user set, similar to the
+behaviour of the [Device List Update] type. `recipients` may only be present if `prev_id` is present. If a user's
```
HarHarLinks:
```suggestion
behaviour of the [Device List Update] type. `recipients` MUST NOT be present if `prev_id` is not present. If a user's
```

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,441 @@
...
+
+If `prev_id` is present but differs from the previous `stream_id`, or `stream_id` is present on its own but
+unrecognised, servers SHOULD query the `GET /_matrix/federation/v1/query/presence_recipients` endpoint on the
+origin server.
+
+### Server-Server Endpoint `GET /_matrix/federation/v1/query/presence_recipients`
+
+Servers that do not recognise a `stream_id` or `prev_id` in an incoming `m.presence` EDU need a way to ask the
+origin server for the user's latest `stream_id` and a snapshot of its recipient user set. This proposal introduces
+a new federation endpoint to address this need. If a server cannot determine the proper recipient user set \- for
```
HarHarLinks:
```suggestion
a new REQUIRED federation endpoint to address this need. If a server cannot determine the proper recipient user set \- for
```

HarHarLinks (commented):
I suggest to clarify requirement levels for each component

proposals/4495-selective-presence.md
lveneris: Is this necessary? I would think anything that is marked by "MUST" or not clarified as an optional feature otherwise is a requirement. I have not seen this in other proposals, I know that much.

If it is, I would rather clarify this once at the top of the proposal than write it multiple times.

lveneris:
That force push was unintentional and I am not sure where it came from or why it happened. I must now unfortunately make another to rectify it. My sincerest apologies.

proposals/4495-selective-presence.md
HarHarLinks: > I have not seen this in other proposals, I know that much.

Agreed, and I have recently found some issues where an MSC was written without being clear about the requirement levels, and then spec always being written without requirement levels. Reading both with an understanding of "not clarified as an optional feature otherwise is a requirement" turned out incorrect. My conclusion is that proposals (and spec text) should be abundantly clear about this.

Your 889e420181628ad5bcbc533fb456d228c5bf834c solution works for me.

lveneris: I just really didn't want to throw boilerplate all over the place, that's all. I welcome the change if it means Selective Presence has a greater chance of landing the way it is intended to.

martinezvelasco25-art:
I

martinezvelasco25 te invitó a unirte a Ihf

Aceptar invitación [https://join.slack.com/t/ihfespacio/invite/enQtMTE4NTI0OTU5MzU4OTItMmViYjkwMmU5ZjU1YmJkN2E3ODkxYWYzNWMzNzg0MDJjMTY4NmRiMjMxZmJkZWM0ZmMwMjFmOTA0ZjZiODAwMQ?x=x-p11850754013473-11852489625556-11833443566071&t=accept_invitation_cta]

Gratis y sin necesidad de configuración.

Esta invitación vence en 30 días.

Invitación de ***@***.*** Haz clic en la siguiente URL para ver el espacio de trabajo en cualquier momento: ihfespacio.slack.com [https://ihfespacio.slack.com/].

Slack es una herramienta conversacional para reunir equipos.

Traslada las conversaciones aisladas en hilos de correo electrónico al mismo lugar. Slack te ayuda a organizar las comunicaciones para acelerar la toma de decisiones.

martinezvelasco25-art:
Tu invitación se envió ayer

I

Acepta la invitación de martinezvelasco25 para trabajar en equipo en Ihf

Aceptar invitación [https://join.slack.com/t/ihfespacio/invite/enQtMTE4NTI0OTU5MzU4OTItMmViYjkwMmU5ZjU1YmJkN2E3ODkxYWYzNWMzNzg0MDJjMTY4NmRiMjMxZmJkZWM0ZmMwMjFmOTA0ZjZiODAwMQ?x=x-p11850754013473-11852489625556-11856393403381&t=accept_invitation_cta]

Gratis y sin necesidad de configuración.

Esta invitación vence en 29 días.

Enviado por ***@***.*** en ihfespacio.slack.com [https://ihfespacio.slack.com/].

Slack es una herramienta conversacional para reunir equipos.

Traslada las conversaciones aisladas en hilos de correo electrónico al mismo lugar. Slack te ayuda a organizar las comunicaciones para acelerar la toma de decisiones.

nyabinary:
@dasha-uwu I see you and other Tuwnel maintainers downvoted the proposal, what is the reasoning why/what do you disagree with in this proposal?

martinezvelasco25-art:
Tu invitación se envió hace 7 días

I

No te pierdas las actualizaciones de martinezvelasco25 en Ihf

Aceptar invitación [https://join.slack.com/t/ihfespacio/invite/enQtMTE4NTI0OTU5MzU4OTItMmViYjkwMmU5ZjU1YmJkN2E3ODkxYWYzNWMzNzg0MDJjMTY4NmRiMjMxZmJkZWM0ZmMwMjFmOTA0ZjZiODAwMQ?x=x-p11850754013473-11852489625556-11888810741447&t=accept_invitation_cta]

Gratis y sin necesidad de configuración.

Esta invitación vence en 23 días.

Enviado por ***@***.*** en ihfespacio.slack.com [https://ihfespacio.slack.com/].

Slack es una herramienta conversacional para reunir equipos.

Traslada las conversaciones aisladas en hilos de correo electrónico al mismo lugar. Slack te ayuda a organizar las comunicaciones para acelerar la toma de decisiones.

proposals/4495-selective-presence.md
```diff
@@ -0,0 +1,437 @@
...
+presence EDUs for every user they share a room with, even if the users have never interacted or have no interest in
+each other's presence; server operators often have no choice but to disable presence altogether to limit the
+excessive federation load; and users who do _not_ want to share potentially trackable information with public,
+federated rooms \- a limitation [acknowledged by the current specification][presence-v1-privacy] \- the only option
+they typically have is to, again, disable presence outright.
+
+This proposal introduces "Selective Presence," a system for users to decide precisely _who_ their
+presence is shared with. The benefits of this are twofold:
+1. the overall load on homeservers implementing presence is reduced substantially, as users only share their
+   presence with interested parties, and
```
gamesguru, line 15: Hi, I've looked in-depth into presence performance (specifically in `conduwuit` / `continuwuity`). Synapse is generally a mature product and difficult to innovate on, but the hot paths around `get_interested_remotes() -> get_rooms_for_user() -> get_current_hosts_in_room()` and `update_presence` can also be significantly improved. It doesn't really leverage deferred coalescing or in-memory cache to the fullest.

Another idea (probably would blow up the scope, but for a future MSC maybe) if resending redundant/duplicate info is hard to avoid in some paths, is to leverage MSC4521 set reconciliation: identify each EDU with a unique `EduID` and cap the JSON exchanges (both C2S and federated) to only sync configs which have _changed_ or users whose `last_active_ts` has bumped or `status_msg` has changed. This would pave the way for broader _gossip-based_ EDU reconciliation, useful since EDUs like read receipts and status changes today are fire-and-forget and can easily slip thru the cracks or get lost in transit. Non-origin sources would need to forward origin signatures, or else the origin would have to attest to IDs over HTTP since authenticity is otherwise unassured.

gamesguru (commented):
Just a performance note. There may also be some trie-like data structures to help with the domain/localpart globbing, although this is unclear and follow-up material at best.

PR #4496, opened 2026-06-26
MSC4496: Calendar Events, Invites, and Availability in Matrix

MTRNord:
[Rendered](https://github.com/MTRNord/matrix-doc/blob/matrix-caldav/proposals/4496-matrix-caldav.md)

Implementations:

- Client: none
- Server: none
- Bridge: none
- SDK: none

Signed-off-by: MTRNord <support@midnightthoughts.space>

### DIFF
--- a/proposals/4496-matrix-caldav.md
+++ b/proposals/4496-matrix-caldav.md
@@ -1,0 +1,785 @@
+# MSC 4496: Calendar Events, Invites, and Availability in Matrix
+
+_This proposal introduces a general-purpose calendaring system for Matrix, covering event
+creation and sharing, invite/RSVP flows, attendee tracking, location and video conference
+attachments, recurrence, and user availability. It is intended to supersede the earlier
+attempt in MSC1116 and to incorporate the community feedback that drove that proposal toward
+an iCalendar-compatible semantic model. It deliberately avoids a meetings-specific scope so
+that higher-level applications (including video conferencing schedulers) can build on top._
+
+Matrix currently has no first-class representation of calendar data. Users who wish to schedule
+events with others over Matrix must either use an out-of-band calendar system and paste details
+into chat, or rely on bespoke application-layer solutions that do not interoperate. This creates
+friction, breaks privacy (invites often travel via corporate email even when both parties use
+Matrix), and prevents Matrix from serving as a complete communication stack for organisations. A
+general, privacy-respecting, federated calendaring primitive in Matrix would fill this gap while
+also enabling transparent interop with the iCalendar/CalDAV ecosystem that the majority of the
+world already uses.
+
+## Proposal
+
+### Design philosophy
+
+This proposal is guided by four core principles:
+
+1. **Matrix-native first.** Events are Matrix timeline and state events, not iCalendar payloads
+   embedded in Matrix. iCalendar/CalDAV compatibility is a translation concern at the bridge or
+   client layer.
+
+2. **Extensible events throughout.** All event types defined here use the content-block model
+   from [MSC1767](https://github.com/matrix-org/matrix-spec-proposals/pull/1767). Clients
+   without calendar support can render a meaningful `m.text` fallback. The relationship between
+   this MSC and encryption is discussed in §7.
+
+3. **Semantic alignment with JSCalendar.** Rather than inventing new field semantics, this
+   proposal deliberately stays close to the data model of JSCalendar (RFC 8984 /
+   draft-ietf-calext-jscalendarbis), which is itself designed to be a modern, JSON-native
+   alternative and eventual successor to iCalendar (RFC 5545). This makes round-tripping with
+   iCalendar and CalDAV straightforward while keeping the representation idiomatic Matrix JSON.
+   Where JSCalendar and iCalendar semantics agree, we follow them; where JSCalendar improves on
+   iCalendar (e.g. storing named timezones alongside UTC rather than UTC-only), we follow
+   JSCalendar.
+
+4. **Privacy by design (GDPR-compatible).** Availability queries reveal only time-slot
+   occupancy, not event titles or attendee lists, unless the queried user explicitly grants
+   more. No personally identifying information is required in federation-visible room state.
+
+---
+
+### 1. Calendar rooms
+
+A **calendar room** is a Matrix room with `m.room.type` set to `m.calendar`. It acts as a
+container for a user's or group's calendar events. Each user's personal calendar is a
+private (invite-only) `m.calendar` room on their homeserver. Shared/group calendars are
+`m.calendar` rooms with multiple members.
+
+#### 1.1 Room creation
+
+```json
+{
+  "type": "m.room.create",
+  "content": {
+    "type": "m.calendar",
+    "m.federate": true
+  }
+}
+```
+
+Calendar metadata is stored in a dedicated state event:
+
+```json
+{
+  "type": "m.calendar.info",
+  "state_key": "",
+  "content": {
+    "m.text": [{ "body": "Personal calendar", "mimetype": "text/plain" }],
+    "org.matrix.msc4496.calendar_info": {
+      "color": "#3c82f6",
+      "timezone": "Europe/Berlin"
+    }
+  }
+}
+```
+
+The `m.text` block (MSC1767) MUST be present as a human-readable fallback for unaware clients.
+
+---
+
+### 2. Calendar event type: `m.calendar.event`
+
+A calendar event is a **timeline message event** sent into a calendar room. Using timeline
+events preserves the append-only audit log, supports reactions, threading, and E2EE without
+special-casing.
+
+#### 2.1 Content structure
+
+```json
+{
+  "type": "m.calendar.event",
+  "content": {
+    "m.text": [
+      {
+        "body": "Team standup – 2026-07-01 09:00–09:30 (Europe/Berlin)",
+        "mimetype": "text/plain"
+      }
+    ],
+    "org.matrix.msc4496.calendar_event": {
+      "uid": "550e8400-e29b-41d4-a716-446655440000",
+      "title": "Team standup",
+      "start": {
+        "utc": "2026-07-01T07:00:00Z",
+        "localtime": "2026-07-01T09:00:00",
+        "timezone": "Europe/Berlin"
+      },
+      "end": {
+        "utc": "2026-07-01T07:30:00Z",
+        "localtime": "2026-07-01T09:30:00",
+        "timezone": "Europe/Berlin"
+      },
+      "all_day": false,
+      "status": "confirmed",
+      "transparency": "opaque",
+      "description": "Daily sync with engineering team.",
+      "organizer": "@alice:example.com",
+      "location": null,
+      "conference": null,
+      "recurrence": null
+    }
+  }
+}
+```
+
+**Field definitions:**
+
+- `uid`: Stable UUID v4 (RFC 4122) that persists across edits, recurrences, and
+  CalDAV round-trips. Maps to iCalendar `UID` and JSCalendar `uid`.
+- `start` / `end`: Both UTC and named IANA timezone local time are stored. This mirrors
+  JSCalendar's approach and avoids the UTC-only pitfall where stored times become incorrect
+  when a timezone's DST rules change in the future (a known iCalendar pain point). Named IANA
+  timezone identifiers MUST be used; numeric offsets alone are not acceptable.
+- `transparency`: `"opaque"` (blocks availability) or `"transparent"` (does not block).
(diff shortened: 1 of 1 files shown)


proposals/4496-matrix-caldav.md
turt2live, line 1: Implementation requirements:

* [ ] Server

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+## Dependencies
+
+This MSC builds on:
+
+- [MSC1767](https://github.com/matrix-org/matrix-spec-proposals/pull/1767) – Extensible events
+  (required for `m.text` fallback blocks in all event types).
+- [MSC2674](https://github.com/matrix-org/matrix-spec-proposals/pull/2674) – Event
+  relationships (required for recurrence overrides and RSVP threading).
+- [MSC2676](https://github.com/matrix-org/matrix-spec-proposals/pull/2676) – Message edits
+  (required for cancellation and event updates).
```
thetayloredman, line 719: These MSCs are all accepted into the spec. Typically these are not listed as "hard dependencies."

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+- [MSC1840](https://github.com/matrix-org/matrix-spec-proposals/pull/1840) – Typed rooms /
+  `m.room.type` (required for the `m.calendar` room type).
```
thetayloredman, line 721: This MSC was rejected. The `type` field was introduced in MSC1772, which is, again, accepted.

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+that higher-level applications (including video conferencing schedulers) can build on top._
+
+Matrix currently has no first-class representation of calendar data. Users who wish to schedule
+events with others over Matrix must either use an out-of-band calendar system and paste details
+into chat, or rely on bespoke application-layer solutions that do not interoperate. This creates
+friction, breaks privacy (invites often travel via corporate email even when both parties use
+Matrix), and prevents Matrix from serving as a complete communication stack for organisations. A
+general, privacy-respecting, federated calendaring primitive in Matrix would fill this gap while
+also enabling transparent interop with the iCalendar/CalDAV ecosystem that the majority of the
+world already uses.
```
thetayloredman, line 17: I still do not understand the use case for this feature. A more fleshed-out rationale (there is interesting value here in e.g. federation and E2EE) would be useful besides just "it's not Matrix."

MTRNord: Will update that in the next days.

The gist is this perspective I am coming from:

- I want something I can use at work that replaces Teams entirely -> Calendaring is missing currently
  - Calendaring in Teams/Outlook is also deeply integrated and not just CalDAV bolted on.
- I want to be able to share events in a group, invite someone via matrix where I don't know their email and I want proper RVSP
- I do recognize that the internet mainly will use CalDAV or CalDAV adjacent protocols. There will be no time where everyone is using Matrix. Therefor I want a solution that is compatible with that and allows easy conversion in either direction (hence the appendix)
  - Additionally in an enterprise setting you likely still have to use email invites so having a bridge possibility here is useful.


As probably is obvious I am coming a lot from the Teams/Enterprise usecase and much less so from a personal usecase here. I don't think it is something everyone wants or needs in their client and that's perfectly reasonable. I can also see partial implementations like event calendars like discourse features for example.

E2EE is "free" in matrix. So that section for the most part exists just for clarity that it is possible to be used without too much effort.

The endpoints for availability are so that there is less information leak. Sure you could do this client side as well but then you cant easily make the homeserver hide hidden events from the other side. This way it is a bit easier to preserve privacy which imho is very important for a calendar. However, I did not want to implement all of CalDAV in matrix. That feels like the wrong direction. Hence historically matrix used bridges to map to third party networks. CalDAV I consider a third party network here. So I kept the API surface in matrix to the absolute bare minimum and instead tried to do as much as possible just using events.

That all being said I will bring this properly into the MSC itself as soon as I get the time :)

proposals/4496-matrix-caldav.md
MTRNord: Will update that. It has been a while since I wrote an MSC and much less so one of this size :)

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
+# MSC 4496: Calendar Events, Invites, and Availability in Matrix
+
+_This proposal introduces a general-purpose calendaring system for Matrix, covering event
+creation and sharing, invite/RSVP flows, attendee tracking, location and video conference
+attachments, recurrence, and user availability. It is intended to supersede the earlier
+attempt in MSC1116 and to incorporate the community feedback that drove that proposal toward
+an iCalendar-compatible semantic model. It deliberately avoids a meetings-specific scope so
+that higher-level applications (including video conferencing schedulers) can build on top._
```
Johennes, line 8: Linkifying might help readability here.

```suggestion
_This proposal introduces a general-purpose calendaring system for Matrix, covering event
creation and sharing, invite/RSVP flows, attendee tracking, location and video conference
attachments, recurrence, and user availability. It is intended to supersede the earlier
attempt in [MSC1116] and to incorporate the community feedback that drove that proposal toward
an iCalendar-compatible semantic model. It deliberately avoids a meetings-specific scope so
that higher-level applications (including video conferencing schedulers) can build on top._

[MSC1116]: https://github.com/matrix-org/matrix-spec-proposals/pull/1116
```

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+This proposal is guided by four core principles:
+
+1. **Matrix-native first.** Events are Matrix timeline and state events, not iCalendar payloads
+   embedded in Matrix. iCalendar/CalDAV compatibility is a translation concern at the bridge or
+   client layer.
+
+2. **Extensible events throughout.** All event types defined here use the content-block model
+   from [MSC1767](https://github.com/matrix-org/matrix-spec-proposals/pull/1767). Clients
+   without calendar support can render a meaningful `m.text` fallback. The relationship between
+   this MSC and encryption is discussed in §7.
```
Johennes, line 32: Given that extensible events are still unreleased, this fallback will only work in future clients while current clients would probably just not display the events at all?

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+
+Calendar metadata is stored in a dedicated state event:
+
+```json
+{
+  "type": "m.calendar.info",
+  "state_key": "",
+  "content": {
+    "m.text": [{ "body": "Personal calendar", "mimetype": "text/plain" }],
+    "org.matrix.msc4496.calendar_info": {
```
Johennes, line 76: I think for readability, the snippets shouldn't include the unstable prefixes. The section at the very end can just provide the mapping.

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+      "color": "#3c82f6",
+      "timezone": "Europe/Berlin"
```
Johennes, line 78: What are the semantics of `timezone`, especially for shared / group calendars?

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+Calendar metadata is stored in a dedicated state event:
+
+```json
+{
+  "type": "m.calendar.info",
+  "state_key": "",
+  "content": {
+    "m.text": [{ "body": "Personal calendar", "mimetype": "text/plain" }],
+    "org.matrix.msc4496.calendar_info": {
+      "color": "#3c82f6",
```
Johennes, line 77: Would color not rather be something to be determined on the displaying client? What if I display two calendars that happened to pick the same `color` value?

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+```json
+"location": {
+  "type": "virtual",
+  "name": "Jitsi Meet",
+  "uri": "https://meet.example.com/standup"
+}
+```
+
+For virtual locations, `uri` is a plain HTTPS URL rather than a `geo:` URI. Clients SHOULD
+display virtual locations as clickable links.
```
Johennes, line 190: I think the spec may not have to be concerned with how clients display the content. A CLI client may not even be able to display clickable links.

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+}
+```
+
+The `room_id` MUST be the Matrix room ID of an active MatrixRTC-enabled room. The
+`via` field is a list of server names through which the room can be joined, following the
+same convention as `m.space.child` state events and `matrix.to` links. Clients MUST include
+at least one `via` entry when constructing this block; a `conference` block without a `via`
+array SHOULD be treated as unresolvable and clients SHOULD disable the join button rather
+than silently failing. The call room is managed independently of the calendar event; this
+MSC does not specify how it is created. Clients SHOULD display the room name with a fallback to the room ID and
```
Johennes, line 214: What does "display the room name with a fallback to the room ID" mean?

proposals/4496-matrix-caldav.md
```diff
@@ -0,0 +1,785 @@
...
+"Join call" action and allows bridges to emit a `CONFERENCE` property (RFC 9073) in
+iCalendar output.
+
+```json
+"conference": {
+  "type": "m.call",
+  "room_id": "!callroom:example.com",
+  "via": ["example.com", "matrix.org"],
+  "label": "Join call"
+}
```
Johennes, line 205: MatrixRTC can have several slots per room. Maybe we should include the `slot_id` from `m.rtc.slot` here, too? Also, since MatrixRTC has its own `m.call` application type and will probably have further application types in future, maybe the `type` here should be something more generic such as `m.rtc`?

proposals/4496-matrix-caldav.md
MTRNord: So the main reason this exists is from caldav spec where this is as well defined by the calendar rather than clients (though clients sometimes allow overriding it locally).

In this MSC it mostly is carried over for the interop capabilities. Though I think it would be fine if clients can just override it locally if they prefer. I will try to spell that out :)

Though still open to discuss this further.

proposals/4496-matrix-caldav.md
MTRNord: Thanks :) Fixed it.

proposals/4496-matrix-caldav.md
MTRNord: I added some text in [54e6001](https://github.com/matrix-org/matrix-spec-proposals/pull/4496/commits/54e6001ec27692be1468c57275345ed49c0f93b4) explaining this.

Its another thing from CalDav world. Its essentially the default setting for a calendar. It can be overridden on events and can be used when rendering to define what "home" TZ it should render things in.
In terms of contents its the usual IANA timezone identifier format.

proposals/4496-matrix-caldav.md
MTRNord: dropped the client recommendation in [daae001](https://github.com/matrix-org/matrix-spec-proposals/pull/4496/commits/daae001b8115276a0e5806e49773a3a143564b31)

proposals/4496-matrix-caldav.md
MTRNord: I tried to clarify this in [3c32e47](https://github.com/matrix-org/matrix-spec-proposals/pull/4496/commits/3c32e47efb804816ad182537cfe171f61333053c) and hope that it makes more sense :)

proposals/4496-matrix-caldav.md
MTRNord: I updated it in [9a67780](https://github.com/matrix-org/matrix-spec-proposals/pull/4496/commits/9a67780049655b18c17af645b69b24027896af27) which hopefully is closer to whats the MatrixRTC reality we currently have :)

proposals/4496-matrix-caldav.md
MTRNord: At this time yes thats (sadly) expected. I dont think its necessarily an issue for now though. It certainly could be nice to have fallback too in the current event format but I think that would bloat the MSC quite a bit.

proposals/4496-matrix-caldav.md
MTRNord: Updated the MSC in [a4f4b30](https://github.com/matrix-org/matrix-spec-proposals/pull/4496/commits/a4f4b3026f8bc86ffc63222c587393d82c6a1ed9) to reflect this better now :) I hope that makes things a bit clearer in terms of intent and motivation I had to write this.

PR #4497, opened 2026-06-27
MSC4497: Filter state events by type in /state endpoint

pernydev:
[Rendered](https://github.com/pernydev/matrix-spec-proposals/blob/main/proposals/4497-state-event-type-filter.md)

Implementations:
- Synapse: https://github.com/element-hq/synapse/pull/19899

Signed-off-by: Perny <hey@perny.dev>

### DIFF
--- a/proposals/4497-state-event-type-filter.md
+++ b/proposals/4497-state-event-type-filter.md
@@ -65,15 +65,15 @@ filtering concept.
 
 ### Use /sync with a filter
 
-Clients could use `/sync` with a `RoomFilter` that specifies the desired `rooms` and a `StateFilter`
+Clients could use `/sync` with a `RoomFilter` that specifies the desired `rooms` and a `RoomEventFilter`
 with the desired `types`. However, the `/sync` response structure is significantly more complex than
 `/state`, and filtering to a single room requires either listing it explicitly or using a filter ID
 created via `POST /filter`.
 
-### Accept a full StateFilter
+### Accept a full RoomEventFilter
 
-Instead of just `types`, the endpoint could accept a full `StateFilter` object (with `types`,
-`not_types`, `senders`, `not_senders`, etc.). However, for most usecases `StateFilter` provides
+Instead of just `types`, the endpoint could accept a full `RoomEventFilter` object (with `types`,
+`not_types`, `senders`, `not_senders`, etc.). However, for most usecases `RoomEventFilter` provides
 a lot more than what is needed to read state.
 
 ### Add individual query parameters per type
--- a/proposals/4497-state-event-type-filter.md
+++ b/proposals/4497-state-event-type-filter.md
@@ -1,0 +1,100 @@
+# Filter state events by type in /state endpoint
+
+Currently, the `GET /_matrix/client/v3/rooms/{roomId}/state` endpoint returns all state events in a
+room. For rooms with many state events, this can be inefficient when a client only needs events of
+specific types.
+
+This proposal adds an optional `types` query parameter to the `/state` endpoint, allowing clients to
+request only state events matching the specified event types.
+
+
+## Proposal
+
+Add an optional `types` query parameter to the `GET /_matrix/client/v3/rooms/{roomId}/state` endpoint.
+
+### Query Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| `types` | `[string]` | A JSON-encoded array of event types to include. If this list is absent then all event types are included. A `*` can be used as a wildcard to match any sequence of characters. |
+
+This follows the same semantics as the `types` field in `RoomEventFilter`:
+
+> A list of event types to include. If this list is absent then all event types are included. 
+> A `*` can be used as a wildcard to match any sequence of characters.
+
+### Examples
+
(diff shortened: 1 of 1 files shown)


proposals/4497-state-event-type-filter.md
turt2live, line 1: Implementation requirements:

* [ ] Client (using)
* [ ] Server (serving)

turt2live, line 1: @pernydev For this MSC to be eligible for a future FCP merge it will need [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md)

turt2live (changes requested):
This review is for early requirements gathering - it's not a review of the MSC itself.

proposals/4497-state-event-type-filter.md
pernydev: Added signoff to PR message :)

pernydev: Will a reverse proxy based server implementation suffice for this MSC? I am personally not very familiar with the codebase of any major server implementation, so contributing to them would be difficult.

The limitation proxy would have here is the loss of the database efficiency angle, since a homeserver would need to still load all states.

turt2live: Implementations don't need to be done by the MSC author. For an MSC like this we're primarily looking for evidence that a use case exists and that this MSC fulfills that. This ideally means that multiple clients require support for the feature, and at least one adds support for it. A proper server implementation would be best to demonstrate that the solution is desirable to the ecosystem.

Half-Shot: Consider it done https://github.com/element-hq/synapse/pull/19899

proposals/4497-state-event-type-filter.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+## Alternatives
+
+### Use /sync with a filter
+
+Clients could use `/sync` with a `RoomFilter` that specifies the desired `rooms` and a `StateFilter`
+with the desired `types`. However, the `/sync` response structure is significantly more complex than
+`/state`, and filtering to a single room requires either listing it explicitly or using a filter ID
+created via `POST /filter`.
```
Johennes: I think the type in the spec is [`RoomEventFilter`](https://spec.matrix.org/v1.18/client-server-api/#post_matrixclientv3useruseridfilter_request_roomeventfilter), not `StateFilter`?

(Same in the paragraph below.)

```suggestion
Clients could use `/sync` with a `RoomFilter` that specifies the desired `rooms` and a [`RoomEventFilter`]
with the desired `types`. However, the `/sync` response structure is significantly more complex than
`/state`, and filtering to a single room requires either listing it explicitly or using a filter ID
created via `POST /filter`.

[`RoomEventFilter`]: https://spec.matrix.org/v1.18/client-server-api/#post_matrixclientv3useruseridfilter_request_roomeventfilter
```

proposals/4497-state-event-type-filter.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Accept a full StateFilter
+
+Instead of just `types`, the endpoint could accept a full `StateFilter` object (with `types`,
+`not_types`, `senders`, `not_senders`, etc.). However, for most usecases `StateFilter` provides
+a lot more than what is needed to read state.
```
Johennes: Given that we already have [`RoomEventFilter`](https://spec.matrix.org/v1.18/client-server-api/#post_matrixclientv3useruseridfilter_request_roomeventfilter) established on `/sync` and `/messages`, I wonder if reusing it wouldn't be the cleaner design? It has no required fields. So even though it involves significantly more options than the single query parameter, clients wouldn't be forced to use all of them. Servers should also already have support for it, which should simplify the implementation.

pernydev: Yeah that is true and something I thought of while writing this. I'd love to hear what people think about the usefulness of `RoomEventFilter` for state, it's a conversation worth having here.

proposals/4497-state-event-type-filter.md
pernydev: This should be fixed now.

proposals/4497-state-event-type-filter.md
```diff
@@ -0,0 +1,100 @@
...
+### Query Parameters
+
+| Name | Type | Description |
+|------|------|-------------|
+| `types` | `[string]` | A JSON-encoded array of event types to include. If this list is absent then all event types are included. A `*` can be used as a wildcard to match any sequence of characters. |
+
+This follows the same semantics as the `types` field in `RoomEventFilter`:
+
+> A list of event types to include. If this list is absent then all event types are included. 
+> A `*` can be used as a wildcard to match any sequence of characters.
```
reivilibre, line 24: For a server implementer, this hinders the ability to optimise the implementation. For that reason I'm not a fan; it's also not clear to me what a good motivating use case is.

I'd personally suggest the removal of this wildcard functionality. A later MSC could propose it as a `type_wildcard=...` param if we think it's worthwhile.

PR #4500, opened 2026-07-05
MSC4500: State accumulator endpoint and transaction digests

gamesguru:
**Update 2026/08/05:** Pushing some minor edits (soon-ish), but the document is in good shape. I'll advise implementations break it into two phases:

1. (Estimated developer effort: 3-4 days) Implement the database columns, add `rezzy` / `gomatrixcrypto` dependency or library implementation of choice, and return the BLAKE digest in `txn` requests (minimum to pass Complement tests).
2. [Later phase, deferred]. (Est. effort: 5-25 days) Use suggested optimizations/performance gains in real-time federation and other relevant code places. Emit alerts or logs when digests mismatch (this will cause noise at first, even between Synapse servers, since we all have state issues in bigger rooms). Decide the role of the "bisection" strategy or pathway, whether it's worth implementing, and if so, as real-time federation or only in an admin command.

Reviewers may wish to use my preliminary self-review as a series of guiding questions:

- What benefits does a cryptographic accumulator provide over a less secure one? Do we need that protection here? Can it be falsified/spoofed/misrepresented by a dishonest node anyway? What makes `LtHash` a better choice than `MuHash3072`? Is there an alternative which makes the subtraction/addition more intuitive and less error-prone, i.e., a _non_-multiset, plain set, implementation? Should an auxiliary `checksum` or strict database transactions be advised/required to prevent corruption/drift during power loss or potential software bugs?
- How practical is the bisection strategy proposed here? Does the full 2 KB `/state_accumulator` endpoint have any benefit over 32-byte BLAKE digest comparison? How does it compare or compete with and is it nullified by my MSC4511 or MSC4521 proposals with similar ability to determine differences or divergence?
- Is the `LtHash` input encoding ideal? Why not just use `$eventID`, why use the `msc4500\x00` DST-prefixed tuple `(type, state_key, event_id)`? What's the advantage of this, other than visually resembling Synapse's state group layout?
- What are the approximate storage requirements and expected performance gains? Is there a point where small rooms cease to benefit? Are smaller state groups (< 20 elements) better to maintain the old way, than by a 2 KB accumulator?
- Are migrations advised? If not, is back-indexing of rooms optional? Is this only strictly required on future room versions? Explain how this could be rolled out, and assess whether larger deployment would be okay with this or if it would require prohibitively long downtime (e.g., 10 million rooms, normally distributed).

---

A back-end optimization and early-warning system for room state divergence. Based on prior art.

Byte-for-byte compatible incremental global state accumulator used by Facebook and Ethereum.

[Rendered](https://github.com/gamesguru/matrix-spec-proposals/blob/guru/4500-state-accumulators/proposals/4500-state-accumulators.md).

### Implementations

- **Library:**
    + [LtHash16 accumulator Rust module (gamesguru/rezzy)](https://gitlab.com/gamesguru/rezzy/-/blob/master/src/state/lthash.rs)
    + [LtHash16 Golang primitives (Wombat-Foundation/gomatrixcrypto)](https://gitlab.com/wombat-foundation/gomatrixcrypto/-/blob/main/lthash/lthash.go)

- **Complement tests:** [complement/tests/msc4500 at main · gamesguru/complement](https://github.com/gamesguru/complement/tree/main/tests/msc4500)

- **Homeserver:** [Implement MSC4500 (State accumulators) [on top of development branch] gamesguru/continuwuity](https://github.com/gamesguru/continuwuity/pull/84)

### DIFF
--- a/proposals/4500-state-accumulators.md
+++ b/proposals/4500-state-accumulators.md
@@ -1,0 +1,636 @@
+# MSC4500: State accumulator endpoint and transaction digests
+
+Matrix servers replicate a room as a DAG of events and rely on state resolution
+to eventually converge on a shared state. When servers diverge, the result can
+be a serious nuisance. Matrix lacks an out-of-band or real-time mechanism for
+state verification or re-alignment; servers often only learn of
+de-synchronization once they disagree on a much later authorization failure
+(e.g., another user's join is incorrectly rejected).
+
+I present an "early-warning system" which rapidly confirms incremental state
+consensus, or signals that divergence exists, so servers know they share the
+exact same view of a room at a given point in the DAG.
+
+This proposal does not impose any verification requirements on PDU handling. It
+seeks to act as a secondary state convergence mechanism, while simultaneously
+**relegating state group transitions** and naive iterative BFS implementations
+to storage/retrieval with a cheap, bitwise, commutative, subtractable (supports
+element removal), collision-resistant 2048-byte `LtHash16` accumulator function
+[^3], [^6]. Similar additive lattice accumulators are increasingly used in
+production blockchain architectures to compute real-time, incremental
+cryptographic state commitments under high transactional volume [^5], [^6].
+
+Avoiding diff chain reconstruction for point lookups will reduce Synapse's
+electricity consumption across a wide range of API state endpoints.
+
+The accumulator under question may be called 'homomorphic' and solves the
+following hashing problem: "Given the hash of an input, along with a small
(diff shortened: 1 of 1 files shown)


proposals/4500-state-accumulators.md
turt2live, line 1: Implementation requirements:

* [ ] Complement tests
* [ ] Server (multiple)

It's unclear to me at a glance if this would require a room version or not.

PR #3639, opened 2022-01-12
WIP: matrix for the social media use case

Henri2h:
[Rendered](https://github.com/Henri2h/matrix-doc/blob/henri2h-matrix-for-social-media/proposals/MSCZZZZ-matrix-for-the-social-media-use-case.md)

Thanks, @MTRNord and @frandavid100 for your help writing this MSC.

PR #4501, opened 2026-07-06
MSC4501: Social Media on Matrix

q-TheDog:
[Rendered](https://github.com/Haven-Organization/matrix-spec-proposals/blob/msc-social-media-pages/proposals/4501-rooms-as-social-media-pages.md)

A spiritual successor to [MSC3639: Matrix for the social media use case](https://github.com/matrix-org/matrix-spec-proposals/pull/3639) (abandoned). This MSC provides the event/room/profile types, conventions, and guidelines to build a modern social media app on Matrix:
- Profiles
- Groups
- Posts/Reposts/Quote Posts/Replies
- Likes
- Repost/Like count aggregation
- Discoverability 
- Interoperability with existing clients

Signed off by: q havensoftware@zohomail.com

Implementations:
- Client ([Haven Desktop](https://github.com/Haven-Organization/haven-desktop)):
  - https://github.com/Haven-Organization/haven-desktop/tree/develop/src/apps/social
  - https://github.com/Haven-Organization/haven-desktop/blob/develop/src/apps/social/utils/room-classifier.ts
- Bridge ([matrix-appservice-activitypub](https://github.com/Haven-Organization/matrix-appservice-activitypub)):
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/note_mirroring.py
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/inbox_dispatch.py
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/reaction_bridge.py

### DIFF
--- a/proposals/4501-rooms-as-social-media-pages.md
+++ b/proposals/4501-rooms-as-social-media-pages.md
@@ -22,7 +22,8 @@ client can build a social media experience on top of:
 - Room membership doubling as "following", with no new event needed to represent it.
 - A post event type that is structurally identical to `m.room.message`, so existing message-handling
   code can be reused with no new parsing logic.
-- A minimal, single mechanism for reposts, boosts, retweets, and quote-posts.
+- A minimal, generic mechanism for referencing another post across rooms, covering reposts, boosts,
+  retweets, and quote-posts, as well as cross-posting a reply into your own profile feed.
 - Reusing the existing `m.reaction` mechanism for "liking" a post, rather than a new event type.
 - Guidance for how a client builds a "feed" out of these rooms.
 - A phased rollout plan for when clients should treat `m.room.message` as an interchangeable stand-in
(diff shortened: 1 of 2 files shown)


proposals/4501-rooms-as-social-media-pages.md
turt2live, line 1: Implementation requirements:

* [ ] Chat client (backwards compatibility), or MSC mechanisms to prevent/discourage chat clients from encountering the data
* [ ] Social media client (first party compatibility, multiple)

turt2live, line 1: This MSC requires [sign off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md#sign-off) before it can enter a proposed FCP merge.

turt2live (changes requested):
Early review for early process requirements.

proposals/4501-rooms-as-social-media-pages.md
q-TheDog: I have included a client that might cover both. A social media client added onto Element Web.

- Client ([Haven Desktop](https://github.com/Haven-Organization/haven-desktop)):
  - https://github.com/Haven-Organization/haven-desktop/tree/develop/src/apps/social
  - https://github.com/Haven-Organization/haven-desktop/blob/develop/src/apps/social/utils/room-classifier.ts
- Bridge ([matrix-appservice-activitypub](https://github.com/Haven-Organization/matrix-appservice-activitypub)):
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/note_mirroring.py
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/inbox_dispatch.py
  - https://github.com/Haven-Organization/matrix-appservice-activitypub/blob/main/bridge/reaction_bridge.py

q-TheDog: Signed. Thanks!

q-TheDog:
Did a little restructuring to the document to give each distinct flavor of event being proposed its own header. Also changed the language around reposts to be consistent throughout instead of interchangeably using other social media platforms' terminology. Other terms for these actions are now isolated in "Also Known As" sections under each type. No changes to any of the JSON schemas

proposals/4501-rooms-as-social-media-pages.md
```diff
@@ -0,0 +1,683 @@
...
+    }
+  }
+}
+```
+
+This is what lets a profile have moderators (e.g. at power level 50, able to redact posts or ban
+abusive members) without those moderators being able to reassign or strip the profile's ownership out
+from under its actual owner.
+
+`m.social.profile_user_id` is the inverse of `m.social.profile_room_id` (see Profile/Group
```
ErikMichelson, line 105: Couldn't `m.social.profile_room_id` also be an array to support "owning" multiple feeds?

One use-case I could think of is having two different feeds for different topics where some people might prefer to only follow one of these.

ErikMichelson (commented):
Another question: How can this social media feed approach deal with per-post-visibility?

It is common for social media platforms to allow specifying whether a single post is publicly visible or only for followers. In the same way, replies could be public or only visible to the post author or even just visible to participants of the specific reply thread. 

I don't see any mechanism in matrix currently to handle per-message overrides for readability of a message, so I guess this wouldn't work and therefore should probably be noted as a limitation.

proposals/4501-rooms-as-social-media-pages.md
q-TheDog: > Couldn't m.social.profile_room_id also be an array to support "owning" multiple feeds?

Yes, and because the room state event points back to the user it's possible to have multiple profile rooms that indicate you as the owner under the current proposal (and I've run into that just by remaking them when testing my client). The only limitation is that yeah on your /profile data there's currently only one advertised as the true main profile room. I can think of a few UI examples where having a single main room on your /profile has been better than housing multiple:

1. On the user RightPanel, I can have a single big button for the one profile room. Trying to have a good UI to show that you have multiple here may not be as easy to convey:
image


2. It's a common convention on apps like Twitter, Mastodon, etc to have a "My Profile" button that takes you to it to see your posts, manage it, etc. Currently I have emulated this in Haven Desktop by making it so the Profile button uses the m.social.profile_room_id to decide which profile room to select:
image

3. When you going to the post composer on your Feed or when you click Post to start a new post, I've found it's best to have your "main" profile room be the default selection when choosing where to post to:
image


If you have multiple profiles you're not going to know consistently which room will be picked here.


Some ways to keep these features intact while still being able to fit multiple room IDs would be:
1. To also have a way to mark one of them as your main one
2. To make the order of them in the array matter and say the one at the first position should be treated as your main one.

If someone wants to throw out an idea for a JSON structure that achieves the above I'll try it out in a client.

q-TheDog: > How can this social media feed approach deal with per-post-visibility?

Currently we're limited by Matrix permissions here. Even with events of any kind today event visibility is always all-or nothing for that room. So when choosing to have a profile room you have to decide how you want ALL posts there to be accessed going forward. Not all social media platforms have post privacy settings though which doesn't make this too uncommon. You can't set privacy on a tweet for example.

Even worse of a flaw than just visibility, being limited by the Matrix permission model makes it so right now there's nothing stopping someone from posting new threads on your profile unless you also remove the ability to reply to your posts by blocking all messages. This can only be worked around with appservice moderation to redact people making new threads. There was a [discussion in element-meta](https://github.com/element-hq/element-meta/issues/209) about having a permission so only certain users can create top-level threads in rooms but I think work on it stagnated. My workaround for this was adding reposts/quote posts/cross-room replies since at least you can take a post elsewhere if someone does lock down posting on their profile.

q-TheDog:
Made the following changes:

- Added `m.social.body` and `m.social.formatted_body` as a place to store an alternative body for a m.room.message when the body you want to show in the social client is different than the body you want to show in non-MSC4501 clients. This came out of necessity when working on [matrix-appservice-activitypub](https://github.com/Haven-Organization/matrix-appservice-activitypub) where I was unable to include information like  "[user] reposted [other user]'s post: [blockquote]" while also being able to cleanly remove the header and blockquote from the body that was already being displayed via the `m.social.relates_to` block in the social media client. With a separate body field you can just store each version of them separately. Included a JSON to demonstrate this.

- Clarified usage of `content_inline` to make it clear that it applies to file attachments on the outer event too. Provided an example JSON for that.

- Moved the above two things into their own section: [Client Backwards Compatibility](https://github.com/Haven-Organization/matrix-spec-proposals/blob/msc-social-media-pages/proposals/4501-rooms-as-social-media-pages.md#client-backwards-compatibility), since they both only exist for the purpose of making events render nicely in non-MSC4501 clients. 

No changes to any of the already-proposed types.

q-TheDog:
Added a place for `via` entries for all cross-room references. This did mean changing one proposed key. `m.social.profile_room_id` is now `m.social.profile_room`, and is now a block containing `room_id` and `via`.

PR #4502, opened 2026-07-08
MSC4502: Targeted and unrestricted room member queries

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/as-unrestricted-membership/proposals/4502-as-unrestricted-membership.md)

Thanks @Half-Shot for participating in the brainstorming and design.

### DIFF
--- a/proposals/4502-as-unrestricted-membership.md
+++ b/proposals/4502-as-unrestricted-membership.md
@@ -5,7 +5,7 @@ Client-Server API. This can, among others, be used to verify a user's membership
 only works for local users, however. In some cases an application service might need to verify that
 a remote user is joined to a room known to the local homeserver.
 
-One example for this is the Matrix RTC Authentication Service from [MSC4195: MatrixRTC Transport
+One example for this is the LiveKit SFU Authorisation Service from [MSC4195: MatrixRTC Transport
 Using LiveKit Backend]. A remote user hits this service to get access to a local user's RTC media
 stream. A precondition for this is that both users share a room where a MatrixRTC session is
 established. In order to verify this, the service needs to check if the remote user is actually
--- a/proposals/4502-as-unrestricted-membership.md
+++ b/proposals/4502-as-unrestricted-membership.md
@@ -19,7 +19,7 @@ room memberships on any room known to the homeserver.
 A new optional property `scopes` is added at the top level of the service [registration file].
 `scopes` is an array of strings. The only permitted value is `unrestricted_member_access`.
 
-``` yaml
+```yaml
 id: "My Service"
 url: null # No traffic required
 as_token: "..."
--- a/proposals/4502-as-unrestricted-membership.md
+++ b/proposals/4502-as-unrestricted-membership.md
@@ -25,7 +25,7 @@ Exactly one of these parameters can be supplied at the same time. If both or non
 server MUST respond with HTTP 400 / `M_MISSING_PARAM`. If either `roomId`, `mxid` or `server_name`
 contain invalid values, the server MUST respond with HTTP 400 / `M_INVALID_PARAM`.
 
-If all paremeters are valid, the server checks whether the supplied user identifier or server name
+If all parameters are valid, the server checks whether the supplied user identifier or server name
 is joined to the room and returns the following response:
 
 ``` json5
--- a/proposals/4502-as-unrestricted-membership.md
+++ b/proposals/4502-as-unrestricted-membership.md
@@ -1,25 +1,50 @@
-# MSC4502: Unrestricted room member access for application services
+# MSC4502: Targeted and unrestricted room member queries
 
-Application services have the ability to [masquerade] as a local user when making requests to the
-Client-Server API. This can, among others, be used to verify a user's membership in a room. This
-only works for local users, however. In some cases an application service might need to verify that
-a remote user is joined to a room known to the local homeserver.
+Matrix allows users to query room members via either [/rooms/{roomId}/joined_members] or
+[/rooms/{roomId}/members]. Both endpoints require the user to be joined to the room themselves. In
+some cases though, users can have a legitimate reason to query room members for rooms that they are
+not joined to. This can, for instance, apply to admin users or to privileged application services
+such as moderation tooling.
 
-One example for this is the LiveKit SFU Authorisation Service from [MSC4195: MatrixRTC Transport
-Using LiveKit Backend]. A remote user hits this service to get access to a local user's RTC media
-stream. A precondition for this is that both users share a room where a MatrixRTC session is
-established. In order to verify this, the service needs to check if the remote user is actually
-joined to that room.
+Another problem of the existing endpoints is that they require loading all members. This is wasteful
+in cases where only a specific user's or server's membership in a room is required.
 
-This proposal extends the [Application Service API] with a new scope to permit services to look up
-room memberships on any room known to the homeserver.
+The present proposal addresses these issues by introducing a new endpoint for querying a single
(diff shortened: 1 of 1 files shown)


proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,69 @@
...
+
+This proposal extends the [Application Service API] with a new scope to permit services to look up
+room memberships on any room known to the homeserver.
+
+## Proposal
+
+A new optional property `scopes` is added at the top level of the service [registration file].
+`scopes` is an array of strings. The only permitted value is `unrestricted_member_access`.
+
+``` yaml
```
turt2live, line 47:
```suggestion
```yaml
```

proposals/4502-as-unrestricted-membership.md
turt2live, line 1: Implementation requirements:

* [ ] Server (respecting scopes)
* [ ] Appservice (using scopes)

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,69 @@
...
+id: "My Service"
+url: null # No traffic required
+as_token: "..."
+hs_token: "..."
+sender_localpart: "_my_service"
+namespaces: # No namespaces required
+  users: []
+  aliases: []
+  rooms: []
+scopes: [ "unrestricted_member_access" ]
```
Johennes: @turt2live remarked that the word "scopes" might be a little too general and cause tension with identically named things in the authentication / authorization world.

Half-Shot: Happy to go with something else, albeit I was rather hoping we could directly hook into OIDC scopes to reduce the implementation burdern on homeservers. Whereby ASes now have hardcoded scopes in their registration file, which neatly hooks into the permissions HS developers will already be building for clients.

There may also be some AS-only scopes like the one in this MSC.

Johennes: > [...] albeit I was rather hoping we could directly hook into OIDC scopes to reduce the implementation burdern on homeservers. 

Oh. So that would mean defining something like `urn:matrix:client:rooms:joined_members:unrestricted` in the CS API to give the calling user access to `/joined_members` on any room and reusing that scope value in the AS API?

Half-Shot: yeah, pretty much. Has the side benefit that the spec could in theory allow users with permission (server admins) to request the same perm?

Johennes: Yeah, that makes a lot of sense. And I think it's not actually a big change to the MSC. We'll basically just anchor the scope in another place where it can see further reuse.

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,69 @@
+# MSC4502: Unrestricted room member access for application services
+
+Application services have the ability to [masquerade] as a local user when making requests to the
+Client-Server API. This can, among others, be used to verify a user's membership in a room. This
+only works for local users, however. In some cases an application service might need to verify that
+a remote user is joined to a room known to the local homeserver.
+
+One example for this is the Matrix RTC Authentication Service from [MSC4195: MatrixRTC Transport
```
HarHarLinks:
```suggestion
One example for this is the LiveKit SFU Authorisation Service from [MSC4195: MatrixRTC Transport
```

...I maintain calling this software *also* MAS is a bad idea.

proposals/4502-as-unrestricted-membership.md
Johennes: Fixed in e302d6df125dfc6b11b4941a337463e291778729.

proposals/4502-as-unrestricted-membership.md
Johennes: Server: https://github.com/element-hq/synapse/pull/19974

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,111 @@
...
+The present proposal addresses these issues by introducing a new endpoint for querying a single
+subject's membership in a room and guarding the endpoint behind a dedicated OAuth scope token
+available to both users and application services.
+
+## Proposal
+
+A new Client-Server endpoint `GET /_matrix/client/v3/rooms/{roomId}/is_joined` is introduced. The
+endpoint accepts the following query parameters:
+
+- `mxid` (string): A [user identifier].
```
reivilibre, line 21: `mxid` vs `user_id`

Seems we actually (surprise to me) use both in the spec, I would have thought we used `user_id` in most places though.

I know it's minor but this sort of inconsistency, when stacked up with others, makes a spec feel a bit eclectic

Johennes: Good call. `user_id` also feels slightly clearer to me. Have made the switch in 02a05d76b7e9a48c6755a38ce3d939e5e6cebdf2.

Johennes: Oh actually, I needed to roll that back in a420cc3cdd747ccd081bf46030777b52db56436e. `user_id` clashes with the parameter that application services can use to emulate other users.

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,111 @@
...
+  "joined": true
+}
+```
+
+- `joined` (required, boolean): `true` if the subject is joined to the room. `false` otherwise
+  including when the server cannot verify membership because it doesn't know the subject or the
+  room.
+
+Access to the endpoint is guarded by a new OAuth scope token `urn:matrix:client:rooms:is_joined`.
+Usage of this scope SHOULD be reserved to admins and application services.
```
tonkku107, line 42: Why not allow regular users to use this endpoint for rooms they are joined in without a special scope and only require the scope when checking rooms the user isn't a part of?

The scope should also be clear about being an administrative scope. See [this comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4484#discussion_r3312487798) in MSC4484.

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,111 @@
...
+Another problem of the existing endpoints is that they require loading all members. This is wasteful
+in cases where only a specific user's or server's membership in a room is required.
+
+The present proposal addresses these issues by introducing a new endpoint for querying a single
+subject's membership in a room and guarding the endpoint behind a dedicated OAuth scope token
+available to both users and application services.
+
+## Proposal
+
+A new Client-Server endpoint `GET /_matrix/client/v3/rooms/{roomId}/is_joined` is introduced. The
```
tonkku107, line 18: If the intention is to only be an administrative endpoint then it should be under [/_matrix/client/v1/admin](https://spec.matrix.org/v1.19/client-server-api/#server-administration)

proposals/4502-as-unrestricted-membership.md
```diff
@@ -0,0 +1,111 @@
...
+id: "My Service"
+url: null # No traffic required
+as_token: "..."
+hs_token: "..."
+sender_localpart: "_my_service"
+namespaces: # No namespaces required
+  users: []
+  aliases: []
+  rooms: []
+scopes: [ "urn:matrix:client:rooms:is_joined" ]
```
tonkku107, line 57: If we are to introduce oauth scopes to application services, then they should also request `urn:matrix:client:api:*` for the rest of the client-server api, etc.
That complicates device masquerading though as device access is also a scope. Perhaps `urn:matrix:client:device:*` could be acceptable for an application service.

PR #4504, opened 2026-07-14
MSC4504: Using a global version number for the entire specification (v2)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/global-versioning-v2/proposals/4504-global-versioning-v2.md)

**Disclaimer**: I don't expect this is a complete or fully adequate solution to all of the versioning problems/concerns, but it should at least help set the protocol in a direction of addressing them.

### DIFF
--- a/proposals/4504-global-versioning-v2.md
+++ b/proposals/4504-global-versioning-v2.md
@@ -1,7 +1,7 @@
 # MSC4504: Using a global version number for the entire specification (v2)
 
 **Note**: This proposal changes process rather than specification. It may be tested in production
-ahead of formal acceptance under the supervison of the Spec Core Team.
+ahead of formal acceptance under the supervision of the Spec Core Team.
 
 [MSC2844](https://github.com/matrix-org/matrix-spec-proposals/pull/2844) ([merged](https://github.com/matrix-org/matrix-doc/blob/master/proposals/2844-global-versioning.md))
 introduced the [modern vX.Y versioning scheme](https://spec.matrix.org/v1.19/#specification-versions)
@@ -134,7 +134,7 @@ A third option would be to label certain releases as Long-Term Stable (LTS) rele
 could allow for non-LTS releases to potentially make more breaking changes, but might carry higher
 administrative burden to operate or determine the impact to server implementations. This proposal
 still introduces the essence of LTS releases to the spec, but does not call them LTS releases because
-the intermediary releases are requried to be compatible with the last "LTS" release.
+the intermediary releases are required to be compatible with the last "LTS" release.
 
 Finally, we could require implementations to drop old spec versions on a schedule. As the introduction
 mentions, this would potentially cause issues within local server ecosystems where they're not ready
@@ -176,7 +176,7 @@ error code to encounter in the wild (as is `500 M_UNKNOWN`).
 
 It's expected that the first removal-containing release after this MSC is merged to the spec will be
 at least 6 months later. This is to give time for implementations to discover the new requirements
-and adapt accordingly. That release MAY be out of cycle (ie: not divisble by 6), but MUST be the last
+and adapt accordingly. That release MAY be out of cycle (ie: not divisible by 6), but MUST be the last
 out of cycle release until the process is changed by another MSC. If that release is out of cycle,
 that fact SHOULD be advertised to developers during the minimum 6 month time period.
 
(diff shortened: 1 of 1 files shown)


proposals/4504-global-versioning-v2.md
turt2live, line 1: Implementation requirements:

* [ ] SCT comfortable with the process (see Unstable Prefix section)

proposals/4504-global-versioning-v2.md
```diff
@@ -0,0 +1,186 @@
...
+wouldn't continue encountering errors.
+
+Where removals or sufficiently breaking changes happen to events or rooms, clients will need to be
+differently aware of their compatibility. Proposals like [MSC4292](https://github.com/matrix-org/matrix-spec-proposals/pull/4292)
+attempt to address that as a distinct problem case.
+
+Removals from the specification MUST only happen when the `Y` (`vX.Y`) version is divisible by 6.
+Implementations can (and SHOULD) also remove functionality when they support such a spec version.
+This is done to reduce the number of specification versions that clients need to check compatibility
+for - they can assume that any spec release is compatible with the 5 prior at a minimum.
```
tulir, line 69: This feels a bit too close to semver without actually using semver. Should we instead just switch to semver and suggest that we SHOULD make a major bump with all the queued removals every 1.5 years?

We'd probably need to come up with a suitable definition of "breaking change" for the purposes of versioning. Wouldn't want new default room versions or things like authed media to force a major version bump even though some people might interpret them as breaking.

The main downside is that the major version would start increasing and marketing releases like Matrix 2.0 wouldn't really be possible anymore. I don't see any other problems with a semver-like scheme though.

turt2live: For added context: we've historically wanted to use semver, but didn't want to give up control over when the major version bump happens to the versioning scheme. This has resulted in us having a very semver-like scheme without being semver compatible. 

This MSC is a mix of more semver properties (still without being semver compatible) and LTS properties (also without being LTS compatible). 

I'm not confident in the solution, and adopting semver would reduce the awkward semver-like properties in the spec. It wouldn't necessarily fix being unable to remove things from implementations in practice, but maybe we can advocate that old spec versions be dropped more quickly than never as a separate problem.

proposals/4504-global-versioning-v2.md
```diff
@@ -0,0 +1,186 @@
...
+for the user because the client is unable to continue working. For example, receiving `404 M_UNRECOGNIZED`
+on [`/sync`](https://spec.matrix.org/v1.19/client-server-api/#get_matrixclientv3sync) would prevent
+most clients from operating at all, so they might need to get the user to a screen where the client
+wouldn't continue encountering errors.
+
+Where removals or sufficiently breaking changes happen to events or rooms, clients will need to be
+differently aware of their compatibility. Proposals like [MSC4292](https://github.com/matrix-org/matrix-spec-proposals/pull/4292)
+attempt to address that as a distinct problem case.
+
+Removals from the specification MUST only happen when the `Y` (`vX.Y`) version is divisible by 6.
```
richvdh, line 66: Whether you use semver or a weird `Y % 6 == 0` system, this sounds like it is going to introduce a Python 3 problem to me: the upgrade sounds scary (even if it's not), and implementations are going to be stuck on the old version. Let's face it, the Element client implementations lag badly as it is.

proposals/4504-global-versioning-v2.md
richvdh: This will be the point at which someone suggests we have a separate "marketing" version: in other words, Matrix 2.0 will actually correspond to spec version 3.15. I would not be in favour of that: we'll be forever trying to understand which one people are talking about.

I think @ara4n is the main advocate of "Matrix 2.0" being something we can shout about.

tulir: Definitely shouldn't have marketing versions that are confusable with real versions. Something like "Matrix 2026 edition" as the marketing version could work, though not sure how nice that is

clokep: > For added context: we've historically wanted to use semver, but didn't want to give up control over when the major version bump happens to the versioning scheme. This has resulted in us having a very semver-like scheme without being semver compatible.

Another solution is to add a number, e.g. vX.Y.Z:
* X is the "manual" control knob
* Y is breaking changes
* Z is non-breaking changes

🤷 It makes the version longer, but maybe helps w/ the problem. (I think we'd need to then wait for v2 to adopt it though.)

turt2live: that's also arguably semver with extra steps :p

clokep: Maybe, but having a magic `% 6` for breaking changes seems pretty confusing for everyone.

PR #4505, opened 2026-07-14
MSC4505: Push Rules for Live Location Sharing

ara4n:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/beacon-push-rules/proposals/4505-lls-push-rules.md)

Written with Matrix SCT lead hat on.

(implementations moved to thread: https://github.com/matrix-org/matrix-spec-proposals/pull/4505/changes#r3580285710 )

### DIFF
--- a/proposals/4505-lls-push-rules.md
+++ b/proposals/4505-lls-push-rules.md
@@ -86,7 +86,7 @@ de-duplicate notifications for a beacon they already know about.
 As with [MSC3930](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3930-polls-notifications.md),
 these rules are a stop-gap of the current push rules system rather than a step
 towards a general solution for extensible events, and future work on notifications for extensible
-events would supercede them.
+events would supersede them.
 
 ## Alternatives
 
--- a/proposals/4505-lls-push-rules.md
+++ b/proposals/4505-lls-push-rules.md
@@ -1,4 +1,4 @@
-# MSCXXXX: Push rules for live location sharing
+# MSC4505: Push rules for live location sharing
 
 [MSC3489](https://github.com/matrix-org/matrix-spec-proposals/pull/3489) and
 [MSC3672](https://github.com/matrix-org/matrix-spec-proposals/pull/3672) describe live location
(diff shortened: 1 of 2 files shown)


proposals/4505-lls-push-rules.md
turt2live, line 1: Implementation requirements:

* [ ] Client (using)
* [ ] Server (respecting)

UX for setting the push rules is not required. We have a number of push rules already which demonstrate the UX possibilities (including a flight console).

turt2live: From PR description:

 * Synapse impl: https://github.com/element-hq/synapse/pull/19954
 * EXI impl: https://github.com/element-hq/element-x-ios/pull/5872
 * EXA impl: https://github.com/element-hq/element-x-android/pull/7203
 * EW impl: https://github.com/element-hq/element-web/pull/34247

proposals/4505-lls-push-rules.md
```diff
@@ -0,0 +1,120 @@
...
+The `.m.rule.beacon` suppression cannot take effect in encrypted rooms, where location updates are
+sent as `m.room.encrypted` events and therefore match the `.m.rule.encrypted` underride on the
+server. Clients which re-evaluate push rules against the decrypted event (as the major
+implementations do) will suppress the notification via `.m.rule.beacon` at that point; the cost of
+the spurious push wakeups themselves is a pre-existing problem for all encrypted event types that
+are uninteresting once decrypted, and is out of scope here.
+
+Re-sending `m.beacon_info` with `live: true` (e.g. extending a share's timeout) will notify again.
+This is arguably correct ("X is (still) sharing their live location"); clients may wish to
+de-duplicate notifications for a beacon they already know about.
```
Johennes, line 84: Du-duplicating itself seems ok. The wording here doesn't sound fully convincing though. A notification for something that hasn't changed feels incorrect rather than correct to me. 🤔

Maybe we could just omit the correctness part, accept duplicates as a fact and RECOMMEND clients to handle them?

```suggestion
Re-sending `m.beacon_info` with `live: true` (e.g. extending a share's timeout) will notify again.
Clients SHOULD de-duplicate notifications for a beacon they already know about.
```

Susurrus: On the client side, I don't think we'd want to send another notification in this case. We do want to send the initial notification, but we won't put the duration in it. If the person sharing extends it, I don't think we'd notify at all. So this will end up being more code on the client side to handle this. Unless there's a clear use case for notifications in this scenario, I'd suggest we go the other way and eliminate this as the default.

PR #4507, opened 2026-07-14
MSC4507: Return additional room state from `/_matrix/federation/v1/hierarchy/{roomId}`

anoadragon453:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/anoa/hierarchy_additional_room_state/proposals/4507-hierarchy-extra-room-state.md)

Disclosure: This is done primarily with my [Element](https://element.io/) hat on.

### DIFF
--- a/proposals/4507-hierarchy-extra-room-state.md
+++ b/proposals/4507-hierarchy-extra-room-state.md
@@ -1,0 +1,197 @@
+# MSC4507: Return additional room state from `/_matrix/federation/v1/hierarchy/{roomId}`
+
+The [`GET
+/_matrix/federation/v1/hierarchy/{roomId}`](https://spec.matrix.org/v1.19/server-server-api/#get_matrixfederationv1hierarchyroomid)
+federation endpoint allows homeservers to ask another homeserver for certain
+information about a room or a space. This is most useful when a homeserver does
+not (currently) have any users in the room, and thus does not have the latest
+room state synchronised.
+
+The information that is returned is maintained as part of the Matrix spec, and
+includes fields such as `join_rule`, `room_type`, `topic`. However, there may be
+times when it's useful for a homeserver to request arbitrary state events from a
+room without joining it.
+
+This proposal adds a new query parameter and response field to [`GET
+/_matrix/federation/v1/hierarchy/{roomId}`](https://spec.matrix.org/v1.19/server-server-api/#get_matrixfederationv1hierarchyroomid)
+which allows homeservers to ask for certain pieces of state from a room, or any
+space child rooms, that they may not be in.
+
+It also defines a new state event, which determines which state event types and
+state keys a homeserver that isn't in the room is allowed to see.
+
+## Use case
+
+The primary use case that motivated this proposal is the concept of "security
+policies" from Element's implementation of [Mission Partner
+Environments](https://element.io/en/mission-partner-environments). Broadly, one
+defines a "security policy" for a room which, among other things, states the
+security clearance needed to access the room. This security policy is
+implemented as a state event, and clearance information is stored in local
+OIDC identity providers or LDAP instances as user attributes.
+
+When a user tries to join a room, their own homeserver must approve/deny the join
+based on the user's security clearance and the security policy in the room. But
+the homeserver may not have access to the room's security policy if this is the
+first user to join the room. Therefore, homeservers need a way to request this
+information from resident servers before carrying out the room join.
+
+This unlocks other use cases as well, of course. No longer does a bespoke field
+for every type of state event need to be defined in the response body of this
+request. In addition, homeservers can choose to "opt out" of information they
+don't need by not specifying the state event. The current shape of the endpoint
+has no method of doing that.
+
+## Proposal
+
+### Query parameter
+
+Add a `additional_state` query parameter to [`GET
+/_matrix/federation/v1/hierarchy/{roomId}`](https://spec.matrix.org/v1.19/server-server-api/#get_matrixfederationv1hierarchyroomid).
+The value is a JSON object in form:
+
+```json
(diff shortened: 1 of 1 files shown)


proposals/4507-hierarchy-extra-room-state.md
turt2live, line 1: Implementation requirements:

* [ ] Server (serving)
* [ ] Server (requesting)
* [ ] Client (rendering/setting state)

proposals/4507-hierarchy-extra-room-state.md
```diff
@@ -0,0 +1,197 @@
...
+
+If a state event is marked as public via `m.public_state`, then marked private
+again by another `m.public_state`, it is possible for the event to still be
+exposed if a homeserver in the room has not yet received the second
+`m.public_state` state event. Likewise, a state event may not be made public yet
+if the first state event has not yet reached a given resident homeserver.
+
+This edge case is considered acceptable for the above use case.
+
+## Alternatives
```
turt2live, line 171: In the T&S world we're also looking at something called Join Gates: arbitrary server-defined policies which can prevent another user from joining or being invited to a room. Some rooms implement a basic form of join gates through either `restricted` join rules or a bot which kicks & DM's the joiner. Obviously neither of these is suitable for the intended use case of this MSC, but proper Join Gates would enable the use case nicely I think.

For T&S, the imagined scenario is a Join Gate can send a user through a web browser (if needed) to complete a captcha, acknowledge community guidelines, etc. The same mechanism can be used to allow/deny users joining a room based on clearance and classification, and without using a web browser, all before an event is sent to the DAG (so a denied user never joins the room).

A benefit that this MSC brings over Join Gates is a definition of per-event history visibility, though limited to state events. We do want per-event history visibility, but limiting it to certain state events feels like a bandaid.

turt2live (commented):
Early review with a combination T&S and SCT hats. Apologies that I don't have much to link to for what Join Gates are, but hopefully the comment covers it.

proposals/4507-hierarchy-extra-room-state.md
anoadragon453: All good context, thank you!

For the timescales of the implementation work involved here, we can't afford to wait for join Gates to flourish. But I think we can point the implementation to eventually make use of it once the concept is more concrete.

At that point, happy to pivot this MSC towards (in close it in favour of) something that's useful for more granular concept of event visibility.

turt2live: Another options which occurred to me: could just use policy servers as they are built today :) Aside from [a bug in Synapse](https://github.com/element-hq/synapse/pull/19593), policy servers can prevent joins in rooms too. They'd also naturally have access to the room state.

anoadragon453: Fair! Currently the implementation is relying on a synapse module, with a `user_may_join_room` implementation.

The main problem addressed by this MSC is that for a federated join, each homeserver (including a policy server) only has access to either:

* The security policy as defined in the room state.
* Details about the user (stored in a local IdP) which grant or deny them permission to join the room.

This MSC allows a homeserver with access to a user's details the ability to fetch the room's security policy, before allowing/denying the join locally.

If a policy server (on the resident server's side?) were to take on approving/denying the request, then it'd somehow need to fetch the user's attributes from the remote homeserver. And that'd require an entirely different MSC.

turt2live: The policy server is a third party rather than on either particular server - it could be given access to the IdP, and it would still be a resident.

proposals/4507-hierarchy-extra-room-state.md
```diff
@@ -0,0 +1,202 @@
...
+
+This unlocks other use cases as well, of course. No longer does a bespoke field
+for every type of state event need to be defined in the response body of this
+request. In addition, homeservers can choose to "opt out" of information they
+don't need by not specifying the state event. The current shape of the endpoint
+has no method of doing that.
+
+## Proposal
+
+### Query parameter
```
clokep, line 47: This kind of feels like it should be a body field, maybe we should revisit the endpoint and make it a POST?

Repeating a field and JSON encoding into query parameters is a lot of parsing.

anoadragon453: Agreed, as I've found while writing up an initial implementation.

proposals/4507-hierarchy-extra-room-state.md
```diff
@@ -0,0 +1,202 @@
...
+A state event is preferred to individual homeserver configuration, as that can
+lead to different responses from different homeservers. As well as lack of easy
+per-room configuration, or control by room admins (rather than homeserver
+operators).
+
+### Response
+
+The remote homeserver SHOULD return state events from the room that match the
+requested state events and are authorised to be shared by the `m.public_state`
+state event, as known to the resident homeserver.
```
anoadragon453, line 98: Should these be stripped state events, rather than full PDUs? My use case only needs the `content` of the event, rather than anything else.

And full PDUs would reveal more information about the room (event IDs in `prev_events`, `auth_events`), which the requester doesn't really need.

proposals/4507-hierarchy-extra-room-state.md
```diff
@@ -0,0 +1,203 @@
...
+response fields, and only relies on `additional_state` semantics. This allows
+homeservers to only ask for the fields they need, eliminating extra bandwidth
+used for other fields. This is probably best paired with an updated
+Client-Server `/hierarchy` request, so clients can specify which fields they're
+interested in.
+
+It'd also be best to make the request a POST, so that we don't have to abuse
+query parameters for the `(event type, state_key)` definitions.
+
+## Security considerations
```
anoadragon453, line 185: This MSC should respect room ACLs.

PR #4508, opened 2026-07-15
MSC4508: Sliding Sync Extensions: Common format and Typing

erikjohnston:
MSC4186 (Simplified Sliding Sync) only includes core room data and omits other data, such as typing notifications, from the sync response. Instead, such data is left to "extensions", which clients opt into individually via the `extensions` field of the sync request.

This MSC defines the extension for typing notifications, so that clients using sliding sync can show typing indicators in rooms.

Supersedes MSC3961.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sliding-sync-typing-extension/proposals/4508-sliding-sync-extension-typing.md)

---

Note: LLM was used in the very *initial* barebones draft and then subsequently to check grammar and consistency. The MSC has been heavily edited and reviewed since then.

---

SCT Stuff:

[MSC Checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4508#issuecomment-5118243075)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4508#issuecomment-5118963462)


proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,232 @@
...
+
+- it is currently within the range of one of the lists named in `lists` (i.e. after applying the
+  list's filters, the room's index in the filtered list is within the list's range), or
+- it is one of the room subscriptions named in `rooms`.
+
+The special value `"*"` acts as a wildcard: `{"lists": ["*"]}` matches all lists in the request and
+`{"rooms": ["*"]}` matches all room subscriptions in the request. An empty array matches nothing,
+e.g. `{"lists": [], "rooms": ["*"]}` applies the extension only to room subscriptions. List keys
+that are not in the request, and rooms that are either unknown or inaccessible to the user, are
+ignored.
```
erikjohnston: This is the situation today, but in the SSS MSC we removed these sorts of wildcards. I'm very tempted to remove them here as well and rely on the whole "missing key means all" behavious.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,232 @@
...
+| Name | Type | Required | Comment |
+| - | - | - | - |
+| `enabled` | `bool` | No | Whether the extension is enabled. Defaults to `false`. The server only processes the extension if this is `true`. |
+
+Extension MSCs MUST address the behaviour when an extension that was absent or disabled is later
+enabled on an existing connection. Extensions SHOULD leave the behaviour up to server
+implementations unless there is a client UX need, as clients SHOULD enable the extensions they need
+at the beginning of the connection. This gives the server flexibility to track extension data
+efficiently, regardless of whether extensions are toggled. Servers MUST allow toggling extensions,
+even if they only send newly changed data, and MUST NOT reject the request.
```
erikjohnston, line 43: This is mostly from the fact that I don't think current extensions really deal well with toggling of extensions. However, clients *deferring* enabling extensions may be a valid use case. For example, when doing the initial sync they may first ignore receipts etc to speed up that initial response, and then subsequently enable it. That is likely easier for servers to implement.

proposals/4508-sliding-sync-extension-typing.md
anoadragon453, line 1: Heads up this file is still prefixed with `xxxx-`.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,232 @@
...
+
+A new sliding sync extension is added, with the extension key `typing`.
+
+## Common extension semantics
+
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) defines the `extensions`
+request field as a map from extension key to an extension config, but leaves the format of the
+config to the MSCs defining each extension. As this is the first such MSC, it defines the following
+semantics, which are expected to be common to all extensions. Future extension MSCs may reference
+this definition rather than redefine them.
```
anoadragon453, line 23: Thanks for defining these semantics! I'll follow them in MSC4262 and clarify what to do in each edge case.

> the MSCs defining each extension. As this is the first such MSC, ...

Pretty sure this isn't the first SS extension MSC :P

proposals/4508-sliding-sync-extension-typing.md
turt2live: (converting to draft while true)

turt2live, line 1: Implementation requirements:

* [x] Client (using)
* [x] Server

proposals/4508-sliding-sync-extension-typing.md
lveneris, line 23: Why are these be defined here, particularly as there are several existing Sliding Sync Extension proposals? In the interests of keeping an atomic goal, I feel that this would be best suited to a separate proposal, which this proposal would then rely on as a dependency.

erikjohnston: Maybe, I think it's easier for people to see a real (albeit simple) extension defined alongside so they can see how it works in practice. Otherwise it becomes all a bit abstract.

lveneris: Yes, I presume a proposal that defines the shape of a sliding sync extension would feature an example of its usage. I would not, however, expect a proposal that offers a sliding sync extension to sneak in a general structure for all other extensions.

erikjohnston: It was decided that we should rename this PR to cover the fact it defines both the common format and typing

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: Done in 060f9e4

erikjohnston:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

erikjohnston:
@mscbot fcp merge

proposals/4508-sliding-sync-extension-typing.md
turt2live: @mscbot concern Implementations not cited for review

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+
+Extension MSCs MUST specify the expected behaviour of clients when the sliding sync connection is
+reset (via `M_UNKNOWN_POS`).
+
+Extension MSCs MUST specify when the sync response should be returned immediately versus continuing
+with long-polling. Some extensions MAY not require certain updates to be returned immediately.
+
+As with unknown fields elsewhere in the client-server API, servers MUST ignore extensions with keys
+they do not recognise (rather than rejecting the request), so that newer clients remain compatible
+with older servers.
```
MadLittleMods, line 53: Original discussion: https://github.com/matrix-org/matrix-spec-proposals/pull/4186#discussion_r2207912899

I think it would also be useful to figure out a way for clients to know which extensions the homeserver supports.

For example, when I make a request with the `typing` extension enabled, and the server doesn't respond with anything, that could mean there is no (new) data or that the homeserver doesn't understand the extension (no way to distinguish).

I'm thinking more about something in the Sliding Sync response, and not some separate endpoint like checking capabilities.

turt2live: The homeserver would support whatever extensions are included in its supported spec version, so clients can use `/versions` to determine which extensions are supported. 

Unstable extensions would also be discovered through `/versions`.

MadLittleMods: You can have a stable endpoint without the Matrix spec version being advertised on your homeserver. For example with Synapse, we only advertise a spec version is supported once we've gone through stabilizing things piecemeal over time.

tulir: MSCs should always define both unstable and stable `/versions` flags, so that servers are able to advertise support for the stable form before supporting the entire spec release. The convention is to append `.stable` to the flag.

MadLittleMods: Sounds workable.

And the the action here is to document that each Sliding Sync extension should define unstable/stable `/versions` flags.

MadLittleMods: As a note: the `/versions` solution doesn't help if you accidentally typo an extension.

turt2live: > You can have a stable endpoint without the Matrix spec version being advertised on your homeserver. 

I consider this an implementation bug, personally. We have process to deal with it, as tulir mentions, but I'd honestly like to stop using it.

> For example with Synapse, we only advertise a spec version is supported once we've gone through stabilizing things piecemeal over time.

This is expected, but being 2 years out of date from the spec (in the case of Synapse) is unreasonable.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) (Simplified Sliding Sync)
+only includes core room data and omits other data, such as typing notifications, from the sync
+response. Instead, such data is left to "extensions", which clients opt into individually via the
+`extensions` field of the sync request.
+
+This MSC defines the extension for typing notifications, so that clients using sliding sync can show
+typing indicators in rooms.
+
+Supersedes [MSC3961](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/ssext-typing/proposals/3961-sliding-sync-typing.md).
```
turt2live, line 11:
```suggestion
Supersedes [MSC3961](https://github.com/matrix-org/matrix-spec-proposals/pull/3961).
```

linking to the PR is preferred when the linked MSC hasn't been accepted yet for comment history, etc.

proposals/4508-sliding-sync-extension-typing.md
turt2live: > > the MSCs defining each extension. As this is the first such MSC, ...
> 
> Pretty sure this isn't the first SS extension MSC :P

It is from a process perspective: the extensions system hasn't landed yet.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+            "rooms": ["!abcd:example.com"]
+        }
+    }
+}
+```
+
+## Extension response
+
+If the extension is enabled, the server MAY include a `typing` section in the `extensions` response
+field, with the following format:
```
turt2live, line 122: Why is this a MAY? Surely if the extension is enabled and the server recognizes it, there should be support?

If this is trying to save bandwidth when there's no typing updates, can we say "MUST include ... unless empty"?

turt2live (approved):
overall lgtm - I don't have overly blocking concerns

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+
+| Name | Type | Required | Comment |
+| - | - | - | - |
+| `rooms` | `{string: Event}` | No | A map of room ID to the `m.typing` ephemeral event for that room. |
+
+The value for each room is the same `m.typing` event that would appear in the room's
+`ephemeral.events` array in a [`/v3/sync`
+response](https://spec.matrix.org/v1.17/client-server-api/#typing-notifications), e.g.:
```
clokep, line 130: Does the shape of this EDU make sense to reuse here? It feels pretty duplicative to include a type of `m.typing` and the `content` block.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) defines the `extensions`
+request field as a map from extension key to an extension config, but leaves the format of the
+config to the MSCs defining each extension. As this is the first such MSC, it defines the following
+semantics, which are expected to be common to all extensions. Future extension MSCs may reference
+this definition rather than redefine them.
+
+> [!NOTE]
+>
+> There are a myriad of different use cases for extensions, e.g. typing notifications, to-device
```
dbkr, line 27: grammar nit

```suggestion
> There are myriad different use cases for extensions, e.g. typing notifications, to-device
```

dbkr (commented):
Otherwise nothing that hasn't already been mentioned.

proposals/4508-sliding-sync-extension-typing.md
richvdh: ("a myriad of" is valid: https://dictionary.cambridge.org/dictionary/english/myriad)

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+> messages, etc, all of which have different semantics. For typing notifications the client only
+> needs the latest state (rather than all updates), while for to-device messages the client must
+> (eventually) receive every missed to-device message in order.
+
+All extensions include the following fields:
```
richvdh, line 32: I wondered if this was talking about request, response, or something else.

```suggestion
All extensions include the following fields as members of their `ExtensionConfig`:
```

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+As with unknown fields elsewhere in the client-server API, servers MUST ignore extensions with keys
+they do not recognise (rather than rejecting the request), so that newer clients remain compatible
+with older servers.
+
+## Common per-room extension semantics
+
+Some extensions apply on a per-room basis, of which this is the first. We define the generic
+per-room semantics here.
+
+The per-room extensions include the following fields (as well as the ones above):
```
richvdh, line 60:
```suggestion
The per-room extensions include the following fields as members of their `ExtensionConfig` (as well as the ones above):
```

proposals/4508-sliding-sync-extension-typing.md
richvdh: > Does the shape of this EDU make sense to reuse here?

An aside, but note that this is *not* the shape of an [`m.typing` EDU](https://spec.matrix.org/v1.19/server-server-api/#typing-notifications).  I think we call this sort of `m.typing` an "ephemeral event". There is very much not a 1:1 mapping between EDUs and ephemeral events.

I could go either way on the actual point here. On the one hand, yes this format is duplicative and there doesn't seem to be any good reason to cargo-cult it from sync v3. On the other hand... there is that existing definition of an `m.typing` "event" and how clients should behave when they receive one.

On balance, I'd prefer to see a simplification here.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) keeps rooms the user has
+been kicked or banned from in lists. Such rooms MUST NOT have typing notifications sent.
+
+For the rooms that remain:
+
+- Whenever a room enters scope on a connection, the server MUST send the room's current typing
+  state. This applies on an initial sync (no `pos`), when the room first comes into scope on the
+  connection, and when it re-enters scope after having dropped out (e.g. after falling out of a
+  list's range). As a minor optimisation, the server MAY omit the room if there are no users typing
+  and the previously sent state (if any) was also empty. If the extension is enabled partway through
```
richvdh, line 167: > and the previously sent state (if any) was also empty

Why is this condition required, given we tell clients that they must clear the typing notifications locally whenever a room is out of scope?

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,235 @@
...
+
+# Unstable prefix
+
+The experimental implementations of
+[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) (e.g. Synapse, as used by
+Element X) have supported this extension with the unprefixed `typing` key on the
+`/_matrix/client/unstable/org.matrix.simplified_msc3575/sync` endpoint.
+
+Until this MSC is accepted, implementations MUST use `org.matrix.msc4508.typing` (substituting this
+proposal's MSC number) as the extension key on the stable
```
richvdh, line 224: > (substituting this proposal's MSC number)

given this proposal's number is already in the identifier... what does this mean?

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: Both are in released Synapse and EX code.

Synapse: https://github.com/element-hq/synapse/blob/556154848c54ccca9c1cff5bfa4a0baad8d752aa/synapse/handlers/sliding_sync/extensions.py#L864 / https://github.com/element-hq/synapse/pull/17505
Not quite sure where in the Rust SDK it is, but there are lots of references in: https://github.com/matrix-org/matrix-rust-sdk/tree/main/crates/matrix-sdk/src/sliding_sync

turt2live: @mscbot resolve Implementations not cited for review

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: Done this in ddd828e7b3f0b7136dd7fce18b734055db0f64ab

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: 46eeef13aee3bbd3d25d1b4ce52bf547857965f1

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: 9d905b722b5e0629b99ebe6d9b12549b9c8eff11

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: Yeah, we don't need it 76c3387f9e0727fdbca8dd85c52a4edc879ea2e9

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: Oops, that was a leftover from when I was first writing it. 2b9f28e4a7a5f624768a2a124bd088b8006dd548

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: I'm wondering if we should say that when a request enables an extension for the first time the server MUST include it in the response? That way clients can differentiate between an unsupported extension vs one that has no data to respond with?

turt2live: We already have `/versions` detection though - why can't we use that?

erikjohnston: Yeah, we can. Though having the server also respond means that clients don't have to wait for a new spec version (assuming we don't do `.stable`).

It also makes unstable-to-stable transition quite easy: simply send both unstable and stable extensions and see what is returned. That can then be stripped once the new spec version is announced.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -29,18 +29,28 @@ this definition rather than redefine them.
...
-implementations unless there is a client UX need, as clients SHOULD enable the extensions they need
-at the beginning of the connection. This gives the server flexibility to track extension data
-efficiently, regardless of whether extensions are toggled. Servers MUST allow toggling extensions,
-even if they only send newly changed data, and MUST NOT reject the request.
+Extension MSCs MUST address the behaviour when an extension is enabled partway through an existing
+connection. There are two distinct cases:
+
+- The extension is enabled for the first time on the connection, having been absent or disabled on
+  every previous request. Clients may choose to defer an extension deliberately to allow the initial
+  request to complete faster.
```
richvdh: ... or maybe if a new release of a client adds support for a new extension?

proposals/4508-sliding-sync-extension-typing.md
richvdh: I am very dubious about this MUST; I don't really see that SSS extension MSCs need a different set of rules to other MSCs.

(My understanding of the normal rule is: if you have a usecase for unstable or stable feature flags, then feel free to list them in the MSC, but neither is required.)

This is not a hill I'm going to die on though.

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,284 @@
...
+
+Extension MSCs MUST specify when the sync response should be returned immediately versus continuing
+with long-polling. Some extensions MAY not require certain updates to be returned immediately.
+
+As with unknown fields elsewhere in the client-server API, servers MUST ignore extensions with keys
+they do not recognise (rather than rejecting the request), so that newer clients remain compatible
+with older servers.
+
+Since an ignored extension is indistinguishable from an enabled extension that has no data to send,
+extension MSCs MUST define both an unstable and a stable feature flag in
```
richvdh: I am very dubious about this MUST; I don't really see that SSS extension MSCs need a different set of rules to other MSCs.

(My understanding of the normal rule is: if you have a usecase for unstable or stable feature flags, then feel free to list them in the MSC, but neither is required.)

This is not a hill I'm going to die on though.

richvdh (approved):

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,284 @@
...
+                "!abcd:example.com": {
+                    "user_ids": ["@alice:example.com", "@bob:example.com"]
+                }
+            }
+        }
+    }
+}
+```
+
+As in [`/v3/sync`](https://spec.matrix.org/v1.17/client-server-api/#typing-notifications), the
```
ara4n, line 176: I'm surprised that we don't say anything about how to actually figure out whether someone is typing or not - e.g. idle for N seconds = stopped typing; whether we coalesce to stop people flickering on and off or not; etc.  I assume this is already in the spec somewhere, in which case can we be explicit that the calculation should not be changed?

ara4n:
this LGTM, albeit with one minor proviso that we should be explicit about whether the way to determine if a user is typing has changed or not, and backref it to the spec if needed: https://github.com/matrix-org/matrix-spec-proposals/pull/4508/changes#r3916956193.

It'd also be good to actually say something, anything, about why 3961 got closed.

I'm also not a fan of mashing together "how to do extensions" with "here's a typing extension", but not going to die on that hill.

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: I actually don't think there is any client timeouts on typing at all currently. I'm somewhat keen to not make it a problem of this MSC as it's an existing issue with typing. I'd be OK if there was an obvious correct way of doing it, but I don't think its clear that clients should just do local timeouts (and if so what the value should be).

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: 990163ead

proposals/4508-sliding-sync-extension-typing.md
erikjohnston: aa2e599f1

proposals/4508-sliding-sync-extension-typing.md
```diff
@@ -0,0 +1,284 @@
...
+
+Extension MSCs MUST specify when the sync response should be returned immediately versus continuing
+with long-polling. Some extensions MAY not require certain updates to be returned immediately.
+
+As with unknown fields elsewhere in the client-server API, servers MUST ignore extensions with keys
+they do not recognise (rather than rejecting the request), so that newer clients remain compatible
+with older servers.
+
+Since an ignored extension is indistinguishable from an enabled extension that has no data to send,
+extension MSCs SHOULD define both an unstable and a stable feature flag in
```
erikjohnston, line 66: Note: this is now a SHOULD.

PR #4195, opened 2024-09-16
MSC4195: LiveKit transport for MatrixRTC

hughns:
[Rendered](https://github.com/hughns/matrix-spec-proposals/blob/hughns/matrixrtc-livekit/proposals/4195-matrixrtc-livekit.md)

----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4195#issuecomment-5577346882)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4195#issuecomment-5577344421)

----

> [!NOTE]
> All implementations use `livekit` rather than `m.livekit` for the transport's type.

### Server

- `m.livekit` in `/transports`: https://github.com/element-hq/synapse/pull/20146

> [!NOTE]
> For backwards compatibility, the endpoint also returns `livekit_service_url`. The stems from a previous iteration of the proposal and has been fully replaced by `url` in the current text.

- Client-Server & Server-Server endpoints: https://github.com/element-hq/lk-jwt-service implements the endpoints introduced by this proposal. People interested in reviewing the implementation should start in https://github.com/element-hq/lk-jwt-service/blob/main/src/handler.rs#L1245. The service is integrated into https://github.com/element-hq/synapse/ as an application service by means of:
  - https://github.com/element-hq/synapse/pull/19972
  - https://github.com/element-hq/synapse/pull/19977
  - https://github.com/element-hq/synapse/pull/19974.
- The three pull requests above depend on [MSC4512](https://github.com/matrix-org/matrix-spec-proposals/pull/4512) and [MSC4502](https://github.com/matrix-org/matrix-spec-proposals/pull/4502). This is an implementation detail of the Synapse implementation, however. There is no dependency between the current proposal and these two MSCs.

> [!NOTE]
> For backwards compatibility with implementations of previous versions of this MSC, https://github.com/element-hq/lk-jwt-service exhibits the following differences with respect to the proposal text:
>
> - The legacy endpoints `/sfu/get`, `/get_token` and `/delegate_delayed_leave` are still included. These have been replaced by proper Client-Server and Server-Server endpoints in the proposal which are _also_ implemented in https://github.com/element-hq/lk-jwt-service.
> - The device ID is still included in the LiveKit participant identity derivation. On requests, the ID is supplied as `claimed_device_id` wrapped together with the member ID in a `member` object. The proposal has eliminated `claimed_device_id` and replaced the `member` object with `member_id`.
[...]

PR #4512, opened 2026-07-20
MSC4512: Delegating parts of the Client-Server and Server-Server API to application services

Johennes:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/johannes/app-service-takeover/proposals/4512-app-service-takeover.md)

Thanks to @erikjohnston for ideating and brainstorming on this.

### DIFF
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -35,9 +35,9 @@ application service and allows the service to trigger S-S requests under its own
 
 ### Proxying client-server requests
 
-For any authenticated C-S request under `/_matrix/client/{proxy}/*`, the server first authorises the
-request as usual. If authorization succeeds, the server proxies the request to
-`{url}/_matrix/client/{proxy}/*` and streams the response back to the requesting client.
+For any authenticated C-S request under `/_matrix/client/(unstable/[^/]+|v[^/]+)/{proxy}/.*`, the
+server first authorises the request as usual. If authorization succeeds, the server proxies the
+request to the same path anchored on `url` and streams the response back to the requesting client.
 
 Any "hop-by-hop" headers as defined by [RFC2616] MUST be stripped both before forwarding the request
 to the service and before streaming the response back to the requesting client.
@@ -50,9 +50,9 @@ server supplies the MXID of the requesting client to the application service in 
 
 The process for proxying server-server requests is analogous.
 
-For any authenticated S-S request under `/_matrix/federation/{proxy}/*`, the server first authorises
-the request as usual. If authorization succeeds, the server proxies the request to
-`{url}/_matrix/federation/{proxy}/*` and streams the response back to the requesting server.
+For any authenticated S-S request under `/_matrix/federation/(unstable/[^/]+|v[^/]+)/{proxy}/.*`, the
+server first authorises the request as usual. If authorization succeeds, the server proxies the
+request to the same path anchored on `url` and streams the response back to the requesting server.
 
 Any "hop-by-hop" headers as defined by [RFC2616] MUST be stripped both before forwarding the request
 to the service and before streaming the response back to the requesting server.
@@ -73,7 +73,7 @@ POST /_matrix/client/v1/appservice/fed_proxy HTTP/1.1
 {
   "destination": "example.org",
   "method": "POST",
-  "path": "/_matrix/federation/{proxy}/foo/bar/",
+  "path": "/_matrix/federation/v1/{proxy}/foo/bar/",
   "query": { ... },
   "body": { ... }
 }
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -35,6 +35,9 @@ proxy_url: "http://127.0.0.1:1234"
 When both `proxy_prefix` and `proxy_url` are set, the server proxies matching C-S and S-S requests
 to the application service and allows the service to trigger S-S requests under its own prefix.
 
+For now, authentication is REQUIRED for any C-S and S-S requests under the `proxy_prefix`. A future
+proposal may introduce a way for application services to define when authentication isn't required.
+
 ### Proxying client-server requests
 
 For any authenticated C-S request under `/_matrix/client/(unstable/[^/]+|v[^/]+)/{proxy_prefix}/.*`,
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -42,7 +42,7 @@ the server first authorises the request as usual. If authorization succeeds, the
 request to the same path anchored on `proxy_url` and streams the response back to the requesting
 client.
 
-Any "hop-by-hop" headers as defined by [RFC2616] MUST be stripped both before forwarding the request
+Any "hop-by-hop" headers as defined by [RFC 2616] MUST be stripped both before forwarding the request
 to the service and before streaming the response back to the requesting client.
 
 Additionally, the `Authorization` header MUST be stripped on the forwarded request. Instead, the
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -46,7 +46,7 @@ Any "hop-by-hop" headers as defined by [RFC 2616] MUST be stripped both before f
 to the service and before streaming the response back to the requesting client.
 
 Additionally, the `Authorization` header MUST be stripped on the forwarded request. Instead, the
-server supplies the MXID of the requesting client to the application service in a new request header
+server supplies the User ID of the requesting client to the application service in a new request header
 `X-Matrix-User-Identifier`.
 
 ### Proxying server-server requests
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -21,7 +21,7 @@ file] for application services. `proxy_prefix` is a string and defines the path 
 service is claiming. `proxy_url` is a string that determines the `url` to send proxied requests to.
 `proxy_prefix` and `proxy_url` are optional but if used, MUST always be specified together.
 
-``` yaml
+```yaml
 id: "My app service"
 url: null
 as_token: "30c05ae90a248a4188e620216fa72e349803310ec83e2a77b34fe90be6081f46"
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -139,9 +139,11 @@ design as no associated use cases are known yet.
 
 ## Alternatives
 
-The existing `url` property could be reused to avoid having to specify a separate `proxy_url`. This
-would disallow using proxying while disabling namespace-related traffic from the server by setting
-`url` to `null`, however.
+The existing `url` property could be reused to avoid having to specify a separate `proxy_url`. However,
+`url` is also used to forward events to the application service as per its `namespaces` setting. An
+application service may not be interested in receiving regular Matrix events though. Instead, it could
+only be using `namespaces` to be able to puppet users on the C-S API. The separate `proxy_url` property
+resolves this conflict and avoids excess traffic between the homeserver and the application service.
 
 ## Security considerations
 
--- a/proposals/4512-app-service-takeover.md
+++ b/proposals/4512-app-service-takeover.md
@@ -167,5 +167,5 @@ None.
   [lk-jwt-service]: https://github.com/element-hq/lk-jwt-service
   [MSC4195]: https://github.com/matrix-org/matrix-spec-proposals/pull/4195
   [registration file]: https://spec.matrix.org/v1.19/application-service-api/#registration
-  [RFC2616]: https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
+  [RFC 2616]: https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
   [server name]: https://spec.matrix.org/v1.19/appendices/#server-name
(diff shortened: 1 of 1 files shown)


proposals/4512-app-service-takeover.md
Johennes, line 1: Implementation requirements:

- [ ] Server
- [ ] Application service

Johennes: - Server (proxying incoming requests): https://github.com/element-hq/synapse/pull/19972
- Server (sending outgoing requests): https://github.com/element-hq/synapse/pull/19977

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Unstable prefix
+
+| Stable identifier | Purpose | Unstable identifier |
+|----|----|----|
+| `proxy` | Registration file property | `io.element.msc4512.proxy` |
+| `/_matrix/client/v1/appservice/fed_proxy` | Endpoint | `/_matrix/client/unstable/io.element.msc4512/appservice/fed_proxy` |
+| `M_FEDPROXY_DESTINATION_DENIED` | Error code | `IO.ELEMENT.MSC4512.M_FEDPROXY_DESTINATION_DENIED` |
+| `M_FEDPROXY_PATH_NOT_ALLOWED` | Error code | `IO.ELEMENT.MSC4512.M_FEDPROXY_PATH_NOT_ALLOWED` |
+| `M_FEDPROXY_CONNECTION_FAILED` | Error code | `IO.ELEMENT.MSC4512.M_FEDPROXY_CONNECTION_FAILED` |
```
tulir, line 167:
```suggestion
| `M_FEDPROXY_DESTINATION_DENIED` | Error code | `IO.ELEMENT.MSC4512_FEDPROXY_DESTINATION_DENIED` |
| `M_FEDPROXY_PATH_NOT_ALLOWED` | Error code | `IO.ELEMENT.MSC4512_FEDPROXY_PATH_NOT_ALLOWED` |
| `M_FEDPROXY_CONNECTION_FAILED` | Error code | `IO.ELEMENT.MSC4512_FEDPROXY_CONNECTION_FAILED` |
```

`M_` is equivalent to `m.` in other identifiers, it's not supposed to be in the unstable prefix

Johennes: Actually it's listed the way I've done it in https://spec.matrix.org/proposals/#unstable-endpoints-features-and-vendor-prefixes. Is it just wrong there then?

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+would be beneficial is [lk-jwt-service], originally introduced in [MSC4195]. Reusing this service
+spares server maintainers having to integrate LiveKit SDKs and converse with the LiveKit SFU.
+Instead they could just run lk-jwt-service as a sidecar and proxy the respective endpoints to it.
+
+This proposal aims to formalise this delegation process in a generic way by enabling app services to
+claim parts of the Client-Server and Server-Server API.
+
+## Proposal
+
+Two new properties `proxy_prefix` and `proxy_url` are added to the top level of the [registration
```
thetayloredman, line 19: What benefits does this provide in contrast to just using a reverse proxy?

Johennes: Going through the homeserver means the server can handle request authorisation by checking access tokens and verifying federation signatures which spares the application service from doing that itself. As explained further down in the proposal text, the server strips the `Authorization` header before forwarding the request.

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+file] for application services. `proxy_prefix` is a string and defines the path prefix that the
+service is claiming. `proxy_url` is a string that determines the `url` to send proxied requests to.
+`proxy_prefix` and `proxy_url` are optional but if used, MUST always be specified together.
+
+``` yaml
```
turt2live:
```suggestion
```yaml
```

the syntax highlighting is broken after this code block for some reason. I'm hopeful it's the space?

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+to the application service and allows the service to trigger S-S requests under its own prefix.
+
+### Proxying client-server requests
+
+For any authenticated C-S request under `/_matrix/client/(unstable/[^/]+|v[^/]+)/{proxy_prefix}/.*`,
+the server first authorises the request as usual. If authorization succeeds, the server proxies the
+request to the same path anchored on `proxy_url` and streams the response back to the requesting
+client.
+
+Any "hop-by-hop" headers as defined by [RFC2616] MUST be stripped both before forwarding the request
```
turt2live:
```suggestion
Any "hop-by-hop" headers as defined by [RFC 2616] MUST be stripped both before forwarding the request
```

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+
+None.
+
+[^1]: Some statistics on what spec versions homeservers support and how long it took them to declare
+    support may be found on <https://patrick.cloke.us/homeserver-spec-versions/>.
+
+  [lk-jwt-service]: https://github.com/element-hq/lk-jwt-service
+  [MSC4195]: https://github.com/matrix-org/matrix-spec-proposals/pull/4195
+  [registration file]: https://spec.matrix.org/v1.19/application-service-api/#registration
+  [RFC2616]: https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
```
turt2live:
```suggestion
  [RFC 2616]: https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
```

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+For any authenticated C-S request under `/_matrix/client/(unstable/[^/]+|v[^/]+)/{proxy_prefix}/.*`,
+the server first authorises the request as usual. If authorization succeeds, the server proxies the
+request to the same path anchored on `proxy_url` and streams the response back to the requesting
+client.
+
+Any "hop-by-hop" headers as defined by [RFC2616] MUST be stripped both before forwarding the request
+to the service and before streaming the response back to the requesting client.
+
+Additionally, the `Authorization` header MUST be stripped on the forwarded request. Instead, the
+server supplies the MXID of the requesting client to the application service in a new request header
```
turt2live:
```suggestion
server supplies the User ID of the requesting client to the application service in a new request header
```

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+proxy_prefix: "foo/bar"
+proxy_url: "http://127.0.0.1:1234"
+```
+
+When both `proxy_prefix` and `proxy_url` are set, the server proxies matching C-S and S-S requests
+to the application service and allows the service to trigger S-S requests under its own prefix.
+
+### Proxying client-server requests
+
+For any authenticated C-S request under `/_matrix/client/(unstable/[^/]+|v[^/]+)/{proxy_prefix}/.*`,
```
turt2live: How does the server know the request is meant to be authenticated?

turt2live: likewise for Federation API

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
+# MSC4512: Delegating parts of the C-S and S-S API to application services
```
turt2live: nit: abbreviating "Client-Server API" and "Federation API" like this increases review time because it needs constant unpacking and careful reading. 

I strongly suggest not abbreviating API names, especially when the proposal works with both. This applies throughout the proposal.

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+- `status` (required, number): The HTTP status of the federation response.
+- `content` (object): A JSON object representing the body of the federation response.
+
+## Potential issues
+
+Running sidecar services reduces implementation work but complicates deployment of the homeserver.
+This is a tradeoff that server maintainers need to consider on a case by case basis.
+
+This proposal doesn't permit application services to proxy unauthenticated endpoints. This is by
+design as no associated use cases are known yet.
```
turt2live, line 141: Per other thread: the proposal doesn't actually say if the endpoints are authenticated or not, just "if they're authenticated...". If the intention is that all requests are authenticated, that should be stated in the proposal.

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Alternatives
+
+The existing `url` property could be reused to avoid having to specify a separate `proxy_url`. This
+would disallow using proxying while disabling namespace-related traffic from the server by setting
+`url` to `null`, however.
```
turt2live: This paragraph isn't clear, possibly because of all the negatives.

turt2live (changes requested):
Drive-by editorial review while looking at the MSC for potential use in another project. None of this is critical, but would block FCP down the line.

proposals/4512-app-service-takeover.md
Johennes: Good spot. Yes, that was the intention. Have clarified in 6f66ef1a011cc5469f2722bf80ac4c90582f7eea.

proposals/4512-app-service-takeover.md
Johennes: Have clarified in 6f66ef1a011cc5469f2722bf80ac4c90582f7eea.

proposals/4512-app-service-takeover.md
Johennes: Have tried to clarify it in 2d6005034c0d1489d5105cf36313e4c34fdeebb5. Does that help?

proposals/4512-app-service-takeover.md
Johennes: Ok, makes sense. I have spelled out all the abbreviations in b7f834bd4c0073ef1441ae8989d1a6fa44ae4b89.

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,180 @@
...
+spares server maintainers having to integrate LiveKit SDKs and converse with the LiveKit SFU.
+Instead they could just run lk-jwt-service as a sidecar and proxy the respective endpoints to it.
+
+This proposal aims to formalise this delegation process in a generic way by enabling app services to
+claim parts of the Client-Server and Server-Server API.
+
+## Proposal
+
+Two new properties `proxy_prefix` and `proxy_url` are added to the top level of the [registration
+file] for application services. `proxy_prefix` is a string and defines the path prefix that the
```
tonkku107, line 20: Maybe the prefix should be an array so that multiple prefixes can be claimed by the same application service. Some functionalities might require multiple routes but using a common prefix might end up consuming too many paths.

Alternatively, you could use a regex to claim paths like how users, aliases and rooms are done in namespaces.

Johennes: Hm, I wonder if just registering for the lowest required path could suffice?

For instance, when using this to implement the endpoints from #4195 in https://github.com/element-hq/lk-jwt-service, we register for `rtc/livekit` so that we can serve the `/get_token` and `/delegate_delayed_leave` endpoints without occupying anything else under `rtc/$foo`.

I guess if we'd add another endpoint under `rtc/livekit` that is _not_ supposed to be proxied by the same application service, that would be a problem though.

tonkku107: This is a hypothetical example to illustrate the point.

Let's say a moderation bot application service wants to implement `/_matrix/client/v3/rooms/{roomId}/ban` and `/_matrix/client/v3/rooms/{roomId}/kick`. The lowest required path here is `/rooms` which would proxy everything room related like `/_matrix/client/v3/rooms/{roomId}/messages` as well.

Now let's add support for the reporting endpoints which includes `/_matrix/client/v3/users/{userId}/report`. Now the lowest required path is `/` which means everything gets proxied to the application service.

Johennes: Yeah, sounds like allowing an array of prefixes will be easiest.

proposals/4512-app-service-takeover.md
```diff
@@ -0,0 +1,180 @@
...
+would be beneficial is [lk-jwt-service], originally introduced in [MSC4195]. Reusing this service
+spares server maintainers having to integrate LiveKit SDKs and converse with the LiveKit SFU.
+Instead they could just run lk-jwt-service as a sidecar and proxy the respective endpoints to it.
+
+This proposal aims to formalise this delegation process in a generic way by enabling app services to
+claim parts of the Client-Server and Server-Server API.
+
+## Proposal
+
+Two new properties `proxy_prefix` and `proxy_url` are added to the top level of the [registration
```
nico-famedly, line 19: I feel like always proxying from both APIs is a bit scary. I think most usecases will want to proxy only one or the other or even proxy different endpoints for federation vs client API. So maybe this could be split up into 2 options or if we want a list anyway, have it similar to the listeners in Synapse, where you can specify "client", "federation" or both? (And in the future potentially further options like authentication, etc)

proposals/4512-app-service-takeover.md
nico-famedly: it also significantly simplifies deployments imo. You just add the appservice with its configuration instead of also having to add appservice paths to the reverse proxy configuration.

proposals/4512-app-service-takeover.md
tulir: Oh, I had missed this reply. Yes, that doc is wrong. It looks like a comment on the PR even pointed that out, but it was missed? https://github.com/matrix-org/matrix-spec/pull/2222#discussion_r2405506479

tulir: https://github.com/matrix-org/matrix-spec/pull/2446

Johennes: Oh, indeed! 🤦‍♂️

PR #4513, opened 2026-07-21
MSC4513: Standard HTTP status codes for all endpoints, current and future

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/standard-error-codes/proposals/4513-standard-http-status-codes.md)

Owning team: SCT


proposals/4513-standard-http-status-codes.md
turt2live, line 1: Implementation requirements: none - the MSC is already effectively implemented by Synapse for ~11 years.

PR #4514, opened 2026-07-21
MSC4514: EDU size limit

### LINKED ISSUE #807: We should spec the max size of an EDU over federation
Currently, there is effectively no limit to the size of a `/_matrix/federation/v1/send/{txnId}` request, because some EDUs (eg presence, receipts) can reasonably contain an unlimited quantity of data. I think we should specify some limits to the amount of data such EDUs can contain.

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/edu-limit/proposals/4514-edu-size-limit.md)

Owning team: SCT

Fixes https://github.com/matrix-org/matrix-spec/issues/807

----

SCT Stuff

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4514#issuecomment-5048928145)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4514#issuecomment-5050747212)

### DIFF
--- a/proposals/4514-edu-size-limit.md
+++ b/proposals/4514-edu-size-limit.md
@@ -1,0 +1,71 @@
+# MSC4514: EDU size limit
+
+This proposal clarifies/specifies that [EDUs](https://spec.matrix.org/v1.19/server-server-api/#edus)
+are limited to 65535 bytes. See [this issue](https://github.com/matrix-org/matrix-spec/issues/807)
+for more detail.
+
+
+## Proposal
+
+EDUs MUST NOT exceed 65536 bytes. This matches the [size limit for events](https://spec.matrix.org/v1.19/client-server-api/#size-limits).
+
+This implicitly means that [to-device messages](https://spec.matrix.org/v1.19/server-server-api/#send-to-device-messaging)
+and other information bundled into an EDU cannot exceed 65535 bytes. Clients MAY encounter errors if
+they attempt to send information larger than the implied EDU permits. For example, a large to-device
+message.
+
+Note that this limit also implicitly creates a maximum size for a [`/send`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv1sendtxnid)
+transaction. Servers SHOULD split transactions over multiple requests if they encounter
+[413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413) errors on the endpoint.
+
+Parallel to [MSC4513](https://github.com/matrix-org/matrix-spec-proposals/pull/4513), this proposal
+permits 413 HTTP status codes to be returned on the following endpoints because they (sometimes)
+generate EDUs in normal operation:
+
+* [`PUT /_matrix/federation/v1/send/:txnId`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv1sendtxnid)
+* [`PUT /_matrix/client/v3/sendToDevice/:eventType/:txnId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid)
+* [`POST /_matrix/client/v3/rooms/:roomId/receipt/:receiptType/:eventId`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3roomsroomidreceiptreceipttypeeventid)
+* [`PUT /_matrix/client/v3/presence/:userId/status`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3presenceuseridstatus)
+* [`PUT /_matrix/client/v3/rooms/:roomId/typing/:userId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3roomsroomidtypinguserid)
+* [`PUT /_matrix/client/v3/devices/:deviceId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3devicesdeviceid)
+* [`POST /_matrix/client/v3/keys/device_signing/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysdevice_signingupload)
+* [`POST /_matrix/client/v3/keys/signatures/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keyssignaturesupload)
+
+Servers SHOULD return a [standard Matrix error](https://spec.matrix.org/v1.19/client-server-api/#standard-error-response)
+with the 413 status code, but are not required to. This is in alignment with MSC4513.
+
+
+## Potential issues
+
+Large EDUs were possible before, but were somewhat likely to encounter 413 or similar status codes in
+the wild due to reverse proxy configurations. Those applications will need to be adjusted to fit their
+data into a smaller form factor.
+
+Noted by [synpase#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
+implementations to split up to-device messages such that they're larger than the individual EDU size
+limit. This proposal doesn't carve out an exception for this because it's believed that a to-device
+message *could* be rejected at client send time with the size of the EDU's boilerplate being known.
+If it ends up being that some EDUs need to be larger than the target, the target SHOULD be adjusted
+by this proposal.
+
+
+## Alternatives
+
+Alternatives largely amount to "don't do this", so aren't discussed here.
+
+
+## Security considerations
+
+This MSC is a security consideration.
+
+
+## Unstable prefix
+
+None possible - implementations need to be able to reject large EDUs without namespacing.
+
+
+## Dependencies
+
+This proposal would benefit from [MSC4513](https://github.com/matrix-org/matrix-spec-proposals/pull/4513),
+but is deliberately not blocked by it. The relevant portions of MSC4513 have
+been minimally incorporated into this proposal.
--- a/proposals/4514-edu-size-limit.md
+++ b/proposals/4514-edu-size-limit.md
@@ -42,7 +42,7 @@ Large EDUs were possible before, but were somewhat likely to encounter 413 or si
 the wild due to reverse proxy configurations. Those applications will need to be adjusted to fit their
 data into a smaller form factor.
 
-Noted by [synpase#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
+Noted by [synapse#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
 implementations to split up to-device messages such that they're larger than the individual EDU size
 limit. This proposal doesn't carve out an exception for this because it's believed that a to-device
 message *could* be rejected at client send time with the size of the EDU's boilerplate being known.
--- a/proposals/4514-edu-size-limit.md
+++ b/proposals/4514-edu-size-limit.md
@@ -1,16 +1,17 @@
 # MSC4514: EDU size limit
 
 This proposal clarifies/specifies that [EDUs](https://spec.matrix.org/v1.19/server-server-api/#edus)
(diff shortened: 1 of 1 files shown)


proposals/4514-edu-size-limit.md
turt2live, line 1: Implementation requirements: none - this is already implemented by Synapse (as noted in the MSC).

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,71 @@
+# MSC4514: EDU size limit
+
+This proposal clarifies/specifies that [EDUs](https://spec.matrix.org/v1.19/server-server-api/#edus)
+are limited to 65535 bytes. See [this issue](https://github.com/matrix-org/matrix-spec/issues/807)
+for more detail.
+
+
+## Proposal
+
+EDUs MUST NOT exceed 65536 bytes. This matches the [size limit for events](https://spec.matrix.org/v1.19/client-server-api/#size-limits).
```
clokep: The rest of the proposal uses 6553*5*.

Is this when formatted using canonical JSON?

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,72 @@
...
+with the 413 status code, but are not required to. This is in alignment with MSC4513.
+
+
+## Potential issues
+
+Large EDUs were possible before, but were somewhat likely to encounter 413 or similar status codes in
+the wild due to reverse proxy configurations. Those applications will need to be adjusted to fit their
+data into a smaller form factor.
+
+Noted by [synpase#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
```
nico-famedly:
```suggestion
Noted by [synapse#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
```

turt2live: thnaks

clokep (approved):

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,72 @@
...
+* [`POST /_matrix/client/v3/rooms/:roomId/receipt/:receiptType/:eventId`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3roomsroomidreceiptreceipttypeeventid)
+* [`PUT /_matrix/client/v3/presence/:userId/status`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3presenceuseridstatus)
+* [`PUT /_matrix/client/v3/rooms/:roomId/typing/:userId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3roomsroomidtypinguserid)
+* [`PUT /_matrix/client/v3/devices/:deviceId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3devicesdeviceid)
+* [`POST /_matrix/client/v3/keys/device_signing/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysdevice_signingupload)
+* [`POST /_matrix/client/v3/keys/signatures/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keyssignaturesupload)
+
+Servers SHOULD return a [standard Matrix error](https://spec.matrix.org/v1.19/client-server-api/#standard-error-response)
+with the 413 status code, but are not required to. This is in alignment with MSC4513.
```
clokep: Is there any recommended `errcode`?

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## Potential issues
```
clokep, line 40: We should probably at least mention this is technically not backwards compatible?

clokep:
MSCs proposed for Final Comment Period (FCP) should meet the requirements outlined in the [checklist](https://github.com/matrix-org/matrix-spec-proposals/blob/main/MSC_CHECKLIST.md?rgh-link-date=2026-06-30T14%3A25%3A30Z) prior to being accepted into the spec. This checklist is a bit long, but aims to reduce the number of follow-on MSCs after a feature lands.

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Unstable prefix
  - [x] Dependencies

turt2live:
This has received relatively positive review since being opened, so to avoid it becoming further delayed:

@mscbot fcp merge

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,79 @@
...
+
+EDUs MUST NOT exceed 65536 bytes when formatted as [Canonical JSON](https://spec.matrix.org/v1.19/appendices/#canonical-json).
+This matches the [size limit for events](https://spec.matrix.org/v1.19/client-server-api/#size-limits).
+
+This implicitly means that [to-device messages](https://spec.matrix.org/v1.19/server-server-api/#send-to-device-messaging)
+and other information bundled into an EDU cannot exceed 65536 bytes. Clients MAY encounter errors if
+they attempt to send information larger than the implied EDU permits. For example, a large to-device
+message.
+
+Note that this limit also implicitly creates a maximum size for a [`/send`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv1sendtxnid)
```
richvdh, line 18: > this limit also implicitly creates a maximum size for a `/send` transaction

How so? What is to stop me creating a huge transaction?

If you want to create a max size for a `/send` transaction, I'd be supportive of that, but I don't think there is any *implicit* limit.

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,79 @@
...
+transaction. Servers SHOULD split transactions over multiple requests if they encounter
+[413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413) errors on the endpoint.
```
richvdh, line 20: If the 413 error is happening due to an overlarge EDU rather than an overlarge transaction, then splitting the transaction ain't gonna help

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+## Potential issues
+
+Large EDUs were possible before, but were somewhat likely to encounter 413 or similar status codes in
+the wild due to reverse proxy configurations. Those applications will need to be adjusted to fit their
+data into a smaller form factor.
+
+Noted by [synapse#19617](https://github.com/element-hq/synapse/pull/19617), it's still possible for
+implementations to split up to-device messages such that they're larger than the individual EDU size
```
richvdh, line 47: what? How does *splitting* to-device messages produce overlarge EDUs?

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,79 @@
...
+
+Note that this limit also implicitly creates a maximum size for a [`/send`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv1sendtxnid)
+transaction. Servers SHOULD split transactions over multiple requests if they encounter
+[413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/413) errors on the endpoint.
+
+Parallel to [MSC4513](https://github.com/matrix-org/matrix-spec-proposals/pull/4513), this proposal
+permits 413 HTTP status codes to be returned on the following endpoints because they (sometimes)
+generate EDUs in normal operation:
+
+* [`PUT /_matrix/federation/v1/send/:txnId`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv1sendtxnid)
```
richvdh, line 26: Is this trying to say: "If you receive an overlarge EDU in `/send`, the correct behaviour is to return a 413"? That would be a sensible thing to say, but it would be good to make it explicit.

proposals/4514-edu-size-limit.md
```diff
@@ -0,0 +1,79 @@
...
+* [`PUT /_matrix/client/v3/sendToDevice/:eventType/:txnId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid)
+* [`POST /_matrix/client/v3/rooms/:roomId/receipt/:receiptType/:eventId`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3roomsroomidreceiptreceipttypeeventid)
+* [`PUT /_matrix/client/v3/presence/:userId/status`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3presenceuseridstatus)
+* [`PUT /_matrix/client/v3/rooms/:roomId/typing/:userId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3roomsroomidtypinguserid)
+* [`PUT /_matrix/client/v3/devices/:deviceId`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3devicesdeviceid)
+* [`POST /_matrix/client/v3/keys/device_signing/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysdevice_signingupload)
+* [`POST /_matrix/client/v3/keys/signatures/upload`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keyssignaturesupload)
```
richvdh, line 33: These guys don't directly create EDUs. Rather, they create data, which may later be turned into EDUs for sending over federation.

Rather than saying "they (sometimes) generate EDUs in normal operation", could we instead say "they generate data which must be mapped onto EDUs"?

proposals/4514-edu-size-limit.md
nico-famedly: You are limited to 100 EDUs of a specific maximum size and 50 PDUs of a specific maximum size. The only other fields are origin and the timestamp, which both have reasonable maximum sizes as well as headers. As such there can now be a reasonable limit assumed for the transaction, even if you allow for some unsupported, namespaced, custom fields. The only way to create a huge transaction is via additional properties. Those don't have a size limit though, so not sure, if it would be safe to reject such txns, if they exceed a certain size.

richvdh: Yes, you can come up with a reasonable limit for the maximum size of an incoming transaction. If that's what the intention of this section of the proposal is, then it would be good to call it out.

Currently I could be sending transactions which include an unspecified key with a very very long value. According to the wording of the spec, that is spec-compliant, and changing the max size of an EDU does nothing to change that.

It would be entirely reasonable to change the spec here, but let's make it explicit that's what we're doing rather than \<handwave> implicit limit \</handwave>

turt2live: ah, true, I forgot about the other fields in the transaction. This should probably clarify that the transaction contents have a limit, but that doesn't help the overall `/send` request body limit. It does get us closer though.

proposals/4514-edu-size-limit.md
turt2live: this MSC doesn't aim to solve that problem: if a server sends invalid data, it gets to deal with those consequences.

proposals/4514-edu-size-limit.md
turt2live: I suppose this means after *attempting to* split them up - the PR mentions that some to-device messages can't be broken up (for reasons I don't quite follow myself).

proposals/4514-edu-size-limit.md
richvdh: which is fine, but saying "if you get a 413, split your transaction" is just bad advice.

proposals/4514-edu-size-limit.md
richvdh: I think the confusion here is exactly what is meant by a "message". I think element-hq/synapse#19617 talks about a "message" as "a per-user payload of [`/sendToDevice`](https://spec.matrix.org/v1.17/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid)", whereas the spec would call that *multiple* to-device messages.

PR #4516, opened 2026-07-23
Adapt MatrixRTC encryption to latest agreements

Johennes:
Changes from the latest discussion with @BillCarsonFr and @gewitternacht. Thanks @andybalaam for helping me figure out the crypto terminology.

Changes:

- Remove `claimed_device_id`
- Make encryption mandatory in encrypted rooms
- Forbid encryption in unencrypted rooms

### DIFF
--- a/proposals/4143-matrix-rtc.md
+++ b/proposals/4143-matrix-rtc.md
@@ -450,7 +450,10 @@ The schema for `m.rtc.encryption_key` to-device messages is as follows:
       "index": {index},
       "key": "{encoded_key}",
     },
-    "format": "0"
+"index": <index>,
+"key": "{encoded_key}",
+},
+"format": 0
 }
 ```
 
--- a/proposals/4143-matrix-rtc.md
+++ b/proposals/4143-matrix-rtc.md
@@ -111,7 +111,7 @@ neither `application_type` nor `application_slot_id` can contain the `#` charact
   - Optionally includes further properties for settings that are specific to the application
     `type`. The concrete properties are defined by the application's specification. A calling
     application, for instance, could include properties for constraining the call to be voice-only.
-- `encryption` (object): If present, describes the encryption mechanism to use in this slot. Further,
+- `encryption` (object): If present, describes the encryption mechanism to use in this slot. Further
   details on the available mechanisms can be found in the [encryption section] below. If absent,
   encryption is disabled.
   - `type` (required, string): The globally unique identifier of the encryption mechanism.
--- a/proposals/4143-matrix-rtc.md
+++ b/proposals/4143-matrix-rtc.md
@@ -239,7 +239,7 @@ joined if all of the following conditions apply:
 - `member.membership` equals `join`.
 - An open slot exists in the room state as an `m.rtc.slot` state event with `state_key` equalling
   the `m.rtc.member` event's `slot_id`.
-- The sender is currently a member of the room (i.e. has membership `join`).
+- The sender is currently a member of the room (i.e. has room membership `join`).
 - The event is currently sticky, meaning that its stickiness duration as per [MSC4354] has not expired.
   This is to ensure that the membership view is as consistent as possible across all members.
 
--- a/proposals/4143-matrix-rtc.md
+++ b/proposals/4143-matrix-rtc.md
@@ -422,23 +422,8 @@ type `m.rtc.encryption_key`.
 
 The recipient devices are determined from the `m.rtc.member` events that are considered to be
 joined to the slot. The conditions for considering a member joined were given
-[above](#joining-a-slot). Once the member events are determined, the target device IDs are
-obtained by:
-
-1. Identifying the Megolm session that was used to encrypt the `m.rtc.member` event.
-1. Looking up the [`m.room_key`] to-device message on which that session was received.
-1. If `sender_device_keys` exists in that message's decrypted [`OlmPayload`] and has
-   been [validated], take the device ID from the `device_id` property in `sender_device_keys`.
-1. If `sender_device_keys` does not exists, take the Curve25519 device identity key from the
-   `sender_key` property in the message's `m.room.encrypted` event body and look up the device
-   ID associated with that key via [`/keys/query`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysquery).
-
-Note that usually SDKs should already provide methods for obtaining and validating device IDs so that
-clients should not have to implement the logic above themselves.
-
-[`m.room_key`]: https://spec.matrix.org/v1.19/client-server-api/#mroom_key
-[`OlmPayload`]: https://spec.matrix.org/v1.19/client-server-api/#molmv1curve25519-aes-sha2
-[validated]: https://spec.matrix.org/v1.19/client-server-api/#validation-of-incoming-decrypted-events
+[above](#joining-a-slot). Once the member events are determined, the `m.rtc.encryption_key`
+to-device messages are sent to the devices that were used to encrypt these member events.
 
 The schema for `m.rtc.encryption_key` to-device messages is as follows:
 
@@ -476,16 +461,8 @@ Upon receipt, clients SHOULD discard any `m.rtc.encryption_key` events that were
 
 Receiving clients can determine the sender's `m.rtc.member` event by matching its `member.id` with
 the value of `member_id` in the `m.rtc.encryption_key` message. Once the member event was determined,
-clients perform the following checks:
-
-- The `sender` property in the message's [`OlmPayload`] matches the `sender` of the `m.rtc.member`
-  event.
-- The Curve25519 device identity key that was used to send the message matches the key that was
-  used to send the `m.rtc.member` event. Similar to the steps given above for obtaining the device
-  ID, the sender key can be acquired and validated using `sender_device_keys` or failing that by
-  taking the value from the encryption envelope and verifying it against `/keys/query`.
-
-Any `m.rtc.encryption_key` event that does not pass these checks MUST be discarded.
+clients verify that the sender and device that was used to send the member event match the sender
+and device of the to-device message. Otherwise the message MUST be discarded.
 
 In keeping with [MSC4153: Exclude non-cross-signed devices][MSC4153], clients SHOULD also discard
 `m.rtc.encryption_key` events when the sending device is not cross-signed.
--- a/proposals/4143-matrix-rtc.md
+++ b/proposals/4143-matrix-rtc.md
@@ -167,8 +167,7 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
       ... // Further application-specific properties (if required)
     },
     "member": {
-      "id": "{member_id}",
-      "claimed_device_id": "{device_id}"
+      "id": "{member_id}"
     },
     "transports": {
       "published": [
@@ -200,9 +199,6 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
   - `id` (required, string): Identifier to distinguish multiple members, even for the same user
     and device. MUST be unique for each join of the same user. This means that clients need to use a different
     identifier when leaving and then rejoining a slot.
-  - `claimed_device_id` (required, string) — Matrix device identifier. This is used to exchange
-    encryption keys as explained later in the [encryption section]. The device ID is untrusted ("claimed"
-    by the sender) and must be cross checked against the message encryption envelope for confirmation.
 - `transports` (object): Details on the MatrixRTC transports of this member. Other clients use the
   information in this object to determine how to connect to and exchange real-time data with this
   member. Clients should be prepared to connect to as many transports as there are members
@@ -227,10 +223,9 @@ joined if all of the following conditions apply:
 - The sender is currently a member of the room (i.e. has membership `join`).
 - The event is currently sticky, meaning that its stickiness duration as per [MSC4354] has not expired.
   This is to ensure that the membership view is as consistent as possible across all members.
-- If the room is encrypted, the `m.rtc.member` event was sent encrypted rather than in clear.
 
 If these conditions are not fulfilled, clients MUST treat the member as left and refrain
-from sending them encryption keys or connecting to their transports.
+from connecting to their transports.
 
 #### Leaving a slot
 
@@ -270,8 +265,7 @@ schema:
   in the addendum of [MSC4354]. MUST have the same value as `member.id` in the previously
   joined `m.rtc.member` event.
 
-Again, once a member has left, clients SHOULD refrain from sending them encryption keys
-or connecting to their transports.
+Again, once a member has left, clients SHOULD refrain from connecting to their transports.
 
 [lifecycle]: #membership-lifecycle
 
@@ -375,26 +369,26 @@ server-supported transport types:
 
 ### End-to-end encryption
 
-`m.rtc.member` events MUST be encrypted when sent in an encrypted room.
+Encryption in MatrixRTC has two layers. On the one hand, room events use the [existing mechanisms]
+for encrypting messages in rooms. On the other hand, applications also need a way to encrypt the RTC
+data itself. This process is generally specific to the transport being used, but often requires session
+members to agree on key material, at a minimum. To support this, MatrixRTC provides a generic system
+for establishing shared key material between members. Transports can then define how to actually use
+this key material, which may involve deriving further secrets from it. The concrete mechanism for
+agreeing on the shared key material within a slot is prescribed through the `encryption` object in
+`m.rtc.slot` events.
 
-Additionally, applications need a way to encrypt the RTC data itself. The process is generally
-specific to the transport being used, but often requires session members to agree on key material, at a minimum.
-To support this, MatrixRTC provides a generic system for establishing shared key material between
-members. Transports can then define how to actually use this key material, which may involve
-deriving further secrets from it.
+[existing mechanisms]: https://spec.matrix.org/v1.19/client-server-api/#end-to-end-encryption
 
-The concrete mechanism for agreeing on the shared key material within a slot is prescribed through
-the `encryption` object in `m.rtc.slot` events. Clients SHOULD enforce the use of encryption when
-opening a slot in encrypted rooms. When a client observes encryption being enabled in an `m.rtc.slot`
-event, it SHOULD set a flag to indicate that encryption should be used when being joined to this slot. This flag
-SHOULD NOT be cleared if a later `m.rtc.slot` event disables encryption. In other words, once encryption
-is enabled on a slot, it can never be disabled. This is to avoid a situation where a MITM can simply
-ask members to disable encryption.[^e2eeguide]
+Use of encryption in MatrixRTC is REQUIRED in encrypted rooms. This means that `m.rtc.member` events
+MUST be encrypted and `m.rtc.slot` events MUST contain an `encryption` object when sent in an encrypted
+room. Member / slot events that violate these conditions MUST be considered left / closed.
 
-[^e2eeguide]: This is aligned with the recommendation for handling the `m.room.encryption` state
-              event for normal room messaging in https://matrix.org/docs/matrix-concepts/end-to-end-encryption.
+Conversely, MatrixRTC encryption MUST NOT be used in unencrypted rooms. This is because the specific
+encryption mechanism introduced in this proposal is not well suited for unencrypted rooms. A future MSC
+may introduce another mechanism that lends itself better to unencrypted rooms.
 
-The only available mechanism for now is `m.per_member`.
+The only available encryption mechanism for now is `m.per_member`.
 
 ```json5
 {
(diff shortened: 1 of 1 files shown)


proposals/4143-matrix-rtc.md
```diff
@@ -442,7 +445,7 @@ is taken from the `member.claimed_device_id` property of the respective `m.rtc.m
 ```
 
 - `room_id` (required, string): The ID of the room that the slot is located in.
-- `member_id` (required, string): The `member.id` value of the target's `m.rtc.member` event.
+- `member_id` (required, string): The `member.id` value of the sender's `m.rtc.member` event.
```
Johennes, line 449: I think this was a bug in the proposal text that we didn't spot so far. This must be the sender's not the recipient's member ID. The recipient needs this to know on which member event to apply the key to decrypt the streams.

proposals/4143-matrix-rtc.md
```diff
@@ -423,8 +417,17 @@ type `m.rtc.encryption_key`.
...
 The recipient devices are determined from the `m.rtc.member` events that are considered to be
 joined to the slot. The conditions for considering a member joined were given
-[above](#joining-a-slot). Once joined members are determined, the target device ID
-is taken from the `member.claimed_device_id` property of the respective `m.rtc.member` event.
+[above](#joining-a-slot). Once the member events are determined, the target device IDs are
+obtained by:
+
+1. Identifying the Megolm session that was used to encrypt the `m.rtc.member` event.
+1. Looking up the [`m.room_key`] to-device message on which that session was received.
+1. Taking the device ID from the `OlmPlaintext` (formerly `OlmPayload`) of that message.
```
BillCarsonFr: 3 times 1. bullet point.

I would remove the last one
~~Taking the device ID from the `OlmPlaintext` (formerly `OlmPayload`) of that message.~~
- Use the device that sends the room key as target for the rtc key to device message. Apply the current user trust requirement (i.e. exclude insecure device, verification violations, etc.., similar to what is done when sharing room keys)

Johennes: > Use the device that sends the room key as target for the rtc key to device message

Is that not taking the device ID from the `OlmPayload`? If not, can we be specific about where it is taken from?

BillCarsonFr: > > Use the device that sends the room key as target for the rtc key to device message
> 
> Is that not taking the device ID from the `OlmPayload`? If not, can we be specific about where it is taken from?

What is that you are calling olm payload?

Techinically it when receiving an encrypted to device, you have the sender_key in the encrypted payload. Depending on the message type (Prekey / Message), you create an inbound session or you try to decrypt with the known existing session for that sender_key.

If you can successfully decrypt you know that the message was sent by the device owning the private part of that sender_key. You can find the device that is associated to this sender_key and get it's device_id.

While decrypting if you find any mismatch of what is in the encrypted payload you will have a decryption error, so at the end you could say that you take what is in the payload. But the truth is that you take the DecryptionResult to see what is the sender not anything in any json.  See here https://github.com/matrix-org/matrix-rust-sdk/blob/fd95d90bf9f93a3aa68376b1cbdffc3cada2c7b3/crates/matrix-sdk-crypto/src/olm/account.rs#L121

Johennes: > What is that you are calling olm payload?

It's the `OlmPayload` type defined in the spec under https://spec.matrix.org/v1.18/client-server-api/#molmv1curve25519-aes-sha2 (closest link, need to scroll down a little).

> Techinically it when receiving an encrypted to device, you have the sender_key in the encrypted payload. Depending on the message type (Prekey / Message), you create an inbound session or you try to decrypt with the known existing session for that sender_key.
> 
> If you can successfully decrypt you know that the message was sent by the device owning the private part of that sender_key. You can find the device that is associated to this sender_key and get it's device_id.
> 
> While decrypting if you find any mismatch of what is in the encrypted payload you will have a decryption error, so at the end you could say that you take what is in the payload. But the truth is that you take the DecryptionResult to see what is the sender not anything in any json. See here [matrix-org/matrix-rust-sdk@`fd95d90`/crates/matrix-sdk-crypto/src/olm/account.rs#L121](https://github.com/matrix-org/matrix-rust-sdk/blob/fd95d90bf9f93a3aa68376b1cbdffc3cada2c7b3/crates/matrix-sdk-crypto/src/olm/account.rs#L121)

The issue is that for the MSC, we should ideally describe it in general terms without pointing to a specific SDK implementation. I have tried to trace what the Rust SDK does and align it with spec terminology. It looks like the SDK takes the device ID on `DecryptionResult` from `sender_device_keys` on the decrypted and validated `OlmPayload` if it exists. Otherwise, it takes the identity key from the encryption envelope and looks up the device ID from `/keys/query`.

I have now adapted the wording as follows:

> 1. Identifying the Megolm session that was used to encrypt the `m.rtc.member` event.
> 1. Looking up the [`m.room_key`] to-device message on which that session was received.
> 1. If `sender_device_keys` exists in that message's decrypted [`OlmPayload`] and has
>    been [validated], take the device ID from the `device_id` property in `sender_device_keys`.
> 1. If `sender_device_keys` does not exists, take the Curve25519 device identity key from the
>    `sender_key` property in the message's `m.room.encrypted` event body and look up the device
>    ID associated with that key via [`/keys/query`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysquery).

Does that look correct?

Johennes:
Since we were talking about it in parallel and it touches the same area, I have also introduced a `membership = join | leave` property on `m.rtc.member` events. This makes the intended membership more explicit and aligns with `m.room.member` events.

proposals/4143-matrix-rtc.md
```diff
@@ -423,12 +423,26 @@ type `m.rtc.encryption_key`.
 
 The recipient devices are determined from the `m.rtc.member` events that are considered to be
 joined to the slot. The conditions for considering a member joined were given
-[above](#joining-a-slot). Once joined members are determined, the target device ID
-is taken from the `member.claimed_device_id` property of the respective `m.rtc.member` event.
+[above](#joining-a-slot). Once the member events are determined, the target device IDs are
+obtained by:
+
+1. Identifying the Megolm session that was used to encrypt the `m.rtc.member` event.
```
BillCarsonFr: I still think that it is not right, and I don't think we should go in that many details here. Let's say:

The device that encrypted the `m.rtc.member` event should be the one that the `m.rtc.encryption_key` is sent to.

Check the [megolm spec](https://spec.matrix.org/v1.12/client-server-api/#mmegolmv1aes-sha2) for more details on authentication. (~The sender device is the device that sent the megolm key -that encrypted the member event- as a m.room_key message via the olm 1:1 to-device channel).

Other MSC that speak about this authentication never go in such details:
https://github.com/uhoreg/matrix-doc/blob/include_device_keys/proposals/4147-including-device-keys-with-olm-encrypted-events.md

Or maybe say that the cryptographic sender of a message can be obtain from the crypto SDK, usually as part of an EncryptionInfo structure?

BillCarsonFr (changes requested):

proposals/4143-matrix-rtc.md
```diff
@@ -153,23 +153,68 @@ in the addendum of [MSC4354] to construct a state-like store of membership event
...
+    and device. MUST be unique for each join of the same user. This means that clients need to use a different
+    identifier when leaving and then rejoining a slot.
+  - `membership` (required, string): The intended membership status. One of `join`, `leave`.
+- `application` (object): Describes the application that is running in the slot. REQUIRED if `membership = join`.
+  - `type` (required, string): The application's globally unique identifier; same as in `m.rtc.slot`.
+  - Optionally includes further properties for settings that are specific to the application
+    `type`. As in `m.rtc.slot`, the concrete properties are defined by the application's specification.
+    For example, a [Third Room](https://thirdroom.io) application could include approximate map positions,
+    allowing clients to avoid connecting to members outside their area of interest.
+- `transports` (object): Details on the MatrixRTC transports of this member. Other clients use the
```
gewitternacht, line 170: Just to be sure: It's intended that in contrast to `application`, `transports` is *not* required if `membership = join`?

proposals/4143-matrix-rtc.md
```diff
@@ -182,66 +227,40 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
         "{transport_type}",
         ...
       ]
-    },
-    "sticky_key": "{member_id}" // = member.id
```
gewitternacht, line 186: This makes the order of properties differ between the textual description and the leave-JSON snippet, and this JSON snippet, making it a bit harder to follow.

proposals/4143-matrix-rtc.md
```diff
@@ -182,66 +227,40 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
...
-- `sticky_key` (required, string): The sticky key for the ephemeral map algorithm as defined
-  in the addendum of [MSC4354]. MUST have the same value as `member.id`.
-
 Apart from having to match the above schema, an `m.rtc.member` event MUST only be considered to be
 joined if all of the following conditions apply:
 
+- `member.membership` equals `join`.
 - An open slot exists in the room state as an `m.rtc.slot` state event with `state_key` equalling
   the `m.rtc.member` event's `slot_id`.
 - The sender is currently a member of the room (i.e. has membership `join`).
```
gewitternacht: The part in parentheses is now easily confusable with the first bullet point (rtc membership vs room membership).


```suggestion
- The sender is currently a member of the room (i.e. has room membership `join`).
```

maybe?

proposals/4143-matrix-rtc.md
```diff
@@ -182,66 +227,40 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
...
 If these conditions are not fulfilled, clients MUST treat the member as left and refrain
-from sending them encryption keys or connecting to their transports.
+from connecting to their transports.
 
 #### Leaving a slot
 
-To voluntarily leave a slot, the client sends an `m.rtc.member` event with the following
-schema:
+To voluntarily leave a slot, the client sends an `m.rtc.member` event with `membership = leave`.
+The event's `sticky_key` MUST have the same value as `member.id` in the previously joined
```
gewitternacht: Is this requirement necessary?
As there's already the general
> `sticky_key` (required, string): The sticky key for the ephemeral map algorithm as defined in the addendum of [MSC4354]. MUST have the same value as `member.id`.

above, wouldn't it be better to drop this, and instead mention explicitly that an `m.rtc.member` event is concerning only the specified `member.id`?

For example, like so?
```suggestion
To voluntarily leave a slot, the client sends an `m.rtc.member` event for the desired `member.id` and with `membership = leave`.
```

proposals/4143-matrix-rtc.md
gewitternacht: Should it be mentioned here that the `encryption` object MUST be present, i.e. RTC encryption must be enabled, if room encryption is enabled, and the object MUST NOT be present, i.e. RTC encryption must be disabled, if room encryption is disabled?
But maybe that's undesired duplication of information with the section on encryption below, I'm not sure.

gewitternacht: Irritating comma after "Further".
```suggestion
- `encryption` (object): If present, describes the encryption mechanism to use in this slot. Further
```

proposals/4143-matrix-rtc.md
```diff
@@ -182,66 +227,40 @@ To join a slot, the client sends an `m.rtc.member` event with the following sche
...
-      [Common Namespaced Identifier Grammar] but without the namespacing requirements.
-    - Optionally includes further properties specific to the transport `type`. The concrete properties
-      are defined by the transport's specification. This could, for instance, include WebSocket URLs.
-  - `can_subscribe` (array): An array of transport types that the member is able to subscribe to.
-    Other members can use this as cue for deciding which transports to use to accommodate this member.
-- `sticky_key` (required, string): The sticky key for the ephemeral map algorithm as defined
-  in the addendum of [MSC4354]. MUST have the same value as `member.id`.
-
 Apart from having to match the above schema, an `m.rtc.member` event MUST only be considered to be
 joined if all of the following conditions apply:
```
gewitternacht, line 237: Not sure if this is mentioned anywhere already: Maybe this should include a bullet point on how several `m.rtc.member` events with the same `member.id` are to be treated? This feels kinda relevant for the section on distributing encryption keys below, which references back to here. But maybe we're fine leaving that unspecified and sending encryption keys to whatever device sent an otherwise valid join-`m.rtc.member` event, even if that means we send the keys to two different devices both claiming the same `member.id`?

proposals/4143-matrix-rtc.md
gewitternacht: Only noticing this now – the snippet is a bit confusing:
`index` and `format` are supposed to be numbers, but the placeholder of the former makes it look like an object, and the latter is enclosed in quotes, so maybe
```suggestion
"index": <index>,
"key": "{encoded_key}",
},
"format": 0
```

proposals/4143-matrix-rtc.md
Johennes: I think us historically having been a bit lax in spelling out the nitty-gritty details on crypto MSCs is what has led to crypto in the spec being a bit tedious to piece together. I'd very much be in favor of making this as explicit as we can because it'll help us validate that what we're proposing actually works and is safe.

For whatever it's worth, I asked Claude and it appears to think the description aligns with what the Rust SDK does.

gewitternacht: I'd agree with Johannes here: Such details (or the lack thereof) are exactly what might cause vulnerabilities in implementations later on.
I also agree that it's not really *pretty* to have that level of detail in here, and the ideal way would probably to link to some dedicated section of the spec that explains how to go from one piece of information to another one in a cryptographically safe way. But as long as that doesn't exist, I'd prefer to spell the details out here.

Maybe we could _additionally_ mention something like

> Or maybe say that the cryptographic sender of a message can be obtain from the crypto SDK, usually as part of an EncryptionInfo structure

though, i.e., hint at the fact that these steps usually don't need to be reimplemented manually, as there should be functionality like that in a client library?

proposals/4143-matrix-rtc.md
```diff
@@ -455,14 +469,14 @@ is taken from the `member.claimed_device_id` property of the respective `m.rtc.m
...
+the value of `member_id` in the `m.rtc.encryption_key` message. Once the member event was determined,
 clients perform the following checks:
 
-- The `sender` property from the decryption result matches the `sender` of the `m.rtc.member`
+- The `sender` property in the message's [`OlmPayload`] matches the `sender` of the `m.rtc.member`
   event.
-- The `device_id` property from the decryption result matches the `member.claimed_device_id`
-  value of the `m.rtc.member` event.
+- The `device_id` property in the message's [`OlmPayload`] matches the device ID that was used to
+  send the `m.rtc.member` event (as determined by following the same steps given for the sender above).
```
gewitternacht: I'm not entirely sure this is what we should do here. As @dkasak mentioned in an internal discussion, device IDs are rather insecure, and can't really provide security guarantees.

For determining who to send the `encryption_key` *to*, I don't think we have much of a choice, as that's what the [`/sendToDevice`](https://spec.matrix.org/v1.19/client-server-api/#put_matrixclientv3sendtodeviceeventtypetxnid) endpoint takes as input.

But when receiving an `encryption_key` *from* someone, we maybe could do better, and instead of comparing `device_id`s, we could compare the actual curve25519 _encryption keys_ that were used to send the `m.rtc.encryption_key` message and to establish the Megolm session for the `m.rtc.member` event.

What I think this would protect against is a malicious homeserver swapping out the keys associated with the `device_id` that the `m.rtc.member` event was sent with, and then sending an `m.rtc.encryption_key` message encrypted with those new keys to the victim, and subsequently publishing an RTC stream using that `m.rtc.encryption_key`, thereby impersonating the real sender of the `m.rtc.member` event.

(I'm not too sure about all of this myself, as, like Johannes remarked, the crypto in the spec is a bit tedious to piece together, and I'm not _that_ familiar with all of it yet 😅)

gewitternacht: Though the homeserver should not be able to sign those new keys it wants to associate with the `device_id`, I suppose, so all of this is only a concern as long as we allow for non-cross-signed devices. And if we do allow for non-cross-signed devices, a homeserver can always impersonate its users anyway, I believe, so maybe it *is* okay to use the `device_id` here after all...

proposals/4143-matrix-rtc.md
Johennes: Actually yes, because you may send your membership event before connecting and publishing your transport.

proposals/4143-matrix-rtc.md
Johennes: Ah, good spot. Fixed in 9a42ef5f8b5ba7aea6e957a4e69f5fc18687c0a7.

proposals/4143-matrix-rtc.md
Johennes: Yes, that's better.

proposals/4143-matrix-rtc.md
Johennes: Yes, you're right. I suppose that's sufficient and less confusing.

proposals/4143-matrix-rtc.md
Johennes: Yeah, that was exactly my thinking. I've tried to keep all the encryption specific info in that section specifically because the context is already large and explaining it here, too, might negatively impact the overall readability (because it raises the question of why the requirement exists). I think this is made sufficiently clear and motivated in the encryption section.

proposals/4143-matrix-rtc.md
Johennes: Yeah, I think it's probably fine because we also scope on the user ID. So it would be the same user which means little security risk. The situation should also only occur if the client implementation is non-conforming. So I'd lean towards not spelling it out.

proposals/4143-matrix-rtc.md
Johennes: I think this may be a good point nevertheless.

> But when receiving an `encryption_key` _from_ someone, we maybe could do better, and instead of comparing `device_id`s, we could compare the actual curve25519 _encryption keys_ that were used to send the `m.rtc.encryption_key` message and to establish the Megolm session for the `m.rtc.member` event.

The identity key is either obtained from `sender_device_keys` or if that's missing it's taken from the envelope and validated against `/keys/query`. So that's the same process we currently describe just without the additional identity key > device ID mapping. In other words, I think we should use that instead of the current text when receiving messages as resolving the device ID doesn't add any extra security.

Johennes: How about this?

> - The Curve25519 device identity key that was used to send the message matches the key that was
>   used to send the `m.rtc.member` event. Similar to the steps given above for obtaining the device
>   ID, the sender key can be acquired and validated using `sender_device_keys` or failing that by
>   taking the value from the encryption envelope and verifying it against `/keys/query`.

proposals/4143-matrix-rtc.md
gewitternacht: To my understanding, these steps look correct.

However, in the context of [this comment](https://github.com/matrix-org/matrix-spec-proposals/pull/4516#discussion_r3645441633), I noticed that even though we're specifying how to determine the device ID that the `m.rtc.encryption_key` message is sent to, we still have to actually encrypt the to-device message with Olm, and for that, we also need to know what the receiving device's keys are.
For any one-time keys, we need to call `/keys/claim`. For obtaining the device identity key, I see two possibilities:
1. determining it analogously to the `device_id` steps given above (which actually already mention the device identity key), or
2. analogously to determining the one-time key, calling [`/keys/query`](https://spec.matrix.org/v1.19/client-server-api/#post_matrixclientv3keysquery) for the determined `device_id`.

My feeling is that 1. is better, but again, not sure, and I'm starting to get a bit confused about all of the ways to obtain keys right now :sweat_smile: 

edit: just saw that you actually [commented](https://github.com/matrix-org/matrix-spec-proposals/pull/4516#discussion_r3645596630) down below about identity key determination, not sure if this is exactly what I'm asking here?

Johennes: My understanding is that 1. is preferred if `sender_device_keys` exists (because it doesn't require another network request) and 2. is the fallback. I think that also aligns with https://spec.matrix.org/v1.19/client-server-api/#verification-of-the-sending-user-for-incoming-events.

However, I think that part I would consider already established in the spec. So we shouldn't have to repeat it?

proposals/4143-matrix-rtc.md
gewitternacht: Sounds good to me.
And agreed that even if my conclusion is correct that it might be okay to compare the `device_id` here, comparing the actual identity key is probably better in any case.

proposals/4143-matrix-rtc.md
Johennes: > though, i.e., hint at the fact that these steps usually don't need to be reimplemented manually, as there should be functionality like that in a client library?

Good point. Have done this in e1c6afeb1bc15383ec826fd61becdf903806ae2b.

I'll consider this resolved. It feels like we have enough confidence that the steps are correct and the level of details justified.

proposals/4143-matrix-rtc.md
Johennes: Great. Have applied this in 6de1680953ce1a7d740a1174e50e4de0585a621e.

proposals/4143-matrix-rtc.md
gewitternacht: Okay, yes, as 1. sort of includes 2. as a fallback already, it's really just the above steps again, and should be derivable from there and the rest of the spec :+1:

Johennes:
@BillCarsonFr and I had a discussion about the explicit steps for validating senders and devices. @BillCarsonFr would like these to be excluded from the MSC because he thinks they are incomplete. In the interest of time, I replace the steps with a generalized summary in e7b229a12ff3208c77dc32ce9a3203b9a1b4d46f. We can still spell them out in the spec PR if we want to.

Johennes:
Otherwise, I believe we have accumulated enough confidence here. Thanks both.

gewitternacht:
> In the interest of time, I replace the steps with a generalized summary in https://github.com/matrix-org/matrix-spec-proposals/commit/e7b229a12ff3208c77dc32ce9a3203b9a1b4d46f. We can still spell them out in the spec PR if we want to.

Okay, I trust you two had good reason to come to this conclusion, though I want to stress that I think it's important to get the details right here, and this should be picked up again for the spec PR then.

BillCarsonFr:
> > In the interest of time, I replace the steps with a generalized summary in [e7b229a](https://github.com/matrix-org/matrix-spec-proposals/commit/e7b229a12ff3208c77dc32ce9a3203b9a1b4d46f). We can still spell them out in the spec PR if we want to.
> 
> Okay, I trust you two had good reason to come to this conclusion, though I want to stress that I think it's important to get the details right here, and this should be picked up again for the spec PR then.

I think that the validation of megolm and olm message is out-of-scope of this MSC. There are exisiting spec effort to clarify that, like here https://github.com/matrix-org/matrix-spec/pull/2122 (verification of the sending user)  and we don't need to duplicate with partial info?

gewitternacht:
I see your point, and maybe the way we tried to list the steps here wasn't ideal, or focused on the wrong aspects. However, why I think it's not completely out-of-scope:

These details are needed for what this MSC tries to achieve, and they are not anywhere in the spec currently (to my knowledge), so I think it's a good and important opportunity to have a try at spelling them out – maybe, like I said above
> the ideal way would probably to link to some dedicated section of the spec that explains how to go from one piece of information to another one in a cryptographically safe way.

But even more importantly, I would argue that we're not only talking about "the validation of megolm and olm messages" here, but also the details on how new components (RTC encryption keys and the to-device messages to distribute them) tie into existing components (Olm/Megolm sessions), and it's crucial to make clear how they connect to one another, and which exact pieces of the existing components are used to guarantee security of the new components.

But like Johannes said, we can go with the more vague description for now, and revisit for the spec PR.

PR #4517, opened 2026-07-23
MSC4517: Federated thirdparty lookups

ara4n:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/federated-thirdparty-lookups/proposals/4517-federated-thirdparty-lookups.md)

Implementations:
 * https://github.com/matrix-org/matrix-bifrost/pull/386
 * https://github.com/element-hq/synapse/pull/19994
 * https://github.com/element-hq/element-web/pull/34406

### DIFF
--- a/proposals/4517-federated-thirdparty-lookups.md
+++ b/proposals/4517-federated-thirdparty-lookups.md
@@ -1,0 +1,135 @@
+# MSCXXXX: Federated third-party network lookups
+
+## Problem
+
+Matrix has always had [third-party lookup
+APIs](https://spec.matrix.org/v1.15/client-server-api/#third-party-networks): a client asks its
+homeserver to map an identifier on a bridged network (an XMPP JID, an IRC channel, a phone number)
+to the corresponding Matrix ghost user or portal room, and the homeserver asks its application
+services via the corresponding [AS API](https://spec.matrix.org/v1.15/application-service-api/#third-party-networks).
+
+However, this only works for bridges attached to *your own* homeserver. In practice bridges are
+run by somebody else: if montague.example runs an XMPP bridge, only montague.example's users can
+resolve `juliet@capulet.example` to `@_xmpp_juliet=40capulet.example:montague.example` - even
+though, once you know that MXID, DMing it federates perfectly well. Everyone else has to guess the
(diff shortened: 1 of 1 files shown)


proposals/4517-federated-thirdparty-lookups.md
turt2live, line 1: Implementation requirements:

* [x] Client (using)
* [x] Server (federating)
* [x] Appservice (offering)

proposals/4517-federated-thirdparty-lookups.md
```diff
@@ -0,0 +1,122 @@
...
+### Client behaviour
+
+Clients which offer a server picker for the room directory can now offer the same picker (and the
+same bridged-network dropdown) when starting DMs or searching for rooms, resolving typed
+third-party identifiers via whichever server the user selected. `instances` (and their
+`instance_id`s) are only meaningful on the server that reported them, so clients must track which
+server each instance came from and pass the same `server` when querying it.
+
+## Potential issues
+
```
Half-Shot, line 89: This isn't so much a potential issue with this proposal but; one of the reasons the thirdparty endpoints never gained traction is that appservices can't identify who made the request. Bridges which do not operate on a public network can't really make use of this feature at all, so at best this solves the problem of thirdparty lookups for a minority (XMPP, IRC and....obscurer protocols?)

So while this proposal doesn't really make things worse, it's bolting onto an API that we've been unable to really make use of because it's not very useful.

proposals/4517-federated-thirdparty-lookups.md
```diff
@@ -0,0 +1,122 @@
...
+    }
+  ]
+}
+```
+
+The queried server consults only its own application services when answering: it MUST NOT
+re-delegate the query to a third server, both to avoid loops and so that the answering server is
+always authoritative for the ghosts/portals it returns. These endpoints should be rate limited.
+
+Failures (timeouts, unknown endpoints on older servers) should be treated by the proxying server
```
reivilibre, line 76: This seems wrong: you requested your homeserver to talk to `another-server.example` and if it fails, you get a useless empty response.
This can't be distinguished from a genuine empty response.

It would be better for an error to be given back to the client and for clients to deal with the UI concern. This may include giving some feedback about the error.

PR #4518, opened 2026-07-23
MSC4518: Registries

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/registry/proposals/4518-registries.md)

----

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4518#issuecomment-5170101043)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4518#issuecomment-5170106045)

### DIFF
--- a/proposals/4518-registries.md
+++ b/proposals/4518-registries.md
@@ -1,0 +1,181 @@
+# MSC4518: Registries
+
+**Note**: This proposal formalizes a process rather than significant specification. It may be
+tested in production under the supervision of the Spec Core Team (SCT).
+
+As Matrix continues to grow and expand its capability, there is increased want for a structured place
+to put coordinated implementation-specific details. For example, event types which cover a feature
+that wouldn't necessarily be accepted into the "main" spec, transport options for VoIP media, or a
+set of common identifiers which aren't explicitly needed in Matrix itself.
+
+This proposal calls those structured places "registries", using [BCP 26 (RFC 8126)](https://datatracker.ietf.org/doc/html/rfc8126)
+as a set of guiding principles. The major difference between this proposal and BCP 26 is reduced governance
+overhead. Creation, population, removal, etc of a registry is already quite easy under BCP 26, though
+other IETF and IANA policies and procedures often layer to create more involved steps - Matrix does
+not have the same process requirements.
+
+More information about BCP 26 and this proposal's base principles can be found from the IETF Internet-Draft
+Author Resources: <https://authors.ietf.org/en/iana-considerations>
+
+
+## Proposal
+
+Registries consist of usually-optional functionality for a Matrix implementation. They are not part
+of the versioned specification, similar to how the [proposals page](https://spec.matrix.org/proposals/)
+is set up, but still have light governance to manage additions, removals, and changes. The precise
+deployment of a registry is left as an editorial detail - a GitHub repo with a bunch of Markdown docs
+is just as equally suitable as `https://spec.matrix.org/registry/whatever`.
+
+How functionality contained in a registry is discovered is left as a detail for the proposal(s) which
+define it, if one is needed. For example, a VoIP transport registry might also introduce a `/transports`
+endpoint so clients can discover those transports. Or, in the case of an event type-like registry, no
+discovery mechanism is necessarily required - clients would render the event types they support without
+needing to know what the server supports.
+
+When a registry should be used is not strictly defined. The intention is to give proposals an additional
+tool for improving interoperability without bloating the versioned specification. In general, a registry
+SHOULD NOT be used to list required identifiers, like join rules, but SHOULD be used to record optional
+extensions, like VoIP transports.
+
+The structure of a registry is left as a detail for the proposal(s) which define that registry. Some
+non-normative examples are included later in this proposal.
+
+The operation and what types of information are included in a registry are defined by normal MSCs for
+SCT review. The actual contents of a registry do not need to use MSCs, but unless otherwise defined
+by the registry's processes, MSCs are required.
+
+This setup ensures that the SCT is involved in changes to the "schema" of a registry, but is not
+required to be involved in the population of data in the registry. An external team or Governing Board
+Working Group could be given the ability to manage the contents of a registry. Though, to ensure
+registries can continue operating when volunteers are unavailable, the SCT is always able to modify
+the contents of a registry through normal MSCs.
+
+**Note**: If a non-SCT body chooses to use MSCs as the way to manage a registry's contents, those
+MSCs will still require SCT review and normal MSC process. Non-SCT bodies are encouraged to develop
+external approval/submission processes to avoid MSC process delay where needed.
+
+As a (deliberately lengthy) worked example:
+
+* MSC0001 defines an event types registry. The proposal states that the registry contains an event
+  `type`, description of what that event type is, and a schema for the event `content`. The proposal
+  is successfully merged through FCP review.
+* Later, the Event Types Working Group is established by the Governing Board and wants to take on
+  management of the registry from the SCT. The SCT has been managing the registry all this time
+  because MSC0001 didn't specify an alternative approval/curation process.
+
+  The new Working Group opens MSC0002 to change the registry's curation to the Working Group, but
+  doesn't remove the MSC requirement to add/remove/change contents.
+* MSC0003 is opened by a contributor to add their event type to the registry. The Working Group and
+  SCT both review the proposal. Ultimately, the MSC is merged after a successful FCP started by the
+  SCT.
+* The Working Group finds this dual review a little hard to manage, so they open MSC0004 to move
+  curation to a dedicated GitHub repo. The SCT reviews the proposal under normal process. The MSC
+  is successful during a FCP merge.
+* Another contributor wants to add to the event types registry, so they open an issue against the
+  dedicated GitHub repo. The SCT is no longer directly involved in review, so it's just the Working
+  Group which manages approval.
+* Yet another contributor wants to make a change to the registry, so they open a similar issue on
+  the dedicated GitHub repo. However, this time the Working Group isn't able to review it for some
+  reason. The SCT steps in and reviews the change, approving or denying it as required.
+* Later, Extensible Events lands in the specification and the Working Group wants to change the
+  schema of their registry to support Content Blocks and rendering options. Because they only have
+  curation capability, they need to open MSC0005 to get SCT review on the proposed schema changes.
+
+As shown, it's preferable that if a non-SCT body is to have curation capability then they should also
+avoid using MSCs as it creates extra process steps.
+
+### Non-normative registry examples
+
+The following are example (minimal) MSCs which demonstrate how a registry might be created and what
+they might contain. Any resemblance to other proposals, ideas, or thoughts is coincidental - this
+section is non-normative.
+
+#### Event type registry
+
+A new "Event Types" registry is established with the following details:
+
+* The event `type` using [Common Namespaced Identifier Grammar](https://spec.matrix.org/unstable/appendices/#common-namespaced-identifier-grammar).
+* A description of what the event is.
+* A schema for the event's `content`.
+
+The registry contains an additional table for Extensible Event Content Blocks:
+
+* The block `type`, also using Common Namespaced Identifier Grammar.
+* A description of what the block is used for/represents.
+* A schema for the block.
+
+The registry is curated through MSCs and normal process.
+
+#### VoIP transports registry
+
+A new `/available_transports` endpoint is made available. It returns a `transports` array containing
+identifiers from the below-defined transports registry. The array MAY be empty to indicate no transports
+are supported.
+
+The transports registry contains:
+* A transport ID using Common Namespaced Identifier Grammar.
+* A link to the proposal or other specification-like document which defines it. The contents of the
+  linked document MUST be backwards compatible with when the transport ID was first introduced.
+* A flag to indicate whether server implementations SHOULD support the transport.
+
+The registry is curated by the VoIP Working Group via the `matrix-org/voip-transports` GitHub repo.
+
+This MSC also introduces an initial `io.element.msc0006.livekit` transport to be included in the
+registry: ...
+
+#### T&S harms registry
+
+A new harm identifiers registry is established with the following details:
+
+* The identifier using Common Namespaced Identifier Grammar.
+* The minimum specification version where support for the identifier is mandatory. For example, if
+  this were to be `v1.19`, then servers MUST support receiving reports with this `harm` if they also
+  support at least `v1.19`.
+* A user-facing description for the identifier.
+
+The categories which contain the identifiers are left as an editorial detail.
+
+The registry is curated by the Foundation's T&S team via normal MSCs. This registry is not expected
+to be updated frequently.
+
+The following identifiers form the start of the registry: ...
+
+
+## Potential issues
+
+Registries can increase optionality too much in Matrix, leading to several capability negotiation
+endpoints being available. The SCT is expected to carefully review whether a registry is required,
+and ensure their prompt removal (via another MSC) as soon as they aren't needed.
+
+Registries might also increase burden on the existing MSC process by default. Proposals which create
+registries SHOULD aim to include as many starting values as possible to avoid flooding the MSC process.
+A future MSC can change the use of MSCs to curate/modify registries if needed.
+
+
+## Alternatives
+
+The only meaningful alternative is incorporating the functionality into the spec itself. Placing
+features like VoIP transports directly into the spec can hinder development because it's hard to change,
+remove, or adapt each transport when there's only 4 releases a year. Instead, by having a registry
+it's possible to iterate on the transport options faster. Later, when the set of transports is believed
+stable or ready for inclusion in the spec, the registry can be removed too.
+
+
+## Security considerations
+
+There is opportunity for spammy/malicious submissions to registries, just like there are for regular
+proposals. The SCT retains curation capability of all registries to remove such submissions without
+placing burden on the registry's managing team.
+
+
+## Unstable prefix
+
+This proposal defines a process, not something which can be prefixed. The SCT MAY explore using this
+proposal in production ahead of formal acceptance.
+
+Prefixes might still be used within a registry.
+
+
+## Dependencies
+
+This proposal has no direct dependencies.
--- a/proposals/4518-registries.md
+++ b/proposals/4518-registries.md
@@ -20,11 +20,26 @@ Author Resources: <https://authors.ietf.org/en/iana-considerations>
 
 ## Proposal
 
-Registries consist of usually-optional functionality for a Matrix implementation. They are not part
-of the versioned specification, similar to how the [proposals page](https://spec.matrix.org/proposals/)
-is set up, but still have light governance to manage additions, removals, and changes. The precise
-deployment of a registry is left as an editorial detail - a GitHub repo with a bunch of Markdown docs
-is just as equally suitable as `https://spec.matrix.org/registry/whatever`.
+Registries MUST only consist of optional functionality for a Matrix implementation. They have light
+governance to manage additions, removals, and changes. The precise deployment of a registry is left
+as an editorial detail - a GitHub repo with a bunch of Markdown docs is just as equally suitable as
+`https://spec.matrix.org/registry/whatever`. Registries nominally operate outside of the versioned
+specification, but MAY be included in the versioned spec for editorial ease, versioning, and visibility.
+
+**Note**: Required (non-optional) functionality MUST be placed into the main specification rather than
+a registry.
+
+The main specification continues to describe namespacing characteristics of its fields, including
+whether it uses a registry. When a registry is used is not strictly defined, but are most useful to
+describe allowed extensions, such as event types and VoIP transports. Registries MUST NOT be used to
+record required functionality, like join rules.
+
+The above critically means that existing identifiers under the [Common Namespaced Identifier Grammar](https://spec.matrix.org/v1.19/appendices/#common-namespaced-identifier-grammar)
+retain the ability to have unregistered custom extensions, if the field permits. As always, the presence
+of `m.*` does *not* automatically mean that unregistered custom extensions are permitted.
+
+Registries do not allow unregistered custom extensions by default. If a registry needs to support
+custom extensions, like an event types registry would, it MUST declare that.
 
 How functionality contained in a registry is discovered is left as a detail for the proposal(s) which
 define it, if one is needed. For example, a VoIP transport registry might also introduce a `/transports`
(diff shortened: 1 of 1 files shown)


proposals/4518-registries.md
turt2live, line 1: Implementation requirements:

* [x] Ideally 2+ MSCs wanting to use this

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+
+The registry contains an additional table for Extensible Event Content Blocks:
+
+* The block `type`, also using Common Namespaced Identifier Grammar.
+* A description of what the block is used for/represents.
+* A schema for the block.
+
+The registry is curated through MSCs and normal process.
+
+#### VoIP transports registry
```
Johennes, line 121: BCP 26 appears to define registries as a place to collect protocol parameters and this fits well with the examples of event types and harms taxonomies given above and below. VoIP / MatrixRTC feels a little different though. Transports involve both clients and servers and will often need endpoint support. This goes beyond pure identifiers and could have larger implications for e.g. compatibility or security. That makes it a less obvious registry example in my opinion.

Is it actually the intention that Matrix registries can also contain endpoints?

turt2live: they could, yes. The hypothetical example here is essentially doing that already by requiring some link to a specification in the registry. That specification can then say "the `org.example.msc0000.livekit` transport requires supporting the `/livekit` endpoint".

This is how the ports registry works too: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=ntp+5905 (NTP used as an example - for some reason despite "reference" being a required field, our reference is lost on `matrix-fed`). Another is `.well-known` URIs: https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml#:~:text=permanent-,%5BThe_Matrix.org_Foundation_C.I.C.%5D,-2019%2D04%2D16

proposals/4518-registries.md
turt2live: https://github.com/matrix-org/matrix-spec-proposals/pull/4519
https://github.com/matrix-org/matrix-spec-proposals/pull/4456

proposals/4518-registries.md
turt2live: ... as shown here: https://github.com/matrix-org/matrix-spec-proposals/pull/4519

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+
+More information about BCP 26 and this proposal's base principles can be found from the IETF Internet-Draft
+Author Resources: <https://authors.ietf.org/en/iana-considerations>
+
+
+## Proposal
+
+Registries consist of usually-optional functionality for a Matrix implementation. They are not part
+of the versioned specification, similar to how the [proposals page](https://spec.matrix.org/proposals/)
+is set up, but still have light governance to manage additions, removals, and changes. The precise
```
erikjohnston: What's the rationale for keeping it outside the versioned spec? I think with e.g. MatrixRTC transports it would be useful to have them defined in versioned specs; i.e. if you printed out spec version v1.19 then having the registries would be likely be very useful, as they are a necessary part of implementing fully fledged clients and servers.

They should live outside the CS/SS API versioned specs, as they use different capability/negotiations to determine support (rather than looking at the supported spec version).

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+
+#### T&S harms registry
+
+A new harm identifiers registry is established with the following details:
+
+* The identifier using Common Namespaced Identifier Grammar.
+* The minimum specification version where support for the identifier is mandatory. For example, if
+  this were to be `v1.19`, then servers MUST support receiving reports with this `harm` if they also
+  support at least `v1.19`.
+* A user-facing description for the identifier.
```
erikjohnston, line 148: Hmm, this feels like a dubious fit for a registry. IMO anything that is required in a CS API specific version should be in the main CS API spec (even if it's on a separate page).

Given that it's expected that this would still be updated via MSCs, it feels like the only benefit is that this is a separate page.

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+
+The operation and what types of information are included in a registry are defined by normal MSCs for
+SCT review. The actual contents of a registry do not need to use MSCs, but unless otherwise defined
+by the registry's processes, MSCs are required.
+
+This setup ensures that the SCT is involved in changes to the "schema" of a registry, but is not
+required to be involved in the population of data in the registry. An external team or Governing Board
+Working Group could be given the ability to manage the contents of a registry. Though, to ensure
+registries can continue operating when volunteers are unavailable, the SCT is always able to modify
+the contents of a registry through normal MSCs.
```
erikjohnston: I get  that this is just an example, but I'm slightly dubious about even suggesting we might give this to an external team or working group. Ultimately, the spec (including the registries) are our responsibility and so I think we need to think very carefully before allowing others to edit without any SCT input.

I certainly see the benefits of avoiding a full MSC process for some registries. For example, with the events registry we could make it so that no MSC is required if a) the event type is on a custom domain owned by the author, and b) at least one SCT member has signed off on it.

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+How functionality contained in a registry is discovered is left as a detail for the proposal(s) which
+define it, if one is needed. For example, a VoIP transport registry might also introduce a `/transports`
+endpoint so clients can discover those transports. Or, in the case of an event type-like registry, no
+discovery mechanism is necessarily required - clients would render the event types they support without
+needing to know what the server supports.
+
+When a registry should be used is not strictly defined. The intention is to give proposals an additional
+tool for improving interoperability without bloating the versioned specification. In general, a registry
+SHOULD NOT be used to list required identifiers, like join rules, but SHOULD be used to record optional
+extensions, like VoIP transports.
```
erikjohnston: This goes against the words below, but personally I think we should add:

> Registries must never require clients/servers to support something, everything must be optional. Any required functionality must be in the main spec.

And maybe:

> Registries are useful to document allowed extensions, such as event types or VoIP transports. The main spec MUST explicitly document what extensions are allowed, e.g. event types are explicitly allowed to have custom types, while join rules are not. Specifically, the presence of a `m.*` prefix on fields is *not* enough to infer that extensions are allowed (outside of experimental features).


> Registries MAY indicate (non-normatively) which items are common or recommended to be currently supported. E.g. the VoIP transport may indicate which is the currently most widely used transport, or which is the "latest generation".

erikjohnston (commented):
Mostly looks sane, though I have a few concerns with it as drafted.

proposals/4518-registries.md
turt2live: They live outside the versioned spec primarily to let them get updated faster than spec releases allow. https://github.com/matrix-org/matrix-spec-proposals/pull/4519 (RTC transports) as currently written for example auto-registers all unstable MSCs so they can be supported by implementations - waiting for a spec release would mean that those registrations are stuck waiting 3-6 months before they can be used. 

Versioned things that exist outside of the Client-Server or Federation API would just be appendices. We can also have those, but their purpose is slightly different.

proposals/4518-registries.md
turt2live: There's strong desire to have harm identifiers in particular outside the spec, but we also need to make sure that servers actually respect the identifiers themselves. We could invert the association here such that the versioned spec states which identifiers are supported, though administrative effort would be higher. 

More discussion about the specific case can be found here: https://github.com/matrix-org/matrix-spec-proposals/pull/4456#discussion_r3174286997

proposals/4518-registries.md
turt2live: Registries would require an MSC to change or set their governance model, so we can review this on a case-by-case basis. There may be some registries we absolutely do not want to hand over to an external body - we can raise those concerns on the MSCs which attempt to do so. This proposal gives us the flexibility to do that in the future, if we want/need it.

Using normal spec PRs for an events registry is entirely possible.

proposals/4518-registries.md
turt2live: This could be a good middle ground for https://github.com/matrix-org/matrix-spec-proposals/pull/4518#discussion_r3664918019

proposals/4518-registries.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Potential issues
+
+Registries can increase optionality too much in Matrix, leading to several capability negotiation
+endpoints being available. The SCT is expected to carefully review whether a registry is required,
+and ensure their prompt removal (via another MSC) as soon as they aren't needed.
+
+Registries might also increase burden on the existing MSC process by default. Proposals which create
+registries SHOULD aim to include as many starting values as possible to avoid flooding the MSC process.
+A future MSC can change the use of MSCs to curate/modify registries if needed.
```
anoadragon453: Can an MSC introducing a registry also simultaneously declare that the registry may be updated outside of the MSC process?

I see no reason to create such a rule, but this sentence sounds like that may be the case.

turt2live: Yes, a proposal can seed the registry and declare that other updates need to go through a different process. If the proposal isn't sure about the initial values, it would be best to exclude them and leave it to the post-acceptance review process (whatever that is for the registry).

proposals/4518-registries.md
erikjohnston: Ah, so there are I think two distinct questions here:
1. How do we publish things with minimal lag
2. Do we version the registries alongside the main spec.

I don't think it's just registries that suffer from us only updating the main spec on version releases. The "obvious" answer is to have a `/latest/` which is updated immediately on a spec PR merging, but that has issues around how we document new APIs etc.

---

> Versioned things that exist outside of the Client-Server or Federation API would just be appendices. We can also have those, but their purpose is slightly different.

Well, appendices are an example of things that live outside the CS or Fed API specs, but that does not mean they are the only thing that can. There is even an argument that registries can be/are basically just appendices.

---

> #4519 (RTC transports) as currently written for example auto-registers all unstable MSCs so they can be supported by implementations - waiting for a spec release would mean that those registrations are stuck waiting 3-6 months before they can be used.

As a side note: we support that today with namespacing and unmerged MSCs? Nothing is stopping people from using unstable stuff that isn't in the registry. Having a table of unstable types pointing to the MSCs could be useful, but that doesn't block people experimenting with it.

proposals/4518-registries.md
erikjohnston: I have a very strong feeling that if something is required to be implemented in a spec version it is defined in that spec version. In fact, I'd go so far as to say that if something is required in a spec version it is de facto *in* that spec version regardless of where it lives. Having it documented separately in an unversioned page feels confusing to consumers of the spec.

(Will comment on that thread as well).

proposals/4518-registries.md
turt2live: To clarify: there is no distinction between Appendices and Client-Server or Federation APIs from a core spec perspective. It's an editorial decision where to put their contents, including what the pages are called. We could split the Client-Server API into a dozen different documents without needing an MSC, for example. The "core spec" is everything under the spec version, regardless of what the page's title is.

I don't think we want registries to be part of the core spec, which means we don't bring them under the versioned spec. If we want them as part of the core spec, then we'd indeed just use appendices and ignore this MSC entirely. 

Our spec process does support namespacing and unstable features, but given registries are expected to have ties to the core spec it feels important to explicitly list what's in there, including the unstable stuff. Otherwise, implementations would need to go on an adventure to figure out why `/transports` is returning unknown transports.

erikjohnston: To be clear I think it's a relatively minor point *in practice* about whether we should consider registries as part of the versioned spec or not.

> To clarify: there is no distinction between Appendices and Client-Server or Federation APIs from a core spec perspective. It's an editorial decision where to put their contents, including what the pages are called. We could split the Client-Server API into a dozen different documents without needing an MSC, for example. The "core spec" is everything under the spec version, regardless of what the page's title is.

They're all part of the wider spec, yes. The CS and SS API spec though are very much a logical groupings and not just editorial. The CS API in particular, where the capability negotiation is done by clients querying the supported spec version via `/versions`.

The SCT does have broad remit to organise the spec how we like.

> I don't think we want registries to be part of the core spec, which means we don't bring them under the versioned spec.

Can you expand as to why? And also what you mean by them being outside the core spec in practice? Is it just that you wouldn't be able to see older versions of the registry pages on spec.matrix.org (like the "Spec Change Proposals" section that takes you to `/unstable/`)?

For me, conceptually, I would see registries as part of the "spec" as a whole. It documents how client and servers can interact with one another. The main difference with the existing CS spec is that they are optional and have different capability negotiations that make them less suitable to be documented alongside the other CS APIs.

Though I agree that ideally we would by default always show the current/latest/unstable registries so that updates get seen immediately.

> If we want them as part of the core spec, then we'd indeed just use appendices and ignore this MSC entirely.

It's still useful to have an MSC to outline a new concept, even if we could fudge it. I don't think there is anything in the MSC as it stands that is necessary to do as a standalone MSC either.

Probably the only *practical* difference between having registries part of the "versioned" spec or not is simply if you could look at the registries as they were at past spec versions. This helps e.g. seeing how things changed over time, like what were the known transports at the time, when did something get deprecated, etc.

> Our spec process does support namespacing and unstable features, but given registries are expected to have ties to the core spec it feels important to explicitly list what's in there, including the unstable stuff. Otherwise, implementations would need to go on an adventure to figure out why `/transports` is returning unknown transports.

I don't follow this, what do you mean "have ties to the core spec"? You have the same issue with unstable features today, which is why the spec says to ignore unknown types (where applicable). I don't think it causes any particular confusion? Especially when people put the MSC number in the type. I don't think we can expect the registries to ever be fully complete, I don't believe we would require a type to be listed in a registry before it could be used (experimentally)?

Could you expand on why this would be useful, given it's different from what we do today?

turt2live: > > I don't think we want registries to be part of the core spec, which means we don't bring them under the versioned spec.
> 
> Can you expand as to why? And also what you mean by them being outside the core spec in practice? Is it just that you wouldn't be able to see older versions of the registry pages on spec.matrix.org (like the "Spec Change Proposals" section that takes you to `/unstable/`)?

Registries would just take up too much space relative to their importance in the spec, so it's best to put them nearby but outside the spec. Implementation-wise I'd expect registries to behave exactly like the proposals page, so we'd have git history if we wanted to see changes over time (I don't think we need to see history of registries as often as history of API endpoints).


> > Our spec process does support namespacing and unstable features, but given registries are expected to have ties to the core spec it feels important to explicitly list what's in there, including the unstable stuff. Otherwise, implementations would need to go on an adventure to figure out why `/transports` is returning unknown transports.
> 
> I don't follow this, what do you mean "have ties to the core spec"? You have the same issue with unstable features today, which is why the spec says to ignore unknown types (where applicable). I don't think it causes any particular confusion? Especially when people put the MSC number in the type. I don't think we can expect the registries to ever be fully complete, I don't believe we would require a type to be listed in a registry before it could be used (experimentally)?
> 
> Could you expand on why this would be useful, given it's different from what we do today?

Currently unstable features are conditional on `/versions` flags, meaning the endpoints required to use the feature are gated by a version check. With registries though, the endpoints themselves would be stable instead of unstable, but would potentially return or accept unstable identifiers too - these are the ties. Ignoring unknown data doesn't really work when the client is trying to decide what to submit to a server, if anything, on an existing stable endpoint. Clients receiving unknown data can indeed ignore it, but that's still only half the problem.

---

fwiw, I think we're both describing the same functional behaviour, but unsure we agree about where exactly the page titled "Registries" would live (and the process implications of that).

proposals/4518-registries.md
HarHarLinks, line 1: more prior art: https://matrix.directory/

turt2live: is there a repo behind this website?

HarHarLinks: https://github.com/BramvdnHeuvel/Matrix-Events-Directory/

proposals/4518-registries.md
turt2live: I've added text to make versioning an editorial decision rather than an operating one. It's a bit nuanced, but the distinction is to allow us to split it out of the versioned spec later if we want.

proposals/4518-registries.md
turt2live: I've adjusted the example for now, but the actual harms proposal will need further work.

proposals/4518-registries.md
turt2live: I've adjusted this to say that the SCT must always be involved somewhere along the process, but curation doesn't need to happen within the MSC process.

proposals/4518-registries.md
turt2live: I've incorporated all 3 suggestions into different parts of the proposal

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
Having addressed primarily Erik's review, I think this is ready to enter FCP.

@mscbot fcp merge

proposals/4518-registries.md
```diff
@@ -0,0 +1,203 @@
...
+  expected, another MSC can always be opened to change the curation process for that registry.
+
+* Registries which need to record required functionality are unable to do so. The best they can do is
+  flag which registry items are "suggested", though the core spec which actually uses the registry
+  can limit values which are inside the registry. For example, "this endpoint MUST only accept values
+  in the X registry". This is especially helpful if the registry itself is versioned under/at the same
+  time as the spec - the spec can reference a "version" of the registry.
+
+
+## Alternatives
```
HarHarLinks, line 177: This appears to be out of scope or I didn't understand if the intention was to include this, but anyway it's related enough to ask here: has the author/SCT considered indexing entirely off-spec registries as well? E.g. could say Nordeck establish a registry of Nordeck widget event types, or gematik a registry of medical event types, or FITKO of administration event types, etc, that would end up listed on spec.matrix.org in the interest of increasing discoverability.

turt2live: The overall intention is that registries are used for specified features, not for off-spec things. However, the event types registry, when/if it exists, would likely support all of your examples. Where things don't fit in a registry and aren't destined for the main spec itself, the intention is that those external parties create their own specification and publish it external to Matrix.

proposals/4518-registries.md
```diff
@@ -0,0 +1,203 @@
...
+Author Resources: <https://authors.ietf.org/en/iana-considerations>
+
+
+## Proposal
+
+Registries MUST only consist of optional functionality for a Matrix implementation. They have light
+governance to manage additions, removals, and changes. The precise deployment of a registry is left
+as an editorial detail - a GitHub repo with a bunch of Markdown docs is just as equally suitable as
+`https://spec.matrix.org/registry/whatever`. Registries nominally operate outside of the versioned
+specification, but MAY be included in the versioned spec for editorial ease, versioning, and visibility.
```
clokep, line 27: Should I read this as "it is OK if they appear under https://spec.matrix.org/v1.XX/"?

proposals/4518-registries.md
```diff
@@ -0,0 +1,203 @@
+# MSC4518: Registries
+
+**Note**: This proposal formalizes a process rather than significant specification. It may be
+tested in production under the supervision of the Spec Core Team (SCT).
+
+As Matrix continues to grow and expand its capability, there is increased want for a structured place
+to put coordinated implementation-specific details. For example, event types which cover a feature
+that wouldn't necessarily be accepted into the "main" spec, transport options for VoIP media, or a
+set of common identifiers which aren't explicitly needed in Matrix itself.
```
clokep, line 9: These don't really sound "implementation-specific"? I'm a bit confused by the rationale here.

proposals/4518-registries.md
turt2live: yes. I'll add more words.

proposals/4518-registries.md
turt2live: I did struggle to define what this stuff is: alternative wording suggestions are welcome. The general idea is we probably won't put `livekit` into the main spec, nor would we put `org.example.xray` event types into the main spec. We would instead put them somewhere else for the main spec to reference ("hey, if you want to do VoIP, check out the transports registry for common transports").

proposals/4518-registries.md
richvdh, line 1: Honestly I'm finding this a bit hard to draw any conclusions over. Everything of value seems to be left up to the MSC that introduces the actual registry (which I think is fine). It ends up feeling like a rather process-heavy way to say "MSC4519 is allowed to add a section to the Appendices and define how it will be maintained".

On which basis: no objections, but I feel like we could live without an MSC to do this. Let MSC4519 do its thing, and we'll learn from it and apply any conclusions next time we want to do something similar (MSC4456, I guess).

Perhaps I'm missing something though. In fairness I think I missed the SCT meeting when this idea was discussed.

turt2live: The end goal here is to *not* use appendices, but rather something adjacent. It's certainly nuanced, but I think an MSC for this is important to get broad agreement that we're okay with the concept of registries. 

Registries shouldn't be any more process heavy than the rest of the spec. If we want to do different though, that's where commentary on this MSC is important.

proposals/4518-registries.md
turt2live: 9f7b02d085380705b886a6afd71dc0a18580755b

proposals/4518-registries.md
Johennes: I can see how "implementation-specific" could be misinterpreted. In my mind, what we could use registries for are features that are too specialized to be included in the main spec but used widely enough to make a plain unmerged MSC insufficient for coordinating compatibility. Maybe something like this would help express that?

```suggestion
As Matrix continues to grow and expand its capability, there is increased want for a structured place
to coordinate details of features that are too specialized to warrant inclusion in Matrix itself but have
a large enough user base to require coordination in a more regulated form than a plain unmerged MSC.
Examples for this include event types that wouldn't necessarily be accepted into the "main" spec (such
as the FHIR events from [MSC4302]), transport options for VoIP media, or a set of common identifiers
which aren't explicitly needed in Matrix itself.

[MSC4302]: https://github.com/matrix-org/matrix-spec-proposals/pull/4302
```

proposals/4518-registries.md
```diff
@@ -0,0 +1,204 @@
...
+specification, but MAY be included in the versioned spec (`https://spec.matrix.org/vX.Y/registry/whatever`)
+for editorial ease, versioning, and visibility.
+
+**Note**: Required (non-optional) functionality MUST be placed into the main specification rather than
+a registry.
+
+The main specification continues to describe namespacing characteristics of its fields, including
+whether it uses a registry. When a registry is used is not strictly defined, but are most useful to
+describe allowed extensions, such as event types and VoIP transports. Registries MUST NOT be used to
+record required functionality, like join rules.
```
Johennes, line 36: I think there's a word missing here?

```suggestion
The main specification continues to describe namespacing characteristics of its fields, including
whether it uses a registry. When a registry is used is not strictly defined, but registries are most useful to
describe allowed extensions, such as event types and VoIP transports. Registries MUST NOT be used to
record required functionality, like join rules.
```

proposals/4518-registries.md
```diff
@@ -0,0 +1,204 @@
...
+  doesn't remove the MSC requirement to add/remove/change contents.
+* MSC0003 is opened by a contributor to add their event type to the registry. The Working Group and
+  SCT both review the proposal. Ultimately, the MSC is merged after a successful FCP started by the
+  SCT.
+* The Working Group finds this dual review a little hard to manage, so they open MSC0004 to move
+  curation to a dedicated GitHub repo. The SCT reviews the proposal under normal process. The MSC
+  is successful during a FCP merge.
+* Another contributor wants to add to the event types registry, so they open an issue against the
+  dedicated GitHub repo. The SCT is no longer directly involved in review, so it's just the Working
+  Group which manages approval.
```
Johennes, line 87: This appears to be at odds with the requirement for SCT involvement from above?

> If the curation happens outside of the MSC process, SCT review MUST happen elsewhere in that curation process.

proposals/4518-registries.md
HarHarLinks: I am increasingly confused about what ridge this MSC is trying to walk. The spec leading me to discover an SCT-sanctioned list of VoIP transports sounds very in-spec to me, except with less bureaucracy to update said list.

proposals/4518-registries.md
richvdh: > It's certainly nuanced, but I think an MSC for this is important to get broad agreement that we're okay with the concept of registries.

I find it very hard to have an opinion on this in the abstract. Feels like the devil is in the detail, and there is no detail here.

turt2live: I don't feel it's fair to say there's no detail - time has been put into trying to collect the wide range of opinions and thoughts into something workable. If more detail is required in particular places, happy to add it, but it's hard to respond to "fix all of it".

Johennes: I think for practical reasons we should probably consider this and https://github.com/matrix-org/matrix-doc/pull/4519 in tandem rather than in isolation. In hindsight, one MSC might have sufficed but I think this proposal still helps in fleshing out a joint understanding of registries and help decide whether we want to pursue this path or not.

For added context, some of the core ideas came out of conversations between the VoIP and backend teams at Element and @turt2live thankfully volunteered to codify the thoughts into MSC(s) to help us move things along.

richvdh: > I don't feel it's fair to say there's no detail - time has been put into trying to collect the wide range of opinions and thoughts into something workable.

I appreciate that you've put a bunch of time and effort into this, and it certainly wasn't my intention to belittle that -- apologies if it came across that way.

But it's precisely *because* you've detailed a bunch of different options and opinions that I'm finding it hard to reach a conclusion on this MSC. Effectively, the MSC says "We could do A, or B, or C, or D," and -- particularly without concrete examples of how any of those options might best be used -- I find each of those options hard to picture.

turt2live:
Per SCT internal discussion, we do not believe this is required for Matrix 2.0. It also likely needs more thought, so:

@mscbot fcp cancel

PR #4515, opened 2026-07-22
MSC4515: RTC Transports discovery for widgets

BillCarsonFr:
[Rendered](https://github.com/BillCarsonFr/matrix-doc/blob/valere/rtc/widget_action_rtc_transport/proposals/4515-rtc-transports-widget-action.md)

Implementations:
 - Widget API https://github.com/matrix-org/matrix-widget-api/pull/165
 - matrix-js-sdk https://github.com/matrix-org/matrix-js-sdk/pull/5444
 - rust sdk https://github.com/matrix-org/matrix-rust-sdk/pull/6791
 - Clients:
     - EW https://github.com/element-hq/element-web/pull/34393
     - Gomuks https://github.com/gomuks/gomuks/pull/737 (with https://github.com/matrix-org/matrix-widget-api/pull/165)
 - Widget (using):
     - EC https://github.com/element-hq/element-call/pull/4135

PR #4519, opened 2026-07-28
MSC4519: MatrixRTC Transports Registry

turt2live:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/voip-transports-registry/proposals/4519-rtc-transports-registry.md)

----

SCT Stuff:

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4519#issuecomment-5169305358)

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4519#issuecomment-5169318925)

### DIFF
--- a/proposals/4519-rtc-transports-registry.md
+++ b/proposals/4519-rtc-transports-registry.md
@@ -93,6 +93,8 @@ The endpoint requires authentication to permit different transports to be return
 not in the registry are not permitted. Additional properties/schema are as per the specification for
 the `type`.
 
+If there are no transports available to the user, `rtc_transports` MUST be an empty array.
+
 ### Exception: MSC4195
 
 Because MSC4195 has been used to test the unstable implementation of MSC4143 and has no unstable
--- a/proposals/4519-rtc-transports-registry.md
+++ b/proposals/4519-rtc-transports-registry.md
@@ -87,6 +87,14 @@ is added to the Client-Server API (core spec):
 not in the registry are not permitted. Additional properties/schema are as per the specification for
 the `type`.
 
+### Exception: MSC4195
+
+Because MSC4195 has been used to test the unstable implementation of MSC4143 and has no unstable
+prefixes itself, `livekit` is registered as an unstable type instead of `msc4195.livekit`. When
+MSC4195 becomes accepted, its type is to be upgraded to `stable` with relevant changes to the spec
+link. No implementation changes would be required outside what they'd need to do to support MSC4143
+becoming stable itself.
+
 
 ## Potential issues
 
--- a/proposals/4519-rtc-transports-registry.md
+++ b/proposals/4519-rtc-transports-registry.md
@@ -1,0 +1,135 @@
+# MSC4519: MatrixRTC Transports Registry
+
+[MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) describes the framework for
+Real-Time Communication (RTC) applications to operate over Matrix. The MSC also allows for alternate
+transports to be made available for accessing media streams of RTC application members. See MSC4143
+for definitions and more information.
+
+[MSC4518](https://github.com/matrix-org/matrix-spec-proposals/pull/4518) describes a process to
+support interoperability of non-core Matrix components. The MSC supports a variety of data types
+being stored within a registry, keyed by a string identifier, including chunks of specification.
+See MSC4518 for more details on what registries are and how they intend to operate.
+
+This proposal creates a new registry to hold RTC transports and their specifications, encouraging
+interoperability without requiring those transports to enter the core Matrix specification. Transports
+suitable for the core specification can still land there, though they'll also need placing in the
+registry.
+
+The primary motivation for a registry is to allow transports such as [MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195)
+to land with relative stability, but without placing a dependency on LiveKit into the core spec.
+
+
+## Proposal
+
+A new "MatrixRTC Transports" registry is created with the following information:
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
+added when a new proposal is opened which introduce a transport. They MUST have a `type` of `msc0000.{stable_type}`
+and implementations MUST use the "Unstable Prefix" identifiers from the proposal rather than any
+stable identifiers. For example, MSC4195's `livekit` type would have an unstable registration of
+`msc4195.livekit`, using unstable identifiers. Later, if accepted, `livekit` can be registered as a
+stable `type` with stable identifiers in its specification.
+
+Twelve months after the FCP acceptance of the MSC the unstable type is automatically removed from the
+registry. Early removal is possible if the public federation has adopted the stable type sooner than
+expected - a normal spec PR with normal review requirements is used to propose that removal.
+
+**Note**: The automation above is process automation. Someone will still need to actually open the
+spec PR which adds/removes/changes a transport's registration.
+
+This structure allows for registry changes to be as easy as modifying the specification itself. As
+a worked example:
+
+* MSC4195 gets opened, introducing the `livekit` type.
+* The author opens a spec PR to register the `msc4195.livekit` unstable type, using the MSC URL as
+  the specification reference (specifically the "Unstable Prefix" section/components).
+* MSC4195 goes through normal review, and is eventually proposed for FCP (merge).
+* MSC4195 succeeds at FCP and becomes accepted.
+* A contributor opens a spec PR to register the `livekit` stable type. They also write a new section
+  on the registry's page with MSC4195's contents. That new section's anchor becomes the spec URL.
+
+  **Note**: If the MSC is detailed enough, a link to the rendered Markdown in the spec proposals repo
+  is probably suitable too.
+* Upon that spec PR being merged, MSC4195 also becomes merged (per MSC process).
+* A year later, a contributor removes the `msc4195.livekit` unstable type through a spec PR.
+
+Mentioned in this proposal's introduction, this process does not prohibit a transport from entering
+the core spec. Instead of the spec PR for MSC4195 (for example) being an added section on the registry's
+page, it would be somewhere within the core spec. The registry would then link to that versioned
+section. Transports which intend to enter the core spec MUST declare that in their proposals - the
+default assumption is that transports will not enter the core spec.
+
+### Transport discovery
+
+Some RTC transports might require server-side infrastructure such as SFUs and TURN servers. Clients
+need a mechanism to discover the availability of such infrastructure and any potentially required
+connection details. To enable this, a new authenticated, rate limited, and guest accessible endpoint
+is added to the Client-Server API (core spec):
+
+```json5
+// GET /_matrix/client/v1/rtc/transports
+// 200 OK
+{
+  "rtc_transports": [
+    {
+      "type": "{transport_type}",
+      ... // Further transport-specific properties (if required)
+    }
+  ]
+}
+```
+
+`rtc_transports` is required and MUST be an array of objects with at least a registered `type`. Types
+not in the registry are not permitted. Additional properties/schema are as per the specification for
+the `type`.
+
+
+## Potential issues
+
+See MSC4518 for potential issues with registries.
+
+Transports are not expected to be proposed often, so the regular MSC process should support the
+registry well. Most transports would have tried to enter the core specification anyway, which
+would generate MSCs. Per MSC4518, the team which reviews additions/changes/removals to the registry
+can change at any time through another MSC.
+
+
+## Alternatives
+
+See MSC4518 for alternatives to registries. In the case of MSC4195 (LiveKit Transport), the alternative
+is effectively putting the dependency into the core spec for all implementations.
+
+`.well-known` was considered for the transport discovery endpoint, but as a traditionally unauthenticated
+endpoint it doesn't fit infrastructure requirements for some transports. Well-Known URLs are often not
+served by the homeserver (which would have transport information), but rather a different web server,
+making populating the resource much harder. Authentication further allows for user-specific transports.
+
+
+## Security considerations
+
+Individual transports are expected to have their own security considerations. The presence of a
+registry does not carry inherent security challenges, per MSC4518.
+
+
+## Unstable prefix
+
+Registries do not have an effective unstable prefix. To assist in editorial ease however, this
+proposal's registry can be implemented as an added section in the [appendices](https://spec.matrix.org/v1.19/appendices/).
+Later, when either more proposals use registries or the infrastructure is easier to use, the section
+would be moved out.
+
+The `GET /_matrix/client/v1/rtc/transports` endpoint should be accessed as `GET /_matrix/client/unstable/org.matrix.msc4143/rtc/transports`
+while this proposal is considered unstable. Note that the namespace is from MSC4143 - this is to
+reflect the endpoint being lifted out of that proposal and into this one.
+
+
+## Dependencies
+
+This proposal depends on MSC4518 (Registries).
+
+This proposal SHOULD NOT be accepted without MSC4143 (MatrixRTC) being accepted
+concurrently. Note that MSC4143 also depends on this proposal.
--- a/proposals/4519-rtc-transports-registry.md
+++ b/proposals/4519-rtc-transports-registry.md
@@ -87,6 +87,8 @@ is added to the Client-Server API (core spec):
 }
 ```
 
+The endpoint requires authentication to permit different transports to be returned to different users.
+
 `rtc_transports` is required and MUST be an array of objects with at least a registered `type`. Types
 not in the registry are not permitted. Additional properties/schema are as per the specification for
 the `type`.
--- a/proposals/4519-rtc-transports-registry.md
+++ b/proposals/4519-rtc-transports-registry.md
@@ -28,11 +28,15 @@ A new "MatrixRTC Transports" registry is created with the following information:
 * A link to the specification for that transport, per MSC4143.
 
 `stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
-added when a new proposal is opened which introduce a transport. They MUST have a `type` of `msc0000.{stable_type}`
-and implementations MUST use the "Unstable Prefix" identifiers from the proposal rather than any
-stable identifiers. For example, MSC4195's `livekit` type would have an unstable registration of
-`msc4195.livekit`, using unstable identifiers. Later, if accepted, `livekit` can be registered as a
-stable `type` with stable identifiers in its specification.
+added when a new proposal is opened which introduce a transport. They by default have a `type` of `msc0000.{stable_type}`,
+but MAY be extended upon per the MSC's "Unstable Prefix" section provided they maintain the `msc0000`
+prefix. For example, if a proposal changes substantially after the registration was made, the proposal 
+might suggest using `msc0000.livekit.v2`.
+
+Like with regular unstable implementation, `unstable` transport implementations MUST use the "Unstable
+Prefix" section of the relevant MSC. For example, if the transport introduces `/new_endpoint`, then
+implementations would use `/unstable/org.example.msc0000/new_endpoint` in the `msc0000.whatever`
+transport. `/v1/new_endpoint` could only be used in the `whatever` transport (the `stable` type).
 
 Twelve months after the FCP acceptance of the MSC the unstable type is automatically removed from the
 registry. Early removal is possible if the public federation has adopted the stable type sooner than


proposals/4519-rtc-transports-registry.md
turt2live, line 1: Implementation requirements:

* [x] Transports endpoint: client
* [x] Transports endpoint: server

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,143 @@
...
+
+The primary motivation for a registry is to allow transports such as [MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195)
+to land with relative stability, but without placing a dependency on LiveKit into the core spec.
+
+
+## Proposal
+
+A new "MatrixRTC Transports" registry is created with the following information:
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
```
robintown, line 26: Currently this proposal says, "Types not in the registry are not permitted," but I hope that the spec could eventually make room for unregistered, proprietary transports which might not be suitable for making public. This seems easiest to manage if we recommend the Common Namespaced Identifier Grammar.

Previously we were recommending CNIG without the namespacing requirements so that we could still have an un-namespaced `livekit` as in this MSC, but I & others on the Element VoIP team would be interested in your judgement on whether it would be preferable to *keep* the namespacing requirement, if we say `type` uses CNIG.

turt2live: Explicitly supporting proprietary transports doesn't really fit the [Guiding Principles](https://spec.matrix.org/proposals/#guiding-principles) in my opinion. We can support namespacing, but I would still probably keep the limitation that only registered types can be used, at least until Matrix itself has a transport type that's part of the core specification. 

Without a core transport type in the spec we run the risk of only having proprietary transports, which defeats the purpose of this registry.

A later MSC can remove the condition easily enough without it being a breaking change.

turt2live: (leaving open for further opinions, but mentally considering resolved for now)

turt2live:

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Dependencies

turt2live:
@mscbot fcp merge
@mscbot concern Dependency - Registries
@mscbot concern Coordinate FCP with MatrixRTC

proposals/4519-rtc-transports-registry.md
turt2live: This is effectively implemented by Synapse via MatrixRTC.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,143 @@
...
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
+added when a new proposal is opened which introduce a transport. They MUST have a `type` of `msc0000.{stable_type}`
+and implementations MUST use the "Unstable Prefix" identifiers from the proposal rather than any
+stable identifiers. For example, MSC4195's `livekit` type would have an unstable registration of
+`msc4195.livekit`, using unstable identifiers. Later, if accepted, `livekit` can be registered as a
```
richvdh, line 34: I don't understand what "implementations MUST use the "Unstable Prefix" identifiers" means here: it seems to conflict with the requirement of using `msc0000.{stable_type}`.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,143 @@
...
+## Proposal
+
+A new "MatrixRTC Transports" registry is created with the following information:
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
+added when a new proposal is opened which introduce a transport. They MUST have a `type` of `msc0000.{stable_type}`
```
richvdh, line 31: > They MUST have a `type` of `msc0000.{stable_type}`

This means that, if an MSC evolves such that we need a different transport identifier, we'd have to open a new MSC? And conversely, if one MSC gets abandoned, but someone else decides to further the work (via a new MSC), they'd have to use a new transport identifier?

The MUST just seems a bit constraining.

proposals/4519-rtc-transports-registry.md
richvdh: would using `m.livekit` really be that bad?

proposals/4519-rtc-transports-registry.md
turt2live: It's saying that if your MSC adds `/some_endpoint`, then when `msc0000.stable` is used you're also using `/unstable/some_endpoint` rather than `/v1/some_endpoint`.

proposals/4519-rtc-transports-registry.md
turt2live: `m.livekit` is fine in isolation, but opening up the registry to have unregistered custom types I feel would be a mistake.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,143 @@
...
+the core spec. Instead of the spec PR for MSC4195 (for example) being an added section on the registry's
+page, it would be somewhere within the core spec. The registry would then link to that versioned
+section. Transports which intend to enter the core spec MUST declare that in their proposals - the
+default assumption is that transports will not enter the core spec.
+
+### Transport discovery
+
+Some RTC transports might require server-side infrastructure such as SFUs and TURN servers. Clients
+need a mechanism to discover the availability of such infrastructure and any potentially required
+connection details. To enable this, a new authenticated, rate limited, and guest accessible endpoint
```
timedoutuk, line 70: Why is authentication required? I currently can't see a reason for this

timedoutuk, line 70: Why is this endpoint rate-limited? what is a suitable rate-limit?

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,143 @@
...
+  "rtc_transports": [
+    {
+      "type": "{transport_type}",
+      ... // Further transport-specific properties (if required)
+    }
+  ]
+}
+```
+
+`rtc_transports` is required and MUST be an array of objects with at least a registered `type`. Types
```
timedoutuk, line 86: What about when the server does not have any RTC transports? Should 404 M_NOT_FOUND be returned here? or an empty array of `rtc_transports`?

proposals/4519-rtc-transports-registry.md
tulir: Servers may want to give different transports to different users, plus unauthenticated users never need that info, so requiring authentication makes sense.

turt2live: Because all endpoints should support rate limiting. Implementations are free to choose their own rate limits, per usual.

tulir: That sounds like the "rate limited" flag is effectively completely meaningless. I guess we should make an MSC to replace it with something more useful, like a recommended rate limit? (or if the spec doesn't want to define suitable limits, then remove the flag entirely)

turt2live: That's effectively what https://github.com/matrix-org/matrix-spec-proposals/pull/4513 does.

In any case, it's not a concern for this MSC or the hundreds before it - the endpoint is rate limited.

timedoutuk: I suppose this is more of a broad spec problem (rate-limiting is a yes/no suggestion rather than some standard rate suggestion), fair enough

proposals/4519-rtc-transports-registry.md
turt2live: 34f78fb1b68bf594b78d7d4ac50a39bda520e227

proposals/4519-rtc-transports-registry.md
turt2live: f0f4bf79c0e4b95fd20dc42885bbb51addf7c13f

proposals/4519-rtc-transports-registry.md
turt2live: df0c115c3a6e7ad71ae7346e7a9562604096240d

proposals/4519-rtc-transports-registry.md
turt2live: df0c115c3a6e7ad71ae7346e7a9562604096240d

proposals/4519-rtc-transports-registry.md
Johennes: - Server: https://github.com/element-hq/synapse/pull/18967
- JS SDK: https://github.com/matrix-org/matrix-js-sdk/pull/5104
- Element Call: [`RtcTransportAutoDiscovery.ts`](https://github.com/element-hq/element-call/blob/e20abb19fbdb2b6fd729e02ab66b53c70fec48d0/src/state/CallViewModel/localMember/RtcTransportAutoDiscovery.ts)
- Element Web: [`CallStore.fetchTransports`](https://github.com/element-hq/element-web/blob/5bcfe5aee656937b068b867230802e2eda2f7c0b/apps/web/src/stores/CallStore.ts#L60)

> [!NOTE]
> When ran in widget mode, Element Call cannot yet use the `/transports` endpoint because the widget API doesn't expose access to it. This is covered by https://github.com/matrix-org/matrix-spec-proposals/pull/4515.

proposals/4519-rtc-transports-registry.md
Johennes: I think we should switch to `m.livekit` because we already use the namespacing elsewhere in MatrixRTC (e.g. `m.per_member`) and we would probably want to allow extensibility at some point.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
+added when a new proposal is opened which introduce a transport. They by default have a `type` of `msc0000.{stable_type}`,
+but MAY be extended upon per the MSC's "Unstable Prefix" section provided they maintain the `msc0000`
+prefix. For example, if a proposal changes substantially after the registration was made, the proposal 
+might suggest using `msc0000.livekit.v2`.
```
Johennes, line 34: Is there a benefit to customizing the prefixing rules like this other than making identifiers a little shorter? I would have expected that we'd just use the [existing prefixing rules](https://spec.matrix.org/proposals/#unstable-endpoints-features-and-vendor-prefixes).

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+**Note**: The automation above is process automation. Someone will still need to actually open the
+spec PR which adds/removes/changes a transport's registration.
+
+This structure allows for registry changes to be as easy as modifying the specification itself. As
+a worked example:
+
+* MSC4195 gets opened, introducing the `livekit` type.
+* The author opens a spec PR to register the `msc4195.livekit` unstable type, using the MSC URL as
+  the specification reference (specifically the "Unstable Prefix" section/components).
```
Johennes, line 53: The unstable transport is registered manually via a spec PR here but further up it says

> [...] `unstable` transports are automatically added when a new proposal is opened which introduce a transport.

Having the registry change automatically as MSCs are opened worries me a little. I also wonder if we actually need to register unstable types in the registry at all? It feels like while something is not stable, it shouldn't land in the registry just like normal MSCs don't land in the spec either without being accepted?

That would also avoid the clean-up step 12 months after the transport became stable.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+* MSC4195 goes through normal review, and is eventually proposed for FCP (merge).
+* MSC4195 succeeds at FCP and becomes accepted.
+* A contributor opens a spec PR to register the `livekit` stable type. They also write a new section
+  on the registry's page with MSC4195's contents. That new section's anchor becomes the spec URL.
+
+  **Note**: If the MSC is detailed enough, a link to the rendered Markdown in the spec proposals repo
+  is probably suitable too.
```
Johennes, line 60: Linking by SHA would be ok but linking by branch name (which is normally done for the \[Rendered\] links) would allow for further changes to the registry-spec without new MSCs. I think for stable transports, I'd prefer the registry to just contain the required spec text – just like the normal spec.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+If there are no transports available to the user, `rtc_transports` MUST be an empty array.
+
+### Exception: MSC4195
+
+Because MSC4195 has been used to test the unstable implementation of MSC4143 and has no unstable
+prefixes itself, `livekit` is registered as an unstable type instead of `msc4195.livekit`. When
+MSC4195 becomes accepted, its type is to be upgraded to `stable` with relevant changes to the spec
+link. No implementation changes would be required outside what they'd need to do to support MSC4143
+becoming stable itself.
```
Johennes, line 104: Should this rather move into #4195? Otherwise both MSCs mutually depend on each other?

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+## Security considerations
+
+Individual transports are expected to have their own security considerations. The presence of a
+registry does not carry inherent security challenges, per MSC4518.
+
+
+## Unstable prefix
+
+Registries do not have an effective unstable prefix. To assist in editorial ease however, this
+proposal's registry can be implemented as an added section in the [appendices](https://spec.matrix.org/v1.19/appendices/).
```
Johennes, line 137: The "can" makes it seem like we'll leave the concrete decision open until spec PR time?

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+being stored within a registry, keyed by a string identifier, including chunks of specification.
+See MSC4518 for more details on what registries are and how they intend to operate.
+
+This proposal creates a new registry to hold RTC transports and their specifications, encouraging
+interoperability without requiring those transports to enter the core Matrix specification. Transports
+suitable for the core specification can still land there, though they'll also need placing in the
+registry.
+
+The primary motivation for a registry is to allow transports such as [MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195)
+to land with relative stability, but without placing a dependency on LiveKit into the core spec.
```
clokep, line 19: The spec has some optional modules, this sentence along with the potential issue of there not being many RTC transports makes this feel like overkill to have this be a registry?

clokep (commented):
@mscbot concern Unclear why this needs a registry instead of the normal spec process

proposals/4519-rtc-transports-registry.md
Johennes: Yeah, I think we need to improve the motivation a little. The concrete advantages of the registry I can currently think of are:

- Releasing transports is not dependent on the spec release cycle
  This proposal doesn't currently clarify whether the registry will be released together with the spec or not though. See also https://github.com/matrix-org/matrix-spec-proposals/pull/4519/changes#r3813858452.
- Transports can more easily be removed from the registry than from the main spec
  Homeservers declare support for entire spec versions. Even if a feature is removed, servers often retain their implementations to be able to declare support for older spec versions. This came up as a concern when discussing the integration of the LiveKit transport with the Synapse team.

Having said that, I acknowledge the concern that we may not have a lot of transports, however. The transports the VoIP team at Element can foresee right now are:

- LiveKit ([MSC4195](https://github.com/matrix-org/matrix-spec-proposals/pull/4195))
  This is the only one we have today.
- A cascading SFU transport where users connect to their local SFU only and SFUs federate
  @fkwp has made some explorations on this. It'll require a dedicated SFU implementation. If proven worthwhile, this would be an alternative (if not a full replacement) for the LiveKit transport. Currently, it seems likely that we'll work on this after Matrix 2.0.
- A full-mesh transport where members connect to each other directly as in [legacy VoIP](https://spec.matrix.org/v1.18/client-server-api/#voice-over-ip)
  This would be an alternative for small calls without LiveKit. Currently, it does _not_ seem overly likely that we'll work on this in the near future unless somebody sponsors it.

It's difficult to predict what transports other people might come up with though.

clokep: > Transports can more easily be removed from the registry than from the main spec
Homeservers declare support for entire spec versions. 

I don't buy this. If we add a capability for them then just rip it out and mark it as unsupported via the capability.

clokep:
@mscbot concern Unclear why this needs a registry instead of the normal spec process

proposals/4519-rtc-transports-registry.md
turt2live: @erikjohnston is the main proponent for not putting LiveKit into the spec, so I'd appreciate talking points from him that can be included in this MSC.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
+# MSC4519: MatrixRTC Transports Registry
+
+[MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143) describes the framework for
+Real-Time Communication (RTC) applications to operate over Matrix. The MSC also allows for alternate
```
dbkr, line 4: Not really 'alternate': it needs a transport to work at all.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+## Proposal
+
+A new "MatrixRTC Transports" registry is created with the following information:
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
+added when a new proposal is opened which introduce a transport. They by default have a `type` of `msc0000.{stable_type}`,
```
dbkr, line 31: Presumably the '0000' is a placeholder for the msc's number, in which case I'd suggest 'xxxx'?

proposals/4519-rtc-transports-registry.md
dbkr: …which does somewhat conflict with the registry msc's requirement that only optional functionality be in registries… I suppose it's optional in that you don't have to implement *all* of them, even if you have to implement *one*?

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+`.well-known` was considered for the transport discovery endpoint, but as a traditionally unauthenticated
+endpoint it doesn't fit infrastructure requirements for some transports. Well-Known URLs are often not
+served by the homeserver (which would have transport information), but rather a different web server,
+making populating the resource much harder. Authentication further allows for user-specific transports.
+
+
+## Security considerations
+
+Individual transports are expected to have their own security considerations. The presence of a
+registry does not carry inherent security challenges, per MSC4518.
```
dbkr, line 131: I would argue that they very much do since they offer a way for protocol to become spec with little or no review.

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Proposal
+
+A new "MatrixRTC Transports" registry is created with the following information:
+
+* A transport `type`. MUST be an [Opaque Identifier](https://spec.matrix.org/v1.19/appendices/#opaque-identifiers).
+* Stability (`stable` or `unstable`).
+* A link to the specification for that transport, per MSC4143.
+
+`stable` transports are added/changed/removed through normal MSCs. `unstable` transports are automatically
```
dbkr, line 30: I have to admit I'm somewhat struggling with what the MSC really achieves since clients could implement an open MSC today, so this feels like just a way to bypass the review process, but then isn't that process there for a reason?

proposals/4519-rtc-transports-registry.md
turt2live: The review requirements are the same as any other MSC, so I'm not sure that holds true. It's also not a Security Consideration, but would rather be a Potential Issue.

proposals/4519-rtc-transports-registry.md
turt2live: Registries are outside of the spec, so they need their own mechanisms to track unstable prefixes.

proposals/4519-rtc-transports-registry.md
turt2live: I'm not following the concern here, sorry - linking to a branch name to get live updates is the desired behaviour.

proposals/4519-rtc-transports-registry.md
turt2live: This MSC is better positioned to state the namespacing requirements. Stating them in this way doesn't create a dependency on MSC4195 - it's just saying "if/when MSC4195 enters the registry, it uses this type".

proposals/4519-rtc-transports-registry.md
turt2live: correct.

proposals/4519-rtc-transports-registry.md
turt2live: "alternate" is in contrast to the specified transport itself, if any. We don't currently have that, so this might be confusing.

proposals/4519-rtc-transports-registry.md
turt2live: I use `0000` as placeholder on other MSCs - this feels like editorial preference.

proposals/4519-rtc-transports-registry.md
turt2live: With the idea that registries exist outside of the spec, we need to define the process by which the registry is updated/used. This includes defining how unstable types are reflected in the proposed endpoint and how experimental implementations are worked on. Without automatic registration, we'd need 2 MSCs before implementations could start using a new transport: one to define the transport (already required) and another to add that to the registry.

turt2live:
For documentation: https://github.com/matrix-org/matrix-spec-proposals/pull/4518 is no longer part of Matrix 2.0, but this MSC remains in scope (currently). Dependencies on MSC4518 will be removed in a later editorial pass, pending the outcome of further conversations around registries as a concept.

proposals/4519-rtc-transports-registry.md
Johennes: My concern is that the registry itself should be stable. It shouldn't change unexpectedly without a release. Anything that aims to enter the registry requires an MSC. While the MSC is unstable / unmerged, the MSC alone is sufficient to coordinate any early implementations. We don't need to track unstable transports in the registry – just like we don't track unstable features in the Matrix spec itself.

proposals/4519-rtc-transports-registry.md
dbkr: If a transport is added to the registry as soon as an MSC is created for it, doesn't that mean it's fair game to start using it as soon as you've created an MSC, regardless of whether anyone has looked at the MSC?

proposals/4519-rtc-transports-registry.md
dbkr: Well my point is that if there was no registry, you could just create the MSC and then start using it unstable-prefixed as we do now?

proposals/4519-rtc-transports-registry.md
```diff
@@ -0,0 +1,151 @@
...
+Some RTC transports might require server-side infrastructure such as SFUs and TURN servers. Clients
+need a mechanism to discover the availability of such infrastructure and any potentially required
+connection details. To enable this, a new authenticated, rate limited, and guest accessible endpoint
+is added to the Client-Server API (core spec):
+
+```json5
+// GET /_matrix/client/v1/rtc/transports
+// 200 OK
+{
+  "rtc_transports": [
```
Johennes, line 81: We should rename this to just `transports` for parity with the `m.rtc.member` events from #4143.

proposals/4519-rtc-transports-registry.md
Johennes: #4195 has switch to `m.livekit` now.

turt2live:
This MSC is being removed from the Matrix 2.0 scope, so is being removed from FCP too.

@Johennes - the `/transports` endpoint will need to be returned to the original MSC it came from. This MSC can be left to bitrot.

@mscbot fcp cancel

See [this post in the SCT Office](https://matrix.to/#/!0KNSXYXB_2xtEUkQ9MGBRy5oNIOfAKoq2uIqPZCJbI8/$u69GUT68iYVXwF0cO9s_gskaikQjF_OJp-KrC_qcxsA?via=element.io&via=matrix.org&via=federated.nexus) for more information.

PR #4520, opened 2026-07-28
Move transport types out of MSC4143 and into MSC4519

turt2live:
See title.


robintown (approved):
Thanks, this looks good considering the current text of MSC4519. (Though I am noting that in case https://github.com/matrix-org/matrix-spec-proposals/pull/4519#discussion_r3682414975 leads anywhere, the text here implying that `type`s are always registered should perhaps change.)

PR #4521, opened 2026-07-28
MSC4521: Adaptive set reconciliation

gamesguru:
### MSC4521: Adaptive set reconciliation

*NOTE: Current revision (bb76dade3) contains minor mistakes, such as the all zero 128-bit accumulator edge case defined wrong. It is generally in good shape, but if your time is limited I advise WAITING until further commits are added and I make further efforts to circulate or socialize it in the spec room. The summary/description below is fully up to date, 8/20/26.*

---

MSC enabling arbitrary over-the-wire set reconciliation approaching the Shannon limit[^1] for communication.

This proposal is an alternative to but shares most of the same use-cases of a previously discussed (but not formally documented) bloom filter and invertible bloom table lookup approach, which may still have merits and perform better at "medium" set differences (very approximately defined as between 1,000 and 1 million elements or events in one but not both sets). This proposal accordingly leaves open that possibility and allows the establishment and standardization of complementary and alternative reconciliation profiles, making drop in replacement at least less painful. It aims to promote, for example, the rapid ~~eventual~~ consistency that MSC4242 State DAGs also seeks[^2], by achieving fast synchronization of state sets and DAGs (either at the `/send` txn level, or through a dedicated endpoint).

With applications to both federated and C2S endpoints, costing only 50-200 ms network latency for typical divergence, requiring minimal compute time, and adding only ~2 KB response overhead—this MSC allows fast over-the-wire set comparisons and reconciliations, _without_ exchanging the full set or lists of IDs.

So, you can compare two sets of events or rooms or `(state_key, event_type, event_id)` tuples over the network between federations or C2S and you don't have to send the full list. Requires some library/SDK usage, persisting a local sketch, and other implementation details outlined in the document.

By the use of 64-bit Galois fields and error correcting BCH-codes, the MSC allows finding the exact symmetric difference between two sets, A and B, and returning the exact list of missing or extra IDs with near theoretic optimal network/bandwidth complexity.

[Rendered](https://github.com/gamesguru/matrix-spec-proposals/blob/guru/4521-algebraic-set-reconciliation/proposals/4521-algebraic-set-reconciliation.md).

### Proposed use cases

- Fast state diffing over `/state_ids` endpoint.
- Required for _real-time_ guarantees in MSC4242
- Full room missed PDU reconciliation
- C2S checks (verify list of rooms, room states, and timeline segments)
- Media synchronization (currently we only query the origin, but combined with MSC2846 or MSC3468 this could enable broad, fast syncs between "heavy" nodes)

### Implementations

- Library/Rust SDK: https://github.com/gamesguru/rezzy/blob/master/src/reconcile/algebraic.rs
- Homeserver, WIP: _feat: room reconciliation surface (MSC4521 primitives) · gamesguru/congruent_ <https://github.com/gamesguru/congruent/pull/87>

Sign-off of this MSC **deferred** until stable and efficient implementations are proven. *Dated: 8/20/26*.

[^1]:
    _Explained: The Shannon limit - MIT News - Massachusetts Institute of Technology_
    https://news.mit.edu/2010/explained-shannon-0115

[^2]:
    [matrix-spec-proposals/proposals/4242-state-dags.md, commit af93906d3 line 51](https://github.com/matrix-org/matrix-spec-proposals/blob/af93906d335763e6723de5a63caeefbdf36beb89/proposals/4242-state-dags.md?plain=1#L51)

### DIFF
--- a/proposals/4521-algebraic-set-reconciliation.md
+++ b/proposals/4521-algebraic-set-reconciliation.md
@@ -1,0 +1,756 @@
+# MSC4521: Adaptive Set Reconciliation via PinSketch
+
+Several federation mechanisms need to know whether two servers contain the same
+set of identifiers. With a lot of work, this MSC lets them compute the exact
+symmetric difference between large populations, without any probabilistic
+errors. This MSC helps ensure network synchronization.
+
+Some consumers need that over a room's known event or resolved state set; others
+use it to synchronize key IDs between notaries, or to reconcile ephemeral and
+other identifier populations. This MSC defines the primitive once, as a named
+digest profile, so that consumers can share the field, hash derivation, wire
+encoding, and decoding contract instead of rebuilding them from scratch.
+
+`algebraic_v1` couples a strata estimator, extraction sketch, and 128-bit
+accumulator into one ladder.
+
+The profile targets differences up to 4,096 elements per round in populations up
+to $10^7$, keeps the initial depth-0 sketch under 256 B, and keeps a fully
+saturated round under 32 KiB, excluding object payloads. Decoding a capacity-`k`
+node costs $O(k^2 \log k)$. A difference of size $d$ spread over $n$ nodes
+therefore costs $O\!\left(\frac{d^2}{n}\log\frac{d}{n}\right)$. Quadratic
+complexity means that invertible bloom filters will outscale this MSC
+asymptotically, but at smaller deltas, this MSC wins (nearly all typical use
+cases). Larger differences are a frame problem, not a reconciliation problem
+(see [Scalability](#scalability)); the baseline 20-round, 4096-capacity sequence
+reaches about 82,000 differing elements under the default ceiling parameters.
+More capacity extends a round; it does not restart it.
+
+## Scope
+
+This profile defines identifier derivation, the 64-bit field and `libminisketch`
+compatibility contract, the level-0 accumulator, the syndrome sketch and its
+capacity bounds, dynamic tree extraction and the strata estimator, the
+decode-and-verify contract, capacity budgets, and the resident structure.
+
+This profile does **not** define frames, negotiation, scheduling, endpoints, or
+authorization; those belong to the consuming MSC. Consumers MUST still verify
+that both sides digest the same population before comparing them.
+
+## Element derivation
+
+The profile operates over a set `S` of opaque elements. Each consumer MUST map
+every element to a canonical 32-byte digest before applying this profile.
+Consumers define what the elements mean; the kernel treats them as an opaque set
+and does not interpret their content.
+
+Let `D(e)` be the consumer-defined 32-byte digest for element `e`.
+
+`libminisketch` requires non-zero inputs over $\mathbb{F}_{2^{64}}$.
+Implementations derive $h_{64}(e)$ and $h_{128}(e)$ from `D(e)` using network
+byte order (big-endian):
+
+- **$h_{64}(e)$ (64-bit field element):** Scan `D(e)` in four 8-byte big-endian
+  chunks. $h_{64}(e)$ MUST be the first non-zero chunk interpreted as an
+  unsigned 64-bit integer, or `1` if all four chunks are zero.
+- **$h_{128}(e)$ (128-bit accumulator element):** Take the first 16 bytes of
+  `D(e)` as an unsigned 128-bit big-endian integer. Zero is permitted here; the
+  accumulator is a plain XOR sum, so it does not need the $h_{64}$ non-zero
+  fallback.
+
+### Matrix event-ID binding
+
+For event sets, `D(e)` is derived based on the room version:
+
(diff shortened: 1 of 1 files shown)


proposals/4521-algebraic-set-reconciliation.md
turt2live, line 1: Implementation requirements unknown: MSC unclear.

proposals/4521-algebraic-set-reconciliation.md
```diff
@@ -0,0 +1,756 @@
+# MSC4521: Adaptive Set Reconciliation via PinSketch
```
Half-Shot, line 1: As much as I can't claim to be a great mathematician, I think this opening needs be much more high level on what the MSC is aiming to solve. I roughly understand the *mechanism* of what's being proposed from the intro, but not what problem it's solving in Matrix.

Please try and ground the proposal in some use cases, as the document is very long and I suspect most will struggle to understand what you are proposing here.

KitsuneRal, line 1: This text desperately needs de-slopping and better structure. Aside from "Possible consumers" (buried somewhere at the very end), there's not much discussion into the "why" of this MSC. No exploration of the current state; extremely limited practical application of the concept to Matrix; normative sections intermingled with original ones. Since this is also, apparently, generated by AI, I'm afraid I can't trust the line of reasoning; AI-generated text may contain subtle errors, and the onus is on the MSC author to prove this particular text doesn't have them.

proposals/4521-algebraic-set-reconciliation.md
gamesguru: This MSC targets an SDK or toolkit, like `ruma` or the `matrix-rust-sdk`. It is not meant for homeservers to directly roll their own implementations from scratch (they _can_, however). Hence the "low level" or "primitive" presentation.

It aims to provide Matrix with a reliable, fast synchronization method. Whether that be notaries syncing key IDs, servers booting up after extended down time looking to catch up missed event IDs, or missed read receipts (EDUs) that server owners want to repair or fill in by "gossiping" with other nodes/servers.

This MSC also lays the groundwork to optimize the traditionally sluggish `GET /state_ids` with a nearly instantaneous symmetric difference (in most cases — even in rooms with 1000 or 10,000 people — servers rarely disagree on resolved state by more than 50 or 100 event entries... but requesting the full list of state IDs often takes 60 seconds or more).

---

I'm open to guidance on the proper procedure for MSCs that are foundational building blocks for SDKs more than directly under the purview of Synapse or a homeserver.

This problem has been discussed previously in the Spec channel, at least one early discussion can be found searching "bloom [filters]" or "reconciliation." Tracing this reveals the progression of my thought and some motivations behind it.

I'm planning to post the PDU reconciliation MSC (and possibly one to optimize `GET /state_ids` and `GET /state`), which should hopefully bring clarity to the motivation behind this MSC and allow for tangible discussion about its real world uses and any potential complications or edge cases.

proposals/4521-algebraic-set-reconciliation.md
gamesguru: Hi, thanks. I can update the opening to ground the proposal in some real-world use cases (e.g., optimizing `GET /state_ids`, or PDU reconciliation).

I'll post a consumer MSC to strengthen the connections and justification for it. I'll also restructure the math-heavy portions to make the ideas more accessible so more people can review and relate to the document

gamesguru: Hi, please see the comments above for some context, as they also had similar concerns. I'll add some reasons for what motivates the MSC and the use cases it can solve or improve, as well as linking to consumer MSCs and making the document as a whole easier on the eyes.

Regarding AI, there is little doubt it is involved in shaping the prose of my draft, but the equations, ideas, and intuitions were mine from the start. The design is uniquely mine; there's not a model on Earth producing work as idiosyncratic and peculiar as mine.

That said, you shouldn't have to take my word for it. The kernel is standard PinSketch / BCH syndrome decoding; the math is independently checkable against the literature and the reproducible test vectors already in the draft. I certainly expect aspects will need to be questioned and rigorously defended. Rather than me attempting a reproof of the entire document from scratch, could you flag any specific areas where you'd like to see improved clarity or rigor? Thanks.

lveneris:
I am familiar with many of the mathematical principles here, and with great difficulty, I have waded through this MSC. You will notice that you are lacking something many of your citations have: justification. Other people have already given their takes on this, but my best advice is to state plainly **which** parts of the spec you are changing and **why** you are changing them before you launch into technicalities.

As this is intended to be a new primitive (although, it seems to be a whole series of primitives in a rather baggy trench coat), I would suggest giving concrete examples of what the primitive would be useful to, how it compares to any existing work, and more practical references than other MSCs that haven't left the oven yet. Said plainly, this proposal currently reads like a suggestion on how certain endpoints that operate on identifier sets might work in implementation, rather than any real change to how entities in the ecosystem talk to each other. Fixing this meaningfully will require an almost complete redraft of what you have here.

I wish you the best in this effort.

PR #4522, opened 2026-07-28
MSC4522: Username colors

nushea:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/106542900bf5862ad5f84dcfc78e8e047ff08939/proposals/4522-username-colors.md)

Implementations:
 - Setting:
   - Sable https://github.com/SableClient/Sable/pull/1526
 - Rendering:
   - Sable https://github.com/SableClient/Sable/pull/1526

Signed-off by: Shea <nu@she-a.eu>

### DIFF
--- a/proposals/4522-username-colors.md
+++ b/proposals/4522-username-colors.md
@@ -20,13 +20,9 @@ For standardization:
 If a custom color is wished to be set for an ID of a PMP it may be set through a `m.colors` field within the 
 `m.per_message_profile` of the message
 
-  The `m.colors` and `color` fields represent arrays of objects for the color. Each object of the array
-should contain the color that is to be displayed and may contain a `background` that it is meant to 
-represent the ideal background the `color` should be displayed over. The background SHOULD NOT be 
-displayed but should only serve as a way for the client to determine which color best fits the background 
-it is displayed against. 
-  The `m.colors` array and color arrays MUST have at most one object that has no background color that SHOULD
-serve as a fallback.
+  The `m.colors` and `color` fields represent an object that contains the colors that may be choosen depending
+on the active theme of the client. The light_color represents the color that is lighter, and dark_color the color that is darker,
+so a light_color should ideally be used when the user has a dark theme and vice versa.
 
 The field within the extended profile can be for example:
 ```json
@@ -35,12 +31,10 @@ The field within the extended profile can be for example:
   "displayname": "Shea Butter",
   "m.banner_url": "mxc://matrix.org/example_banner",
   "m.tz": "Europe/Troll",
-  "m.colors": [
-    { "color": "#ffd9f5", "background": "#000000" },
-    { "color": "#440000", "background": "#fff" },
-    { "color": "#f0f", "background": "#888" },
-    { "color": "#ff0000" }
-  ]
+  "m.colors": {
+    "light_color": "#ffd9f5",
+    "dark_color": "#440000" 
+  }
 }
 ```
 
@@ -50,9 +44,10 @@ The room state `m.room.member` may be updated to for example:
   "membership": "join",
   "displayname": "User",
   "avatar_url": "mxc://example.com/code",
-  "m.colors": [
-    {"color": "#ff0", "background": "000"}
-  ]
+  "m.colors": {
+    "light_color": "#f00",
+    "dark_color": "#400" 
+  }
 }
 ```
 
@@ -63,17 +58,17 @@ The room state `m.powerlevels_style` represents a potential override, and an arr
     "powerlevels": [
       {
         "powerlevel": 51,
-        "colors": [
-          { "color": "#00f", "background": "#818181" },
-          { "color": "#88f", "background": "#808080" }
-        ]
+        "colors": {
+          "light_color": "#f00",
+          "dark_color": "#400" 
+        }
       },
       {
         "powerlevel": 1,
-        "colors": [
-          { "color": "#000", "background": "#ffffff" },
-          { "color": "#fff", "background": "#000000" }
-        ]
+        "colors": {
+          "light_color": "#f8f",
+          "dark_color": "#404" 
+        }
       }
     ]
   },
@@ -86,10 +81,10 @@ The `m.colors` for PMPs may look for example:
   "displayname": "Nough",
   "avatar_url": "mxc://example.com/example_url",
   "has_fallback": true,
-  "m.colors": [
-    { "color": "#f0f", "background": "#818181" },
-    { "color": "#88f", "background": "#808080" }
-  ]
+  "m.colors": {
+    "light_color": "#82F2A3",
+    "dark_color": "#11403A" 
+  }
 },
 ```
 
@@ -97,26 +92,26 @@ The `m.colors` for PMPs may look for example:
 to the other colors in order to reduce ambiguity between the moderators of a community and its participants,
 but if it is missing clients should consider that as being set to false.
 
-  Every `color` and `background` field must be a hex color for consistency.
+  The color fields must be a hex color for consistency.
 
-  The order of the color values should be: PMP color, per-room color, 
-account profile color, powerlevel color, default fallback that a client SHOULD have. This order may be
-overridden for powerlevels to take precedence above every other color however.
+  The order of the color values should be: PMP color, per-room color, account profile color, 
+powerlevel color, default fallback that a client SHOULD have. This order may be overridden 
+for powerlevels to take precedence above every other color.
 
 ## Potential issues
 
 A user may choose to set their per-room or per account color to match one of a modertator of a room, but this
 may be mitigated by setting `override_other_colors` to true if needed.
 
-A user may choose to set the color and background within one item to the same value but this may be mitigated
-by clients by removing these values from the pool of potential colors to choose from.
+A user may choose to set the color that has poor visibility to the background but that may be mitigated by
+either choosing the color for the opposite theme if it fits better or refusing both and going one step lower
+whenever that is detected (ie so it would ignore the PMP color and use the per-room color)
 
 ## Alternatives
 
-One alternative is setting only one hex value for the color but that was dismissed as that would force clients
-to either fully dismiss the color option of an individual when using a color that clashes with the background it
-is set over or display a color that has poor visibility, and does not offer an alternative color that the user 
-might prefer instead.
+One alternative is setting only one hex value for the color but that was dismissed as it reduces the pool of
+potential colors that a client might have to choose from when there is poor visiblity resulting in a less
+personalized experience.
 
 ## Security considerations
 
--- a/proposals/4522-username-colors.md
+++ b/proposals/4522-username-colors.md
@@ -13,16 +13,15 @@ that could be specific to a room, a powerlevel, or PMP
 Color selection should have a unified cross platform solution in order to best represent every case.
 
 For standardization:
- - per-account colors, profiles may have an optional `m.colors` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
- - per-room colors may be set through a `m.colors` field within the `m.room.member` state event
- - per-powerlevel colors may be set through a `colors` field within a new `m.powerlevels_style` room state object
+ - per-account colors, profiles may have an optional `m.color` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+ - per-room colors may be set through a `m.color` field within the `m.room.member` state event
+ - per-powerlevel colors may be set through a `color` field within a new `m.powerlevels_style` room state object
 
-If a custom color is wished to be set for an ID of a PMP it may be set through a `m.colors` field within the 
+If a custom color is wished to be set for an ID of a PMP it may be set through a `m.color` field within the 
 `m.per_message_profile` of the message
 
-  The `m.colors` and `color` fields represent an object that contains the colors that may be choosen depending
-on the active theme of the client. The light_color represents the color that is lighter, and dark_color the color that is darker,
-so a light_color should ideally be used when the user has a dark theme and vice versa.
+  The `m.color` and `color` fields represent the theme that the color is to be displayed over, 
+the `on_dark` value represents the color that should be used when the client uses a dark theme and `on_light` on light themes.
 
 The field within the extended profile can be for example:
 ```json
@@ -31,9 +30,9 @@ The field within the extended profile can be for example:
   "displayname": "Shea Butter",
   "m.banner_url": "mxc://matrix.org/example_banner",
   "m.tz": "Europe/Troll",
-  "m.colors": {
-    "light_color": "#ffd9f5",
-    "dark_color": "#440000" 
+  "m.color": {
+    "on_dark": "#ffd9f5",
+    "on_light": "#440000" 
   }
 }
 ```
@@ -44,9 +43,9 @@ The room state `m.room.member` may be updated to for example:
   "membership": "join",
   "displayname": "User",
   "avatar_url": "mxc://example.com/code",
-  "m.colors": {
-    "light_color": "#f00",
-    "dark_color": "#400" 
+  "m.color": {
+    "on_dark": "#f00",
+    "on_light": "#400" 
   }
 }
 ```
@@ -58,32 +57,32 @@ The room state `m.powerlevels_style` represents a potential override, and an arr
     "powerlevels": [
       {
         "powerlevel": 51,
-        "colors": {
+        "color": {
           "light_color": "#f00",
           "dark_color": "#400" 
         }
       },
       {
         "powerlevel": 1,
-        "colors": {
-          "light_color": "#f8f",
-          "dark_color": "#404" 
+        "color": {
+          "on_dark": "#f8f",
+          "on_light": "#404" 
         }
       }
     ]
   },
 ```
 
-The `m.colors` for PMPs may look for example:
+The `m.color` for PMPs may look for example:
 ```json
 "m.per_message_profile": {
   "id": "nough",
   "displayname": "Nough",
   "avatar_url": "mxc://example.com/example_url",
   "has_fallback": true,
-  "m.colors": {
-    "light_color": "#82F2A3",
-    "dark_color": "#11403A" 
+  "m.color": {
+    "on_dark": "#82F2A3",
+    "on_light": "#11403A" 
   }
 },
 ```
@@ -121,7 +120,7 @@ There are no new security issues introduced by this proposal
 
 ## Unstable prefix
 
-`eu.she-a.colors` should be used instead of `m.colors`       
+`eu.she-a.color` should be used instead of `m.color`       
 
 `eu.she-a.powerlevels_style` should be used instead of `m.powerlevels_style`
 
--- a/proposals/4522-username-colors.md
+++ b/proposals/4522-username-colors.md
@@ -5,7 +5,7 @@ and with the advent of extensible events clients have started to create unique i
 that are not cross compatible.
 
 There needs to be a way to tell a client that your name should have a specific color, ideally one 
-that could be specific to a room, a role, or PMP
+that could be specific to a room, a powerlevel, or PMP
 
 
 ## Proposal
@@ -14,12 +14,11 @@ Color selection should have a unified cross platform solution in order to best r
 
 For standardization:
  - per-account colors, profiles may have an optional `m.colors` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
- - per-room colors may be set through a `m.users_colors` room state object
- - per-role colors may be set through a `colors` field within a new `m.roles_style` room state object
+ - per-room colors may be set through a `m.colors` field within the `m.room.member` state event
+ - per-powerlevel colors may be set through a `colors` field within a new `m.powerlevels_style` room state object
 
-If a custom color is wished to be set for an ID of a PMP, that may be done as follows:
- - per-message colors may be set through a `m.colors` field within the `m.per_message_profile` of the message
- - per-room colors may be set through the same means as the account ones, but prefixed with the id (eg. @foo@bar:example.org )
+If a custom color is wished to be set for an ID of a PMP it may be set through a `m.colors` field within the 
+`m.per_message_profile` of the message
 
   The `m.colors` and `color` fields represent arrays of objects for the color. Each object of the array
 should contain the color that is to be displayed and may contain a `background` that it is meant to 
@@ -45,28 +44,19 @@ The field within the extended profile can be for example:
 }
 ```
 
-The room state `m.users_colors` is an array of objects with the name and colors array as for example:
+The room state `m.room.member` may be updated to for example:
 ```json
 {
-    "m.users": [
-      {
-        "userId": "@foo:example.com",
-        "colors": [
-          { "color": "#00f", "background": "#818181" },
-          { "color": "#88f", "background": "#808080" }
-        ]
-      },
-      {
-        "userId": "@bar@foo:example.com",
-        "colors": [
-          { "color": "#f00" }
-        ]
-      }
-    ]
-  },
+  "membership": "join",
+  "displayname": "User",
+  "avatar_url": "mxc://example.com/code",
+  "m.colors": [
+    {"color": "#ff0", "background": "000"}
+  ]
+}
 ```
 
-The room state `m.roles_style` represents an array of objects with powerlevel, an optional override, and a colors array as for example:
+The room state `m.powerlevels_style` represents a potential override, and an array of objects with powerlevels with and a colors array as for example:
 ```json
 {
     "override_other_colors": false,
@@ -91,7 +81,7 @@ The room state `m.roles_style` represents an array of objects with powerlevel, a
 
 The `m.colors` for PMPs may look for example:
 ```json
-"com.beeper.per_message_profile": {
+"m.per_message_profile": {
   "id": "nough",
   "displayname": "Nough",
   "avatar_url": "mxc://example.com/example_url",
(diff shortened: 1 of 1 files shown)


proposals/4522-username-colors.md
turt2live, line 1: Implementation requirements:

* [ ] Client (setting)
* [ ] Client (showing) - ideally multiple

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+that could be specific to a room, a role, or PMP
+
+
+## Proposal
+
+Color selection should have a unified cross platform solution in order to best represent every case.
+
+For standardization:
+ - per-account colors, profiles may have an optional `m.colors` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+ - per-room colors may be set through a `m.users_colors` room state object
```
thetayloredman: why `users_colors` in contrast to a user's `m.room.member` event? this seems like it would allow any user to override any other user's color chosen.

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+ - per-role colors may be set through a `colors` field within a new `m.roles_style` room state object
```
thetayloredman: Matrix has no concept of a "role"

turt2live:
(please avoid force pushing now that the MSC has been labeled - we're not afraid of having lots of commits)

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+
+If a custom color is wished to be set for an ID of a PMP, that may be done as follows:
+ - per-message colors may be set through a `m.colors` field within the `m.per_message_profile` of the message
+ - per-room colors may be set through the same means as the account ones, but prefixed with the id (eg. @foo@bar:example.org )
```
thetayloredman: why do PMPs need a room-scoped color setting, when color should already be set on each message?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+
+  The `m.colors` and `color` fields represent arrays of objects for the color. Each object of the array
+should contain the color that is to be displayed and may contain a `background` that it is meant to 
+represent the ideal background the `color` should be displayed over. The background SHOULD NOT be 
+displayed but should only serve as a way for the client to determine which color best fits the background 
+it is displayed against.
```
thetayloredman: how is this calculated?

nushea:
Oh sorry, i will keep this in mind for the future

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+        ]
+      },
+      {
+        "userId": "@bar@foo:example.com",
+        "colors": [
+          { "color": "#f00" }
+        ]
+      }
+    ]
+  },
```
thetayloredman: why not just use an object keyed by MXID?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+          { "color": "#fff", "background": "#000000" }
+        ]
+      }
+    ]
+  },
+```
+
+The room state `m.colors` for PMPs may look for example for example:
+```json
+"com.beeper.per_message_profile": {
```
thetayloredman: PMPs are defined in account_data, not room state. additionally the stable identifier should probably be used here.

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,149 @@
...
+  "m.colors": [
+    { "color": "#ffd9f5", "background": "#000000" },
+    { "color": "#440000", "background": "#fff" },
+    { "color": "#f0f", "background": "#888" },
+    { "color": "#ff0000" }
+  ]
+}
+```
+
+The room state `m.users_colors` is an array of objects with the name and colors array as for example:
```
timedoutuk: why not just make this a field in the `m.room.member` event? as it stands this means anyone can overwrite anyone else's colours, especially since the state key isn't even restricted to per-member

proposals/4522-username-colors.md
nushea: the objective was to be override-able for a room administrator in case that they enforce a coloring scheme, but i will just let the room color override then

proposals/4522-username-colors.md
nushea: Removed it and set it so that per-room colors are now only within `m.room.member`

proposals/4522-username-colors.md
nushea: It is left to the discretion of a client to decide which color fits the background, but it is essentially a way to guide the client towards which color would look best over the theme that it has for the background of the object the name is displayed on

lveneris: in that case, why is it here? there are standard formulae for accessible contrast ratios, and if clients are going to pick a value with their theme in mind anyway, i do not see a reason to have a value in the event itself.

also, has UX been considered? it would be far more understandable to allow people to pick a value for a light theme and a value for a dark theme, rather than an array of arbitrary backgrounds and which colour they'd want to appear over them. if you want to be extra snazzy, you can use the aforementioned formulae to extrapolate a desired value for *any* background colour creating a gradient with these two points.

nushea: Removed the options in order to lower the implementation complexity so that now there are only 2 colors that are specific to themes so there in no reason in standardizing the contrast rules

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,131 @@
...
+
+The field within the extended profile can be for example:
+```json
+{
+  "avatar_url": "mxc://matrix.org/example",
+  "displayname": "Shea Butter",
+  "m.banner_url": "mxc://matrix.org/example_banner",
+  "m.tz": "Europe/Troll",
+  "m.colors": {
+    "light_color": "#ffd9f5",
```
j0lol: Maybe nitpicking, but the “light color” being for a dark theme could be confusing? I almost want to suggest something like `on_dark` but that is a bit awkward too.

j0lol: I think it might be confusing either way…

nushea: True, but im not sure if it would be better as either "on_dark_color" or "color_on_dark", maybe something akin to dark_themed_color and light_themed_color but that isnt much better and is also longwinded so idk what i should name the 2 options

j0lol: The UI for this in PMPs looks like this right now. As you can see the language is a bit odd (the light one is for the dark theme, vice versa)
IMG_7449

nushea: Changed it to a keyname that is more clear

subpop:
Clients tend to follow a color scheme either set by the overall design/vision, or by following conventions and guidelines set by the deployment target. Having colors that don't render well on the client's background or with the client's fonts reduces the client's ability to present a consistent UX. Would this proposal permit clients to adjust the specified color to fit into their color palette (i.e. adjust the saturation or contrast to better render with a certain background color)? Or even disregard colors entirely?

I developed a [stable color algorithm derived from a matrix ID](https://sub-pop.net/why-is-everyone-green-or-purple/) and use that color to render colored message bubbles in my macOS Client. I'm not sure how the client UX would change if I didn't have the ability to control the color palette of the messages anymore.

j0lol:
> Would this proposal permit clients to adjust the specified color to fit into their color palette [...]? 

One interesting way to spin this would be to split the color into HSL (or maybe OKLCH) it as more of a:

- Clients SHOULD render hue unless disregarded by the user for accessibility reasons
- Clients MAY render Saturation/Lightness (or in LCH, Chroma/Lightness), or alter/disregard these to fit their color scheme

nushea:
> Clients tend to follow a color scheme either set by the overall design/vision, or by following conventions and guidelines set by the deployment target. Having colors that don't render well on the client's background or with the client's fonts reduces the client's ability to present a consistent UX. Would this proposal permit clients to adjust the specified color to fit into their color palette (i.e. adjust the saturation or contrast to better render with a certain background color)? Or even disregard colors entirely?
> 
> I developed a [stable color algorithm derived from a matrix ID](https://sub-pop.net/why-is-everyone-green-or-purple/) and use that color to render colored message bubbles in my macOS Client. I'm not sure how the client UX would change if I didn't have the ability to control the color palette of the messages anymore.

This MSC is meant for the name but yes accessibility should be considered by the client implementing the MSC, and especially for accessibility reasons colors may be disregarded, following the list of fallback colors (ie if a per-room color is disregarded for theming or accessibility purposes the per-account color could be used instead).

However for modifying the colors themselves that would limit the real options that a user might have when selecting the color but I could change the MSC to mention that 

"a client may alter the selected color, but should try to avoid altering the hue if possible, and may alter/discard the Saturation/Lightness in order to better fit the overall theme of the client" or anything similar following the principles that [j0lol](https://github.com/j0lol) described above

subpop:
> > Would this proposal permit clients to adjust the specified color to fit into their color palette [...]?
> 
> One interesting way to spin this would be to split the color into HSL (or maybe OKLCH) it as more of a:
> 
> * Clients SHOULD render hue unless disregarded by the user for accessibility reasons
> * Clients MAY render Saturation/Lightness (or in LCH, Chroma/Lightness), or alter/disregard these to fit their color scheme

I'm partial to OKLCH since I settled on that in my client, but I like the general idea of storing the hue and letting clients finalize the rendered color.

> This MSC is meant for the name

In my client, names are rendered in small gray text above the message bubble. The bubble gets the color treatment, as does the user's profile/avatar if they don't have one. I think that's your intention with this spec though.

image

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,130 @@
...
+  "displayname": "User",
+  "avatar_url": "mxc://example.com/code",
+  "m.color": {
+    "on_dark": "#f00",
+    "on_light": "#400" 
+  }
+}
+```
+
+The room state `m.powerlevels_style` represents a potential override, and an array of objects with powerlevels with and a colors array as for example:
```
Half-Shot: This feels like it's serving a bit of a different purpose to the other two. While setting your own color is purely decorative, this one feels like it's giving meaning to the colors.

This is a problem because now it means users who are not able to see the color will not see the meaning. This feels a bit like you've started to define powerlevels-as-roles (not too far off https://github.com/matrix-org/matrix-spec-proposals/pull/4056/changes).

I'd probably suggest splitting out the power level stuff to another MSC as I think it's going to be hard to drive this change in too.

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,130 @@
...
+```json
+{
+    "override_other_colors": false,
+    "powerlevels": [
+      {
+        "powerlevel": 51,
```
Half-Shot: This is a bit confusing because, if I were powerlevel 52 then I would have all the abilities of PL 51 without the color. Is this a minimum PL?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,130 @@
...
+  The color fields must be a hex color for consistency.
+
+  The order of the color values should be: PMP color, per-room color, account profile color, 
+powerlevel color, default fallback that a client SHOULD have. This order may be overridden 
+for powerlevels to take precedence above every other color.
+
+## Potential issues
+
+A user may choose to set their per-room or per account color to match one of a modertator of a room, but this
+may be mitigated by setting `override_other_colors` to true if needed.
```
Half-Shot: This seems a bit perplexing. If you set `override_other_colors` to false or omit it then any user can impersonate an admin quite easily, and if you enable it in practice nobody can set custom colours?

I'd imagine in nearly all cases you *wouldn't* want impersonations except for DMs and maybe small groups, which probably wouldn't have any `m.powerlevels_style` events anyway so it's kinda moot.

Essentially: What case would there be for `override_other_colors` to be anything other than true?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,130 @@
...
+There needs to be a way to tell a client that your name should have a specific color, ideally one 
+that could be specific to a room, a powerlevel, or PMP
+
+
+## Proposal
+
+Color selection should have a unified cross platform solution in order to best represent every case.
+
+For standardization:
+ - per-account colors, profiles may have an optional `m.color` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
```
Half-Shot: Nitpicking but, I'd probably have a more descriptive field here like `m.name_color`

j0lol:
I guess maybe “name color” is a bit too client dependent? Maybe like, “color preference” is a better goal for this MSC.

proposals/4522-username-colors.md
nushea: The usecase for it being set to false is smaller or friend-oriented communities, where the structure is more happenstance than a stance of one individual as more managerial than anyone else, and communities where there is a more relaxed atmosphere in general. 

I'd imagine most groups would have it turned off besides for software distribution groups where provenance is most important since for other situations of impersonation there would be little impact but would be very easy punish, similar to the current situation where impersonation is very easy to both do and to punish

Though there would be other uses such a room having specific level that unlocks permision, so a lower pl user might not be allowed to respond to polls etc and also not have a unique color until they are trusted by the community

proposals/4522-username-colors.md
nushea: Hi, just wondering, i have changed it to the final prefix to color_preference since it theoretically might be used for more than strictly names, what do you think of this name?

proposals/4522-username-colors.md
nushea: I have decided to remove this and instead hope that more clients will implement #3949 even though it is more limited. Alongside the reason above it was removed to lower implementation costs and potential conflicts with other MSCs, hoping that clients will naturally come behind another powerlevel specific color standard since currently the situation is very divided in that regard as well

nushea:
> I'm partial to OKLCH since I settled on that in my client, but I like the general idea of storing the hue and letting clients finalize the rendered color.

I have modified the msc to endorse the idea of modifying the color to fit technical, theming and accessibility reasons, based on the conversation.           
Would you mind checking it out and advising if there are any concerns or ambiguities with the new form?        
The idea of the MSC is to create a sense of unity between the clients rather than regulating hard limits of coloring.

> 
> In my client, names are rendered in small gray text above the message bubble. The bubble gets the color treatment, as does the user's profile/avatar if they don't have one. I think that's your intention with this spec though.
> image

That makes sense and looks very nice, would love to see how this will look with user input on the colors while keeping client theming guides!

proposals/4522-username-colors.md
nushea: The initial idea would be that every PL that is in use would have its own colors or otherwise would be unset so that it would fall down to the next coloring level, but this is irrelevant now.

(ie if you had override_other_colors: true, you could have role 52 that had no color input and thus people assigned that PL would be able to choose a per-room color)

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Proposal
+
+Color selection should have a unified cross platform solution in order to best represent every case.
+
+For standardization:
+ - per-account colors, profiles may have an optional `m.color_preference` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+ - per-room colors may be set through a `m.color_preference` field within the `m.room.member` state event
+
+If a custom color is wished to be set for an ID of a PMP it may be set through a `m.color_preference` field within the
```
KitsuneRal: Please spell out PMP on the first use.

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,111 @@
...
+
+For standardization:
+ - per-account colors, profiles may have an optional `m.color_preference` field object relying on [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133).
+ - per-room colors may be set through a `m.color_preference` field within the `m.room.member` state event
+
+If a custom color is wished to be set for an ID of a PMP it may be set through a `m.color_preference` field within the 
+`m.per_message_profile` of the message
+
+  The `m.color_preference` object contains 2 fields for the potential colors to be displayed but are named by the theme kind that the name/message is to be displayed over, 
+the `on_dark` value represents the color that should be used when the client uses a dark theme (ie the color itself shoulds be bright) and `on_light` on light themes (ie the color itself should be dark).
```
KitsuneRal: I suggest doing a small research on how the same issue is solved on other platforms and summarising the results in this text. Picking good colours (or rather tones) that would work without knowing the other side's environment is more complex than just picking one for dark backgrounds and one for light ones. You may start with an exploration of what has been done in Matrix clients so far and also look at what people with substantial experience in UI theming ended up with, e.g. https://m3.material.io/styles/color/.

KitsuneRal (changes requested):

proposals/4522-username-colors.md
nushea: The point of the MSC is not to dictate the ways that the colors should be rendered exactly, but rather to give the users choices in how it should be done so, if you would not mind please advise on the issues within the section of alternatives considered

proposals/4522-username-colors.md
nushea: Sable has added support for setting and rendering colors per the MSC https://github.com/SableClient/Sable/pull/1526

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+  The order of precedence for the color values must be: 
+   - PMP color, 
+   - per-room color, 
+   - account profile color, 
+   - [MSC3949](https://github.com/ajbura/matrix-spec-proposals/blob/power-level-tag/proposals/3949-power-level-tag.md) colors if the client implements this MSC, 
+   - default fallback that a client SHOULD have
+
+  The color fields must be a hex color (#RGB or #RRGGBB) for consistency and ease of use for the end-user.
```
lveneris, line 73: Why was RGB chosen? Especially because in the following paragraph, you mention that the hue angle should be preserved, while saturation and lightness can be adjusted, while other systems would make this implementation easier.

Consider: now, the client is likely to convert the RGB colour value into a system with an explicit hue angle, so it can modify the other values trivially without impacting it, and then either display that or convert back.

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,116 @@
...
+    "on_light": "#11403A" 
+  }
+},
+```
+
+  The order of precedence for the color values must be: 
+   - PMP color, 
+   - per-room color, 
+   - account profile color, 
+   - [MSC3949](https://github.com/ajbura/matrix-spec-proposals/blob/power-level-tag/proposals/3949-power-level-tag.md) colors if the client implements this MSC,
```
lveneris: Why was this fallback chosen? What should a fallback be for clients that do NOT implement this MSC, given that the PR indicates only Cinny does?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,116 @@
...
+   - default fallback that a client SHOULD have
+
+  The color fields must be a hex color (#RGB or #RRGGBB) for consistency and ease of use for the end-user.
+
+  A client may adjust the final displayed color to accommodate theming, technical and accessibility requirements.
+However, The client should preserve the hue where possible, and may alter the Saturation/Chroma/Lightness.
+  
+  For example a terminal-based client might have a very limited color palette and might end up compressing 
+all the color options set by the users to an 8-32 selection and that client could still be compliant with the 
+msc as it would be within its limitations.
```
lveneris, line 80: Have you considered allowing users to give a free list of colours in order of preference and have the client pick the first one that they can render and that suits their currently loaded theme? It would be interesting to explore this alternative, even if you do not land on it.

proposals/4522-username-colors.md
nushea: The fallback after that is the default fallback, if a client does not implement that MSC it should skip over the [MSC3349](https://github.com/matrix-org/matrix-doc/pull/3349) option and go straight to its fallback, that MSC it is useful however as it is the only powerlevel-specific MSC that allows you to set colors for each PL, if there were others i would've mentioned them as well

proposals/4522-username-colors.md
nushea: For consideration to the user. HSL and other standards are less understandable to the end user than choosing an RGB color during input, and RGB gives them a unique color.

And also to not make clients support multiple potentially obscure standards for coloring or forcing a specific hue based standard. A client might choose to use HSL or OKLCH or HWB or HSB or a myriad of other available standards for their conversion and there should be nothing forcing to one standard but it would create an unreasonable burden on a client to ensure that it is able to read every single other color standard and convert it to their own one while a conversion to RGB is commonplace for every color standard

proposals/4522-username-colors.md
nushea: That would be an interesting alternative but it would likely make it more costly to implement, the original version of the MSC also proposed having a list of colors based on the theme that was used when the color was selected and people have asked me to reduce it to 2 due to implementation and ux costs. The 2 options are being used since almost every platform offers light and dark modes so it would reduce the implementation burden

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,116 @@
...
+
+The field within the extended profile can be for example:
+```json
+{
+  "avatar_url": "mxc://matrix.org/example",
+  "displayname": "Shea Butter",
+  "m.banner_url": "mxc://matrix.org/example_banner",
+  "m.tz": "Europe/Troll",
+  "m.color_preference": {
+    "on_dark": "#ffd9f5",
```
lveneris, line 33: Another thought that occurred to me: why strings, even? Colours always map to numbers or a series of them, doesn't this just introduce the possibility for garbage data?

I do not think enough weight has been given to alternatives or justification in this MSC. It would be good to see more explanation for the design decisions made here.

proposals/4522-username-colors.md
lveneris: And, is there an intended default fallback, or is that up to clients to decide? Are there no recommendations?

proposals/4522-username-colors.md
lveneris: There are coherent colour pickers available in most systems. RGB is as understandable to most people as any series of numbers. I would need to see some evaluation of these design potentials before I could get on board with any particular solution, although I probably would have shot for HSL or OKLCH, which are both broadly used and do not have the shortcomings of RGB in this regard.

proposals/4522-username-colors.md
nushea: The point of the MSC it to give clients more color options, not to restrict their behavior beyond what is absolutely necessary

proposals/4522-username-colors.md
nushea: Most other platforms that allow a user to set a color use RGB, 

There is no benefit from choosing anything else, a client would need to convert into their own preferred color standard regardless, and while the burden of converting between OKLCH to a color that you may display on a terminal is unjustifiable, 24bit RGB is even supported and understood by most terminals through ANSI. A client that already uses a more complex coloring system such as clients that would employ OKLCH would also have a very simple RGB to OKLab colors.

proposals/4522-username-colors.md
nushea: Numbers can also produce incorrect data, using numbers would just add unnecessary steps in the processing of the colors

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,116 @@
...
+
+`eu.she-a.color` should be used instead of `m.color_preference`      
+
+## Dependencies
+
+This MSC builds on:
+ - [MSC4133](https://github.com/matrix-org/matrix-spec-proposals/pull/4133) for per-account colors,
+ - [MSC4144](https://github.com/matrix-org/matrix-spec-proposals/pull/4144) for the per-message-profile object, 
+
+And wishes that more clients would implement [MSC3949](https://github.com/ajbura/matrix-spec-proposals/blob/power-level-tag/proposals/3949-power-level-tag.md) or some other standard powerlevel specific customization
```
bitnomad2:
```suggestion
And wishes that more clients would implement [MSC3949](https://github.com/matrix-org/matrix-spec-proposals/pull/3949) or some other standard powerlevel specific customization
```

bitnomad2 (changes requested):
Bit nitpick but PR page seems to be norm when mentioning proposals.

proposals/4522-username-colors.md
nushea: Hi, i have applied the change, initially i only linked to the proposal itself, thank you for pointing it out!

proposals/4522-username-colors.md
TheArcaneBrony: making it a string makes it harder to process, making it a number probably makes more sense?

proposals/4522-username-colors.md
```diff
@@ -0,0 +1,130 @@
...
+  ```json
+  m.preferred_colors = ["#ffd9f5", "#440000", "#460333"]
+  ```
+  And while that method would be a very pleasant and elegant way to set colors, it would present an increased cost of implementation, allowing arbitrarily long lists of 
+colors across 3 of the stages of the order of precedence.
+
+## Security considerations
+
+There are no new security issues introduced by this proposal
+
```
TheArcaneBrony, line 119: Should probably mention that this makes impersonation easier, given that as far as I'm aware, the colors were meant to disambiguate similar or equal profiles, since it's usually a hash of the mxid.

nushea: While some clients implement powerlevel based colors (such as regular cinny) and some use pseudorandom colors for their users (such as gomuks/commet, with them being different between clients), the spec has never given them a role in disambiguation, and the usage of colors has been completely unreliable between clients rendering any usage in disambiguation not useful.

The spec itself does define how clients should ensure that impersonation should not happen with https://spec.matrix.org/v1.18/client-server-api/#calculating-the-display-name-for-a-user

So while i could mention it i am not sure if it would be relevant to the MSC as it does not introduce new to the spec ways of impersonation

proposals/4522-username-colors.md
nvnvnvnvnvnvnv: Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/18608e3f2084add531d3f341c0b790999ec864d1

PR #4525, opened 2026-08-06
MSC4525: Paginated Sync

ara4n:
[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/matthew/paginated-sync/proposals/4525-paginated-sync.md)

This is an experiment in further simplifying SSS based on input from @erikjohnston and others, drafted with my SCT lead hat on.

Experimental implementations:
 * synapse: https://github.com/element-hq/synapse/pull/20073
 * matrix-rust-sdk: https://github.com/matrix-org/matrix-rust-sdk/compare/matthew/preview-prefill...matthew/paginated-sync?expand=1
 * element-x-ios: https://github.com/element-hq/element-x-ios/compare/matthew/preview-prefill...matthew/paginated-sync?expand=1

### DIFF
--- a/proposals/4525-paginated-sync.md
+++ b/proposals/4525-paginated-sync.md
@@ -27,11 +27,11 @@ most of the remaining problems:
 
 * ...and connection expiry is very disruptive: When a connection expires (after ~30 minutes 
   offline, or when too many updates have stacked up), the client starts from scratch: it re-sends 
-  the full request, re-downloads account data, push rules, read receipts, and re-grows its ranges 
-  over the entire room list, re-fetching a page of rooms it almost entirely already has. All of that
-  bandwidth and battery is spent recovering state the client never lost.  Moreover, the UX of
-  "timeline resets" on clients can be bad if they throw away their timeline view/cache; this removes
-  the concept of timeline resets entirely.
+  the full request, re-downloads account data, push rules, read receipts, marks all E2EE as dirty
+  (obligating /key/query on all known devices), and re-grows its ranges 
+  over the entire room list, re-fetching a page of rooms it almost entirely already has.
+  Moreover, the UX of "timeline resets" on clients can be bad if they throw away their timeline
+  view/cache; this removes the concept of timeline resets entirely, radically speeding up launches.
 
 * `timeline_limit: 1` blows holes in the timeline: To keep list responses small when initially
   growing the all_rooms list, clients run their lists with a timeline limit of 1. This means that
--- a/proposals/4525-paginated-sync.md
+++ b/proposals/4525-paginated-sync.md
@@ -29,7 +29,9 @@ most of the remaining problems:
   offline, or when too many updates have stacked up), the client starts from scratch: it re-sends 
   the full request, re-downloads account data, push rules, read receipts, and re-grows its ranges 
   over the entire room list, re-fetching a page of rooms it almost entirely already has. All of that
-  bandwidth and battery is spent recovering state the client never lost.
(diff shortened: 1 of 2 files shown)


proposals/4525-paginated-sync.md
```diff
@@ -0,0 +1,257 @@
...
+  `pending` gives the server an obvious signal for genuinely dead connections.
+* Room list *ordering* on the client is by `bump_stamp`, computed from at most `history` events on
+  cold start; if a room's most recent events are all non-"proper" activity (reactions, state), the
+  client may need `/messages` to find a preview-worthy event. This is already true of MSC4186 with
+  `timeline_limit: 1` - and unlike MSC4186, the client can actually fix it, cheaply, per-room.
+* Losing subscriptions means a client cannot ask for elevated `required_state`/timeline for one
+  specific room. In practice Element X used subscriptions primarily to hear about visible rooms at
+  all (because they were outside the ranges) - which this design makes automatic, as every room
+  with traffic is delivered. Clients wanting deep state for an opened room fetch it with existing
+  per-room APIs (`/state`, `/members`, `/messages`).
```
kegsay, line 215: ..none of which provide live updates. This design forces you to either pull every room's topic or pull no room's topic and then poll for it which isn't great.

proposals/4525-paginated-sync.md
```diff
@@ -0,0 +1,257 @@
...
+A room gets 500 messages while the app is foregrounded: each sync response delivers at most
+`limit` of them; if the client falls behind the room comes down `limited` with a `prev_batch`,
+i.e. the gap is explicit and local to that room, rather than being escalated into a connection
+reset that takes the other 2,599 rooms' state with it.
+
+## Potential issues
+
+* A client which never drains its backlog (e.g. syncs once then sleeps) accumulates pending rooms
+  server-side; but the tracking cost is the same per-room bookkeeping MSC4186 already pays, and
+  `pending` gives the server an obvious signal for genuinely dead connections.
```
kegsay, line 206: This skirts around the problem. Losing the ability to force expire connections means you can't ever safely clean up streams because the client may come back at some point in time. 

> `pending` gives the server an obvious signal for genuinely dead connections

no it doesn't, it just let's you know how far behind that client is, it does not take frequency or durations into account at all.

MatMaul:
For the record, here is a verbatim on my Q&A in the SCT room:

> Me: My concern is about maintainability of the server impls and the spec. Since https://github.com/matrix-org/matrix-spec-proposals/pull/4186 will reach the spec soon, then it will be really complicated to deprecate. if we have something that make sense and would notably simplify both the spec and the impls in sight, shouldn't we wait for it instead of getting 4186 over the line ?

> richvdh: My understanding is that MSC4525 is just Matthew experimenting with new ideas and it will not be replacing MSC4186 any time soon. The SCT remains committed to MSC4186 (and the extensions thereof).

> Matthew: MSC4186 is what we're merging into the spec for the long-overdue 2.0 release, complete with its extensions. however, 4186's alternatives section did outline the pagination idea, and i experimented with it in 4525 while on holiday - but it is very much a post-2.0 thing.

> Me: thanks. I guess the burden of what would be removed by 4525 is not that high compared to the overall complexity of the full MSC anyway.

nvnvnvnvnvnvnv:
Client implementation ([Voyage](https://github.com/VoyageClient/Voyage)):

https://github.com/VoyageClient/Voyage/commit/6f92d206b9f6c1f0424c6721cc59a05cfec83c4e

PR #4527, opened 2026-08-09
MSC4527: State event power levels

thetayloredman:
[Rendered](https://github.com/thetayloredman/matrix-spec-proposals/blob/ln/state-power-levels/proposals/4527-state-event-power-levels.md)

Implementations:
- [Complement](https://github.com/thetayloredman/complement/tree/msc4527)
- [Ruma](https://github.com/thetayloredman/ruma/tree/ln/msc4527)
- [Continuwuity](https://forgejo.ellis.link/thetayloredman/continuwuity/src/branch/msc4527)

### DIFF
--- a/proposals/4527-state-event-power-levels.md
+++ b/proposals/4527-state-event-power-levels.md
@@ -1,0 +1,169 @@
+# MSC4527: State event power levels
+
+Many communities on Matrix choose to enforce a strict [`m.room.power_levels`] that restricts members from sending
+unknown or future event types by setting `events_default` > `users_default` and setting specific `events` overrides
+for event types they desire to allow (e.g. `m.reaction` and `m.room.message`).
+
+When a room's power levels are configured in this way, the value in `events` overrides `state_default`, hence an
+unprivileged user can send a permitted event type as a state event. This allows that event to become part of room state
+permanently, which is behavior often undesirable in large public rooms as it increases room complexity.
+
+This proposal introduces a separate map for controlling the power levels required to send state events separately from
+message events, alongside the required changes to [auth rules] and the [redaction algorithm].
+
+**Note:** At the time of writing, [Room Version 12] is the latest room version available. This proposal uses v12 as a
+base room version for ease of reference, but the concepts/changes can apply to future room versions as well.
+
+## Problem
+
+Given this power levels event:
+```json
+{
+    "events": {
+        "m.room.message": 0
+    },
+    "events_default": 1,
+    "state_default": 50,
+    "users_default": 0
+}
+```
+
+An unprivileged user can legally send the following event:
+
+```jsonc
+{
+    "type": "m.room.message",
+    "state_key": "test",
+    "content": {
+        "body": "Hello, room state!",
+        "msgtype": "m.text" // required by `m.room.message`
+    }
+}
+```
+
+## Proposal
+
+This proposal makes breaking changes to the event authorization rules, and therefore requires a new room version.
+
+### Changes to the `m.room.power_levels` schema
+
+A new key `state` is added to the [`m.room.power_levels`] event, of type `{string: integer}`. Identically to `events`,
+the `state` key consists of a mapping from event type to power level required.
+
+If the field is missing, or the room has no `m.room.power_levels` event, the value of `state` is assumed to be `{}`.
+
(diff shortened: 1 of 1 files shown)


proposals/4527-state-event-power-levels.md
turt2live, line 1: Implementation requirements:
* [ ] Client backwards compatibility
* [ ] Client forwards compatibility

proposals/4527-state-event-power-levels.md
```diff
@@ -0,0 +1,169 @@
...
+unprivileged user can send a permitted event type as a state event. This allows that event to become part of room state
+permanently, which is behavior often undesirable in large public rooms as it increases room complexity.
+
+This proposal introduces a separate map for controlling the power levels required to send state events separately from
+message events, alongside the required changes to [auth rules] and the [redaction algorithm].
+
+**Note:** At the time of writing, [Room Version 12] is the latest room version available. This proposal uses v12 as a
+base room version for ease of reference, but the concepts/changes can apply to future room versions as well.
+
+## Problem
```
Susurrus, line 17: Might be also mentioning things that are proposed that require both State & Room messages: e.g. [MSC3489](https://github.com/matrix-org/matrix-spec-proposals/pull/3489). With it as-proposed, you can't properly implement access restrictions for it, as the state of the sharing is a State event, which can't be restricted, whereas the individual location updates are Room events (if that's the right terminology). So this also means you get bad UX for users, as they don't get any errors when initially starting a share, but do later..

thetayloredman: The same could occur with current rooms if an administrator changes power levels while sharing is in progress. Clients should likely check proactively if they can send both types - as they already do for sending messages.

proposals/4527-state-event-power-levels.md
```diff
@@ -0,0 +1,169 @@
+# MSC4527: State event power levels
```
kegsay, line 1: I like this MSC, and it's surprising that we don't have this ability already.

I worry that this change is invasive though due to the ecosystem-wide impact it has, as literally everything reads the content of a PL event. Alternatives like "Global minimum power level to send state" are more compatible with the ecosystem at the cost of it being less expressive. I would like to explore what use cases are broken/not possible with the "global minimum level to send state" alternative, as I prefer that alternative to fix the stated problem.

On my side, there is a security angle here because state events are more expensive than message events. Users may configure the PL event thinking that unprivileged users can't send state but yet as you show, they can.

thetayloredman: @kegsay 
The answer to what use case that would forbid is simple, and on my mind in the writing process: `org.matrix.msc3401.call.member`. I anticipate some rooms would like to allow all users to join MatrixRTC calls, while still gaining from the benefits of a strict power levels configuration.

Susurrus: You may also want to do the same for `m.beacon_info`, disallow it while allowing `m.call.member` for example

proposals/4527-state-event-power-levels.md
```diff
@@ -0,0 +1,169 @@
...
+## Proposal
+
+This proposal makes breaking changes to the event authorization rules, and therefore requires a new room version.
+
+### Changes to the `m.room.power_levels` schema
+
+A new key `state` is added to the [`m.room.power_levels`] event, of type `{string: integer}`. Identically to `events`,
+the `state` key consists of a mapping from event type to power level required.
+
+If the field is missing, or the room has no `m.room.power_levels` event, the value of `state` is assumed to be `{}`.
```
kegsay, line 53: For backwards compat this is pretty bad. An old bot who only ever knows about `events` will expect `"m.room.name": 50` to apply to both message and state events. From a security perspective, it's _state events_ which need protecting here. If we accidentally allow anyone `>= state_default` to tombstone rooms, set server ACLs or change PLs that's pretty bad.

We might want to invert the responsibility of the new field so the existing `events` field affects state events and the new field affects message events. This at least ensures that bots/bridges/clients that fail to understand the new format don't set insecure configurations.

thetayloredman: It could be done using an `events` and a `messages` map, if that's what would satisfy this?

PR #4528, opened 2026-08-11
MSC4528: `M_INCOMPATIBLE_SERVER` error code

erikjohnston:
Specify appropriate error codes for MSC4291 and MSC4311 where it was left as a 5xx.

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/incompatible_server_error_code/proposals/4528-incompatible-server-error.md)

Disclaimer: Claude was used for the rough first draft. This has been subsequently edited a fair bit.

### DIFF
--- a/proposals/4528-incompatible-server-error.md
+++ b/proposals/4528-incompatible-server-error.md
@@ -1,0 +1,117 @@
+# MSC4528: `M_INCOMPATIBLE_SERVER` error code
+
+Following [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291) and
+[MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311), the Server-Server API
+tells sending servers to convert certain federation errors into 5xx errors over the Client-Server
+API. [`PUT /_matrix/federation/v2/send_join/{roomId}/{eventId}`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv2send_joinroomideventid)
+and [`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv2inviteroomideventid)
+both say of their `400` response:
+
+> If `M_MISSING_PARAM` or `M_INVALID_PARAM` is returned and the request is associated
+> with a Client-Server API request, the Client-Server API request SHOULD fail
+> with a 5xx error rather than being passed through.
+
+These rejections occur when the events in the request fail the validation introduced by those MSCs,
+for example when an invite lacks a full-PDU `m.room.create` event in its stripped state. The
+rationale for a 5xx is that there is nothing the client can do differently to make the request
+succeed.
+
+This translation has three problems:
+
+1. No specific status code or `errcode` is given, so clients cannot detect this case and give a more
+   helpful error message.
+
+2. Many clients and client SDKs automatically retry requests which fail with a 5xx status code.
+   However, the failures in the MSCs are not transient as the two servers will remain incompatible
+   until one of them is updated, which will not happen within any retry window.
+
+3. There is no good 5xx error code to use. 500 response rates are often monitored and assumed to
+   mean something has gone wrong and needs fixing. 502/503 response are often interpreted by reverse
+   proxies and load balancers as the backend being down, and will stop sending traffic to said
+   backend. The other defined 5xx errors don't really fit and are rarely used.
+
+This proposal replaces the 5xx translation with a `400` response carrying a new standard error code,
+`M_INCOMPATIBLE_SERVER`, meaning the request failed because the local server and a remote server it
+needed to communicate with are incompatible with each other.
+
+The spec already uses `400` error codes to indicate final error responses that are not the fault of
+the sending client, such as `M_INCOMPATIBLE_ROOM_VERSION`.
+
+## Proposal
+
+A new standard error code is added to the Client-Server API:
+
+* `M_INCOMPATIBLE_SERVER`: The request could not be completed because it required communicating with
+  a remote server which is incompatible with the local server.
+
+Clients SHOULD NOT automatically retry a request which failed with `M_INCOMPATIBLE_SERVER` and
+SHOULD surface the failure to the user. Servers SHOULD use the human-readable `error` field to describe
+the incompatibility, which clients MAY show to the user.
+
+Future proposals SHOULD continue to strive to maintain backwards compatibility where feasible,
+rather than use `M_INCOMPATIBLE_SERVER`.
+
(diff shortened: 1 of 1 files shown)


proposals/4528-incompatible-server-error.md
turt2live, line 1: Implementation requirements:

* [x] Server (throwing) https://github.com/element-hq/synapse/pull/19723/commits/f6beafb46ce453842f7b17646a47646675874e15
* [ ] Client (backwards compatibility)
* [ ] Client (forwards compatibility)

proposals/4528-incompatible-server-error.md
```diff
@@ -0,0 +1,117 @@
...
+Following [MSC4291](https://github.com/matrix-org/matrix-spec-proposals/pull/4291) and
+[MSC4311](https://github.com/matrix-org/matrix-spec-proposals/pull/4311), the Server-Server API
+tells sending servers to convert certain federation errors into 5xx errors over the Client-Server
+API. [`PUT /_matrix/federation/v2/send_join/{roomId}/{eventId}`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv2send_joinroomideventid)
+and [`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.19/server-server-api/#put_matrixfederationv2inviteroomideventid)
+both say of their `400` response:
+
+> If `M_MISSING_PARAM` or `M_INVALID_PARAM` is returned and the request is associated
+> with a Client-Server API request, the Client-Server API request SHOULD fail
+> with a 5xx error rather than being passed through.
```
MadLittleMods, line 12: This MSC is spawning from trying to implement the spec, see https://github.com/element-hq/synapse/pull/19723#discussion_r3630750734

proposals/4528-incompatible-server-error.md
```diff
@@ -0,0 +1,117 @@
+# MSC4528: `M_INCOMPATIBLE_SERVER` error code
```
MadLittleMods, line 1: Fixed link to the rendered MSC: https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/incompatible_server_error_code/proposals/4528-incompatible-server-error.md

(I'm unable to edit the PR description)

erikjohnston: Every. Single. Time.

proposals/4528-incompatible-server-error.md
erikjohnston: Not sure if this really requires an impl?

proposals/4528-incompatible-server-error.md
```diff
@@ -0,0 +1,117 @@
...
+gateway to a remote server which returned an unusable response. However, using a 502 here goes
+against the common assumptions made by clients and proxies about such errors.
+
+## Security considerations
+
+None foreseen.
+
+## Unstable prefix
+
+None needed.
```
MadLittleMods, line 113: Unstable error code needed before this MSC lands.

Guessing `ORG.MATRIX.MSC4528.M_INCOMPATIBLE_SERVER`

erikjohnston: Can you elaborate why this would be useful? Technically unstable prefixes should only be used behind experimental feature flags, and I do not think anyone is going to bother implementing this behind feature flags (nor is it really needed).

tulir: ftr `ORG.MATRIX.MSC4528_INCOMPATIBLE_SERVER` is the correct format for an unstable error code (but no need to change it since it was already committed in the synapse PR)

The `M_` prefix is the equivalent of `m.` in other identifiers, which should never be a part of unstable prefixes.

proposals/4528-incompatible-server-error.md
turt2live: Clients have historically fallen over with error codes, so we should ensure that those which are unaware of the new code handle it reasonably well (ie: don't implode).

erikjohnston: The client endpoints (/invite and /join) can already fail with a variety of error codes, so I don't think there is much to check? The existing MSC already requires clients to generically handle errors (as it specifies it SHOULD be any 5xx error code).

turt2live: It'd still be good to verify that case, especially because assuming that a new error code is fine has bitten us in the past.

erikjohnston: OOB, the ask here is to check each endpoint against a representative set of client implementations to ensure they don't explode on this new error code. It is not enough that the endpoints must already handle other 400 errors, nor enough they must handle any 5xx error.

proposals/4528-incompatible-server-error.md
MadLittleMods: We probably shouldn't be using a stable version of a new error from an MSC that isn't merged yet. If this ends up not being merged and a competing MSC arises with a different meaning, we're just creating a mess.

I don't think unstable prefixes require feature flags. Feature flags are a homeserver specific thing. In terms of how Synapse will use this error code, that's our own judgement and I think it would be perfectly fine to return the nice distinguishable unstable error.



> ftr `ORG.MATRIX.MSC4528_INCOMPATIBLE_SERVER` is the correct format for an unstable error code

Thanks for the correct format! (I guessed based on our historical mixed usage of both)

I might as well change it :fast_forward:

tulir: > I guessed based on our historical mixed usage of both

oh yeah, I noticed that in the Synapse code and found that it was actually documented wrong last year, fixed now in https://github.com/matrix-org/matrix-spec/pull/2446

erikjohnston: > I don't think unstable prefixes require feature flags. Feature flags are a homeserver specific thing. In terms of how Synapse will use this error code, that's our own judgement and I think it would be perfectly fine to return the nice distinguishable unstable error.

Synapse generally tries very hard to avoid using anything unstable without putting it behind an experimental feature flag. This is to try and ensure the ecosystem doesn't start using unstable things (like error codes) just because Synapse has the largest server share.

As @turt2live [noted](https://github.com/matrix-org/matrix-spec-proposals/pull/4528#discussion_r4047212103), we need to be careful to not start returning new error codes without testing.

The interim position should likely be returning `M_UNKNOWN`, which is not great but is acceptable.

MadLittleMods: Clients should be treating unknown error codes the same as `M_UNKNOWN` anyway. Clients being brittle enough to break and not degrade gracefully is a smell.

turt2live: We fully agree it's a smell and not great, but the historical lack of error specification in Matrix has led to clients exploding when faced with unknown error codes, even within `catch` blocks. 

Fixing the problem is harder than testing return values in advance.

PR #3389, opened 2021-09-09
MSC3389: Relation redactions

bwindels:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/bwindels/relation-redactions/proposals/3389-relation-redactions.md)

PR #4530, opened 2026-08-17
MSC4530: Do not redact relationships

bxdxnn:
[Rendered](https://github.com/bxdxnn/matrix-spec-proposals/blob/dont-redact/proposals/4530-do-not-redact-relations.md)

Implementations:
* [Complement](https://github.com/bxdxnn/complement/commit/6485f3d2d2902009d6fa5f15f98588a8d817dd84)
* [Synapse](https://github.com/bxdxnn/synapse/commit/a00f4902b99ce91839534a94460543fa4c0eb0f3)

### DIFF
--- a/proposals/4530-do-not-redact-relations.md
+++ b/proposals/4530-do-not-redact-relations.md
@@ -8,24 +8,13 @@ the client will render these events as normal redacted messages on the main time
 This leads to a lot of flood on the main timeline of the client without any context.
 
 To fix this, in the `m.relates_to` object of the `content` object of an event,
-these keys MUST NOT be stripped upon receipt of a redaction event:
+these keys (if they're a non-empty string) and the `m.relates_to` itself (if it's a non-empty object) MUST NOT be
+stripped upon receipt of a redaction event:
 * `rel_type`
 * `event_id`
 
-For [fallback](https://spec.matrix.org/v1.19/client-server-api/#fallback-for-unthreaded-clients) and context reasons,
-`m.in_reply_to` and `is_falling_back` keys MUST NOT be stripped in the `m.relates_to` object of the `content` object
-if the `rel_type` key's value is `m.thread` in the `m.relates_to` object of the `content` object.
-
 With these keys present, the client will have enough information to correctly categorize and render the redacted event,
 e.g. render the redacted event as related to a thread or an "edit revisions" list.
-
-## Alternatives
-
-### Do not strip `m.in_reply_to` unconditionally as well
-
-I believe this key should be stripped for unthreaded messages, as it doesn't affect the problem described and looks
-like the expected behavior in simple replies for privacy reasons.
-Compared to these simple replies, for threads it's important to keep all of the context.
 
 ## Security considerations
 
--- a/proposals/4530-do-not-redact-relations.md
+++ b/proposals/4530-do-not-redact-relations.md
@@ -30,3 +30,7 @@ Compared to these simple replies, for threads it's important to keep all of the 
 ## Security considerations
 
 None.
+
+## Unstable prefix
+
+The room version is `org.matrix.msc4530.12` based on room version 12.
--- a/proposals/4530-do-not-redact-relations.md
+++ b/proposals/4530-do-not-redact-relations.md
@@ -1,0 +1,32 @@
+# MSC4530: Do not redact relationships
+
+## Proposal
+
+When someone redacts an event that is inside of a thread or is an edit revision
+(i.e. `m.thread` and `m.replace` (this problem also applies to `m.annotation`) relationships),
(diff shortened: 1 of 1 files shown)


proposals/4530-do-not-redact-relations.md
turt2live, line 1: Implementation requirements:

bxdxnn: @turt2live
* [Complement](https://github.com/bxdxnn/complement/commit/6485f3d2d2902009d6fa5f15f98588a8d817dd84)
* [Synapse](https://github.com/bxdxnn/synapse/commit/a00f4902b99ce91839534a94460543fa4c0eb0f3)

proposals/4530-do-not-redact-relations.md
```diff
@@ -0,0 +1,32 @@
+# MSC4530: Do not redact relationships
```
clokep, line 1: This will need some notes on an unstable identifier.

proposals/4530-do-not-redact-relations.md
```diff
@@ -0,0 +1,36 @@
...
+This leads to a lot of flood on the main timeline of the client without any context.
+
+To fix this, in the `m.relates_to` object of the `content` object of an event,
+these keys MUST NOT be stripped upon receipt of a redaction event:
+* `rel_type`
+* `event_id`
+
+For [fallback](https://spec.matrix.org/v1.19/client-server-api/#fallback-for-unthreaded-clients) and context reasons,
+`m.in_reply_to` and `is_falling_back` keys MUST NOT be stripped in the `m.relates_to` object of the `content` object
+if the `rel_type` key's value is `m.thread` in the `m.relates_to` object of the `content` object.
```
tulir: This doesn't sound very convincing. Clients that support threads already use the relation `event_id` instead of the fallback reply, and preserving the relation isn't important for clients that don't support threads, so why preserve replies ever?

proposals/4530-do-not-redact-relations.md
```diff
@@ -0,0 +1,36 @@
...
+
+When someone redacts an event that is inside of a thread or is an edit revision
+(i.e. `m.thread` and `m.replace` (this problem also applies to `m.annotation`) relationships),
+the client will render these events as normal redacted messages on the main timeline of the room.
+This leads to a lot of flood on the main timeline of the client without any context.
+
+To fix this, in the `m.relates_to` object of the `content` object of an event,
+these keys MUST NOT be stripped upon receipt of a redaction event:
+* `rel_type`
+* `event_id`
```
tulir, line 14: MSC3389 also defines the behavior of some edge cases, like `m.relates_to` not being an object or not having rel_type/event_id. Those things should probably be defined here too.

I also wonder if the preservation should only apply when rel_type and event_id are strings 🤔

proposals/4530-do-not-redact-relations.md
bxdxnn: > so why preserve replies ever?

Because there are [replies within threads](https://spec.matrix.org/v1.19/client-server-api/#replies-within-threads) which can be rendered by clients as "nested" e.g. like on Reddit.

tulir: Replies within threads are not nested threads, they're the same as replies outside threads

proposals/4530-do-not-redact-relations.md
bxdxnn: I think this won't be much useful and should be rather defined as some default behavior in core for any bogus fields, and there's a [comment](https://github.com/matrix-org/matrix-spec-proposals/pull/3389#discussion_r1193739511) on that PR from another SCT member saying that what was described about handling empty/bogus keys in that MSC isn't the usual behavior.

tulir: Unlike `third_party_invite`, there are no auth rules for `m.relates_to`'s shape. It could be left out, though should be specified in alternatives and/or security considerations (right now if someone sends bogus relations, the event can just be redacted, which would essentially be prevented by this MSC)

Keeping empty objects might be fine, but should also be specified explicitly

proposals/4530-do-not-redact-relations.md
bxdxnn: Nested threads are forbidden in Matrix, so if hypothetically I'd like to create a client that would have nested threads by using `m.in_reply_to` to nest them (which would be technically correct and rendered normally in any other spec-compliant client) after the thread's root, I'd lose this `m.in_reply_to` information on redaction.

tulir: Personally I don't think the redaction algorithm should support those kinds of things if they're not a proper part of the spec. If a client wants nested threads, it should propose a standard way to create them instead (and that proposal could then include updates to the redaction algorithm if necessary)

PR #4534, opened 2026-08-26
MSC4534: Simple room rules

gingershaped:
[Rendered](https://github.com/gingershaped/matrix-spec-proposals/blob/ginger/msc/simple-room-rules/proposals/4534-simple-room-rules.md)

### DIFF
--- a/proposals/4534-simple-room-rules.md
+++ b/proposals/4534-simple-room-rules.md
@@ -10,25 +10,29 @@ explicitly accept them.
 
 A new state event `m.room.rules` is defined, with an empty state key and these top-level keys in `content`:
 
-| Key                  | Type                      | Purpose                                                                                                 |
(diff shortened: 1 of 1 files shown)


proposals/4534-simple-room-rules.md
turt2live, line 1: Implementation requirements:

* [ ] Client (setting rules)
* [ ] Client (showing&accepting rules)
* [ ] Client (rejecting rules)
* [ ] Client (interacting with another user's acceptance/rejection. For example, moderation tooling.)

gingershaped: I'm not sure I understand the last requirement in this list. Rule acceptance is stored in per-room account data, so only a user's own homeserver can check if they've accepted the rules.

turt2live: Moderation tooling will want to see that acceptance, so it may have to be exposed.

PR #4536, opened 2026-09-04
MSC4536: Additional endpoints for fetching user profiles

jaywink:
[Rendered](https://github.com/jaywink/matrix-spec-proposals/blob/jaywink/room-profiles-endpoint/proposals/4536-additional-endpoints-for-fetching-user-profiles.md)

This MSC aims to provide a solution for Matrix clients to fetch user profile data from Matrix servers, to aid clients in populating multiple profiles efficiently. This need may come for example to provide User Status related fields in room membership lists.

The MSC suggests two extra endpoints:
* one to paginate through room members profiles
* one to fetch a number of user profiles in one go from across several rooms

Complements, but does not depend on [MSC4429: Profile updates over legacy sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4429) and [MSC4262: Profile updates over sliding sync](https://github.com/matrix-org/matrix-spec-proposals/pull/4262).

### DIFF
--- a/proposals/4536-additional-endpoints-for-fetching-user-profiles.md
+++ b/proposals/4536-additional-endpoints-for-fetching-user-profiles.md
@@ -1,4 +1,4 @@
-# MSC0000: Additional endpoints for fetching user profiles
+# MSC4536: Additional endpoints for fetching user profiles
 
 This MSC aims to provide a solution for Matrix clients to fetch user profile data from Matrix servers.
 Currently, there are multiple ways a client may end up receiving profile data from the backend.
@@ -175,10 +175,10 @@ None foreseen at this moment.
 
 While this MSC is unstable, the endpoints are:
 
-* `GET /_matrix/client/unstable/org.matrix.mscxxxx/rooms/{roomId}/profiles`
-* `POST /_matrix/client/unstable/org.matrix.mscxxxx/profiles/query`
+* `GET /_matrix/client/unstable/org.matrix.msc4536/rooms/{roomId}/profiles`
+* `POST /_matrix/client/unstable/org.matrix.msc4536/profiles/query`
 
-Support for these endpoints MUST be advertised via the `org.matrix.mscxxxx` flag in `/_matrix/client/versions`.
+Support for these endpoints should be advertised via the `org.matrix.msc4536` flag in `/_matrix/client/versions`.
 
 ## Dependencies
 
--- a/proposals/0000-additional-endpoints-for-fetching-user-profiles.md
+++ b/proposals/0000-additional-endpoints-for-fetching-user-profiles.md
@@ -1,4 +1,4 @@
-# MSC0000: Room profiles endpoint
+# MSC0000: Additional endpoints for fetching user profiles
 
 This MSC builds on the profile updates over legacy ([MSC4429](https://github.com/matrix-org/matrix-spec-proposals/pull/4429))
 and sliding sync ([MSC4262](https://github.com/matrix-org/matrix-spec-proposals/pull/4262)) MSC's.
@@ -56,7 +56,7 @@ when operating in lazy loading mode, for example to show user status of room mem
 
 - **Endpoint**: `GET /_matrix/client/v3/rooms/{roomId}/profiles`
 - **Description**: Retrieve room member user profiles.
-- **Pagination**: *Yes TODO What mechanism?*
+- **Pagination**: *Yes. TODO what mechanism?*
 - **Parameters**:
   - `fields` - (Optional) URL encoded JSON list of fields to include in the response. If not
     given, all profile fields will be returned.
@@ -88,7 +88,7 @@ TODO: error codes
 - **Pagination**: No
 - **Request body**: Request body should contain a list of profiles, and optionally fields, to query.
 
-  * `users` - List of user IDs to query for. The maximum amount of profiles the homeserver should
+  * `users` - (Required) List of user IDs to query for. The maximum amount of profiles the homeserver should
     accept to be requested in one go should be limited to 100.
   * `fields` - (Optional) List of fields to include in the response. If not
     given, all profile fields will be returned.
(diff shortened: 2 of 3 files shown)


proposals/4536-additional-endpoints-for-fetching-user-profiles.md
```diff
@@ -0,0 +1,185 @@
...
+These endpoints would help clients efficiently fetch profiles for things like the room members list,
+when operating in lazy loading mode, for example to show user status of room members.
+
+### Client-Server API Changes
+
+#### Fetch room profiles
+
+- **Endpoint**: `GET /_matrix/client/v3/rooms/{roomId}/profiles`
+- **Description**: Retrieve room member user profiles.
+- **Pagination**: *Yes. TODO what mechanism?*
```
dbkr, line 56: This feels like the key thing here in that this is only useful if it paginates in the same order as the client displays the members which essentially means speccing that order. Given that lazy loading just loads all members at once, maybe this would be okay here too?

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
```diff
@@ -0,0 +1,185 @@
...
+
+These endpoints would help clients efficiently fetch profiles for things like the room members list,
+when operating in lazy loading mode, for example to show user status of room members.
+
+### Client-Server API Changes
+
+#### Fetch room profiles
+
+- **Endpoint**: `GET /_matrix/client/v3/rooms/{roomId}/profiles`
+- **Description**: Retrieve room member user profiles.
```
pixlwave, line 55: As the concept of a "room profile" is new, it's not immediately clear to me whether a room profile would be the room member's name & avatar combined with the profile-specific fields from their global profile or whether this would just be their global profile unchanged (and presumably lives on /rooms to help with tracking requests for /sync).

At first glance the former would be potentially awkward as we'd end up having to merge a room member, room profile and global profile (for the results from `/sync`) into the final displayable item.

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
```diff
@@ -0,0 +1,185 @@
...
+
+* Alice, on a lazy loading client, syncs
+* Alice joins a room
+* Alice does a lazy loading incremental sync
+
+In this situation, Alice would receive the full profiles of any users who had timeline events in the
+lazy loading sync response. Alice's client would not receive profile information regarding users in
+the joined room that the client didn't know about before, unless those users had events in the timeline.
+
+## Proposal
```
pixlwave, line 36: I seem to remember a discussion about the server being able to track the requested profiles so that it could automatically provide future updates via `/sync` for the client to display? Could this be included as part of the proposal, as I think it helps outline the gap we're aiming to fix.

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
turt2live, line 1: Implementation requirements:
* [ ] Client (using)
* [ ] Server

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
jaywink: Just to help me understand, is there a specific reason the pagination would need to be in the same order as the client displays the members?

> Given that lazy loading just loads all members at once, maybe this would be okay here too?

Sorry what are we referring to with "at once"? Are you referring to the `<room>/members` endpoint, to fill in members after a lazy loading response? I think that is paginated in spec at least :thinking: https://spec.matrix.org/v1.19/client-server-api/#get_matrixclientv3roomsroomidmembers

> Defaults to the current state of the room, as determined by the server.

Does that mean "return all members"?

We could do that with profiles, yes, though it feels heavy if a room has thousands of members.

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
jaywink: Sorry, I think the MSC text is potentially not clear here - I'm definitely not suggesting room scoped profiles, only an endpoint (or two!) for clients to fetch the users _global_ profile.

Users room specific display names are afaict not a part of the global profile at any point of time - they also wont be delivered down sync, and AFAICT wont be returned in the `/profile` endpoint. This MSC doesn't suggest returning them in the `<room>/profiles` endpoint either.

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
jaywink: Updates to profile fields of users who share rooms with the users are what MSC4429 (legacy sync) and MSC4262 (sliding sync) already define, and what has been implemented in Synapse. The issue, that this MSC attempt to make easier to solve, is users who didn't share a room when the profile field changed.

We could instead just send all the profiles of the users down sync when a user joins a new room (see [first alternative](https://github.com/matrix-org/matrix-spec-proposals/pull/4536/changes#diff-2e07d353493f524ff1a0f2fbc9282f65a88f11b31dcec738bc10a6f36101cd66R143)), but that feels horribly inefficient and could overload clients if a user joins a room with tens of thousands of users - even if we try to be smart and filter out profiles we _know_ the client has been sent.

Thus, it feels, the clients should be given control on how to pull the profiles it needs.

proposals/4536-additional-endpoints-for-fetching-user-profiles.md
dbkr: Well, the situation in which we'll need to show them will be when we show the member list in the right panel, so at any given time, the set of profiles is the ones that are displayed on screen, which will depend on the order they're displayed in the members panel. So, if the server returns them in whatever order then that's almost certainly not going to be the same as the sort order. It will still work for breaking up the request into several smaller ones, it will just mean that we have to still load the whole list in several requests and the profiles will appear randomly as they load (which is no worse than loading them individually).

PR #4539, opened 2026-09-10
Bullet-proof the RTC invite validity conditions

robintown:
These are some revisions to the invite validity conditions in MSC4075, which I think would be valuable to ensure that clients, no matter when they sync and what events they have observed in the past, will determine whether an invite is valid in a consistent manner.

The issue I saw with the existing wording was essentially that it made the validity of an invite depend on past events which a client may or may not have observed, depending on your definition of 'observe'. 5188e0d377d52fdbfb3e1a4e27fd255d8415e588 and 923bb1846a91ddd6dc879df050e49cee300a1bcf are the important changes.

### DIFF
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -73,7 +73,8 @@ Clients MUST send both `m.rtc.invite` and `m.rtc.decline` as sticky events as pe
 associated delivery guarantee. The sticky durations of `m.rtc.invite` events, `m.rtc.member` events
 which accept an invite, or `m.rtc.decline` events which decline an invite SHOULD NOT be smaller than
 the invite's `lifetime`. Additionally, clients MUST implement the ephemeral map algorithm as per
-[MSC4354] to construct a state-like store of invite events.
+[MSC4354] to construct a state-like store of invite events. Tracking decline events in a map isn't
+necessary because there is no need to update them after being sent.
 
 [mentions]: https://spec.matrix.org/v1.19/client-server-api/#user-and-room-mentions
 [MSC4354]: https://github.com/matrix-org/matrix-spec-proposals/pull/4354
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -316,7 +316,7 @@ In this situation, Alice's smartphone would have ignored the invite, since it wa
 the time, while her laptop would display the invite, because it cannot see Alice's previous join
 event due to it expiring halfway through.
 
-This series of events should be exceedingly rare, but as a mitigation, clients MAY choose to resend
+This series of events should be exceedingly rare, but as a mitigation, clients SHOULD resend
 `m.rtc.member` join events as soon as they receive an invite to the same session, when necessary to
 ensure that their membership will remain sticky throughout the entire lifetime of the invite.
 
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -316,8 +316,9 @@ the time, while her laptop would display the invite, because it cannot see Alice
 event due to it expiring halfway through.
 
 This series of events should be exceedingly rare, but as a mitigation, clients SHOULD resend
-`m.rtc.member` join events as soon as they receive an invite to the same session, when necessary to
-ensure that their membership will remain sticky throughout the entire lifetime of the invite.
+`m.rtc.member` join events 2 minutes before they would expire (matching the maximum lifetime of an
+`m.rtc.invite` event) at the latest, to rule out any possibility of it expiring during the lifetime
+of an invite.
 
 ## Alternatives
 
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -306,7 +306,7 @@ may later reappear, to the user's surprise, on another device:
 
 1. Alice's laptop loses connection to her homeserver
 1. Later, Alice joins a session from her smartphone
-1. Bob joins the same session and sends an invite asking the whole room to join
+1. Bob then joins the same session and sends an invite asking the whole room to join
 1. During the invite's lifetime:
     1. Alice leaves the session from her smartphone
     1. Alice's original join event expires (ceases to be sticky)
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -112,7 +112,7 @@ apply:
 - The invite is the current invite entry in the ephemeral sticky events map for the sender
   and slot and not a withdrawal (that is, an invite event whose `content` is empty except
   for `sticky_key`).
-- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the room
+- An `m.rtc.slot` event with `state_key = slot_id` and `status = "open"` exists in the state of the room
   where the invite was received.
 - The `lifetime`, as measured from `sender_ts`, has not elapsed. If `sender_ts` is more than
   20 seconds ahead of `origin_server_ts`, the `lifetime` SHOULD be measured from `origin_server_ts`
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -119,7 +119,8 @@ apply:
   instead. This limits the impact of a malicious user faking `sender_ts` to trigger long-lived
   notifications. Regardless of the basis for measuring, the remaining lifetime MUST be capped
   at 2 minutes.
-- `m.mentions`  either has `room` set to `true` or contains the current user in `user_ids`.
+- `m.mentions` either has `room` set to `true` (and the sender had a sufficient power level at the
+  time of sending to trigger a `room` notification) or contains the current user in `user_ids`.
 - The user is not already joined to the same slot via a corresponding `m.rtc.member` event.
 
 If the invite is valid, the receiving client has three options:
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -109,6 +109,7 @@ In line with the expected behaviour of sending clients that was outlined in the 
 a receiving client SHOULD only consider an invite valid as long as all of the following conditions
 apply:
 
+- The `sender` is not the same user as the recipient.
 - The invite is the current invite entry in the ephemeral sticky events map for the sender
   and slot, and is not a withdrawal (that is, an invite event whose `content` is empty except
   for `sticky_key`).
--- a/proposals/4075-rtc-notification-event.md
+++ b/proposals/4075-rtc-notification-event.md
@@ -73,8 +73,7 @@ Clients MUST send both `m.rtc.invite` and `m.rtc.decline` as sticky events as pe
 associated delivery guarantee. The sticky durations of `m.rtc.invite` events, `m.rtc.member` events
 which accept an invite, or `m.rtc.decline` events which decline an invite SHOULD NOT be smaller than
 the invite's `lifetime`. Additionally, clients MUST implement the ephemeral map algorithm as per
-[MSC4354] to construct a state-like store of invite events. Tracking decline events in a map isn't
-necessary because there is no need to update them after being sent.
+[MSC4354] to construct a state-like store of invite events and decline events.
 
 [mentions]: https://spec.matrix.org/v1.19/client-server-api/#user-and-room-mentions
 [MSC4354]: https://github.com/matrix-org/matrix-spec-proposals/pull/4354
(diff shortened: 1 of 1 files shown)


proposals/4075-rtc-notification-event.md
```diff
@@ -290,6 +297,28 @@ This might not be true for future MatrixRTC transports, however.
...
+    1. Alice's original join event expires (ceases to be sticky)
+    1. Alice's laptop reestablishes its connection and syncs the invite event
+
+In this situation, Alice's smartphone would have ignored the invite, since it was already joined at
+the time, while her laptop would display the invite, because it cannot see Alice's previous join
+event due to it expiring halfway through.
+
+This series of events should be exceedingly rare, but as a mitigation, clients MAY choose to resend
+`m.rtc.member` join events as soon as they receive an invite to the same session, when necessary to
+ensure that their membership will remain sticky throughout the entire lifetime of the invite.
```
robintown: I'm of half a mind to make this a SHOULD

proposals/4075-rtc-notification-event.md
```diff
@@ -70,10 +70,10 @@ The schema of `m.rtc.decline` is as follows:
...
 
 Clients MUST send both `m.rtc.invite` and `m.rtc.decline` as sticky events as per [MSC4354] for the
-associated delivery guarantee. The sticky duration for `m.rtc.invite` events SHOULD NOT be smaller
-than the invite's `lifetime`. The sticky duration for `m.rtc.decline`, in turn, SHOULD NOT be smaller
-than the declined invite's sticky duration. Additionally, clients MUST implement the ephemeral map
-algorithm as per [MSC4354] to construct a state-like store of both invite and decline events.
+associated delivery guarantee. The sticky durations of `m.rtc.invite` events, `m.rtc.member` events
+which accept an invite, or `m.rtc.decline` events which decline an invite SHOULD NOT be smaller than
+the invite's `lifetime`. Additionally, clients MUST implement the ephemeral map algorithm as per
+[MSC4354] to construct a state-like store of invite events.
```
robintown: Oh, right: I removed the "and decline events" from this sentence because it looked like nothing depends on looking those up in the ephemeral map anyway.

Johennes: Oh, good point. Maybe we should spell that out to make it clear we're deliberately omitting them here?

```suggestion
[MSC4354] to construct a state-like store of invite events. Tracking decline events in a map isn't
necessary because there is no need to update them after being sent.
```

proposals/4075-rtc-notification-event.md
```diff
@@ -109,26 +109,33 @@ In line with the expected behaviour of sending clients that was outlined in the
...
   at 2 minutes.
-- `m.mentions`  either has `room` set to `true` or contains the current user in `user_ids`.
-- The user is not already joined to the same slot via a corresponding `m.rtc.member` event.
-
-If the invite is valid, the receiving client has three options:
+- `m.mentions` either has `room` set to `true` (and the sender had a sufficient power level at the
+  time of sending to trigger a `room` notification) or contains the current user in `user_ids`.
+- The user does not have *any* currently sticky `m.rtc.member` events with a membership of `join`
+  in the same slot, regardless of whether the ephemeral map algorithm considers them the latest
+  state.
```
robintown: For clarity, this rule prevents Alice's laptop from ringing in the following situation:

1. Alice's laptop loses connection to her homeserver
1. Bob joins a session and sends an invite to Alice
1. During the invite's lifetime:
    1. Alice joins and then leaves the session from her smartphone
    1. Alice's laptop reestablishes its connection and syncs the invite, join, and leave events

Whereas the original "The user is not already joined to the same slot via a corresponding `m.rtc.member` event" rule would not, since Alice's laptop sees, upon syncing, that Alice is not joined to the session.

Johennes: What was your thinking with the "regardless of whether the ephemeral map algorithm considers them the latest state" part? Were you thinking of a case where a device hasn't yet seen the joining member event from another device of the same user?

robintown: I was thinking of this case: https://github.com/matrix-org/matrix-spec-proposals/pull/4539#discussion_r3979425839

proposals/4075-rtc-notification-event.md
```diff
@@ -290,6 +298,28 @@ This might not be true for future MatrixRTC transports, however.
 [MSC4028]: https://github.com/matrix-org/matrix-spec-proposals/pull/4028
 [MSC4075]: https://github.com/matrix-org/matrix-spec-proposals/pull/4075
 
+### Invites may reappear when membership expires
+
+Intuitively, an invite that is considered invalid on one device ought to stay invalid on *all* a
+user's devices for the remainder of its lifetime. However, there is an edge case in which an invite
```
Johennes: Nit: I think there might be a word missing here?

```suggestion
Intuitively, an invite that is considered invalid on one device ought to stay invalid on *all* of a
user's devices for the remainder of its lifetime. However, there is an edge case in which an invite
```

proposals/4075-rtc-notification-event.md
Johennes: Yeah, I think that would make sense.

proposals/4075-rtc-notification-event.md
```diff
@@ -290,6 +298,28 @@ This might not be true for future MatrixRTC transports, however.
...
+may later reappear, to the user's surprise, on another device:
+
+1. Alice's laptop loses connection to her homeserver
+1. Later, Alice joins a session from her smartphone
+1. Bob joins the same session and sends an invite asking the whole room to join
+1. During the invite's lifetime:
+    1. Alice leaves the session from her smartphone
+    1. Alice's original join event expires (ceases to be sticky)
+    1. Alice's laptop reestablishes its connection and syncs the invite event
```
Johennes: I think the join step from the laptop might be missing? The later steps talk of a join expiring which I think is this one?

```suggestion
1. Alice joins a session from her laptop
1. Alice's laptop loses connection to her homeserver
1. Later, Alice joins a session from her smartphone
1. Bob joins the same session and sends an invite asking the whole room to join
1. During the invite's lifetime:
    1. Alice leaves the session from her smartphone
    1. Alice's original join event expires (ceases to be sticky)
    1. Alice's laptop reestablishes its connection and syncs the invite event
```

proposals/4075-rtc-notification-event.md
robintown: 8f9c54a39beea03f523bb7bc0772f7eebf234f60

proposals/4075-rtc-notification-event.md
robintown: Mm, no, I was referring to the join event sent by Alice's smartphone

robintown: In other words the scenario is:

- Alice *is* joined at the time of the invite, from whatever device
- but whatever join there was, by the time the laptop observes the invite, has vanished

proposals/4075-rtc-notification-event.md
Johennes: Resolving this one so that we can discuss it in the other thread.

proposals/4075-rtc-notification-event.md
Johennes: Hm, so the join from the smartphone should have a sticky duration at least as large as the invite's lifetime. In order for the leave of the smartphone to replace the join in the ephemeral map, it needs to have a sticky expiry time larger than the join's sticky expiry. So when the laptop reconnects, either the leave has not yet lost its stickiness or it has lost its stickiness in which case the invite has exceeded its lifetime, too. I think in both cases, the laptop can treat the invite as invalid based on the invite event and the _latest_ state of the sticky map?

robintown: > either the leave has not yet lost its stickiness

That is what I'm imagining, yes. For reference:

Screenshot From 2026-09-10 15-45-03

I'm not sure it's correct to decide that an invite is invalid based on there being a *leave* event currently in the ephemeral map (unless… you compute something based on its sending time? hmm)

Johennes: Oh, the invite comes _after_ the join. Oh geez.

Johennes: So we already have the condition that joins need to be sent with a sticky duration that exceeds the invite lifetime. I wonder if we also need a condition for sending leaves with a sticky duration that exceeds the livetime of any invite that may have been sent in the meantime?

robintown: I don't believe we need that, the "clients SHOULD resend join events" mitigation is enough, right? Here is what that same scenario looks like if Alice's smartphone applies the mitigation:

Screenshot From 2026-09-10 16-38-06

---

…except oh no, that's not enough, because I can still construct this scenario, in which the laptop would falsely believe that because a sticky join event exists, the smartphone was still joined at the time of the invite:

Screenshot From 2026-09-10 16-55-57

To fix *that* case I believe it's enough to modify the `m.rtc.member` condition as follows:

```diff
- - The user does not have *any* currently sticky `m.rtc.member` events with a membership of `join`
-   in the same slot, regardless of whether the ephemeral map algorithm considers them the latest
-   state.
+ - For any currently sticky `m.rtc.member` event with a membership of `join` that the user has for
+   the same slot, there also exists a currently sticky `m.rtc.member` event with a membership of
+   `leave`, such that the leave event comes *after* the join event but *before* the invite event.
```

…and *then* that fixes everything? I feel like I should go write a proof :sweat_smile:

Johennes: Hm, but you cannot look that up in the ephemeral map. Leaves kick joins out of the map because they use the same key. So that would require tracking _all_ sticky events in addition to organizing them in a state-like map.

robintown: Indeed, I guess I assumed tracking all sticky events is the most basic operation an SDK would do with sticky events? ~~It's implemented in matrix-js-sdk, at least: https://github.com/matrix-org/matrix-js-sdk/blob/d807240e252edbc1cb4496f5887898fb2cae9b0c/src/models/room-sticky-events.ts#L90~~

Edit: sorry, looking at that function more closely that is not what it does! :confused: It does ultimately track everything in the private `stickyEventsMap` field though.

Johennes: Interesting. The Rust implementation I submitted today only stores the latest. 😅 

It does feel sort of contradictory to me to store the history of sticky events when we've deliberately designed it to be a replacement for state events without using room state.

In any case, I think the current variant

> - For any currently sticky `m.rtc.member` event with a membership of `join` that the user has for
>   the same slot, there also exists a currently sticky `m.rtc.member` event with a membership of
>   `leave`, such that the leave event comes *after* the join event but *before* the invite event.[^order]

is equivalent to saying

> - There is no `m.rtc.member` event with a membership of `leave` for the user and the slot that comes after the invite event.

You're trying to express that any join that may or may not have occurred before the invite has been terminated by a leave that happened after the invite. But leaves always replace joins in the map, so I think we can just look for the `leave` directly?

Johennes: And we could capture the case where the invite came after the join or the user has already accepted the invite and is still joined with something similar to what we previously had in the list:

- There is no `m.rtc.member` event with a membership of `join` for the user and slot in the map.

So in total, I think we can just work with the state-like map directly without having to track or analyze sticky event history?

robintown: I get it now, those two conditions sound like they will work! (Provided that we still define "comes after")

Sorry for all the back-and-forth :)

Johennes: No, don't be sorry at all. This stuff is super intricate and it's not like I even started to think about any of this nor see any obvious solution before you brought it up. So thanks!!

robintown: Alright, then 09b5743b28b6c29d2046652ada181ee314c7dd2d should settle this.

proposals/4075-rtc-notification-event.md
Johennes: This should be resolved alongside https://github.com/matrix-org/matrix-spec-proposals/pull/4539#discussion_r3979504448.

proposals/4075-rtc-notification-event.md
```diff
@@ -290,6 +297,29 @@ This might not be true for future MatrixRTC transports, however.
...
+
+In this situation, Alice's smartphone would have ignored the invite, since it was already joined at
+the time, while her laptop would display the invite, because it cannot see Alice's previous join
+event due to it expiring halfway through.
+
+This series of events should already be quite rare, but as a mitigation, clients SHOULD resend
+`m.rtc.member` join events 2 minutes before they would expire (matching the maximum lifetime of an
+`m.rtc.invite` event) at the latest, to rule out any possibility of it expiring during the lifetime
+of an invite.
+
```
Johennes: I think we don't need this anymore now that we've tweaked the validness conditions to prevent this case. The still sticky leave from the smartphone should make the laptop ignore the invite.

```suggestion
```

Johennes (approved):
Thanks!

PR #2774, opened 2020-09-11
MSC2774: Expose the widget ID to the widget

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/widget-id/proposals/2774-widget-id.md)

PR #4541, opened 2026-09-10
MSC4541: Widget state

Half-Shot:
A rewrite of https://github.com/matrix-org/matrix-spec-proposals/pull/4411 with more structure and more opinions.
May be depended on by https://github.com/matrix-org/matrix-spec-proposals/pull/4412

The intention is to use this as a basis for future widget specs while providing the absolute core functionality to build upon.


proposals/4541-widget-state-event.md
turt2live, line 1: Implementation requirements:

* [ ] Evidence of current state of affairs

proposals/4541-widget-state-event.md
```diff
@@ -0,0 +1,106 @@
...
+}
+```
+
+`state_key` defines a unique identifier for the widget, which can be anything.
+`name` defines the text that appears whenever a widget is described by a client, such as a heading.
+`url` defines the root document of the widget to be displayed, see [URL Templating](#url-templating) for a
+list of available template parameters.
+`avatar_url` defines the a MXC URI for the avatar to be displayed alongside the widget.
+
+### URL Templating
```
Half-Shot, line 38: @yncyrydybyl asks on MSC4411:

>What is the roomID needed for?
The Host is calling the client in the context of a room. A way to pass initial parameters to a client can be part of another MSC.

I think in this case this is essential for two reasons:

1) Without this, the spec literally boils down to "iframe this website" which feels little flimsy as a rationale to include it in the spec without any reasonable interactions with Matrix.
2) In my mind, the URL templating stuff is what starts to demonstrate the usefulness of widgets (being able to embed a document editor with a prefilled name). We can argue about what fields to include specifically, but I think at a base level this isn't a terribly complicated part of the spec unlike say the PostMessage API which has a ton more complexity than simple URL parameter replacement.

yncyrydybyl: I remember finding it difficult, as a user, to understand which parameters I needed to add and how to prefill them. It also increases the implementation complexity.

The use case I have in mind is a web-based Matrix client that acts as a widget player. It wraps the widget and handles all the Matrix complexity. It can be opened simply via a link:

https://widget-spinner.com/?widget-url=https%3A%2F%2Fmywidget.com%2Fview.html

This is already difficult to read because of the URL encoding. Adding template variables would make it even more complex and harder to understand.

Templating should therefore be covered by a separate MSC to keep this one simple.

Half-Shot: Okay, sounds like there is consensus that the URL templating is not much loved (https://github.com/matrix-org/matrix-spec-proposals/pull/4541#issuecomment-5660994119), so we can remove :+1: 

I remain a bit worried it makes this MSC feel a bit weak given we don't have any interaction points between Matrix and the embedded content. I would like to hear from the SCT on whether defining a widget in it's bare components is enough, if we expand its usefulness in other MSCs.

toger5: > Without this, the spec literally boils down to "iframe this website" which feels little flimsy as a rationale to include it in the spec without any reasonable interactions with Matrix.

I think there is a lot to like about such a simple MSC. I think there is a lot of possible discussions to be had on just this part already:
 - is it per room
 - do we need per client
 - do we want to fixate on iframe
 - does it have to live in the room state
 
I think there also should be a "as simple as possible" implementation path for clients. "iframe this website" is simple in that regard to have the most rudimentary widget support.

But you might be right that sth that (in this form) really is nothing else than a web bookmark in a matrix room that opens in an iframe looks as too little of a thing for an MSC? In the widget world however it is a super central piece..

Do you think we should merge the basic widget actions outlined in the postMessage msc into this?

Half-Shot: So something I just realised is that https://github.com/matrix-org/matrix-spec-proposals/pull/2774 is an already accepted MSC that defines a URL templating parameter for widgets....so we do need to either define how that works here, or roll back that MSC in this MSC.

>Do you think we should merge the basic widget actions outlined in the postMessage msc into this?

I think if it came down to it, perhaps. I am torn on that we do need to have a simple MSC to define the basics of widgets, but that should probably include a mechanism for how they would talk to Matrix.

Shall we sync on this externally so the working group is in agreement?

toger5: That is very surprising! (https://github.com/matrix-org/matrix-spec-proposals/pull/2774)
Aren't the widgets themselves not yet in the spec?
How did 2774 make it into the spec?
Is this already written down properly in the actual spec text?
It definitely has to get a call out. Either why we remove it or that it is using this MSC.

yncyrydybyl: > That is very surprising! ([#2774](https://github.com/matrix-org/matrix-spec-proposals/pull/2774)) Aren't the widgets themselves not yet in the spec? How did 2774 make it into the spec? Is this already written down properly in the actual spec text? It definitely has to get a call out. Either why we remove it or that it is using this MSC.

I don't think so. That's why @richvdh has [added](https://github.com/matrix-org/matrix-spec-proposals/pull/2774#event-23041583406) the [spec-pr-missing](https://github.com/matrix-org/matrix-spec-proposals/issues?q=state%3Aopen%20label%3Aspec-pr-missing).

PR #4542, opened 2026-09-15
Clarifications for MSC4140: Cancellable delayed events

Johennes:
This includes some clarification fixes for #4140.

Refer to commit messages for more details.


proposals/4140-delayed-events-futures.md
```diff
@@ -129,7 +129,7 @@ The values of both the maximum allowed delay and the maximum allowed number of s
 the values of non-negative integer valued fields named `max_delay_ms` and `max_scheduled` respectively.
 For any of these limits enforced by the server, its representative field MUST be present in the capability.
 If the server doesn't enforce one of these limits, its representative field MUST be absent from the capability.
-If the server enforces none of these limits, the capability MAY be omitted entirely instead of having an empty body.
+If the server enforces none of these limits, the capability MUST have an empty body.
```
turt2live, line 133:
```suggestion
If the server enforces none of these limits, the capability MUST have an empty body.

> **Process note**: Previously the capability MAY have been omitted if empty, but that
> would have caused problems with discovery of delayed events. Thus, the body MUST be
> empty instead.
```

turt2live (dismissed):
I'm also going to push to the branch to add other fixes. 2+ other SCT members will be asked to review this PR given the MSC has passed FCP and the changes are somewhat lengthy.

proposals/4140-delayed-events-futures.md
```diff
@@ -247,7 +247,7 @@ On success, the homeserver will respond with HTTP 200 and a JSON object containi
 - `delayed_since_ts` - Required. The timestamp (as Unix time in milliseconds) when the delayed event was scheduled or
   last restarted.
 - `content` - Required. The content of the delayed event.
-  This is the body of the original `PUT` request, not a preview of the full event after sending.
+  This is the `content` property from the body of the original `PUT` request, not a preview of the full event after sending.
```
turt2live, line 255:
```suggestion
  This is the `content` property from the body of the original `PUT` request, not a preview of the full event after sending.
  > **Process note**: The proposal had a typo where this previously placed the entire request
  > body into `content`, including the fields communicated elsewhere in this schema.
```

proposals/4140-delayed-events-futures.md
```diff
@@ -129,7 +129,11 @@ The values of both the maximum allowed delay and the maximum allowed number of s
 the values of non-negative integer valued fields named `max_delay_ms` and `max_scheduled` respectively.
 For any of these limits enforced by the server, its representative field MUST be present in the capability.
 If the server doesn't enforce one of these limits, its representative field MUST be absent from the capability.
-If the server enforces none of these limits, the capability MAY be omitted entirely instead of having an empty body.
+If the server enforces none of these limits, the capability MUST have an empty body.
+
+> **Process note**: Previously the capability MAY have been omitted if empty, but that
+> would have caused problems with discovery of delayed events. Thus, the body MUST be
+> empty instead.
```
turt2live, line 137: All this said, ChatGPT has correctly identified that the capability structure doesn't make sense:

> 2. **The capability semantics now contradict the newly mandatory limits.** #4542 currently says both the maximum delay and maximum number of scheduled events **MUST** be enforced. But the capability definition still explicitly describes servers that don't enforce one limit, and even says that a server enforcing neither limit must advertise an empty capability object. Those cases are no longer possible for a conforming implementation if both limits are mandatory. 
   Given the direction of `cff8f24`, I'd simplify the capability so that, when delayed events are supported, **both `max_delay` and `max_pending` are required** and are non-zero. `max_delay: 0` can continue to mean the feature is unavailable if that's desired. Otherwise, if unbounded implementations are actually supposed to remain valid, then the new `MUST`s need to be backed out instead. I wouldn't leave the current halfway state.

I'd agree with ChatGPT's recommendation, but interested in hearing other SCT member opinions.

Johennes: Yeah, I was actually wondering, too, if allowing unrestricted delayed events should even be supported to begin with. So +1.

clokep: I had the same thought -- if the server MUST enforce a limit, how can there be no limit? 🤔

AndrewFerr: With cff8f24 enforcing both the max delay & delayed event limits, it indeed follows that their respective fields in the capability are now mandatory.

However, nothing stops one/both of them from being 0, as there is no contradiction in having one/both of the limits being so strict as to effectively disable delayed events.

So, the only change needed is to enforce that both of the capability fields are mandatory.  That is done by 4693f36e.

---

With all this said, AIUI all that's needed to protect against resource exhaustion is to enforce _at least one_ of these limits, not strictly both of them:
- Having only a max delay limit guarantees that all delayed events will expire eventually (and time-based rate-limiting on the scheduling of delayed events prevents being able to schedule too many of them at once)
- Having a delayed event limit simply prevents the server from being overloaded with delayed events

This would be reflected in the capability by forbidding it from being empty, i.e. requiring it to have at least one field/limit set.

Then, the MUST/SHOULD/etc language relating to delayed event limits could revert to saying that either limit is optional/recommended, but that the usage of at least one of them is required.

The only reason I fret about this is to save server admins the (admittedly slight) hassle of having to pick good values for both of these limits.  But perhaps this is overthinking the issue & the safety+straightforwardness of making both limits mandatory is best.

proposals/4140-delayed-events-futures.md
```diff
@@ -66,6 +66,12 @@ The body for requests to this endpoint is a JSON object containing the following
 - `state_key` - The state key for the event to be sent, if it is to be a state event; absent otherwise.
 - `content` - Required. The content of the event to be sent.
 
+If any field of the request is set to an invalid value
+(such `delay_ms` being set to a non-positive integer),
+the homeserver will respond with HTTP 400
+and a [standard error response](https://spec.matrix.org/v1.19/client-server-api/#standard-error-response)
+with an `errocde` of `M_INVALID_PARAM`.
```
turt2live, line 73: this needs the "process note" informal convention to highlight the change

turt2live (commented):
the other changes are probably small enough where it doesn't matter as much

proposals/4140-delayed-events-futures.md
AndrewFerr: e2dd400

richvdh (changes requested):
The changes here all look plausible to me, but I would like then to be tied back to discussion threads on the MSC (or otherwise justified).

The description of the PR provides some justifications, but it only covers a subset of the commits, and it is in any case hard to cross-reference: that text should be on commit comments, not in the PR description.

Please could you rebase this PR, and:
 - collect different parts of the same change into single commits (e.g., add the "Process notes" in the same commit as the actual change)
 - Update the commit comments to provide justifications/links for the changes.

PR #4543, opened 2026-09-16
MSC4543: Sliding Sync Extensions: End-to-end encryption

erikjohnston:
[MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) (Simplified Sliding Sync) only includes core room data and omits other data, such as the end-to-end encryption fields of `/v3/sync`, from the sync response. Instead, such data is left to "extensions", which clients opt into individually via the `extensions` field of the sync request.

This MSC defines the extension for the three end-to-end encryption fields that [`/v3/sync`](https://spec.matrix.org/v1.17/client-server-api/#e2e-extensions-to-sync) carries outside any room: `device_lists`, `device_one_time_keys_count` and `device_unused_fallback_key_types`. A client uses them to keep its copy of other users' device lists current and to keep the server supplied with one-time and fallback keys. Without them, a client using sliding sync cannot tell when a device list is stale or when its keys need replenishing.

Supersedes [MSC3884](https://github.com/matrix-org/matrix-spec-proposals/pull/3884).

[Rendered](https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/msc/sliding-sync-e2ee/proposals/4543-sliding-sync-extension-e2ee.md)

---

Note: LLM was used in the very initial barebones draft and then subsequently to check grammar and consistency. The MSC has been heavily edited and reviewed since then.

----

SCT Stuff:

[FCP tickyboxes](https://github.com/matrix-org/matrix-spec-proposals/pull/4543#issuecomment-5729087074)

[MSC checklist](https://github.com/matrix-org/matrix-spec-proposals/pull/4543#issuecomment-5729094689)


erikjohnston (commented):
Implementation requirements:

- [x] Server: https://github.com/element-hq/synapse/blob/v1.161.0/synapse/handlers/sliding_sync/extensions.py#L403-L452
- [x] Client: https://github.com/matrix-org/matrix-rust-sdk/pull/1054

proposals/4543-sliding-sync-extension-e2ee.md
erikjohnston, line 1: Implementation requirements:

- [x] Server: https://github.com/element-hq/synapse/blob/v1.161.0/synapse/handlers/sliding_sync/extensions.py#L403-L452
- [x] Client: https://github.com/matrix-org/matrix-rust-sdk/pull/1054

proposals/4543-sliding-sync-extension-e2ee.md
```diff
@@ -0,0 +1,247 @@
...
+
+Per the common extension semantics of
+[MSC4508](https://github.com/matrix-org/matrix-spec-proposals/pull/4508), servers advertise support
+for this extension in `unstable_features` of
+[`/_matrix/client/versions`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientversions):
+
+- `org.matrix.msc4543` while this MSC is unstable; and
+- `org.matrix.msc4543.stable` once this MSC is accepted and the server supports the extension as
+  specified here, under the unprefixed `e2ee` key on the stable endpoint, until it advertises the
+  spec version containing this MSC.
```
anoadragon453, line 227: Please explicitly say that the value of these fields must be `true` for clients to assume the feature is enabled. As otherwise a client may just check for the field existing, even when it's set to `false` (which we commonly do in Synapse).

proposals/4543-sliding-sync-extension-e2ee.md
```diff
@@ -0,0 +1,247 @@
...
+
+Until this MSC is accepted, implementations MUST use `org.matrix.msc4543.e2ee` as the extension key
+on the stable [MSC4186](https://github.com/matrix-org/matrix-spec-proposals/pull/4186) endpoint,
+`/_matrix/client/v4/sync`. The unprefixed `e2ee` key remains in use on the unstable
+`org.matrix.simplified_msc3575` endpoint for compatibility with existing implementations.
+
+Per the common extension semantics of
+[MSC4508](https://github.com/matrix-org/matrix-spec-proposals/pull/4508), servers advertise support
+for this extension in `unstable_features` of
+[`/_matrix/client/versions`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientversions):
```
anoadragon453: Links to the spec should be updated to v1.19.

proposals/4543-sliding-sync-extension-e2ee.md
```diff
@@ -0,0 +1,247 @@
...
+- a request without `pos`, including the first request after `M_UNKNOWN_POS`; and
+- a request that enables the extension when the request that produced its `pos` did not, since the
+  client is only guaranteed changes since that `pos` and nothing earlier on the connection.
+
+In both cases the client MUST bring its tracked device lists up to date before use. A client SHOULD
+persist the `pos` up to which it has applied device list changes, as it would the `next_batch` of
+`/v3/sync`, and MAY call
+[`/keys/changes`](https://spec.matrix.org/v1.17/client-server-api/#get_matrixclientv3keyschanges)
+with that `pos` as `from` and the end of the gap as `to`. The end of the gap is the `pos` returned
+by the request without `pos` in the first case, and the `pos` the client sent in the enabling
```
anoadragon453, line 117: `/keys/changes` doesn't send down users who've only reset their cross-signing keys though. So a client may miss out on that change if they're only relying on `/keys/changes` to catch them up fully.

proposals/4543-sliding-sync-extension-e2ee.md
```diff
@@ -0,0 +1,247 @@
...
+key that no
+[`/keys/claim`](https://spec.matrix.org/v1.17/client-server-api/#post_matrixclientv3keysclaim)
+response has returned. An empty list means the device has no unused fallback key, whether because
+none was uploaded or because it has been used.
+
+Both fields describe current state. A client replaces the values it holds with the values received.
+
+#### Connections
+
+A client MAY enable the extension on more than one connection. Each connection receives
```
anoadragon453, line 165: In practice, are clients going to have more than one sliding sync connection with e2ee enabled? What's the reason for allowing it?

I worry that clients with multiple e2ee-enabled connections might end up with races when maintaining their crypto store, and it may be best not to allow them to do this.

proposals/4543-sliding-sync-extension-e2ee.md
erikjohnston: hmm hmm hmm, I think the correct thing is to make it do that given `/v3/sync` will suffer the same issue?

proposals/4543-sliding-sync-extension-e2ee.md
erikjohnston: I don't really see a use case for allowing clients to do so, but I'm a bit cautious about banning something purely to stop implementations from maybe doing silly things. The naive approach for implementation is to only have a single connection, and so if a client *is* opting into multiple connections they're either a) doing it deliberately and hopefully know what they're doing, or b) doing it by mistake. Either way I don't see much value in banning it at the spec level.

I think the same argument can be made for room lists and all extensions.

proposals/4543-sliding-sync-extension-e2ee.md
erikjohnston: Good catch fc000f784

proposals/4543-sliding-sync-extension-e2ee.md
erikjohnston: e50d808ae

erikjohnston:
@mscbot fcp merge

erikjohnston:
# MSC Checklist

This document contains a list of final checks to perform on an MSC before it is accepted. The purpose is to prevent small clarifications needing to be made to the MSC after it has already been accepted.

  - [x] Introduction
  - [x] Proposal text
  - [x] Potential issues
  - [x] Alternatives
  - [x] Unstable prefix
  - [x] Dependencies

PR #3245, opened 2021-06-15
MSC0000: Voice messages (using extensible events)

turt2live:
[Rendered](https://github.com/matrix-org/matrix-doc/blob/travis/msc/voice-messages/proposals/0000-voice-messages.md)

PR #4546, opened 2026-09-17
MSC4546: Circles

bxdxnn:
[Rendered](https://github.com/bxdxnn/matrix-spec-proposals/blob/circles/proposals/4546-circles.md)

Implementations:
* [Ruma](https://github.com/ruma/ruma/pull/2606)
* [SDK](https://github.com/matrix-org/matrix-rust-sdk/pull/7063)
* [Element X](https://github.com/element-hq/element-x-android/pull/7733)


proposals/0000-circles.md
```diff
@@ -0,0 +1,69 @@
+# MSC0000: Circles
```
turt2live: Your MSC number is 4546. Please reflect that in the title here, PR title, and file name.

bxdxnn: I know, this is a draft right now.

andybalaam:
Rendered link should be https://github.com/bxdxnn/matrix-spec-proposals/blob/circles/proposals/4546-circles.md

proposals/4546-circles.md
turt2live, line 1: Please ensure [sign-off](https://github.com/matrix-org/matrix-spec-proposals/blob/main/CONTRIBUTING.md) is provided for this MSC. Without sign-off, we cannot put this forward for a merge FCP.

turt2live (dismissed):
Early MSC requirements review - contents not reviewed in detail.

proposals/4546-circles.md
```diff
@@ -0,0 +1,69 @@
...
+a normal [`m.video`](https://spec.matrix.org/v1.16/client-server-api/#mvideo) event and fallback to their
+existing video rendering.
+
+## Proposal
+
+A Circle is an [`m.room.message`](https://spec.matrix.org/v1.16/client-server-api/#mroommessage)
+event with `msgtype` `m.video` whose `content` includes:
+
+```json
+"m.circle": true
```
deepbluev7, line 19: I know it is very bikesheddy, but could we pick a more descriptive name for this, like "video message" or so? Circles sounds a lot more like an iteration on spaces or some social network like feature.

proposals/4546-circles.md
```diff
@@ -0,0 +1,69 @@
+# MSC4546: Circles
+
+Voice messages give Matrix a fast, conversational way to send short audio without leaving the app.
+The same desire exists for video: users often want to send a brief camera clip within the app
+and play it quickly, rather than opening another app, recording it, sending, then playing it in a
+media viewer.
+
+Circles implement this. A Circle is a 1:1 video rendered as a circle on the timeline, with recording
```
deepbluev7: Why are they supposed to be limited to a 1:1 aspect ratio? Should the spec really require how they are supposed to be rendered vs what their purpose is?

proposals/4546-circles.md
bxdxnn: "Video message" isn't a good idea either for this problem:

>I believe "Circles" is the most natural way to call this feature. "Video messages" is easy to confuse with a normal video message. "Video Notes" is Telegram-specific and personally doesn't sound right to me.

from https://github.com/element-hq/element-x-android/pull/7733

proposals/4546-circles.md
bxdxnn: > Why are they supposed to be limited to a 1:1 aspect ratio?

Because the video needs to be a (small) centered square so that it will fit on the timeline and every app will be able to render it the same way regardless of the resolution.

> how they are supposed to be rendered

It's a good idea that it can be rendered as a square and so on, I'll try to make it more ambiguous.

proposals/4546-circles.md
bxdxnn: I've put the note about naming in the Alternatives section.

turt2live:
(please avoid force-pushing - it makes it harder to review. We don't mind extra commits, provided they have sign off)

bxdxnn:
> (please avoid force-pushing - it makes it harder to review. We don't mind extra commits, provided they have sign off)

I know, but that was the problem, they didn't have one.

proposals/4546-circles.md
```diff
@@ -0,0 +1,73 @@
...
+
+Clients MUST ignore `m.circle` on events that are not `m.video` messages.
+
+The recommended video resolution is 512x512. The video aspect ratio MUST be 1:1.
+Clients MAY send the video in other 1:1 resolutions.
+
+## Potential issues
+
+Senders can set `m.circle` on a video that is not 1:1.
+Receivers SHOULD still treat the event as a Circle and MAY use crop-to-fit to mitigate this.
```
Johennes: Why does it need to be 1:1 aspect ration actually?

bxdxnn: So that all apps will render the same expected content. If the aspect ratio will be random:
* the center of the video might not be something meaningful (without full context)
* apps will be downloading unnecessary data as it'll be cropped to 1:1 anyway

proposals/4546-circles.md
```diff
@@ -0,0 +1,73 @@
+# MSC4546: Circles
+
+Voice messages give Matrix a fast, conversational way to send short audio without leaving the app.
+The same desire exists for video: users often want to send a brief camera clip within the app
+and play it quickly, rather than opening another app, recording it, sending, then playing it in a
+media viewer.
```
HarHarLinks, line 6: Ok, but why isn't this an implementation detail of the client? Wouldn't you already expect clients to play videos inline/in-app when possible?

And why does it need to be circular?

proposals/4546-circles.md
bxdxnn: I've made an update specifying why it should be 1:1.

proposals/4546-circles.md
bxdxnn: > Wouldn't you already expect clients to play videos inline/in-app when possible?

I would expect apps to open it in a proper media viewer (in-app or not) full-screen, and have the UX of media players for normal videos.
Same analogy with audio: I would expect normal audio to have the UX of audio players: show cover arts, rich metadata, Android's media widget etc.

This MSC is about bringing the UX of [voice messages](https://github.com/matrix-org/matrix-spec-proposals/pull/3245) to videos.

> And why does it need to be circular?

Nowhere in the current MSC it's said that "it need to be circular", "Circle" is just a name.

